PR #30, opened 2019-03-12
add translate rule for standard_gamma

fehiepsi:
This PR adds translation rule for `standard_gamma` and also makes it faster. Now, `standard_gamma` is jittable and the speed of its jitted version is comparable to pytorch's one. The remaining issue is to avoid the condition involving `log+log`, which I don't have a solution for it yet (using `jax.cond` does not seem to solve that problem, as discussed in https://github.com/google/jax/pull/415).

@neerajprad Because this needs the master branch of JAX (in particularly, https://github.com/google/jax/pull/495), let's wait for a new release of jax before merging.

### DIFF
--- a/numpyro/distributions/util.py
+++ b/numpyro/distributions/util.py
@@ -1,5 +1,5 @@
 import jax.numpy as np
-from jax import jit, lax, random, partial
+from jax import jit, lax, random, partial, vmap
 from jax.core import Primitive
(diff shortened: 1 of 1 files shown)


fehiepsi:
Closed! I'll open a new PR because I can't push to this branch (it becomes unknown...).

PR #44, opened 2019-03-15
Add pathwise test

fehiepsi:
This adds pathwise test for `t`, `lognormal`, `gamma`, `beta` distributions. The test works well with `lognormal` and `gamma` but not good for `t` and `beta`. It seems that we have to implement `pathwise` gradient for `t` and `beta` sampler separately due to additional stochastic effects on gamma samples. Although I just compare the `mean` of actual and expected grad, student-t distribution still does not give good answer. :(


fehiepsi:
The original branch has been deleted at some points. >"< I will open another PR for this.

PR #49, opened 2019-03-17
implement warmup adapter

fehiepsi:
After going back and forth, I come up with this design of warmup adapter. I have tried to keep the interface pretty close to pyro's version. The most complicated job is to manipulating states w.r.t. if/else statements. On the other hand, testing and debugging is pretty simple with functional approach because we don't have side effects.

### DIFF
--- a/numpyro/util.py
+++ b/numpyro/util.py
@@ -1,6 +1,7 @@
 from __future__ import division
 
 import random
+from collections import namedtuple
 
 import jax.lax as lax
 import jax.numpy as np
@@ -47,7 +48,8 @@ def dual_averaging(t0=10, kappa=0.75, gamma=0.05):
         g_avg = (1 - 1 / (t + t0)) * g_avg + g / (t + t0)
         # According to formula (3.4) of [1], we have
         #     x_t = argmin{ g_avg . x + loc_t . |x - x0|^2 },
-        # where loc_t := beta_t / t, beta_t := (gamma/2) * sqrt(t)
+        # hence x_t = x0 - g_avg / (2 * loc_t),
+        # where loc_t := beta_t / t, beta_t := (gamma/2) * sqrt(t).
         x_t = prox_center - (t ** 0.5) / gamma * g_avg
         # weight for the new x_t
         weight_t = t ** (-kappa)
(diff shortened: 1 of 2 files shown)


numpyro/util.py
```diff
@@ -163,6 +165,120 @@ def _body_fn(state):
...
+        if inverse_mass_matrix is None:
+            assert mass_matrix_size is not None
+            if diag_mass:
+                inverse_mass_matrix = np.ones(mass_matrix_size)
+            else:
+                inverse_mass_matrix = np.identity(mass_matrix_size)
+        mm_state = mm_init(inverse_mass_matrix.shape[-1])
+
+        window_idx = 0
+        return step_size, inverse_mass_matrix, ss_state, mm_state, window_idx
```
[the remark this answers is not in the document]
fehiepsi: For mass matrix, it is retrieved from mm_state only at the end of each window. For step_size, an additional operator `exp` is required. I think that we can merge `(step_size, ss_state)` and `(inverse_mm, mm_state)` but given that `ss_state`, `mm_state` are not used in the main hmc sampler, separating like this both simplifies this function implementation and hmc sample function. What do you think?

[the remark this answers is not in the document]
neerajprad: That sounds reasonable!

numpyro/util.py
```diff
@@ -163,6 +165,120 @@ def _body_fn(state):
...
+        if adapt_mass_matrix:
+            mm_state = lax.cond(is_middle_window,
+                                (z_flat, mm_state), lambda args: mm_update(*args),
+                                mm_state, lambda x: x)
+
+        t_at_window_end = t == adaptation_schedule[window_idx, 1]
+        window_idx = np.where(t_at_window_end, window_idx + 1, window_idx)
+        state = step_size, inverse_mass_matrix, ss_state, mm_state, window_idx
+        # TODO: enable lax.cond when https://github.com/google/jax/issues/514 is resolved
+        # state = lax.cond(t_at_window_end & is_middle_window,
```
neerajprad: If you didn't pass in `inverse_mass_matrix` and `step_size` i.e. the `_update_at_window_end` only took 2 arguments (I know that wouldn't work when mass matrix or step size adaptation is disabled), would the test pass with jit?

fehiepsi: @neerajprad I just test it. JIT works well given that prox_center of `ss_init` is a constant (not depends on current `ss_state`).

fehiepsi: @neerajprad FYI, the fix is available in JAX master branch and the test with all states passes.

Issue #29, opened 2019-03-11
Checklist for first release


neerajprad:
- [x] Add sphinx to generate documentation.
 - [x] Add docstring to existing modules.
   - [x] Revisit #60 @fehiepsi
 - [x] Update README - what is currently supported, some small examples, and future plans.
   - [ ] Add pypi version badge to readme
   - [x] Add readthedocs badge to readme
 - [x] Include examples in unit-test to ensure compatibility with release.
 - [x] A couple of good notebooks with explanations (not for benchmarking).
    - [x] local global trend. @fehiepsi 
    - [x] bayesian regression. @neerajprad  
    - [x] bayesian NNs. @martinjankowiak who has volunteered to contribute a Bayesian NN tutorial! Not a release blocker though, we can publish it whenever it is ready. 
 - [x] Clean up minor API issues - #144
 - [x] Register docs on readthedocs (or publish on github.io or somewhere public)
    https://numpyro.readthedocs.io/en/latest/distributions.html

Already Done:
 - [x] A few starter issues for external contributors.
 - More interesting examples:
    - [x] SVI - maybe a VAE or SS-VAE example with MNIST.
    - [x] HMC - baseball, hierarchical modeling, bayesian regression.
 - [x] Better coverage for popular distributions - #1 
 - [x] Working HMC, NUTS - doesn't have to be a fully optimized implementation.
 - [x] Benchmarking HMC, NUTS vis-a-vis Pyro, Stan, PyMC3

jpchen:
>More interesting examples

i have a gmm example (forked from the pyro ex) as soon as we have a categorical [edit: i see there's now a multinomial, which i can probably use

neerajprad:
@fehiepsi - After discussing with @fritzo, I think it will be a good idea to move this to pyro-ppl, and develop in public. We should definitely look to address the issues above, and have an initial release at that point. What do you think?

fehiepsi:
@jpchen If you need Dirichlet distribution, please let me know, I will implement it first. :)

neerajprad:
> i have a gmm example (forked from the pyro ex) as soon as we have a categorical 

That's great! I think you should be able to use the multinomial directly, but let me know if you get into any issues.

fehiepsi:
@neerajprad Sure, let's go for it!

fritzo:
[...]

PR #53, opened 2019-03-20
Adding a simple VAE Example

neerajprad:
Addresses #29.

TODOs:
 - [x] binarize MNIST images instead of using the hack of disabling args validation and relying on `Bernoulli.logpmf`'s BCE implementation. 
 - [x] verify that the model is indeed learning something useful - ~the loss seems to remain roughly the same across epochs.~

Later PR:
 - [ ] Remove or figure out a better way to include the `args_check` flag, which is really hard to exercise with our current distribution class organization. Right now this is a global flag, but its really unintuitive since the flag might be off when `distribution.logpmf` method is called. If we have this flag, we should follow the pattern that we have in PyTorch currently. - #56.


neerajprad:
This should be ready for a review now.

numpyro/examples/vae.py
```diff
@@ -0,0 +1,152 @@
...
+    @jit
+    def epoch_train(opt_state, rng):
+        def body_fn(i, val):
+            loss_sum, opt_state, rng = val
+            rng, batch = binarize(rng, train_fetch(i, train_idx)[0])
+            loss, opt_state, rng = svi_update(i, opt_state, rng, (batch,), (batch,),)
+            loss_sum += loss
+            return loss_sum, opt_state, rng
+
+        return lax.fori_loop(0, num_train, body_fn, (0., opt_state, rng))
```
neerajprad: Even here jitting the for loop brings down the runtime to 25% of just jitting `body_fn`. Thanks for the tip, @fehiepsi.

fehiepsi:
Nice! I'll go over this tomorrow.

numpyro/svi.py
```diff
@@ -35,7 +39,12 @@ def update_fn(i, opt_state, rng, model_args=(), guide_args=()):
         rng, = random.split(rng, 1)
         return loss_val, opt_state, rng
 
-    return init_fn, update_fn
+    def evaluate(opt_state, rng, model_args=(), guide_args=()):
```
fehiepsi: +1

numpyro/examples/vae.py
```diff
@@ -0,0 +1,152 @@
...
+        img_loc = decode(params['decoder'], z).reshape([28, 28])
+        plt.imsave(os.path.join(RESULTS_DIR, 'recons_epoch={}.png'.format(epoch)), img_loc, cmap='gray')
+
+    for i in range(args.num_epochs):
+        t_start = time.time()
+        num_train, train_idx = train_init()
+        _, opt_state, rng = epoch_train(opt_state, rng)
+        rng, = random.split(rng, 1)
+        num_test, test_idx = test_init()
+        test_loss = eval_test(opt_state, rng)
```
fehiepsi: nit: I think it is better to split rng into 2 (1 to train and update rng, 1 for eval) to avoid a confusion that: rng here is used for eval and the train of next epoch.

neerajprad: Done.

PR #54, opened 2019-03-20
Implement build_tree function

fehiepsi:
This PR implements build_tree function which is used to create binary tree for NUTS sampler.

TODO list:

- [x] Implement build_tree while loop
- [x] Revise docs/comments
- [x] Add a proper test


fehiepsi:
@neerajprad This is ready to review. Please let me know if more tests should be added or we will verify the correctness when nuts sampler is implemented. I have tried to jit as much as possible (if we only jit the build_basetree function, for some reasons my implementation is very slow; I guess because I used `lax.cond` here and there) and did some benchmarks [here](https://gist.github.com/fehiepsi/453c0e8ed6a24739d6295feecbeb8635). It took 1s to build full tree (~ 1 nuts step) so it is slow to me. :(

fehiepsi:
Opps. For some reason, lax.cond is not available yet.

neerajprad:
> I guess because I used lax.cond here and there) and did some benchmarks here. It took 1s to build full tree (~ 1 nuts step) so it is slow to me. :(

Hmm..1 s. seems quite slow. I'll play around with your code, and comment.

fehiepsi:
@neerajprad I just update the [gist](https://nbviewer.jupyter.org/gist/fehiepsi/453c0e8ed6a24739d6295feecbeb8635). Quote of takeaways:

+ For HMC, it is best to jit each sample call. The overhead 360microsecond is small. Assume that we want to generate 1000 samples, the overhead is expected to be less than 1s. So most of time will be spent for verlet update, which is exactly what we want. To reduce compling time for generating the first trajectory, it is better to use lax.fori_loop instead of native python loop. Anyway, it seems that JAX will works pretty well with HMC.

+ For NUTS, we couldn't jit each sample call yet. If we only jit the verlet update, NUTS is 10x slower than HMC. But doing so in HMC has already involved a lot of overhead (in g1, it took 300ms for 1000 verlet steps, hence 300s to get 1000 samples). So improving NUTS overhead over HMC might be beneficial but not be a good solution for a long term. The best way is still to jit each NUTS sample call.

Right now, the way to jit NUTS `sample` is to resolve the recursive method of `build_subtree`. In the mean time, I will try to convert it into an iterative program to achieve it.

If you have other ideas, please let me know. :)

neerajprad:
Thanks for profiling, and digging into this. This is super interesting (though not exactly what we were expecting)! This probably might also why just JITing the PE computation in PyTorch just gives us a 4X or so improvement, the real benefit come from JITing the entire loop, which makes NUTS more complicated.

I think rather than spending time casting NUTS into a for/while loop, it might be worth exploring with the JAX folks what machinery we might need in JAX to get this to work.

fehiepsi:
Actually, I think that making build_tree iterative will not be so complicated. I can come up with an implementation soon. Here is a sketch in the mean time:
+ Assume that we already choose a direction and want to build a subtree with depth=3 (8 nodes). We will iteratively apply 8 verlet steps but will stop when diverging or turning happens (this is just a while_loop with condition).
+ The diverging condition is straightforward. The trickiest part is to check if the subtree is turning or sub of subtree is turning,... For example, at nodes 3, we check the turning conditions for the following trees: 0->3, 2->2; at the nodes 7, we check the turning conditions for the following trees: 0->7, 4->7, 6->7. The first case requires us store informations at nodes 0, 2. The second case requires us store informations at nodes 0, 4, 6. The information which we need to store will dynamically change when we move:
  - What to store? It is enough to store `r` and `r_sum` to check for turning condition.
  - What is the trade-off? Memory requirement will be strictly O(d) where d is the depth, even when we stop at early stage. To me, this is not an issue given that max_tree_depth~10 (way small comparing to the number of samples which we need).
  - How to check for turning condition? We create two arrays (1 for `r`, 1 for `r_sum`) R and R_sum with shape `3 x mass_matrix_size`. These tables will be updated during verlet steps:
    * Step 0: R[0, :] = r_0
    * Step 1: check turning condition of R[0, :] and r_1
    * Step 2: R[1:, :] = r_2
    * Step 3: check turning condition of R[0, :] and r_3, R[1, :] and r_3
    * Step 4: R[1:, :] = r_4 (**important**: update at index 1)
    * Step 5: check turning condition of R[1, :] and r_5 (not R[0, :] and r_5)
    * Step 6: R[2:, :] = r_6
    * Step 7: check turning condition of R[0, :] and r_7, R[1, :] and r_7, R[2, :] and r_7
  - Note that at step 5, we can check the turning condition of R[0, :] and r_5, and if it happens, we can stop there. This is unable to check using recursive method but is doable with the above iterative scheme.

neerajprad:
> Assume that we already choose a direction and want to build a subtree with depth=3 (8 nodes). We will iteratively apply 8 verlet steps but will stop when diverging or turning happens (this is just a while_loop with condition).

I might be misunderstanding something, but isn't the point to choose the direction at random for each subtree? How would this work in that case?

neerajprad:
I think your algorithm should work great, and doesn't actually seem that much more complicated! Regarding the checks, we can simply populate a table for a tree depth of 10 (there might be analytical expression available too, but if not):
 - Starting from a given node, go up the tree to the root node. 
 - If the current node is the left child of the parent node, then we only need to check with itself. 
 - Otherwise, until the current node is the right child of the parent, find the leftmost node from the parent node and add it to the list. e.g. from 5, we go up left and gather 4 from the parent. From 7, we gather 6, 4, and 0 at each of the parents.

fehiepsi:
Thanks @neerajprad ! There is an analytic solution based on binary representation. For example, max_idx is the number of non-zero element, and max_idx - min_idx is the number of last non-zero elements (e.g. 1000110111 -> 3).

numpyro/util.py
```diff
@@ -270,14 +275,362 @@ def update_fn(t, accept_prob, z_flat, state):
...
+    # TODO: we can create these checkpoints at build_tree method
+    # and reuse it; but let's do this optimization later
+    r_checkpoints = np.zeros((max_tree_depth, inverse_mass_matrix.shape[-1]),
+                             dtype=inverse_mass_matrix.dtype)
+    r_checkpoints = index_update(r_checkpoints, 0, r_init)
+    r_sum_checkpoints = np.zeros((max_tree_depth, inverse_mass_matrix.shape[-1]),
+                                 dtype=inverse_mass_matrix.dtype)
+    r_sum_checkpoints = index_update(r_sum_checkpoints, 0, r_init)
+
+    # FIXME: use native while here is fine
```
fehiepsi: iterative program works correctly now; the only thing which blocks us from jitting NUTS is this while_loop; not sure why we get **RuntimeError: Invalid argument: The parameter of condition and body, the result of the body, and init must all have the same shape; got Condition: ...** with JAX here while the native python `while` loop works. will investigate more

fehiepsi:
@neerajprad I have updated the [gist](https://gist.github.com/fehiepsi/453c0e8ed6a24739d6295feecbeb8635) which shows that we just have ~500 microsecond overhead with iterative NUTS!! The notebook also shows that we get the same result with non-iterative version (though `rng` scheme might be different, I have thought that the results will be a bit different, but the proposals are identical so I think that they might be overlap somehow or accept_prob is too large/small; anyway it might be not important). :D

Because iterative NUTS is not appeared anywhere else, I think that it is better to write a note in wiki to help other developers understand what is going on under the scene (rather than writing a long explanation in code comments). It is also a good chance to explain some functions like `_leaf_idx_to_ckpt_idx`, which I used to get indices of checkpoint tables. I can do that after we finish our alpha release targets. For now, I will greatly appreciate your comments on language choices, design choices, or optimization if any. I am not happy with many terminologies such as `going_right`, `biased_transition_prob`,... or the order of args in many functions. This PR is lengthy so please take your time (in addition, this is still blocked by `lax.cond` issue).

Somehow the recursive method is 5x slower than a few commits before. I'll investigate why in the mean time.

neerajprad:
>> I have updated the gist which shows that we just have ~500 microsecond overhead with iterative NUTS!! 

Wow..that's awesome! I'll do a detailed pass soon.

numpyro/util.py
```diff
@@ -270,14 +275,369 @@ def update_fn(t, accept_prob, z_flat, state):
...
+                         use_multinomial_sampling, get_transition_prob, iterative_build=False)
+
+
+def _leaf_idx_to_ckpt_idx(n):
+    # computes the number of non-zero bits except the last bit
+    # e.g. 6 -> 2, 7 -> 2, 13 -> 2
+    _, idx_max = lax.while_loop(lambda nc: nc[0] > 0,
+                                lambda nc: (nc[0] >> 1, nc[1] + (nc[0] & 1)),
+                                (n >> 1, 0))
+    # computes the number of last non-zero bits
```
neerajprad: nit: maybe add `number of contiguous last non-zero bits`.

numpyro/util.py
```diff
@@ -270,14 +275,369 @@ def update_fn(t, accept_prob, z_flat, state):
...
+    def _body_fn(state):
+        i, _ = state
+        subtree_r_sum = r_sum - r_sum_ckpts[i] + r_ckpts[i]
+        # XXX we might not need to unravel here
+        return i - 1, _is_turning(inverse_mass_matrix, r_ckpts[i], r, subtree_r_sum)
+
+    _, turning = lax.while_loop(lambda it: (it[0] >= ckpt_idx_min) & ~it[1],
+                                _body_fn,
+                                (ckpt_idx_max, False))
+    return turning, r_ckpts, r_sum_ckpts
```
neerajprad: Nice! It will be great to have a simple test for this.

numpyro/util.py
```diff
@@ -270,14 +275,369 @@ def update_fn(t, accept_prob, z_flat, state):
...
+    # computes the number of last non-zero bits
+    # e.g. 6 -> 0, 7 -> 3, 13 -> 1
+    _, num_subtrees = lax.while_loop(lambda nc: (nc[0] & 1) != 0,
+                                     lambda nc: (nc[0] >> 1, nc[1] + 1),
+                                     (n, 0))
+    idx_min = idx_max - num_subtrees + 1
+    return idx_min, idx_max
+
+
+def _is_iterative_turning(leaf_idx, inverse_mass_matrix, r, r_sum, r_ckpts, r_sum_ckpts):
```
neerajprad: If I understand correctly, this is doing two things:
 - updating checkpoints when `leaf_idx` is even.
 - checking for turning condition. I think we should only have to do this when `leaf_idx` is odd, right?

If so, it might be better to separate it out into two conditions `_update_checkpoints` and `_is_iterative_turning` which can be exercised by `lax.cond` based on whether the index is even or odd. Otherwise, we should probably rename this function appropriately.

numpyro/util.py
```diff
@@ -270,14 +275,369 @@ def update_fn(t, accept_prob, z_flat, state):
...
+                             use_multinomial_sampling, max_tree_depth):
+    max_num_proposals = 2 ** depth
+
+    def _cond_fn(state):
+        tree, turning, _, _, _ = state
+        return (tree.num_proposals < max_num_proposals) & ~turning & ~tree.diverging
+
+    def _body_fn(state):
+        current_tree, _, r_checkpoints, r_sum_checkpoints, rng = state
+        rng, transition_rng = random.split(rng)
```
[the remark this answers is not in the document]
fehiepsi: Yes. 2 is the default parameter.

numpyro/util.py
fehiepsi: Yes, I'll add it. How about we use jax master version for a while before starting pinning to a specific version (e.g. next jax release)?

neerajprad:
This is looking great. I'll keep taking more passes at it. If its blocking us, I think we can build the wheels for jax and jaxlib and store it for CI, so that we aren't waiting for jax's next release.

fehiepsi:
> If its blocking us, I think we can build the wheels for jax and jaxlib and store it for CI, so that we aren't waiting for jax's next release.

Please go for it!

numpyro/util.py
neerajprad: I think at least for CI, it will be good to pin some version from master and host it on our end, but for local development, we can just use jax master. I think we shouldn't block ourselves on the next jax release.

numpyro/util.py
fehiepsi: You are correct. When `leaf_idx` is odd, `ckpt_min > ckpt_max` so turning condition won't be updated.

The only thing which blocks me from separating is we have to recalculate `idx_min`, `idx_max`. I'll we move this calculation to the iterative body function and add these idxs to args of the methods `_update_checkpoints` and `_is_iterative_turning`.

numpyro/util.py
```diff
@@ -270,14 +275,369 @@ def update_fn(t, accept_prob, z_flat, state):
...
+    # update depth and turning condition
+    return _TreeInfo(tree.z_left, tree.r_left, tree.z_left_grad,
+                     tree.z_right, tree.r_right, tree.z_right_grad,
+                     tree.z_proposal, tree.z_proposal_pe, tree.z_proposal_grad,
+                     depth, tree.weight, tree.r_sum, turning, tree.diverging,
+                     tree.sum_accept_probs, tree.num_proposals)
+
+
+def build_tree(verlet_update, kinetic_fn, verlet_state, inverse_mass_matrix, step_size, rng,
+               max_sliced_energy=1000., use_multinomial_sampling=True, max_tree_depth=10,
```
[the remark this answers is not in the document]
fehiepsi: That will simplify our code a lot. It's has been a better strategy according to our pyro tests, and in the paper "A  Conceptual  Introduction  toHamiltonian  Monte  Carlo" too. I'll go for it.

fehiepsi:
> I'll keep taking more passes at it.

Please wait for me to address your above comments first. I expect the code will be more simplified so reviewing will be easier for you.

test/test_hmc_util.py
```diff
@@ -4,6 +4,11 @@
 
 import numpy as onp
 import pytest
+<<<<<<< HEAD:test/test_util.py
```
neerajprad: This wasn't merged cleanly.

numpyro/hmc_util.py
```diff
@@ -268,3 +271,321 @@ def update_fn(t, accept_prob, z_flat, state):
...
+        lambda trees: (trees[0].z_left, trees[0].r_left,
+                       trees[0].z_left_grad, trees[1].z_right,
+                       trees[1].r_right, trees[1].z_right_grad)
+    )
+
+    transition_prob = lax.cond(biased_transition,
+                               (current_tree, new_tree, rng),
+                               lambda args: _biased_transition_prob(*args),
+                               (current_tree, new_tree, rng),
+                               lambda args: _uniform_transition_prob(*args))
```
fehiepsi: @neerajprad  I make `biased_transition` becomes a dynamic variable instead of a static fn `_biased_transition_prob`/`_uniform_transition_prob`. This saves a bit of time after compiling.

numpyro/hmc_util.py
```diff
@@ -268,3 +271,321 @@ def update_fn(t, accept_prob, z_flat, state):
...
+    # TODO: we can create these checkpoints at build_tree method
+    # and reuse it; but let's do this optimization later
+    r_checkpoints = np.zeros((max_tree_depth, inverse_mass_matrix.shape[-1]),
+                             dtype=inverse_mass_matrix.dtype)
+    r_checkpoints = index_update(r_checkpoints, 0, r_init)
+    r_sum_checkpoints = np.zeros((max_tree_depth, inverse_mass_matrix.shape[-1]),
+                                 dtype=inverse_mass_matrix.dtype)
+    r_sum_checkpoints = index_update(r_sum_checkpoints, 0, r_init)
+
+    tree, turning, _, _, _ = lax.while_loop(
```
neerajprad: Would `turning` always be true, since we are checking for this in `_cond_fn`? Does this imply that subtree might not be balanced?

neerajprad: It does however seem to me that `_double_tree` could return a non-balanced tree when the new tree begins to turn, for instance. Can you confirm if that could happen?

numpyro/hmc_util.py
```diff
@@ -268,3 +271,321 @@ def update_fn(t, accept_prob, z_flat, state):
...
+
+    def _cond_fn(state):
+        tree, _ = state
+        return (tree.depth < max_tree_depth) & ~tree.turning & ~tree.diverging
+
+    def _body_fn(state):
+        tree, key = state
+        key, direction_key, doubling_key = random.split(key, 3)
+        going_right = random.bernoulli(direction_key)
+        biased_transition = True
```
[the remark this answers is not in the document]
fehiepsi: This argument is still necessary for recursive algorithm (to build subtrees). When we remove recursive implementation, we can move this into `_double_tree` (and remove unnecessary arguments such as `depth` or functions such as `build_subtree`,...)

numpyro/hmc_util.py
fehiepsi: Yes, that's how NUTS works: stop doubling (of subtrees) when subtrees are turning. You can take a look at Figure 2 of NUTS paper to see an example for such unbalanced tree.

neerajprad: Sorry, I didn't mean that this is a problem per se, but IIRC the tree that is finally returned is always balanced. At what point do we snip off the turning sub tree and return the non turning segment?

neerajprad: Okay clarified offline - this doesn't make any difference, since the candidate set will be balanced as per this [line](https://github.com/pyro-ppl/numpyro/pull/54/files#diff-e33a6b3cbaf1b949c73264e0984d9a76R325).

numpyro/hmc_util.py
```diff
@@ -268,3 +271,321 @@ def update_fn(t, accept_prob, z_flat, state):
...
+        v_left = np.multiply(inverse_mass_matrix, r_left)
+        v_right = np.multiply(inverse_mass_matrix, r_right)
+
+    # This implements dynamic termination criterion (ref [2], section A.4.2).
+    turning_at_left = np.dot(v_left, r_sum - r_left) <= 0
+    turning_at_right = np.dot(v_right, r_sum - r_right) <= 0
+    return turning_at_left | turning_at_right
+
+
+def _uniform_transition_prob(current_subtree, new_subtree, rng):
```
neerajprad: You don't need the `rng` here and below.

fehiepsi: nit: rename `prob` to `kernel` to match the notion is NUTS paper, rename `subtree -> tree` too, and drop `rng`.

test/test_hmc_util.py
```diff
@@ -297,3 +300,70 @@ def test_warmup_adapter(jitted):
...
+    def potential_fn(q):
+        return 0.5 * q ** 2
+
+    vv_init, vv_update = velocity_verlet(potential_fn, kinetic_fn)
+    vv_state = vv_init(0.0, 1.0)
+    inverse_mass_matrix = np.array([1.])
+    rng = random.PRNGKey(0)
+
+    def f(vv_state):
+        tree = build_tree(vv_update, kinetic_fn, vv_state, inverse_mass_matrix, step_size, rng)
```
neerajprad: Pass in `iterative_build` param to `build_tree`?

test/test_hmc_util.py
```diff
@@ -297,3 +300,70 @@ def test_warmup_adapter(jitted):
...
+    r_sum_ckpts = r_ckpts + 1
+
+    actual_turning = _is_iterative_turning(inverse_mass_matrix, r, r_sum, r_ckpts, r_sum_ckpts,
+                                           *ckpt_idxs)
+    assert expected_turning == actual_turning
+
+
+@pytest.mark.parametrize('step_size', [0.01, 1., 100.])
+@pytest.mark.parametrize('iterative_build', [True, False])
+def test_build_tree(step_size, iterative_build):
```
neerajprad: I think it will be great to have a test that `iterative_build = True, False` return the same tree for different models (I think with the same rng they would), which can be deleted once we make this the default. But in the meantime, it will be nice to use this code path for testing.

neerajprad:
Maybe the tests should pass now that #68 is merged.

neerajprad:
@fehiepsi - The tests seem to be still violating the thresholds (by a small margin) for some reason. Maybe relax the precision a bit?

fehiepsi:
@neerajprad I'll merge master, then it should pass. :)

Issue #42, opened 2019-03-14
Make basic datasets available for examples


neerajprad:
- [x] mnist
 - [x] baseball

PR #55, opened 2019-03-21
Add data loader for MNIST

neerajprad:
Addresses #42.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,5 +15,8 @@ __pycache__/
 # test related
 .pytest_cache
 
+# data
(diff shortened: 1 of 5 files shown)


fehiepsi:
@neerajprad looking at https://github.com/google/jax/blob/master/.travis.yml, it seems that we dont need libstdc++. It might be that including `dist: xenial` is enough.

neerajprad:
> it seems that we dont need libstdc++. It might be that including dist: xenial is enough. 

Thanks for the pointer! I think it should work now.

PR #59, opened 2019-03-24
separate jax from other third parties

fehiepsi:
This PR mainly separates jax from other third parties and make the import consistence. During the way, there are several small nits addressed.

### DIFF
--- a/numpyro/distributions/patch.py
+++ b/numpyro/distributions/patch.py
@@ -1,4 +1,5 @@
 import scipy
+import scipy.stats
 
 
 def patch_dependency(target, root_module):
--- a/numpyro/distributions/util.py
+++ b/numpyro/distributions/util.py
@@ -1,11 +1,13 @@
 from contextlib import contextmanager
 
+import scipy.special as osp_special
+
 import jax.numpy as np
-import scipy.special as sp
-from jax import jit, lax, partial, random
+from jax import jit, lax, random
 from jax.core import Primitive
 from jax.interpreters import ad, partial_eval, xla
 from jax.numpy.lax_numpy import _promote_args_like, _promote_shapes
+from jax.util import partial
 
 from numpyro.distributions.distribution import jax_discrete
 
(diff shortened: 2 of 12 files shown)


numpyro/distributions/patch.py
```diff
@@ -1,4 +1,5 @@
 import scipy
+import scipy.stats
```
fehiepsi: this import is used to avoid the error "AttributeError: module 'scipy' has no attribute 'stats'" which happens in case
```
import scipy
scipy.stats.uniform.pdf(0.5)  <-- error
```
In particular, we patch `scipy.stats._distn_infrastructure.rv_frozen.__call__`, so we have to import scipy.stats

numpyro/distributions/util.py
```diff
@@ -1,11 +1,13 @@
...
+import scipy.special as osp_special
+
 import jax.numpy as np
-import scipy.special as sp
-from jax import jit, lax, partial, random
+from jax import jit, lax, random
 from jax.core import Primitive
 from jax.interpreters import ad, partial_eval, xla
 from jax.numpy.lax_numpy import _promote_args_like, _promote_shapes
+from jax.util import partial
```
fehiepsi: `partial` is not an exposed api of jax, so it is better to import from its origin file

numpyro/distributions/util.py
```diff
@@ -1,11 +1,13 @@
 from contextlib import contextmanager
 
+import scipy.special as osp_special
```
fehiepsi: for consistence with jax

setup.cfg
```diff
@@ -6,8 +6,10 @@ exclude = docs/src, build, dist, .ipynb_checkpoints
 line_length = 120
 not_skip = __init__.py
 skip_glob = .ipynb_checkpoints
-known_first_party = numpyro, tests
-known_third_party = opt_einsum, six, torch, torchvision
+known_first_party = numpyro, test
+known_third_party = opt_einsum
+known_jax = jax
+sections = FUTURE, STDLIB, THIRDPARTY, JAX, FIRSTPARTY, LOCALFOLDER
```
neerajprad: Nice!

PR #60, opened 2019-03-25
a pass of hmc docs and comments

fehiepsi:
- [x] go for a pass of docs and comments


neerajprad:
Feel free to add docs later, so that we can merge this soon and avoid further merge conflicts on PRs that use this.

neerajprad:
@fehiepsi - Most of these functions are internal ones that users won't need to interact with, so we do not need to expose them in readthedocs. The docstrings will still be useful for users to understand the code and change the internals if needed, but this is not a release blocker, so take your time.

fehiepsi:
Sure! :D

notebooks/time_series_forecasting.ipynb
```diff
@@ -118,27 +118,29 @@
...
     "\n",
-    "$$\n",
     "\\begin{equation*}\n",
+    "\\begin{gathered}\n",
     "\\text{exp_val}_{t} = \\text{level}_{t-1} + \\text{coef_trend} \\times \\text{level}_{t-1}^{\\text{pow_trend}}\n",
     "+ \\text{s}_t \\times \\text{level}_{t-1}^{\\text{pow_season}}, \\\\\n",
     "\\sigma_{t} = \\sigma \\times \\text{exp_val}_{t}^{\\text{powx}} + \\text{offset}, \\\\\n",
-    "y_{t} \\sim \\text{StudentT}(\\nu, \\text{exp_val}_{t}, \\sigma_{t}), \\\\\n",
+    "y_{t} \\sim \\text{StudentT}(\\nu, \\text{exp_val}_{t}, \\sigma_{t}),\n",
+    "\\end{gathered}\n",
```
fehiepsi: This is to make latex rendered correctly in github.

numpyro/mcmc.py
```diff
@@ -103,7 +107,7 @@ def hmc(potential_fn, kinetic_fn=None, algo='NUTS'):
         euclidean kinetic energy.
     :param str algo: Whether to run ``HMC`` with fixed number of steps or ``NUTS``
         with adaptive path length. Default is ``NUTS``.
-    :return init_kernel, sample_kernel: Returns a tuple of callables, the first
+    :returns: a tuple of callables (`init_kernel`, `sample_kernel`): the first
```
neerajprad: This will not get rendered correctly. You will need to enter the return types separately via `:rtype:`. Could you also ensure that we are using `:returns:` instead of `:return` or vice-versa? I think we have a mix of both.

fehiepsi: Thanks, I'll grep and make changes for `:return:` -> `:returns:` in this PR too.

PR #64, opened 2019-03-25
Use JAX master for CI

### LINKED ISSUE #63: Use jax master for CI
Lets host the jax/jaxlib wheels for CI so that we can track JAX master directly for development.

neerajprad:
Fixes #63. Note that I'm using the pypi released jaxlib, so this won't work if JAX master requires a newer build of jaxlib, which we'll need to build and host at our end.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,8 @@ dist: xenial
 
 install:
     - pip install -U pip
+    - pip install jaxlib
+    - pip install git+https://github.com/google/jax.jit@850e8a756a78388605745d241a7f25daa371a23b
     - pip install .[test]
     - pip freeze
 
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -14,6 +14,8 @@ dist: xenial
 
 install:
     - pip install -U pip
+    - pip install jaxlib
+    - pip install -e git+git://github.com/google/jax.git@850e8a756a78388605745d241a7f25daa371a23b
```
fehiepsi: git + https I guess

fehiepsi: you are right, we need an additional egg info.

neerajprad:
@fehiepsi - This should unblock us for now, but we might need to build jaxlib ourselves if the dependency on jax master changes.

PR #67, opened 2019-03-26
Add translation rule for gamma

fehiepsi:
- [x] ~Investigate why the result is different from before.~


fehiepsi:
@neerajprad For some reasons, tests for 0.001, 0.01 does not pass in my local machine (previously it passes). Here Travis passes. So I think that my system is dump again.

neerajprad:
> @neerajprad For some reasons, tests for 0.001, 0.01 does not pass in my local machine (previously it passes)

Is it failing by a large margin? If so, it might mean that some recent jax commit breaks this.

fehiepsi:
@neerajprad Previously grad can get values `1e-38`, now it returns 0, so the rel error will be large (~1). I use the same version as Travis. I'll restart the computer to see if it makes any difference.

neerajprad:
Since the smallest value that can be represented is `1.1754944e-38` with float32, maybe it is underflowing.

fehiepsi:
@neerajprad This is the problem with fast_math mode as reported in https://github.com/google/jax/issues/469. Running the test with `XLA_FLAGS=--xla_cpu_enable_fast_math=false` gives a consistence behaviour.

PR #68, opened 2019-03-27
disable fast math for travis

fehiepsi:
Using fast math gives inconsistent behaviour, especially for tests which require high precision or involve overflow/underflow. In addition, this feature will be disabled by default in JAX in the future.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ sudo: false
 env:
     global:
         - PYTHONPATH=$PWD:$PYTHONPATH
+        - XLA_FLAGS=--xla_cpu_enable_fast_math=false
 
 cache:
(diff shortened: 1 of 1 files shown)


neerajprad:
> Using fast math gives inconsistent behaviour, especially for tests which require high precision or involve overflow/underflow.

Does it result in any significant slowdown in the test suite?

neerajprad:
Also, was this the reason why the gamma grad test was failing on your system?

fehiepsi:
@neerajprad Yes, it is the reason of the failing. I don't see any slowdown.

fehiepsi:
More context about this issue: https://github.com/google/jax/issues/276

PR #70, opened 2019-03-30
Adding HMC kernel to mcmc

neerajprad:
This uses all the HMC related utilities already added by @fehiepsi in previous PRs.

Some design questions to think about:
 - [x] The good thing about using generic pytree data structures is that users can pass in samples using dicts / lists or any other collection type and for as long as their potential energy and kinetic energy computations agree, this is all that is needed. This is a good API to have. However, this unpacking / packing internally can happen at different points in the code. Currently, the adapter code expects flattened `z` but the verlet integrator does the flattening by itself. I think it will be nice if we do this flattening in the outermost MCMC layer and have all HMC utils consume flattened `z` / `r` for consistency. [To be addressed in #74]
 - [x] I think there will be a lot of code duplication between HMC and NUTS and since we don't have inheritance, we should think about how these two classes should share code. [To be addressed in #74]
 - [x] Debugging: I found that even with the JIT disabled using `jax.disable_jit()`, having `lax.cond`, `lax.fori_loop` makes it really hard to debug due to the presence of abstract values. For now, I have manually written the for loops and if conditions that get exercised if the `_jax_jit_disabled` flag is true. This was crucial for debugging and it will be good to have this available for debugging, though I would rather remove this if there is a better way of doing it.

I have commented on the code for other points of discussion.

TODOs:
 - [x] Fix failing `CircularPlanetaryMotion` model test in `test_velocity_verlet`. ~Its quite odd that this test is failing with the changes in this PR.~ Done  in #71. 
 - [x] A couple of more tests - added a logistic regression test. Any other tests?
 - [x] Investigate why HMC returns NaN, Inf values with larger step size. This seems like a bug in the way we are accepting proposals.

Future PRs:
 - Allow utilities so that Pyro like model syntax can be supported easily.

### DIFF
--- a/numpyro/hmc_util.py
+++ b/numpyro/hmc_util.py
@@ -128,13 +128,13 @@ def velocity_verlet(potential_fn, kinetic_fn):
         potential_energy, z_grad = value_and_grad(potential_fn)(z)
         return IntegratorState(z, r, potential_energy, z_grad)
 
-    def update_fn(step_size, state):
+    def update_fn(step_size, state, inv_mass_matrix):
         """
         Single step velocity verlet.
         """
         z, r, _, z_grad = state
         r = tree_multimap(lambda r, z_grad: r - 0.5 * step_size * z_grad, r, z_grad)  # r(n+1/2)
-        r_grad = grad(kinetic_fn)(r)
+        r_grad = grad(kinetic_fn)(r, inv_mass_matrix)
         z = tree_multimap(lambda z, r_grad: z + step_size * r_grad, z, r_grad)  # z(n+1)
         potential_energy, z_grad = value_and_grad(potential_fn)(z)
         r = tree_multimap(lambda r, z_grad: r - 0.5 * step_size * z_grad, r, z_grad)  # r(n+1)
--- a/numpyro/hmc_util.py
+++ b/numpyro/hmc_util.py
@@ -7,7 +7,6 @@ from jax.tree_util import tree_multimap
 
 from numpyro.util import laxtuple
 
-
 AdaptWindow = laxtuple("AdaptWindow", ["start", "end"])
 AdaptState = laxtuple("AdaptState", ["step_size", "inverse_mass_matrix", "ss_state", "mm_state", "window_idx"])
 IntegratorState = laxtuple("IntegratorState", ["z", "r", "potential_energy", "z_grad"])
@@ -128,13 +127,13 @@ def velocity_verlet(potential_fn, kinetic_fn):
         potential_energy, z_grad = value_and_grad(potential_fn)(z)
         return IntegratorState(z, r, potential_energy, z_grad)
 
-    def update_fn(step_size, state, inv_mass_matrix):
+    def update_fn(step_size, inverse_mass_matrix, state):
         """
         Single step velocity verlet.
         """
         z, r, _, z_grad = state
         r = tree_multimap(lambda r, z_grad: r - 0.5 * step_size * z_grad, r, z_grad)  # r(n+1/2)
-        r_grad = grad(kinetic_fn)(r, inv_mass_matrix)
+        r_grad = grad(kinetic_fn)(r, inverse_mass_matrix)
         z = tree_multimap(lambda z, r_grad: z + step_size * r_grad, z, r_grad)  # z(n+1)
         potential_energy, z_grad = value_and_grad(potential_fn)(z)
         r = tree_multimap(lambda r, z_grad: r - 0.5 * step_size * z_grad, r, z_grad)  # r(n+1)
@@ -143,7 +142,8 @@ def velocity_verlet(potential_fn, kinetic_fn):
     return init_fn, update_fn
 
 
-def find_reasonable_step_size(potential_fn, kinetic_fn, momentum_generator, position, init_step_size):
+def find_reasonable_step_size(potential_fn, kinetic_fn, momentum_generator, inverse_mass_matrix,
+                              position, init_step_size):
     # We are going to find a step_size which make accept_prob (Metropolis correction)
     # near the target_accept_prob. If accept_prob:=exp(-delta_energy) is small,
     # then we have to decrease step_size; otherwise, increase step_size.
@@ -163,15 +163,18 @@ def find_reasonable_step_size(potential_fn, kinetic_fn, momentum_generator, posi
         step_size = (2.0 ** direction) * step_size
         r = momentum_generator()  # generate r upon calling
         _, r_new, potential_energy_new, _ = vv_update(step_size,
+                                                      inverse_mass_matrix,
                                                       (z, r, potential_energy, z_grad))
-        energy_current = kinetic_fn(r) + potential_energy
-        energy_new = kinetic_fn(r_new) + potential_energy_new
+        energy_current = kinetic_fn(r, inverse_mass_matrix) + potential_energy
+        energy_new = kinetic_fn(r_new, inverse_mass_matrix) + potential_energy_new
         delta_energy = energy_new - energy_current
         direction_new = np.where(target_accept_prob < -delta_energy, 1, -1)
         return step_size, direction, direction_new
 
-    step_size, _, _ = lax.while_loop(lambda sdd: (sdd[1] == 0) | (sdd[1] == sdd[2]),
-                                     _body_fn, (init_step_size, 0, 0))
+    def _cond_fn(state):
+        return (state[1] == 0) | (state[1] == state[2])
+
+    step_size, _, _ = lax.while_loop(_cond_fn, _body_fn, (init_step_size, 0, 0))
     return step_size
 
 
@@ -372,16 +375,17 @@ def _combine_tree(current_tree, new_tree, inverse_mass_matrix, going_right, rng,
                      sum_accept_probs, num_proposals)
 
 
-@partial(jit, static_argnums=(0, 1, 10))
-def _build_basetree(vv_update, kinetic_fn, z, r, z_grad, step_size, going_right,
+@partial(jit, static_argnums=(0, 1, 11))
+def _build_basetree(vv_update, kinetic_fn, inverse_mass_matrix, z, r, z_grad, step_size, going_right,
                     energy_current, max_delta_energy):
     step_size = np.where(going_right, step_size, -step_size)
     z_new, r_new, potential_energy_new, z_new_grad = vv_update(
         step_size,
-        (z, r, energy_current, z_grad)
+        inverse_mass_matrix,
+        (z, r, energy_current, z_grad),
     )
 
-    energy_new = potential_energy_new + kinetic_fn(r_new)
+    energy_new = potential_energy_new + kinetic_fn(r_new, inverse_mass_matrix)
     # Handles the NaN case.
     energy_new = np.where(np.isnan(energy_new), np.inf, energy_new)
     delta_energy = energy_new - energy_current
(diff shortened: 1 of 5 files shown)


numpyro/mcmc.py
```diff
@@ -0,0 +1,124 @@
...
+            transition = random.bernoulli(rng_transition, accept_prob)
+            if jax.api._jit_is_disabled:
+                if transition:
+                    vv_state = vv_state_new
+            else:
+                vv_state = lax.cond(transition,
+                                    vv_state_new, lambda state: state,
+                                    vv_state, lambda state: state)
+            z_flat, _ = ravel_pytree(vv_state.z)
+            wa_state = wa_update(t, accept_prob, z_flat, wa_state)
```
neerajprad: I haven't looked at this in detail, but should we pass in the proposal state here along with the acceptance probability or the selected state (after the biased coin flip). I have stuck with what we are doing in Pyro, but I am not sure about this.

fehiepsi: I think we'll put the selected state. It is what Stan does AFAIK. Unfortunately, I couldn't find a reference for it. :(

fehiepsi:
> Currently, the adapter code expects flattened z but the verlet integrator does the flattening by itself.

We can just let adapter code accepts unflattened z (similar to what is in Pyro) for consistency.

> I think it will be nice if we do this flattening in the outermost MCMC layer and have all HMC utils consume flattened z / r for consistency.

This is a nice idea. I am in with this strategy. We can also keep the current API because flattening/unflattening is cheap when the input is an array.

> even with the JIT disabled using jax.disable_jit(), having lax.cond, lax.fori_loop makes it really hard to debug due to the presence of abstract values

Yeah, I also get stuck with this problem. To debug, I have to write code in python for/while loop to verify things are correct first and then using while_loop after that. Fortunately, each function does it own job so debugging can be isolated.

numpyro/hmc_util.py
```diff
@@ -128,13 +127,13 @@ def init_fn(z, r):
...
 
-    def update_fn(step_size, state):
+    def update_fn(step_size, inverse_mass_matrix, state):
         """
         Single step velocity verlet.
         """
         z, r, _, z_grad = state
         r = tree_multimap(lambda r, z_grad: r - 0.5 * step_size * z_grad, r, z_grad)  # r(n+1/2)
-        r_grad = grad(kinetic_fn)(r)
+        r_grad = grad(kinetic_fn)(r, inverse_mass_matrix)
```
fehiepsi: I have thought that we can avoid this by defining: `kinetic_fn = partial(..., inverse_mass_matrix)` but I might have been wrong. Let's do this way first and discuss about the changes (if possible) in later PRs.

numpyro/mcmc.py
```diff
@@ -0,0 +1,124 @@
...
+def _get_num_steps(step_size, trajectory_length):
+    num_steps = np.array(trajectory_length / step_size, dtype=np.int32)
+    return np.where(num_steps < 1, np.array(1, dtype=np.int32), num_steps)
+
+
+def _sample_momentum(inverse_mass_matrix, pack_fn, rng):
+    if inverse_mass_matrix.ndim == 1:
+        return pack_fn(dist.norm(0., np.sqrt(np.reciprocal(inverse_mass_matrix))).rvs(random_state=rng))
+    elif inverse_mass_matrix.ndim == 2:
+        raise NotImplementedError
```
fehiepsi: I think that we can sample `epsilon ~ N(0, 1)` and instead of `inverse_mass_matrix`, we require an input "scale" here. We might have a MassMatrixState(inverse, sqrt) laxtuple and will update it at warmup_adapter method. The inverse one is used for kinetic_fn and sqrt is used for sampling momentum.

numpyro/hmc_util.py
neerajprad: I considered that. I think the issue there is that `kinetic_fn` will not be a pure function, and will need to passed around. However that has other issues because it will need to be a static arg, but since it will keep changing as `mass_matrix` is adapted, it will keep triggering recompilation. As such, I thought it best to keep it simple and just pass the `mass_matrix` which can be easily jitted.

fehiepsi:
> there will be a lot of code duplication between HMC and NUTS and since we don't have inheritance, we should think about how these two classes should share code.

I think we can resolve this by separating hmc and mcmc. So hmc/nuts function will return hmc_init/nuts_init and hmc_update/nuts_update.

numpyro/hmc_util.py
fehiepsi: I am curious that if `kinetic_fn` is changed inside the body of a while_loop, then does it cause any issue? Right now, recursive build requires jitting functions having kinetic_fn as an arg but how about iterative build? I just want to know your opinion. There is no need to test it out in this PR. :)

numpyro/mcmc.py
neerajprad: That's a good way to do it.

numpyro/hmc_util.py
neerajprad: I'm not completely sure - my guess would be that since traced function applications are cached by their arguments, every time we apply partial to `kinetic_fn`, we will change the hash of this partially applied function. Since the key will be missing in the cache, it will still force recompilation. Unlike some other issues, we know that this will only happen a few times, so for large enough problems it may not be as much of an overhead. I would be very surprised and curious if this wasn't the case. :)

fehiepsi: I go ahead and test our hypothesis in [this notebook](https://gist.github.com/fehiepsi/2fdbdd7db13183a0b3fd9fad53287cc8). Despite that I have changed `pf`, jitted h does not know about this change and still runs fast.

numpyro/mcmc.py
neerajprad: I would like to punt this to a follow-up PR, if that's okay.

test/test_mcmc.py
```diff
@@ -0,0 +1,62 @@
...
+            y_lpdf = dist.bernoulli(probs).logpmf(labels)
+            return - (np.sum(coefs_lpdf) + np.sum(y_lpdf))
+
+        def kinetic_fn(beta, m_inv):
+            return 0.5 * np.dot(m_inv, beta ** 2)
+
+        init_kernel, sample_kernel = hmc_kernel(potential_fn, kinetic_fn)
+        init_samples = np.zeros(dim)
+        hmc_state = init_kernel(init_samples,
+                                step_size=0.01,
```
neerajprad: One thing that I noticed is that HMC in numpyro is way more sensitive to these hyper-parameters and will give a `nan` or `inf` values easily if these parameters are too high.

fehiepsi: The `nan`, `inf` are natural values for this model if step_size is large I guess. Might be find_reasonable_step_size does not work as expected. I'll take a look too.

fehiepsi: neerajprad There are a few things happening here:
+ Starting with step_size=0.01, num_steps=15 is equivalent to set trajectory_length=0.15. If we set step_size=1, num_steps=15, then each trajectories will require a large number of steps. So the inference will be slower. Anyway, this is not the reason why we get `nan`, `inf` so often.
+ I think that the reason is: pytorch clamps parameters while numpyro did not. For example,
`torch_dist.Bernoulli(1).log_prob(0) = -16` while `jax_dist.bernoulli(1).logpmf(0) = -inf`. Of course, -inf is the correct value but having -inf log_prob will easily lead to grad of latent is +-inf or nan.

To fix this issue, I suggest to clamp `probs` in potential_fn:
```
            probs = np.clip(expit(np.sum(beta * data, axis=-1)),
                            a_min=np.finfo(np.float32).tiny,
                            a_max=(1 - np.finfo(np.float32).eps))
```

Later, we should do this automatically for users.

neerajprad: Thanks a lot, @fehiepsi for digging into this! I have followed this suggestion and also adjusted the step size to higher than what it was previously.

neerajprad:
@fehiepsi - Let me know if anything needs to be addressed with this PR before we merge. I think the interface issues with MCMC/NUTS should be handled/discussed separately, and we can have smaller PRs later to handle specific parts of #74.

numpyro/hmc_util.py
neerajprad: @fehiepsi - It turns out that the tracer has no issues with partial function application inside a while loop or JITing the enclosing function, as it correctly traces the applied operations and doesn't need to even store/cache the function. I think it will matter in cases like below, where you should notice recompilation ticking in and increasing the runtime. 

```python
@jax.partial(jax.jit, static_argnums=(0,))
def fn(partial_ke, r):
    ...
```

fehiepsi: Yes, I also think so. Currently we use this pattern (@partial(jit, ...)) for recursive build, but when memory issue is addressed for iterative build, I think we can revert back to the old behaviour without any problem.

test/test_mcmc.py
```diff
@@ -0,0 +1,64 @@
...
+    assert_allclose(np.std(zs), true_std, rtol=0.05)
+
+
+def test_logistic_regression():
+    N, dim = 3000, 3
+    warmup_steps, num_samples = 1000, 8000
+    data = random.normal(random.PRNGKey(0), (N, dim))
+    true_coefs = np.arange(1., dim + 1.)
+    probs = expit(np.sum(true_coefs * data, axis=-1))
+    labels = dist.bernoulli(probs).rvs()
```
fehiepsi: nit: we need `random_state` here

neerajprad: @fehiepsi - We are using the scipy sampler here (for all discrete distributions currently), so we don't need to pass in `random_state`, but let me do that for consistency.

numpyro/hmc_util.py
neerajprad: I still think we shouldn't liberally use partial, specially if we have to do it very often for the same function, because a failure to update the function will cause silent bugs. On the other hand, if there are only a limited number of places where we have to use this, it can help readability. We can refactor this later if that turns out to be the case.

numpyro/mcmc.py
```diff
@@ -0,0 +1,111 @@
...
+                                 lambda i, val: vv_update(step_size, inverse_mass_matrix, val),
+                                 vv_state)
+        energy_old = vv_state.potential_energy + kinetic_fn(vv_state.r, inverse_mass_matrix)
+        energy_new = vv_state_new.potential_energy + kinetic_fn(vv_state_new.r, inverse_mass_matrix)
+        delta_energy = energy_new - energy_old
+        delta_energy = np.where(np.isnan(delta_energy), np.inf, delta_energy)
+        accept_prob = np.clip(np.exp(-delta_energy), a_max=1.0)
+        return accept_prob, vv_state_new
+
+    def sample_kernel(hmc_state, i):
```
neerajprad: @fehiepsi  - I'm not too happy with this unused `i` parameter. Are there other alternatives to `lax.scan` that we can use instead? I'll merge this as is, but was just curious.

test/test_mcmc.py
fehiepsi: I see. I mean that we'd better set `random_state=0` here. Right now, `dist.bernoulli(0.5).rvs()` returns different values for different runs with the same numpy.seed:
```
onp.random.seed(0)
dist.bernoulli(0.5).rvs(size=(100,))
```

neerajprad: That's really strange that even with a fixed numpy seed it returns different results.

numpyro/mcmc.py
fehiepsi: @neerajprad I think that we'd better remove it. And use `lax.scan` with `lambda i, state: sample_kernel(state)` as in `fori_loop`.

test/test_mcmc.py
fehiepsi: Yes, I think it is strange too. While debugging, I thought that `grad` or `value_and_grad` gives inconsistency results but it turns out that the data changed. ^^!

numpyro/hmc_util.py
fehiepsi: Yup, agree. These things can be refactored later if necessary. We should avoid making any confusion which might block us. :)

PR #71, opened 2019-03-30
Add mass matrix inverse as arg to kinetic energy

neerajprad:
This provides an `inverse_mass_matrix` arg to `kinetic_energy` which is required for #70.

### DIFF
--- a/numpyro/hmc_util.py
+++ b/numpyro/hmc_util.py
@@ -376,7 +376,7 @@ def _combine_tree(current_tree, new_tree, inverse_mass_matrix, going_right, rng,
 
 
 @partial(jit, static_argnums=(0, 1, 11))
-def _build_basetree(vv_update, kinetic_fn, inverse_mass_matrix, z, r, z_grad, step_size, going_right,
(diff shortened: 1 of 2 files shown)


numpyro/hmc_util.py
```diff
@@ -372,16 +375,17 @@ def _combine_tree(current_tree, new_tree, inverse_mass_matrix, going_right, rng,
                      sum_accept_probs, num_proposals)
 
 
-@partial(jit, static_argnums=(0, 1, 10))
-def _build_basetree(vv_update, kinetic_fn, z, r, z_grad, step_size, going_right,
+@partial(jit, static_argnums=(0, 1, 11))
+def _build_basetree(vv_update, kinetic_fn, inverse_mass_matrix, z, r, z_grad, step_size, going_right,
```
fehiepsi: I think it is better to move this before `step_size` for consistence.

neerajprad: Sure, I grouped it so that it is close to `kinetic_fn`, but I'm fine with any consistent grouping. Will make the change.

PR #73, opened 2019-03-31
Provide context manager to disable tracing in control flow primitives

### LINKED ISSUE #72: Build wrappers for lax primitives to disable tracing in debug mode
Even with the `disable_jit` flag, it is inconvenient to debug our existing code since we have many utilities that use these lax primitives which need to be manually rewritten each time we want to debug something. Lets build light wrappers around these to make debugging easy. 

We should also move this upstream to JAX if we find it useful.

neerajprad:
Fixes #72. 

This provides a context manager `control_flow_prims_disabled` that converts all control flow primitives into regular python loops / conditionals. Once (if) this is moved to JAX, we can simply change the import to `lax` without adding ugly boilerplate code like I have done in #70. 

Note that jit compilation can already be disabled with `jax.api.disable_jit()` context_manager, and I have kept this change decoupled so that it is more generally useful. To disable both, we will need to use:

```python
with disable_jit(), control_flow_prims_disabled():
```
Suggestions on shorter name is welcome!

### Testing
I have tested this on one of our faster hmc util tests, as this can be really slow!

### DIFF
--- a/numpyro/hmc_util.py
+++ b/numpyro/hmc_util.py
@@ -5,7 +5,7 @@ from jax.ops import index_update
 from jax.scipy.special import expit
 from jax.tree_util import tree_multimap
 
-from numpyro.util import cond, control_flow_prims_disabled, laxtuple, optional, while_loop
+from numpyro.util import cond, laxtuple, while_loop
 
 AdaptWindow = laxtuple("AdaptWindow", ["start", "end"])
 AdaptState = laxtuple("AdaptState", ["step_size", "inverse_mass_matrix", "ss_state", "mm_state", "window_idx"])
--- a/numpyro/hmc_util.py
+++ b/numpyro/hmc_util.py
@@ -5,7 +5,7 @@ from jax.ops import index_update
 from jax.scipy.special import expit
 from jax.tree_util import tree_multimap
 
-from numpyro.util import laxtuple
+from numpyro.util import cond, control_flow_prims_disabled, laxtuple, optional, while_loop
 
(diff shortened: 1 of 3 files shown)


numpyro/util.py
```diff
@@ -36,3 +38,54 @@ def laxtuple(name, fields):
...
+        _DISABLE_CONTROL_FLOW_PRIM = True
+        yield
+    finally:
+        _DISABLE_CONTROL_FLOW_PRIM = stored_flag
+
+
+def cond(pred, true_operand, true_fun, false_operand, false_fun):
+    if _DISABLE_CONTROL_FLOW_PRIM:
+        if pred:
+            return true_fun(true_operand)
```
fehiepsi: else statement is missing here

neerajprad: Oops..let me fix this.

numpyro/hmc_util.py
```diff
@@ -578,9 +578,6 @@ def _body_fn(state):
         return tree, key
 
     state = (tree, key)
-    if iterative_build:
-        tree, _ = lax.while_loop(_cond_fn, _body_fn, state)
-    else:
-        while _cond_fn(state):
-            tree, _ = state = _body_fn(state)
+    with optional(not iterative_build, control_flow_prims_disabled()):
```
neerajprad: Note that this isn't equivalent to the version on the left, because it sets the `_DISABLE_CONTROL_FLOW_PRIM` flag for other operations inside `build_tree` which won't work unless jit is disabled as well. I will revert this change.

fehiepsi:
Thanks @neerajprad !

Issue #74, opened 2019-03-31
Follow-up issues for HMC


neerajprad:
From discussion in #70:

 - [x] Uniform handling of packing / unpacking pytree values. I think we can just flatten everything in the kernel itself and unflatten the result received. Alternatively (if this is relatively cheap), we can also just let the adapter function take in `z` and do the flattening. The latter will be better for debuggability, and I'm not sure if there is a significant overhead under JIT. 
 - [x] The current `hmc_kernel` can have a separate `get_kernel` method that would return either an HMC or NUTS kernel depending on args. This can be called by the user or from inside `mcmc`.
 - [x] Default kinetic energy function. This shouldn't be needed as an arg from the user, at least not while we are using euclidean KE.
 - [ ] Pyro style model/guide syntax and wrapper utilities so that users don't need to specify the PE computation explicitly.
 - [x] Clamp probability values (also, provide the logit parametrization), which will really help stabilize HMC trajectories as pointed out by @fehiepsi.
 - [x] Correct the behaviour of find_reasonable_step_size
 - [x] Either drop z_grad, potential_grad from HMCState or modify the behaviour of `lax.scan` to only scan `z` by default. @fehiepsi
 - [x] ~Add context manager to convert `lax.cond` to `lax.while`.~  @neerajprad unfortunately, using while_loop requires initial_state has the same format as the output; but we don't know the output format without evaluating one of true or false functions.
 - [x] Investigate why test_unnormalized_normal keeps failing in @fehiepsi's system.

fehiepsi:
> Alternatively (if this is relatively cheap), we can also just let the adapter function take in z and do the flattening.

I vote for this way. I think that it is cheap because we have a bunch of this operator in `_is_turning` function. It is easy to detect if it is cheap or not when we start to work with more complicated models.

> The current hmc_kernel can have a separate get_kernel method that would return either an HMC or NUTS kernel depending on args. This can be called by the user or from inside mcmc.

Maybe we'd better discuss about this design choice in message. :)

> Default kinetic energy function.

Yes, I agree.

> Pyro style model/guide syntax

Yup.

fehiepsi:
@neerajprad Look like most issues of HMC have been resolved. In upcoming days, I will address memory issue of iterative NUTS, and try to reduce its compiling time if possible. I think we have some room to improve given that functions will be recompiled when static args are not put in keywords (which we observed in dirichlet pr).

neerajprad:
[...]

PR #75, opened 2019-04-02
Correct the usage of find reasonable step size function

fehiepsi:
This function requires an additional state rng for momentum generator. Addresses points 1 and 6 of #74.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -16,7 +16,7 @@ multi_line_output = 3
 filterwarnings = error
     ignore:numpy.ufunc size changed,:RuntimeWarning
     ignore:Using a non-tuple sequence:FutureWarning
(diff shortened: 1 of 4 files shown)


neerajprad:
Nice refactoring! I think some of the lint checks are failing.

setup.cfg
```diff
@@ -16,7 +16,7 @@ multi_line_output = 3
 filterwarnings = error
     ignore:numpy.ufunc size changed,:RuntimeWarning
     ignore:Using a non-tuple sequence:FutureWarning
-    once:No GPU found:UserWarning
+    once:No GPU:UserWarning
```
fehiepsi: new version of jax has message No GPU/TPU found, so pytest does not detect it

Issue #51, opened 2019-03-18
Add logits support to discrete distributions


neerajprad:
Since we use float32 values by default instead of float64 like `scipy.stats` does, having support for logits in multinomial, bernoulli and binomial distributions is important, and will make  operations that manipulate probabilities more numerically stable.

PR #79, opened 2019-04-03
Support logits to frozen Bernoulli distribution

fehiepsi:
This PR supports a kwarg `is_logits` to Bernoulli distribution. By default, is_logits=False and it will be so for non-frozen distribution. For frozen distribution, by adding `is_logits` argument, the arg `p` will be considered as a logit, instead of prob.

@neerajprad Pls find my reasoning in comments. There might be some edge cases which I have not figured it out yet, so please let me know your thought about this change. :)

### DIFF
--- a/numpyro/distributions/discrete.py
+++ b/numpyro/distributions/discrete.py
@@ -16,8 +16,7 @@ from jax.numpy.lax_numpy import _promote_dtypes
 from jax.scipy.special import expit, gammaln
 
 from numpyro.distributions.distribution import jax_discrete
-from numpyro.distributions.util import (binary_cross_entropy_with_logits, entr, promote_shapes,
-                                        xlog1py, xlogy)
+from numpyro.distributions.util import binary_cross_entropy_with_logits, entr, promote_shapes, xlog1py, xlogy
 
 
 class _binom_gen(jax_discrete, binom_gen):
--- a/numpyro/distributions/discrete.py
+++ b/numpyro/distributions/discrete.py
@@ -13,10 +13,11 @@ from scipy.stats._multivariate import multinomial_gen
 import jax.numpy as np
 from jax import device_put, lax
 from jax.numpy.lax_numpy import _promote_dtypes
-from jax.scipy.special import gammaln
+from jax.scipy.special import expit, gammaln
 
 from numpyro.distributions.distribution import jax_discrete
-from numpyro.distributions.util import entr, promote_shapes, xlog1py, xlogy
+from numpyro.distributions.util import (binary_cross_entropy_with_logits, entr, promote_shapes,
+                                        xlog1py, xlogy)
 
 
 class _binom_gen(jax_discrete, binom_gen):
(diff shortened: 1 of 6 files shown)


test/test_distributions.py
```diff
@@ -140,7 +141,7 @@ def test_logprob(jax_dist, loc_scale):
 
 @pytest.mark.parametrize('jax_dist, dist_args', [
     (dist.bernoulli, (0.1,)),
-    (dist.bernoulli, (np.array([0.3, 0.5]))),
+    (dist.bernoulli, (np.array([0.3, 0.5]),)),
```
fehiepsi: this is typo

numpyro/distributions/util.py
```diff
@@ -273,6 +273,12 @@ def entr(p):
     return e
 
 
+def binary_cross_entropy_with_logits(x, y):
+    # compute -y * log(sigmoid(x)) - (1 - y) * log(1 - sigmoid(x))
+    # Ref: https://www.tensorflow.org/api_docs/python/tf/nn/sigmoid_cross_entropy_with_logits
+    return np.clip(x, 0) - x * y + np.log1p(np.exp(-np.abs(x)))
```
fehiepsi: nit: reorder the terms such as terms involving only `x` is grouped (to save a bit of time when y has batch dimensions).

numpyro/distributions/distribution.py
```diff
@@ -67,11 +67,11 @@ def rvs(self, *args, **kwargs):
         # assert that rng is PRNGKey and not mtrand.RandomState object from numpy.
         assert _is_prng_key(rng)
         kwargs['random_state'] = onp.random.RandomState(rng)
-        sample = super(osp_stats.rv_discrete, self).rvs(*args, **kwargs)
+        sample = super(jax_discrete, self).rvs(*args, **kwargs)
```
fehiepsi: I think this is a typo too

numpyro/distributions/discrete.py
```diff
@@ -33,8 +34,39 @@ def _entropy(self, n, p):
 
 
 class _bernoulli_gen(jax_discrete, bernoulli_gen):
+    def __new__(cls, *args, **kwargs):
+        return super(_bernoulli_gen, cls).__new__(cls)
```
fehiepsi: [rv_discrete](https://github.com/scipy/scipy/blob/master/scipy/stats/_distn_infrastructure.py#L2690) has a signature which prevents us from adding additional keyword `is_logits`.

numpyro/distributions/discrete.py
```diff
@@ -33,8 +34,39 @@ def _entropy(self, n, p):
...
+
+    def __init__(self, *args, **kwargs):
+        self.is_logits = kwargs.pop("is_logits", False)
+        super(_bernoulli_gen, self).__init__(*args, **kwargs)
+
+    def freeze(self, *args, **kwargs):
+        self._ctor_param.update(is_logits=kwargs.pop("is_logits", False))
```
fehiepsi: We update `_ctor_param` so that [frozen distribution](https://github.com/scipy/scipy/blob/master/scipy/stats/_distn_infrastructure.py#L427) makes correct signatures for its self.dist.

test/test_mcmc.py
```diff
@@ -44,10 +44,8 @@ def test_logistic_regression():
         def potential_fn(beta):
             coefs_mean = np.zeros(dim)
             coefs_lpdf = dist.norm(coefs_mean, np.ones(dim)).logpdf(beta)
-            probs = np.clip(expit(np.sum(beta * data, axis=-1)),
-                            a_min=np.finfo(np.float32).tiny,
-                            a_max=(1 - np.finfo(np.float32).eps))
-            y_lpdf = dist.bernoulli(probs).logpmf(labels)
+            logits = np.sum(beta * data, axis=-1)
+            y_lpdf = dist.bernoulli(logits, is_logits=True).logpmf(labels)
```
neerajprad: Nice!

numpyro/distributions/discrete.py
neerajprad: This is a nice work-around.

test/test_distributions.py
```diff
@@ -192,3 +193,20 @@ def test_discrete_logpmf_args_check(jax_dist, dist_args):
...
         jax_dist.logpmf(sample_oos, *dist_args)
+
+
+@pytest.mark.parametrize('jax_dist, dist_args', [
+    (dist.bernoulli, (0.1,)),
+    (dist.bernoulli, (np.array([0.3, 0.5]),)),
+], ids=idfn)
+def test_discrete_with_logits(jax_dist, dist_args):
+    rng = random.PRNGKey(0)
+    logit_args = dist_args[:-1] + (logit(dist_args[-1]),)
```
neerajprad: I think you probably just meant `logit(dist_args)`?

fehiepsi: Yes, I mean so. I expect more distributions will need this as mentioned in https://github.com/pyro-ppl/numpyro/issues/51, so this just serves for future tests.

fehiepsi:
> we should probably only support frozen distributions and keep our code simple

Indeed, I like the current state because we inherit many mechanisms from scipy. Adding more distribution won't be much extra work (except multivariate ones, where each dist has its own mechanism) so it is also convenience for other developers. We indeed do not interact with frozen dists because all the logics belong to unfrozen dists. Except for pyro models (which requires frozen dist), users can define a custom objective with their familiar coding style. But I do agree that we only support extra features for frozen distributions. One of the extra features which I have in mind now is having `scale_tril` arg for multivariate normal distribution.

test/test_distributions.py
neerajprad: @fehiepsi - I just meant why do we need to do `dist_args[:-1] + (logit(dist_args[-1]),)` instead of just using `(logit(dist_args),)`?

fehiepsi: For example, in binom/multinomial distributions, we have 2 args: n and p. So we only want to apply `logit` to `p`.

PR #81, opened 2019-04-04
Implement dirichlet distribution

fehiepsi:
This PR implements dirichlet distribution.

I have bad experience while implementing this distribution. It took me 2 weeks to make things work and compatible with scipy. I can't inherit from dirichlet_gen because the frozen version requires arg check. But in scipy, dirichlet distribution does not support batch so arg check won't work for batched alpha (or batched samples). Inherit from multi_rv_generic requires implementing again both unfrozen and frozen versions, which does not make sense to me. So I decide to make a layer jax_mvcontinuous which inherited from rv_continuous. Luckily, I don't have to reimplement rv_continuous.__ init__ method. Instead, rewriting `_construct_argparser` logic to remove log/scale things is enough.

I am not sure if this will make the work of implementing other multivariate distributions less painful. Hope that when we have a few of them, we can make a better common interface.

### DIFF
--- a/numpyro/distributions/continuous.py
+++ b/numpyro/distributions/continuous.py
@@ -20,8 +20,8 @@ class beta_gen(jax_continuous):
         # XXX the implementation is different from PyTorch's one
         # in PyTorch, a sample is generated from dirichlet distribution
         key_a, key_b = random.split(self._random_state)
-        gamma_a = standard_gamma(key_a, a, self._size)
-        gamma_b = standard_gamma(key_b, b, self._size)
+        gamma_a = standard_gamma(key_a, a, shape=self._size)
+        gamma_b = standard_gamma(key_b, b, shape=self._size)
         return gamma_a / (gamma_a + gamma_b)
 
     def _cdf(self, x, a, b):
@@ -40,7 +40,7 @@ class cauchy_gen(jax_continuous):
     def _rvs(self):
         # TODO: move this implementation upstream to jax.random.standard_cauchy
         # Another way is to generate X, Y ~ Normal(0, 1) and return X / Y
-        u = random.uniform(self._random_state, self._size)
+        u = random.uniform(self._random_state, shape=self._size)
         return np.tan(np.pi * (u - 0.5))
 
     def _pdf(self, x):
@@ -68,7 +68,7 @@ class cauchy_gen(jax_continuous):
 
 class expon_gen(jax_continuous):
     def _rvs(self):
-        u = random.uniform(self._random_state, self._size)
+        u = random.uniform(self._random_state, shape=self._size)
         return -np.log(u)
 
     def _cdf(self, x):
@@ -95,7 +95,7 @@ class expon_gen(jax_continuous):
 
 class gamma_gen(jax_continuous):
     def _rvs(self, a):
-        return standard_gamma(self._random_state, a, self._size)
+        return standard_gamma(self._random_state, a, shape=self._size)
 
     # TODO: add _cdf/_sf methods when incomplete gamma is available
     # https://github.com/google/jax/issues/479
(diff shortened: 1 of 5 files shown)


numpyro/distributions/distribution.py
```diff
@@ -58,6 +63,9 @@ def logpdf(self, x, *args, **kwargs):
 class jax_discrete(osp_stats.rv_discrete):
     args_check = True
 
+    def _support_mask(self, k):
+        return (k >= self.a) & (k <= self.b) & (np.floor(k) == k)
```
fehiepsi: I move this check outside in preparing for using a common support constraint instance as in pytorch

neerajprad:
> Inherit from multi_rv_generic requires implementing again both unfrozen and frozen versions, which does not make sense to me.

I think it was great to get started with the scipy implementations, but once we get to a point where we have already implemented `rvs` and `logpdf` methods, we can just remove the dependency on scipy completely. I think right now, the discrete distributions use the scipy `rvs` so this isn't feasible. I think longer term, we will be held back by trying to work around scipy's design.

neerajprad:
Could you comment on how slow the dirichlet sampler is wrt the scipy implementation? Also, since gamma isn't reparametrized currently, my guess is that the only advantage to using our `standard_gamma` based sampler right now is that it is a deterministic transformation of PRNGKey (which may / may not be important). Is that correct?

fehiepsi:
Yes, agree. We only use doc/rvs/logpdf/freeze mechanism (which is quite simple to rewrite). When we have implemented constraints, we'll use our own logic for arg check and support check. Even for discrete distribution, we just simply call scipy.rvs in each `_rvs` method without having to do "class inheritance" from scipy.

fehiepsi:
@neerajprad Currently, our gamma sampler is reparametrized.

numpyro/distributions/discrete.py
```diff
@@ -33,8 +33,39 @@ def _entropy(self, n, p):
...
+
+    def _argcheck(self, p):
+        if self.is_logits:
+            return np.isfinite(p)
+        else:
+            return (p >= 0) & (p <= 1)
+
     def _logpmf(self, x, p):
-        return xlogy(x, p) + xlog1py(1 - x, -p)
+        if self.is_logits:
```
neerajprad: Could you merge master so that these changes do not show up?

neerajprad:
> @neerajprad Currently, our gamma sampler is reparametrized.

Sorry, I meant dirichlet. As we discussed, it won't give us the correct pathwise gradient.

fehiepsi:
@neerajprad For alpha.shape=3, size=1000, it is 2.5ms vs 1ms from scipy. It is interesting that without specifying `shape` args from jax.random samplers, the samplers will recompile. For example
```
# using standard_gamma(key, alpha, size)
%timeit dist.dirichlet.rvs(alpha, size=1000, random_state=...)  # this is slow
# using standard_gamma(key, alpha, shape=size)
%timeit dist.dirichlet.rvs(alpha, size=1000, random_state=...)  # this is fast
# NB: of course, jitting dirichlet.rvs will give the same speed
```

fehiepsi:
> As we discussed, it won't give us the correct pathwise gradient.

Yes, I think so. I have not got response in jax thread about reparameterizing yet. I think that though it won't give a correct reparameterized gradient, it is still better than not using it (this claim needs verification from authors of implicit gradient reparameterization paper).

neerajprad:
>  It is interesting that without specifying shape args from jax.random samplers, the samplers will recompile.

That seems odd. I think they should only recompile when the shape arg changes. Is that not the case?

numpyro/distributions/continuous.py
```diff
@@ -40,7 +40,7 @@ class cauchy_gen(jax_continuous):
     def _rvs(self):
         # TODO: move this implementation upstream to jax.random.standard_cauchy
         # Another way is to generate X, Y ~ Normal(0, 1) and return X / Y
-        u = random.uniform(self._random_state, self._size)
+        u = random.uniform(self._random_state, shape=self._size)
```
neerajprad: This seems like a fine change and isn't blocking this PR, but it will be good to understand why this leads to recompilation.

fehiepsi: I am curious too so I am investigating it. This is True at least for gamma sampler.

fehiepsi:
@neerajprad There is a small repro for that phenomenon

```
from jax import random

def f(size):
    size = (size,)
    return random.normal(random.PRNGKey(0), size)


def g(size):
    size = (size,)
    return random.normal(random.PRNGKey(0), shape=size)

%timeit f(1000)  # 250ms
%timeit g(1000)  # 800micro second
```

PR #83, opened 2019-04-05
Add nuts update and expose warmup update

fehiepsi:
This PR adds `_nuts_next` method and expose warmup_update function when `run_warmup` flag is set to False in init_kernel.

### DIFF
--- a/numpyro/hmc_util.py
+++ b/numpyro/hmc_util.py
@@ -229,7 +229,7 @@ def warmup_adapter(num_adapt_steps, find_reasonable_step_size=None,
     num_windows = len(adaptation_schedule)
 
     def init_fn(z, rng, step_size=1.0, inverse_mass_matrix=None, mass_matrix_size=None):
-        rng, rng_sub = random.split(rng)
+        rng, rng_ss = random.split(rng)
         if inverse_mass_matrix is None:
             assert mass_matrix_size is not None
             if diag_mass:
@@ -238,7 +238,7 @@ def warmup_adapter(num_adapt_steps, find_reasonable_step_size=None,
                 inverse_mass_matrix = np.identity(mass_matrix_size)
 
         if find_reasonable_step_size is not None:
-            step_size = find_reasonable_step_size(inverse_mass_matrix, z, rng_sub, step_size)
+            step_size = find_reasonable_step_size(inverse_mass_matrix, z, rng_ss, step_size)
         ss_state = ss_init(np.log(10 * step_size))
 
         mm_state = mm_init(inverse_mass_matrix.shape[-1])
@@ -246,7 +246,7 @@ def warmup_adapter(num_adapt_steps, find_reasonable_step_size=None,
         window_idx = 0
         return AdaptState(step_size, inverse_mass_matrix, ss_state, mm_state, window_idx, rng)
 
-    def _update_at_window_end(z, rng_sub, state):
+    def _update_at_window_end(z, rng_ss, state):
         step_size, inverse_mass_matrix, ss_state, mm_state, window_idx, rng = state
 
         if adapt_mass_matrix:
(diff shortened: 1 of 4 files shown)


fehiepsi:
The fail on Travis seems related to point 9 in #74. There are some undetermined behaviour of hmc_kernel.

fehiepsi:
@neerajprad In previous PRs, unnormalized normal model failed in my system while passed in travis. Reversely, in this PR, unnormalized normalized normal model failed in Travis while passed in my system. I looked at traceplots of the results to see what is going on with the walks and observed that the default trajectory length (2pi) is the reason. I make a [gist](https://gist.github.com/fehiepsi/13770a6d72c9e0edc37bc71f9b5bd381) here to show that. The gist only plots different results if we change initial num_steps with this PR. But this phenomenon applied for previous PRs too. You can also change the random seed to see the unstable problem with default trajectory length.

numpyro/mcmc.py
```diff
@@ -38,54 +46,53 @@ def init_kernel(init_samples,
...
-        num_steps = _get_num_steps(wa_state.step_size, trajectory_length)
-        return HMCState(vv_state.z, vv_state.z_grad, vv_state.potential_energy, num_steps,
-                        wa_state.step_size, wa_state.inverse_mass_matrix, rng)
-
-    def _next(num_steps, step_size, inverse_mass_matrix, vv_state):
+        if run_warmup:
+            hmc_state, _ = fori_loop(0, num_warmup_steps, warmup_update, (hmc_state, wa_state))
+            return hmc_state
+        else:
+            return warmup_update, hmc_state, wa_state
```
neerajprad: nit: slight preference for keeping `hmc_state` as the first return arg.

numpyro/hmc_util.py
```diff
@@ -8,7 +8,8 @@
 from numpyro.util import cond, laxtuple, while_loop
 
 AdaptWindow = laxtuple("AdaptWindow", ["start", "end"])
-AdaptState = laxtuple("AdaptState", ["step_size", "inverse_mass_matrix", "ss_state", "mm_state", "window_idx"])
+AdaptState = laxtuple("AdaptState", ["step_size", "inverse_mass_matrix", "ss_state", "mm_state",
+                                     "window_idx", "rng"])
```
neerajprad: I think it will be nice to have a consistent pattern for passing in `rng` to functions. Right now, `HMCState` and `AdaptState` take it as a last arg, and some other functions take it as a first arg. Shouldn't be part of this PR, but something that we can think about for future refactors.

test/test_mcmc.py
```diff
@@ -11,28 +12,29 @@
...
     def kinetic_fn(r, m_inv):
-        return 0.5 * np.sum(m_inv * r[0] ** 2)
+        return 0.5 * np.sum(m_inv * r ** 2)
 
-    init_kernel, sample_kernel = hmc_kernel(potential_fn, kinetic_fn)
-    init_samples = [np.array([0.])]
+    init_kernel, sample_kernel = hmc_kernel(potential_fn, kinetic_fn, algo)
+    init_samples = np.array(0.)
     hmc_state = init_kernel(init_samples,
+                            num_steps=10,
```
neerajprad: Could you elaborate on why `num_steps` is needed? Is it for NUTS or HMC?

neerajprad: I missed your comment above. Let me go through the gist.

fehiepsi: I added a comment above the instability of the default trajectory length

numpyro/hmc_util.py
fehiepsi: I only observe the inconsistency in build_tree functions. Will address it in a later PR. :)

neerajprad:
@fehiepsi - I see very similar trajectories on my system with either initial configuration (both look like your first plot), and am not able to replicate the difference that you observe. Could you paste what `hmc_state` looks like for for your second run (exactly like you did for the first one)? It seems really odd to me that NUTS would be sensitive to this initial parameter at all, since the ideal trajectory length for NUTS should be governed by the turning criterion.

fehiepsi:
@neerajprad I just update the gist. I don't find any problem yet with NUTS. This just happens with HMC where trajectory_length/num_steps is a tuning parameter.

neerajprad:
> @neerajprad I just update the gist. I don't find any problem yet with NUTS. This just happens with HMC where trajectory_length/num_steps is a tuning parameter.

Okay, I missed that this was only for HMC. That makes sense. Any idea why `num_steps` is quite less than `2*pi / step_size` (for the last two cases, even for the first one it seems to be less than `trajectory_length / step_size`). I may be missing something.

fehiepsi:
> Any idea why num_steps is quite less than 2*pi / step_size

That's because step_size, inverse_mass_matrix are infos for the next trajectory, not current trajectory. To be more precisely,
+ z, pe, z_grad: last state of current trajectory / starting state of next trajectory
+ step_size, m_inv: state of next trajectory
+ accept_prob: state of current trajectory
+ num_steps: in this PR/in pyro, it is state of current trajectory (but we can make it to be the state of next trajectory by changing the order in the code; in that case, the meanings of num_steps will be different for HMC and NUTS).

neerajprad:
I see. I think `z`, `pe`, `z_grad` and `accept_prob` referring to the last stage is expected. It will be nice if we had `num_steps` also reflect the starting point for the `verlet_integrator` in the next update, but you are right that it will cause further confusion if it doesn't refer to the same thing in HMC and NUTS.

PR #84, opened 2019-04-05
implement a custom lax.scan (for tuple) which only stores selected fields

fehiepsi:
Addresses point 7 of #74. Having to scan inverse_mass_matrix and z_grad requires 3x memory storage so it would be more effective if we only store some selected fields from HMCState. This is achieve by using `tscan` in this PR. Currently, it is tested for laxtuple but the code is expected to work for general pytree.

### DIFF
--- a/numpyro/util.py
+++ b/numpyro/util.py
@@ -112,8 +112,38 @@ def scan(f, a, bs):
 
 
 def tscan(f, a, bs, fields=(0,)):
+    if _DISABLE_CONTROL_FLOW_PRIM:
+        # convert pytree to flat jaxtuple
+        a, a_tree = pytree_to_flatjaxtuple(a)
+        bs, b_tree = pytree_to_flatjaxtuple(bs)
+        length = tuple(bs)[0].shape[0]
+        state = [lax.full((length,) + a[i].shape, 0, lax._dtype(a[i])) for i in fields]
+
+        def body_fun(i, vals):
+            a, state = vals
+            # select i-th element from each b
+            b = [b[i] for b in bs]
+            ff, _ = pytree_fun_to_flatjaxtuple_fun(wrap_init(f), (a_tree, b_tree))
+            a_out = ff.call_wrapped(a, core.pack(b))
+            # select fields from a_out and update state
+            state_out = [lax.dynamic_update_index_in_dim(s, a[None, ...], i, axis=0)
+                         for a, s in zip([tuple(a_out)[j] for j in fields], state)]
+            return a_out, state_out
+
+        _, state = fori_loop(0, length, body_fun, (a, state))
+
+        out = [None] * len(a)
+        for field, i in zip(fields, range(len(fields))):
+            out[field] = state[i].copy()
+
+        return tree_unflatten(a_tree, core.pack(out))
+    else:
+        return _tscan(f, a, bs, fields)
+
+
+def _tscan(f, a, bs, fields=(0,)):
     """
-    Works as jax.lax.scan but has additional `fields` arguments to select only
+    Works as jax.lax.scan but has additional `fields` argument to select only
     necessary fields from `a`'s structure. Defaults to selecting only the first
     field. Other fields will be filled by None.
     """
@@ -159,7 +189,7 @@ def _tscan_impl(a, bs, fields, consts, aval_out, jaxpr):
         a_out = core.eval_jaxpr(jaxpr, consts, (), a, core.pack(b))
         # select fields from a_out and update state
         state_out = [lax.dynamic_update_index_in_dim(s, a[None, ...], i, axis=0)
-                     for a, s in zip([tuple(a_out)[i] for i in fields], state)]
+                     for a, s in zip([tuple(a_out)[j] for j in fields], state)]
         return a_out, state_out
 
     _, state = lax.fori_loop(0, length, body_fun, (a, state))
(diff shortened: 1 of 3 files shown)


neerajprad:
- Could you also add a naive implementation that disables tracing, similar to `scan`?
 - Could you also profile just to make sure that we aren't any worse than `lax.scan`?

numpyro/util.py
```diff
@@ -104,3 +109,73 @@ def scan(f, a, bs):
         return tree_unflatten(tree_def, as_)
     else:
         return lax.scan(f, a, bs)
+
+
+def tscan(f, a, bs, fields=(0,)):
```
neerajprad: Maybe it's worth having a discussion with the JAX devs to see if they are open to integrating this upstream?

fehiepsi: The problem is this `fields` is used for the order in unflatten tree. So if `a` contains dict of dicts, users have to figure out which indices they have to put into the fields by looking at treedef. This can be generalized but I am not sure how to do it properly.

numpyro/util.py
```diff
@@ -104,3 +109,73 @@ def scan(f, a, bs):
...
+    """
+    Works as jax.lax.scan but has additional `fields` arguments to select only
+    necessary fields from `a`'s structure. Defaults to selecting only the first
+    field. Other fields will be filled by None.
```
neerajprad: Hmm..while putting in None is needed to have a sane function output and signature (e.g. you can't construct a partially filled namedtuple), I wonder if this can be more elegantly handled by specifying a transformation function between whatever pytree data structure we are using and an output data structure which can provided to `tscan`.

numpyro/util.py
neerajprad: See my comment below on taking a transformation function instead of `fields` that would work more generally. e.g. in our case, the transformation function might simply be `lambda state: state.z`.

neerajprad: I'm fine investigating this later too, so this doesn't have to be part of this PR!

numpyro/util.py
fehiepsi: That would be nice if we can do that but I can't figure out how to implement it for general pytree. I imagine given a transformation, we have to get `fields` (which are indices of required fields for output), then from the output of fori_loop, we convert it to the form we need.

fehiepsi:
@neerajprad I have added a version which disables prims. Because values converted to device array before applied the function `f`, if you want to debug, it is required to `print(x.copy())` instead of `print(x)`. I don't think that the current `scan` version works as expected so I leave it as is.

Here is the [gist](https://gist.github.com/fehiepsi/4585e45445e6087e446133a435477386) which shows that there is no regression in performance. Indeed, NUTS is a little bit faster using tscan.

numpyro/util.py
```diff
@@ -104,3 +109,103 @@ def scan(f, a, bs):
...
+        bs, b_tree = pytree_to_flatjaxtuple(bs)
+        length = tuple(bs)[0].shape[0]
+        state = [lax.full((length,) + a[i].shape, 0, lax._dtype(a[i])) for i in fields]
+
+        def body_fun(i, vals):
+            a, state = vals
+            # select i-th element from each b
+            b = [b[i] for b in bs]
+            ff, _ = pytree_fun_to_flatjaxtuple_fun(wrap_init(f), (a_tree, b_tree))
+            a_out = ff.call_wrapped(a, core.pack(b))
```
fehiepsi: Somehow we can't use this pattern 2 times; so ff is created for each iteration. In addition, there is no documentation in jax to show how to manipulate these stuffs. :(

neerajprad: Maybe I'm missing something, but this seems quite a bit more complicated. Would something like this not work?

```python
def tscan(f, a, bs, fields=(0,)):
    if _DISABLE_CONTROL_FLOW_PRIM:
        bs, pack_fn = ravel_pytree(bs)
        as_ = []
        _, tree_def = tree_flatten(bs)
        for b in bs:
            as_.append([x if i in fields else None for i, x in enumerate(f(a, b))])
        return tree_unflatten(tree_def, as_)
    else:
        return _tscan(f, a, bs)
```

Note that since this is only for debugging, we don't have to care about efficiency. Maybe you had something else in mind though?

neerajprad:
> I don't think that the current scan version works as expected so I leave it as is.

I'll take a look at this. Thanks for the heads up!

numpyro/util.py
fehiepsi: This can work for a simple version of lax.scan, where a is a simple pytree, b is a simple pytree. But it seems does not work for a simple example as in the test (I remember I get some errors on the shape mismatch).

fehiepsi:
@neerajprad The following code might be simpler and seem align with your suggestions:
```
length = tree_flatten(bs)[0][0].shape[0]
for i in range(length):
    b = tree_map(lambda x: x[i], bs)
    a = f(a, b)
    a_out = tree_map(lambda x: x[None, ...], a)

    # the following three lines are necessary for tscan;
    # it might be useful if we use `transform` instead of `fields`
    a_flat, a_tree = tree_flatten(a_out)
    a_selected = [field if i in fields else None for i, field in enumerate(a_out)]
    a_out = tree_unflatten(a_tree, a_selected)

    if i == 0:
        out = a_out
    else:
        out = tree_multimap(lambda x, y: np.concatenate((x, y)) if x is not None else None,
                            out, a_out)
out
```

neerajprad:
Looks great! We can update the naive implementation with this

PR #85, opened 2019-04-05
Add default kinetic energy

fehiepsi:
Addresses point 3 of #74 .

### DIFF
--- a/numpyro/mcmc.py
+++ b/numpyro/mcmc.py
@@ -32,7 +32,7 @@ def _sample_momentum(unpack_fn, inverse_mass_matrix, rng):
         raise NotImplementedError
 
 
-def _euclidean_ke(r, inverse_mass_matrix):
+def _euclidean_ke(inverse_mass_matrix, r):
     r, _ = ravel_pytree(r)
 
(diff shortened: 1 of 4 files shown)


numpyro/mcmc.py
```diff
@@ -32,7 +32,20 @@ def _sample_momentum(unpack_fn, inverse_mass_matrix, rng):
         raise NotImplementedError
 
 
-def hmc_kernel(potential_fn, kinetic_fn, algo='NUTS'):
+def _euclidean_ke(r, inverse_mass_matrix):
```
neerajprad: I was wondering if we should reverse the order of arguments. That way, we can apply partial with `inverse_mass_matrix` if we would like to in the future, and with the signature of both `potential_energy` and `kinetic_energy` can be the same.

fehiepsi:
> Could you add a test parameter diag_mass so that both the conditions are exercised?

How about add a todo for it? Currently momentum generator does not work with full mass matrix.

neerajprad:
> How about add a todo for it? Currently momentum generator does not work with full mass matrix.

Oh yes, please feel free to add a TODO. I forgot about that.

PR #86, opened 2019-04-05
Add ability to use models directly to do inference in HMC

### LINKED ISSUE #74: Follow-up issues for HMC

neerajprad:
Resolves #74. 

I have resorted to providing an `initialize_model` utility that returns the `potential_fn` for that model, as well as an initial trace from the prior, which can be used in `init_samples` instead. An alternative design would be to have `hmc_kernel` itself take an optional `model` arg which is used to set the potential energy and initial samples if not specified.

TODO:
 - [x] Fix `test_mcmc` that fails under JIT.
 - [x] Fix slowdown in mcmc test. (I'm getting ~10s for HMC and ~25s for NUTS, but there doesn't seem to be a regression with this PR)

### DIFF
--- a/test/test_mcmc.py
+++ b/test/test_mcmc.py
@@ -3,12 +3,14 @@ from numpy.testing import assert_allclose
 
 import jax.numpy as np
 import jax.random as random
-from jax import lax
 from jax.scipy.special import expit
 
 import numpyro.distributions as dist
 from numpyro.distributions.util import validation_disabled
+from numpyro.handlers import sample
+from numpyro.hmc_util import initialize_model
 from numpyro.mcmc import hmc_kernel
+from numpyro.util import scan
 
 
 # TODO: add test for diag_mass=False
@@ -25,7 +27,7 @@ def test_unnormalized_normal(algo):
     hmc_state = init_kernel(init_samples,
                             num_steps=10,
                             num_warmup_steps=warmup_steps)
-    hmc_states = lax.scan(lambda state, i: sample_kernel(state), hmc_state, np.arange(num_samples))
+    hmc_states = scan(lambda state, i: sample_kernel(state), hmc_state, np.arange(num_samples))
     assert_allclose(np.mean(hmc_states.z), true_mean, rtol=0.05)
     assert_allclose(np.std(hmc_states.z), true_std, rtol=0.05)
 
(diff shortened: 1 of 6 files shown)


test/test_mcmc.py
```diff
@@ -40,19 +43,18 @@ def test_logistic_regression(algo):
...
-        init_samples = np.zeros(dim)
-        hmc_state = init_kernel(init_samples,
+        hmc_state = init_kernel(init_params,
                                 step_size=0.1,
                                 num_steps=15,
                                 num_warmup_steps=warmup_steps)
-        hmc_states = lax.scan(lambda state, i: sample_kernel(state),
-                              hmc_state, np.arange(num_samples))
-        assert_allclose(np.mean(hmc_states.z, 0), true_coefs, atol=0.2)
+        sample_kernel = jit(sample_kernel)
```
neerajprad: @fehiepsi - this brings down the time for HMC from 21s to 10s, but doesn't change it for NUTS, which I think already uses JAX primitives from the outset. Might be worth looking if we are missing out on similar opportunities for optimization.

fehiepsi: my bad, I think that the reason why I don't see performance different with jit+scan and non jit + scan is that I compiled jax with gpu recently. I just uninstalled it to do profiling.

neerajprad:
> Could you add a note somewhere for the future to see if reusing the model_trace (as in Pyro) is worth?

Using the PE function directly is the most optimal, but there isn't any substantial difference between that and this verison, so this seems like a very low overhead operation. 

In any case, the reason why we store the trace and then restore it in Pyro is because `TracePosterior`'s interface requires that kernels take in a trace and return a trace. However, we wanted our utilities like `integrator` to be independent of Pyro internals like the trace data structure. We had to do this wrapping / unwrapping to deal with the external interface that requires trace and the internal interface that was designed to use dicts. Also the trace data structure in numpyro is just a simple dict, that stores a bit more info that what we do in Pyro. So wrapping/unwrapping that seems like a solution to a problem that we don't yet have! 😄

fehiepsi:
Yeah, agree! :D

test/test_mcmc.py
fehiepsi: @neerajprad Could you make a gist to show the difference between jit and no jit for this test? In all of my tries, no-jit + scan is faster than jit + scan. I have observed the inconsistent behaviour of CPU vs GPU. I am curious if jax behaves differently for different CPU systems.

neerajprad: Hmm..this is strange. I tried to recreate it but I am getting roughly the same time now (no-jit isn't faster though). I tried going back a few commits and I am still not able to see a difference, so I am guessing that it was just a noisy run earlier, or whatever was causing the overhead is resolved.

PR #87, opened 2019-04-07
Drop recursive NUTS

fehiepsi:
TODO ~in a following PR~:

- [x] Move `r_checkpoints` and `r_sum_checkpoints` outside of double_tree body function.


numpyro/hmc_util.py
```diff
@@ -332,9 +330,8 @@ def _biased_transition_kernel(current_tree, new_tree):
     return transition_prob
 
 
-@partial(jit, static_argnums=(6,))
-def _combine_tree(current_tree, new_tree, inverse_mass_matrix, going_right, rng,
-                  biased_transition, iterative_build):
+@partial(jit, static_argnums=(5,))
+def _combine_tree(current_tree, new_tree, inverse_mass_matrix, going_right, rng, biased_transition):
```
fehiepsi: We can make `biased_transition` a dynamic arg instead of static and using various `lax.cond` inside the function. I have tried it but don't see any improvement in performance. Explicitly splitting this function into 2 versions (one for biased, one for uniform) also does not add any improvement. And it turns out at the end of this work that these versions just have a small overhead over HMC.

neerajprad: Since compiled functions are cached, this function will get compiled at most twice, so I wouldn't necessarily expect that working around that would lead to any savings. Its great that you dug into this, and verified it though.

test/test_mcmc.py
```diff
@@ -39,8 +37,8 @@ def test_logistic_regression(algo):
     warmup_steps, num_samples = 1000, 8000
     data = random.normal(random.PRNGKey(0), (N, dim))
     true_coefs = np.arange(1., dim + 1.)
-    probs = expit(np.sum(true_coefs * data, axis=-1))
-    labels = dist.bernoulli(probs).rvs(random_state=random.PRNGKey(0))
+    logits = np.sum(true_coefs * data, axis=-1)
```
neerajprad: This should already be in master. Could you merge master into your branch?

fehiepsi: The master branch only changes the code in the model, not data. Either way does not matter because under the hood, logits will be converted to probs. I make this change just for consistence with the model.

PR #91, opened 2019-04-08
Add a transform function to tscan

neerajprad:
This builds on top of @fehiepsi's #84 that takes in a `transform` arg instead of `fields` and can be used more broadly to specify the output data structure in scan. An example is in the `test_util.py`. e.g. in the case of HMC, we use tscan to specify the output fields of interest:

```python
samples = tscan(lambda state, i: sample_kernel(state), hmc_state, np.arange(num_samples),
                transform=lambda x: {'z': x.z, 'potential_energy': x.potential_energy})
```

### DIFF
--- a/numpyro/util.py
+++ b/numpyro/util.py
@@ -115,7 +115,7 @@ def scan(f, a, bs):
         return lax.scan(f, a, bs)
 
 
-def tscan(f, a, bs, transform=None):
+def tscan(f, a, bs, transform):
     if not transform:
         def transform(x): return x
     if _DISABLE_CONTROL_FLOW_PRIM:
@@ -152,9 +152,9 @@ def _tscan(f, a, bs, transform=None):
 
     a, a_tree = pytree_to_flatjaxtuple(a)
     bs, b_tree = pytree_to_flatjaxtuple(bs)
-    f, out_tree = pytree_fun_to_flatjaxtuple_fun(wrap_init(f), (a_tree, b_tree))
-    transform_f, transform_tree = pytree_fun_to_flatjaxtuple_fun(wrap_init(transform),
-                                                                 (a_tree,))
+    f, _ = pytree_fun_to_flatjaxtuple_fun(wrap_init(f), (a_tree, b_tree))
+    transform_f, transform_tree = pytree_fun_to_flatjaxtuple_fun(
+        wrap_init(transform), (a_tree,))
 
     # convert arrays to abstract values
     a_aval, _ = lax._abstractify(a)
--- a/numpyro/util.py
+++ b/numpyro/util.py
@@ -203,12 +203,3 @@ tscan_p = core.Primitive('tscan')
 tscan_p.def_impl(_tscan_impl)
 tscan_p.def_abstract_eval(_tscan_abstract_eval)
 xla.translations[tscan_p] = partial(xla.lower_fun, _tscan_impl)
-
-
-def f(x, y):
-    return {"i": x["i"] + y["i"], "j": x["j"] - y["j"]}
-
-
-print(tscan(f, {"i": np.array([0.]), "j": np.array([1.])},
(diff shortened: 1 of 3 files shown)


numpyro/util.py
```diff
@@ -115,31 +115,26 @@ def scan(f, a, bs):
         return lax.scan(f, a, bs)
 
 
-def tscan(f, a, bs, fields=(0,)):
+def tscan(f, a, bs, transform):
+    if not transform:
+        def transform(x): return x
```
fehiepsi: It might be more readable if we create an `identity` function and set `transform=identity` by default.

numpyro/util.py
```diff
@@ -115,31 +115,26 @@ def scan(f, a, bs):
...
             a = f(a, b)
-            a_out = tree_map(lambda x: x[None, ...], a)
-
-            # the following three lines are necessary for tscan;
-            # it might be useful in the case you want to use `transform` instead of `fields`
-            a_flat, a_tree = tree_flatten(a_out)
-            a_selected = [field if i in fields else None for i, field in enumerate(a_out)]
-            a_out = tree_unflatten(a_tree, a_selected)
-
+            a_out = transform(tree_map(lambda x: x[None, ...], a))
```
fehiepsi: Neat! I am surprised that it is so simple. Could you add a fix for me about the case `x` is scalar here? That is to change `lambda x:...` to `lambda x: np.expand_dims(x, axis=0)`.

numpyro/util.py
```diff
@@ -169,34 +166,33 @@ def _tscan(f, a, bs, fields=(0,)):
...
+    out = tscan_p.bind(a, bs, consts, aval_out=aval_out, jaxpr=jaxpr, transform_jaxpr=transform_jaxpr)
+    return tree_unflatten(transform_tree(), out)
 
 
-def _tscan_impl(a, bs, fields, consts, aval_out, jaxpr):
+def _tscan_impl(a, bs, consts, aval_out, jaxpr, transform_jaxpr):
     length = tuple(bs)[0].shape[0]
-    state = [lax.full((length,) + a[i].shape, 0, lax._dtype(a[i])) for i in fields]
+    template = core.eval_jaxpr(transform_jaxpr, (), (), a)
+    state = [lax.full((length,) + t.shape, 0, lax._dtype(t)) for t in template]
```
fehiepsi: Whoa, this is much simpler than I thought!

numpyro/util.py
fehiepsi: More concretely, it is to resolve: `np.array(1., dtype=np.float32)[None, ...]` works but ``np.array(1.)[None, ...]` does not work as expected.

neerajprad: I saw that in `lax.scan` too. I think we should only expand for the scalar case, right?

fehiepsi: `lax.scan` should be fine because x will be a DeviceArray there (see my above comment)

neerajprad: I meant, something like this `lax.scan(f, np.array(0.), np.array([0., 1., 2.]))` throws an exception for the same reason, but it won't if you add a `dtype=float32` to the first arg. 

I think the issue is that `numpy.array(0.)` is actually a scalar array but `jax.numpy.array(0.)` is a python float without the `dtype` arg.

fehiepsi: Yes, you are right. But we might not face it when using `scan` with HMC. I remember that I get this error while using `tscan` to debug for covtype example with prims disabled, adapt_step_size=False. So `step_size` is float during the tscan. Somehow using `lax.scan` is fine because `step_size` will be converted to a DeviceArray during the way.

neerajprad: I think this is a regression in JAX - https://github.com/google/jax/issues/121, maybe? In any case, I'll put in a work-around for this.

PR #92, opened 2019-04-08
Add benchmark notebook for covtype dataset.

fehiepsi:
The number of data points is ~500000 so precision is important. Currently, HMC/NUTS does not work due to the issue https://github.com/google/jax/issues/585


fehiepsi:
@neerajprad I enable fastmath mode to improve precision for `np.sum` as discussed in google/jax#585. Now, the inference works well but 1.7x slower than edward 1. I expect that this is due to jax is slower than tensorflow for large array? I'll keep a benchmark with NUTS too.

fehiepsi:
@neerajprad I am excited to say that HMC/NUTS is faster than the ones in Edward/Edward 2. For NUTS, it is even 1.6x faster than the result announced in the reference paper (while I believe that my computer is much slower than the computer which was used to benchmark in Edward 2 paper). This is achieved by not using the primitive `lax.scan` but use our custom nonprim implementation. Using nonprim, it opens the chance for us to compile the sample_kernel method. To effectively compiling, I use the hack: feed into the hmc_state a large step_size. This way, HMC num_steps will be 1 and similar for NUTS, the tree is diverging at 1 step. So the compiling time will be fast. :D This suggests that we should revisit our API design when we have more benchmarks (e.g. allow users feed a test step_size to force fast compiling).

Note that if we use `lax.scan` with the original step_size, then 1 step of NUTS takes an hour. @@

neerajprad:
> This suggests that we should revisit our API design when we have more benchmarks (e.g. allow users feed a test step_size to force fast compiling).

That's a very good point. I think it will be ideal if we could do this behind the scenes somehow, lets discuss more.

neerajprad:
Just some minor comments:
 - `with validation disabled` shouldn't be needed now.
 - on this dataset, did you notice any benefits to running on the GPU?

fehiepsi:
Yup, I'll do benchmark in GPU too. I just want to have some overall ideas on how our implementation will work on large dataset. I remember the last time I build jax with gpu, compiling time is really slow. Will check it out again next week.

fehiepsi:
> any idea if the edward2 implementation is any different from the one in experimental?

Right now, experimental is not in the main package of tf probability, so it is not straightforward to add a replicate notebook. Could you suggest a way to properly do it? I intend to make a gist file `covtype_edward2.ipynb` and add a link to that file in this notebook. WDYT?

neerajprad:
I think since the files don't have any other dependencies if you just clone the package, you should be able to use the `experimental` module. Let me know if that isn't the case.

fehiepsi:
Hmm, I can do that but you or other readers will have to do the same for replication in your systems. That's why I suggested to have a separated gist for it.

neerajprad:
> Hmm, I can do that but you or other readers will have to do the same for replication in your systems. That's why I suggested to have a separated gist for it.

It will be hard to keep it updated though. 

How about trying to do an editable install with the commit version at the top of your notebook. You can then import from `experimental`. I think it should work fine?

```python
!pip install -e git+git@github.com:tensorflow/probability.git@7769c6c38b08bc9c5d11138d940e59c9ec104e37#egg=tensorflow-probability
```

fehiepsi:
> How about trying to do an editable install with the commit version at the top of your notebook. You can then import from experimental. I think it should work fine?

This is indeed a very clever suggestion! I'll do this way instead. Thanks @neerajprad !

fehiepsi:
@neerajprad  I have finished doing benchmark on covtype dataset. Here are some results:
+ GPU is 8x faster than CPU
+ The speed depends on num_steps of each trajectory. Average speed of long trajectory is better than of short trajectory.
+ Iterative NUTS is 1.7x faster (in both CPU and GPU) than the reported speed in reference paper. 

I think that the benefit is even more than that. Unfortunately, running the experiment in [tensorflow probability repo](https://github.com/tensorflow/probability/tree/master/experimental) is complicated. We need to install tensorflow 2.0 to be able to run it in eager mode. But anaconda does not have gpu version for tensorflow 2.0 yet. After trying out some resolutions, I give up. It is a boring work indeed to resolve these kinds of things. >"< I defer making a comparison in the same machine for interest readers, though it seems unnecessary to me. Having an idea on the performance is already enough. :)

numpyro/hmc_util.py
```diff
@@ -419,7 +419,7 @@ def _double_tree(current_tree, vv_update, kinetic_fn, inverse_mass_matrix, step_
                                         r_ckpts, r_sum_ckpts)
 
     return _combine_tree(current_tree, new_tree, inverse_mass_matrix, going_right, transition_key,
-                         biased_transition=True)
+                         True)
```
fehiepsi: This change is required for newer version of jax. See https://github.com/google/jax/pull/595 The old version of jax should work fine with this change too.

fehiepsi:
I just discover that it is easy to make a table with Markdown

|               | HMC (n=100, l=10) | HMC (n=10, l=300) | NUTS (n=10, l~300) |
| ------------- |:-----------------:|:-----------------:|:------------------:|
| Edward2 (CPU) |                   |                   |      68.4 ms       |
| Edward2 (GPU) |                   |                   |       9.7 ms       |
| Numpyro (CPU) |      43.4 ms      |      40.4 ms      |      40.5 ms       |
| Numpyro (GPU) |      19.3 ms      |       3.7 ms      |       4.8 ms       |

*Note:* Edward 2 number is obtained from reference [1].

neerajprad:
The GPU numbers are very impressive! I would take a closer look later before merging, and also the `lax.scan` variant that you are using for this.

fehiepsi:
> You mentioned benchmarking w.r.t. edward. Are the numbers from the paper, or did you replicate this on your system as well?

In some previous commits, I did the benchmark of HMC against Edward, which shows that numpyro is faster (43ms vs 54ms). But given that Edward only supports tensorflow 1.6 while the current version of tensorflow is 1.13, it might not be fair for that framework so I remove it. I will add the record to the table (by creating 2 new row Edward (CPU), Edward(GPU)) if you think that it is necessary.

> Let us fix the API issues and update the notebook, so that running NUTS with the default parameters is fast. Hopefully, we should be able to do this without putting in too many hacks in the code.
 
If we use the primitive `lax.scan` (to remove overhead of each HMC step) then I can't figure out how to make it faster unless JAX folks improve while_loop initial step behaviour. Otherwise, we can accept the overhead (which seems small to me), and return a jitted version of sample_kernel (with the fast compiling hack). Then we can use jitted_sample_kernel with tscan_non_prim (which is currently not optimized yet) or similar API.

PR #93, opened 2019-04-08
fix bugs about step size

fehiepsi:
These small bugs are detected when doing covtype benchmark.

### DIFF
--- a/numpyro/hmc_util.py
+++ b/numpyro/hmc_util.py
@@ -239,7 +239,7 @@ def warmup_adapter(num_adapt_steps, find_reasonable_step_size=None,
             else:
                 inverse_mass_matrix = np.identity(mass_matrix_size)
 
-        if find_reasonable_step_size is not None:
+        if adapt_step_size and (find_reasonable_step_size is not None):
             step_size = find_reasonable_step_size(inverse_mass_matrix, z, rng_ss, step_size)
         ss_state = ss_init(np.log(10 * step_size))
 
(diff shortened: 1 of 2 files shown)


numpyro/hmc_util.py
```diff
@@ -239,7 +239,7 @@ def init_fn(z, rng, step_size=1.0, inverse_mass_matrix=None, mass_matrix_size=No
             else:
                 inverse_mass_matrix = np.identity(mass_matrix_size)
 
-        if find_reasonable_step_size is not None:
+        if adapt_step_size and (find_reasonable_step_size is not None):
```
neerajprad: Can we not just fix the `find_reasonable_step_size` function to the default implementation and not have to do checks subsequently, i.e. add the default to the kwarg in the constructor `find_reasonable_step_size=_find..`?

neerajprad: I think it would be rare for users to provide a custom implementation for this, so we might as well just use the default which they can override if they wish.

fehiepsi:
@neerajprad I have addressed your comment. Travis has already passed but it keeps pending signal.

numpyro/hmc_util.py
```diff
@@ -239,7 +243,7 @@ def init_fn(z, rng, step_size=1.0, inverse_mass_matrix=None, mass_matrix_size=No
             else:
                 inverse_mass_matrix = np.identity(mass_matrix_size)
 
-        if find_reasonable_step_size is not None:
+        if adapt_step_size:
```
neerajprad: Nice!

PR #94, opened 2019-04-09
Add support for constraint

fehiepsi:
This PR adds various constraints/transforms which is needed for some benchmarks.

Todo:

- [x] Add tests

Todo in following PRs:
- Incorporate into distributions
- Use support information to define transforms in inference.

### DIFF
--- a/numpyro/distributions/constraint_registry.py
+++ b/numpyro/distributions/constraint_registry.py
@@ -1,0 +1,45 @@
+# The implementation follows the design in PyTorch: torch.distributions.constraint_registry.py
+
+
+class ConstraintRegistry(object):
+    def __init__(self):
+        self._registry = {}
+
+    def register(self, constraint, factory):
+        if isinstance(constraint, constraints.Constraint):
+            constraint = type(constraint)
+
+        self._registry[constraint] = factory
+
+    def __call__(self, constraint):
+        try:
+            factory = self._registry[type(constraint)]
+        except KeyError:
+            raise NotImplementedError
+        return factory(constraint)
+
+
+biject_to = ConstraintRegistry()
+
+
+@biject_to.register(constraints.greater_than)
+def _transform_to_greater_than(constraint):
+    return transforms.ComposeTransform([transforms.ExpTransform(),
+                                        transforms.AffineTransform(constraint.lower_bound, 1)])
+
+
+@biject_to.register(constraints.interval)
+def _transform_to_interval(constraint):
+    return transforms.ComposeTransform([transforms.SigmoidTransform(),
+                                        transforms.AffineTransform(constraint.lower_bound,
+                                                                   constraint.upper_bound)])
+
+
+@transform_to.register(constraints.real)
+def _transform_to_real(constraint):
+    return transforms.IdentityTransform
+
+
+@biject_to.register(constraints.simplex)
+def _transform_to_simplex(constraint):
+    return transforms.StickBreakingTransform()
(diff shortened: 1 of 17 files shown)


numpyro/distributions/transforms.py
```diff
@@ -0,0 +1,157 @@
...
+
+    @property
+    def codomain(self):
+        if self.domain is constraints.real:
+            return constraints.real
+        elif isinstance(self.domain, constraints.greater_than):
+            return constraints.greater_than(self.loc + self.scale * self.domain.lower_bound)
+        elif isinstance(self.domain, constraints.interval):
+            return constraints.interval(self.loc + self.scale * self.domain.lower_bound,
+                                        self.loc + self.scale * self.domain.upper_bound)
```
fehiepsi: this way, we can support AffineTransform for more classes of distributions. Currently, the one in PyTorch just supports real domain.

test/test_distributions.py
```diff
@@ -270,3 +273,78 @@ def test_discrete_with_logits(jax_dist, dist_args):
...
+
+    # test log_abs_det_jacobian
+    actual = transform.log_abs_det_jacobian(x, y)
+    assert np.shape(actual) == batch_shape
+    if len(shape) == transform.event_dim:
+        if constraint is constraints.simplex:
+            return
+            # TODO compare against numerical values when jvp rule for `cumsum`, `cumprod` is available
+            # expected = np.linalg.slogdet(jax.jacobian(transform)(x)[::-1, :])[1]
+            # inv_expected = np.linalg.slogdet(jax.jacobian(transform.inv)(y)[:, ::-1])[1]
```
fehiepsi: currently, jvp rule for cumsum is not available yet. I have compared the result against PyTorch.

numpyro/distributions/constraint_registry.py
```diff
@@ -0,0 +1,45 @@
+# The implementation follows the design in PyTorch: torch.distributions.constraint_registry.py
```
neerajprad: I think it will be a good idea to also add PyTorch license/copyright to the top of these files.

test/test_distributions.py
neerajprad: Hmm..seems like we'll need this for dirichlet, but it should't be that complicated to implement.

numpyro/distributions/transforms.py
neerajprad: Nice, we should also add this to PyTorch!

test/test_distributions.py
fehiepsi: Yes, we'll need to implement a custom primitive for this. For cumsum, it seems not complicated because the jacobian is simply a upper triangular matrix will all entries are 1. It is just a bit more tricky for cumprod. I will make an issue for this.

numpyro/distributions/transforms.py
fehiepsi: PyTorch licence includes [many copyrights](https://github.com/pytorch/pytorch/blob/master/LICENSE). Which one would you suggest?

numpyro/distributions/constraint_registry.py
fehiepsi: PyTorch licence includes [many copyrights](https://github.com/pytorch/pytorch/blob/master/LICENSE). Which one would you suggest?

fehiepsi:
@neerajprad I'll merge this first to unblock #98 and will add copyrights in that PR based on your suggestion. :)

numpyro/distributions/constraint_registry.py
neerajprad: We can add the copyright for PyTorch, and the conditions towards the end, i.e. just leave out the caffe2 stuff.

fehiepsi: Yup, already did it!

Issue #82, opened 2019-04-05
Remove dependency on scipy.stats for distributions


neerajprad:
Due to issues faced in #51, #50 and #81, it seems that we would be better off not inheriting from scipy and having to work around their class hierarchy which was designed to support its own use cases. Instead, I think it will help if our design was closer to PyTorch, in terms of Pyro <--> Numpyro compatibility (#66), and more generally reuse of functionality like constraint checks and bijective transforms that are useful for HMC and SVI. 

I think to begin with we can start by only supporting `sample`, `log_prob` methods and adding more as we go along. It should be relatively straightforward to do so for our existing implementations. In cases where we don't have a JAX sampler available, we could just use scipy's sampler internally (as suggested by @fehiepsi).

neerajprad:
@fehiepsi - Let me know your thoughts. I think we should address this soon, so that we don't pile on more tech debt. I can take a look at it.

fehiepsi:
> I think to begin with we can start by only supporting sample, log_prob methods and adding more as we go along.

Agree. Currently, we only use rvs/logpdf for doing inference so there is no need to get inherited from scipy. I prefer to keep our current api (for rvs/logdf/frozen) and just rewrite the __init__ method so that it only initializes things we need (AFAIK, we only need the `parse_args` method to automatically parse args). Other infos such as `momtype=1, a=None, b=None, xtol=1e-14, badvalue=None, longname=None, extradoc=None, seed=None` are not necessary right now. We can add more functionalities from scipy or from our own usage later.

neerajprad:
>  I prefer to keep our current api (for rvs/logdf/frozen) and just rewrite the init method so that it only initializes things we need (AFAIK, we only need the parse_args method to automatically parse args).

I think this is worth discussing and fleshing out a bit. I used scipy to begin with as it was the choice of least resistance, but I am not convinced that following any of the design patterns from scipy will be a good long term plan. :) 

My thought is that there is little benefit in maintaining frozen / unfrozen distribution objects, and significant maintenance cost. A lot of the convoluted code that we have inherited stems precisely from having to maintain both of these patterns.  

[...]

PR #95, opened 2019-04-10
Add distribution classes that do not depend on scipy

neerajprad:
Initial attempt at #82. Note that this doesn't refactor / change existing behavior in any way.

This adds support for a few distribution classes (normal, uniform and bernoulli) in `contrib.distributions`. Note that these classes are tested but they are not being used anywhere. Once this module is fully developed and we are happy with the interface, we can move these distributions into the main `distributions` module.

I have kept the API largely similar to that of `torch.distributions`. This will help integrate with `funsor` / `pyro`. More importantly it will be easy to do args validation / use constraints etc by following the same patterns as in PyTorch.

### DIFF
--- a/numpyro/distributions/util.py
+++ b/numpyro/distributions/util.py
@@ -1,9 +1,11 @@
 from contextlib import contextmanager
+from numbers import Number
 
+import numpy as onp
 import scipy.special as osp_special
 
 import jax.numpy as np
-from jax import jit, lax, random, vmap
+from jax import canonicalize_dtype, jit, lax, random, vmap
 from jax.core import Primitive
 from jax.interpreters import ad, partial_eval, xla
 from jax.numpy.lax_numpy import _promote_args_like, _promote_shapes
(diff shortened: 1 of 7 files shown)


numpyro/distributions/util.py
```diff
@@ -290,6 +292,11 @@ def promote_shapes(*args, shape=()):
                 if len(s) < num_dims else arg for arg, s in zip(args, shapes)]
 
 
+def get_dtypes(*args):
+    return [canonicalize_dtype(type(arg)) if isinstance(arg, Number)
+            else canonicalize_dtype(onp.dtype(arg)) for arg in args]
```
fehiepsi: I am still not familiar with how jax manages dtypes. Somewhere I see `lax._dtype`, somewhere I see `canonicalize_dtype`. >"<

neerajprad: @fehiepsi - Based on my limited exploration, my conclusion was that `canonicalize_dtype` will take care of default conversions, but can't handle python scalars. I just realized that this function can be simplified to `canonicalize_dtype(onp.result_type(arg))` (will do that in a follow up PR). `lax._dtype` just delegates to `onp.result_type`.

fehiepsi: Thanks for explaining!

PR #98, opened 2019-04-10
Add constraints to continuous distribution

### LINKED ISSUE #80: Add argcheck for continous distributions
Currently, only discrete distributions support argcheck. We should do it for continuous distributions too.

fehiepsi:
Resolves #80. This PR adds constraints/supports to continuous distributions. By introducing `jax_frozen` and `jax_generic`, I hope this PR will resolve some problems related to args.

I have kept the semantic of _support_mask/_argcheck as in scipy.stats, which are:
+ _support_mask: support of the distribution with loc=0, scale=1
+ _argcheck: no check for scale parameter

If the proposed solution in this PR is accepted, let's do the same for discrete and multivariate distributions.


fehiepsi:
@neerajprad Not related to this PR, I think that it is better to group tests for continuous distribution together, tests for discrete together,... so every time we want to add a distribution, we know what to do based on similar ones.

fehiepsi:
Hmm, HMC/NUTS tests are failing. I thought validate_args is False by default. I'll investigate why.

neerajprad:
> I think that it is better to group tests for continuous distribution together, tests for discrete together

I think we are already doing that? Feel free to organize them together as you see fit. For the distributions in contrib, the interface for both is the same (i.e. they inherit from the same class and have the same behavior unlike in scipy), so we don't need to separate them out for most tests (e.g. `log_prob`, `shape`, `sample` tests etc.). The fixtures for the two are separate, so `test_rsample` needs to only be exercised for continuous distributions.

numpyro/distributions/constraint_registry.py
```diff
@@ -1,4 +1,14 @@
...
+#
+# Copyright (c) 2016-     Facebook, Inc            (Adam Paszke)
+# Copyright (c) 2014-     Facebook, Inc            (Soumith Chintala)
+# Copyright (c) 2011-2014 Idiap Research Institute (Ronan Collobert)
+# Copyright (c) 2012-2014 Deepmind Technologies    (Koray Kavukcuoglu)
+# Copyright (c) 2011-2012 NEC Laboratories America (Koray Kavukcuoglu)
+# Copyright (c) 2011-2013 NYU                      (Clement Farabet)
+# Copyright (c) 2006-2010 NEC Laboratories America (Ronan Collobert, Leon Bottou, Iain Melvin, Jason Weston)
+# Copyright (c) 2006      Idiap Research Institute (Samy Bengio)
+# Copyright (c) 2001-2004 Idiap Research Institute (Ronan Collobert, Samy Bengio, Johnny Mariethoz)
```
neerajprad: The license also asks to add the disclaimer: `THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT...`

fehiepsi:
> I think we are already doing that?

Sorry, I think that I didn't make a good explanation. I mean to group continuous distributions sample+logpdf+validate_args tests to the same section in the file, and similarly for discrete ones.

numpyro/distributions/distribution.py
```diff
@@ -9,16 +9,83 @@
...
+
+    # TODO: move the implementation of _construct_argparser in jax_mvcontinuous
+    # to here if everything works.
+
+
+class jax_continuous(jax_generic, osp_stats.rv_continuous):
+    def _support(self, *args, **kwargs):
+        # support of the transformed distribution
+        _, loc, scale = self._parse_args(*args, **kwargs)
+        return AffineTransform(loc, scale, domain=self._support_mask).codomain
```
neerajprad: This seems like a round-about way of doing this, but it is fine since this will be refactored soon.

numpyro/distributions/distribution.py
```diff
@@ -9,16 +9,83 @@
...
+            if not np.all(scale > 0):
+                raise ValueError('Invalid scale parameter provided to the distribution.')
+
+        self.a, self.b = self.dist.a, self.dist.b
+
+    @property
+    def support(self):
+        return self.dist._support(*self.args, **self.kwds)
+
+    def __call__(self, size=None, random_state=None):
```
neerajprad: Could you explain why this is needed?

fehiepsi: This is no need given that we have patched scipy implementation. But doing this way, we don't need to patch scipy. I intentionally put it here to get your notice. ;)

PR #99, opened 2019-04-12
Unify args validation for discrete and continuous distributions

neerajprad:
This refactor addresses the following:
 - Add some more constraints for discrete distributions.
 - Removes existing `_args_check` logic to instead use the `_validate_args` checks from @fehiepsi's earlier PRs #98. 
 - Changes the default behavior to not validate (this is something that can be changed later). Changes the `validation_disabled` contextmanager to `validation_enabled` and moves this to `distribution.py` so as to avoid circular dependencies.

Right now, some dependent constraints cannot be checked, e.g. for `logits` vs `probs`. That will require these properties to be set at the instance level, and it gets complicated with the existing distributions hierarchy. Also note that this doesn't handle the multinomial distribution which is complicated for similar reasons. All of these should be addressed once #82 is fixed. 

Fixes #50.

### DIFF
--- a/numpyro/distributions/constraints.py
+++ b/numpyro/distributions/constraints.py
@@ -31,9 +31,6 @@ class Constraint(object):
 
 
 class _Boolean(Constraint):
-    """
-    Constrain to the two values `{0, 1}`.
-    """
     def __call__(self, value):
         return (value == 0) | (value == 1)
 
--- a/numpyro/distributions/constraints.py
+++ b/numpyro/distributions/constraints.py
@@ -30,6 +30,14 @@ class Constraint(object):
         raise NotImplementedError
 
 
+class _Boolean(Constraint):
+    """
+    Constrain to the two values `{0, 1}`.
+    """
+    def __call__(self, value):
+        return (value == 0) | (value == 1)
+
+
 class _GreaterThan(Constraint):
     def __init__(self, lower_bound):
         self.lower_bound = lower_bound
@@ -45,6 +53,14 @@ class _IntegerInterval(Constraint):
 
     def __call__(self, x):
         return (x >= self.lower_bound) & (x <= self.upper_bound) & (x == np.floor(x))
+
+
+class _IntegerGreaterThan(Constraint):
+    def __init__(self, lower_bound):
+        self.lower_bound = lower_bound
+
+    def __call__(self, value):
+        return (value % 1 == 0) & (value >= self.lower_bound)
 
 
 class _Interval(Constraint):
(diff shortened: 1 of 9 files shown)


numpyro/distributions/discrete.py
```diff
@@ -33,6 +37,9 @@ def _entropy(self, n, p):
 
 
 class _bernoulli_gen(jax_discrete, bernoulli_gen):
+    # TODO: This should depend on initialization
+    arg_constraints = {'p': constraints.real}
```
fehiepsi: To me, the simplest solution right now is to create a new class, named for example: bernoulli_with_logits. ^^! WDYT?

numpyro/distributions/distribution.py
```diff
@@ -127,12 +128,13 @@ def logpdf(self, x, *args, **kwargs):
...
 
-class jax_discrete(osp_stats.rv_discrete):
-    args_check = True
-
+class jax_discrete(jax_generic, osp_stats.rv_discrete):
     def _support_mask(self, k):
         return (k >= self.a) & (k <= self.b) & (np.floor(k) == k)
 
+    def _support(self, *args, **kwargs):
+        return self._support_mask
```
fehiepsi: Though it is not important because we don't want to work with `loc!=0`, could you add a TODO comment here to remind us in case this scipy interface works well on its own?

numpyro/distributions/constraints.py
```diff
@@ -67,9 +83,12 @@ def __call__(self, x):
         return np.all(x > 0, axis=-1) & (x_sum <= 1) & (x_sum > 1 - 1e-6)
 
 
+boolean = _Boolean()
```
fehiepsi: Nit: we can make this _IntegerInterval(0, 1)

numpyro/distributions/discrete.py
neerajprad: That would be one way to do it, but there are many cases where we can only do validation after initialization, e.g. for binomial, where support is in `[0, n]`. Also the multivariate cases are a mess to handle, so I think it would be preferable prioritizing #82 than invest more time in working around these issues. 

For the new distribution classes, I currently have `probs` and `logits` as separate parametrizations for the same reason, and it is worth thinking if it is better to separate out the implementations.

fehiepsi: About support, I think we just mimic the way we did in continuous distribution: get arg and verify. I think it is what PyTorch does (except @lazy decorator). For special case like binomial, we can just overwrite that `_support` property I guess.

numpyro/distributions/constraints.py
neerajprad: The issue is that it will be much slower, almost 5X or so.

numpyro/distributions/discrete.py
neerajprad: `_support` needs to be defined at the correct place (not in gen, but rather in frozen) because it needs to be treated as a property and not something into which arguments will be passed into. In any case, this isn't an issue with the classes in `contrib`, so I'm just focusing on that for the moment.

fehiepsi: Ah, I mean to modify the _support method in gen. Anyway, I’ll do that instead. :)

PR #100, opened 2019-04-12
Separating out tscan_nonprim to make it easier for benchmarking

fehiepsi:
For large model (e.g. logistics regression in the covtype benchmark), using the primitive `tscan` is very slow. This PR separates out nonprim implementation to make it easier to do benchmark.

### DIFF
--- a/numpyro/util.py
+++ b/numpyro/util.py
@@ -121,19 +121,23 @@ def _identity(x):
 
 def tscan(f, a, bs, transform=_identity):
     if _DISABLE_CONTROL_FLOW_PRIM:
-        length = tree_flatten(bs)[0][0].shape[0]
-        for i in range(length):
-            b = tree_map(lambda x: x[i], bs)
-            a = f(a, b)
-            a_out = transform(tree_map(lambda x: np.expand_dims(x, axis=0), a))
-            if i == 0:
-                out = a_out
-            else:
-                out = tree_multimap(lambda x, y: np.concatenate((x, y)) if x is not None else None,
-                                    out, a_out)
-        return out
+        tscan_nonprim(f, a, bs, transform=_identity)
     else:
         return _tscan(f, a, bs, transform)
+
+
+def tscan_nonprim(f, a, bs, transform=_identity):
+    length = tree_flatten(bs)[0][0].shape[0]
+    for i in range(length):
+        b = tree_map(lambda x: x[i], bs)
+        a = f(a, b)
+        a_out = transform(tree_map(lambda x: np.expand_dims(x, axis=0), a))
+        if i == 0:
+            out = a_out
+        else:
+            out = tree_multimap(lambda x, y: np.concatenate((x, y)) if x is not None else None,
+                                out, a_out)
+    return out
 
 
 def _tscan(f, a, bs, transform=None):


numpyro/util.py
```diff
@@ -121,21 +121,25 @@ def _identity(x):
...
-            b = tree_map(lambda x: x[i], bs)
-            a = f(a, b)
-            a_out = transform(tree_map(lambda x: np.expand_dims(x, axis=0), a))
-            if i == 0:
-                out = a_out
-            else:
-                out = tree_multimap(lambda x, y: np.concatenate((x, y)) if x is not None else None,
-                                    out, a_out)
-        return out
+        tscan_nonprim(f, a, bs, transform=_identity)
```
neerajprad: nit: append underscore to deter usage outside this file.

neerajprad:
> using the primitive tscan is very slow.

What is the version that you used for the benchmark? Is it too much of a hack, or is there some suitable version that can be made available for use in the meantime? :)

numpyro/util.py
fehiepsi: Do you mean `_tscan_nonprim`? Indeed, I use this implementation directly (not through tscan because adding a nonprim context manager will force other primitives become nonprim too). ^^!

fehiepsi:
I use `tscan_nonprim` for doing benchmark. The only hack is to add the following two lines before sampling
```
jsample_kernel = jit(sample_kernel)
jsample_kernel(hmc_state.update(step_size=1.))  # or larger step_size if needed
```

numpyro/util.py
neerajprad: Hmm..I thought you were running some other custom implementation. :) So you are saying that this naive version is faster than the actual `lax.scan` version? Why is that the case - is it just due to the initial small step size?

fehiepsi:
Btw, not related, could we rename `init_kernel`, `sample_kernel` to `init_fn`, `sample_fn` and for usage outside implementation, we can name it `hmc_init`, `hmc_update`? The name `init_kernel` or `sample_kernel` looks like we are creating a kernel or sampling a kernel.

numpyro/util.py
fehiepsi: > is it just due to the initial small step size?

yup, I think so. `jit(sample_kernel)(hmc_sate)` took more than 1 hour in my system while `jit(sample_kernel)(hmc_state.update(step_size=1.))` took me 15s. >"<

neerajprad: I'm fine with merging this then, but would like to find a way to resolve this within `tscan` itself finally.

neerajprad:
> The name init_kernel or sample_kernel looks like we are creating a kernel or sampling a kernel.

I think this nomenclature is related to sampling from the transition kernel in MCMC, which is what we are doing. It seems more precise than `update_fn`.

numpyro/util.py
fehiepsi: I think we can use `tscan` for large models again when issues with first step of while_loop is resolved. :)

Hmm, I just thought that my suggestions are not good. I should make a custom tscan in my notebook instead, where I can add print statements too. Sorry for wandering around!

fehiepsi:
Yes, I understand that you want to mean it is a kernel to do sampling. But I think that having `hmc_kernel`, `init_kernel`, `sample_kernel` might cause confusion for other users. Anyway, that's just minor to me because I can just rename these functions for my convenience.

neerajprad:
> I think that having hmc_kernel, init_kernel, sample_kernel might cause confusion for other users. 

I think we can rename `hmc_kernel` to simply `hmc`, and `sample_kernel` to `transition_kernel`, if that helps (I'm not sure if `hmc_update` is any more informative, but I don't have a strong opinion either ways). I think the idea of doing MCMC by starting from some initial sample, and transitioning based on a transition kernel is quite intuitive otherwise.

fehiepsi:
> I think we can rename hmc_kernel to simply hmc, and sample_kernel to transition_kernel

This is a much better idea! I don't mean to use `update` instead of `sample`. `sample_fn` is a better name than `sample_kernel` to me. But `transition_kernel` is a pretty good name and together with renaming `hmc_kernel` to `hmc`, they resolve confusions which I have! I vote for it. :)

PR #101, opened 2019-04-12
Fix various issues of scipy discrete distributions

### LINKED ISSUE #51: Add logits support to discrete distributions

fehiepsi:
This PR does the following

- [x] Support `loc` in `_support_mask`
- [x] Let the api of jax_discrete matches jax_continuous
- [x] Support a flag `is_logits` for frozen instances, including binomial
- [x] Add a stable version for logpmf of binomial based on logits.
- [x] Remove scipy's patch
- [x] Fix tests

Fix #51


numpyro/distributions/distribution.py
```diff
@@ -129,24 +130,49 @@ def logpdf(self, x, *args, **kwargs):
 
 
 class jax_discrete(jax_generic, osp_stats.rv_discrete):
-    def _support_mask(self, k):
-        return (k >= self.a) & (k <= self.b) & (np.floor(k) == k)
+    def __new__(cls, *args, **kwargs):
+        return super(jax_discrete, cls).__new__(cls)
+
+    def __init__(self, *args, **kwargs):
+        self.is_logits = kwargs.pop("is_logits", False)
```
neerajprad: Storing this for all distributions, i.e. those that don't even take in a probability param seems like a hack, but I guess there isn't a cleaner alternative here.

numpyro/distributions/distribution.py
```diff
@@ -129,24 +130,49 @@ def logpdf(self, x, *args, **kwargs):
...
+        super(jax_discrete, self).__init__(*args, **kwargs)
+
+    def freeze(self, *args, **kwargs):
+        self._ctor_param.update(is_logits=kwargs.pop("is_logits", False))
+        return super(jax_discrete, self).freeze(*args, **kwargs)
 
-    # TODO: This will not work for the case that loc != 0
     def _support(self, *args, **kwargs):
```
neerajprad: For my understanding, does `dist.binom.support(p=.., n=..)` also work or is this only meant to be used with frozen distributions?

fehiepsi: The support property is only used in frozen distribution, where all parameters are known: loc, scale, n, p,... I guess using `dist.binorm._support(p=..., n=...)` also works.

Issue #107, opened 2019-04-14
[discussion] use transforms in potential_fn


fehiepsi:
I would like to suggest a way to use support information for default transforms and incorporate it in potential_fn:
+ initialize_model takes an additional argument named transforms, which is a dict map each latent variable to its user-defined transform. Otherwise, we use the default transform.
+ initialize model returns a z transform fn (in addition to init unconstrained param and potential_fn) for tranforming unconstrained variables to contrained variable. So users can use this transform fn in transform argument of tscan: `tscan(..., transform=lambda latent: z_tranform_fn(latent.z))`

neerajprad:
We could handle the default transforms for both SVI and MCMC via an effect handler, the only difference being that for SVI it affects only the `param` statements whereas for MCMC it would transform `sample` statements. Another thing to note is that just transforming in the outside loop is not be enough, since this also changes the PE computation.

Regarding:
> initialize_model takes an additional argument named transforms.

In practice, I haven't seen models where users had to override the default transforms. I can imagine that providing a way to use custom transforms might be useful for certain models. If we provide that, it shouldn't come at the cost of users having to write boiler-plate code for models where they would just use the default transforms anyway. Let me put up something, and we can discuss the interface issues in the PR.

neerajprad:
Actually, on second thoughts, I think your approach above would be the simplest since HMC in numpyro doesn't have access to the model directly.

PR #108, opened 2019-04-15
Use transforms for models with constrained latent variables

neerajprad:
This is based off of @fehiepsi's suggestion in #107. 
 - The nice thing is that no changes are needed within the `hmc` implementation, since that only takes in a potential energy computation. So this only involves a change in `potential_energy` computation and the `initialize_model` functions.
 - The not so nice thing is that the user needs to write some boiler-plate code to transform the latent sites to constrained space, but this doesn't seem too bad, and besides this is unavoidable with the current API where HMC does not access the model directly. We could, in the future, provide some convenience wrappers that could take in a model callable directly to wrap over the boiler-plate code.
 - Made some changes to `tscan` so that the non-primitive version works with more complex transforms.
 - Added a `clipped_expit` without which the beta-bernoulli model gives NaNs.

TODO: When `.support` works with multivariate distributions, I will remove the `xfail` marker from `test_dirichlet_categorical`. I intend to fix this directly in `contrib.distributions`.

### DIFF
--- a/numpyro/handlers.py
+++ b/numpyro/handlers.py
@@ -94,7 +94,7 @@ def apply_stack(msg):
         if msg.get("stop"):
             break
     if msg['value'] is None:
-        msg['value'] = msg['fn'](*msg['args'], **msg['kwargs'])
+        msg['value'] = msg['fn'].rvs(*msg['args'], **msg['kwargs'])
 
     # A Messenger that sets msg["stop"] == True also prevents application
     # of postprocess_message by Messengers above it on the stack
--- a/numpyro/hmc_util.py
+++ b/numpyro/hmc_util.py
@@ -557,10 +557,10 @@ def log_density(model, model_args, model_kwargs, params):
 def potential_energy(model, model_args, model_kwargs, transforms):
     def _potential_energy(params):
         params_constrained = {k: transforms[k](v) for k, v in params.items()}
-        log_joint = jax.partial(log_density, model, model_args, model_kwargs)(params)[0]
+        log_joint = jax.partial(log_density, model, model_args, model_kwargs)(params_constrained)[0]
         for name, t in transforms.items():
             log_joint = log_joint + np.sum(t.log_abs_det_jacobian(params[name], params_constrained[name]))
-        return log_joint
+        return - log_joint
 
     return _potential_energy
 
--- a/numpyro/hmc_util.py
+++ b/numpyro/hmc_util.py
@@ -6,6 +6,7 @@ from jax.ops import index_update
 from jax.scipy.special import expit
 from jax.tree_util import tree_multimap
 
+from numpyro.distributions import biject_to
 from numpyro.distributions.distribution import jax_continuous
 from numpyro.handlers import seed, substitute, trace
 from numpyro.util import cond, laxtuple, while_loop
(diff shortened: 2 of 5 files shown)


fehiepsi:
Nice! Just having two comments on the design:
+ It might be better to have a transform fn out-of-box for z instead of a dict of invidual items in z. This makes the code in tscan simpler.
+ According to the tests, we define both step_size and num_steps. These are unnecessary for NUTS and step_size is unnecessary for HMC (with adaptation). How about we use only trajectory length for init kernel? Num_steps will automatically change when step_size changes, which can cause confusion on its purpose.

> I intend to fix this directly in contrib.distributions.

I already have some works in scipy version. Will fix there too.

neerajprad:
> It might be better to have a transform fn out-of-box for z instead of a dict of invidual items in z. This makes the code in tscan simpler.

Done.

> According to the tests, we define both step_size and num_steps. These are unnecessary for NUTS and step_size is unnecessary for HMC (with adaptation). 

I agree. In practice though, HMC doesn't converge fast enough, and NUTS seems to take forever with the default values for `step_size` and `num_steps`. This is definitely something that we should fix, but it is worth a separate PR (and more thought on how to make HMC/NUTS fast by default based on your lessons from #92). 

> How about we use only trajectory length for init kernel? Num_steps will automatically change when step_size changes, which can cause confusion on its purpose.

I think it is fine to use `step_size` and `trajectory_length` as initialization parameters and dropping `num_steps`. This is an independent issue though that we can address in a small separate PR.

numpyro/handlers.py
```diff
@@ -94,7 +94,7 @@ def apply_stack(msg):
         if msg.get("stop"):
             break
     if msg['value'] is None:
-        msg['value'] = msg['fn'](*msg['args'], **msg['kwargs'])
+        msg['value'] = msg['fn'].rvs(*msg['args'], **msg['kwargs'])
```
fehiepsi: Is this change caused by the drop of scipy's patch?

numpyro/util.py
```diff
@@ -121,12 +121,12 @@ def _identity(x):
 
 def tscan(f, a, bs, transform=_identity):
     if _DISABLE_CONTROL_FLOW_PRIM:
-        _tscan_nonprim(f, a, bs, transform=_identity)
+        return _tscan_nonprim(f, a, bs, transform)
```
fehiepsi: thanks for the fix!

numpyro/hmc_util.py
```diff
@@ -553,13 +554,27 @@ def logp(d, val):
...
+        for name, t in transforms.items():
+            log_joint = log_joint + np.sum(t.log_abs_det_jacobian(params[name], params_constrained[name]))
+        return - log_joint
+
+    return _potential_energy
+
+
+def transform_fn(transforms, params, invert=False):
+    return {k: transforms[k](v) if not invert else transforms[k].inv(v)
+            for k, v in params.items()}
```
fehiepsi: The logic of transforms and log_ab_det_jac looks correct to me! It might be worth to add a comment (to avoid the confusion with transforms belong to in Transform class) about the convention that: transform is from constrained to unconstrained, and inv transform is from unconstrained to constrained. But this can be added to docs later instead.

numpyro/handlers.py
neerajprad: This is because the patch doesn't address frozen instances from multivariate distributions, and I didn't want to add to the `patch` hack. :)

numpyro/hmc_util.py
neerajprad: Good point. Will remember to add this to the docs!

numpyro/handlers.py
fehiepsi: Understood! Will fix it too. :)

PR #111, opened 2019-04-17
Add Categorical distribution

fehiepsi:
This PR does
- [x] Make multivariate a subclass of jax_generic
- [x] Address comment for tests of cumsum, cumprod from @neerajprad 
- [x] Move multinomial distribution to multivariate.py
- [x] Add Categorical distribution
- [x] Add multinomial sampler based on Categorical sampler, so it can take general shape for `p` (numpy only supports 1D array p)
- [x] Add tests and fix bugs.


test/test_mcmc.py
```diff
@@ -67,37 +67,34 @@ def model(data):
         return p_latent
 
     true_probs = np.array([0.9, 0.1])
-    data = dist.bernoulli(true_probs).rvs(size=(1000, 2), random_state=random.PRNGKey(0))
+    data = dist.bernoulli(true_probs).rvs(size=(1000, 2), random_state=random.PRNGKey(1))
```
fehiepsi: to avoid duplicated key with hmc

test/test_mcmc.py
```diff
@@ -67,37 +67,34 @@ def model(data):
...
     init_params, potential_fn, transform_fn = initialize_model(random.PRNGKey(2), model, (data,), {})
     init_kernel, sample_kernel = hmc(potential_fn, algo=algo)
     hmc_state = init_kernel(init_params,
-                            step_size=0.1,
                             trajectory_length=1.,
                             num_warmup_steps=warmup_steps)
     hmc_states = tscan(lambda state, i: sample_kernel(state), hmc_state, np.arange(num_samples),
                        transform=lambda x: transform_fn(x.z))
-    assert_allclose(np.mean(hmc_states['p_latent'], 0), true_probs, rtol=0.05)
+    assert_allclose(np.mean(hmc_states['p_latent'], 0), true_probs, atol=0.05)
```
fehiepsi: to match pyro test

numpyro/distributions/multivariate.py
```diff
@@ -7,27 +7,80 @@
...
+
+    def _support(self, *args, **kwargs):
+        (p,), _, _ = self._parse_args(*args, **kwargs)
+        return constraints.integer_interval(0, p.shape[-1] - 1)
+
+    def _batch_shape(self, p):
+        return p.shape[:-1]
+
+    def _event_shape(self, p):
+        return ()
```
fehiepsi: `event_shape` is not required but it is useful for testing, so I add it just for convenience.

numpyro/distributions/distribution.py
```diff
@@ -217,23 +216,16 @@ def rvs(self, *args, **kwargs):
         size = kwargs.pop('size', args.pop() if len(args) > self.numargs else None)
 
         args, _, _ = self._parse_args(*args, **kwargs)
-        # XXX we might not need to verify that args is empty for multivariate distributions
-        args = _promote_args("rvs", *args)
 
-        # TODO: make this code compatible to mvn distribution
         if not size:
-            size = args[-1].shape[:-1]
+            size = self._batch_shape(*args)
```
neerajprad: I am curious to know what is the functionality of this class? If it is just this behavior of adjusting size to account for the event shape, can we not do this in `jax_discrete` and `jax_continuous` now that you have already defined `event_shape` and `batch_shape`.

test/test_distributions_util.py
```diff
@@ -83,14 +93,17 @@ def test_binary_cross_entropy_with_logits(x, y):
...
+    (5, 3),
 ])
 def test_cumsum_jac(shape):
     rng = random.PRNGKey(0)
     x = random.normal(rng, shape=shape)
-    expected = grad(lambda x: np.sum(x * (np.arange(x.shape[-1]) + 1.)[::-1]))(x)
-    actual = grad(lambda x: np.sum(cumsum(x)))(x)
-    assert_allclose(expected, actual)
+
+    def test_fn(x):
```
neerajprad: Looks great! Thanks for these changes.

test/test_mcmc.py
```diff
@@ -67,37 +67,34 @@ def model(data):
         return p_latent
 
     true_probs = np.array([0.9, 0.1])
-    data = dist.bernoulli(true_probs).rvs(size=(1000, 2), random_state=random.PRNGKey(0))
+    data = dist.bernoulli(true_probs).rvs(size=(1000, 2), random_state=random.PRNGKey(1))
     init_params, potential_fn, transform_fn = initialize_model(random.PRNGKey(2), model, (data,), {})
     init_kernel, sample_kernel = hmc(potential_fn, algo=algo)
     hmc_state = init_kernel(init_params,
```
neerajprad: This seems to take a really long time for NUTS on my system with the default `step_size`. How long does this test take to terminate on your system?

numpyro/distributions/util.py
```diff
@@ -326,6 +326,37 @@ def binomial(key, p, n=1, shape=()):
     return np.sum(mask * lax.lt(uniforms, p), axis=-1, keepdims=False)
 
 
+@partial(jit, static_argnums=(2,))
+def categorical_rvs(key, p, shape=()):
+    # this implementation is fast when event shape is small, and slow otherwise
+    # Ref: https://stackoverflow.com/a/34190035
```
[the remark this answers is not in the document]
fehiepsi: Yeah, a while_loop + reshape_to_2D + vmap should be enough. I'll add a TODO note for further enhancement.

numpyro/distributions/multivariate.py
```diff
@@ -7,27 +7,80 @@
...
 def _lnB(alpha):
     return np.sum(gammaln(alpha), axis=-1) - gammaln(np.sum(alpha, axis=-1))
 
 
-class dirichlet_gen(jax_mvcontinuous):
-    # TODO: use dirichlet doc instead of the default one of rv_continuous
+# TODO: either use multivariate docs instead of the default one of
+# rv_continuous/rv_discrete
+
+class categorical_gen(jax_multivariate, jax_discrete):
```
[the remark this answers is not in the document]
fehiepsi: `jax_discrete` has a special __init__ which includes `is_logits` flag. In addition, because we don't have a common method for `logpdf` `logpmf` so by inheriting from `jax_discrete`, we just need to check `isinstance(dist, jax_discrete)`. Beside these convenience, there is no need for inheritance.

numpyro/distributions/distribution.py
fehiepsi: This class mainly provides a `parse_args` method different from `rv_discrete, rv_continuous` which includes `loc` and `scale`. On the other hand, the `size` pattern is also different from the other two classes.
For continuous dist, if size is not provided, then `size` is the broadcasted shape of args + loc + scale.
For discrete dist, if size is not provided, then `size` is the broadcasted shape of args + loc. For multivariate distribution, this pattern is not true any more:
+ for mvn: `size` = broadcasted(mean.shape, cov.shape[:-1])
+ for dirichlet: `size = alpha.shape[:-1]`
+ for multinomial: `size = p.shape[:-1]`, n is int
+ ...
> can we not do this in jax_discrete and jax_continuous now
Yes, thanks for suggestion! I believe we can make a common `rvs` method for `jax_generic` based on `_batch_shape` method. I'll make a follow-up PR for it.

test/test_mcmc.py
fehiepsi: Hmm, it took 11s for HMC and 15s for NUTS on my CPU. I used jax pip version and might have compiled jaxlib from some commits in the master branch. I am not sure if it affects performance. Looking like Travis also runs well.

numpyro/distributions/distribution.py
neerajprad: Thanks for explaining. 

>  I'll make a follow-up PR for it if you also think so.

I was asking more for my clarification; I actually think we shouldn't spend more time on this than is needed to unblock any work that you are doing, because I am not convinced that hacks like `parse_args` need to be built up on. :)

test/test_mcmc.py
neerajprad: Hmm..for some reason HMC takes ~ 11 s. for me, but NUTS doesn't seem to terminate in any reasonable amount of time. But since its working on travis, I'll merge this.

PR #115, opened 2019-04-18
JIT tscan and HMC warmup steps

### LINKED ISSUE #114: Make tscan jittable
As it stands, unlike other primitives, our tscan implementation is not jittable. It is worth making it work well with `jax.jit` for benchmarking purposes.

neerajprad:
Fixes #114. 

This fixes the issue with `tscan` not being jittable. With this change and jitting the `warmup_update` step, we get significantly faster run times, specially for NUTS. e.g. for `test_beta_bernoulli`, the run time goes down from 21s to 15s. 

I have also added a minimal version of @fehiepsi's #92 as an example to test how this change fares on the benchmark, and also so that it is easy to make changes and see how the benchmark moves (which takes a few more steps in a jupyter notebook). It seems that this is competitive for HMC, but for NUTS, this is still quite slow. @fehiepsi - is it important to start with the step size and initial param values as in the notebook?

TODO:
 - [x] Figure out why NUTS is still slow wrt the original benchmark.

### DIFF
--- a/numpyro/examples/covtype.py
+++ b/numpyro/examples/covtype.py
@@ -16,7 +16,7 @@ from numpyro.util import tscan
 
 
 step_size = 0.00167132
-init_params = {"coefs": np.array(
+init_params = {"coefs": onp.array(
     [+2.03420663e+00, -3.53567265e-02, -1.49223924e-01, -3.07049364e-01,
      -1.00028366e-01, -1.46827862e-01, -1.64167881e-01, -4.20344204e-01,
      +9.47479829e-02, -1.12681836e-02, +2.64442056e-01, -1.22087866e-01,
@@ -63,10 +63,7 @@ def model(data, labels):
 
 
 def benchmark_hmc(args, features, labels):
-    N, dim = features.shape
     trajectory_length = step_size * args.num_steps
-    init_params = {'coefs': random.normal(key=random.PRNGKey(0), shape=(dim,))}
-
     _, potential_fn, _ = initialize_model(random.PRNGKey(1), model, (features, labels,), {})
     init_kernel, sample_kernel = hmc(potential_fn, algo=args.algo)
     t0 = time.time()
--- a/numpyro/examples/covtype.py
+++ b/numpyro/examples/covtype.py
@@ -57,7 +57,6 @@ def benchmark_hmc(args, features, labels):
                                   trajectory_length=trajectory_length, run_warmup=False)
     t1 = time.time()
     print("time for hmc_init: ", t1 - t0)
-    hmc_state = hmc_state.update(step_size=1.)
 
     def transform(state): return {'coefs': state.z['coefs'],
                                   'num_steps': state.num_steps}
--- a/numpyro/examples/covtype.py
+++ b/numpyro/examples/covtype.py
@@ -64,6 +64,7 @@ def benchmark_hmc(args, features, labels):
     def body_fn(state, i):
         return sample_kernel(state)
 
+    hmc_state = hmc_state.update(step_size=1.)
     tscan(body_fn, hmc_state, np.arange(1), transform=transform)
     t2 = time.time()
     print("time to compile sample_kernel: ", t2 - t1)
--- a/numpyro/examples/covtype.py
+++ b/numpyro/examples/covtype.py
@@ -15,7 +15,7 @@ from numpyro.mcmc import hmc
 from numpyro.util import tscan
 
 
-jax.config.update("jax_platform_name", "gpu")
+
 
 
 # TODO: add to datasets.py so as to avoid dependency on scikit-learn
@@ -76,6 +76,7 @@ def benchmark_hmc(args, features, labels):
 
 
 def main(args):
+    jax.config.update("jax_platform_name", args.device)
     features, labels = load_dataset()
     benchmark_hmc(args, features, labels)
 
@@ -85,5 +86,6 @@ if __name__ == '__main__':
     parser.add_argument('-n', '--num-samples', default=100, type=int, help='number of samples')
     parser.add_argument('--num-steps', default=10, type=int, help='number of steps (for "HMC")')
     parser.add_argument('--algo', default='NUTS', type=str, help='whether to run "HMC" or "NUTS"')
+    parser.add_argument('--device', default='cpu', type=str, help='use "cpu" or "cuda".')
     args = parser.parse_args()
     main(args)
--- a/numpyro/examples/covtype.py
+++ b/numpyro/examples/covtype.py
@@ -4,6 +4,7 @@ import time
 import numpy as onp
 from sklearn.datasets import fetch_covtype
 
+import jax
 import jax.numpy as np
 from jax import random
 
@@ -12,6 +13,9 @@ from numpyro.handlers import sample
 from numpyro.hmc_util import initialize_model
 from numpyro.mcmc import hmc
 from numpyro.util import tscan
+
+
+jax.config.update("jax_platform_name", "gpu")
 
 
 # TODO: add to datasets.py so as to avoid dependency on scikit-learn
@@ -69,7 +73,6 @@ def benchmark_hmc(args, features, labels):
     num_leapfrogs = np.sum(hmc_states['num_steps'])
     print('number of leapfrog steps: ', num_leapfrogs)
     print('avg. time for each step: ', (time.time() - t1) / num_leapfrogs)
-    print(hmc_states['coefs'])
 
 
 def main(args):
--- a/numpyro/examples/covtype.py
+++ b/numpyro/examples/covtype.py
@@ -47,7 +47,7 @@ def benchmark_hmc(args, features, labels):
     N, dim = features.shape
     step_size = np.sqrt(0.5 / N)
     trajectory_length = step_size * args.num_steps
-    init_params = {'coefs': np.zeros(dim)}
+    init_params = {'coefs': random.normal(key=random.PRNGKey(0), shape=(dim,))}
 
     _, potential_fn, _ = initialize_model(random.PRNGKey(1), model, (features, labels,), {})
     init_kernel, sample_kernel = hmc(potential_fn, algo=args.algo)
@@ -64,15 +64,12 @@ def benchmark_hmc(args, features, labels):
     def body_fn(state, i):
         return sample_kernel(state)
 
-    hmc_state = hmc_state.update(step_size=1.)
-    tscan(body_fn, hmc_state, np.arange(1), transform=transform)
-    t2 = time.time()
-    print("time to compile sample_kernel: ", t2 - t1)
     hmc_state = hmc_state.update(step_size=step_size)
     hmc_states = tscan(body_fn, hmc_state, np.arange(args.num_samples), transform=transform)
     num_leapfrogs = np.sum(hmc_states['num_steps'])
     print('number of leapfrog steps: ', num_leapfrogs)
-    print('avg. time for each step: ', (time.time() - t2) / num_leapfrogs)
+    print('avg. time for each step: ', (time.time() - t1) / num_leapfrogs)
+    print(hmc_states['coefs'])
 
 
 def main(args):
--- a/numpyro/examples/covtype.py
+++ b/numpyro/examples/covtype.py
@@ -1,0 +1,89 @@
+import argparse
+import time
+
+import numpy as onp
+from sklearn.datasets import fetch_covtype
+
+import jax.numpy as np
+from jax import random
+
+import numpyro.distributions as dist
+from numpyro.handlers import sample
+from numpyro.hmc_util import initialize_model
+from numpyro.mcmc import hmc
+from numpyro.util import tscan
+
+
+# TODO: add to datasets.py so as to avoid dependency on scikit-learn
+def load_dataset():
+    data = fetch_covtype()
+    features = data.data
+    labels = data.target
+
+    # normalize features and add intercept
+    features = (features - features.mean(0)) / features.std(0)
+    features = np.hstack([features, np.ones((features.shape[0], 1))])
+
+    # make binary feature
+    _, counts = onp.unique(labels, return_counts=True)
+    specific_category = np.argmax(counts)
+    labels = (labels == specific_category)
+
+    N, dim = features.shape
+    print("Data shape:", features.shape)
+    print("Label distribution: {} has label 1, {} has label 0"
+          .format(labels.sum(), N - labels.sum()))
+    return features, labels
+
+
+def model(data, labels):
+    N, dim = data.shape
+    coefs = sample('coefs', dist.norm(np.zeros(dim), np.ones(dim)))
+    logits = np.dot(data, coefs)
+    return sample('obs', dist.bernoulli(logits, is_logits=True), obs=labels)
+
+
+def benchmark_hmc(args, features, labels):
+    N, dim = features.shape
+    step_size = np.sqrt(0.5 / N)
+    trajectory_length = step_size * args.num_steps
+    init_params = {'coefs': np.zeros(dim)}
+
+    _, potential_fn, _ = initialize_model(random.PRNGKey(1), model, (features, labels,), {})
+    init_kernel, sample_kernel = hmc(potential_fn, algo=args.algo)
+    t0 = time.time()
+    # Do not run warmup for consistent benchmark
+    hmc_state, _, _ = init_kernel(init_params, num_warmup_steps=0, step_size=step_size,
+                                  trajectory_length=trajectory_length, run_warmup=False)
+    t1 = time.time()
+    print("time for hmc_init: ", t1 - t0)
+    hmc_state = hmc_state.update(step_size=1.)
+
+    def transform(state): return {'coefs': state.z['coefs'],
+                                  'num_steps': state.num_steps}
+
+    def body_fn(state, i):
+        return sample_kernel(state)
+
+    tscan(body_fn, hmc_state, np.arange(1), transform=transform)
+    t2 = time.time()
+    print("time to compile sample_kernel: ", t2 - t1)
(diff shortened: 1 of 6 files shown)


numpyro/examples/covtype.py
```diff
@@ -0,0 +1,89 @@
...
+    coefs = sample('coefs', dist.norm(np.zeros(dim), np.ones(dim)))
+    logits = np.dot(data, coefs)
+    return sample('obs', dist.bernoulli(logits, is_logits=True), obs=labels)
+
+
+def benchmark_hmc(args, features, labels):
+    N, dim = features.shape
+    step_size = np.sqrt(0.5 / N)
+    trajectory_length = step_size * args.num_steps
+    init_params = {'coefs': np.zeros(dim)}
```
neerajprad: @fehiepsi - I notice that you used different initializations for HMC and NUTS, which I missed in the review. Could you elaborate on the reasons for doing so?

numpyro/examples/covtype.py
```diff
@@ -0,0 +1,89 @@
...
+    hmc_state = hmc_state.update(step_size=1.)
+
+    def transform(state): return {'coefs': state.z['coefs'],
+                                  'num_steps': state.num_steps}
+
+    def body_fn(state, i):
+        return sample_kernel(state)
+
+    tscan(body_fn, hmc_state, np.arange(1), transform=transform)
+    t2 = time.time()
```
neerajprad: We can probably remove this step, as it doesn't seem to affect the timings that much.

numpyro/patch.py
```diff
@@ -0,0 +1,28 @@
...
+        setattr(module, name, new_fn)
+        return new_fn
+
+    return decorator
+
+
+# TODO: Remove with jax v0.1.26
+@patch_dependency('jax.interpreters.partial_eval.trace_unwrapped_to_jaxpr', jax)
+def _trace_unwrapped_to_jaxpr(fun, pvals, **kwargs):
+    return pe.trace_to_jaxpr(lu.wrap_init(fun, kwargs), pvals)
```
fehiepsi: +1

test/conftest.py
```diff
@@ -1,5 +1,8 @@
+import os
+
 from numpyro.util import set_rng_seed
 
 
 def pytest_runtest_setup(item):
     set_rng_seed(0)
+    os.environ['XLA_FLAGS'] = '--xla_cpu_enable_fast_math=false'
```
fehiepsi: +1 nice!

numpyro/mcmc.py
```diff
@@ -80,7 +80,8 @@ def init_kernel(init_samples,
                              wa_state.step_size, wa_state.inverse_mass_matrix, rng_hmc)
 
         if run_warmup:
-            hmc_state, _ = fori_loop(0, num_warmup_steps, warmup_update, (hmc_state, wa_state))
+            hmc_state, _ = jit(fori_loop, static_argnums=(2,))(0, num_warmup_steps, warmup_update,
+                                                               (hmc_state, wa_state))
```
fehiepsi: hmm, does this really help? I thought we just run it 1 time so jitting is not useful.

numpyro/examples/covtype.py
fehiepsi: ^^ The only reason is to match the corresponding benchmarks in **two** references (each one has its own way to benchmark @@). I believe that using initialization for HMC as you do here is enough.

numpyro/examples/covtype.py
```diff
@@ -0,0 +1,89 @@
...
+    features, labels = load_dataset()
+    benchmark_hmc(args, features, labels)
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(description="parse args")
+    parser.add_argument('-n', '--num-samples', default=100, type=int, help='number of samples')
+    parser.add_argument('--num-steps', default=10, type=int, help='number of steps (for "HMC")')
+    parser.add_argument('--algo', default='NUTS', type=str, help='whether to run "HMC" or "NUTS"')
+    args = parser.parse_args()
```
fehiepsi: It would be better to have an option to run this benchmark script in GPU: by using `jax.config.update("jax_platform_name", "gpu")`

test/conftest.py
neerajprad: This doesn't work actually, so I have removed it. The problem is that it needs to be set much earlier, so we will need to set the environment variable manually. I had to do this because the NUTS tests wouldn't terminate otherwise.

numpyro/mcmc.py
neerajprad: Both of these help. Yes, please test. My hypothesis is that it helps because it compiles `sample_kernel` essentially which is reused across time steps.

neerajprad:
> We can keep the notebook for a while to compare different strategy, then we should remove it. If we make script, it would be nice to store the benchmark result somewhere (e.g. in wiki) so we can keep track of performance.

We can retain both. I think its nice to have a notebook that gives more details, but a script is more handy for comparison on a day to day basis.

> About the slowness of NUTS, if you let num_samples=100, then it will take sometime to finish because NUTS have larger trajectory length than HMC. In my run (with the version in the notebook, it took 65322 leapfrog steps to get 100 samples), so basically will take 65x longer than HMC. It took me 5m in GPU to finish getting 100 NUTS samples (with HMC init param/step_size) so in CPU, I expect it will take more than half an hour. ^^!

I see. From one of the runs, the time per leapfrog step was also quite a bit higher, let me let it run for half an hour then and see what I get.

neerajprad:
>  it took 65322 leapfrog steps to get 100 samples

This seems a bit surprising. With the benchmark scripts NUTS terminates within 1138 leapfrog steps but it takes around 0.1 s for each step, which is much higher than 0.06 for HMC. Could you disable fast math mode, and try the benchmark on your system?

fehiepsi:
@neerajprad That's the run in GPU, where precision is much better than CPU. For CPU with fast math disable, in my system:
```
100%|██████████| 100/100 [01:08<00:00,  1.92it/s]

number of leapfrog steps: 1138
avg. time for each step : 0.06028635409678642
```
while in HMC,
```
100%|██████████| 100/100 [00:59<00:00,  1.68it/s]

number of leapfrog steps: 1000
avg. time for each step : 0.05941483545303345
```
If you disable fast math and plot the sample of first coef, you will see that HMC samples are zeros constant and NUTS samples are highly correlated.

neerajprad:
It is interesting that you get the same number of leapfrog steps for NUTS but your timings look better (0.06 vs 0.1 sec. per step). Just to confirm, this is with on `covtype.py` using the same initialization for both?

fehiepsi:
Yes, both with `step_size = np.sqrt(0.5 / N)` and `init_params = {"coefs": np.zeros(dim)}`.

numpyro/mcmc.py
fehiepsi: Yup, I can confirm that it helps. :D

fehiepsi:
@neerajprad All the above info from my side is from modifications (step size, init params) of the notebook (which I didn’t use tscan). I have not used your script.

neerajprad:
> @neerajprad All the above info from my side is from modifications (step size, init params) of the notebook (which I didn’t use tscan). I have not used your script.

Then the script seems to be slower than your notebook for NUTS. I'll need to take a look at why that's the case. Hopefully if this (or a similar approach works), we can have a fast implementation by default instead of relying on users to bypass warmup and compile `sample_kernel`.

fehiepsi:
I hope so too. Let me play around with your idea too to see if we miss something.

neerajprad:
Feel free to push to this PR itself; I might not be able to get to this until tomorrow.

numpyro/examples/covtype.py
fehiepsi: I think that this is one of the reasons for the slowdown. `np.arange(1)` has a different shape w.r.t. `np.arange(num_samples)` so tscan will be recompiled.

neerajprad: Actually this step is not needed, and doesn't change the benchmark. I have removed it now.

numpyro/examples/covtype.py
neerajprad: I have added this. This should be only needed to change back to cpu I think, since gpu would be used by default when available.

neerajprad:
@fehiepsi - With fast-math disabled, I'm getting 0.1 sec per time step for HMC (0.06 with fast-math), and 0.09 for NUTS. I'm happy with the benchmark in that both HMC and NUTS have similar times. I haven't compared this with your original benchmark on my system, which I think will be similar to this. All the benchmark numbers are uniformly better on your system though (probably because you have a more powerful CPU). In any case, this is ready to merge, unless you have further comments.

numpyro/examples/covtype.py
neerajprad: Oh..I see you meant having a command line option. Will add.

numpyro/examples/covtype.py
```diff
@@ -0,0 +1,91 @@
...
+    features, labels = load_dataset()
+    benchmark_hmc(args, features, labels)
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(description="parse args")
+    parser.add_argument('-n', '--num-samples', default=100, type=int, help='number of samples')
+    parser.add_argument('--num-steps', default=10, type=int, help='number of steps (for "HMC")')
+    parser.add_argument('--algo', default='NUTS', type=str, help='whether to run "HMC" or "NUTS"')
+    parser.add_argument('--device', default='cpu', type=str, help='use "cpu" or "cuda".')
```
fehiepsi: For the [lastest release](https://github.com/google/jax/blob/03c64af2d531e83c00a6973e90dc3d8a8fdd853a/jax/lib/xla_bridge.py#L122), `device` is either "cpu"/"gpu" or "CUDA", "Host". Because cpu/gpu are the defaults, I think we should replace "cuda" by "gpu" here.

numpyro/examples/covtype.py
```diff
@@ -0,0 +1,91 @@
...
+    t1 = time.time()
+    print("time for hmc_init: ", t1 - t0)
+
+    def transform(state): return {'coefs': state.z['coefs'],
+                                  'num_steps': state.num_steps}
+
+    def body_fn(state, i):
+        return sample_kernel(state)
+
+    hmc_state = hmc_state.update(step_size=step_size)
```
fehiepsi: Could you add a comment into the script for why this step is needed so we can address it in later PR? I guess the reason is we convert step_size to float in `hmc` init implementation?

numpyro/examples/covtype.py
```diff
@@ -0,0 +1,91 @@
...
+import jax.numpy as np
+from jax import random
+
+import numpyro.distributions as dist
+from numpyro.handlers import sample
+from numpyro.hmc_util import initialize_model
+from numpyro.mcmc import hmc
+from numpyro.util import tscan
+
+
```
fehiepsi: nit: lint for 5 blank lines

numpyro/examples/covtype.py
```diff
@@ -0,0 +1,91 @@
...
+    coefs = sample('coefs', dist.norm(np.zeros(dim), np.ones(dim)))
+    logits = np.dot(data, coefs)
+    return sample('obs', dist.bernoulli(logits, is_logits=True), obs=labels)
+
+
+def benchmark_hmc(args, features, labels):
+    N, dim = features.shape
+    step_size = np.sqrt(0.5 / N)
+    trajectory_length = step_size * args.num_steps
+    init_params = {'coefs': random.normal(key=random.PRNGKey(0), shape=(dim,))}
```
fehiepsi: nit: I think it is better to use `init_params` from `initialize_model`

fehiepsi:
> I'm happy with the benchmark in that both HMC and NUTS have similar times. I haven't compared this with your original benchmark on my system, which I think will be similar to this. All the benchmark numbers are uniformly better on your system though (probably because you have a more powerful CPU).

I just want to verify something to make sure we don't have different interpretations of the problem:
+ With the current init_params, first `num_steps` of NUTS is 1. So the hack `step_size=1.` is not important. In addition, we run NUTS with `num_samples=100`, so the effect of first sampling step would be small. When first `num_steps` is large, compiling time will contribute to the benchmark a lot. I think that is the reason you didn't use the NUTS init_params and step_size as in the notebook's benchmark (NUTS will take a bunch of time to finish its first step). If that is the case, then I would prefer to keep init_pamams and step_size as in the notebook for further enhancement (in later PRs) and change it back to the default init_params when problems are resolved.
+ The following timing in my system (cpu, fastmath disabled) shows that first step is important.
```
HMC script (n=10): 122ms
HMC notebook (n=10): 61ms
HMC script (n=100): 64ms
HMC notebook (n=100): 61ms
```

numpyro/examples/covtype.py
neerajprad: Actually the only reason why I had this was to run for 10 steps per iteration in HMC, but since we don't have the `num_steps` arg anymore and trajectory_length is already set, this was the simplest way to ensure that.

neerajprad:
>  When first num_steps is large, compiling time will contribute to the benchmark a lot. I think that is the reason you didn't use the NUTS init_params and step_size as in the notebook's benchmark (NUTS will take a bunch of time to finish its first step). If that is the case, then I would prefer to keep init_pamams and step_size as in the notebook for further enhancement (in later PRs) and change it back to the default init_params when problems are resolved.

Good point. I didn't use `init_params` from the notebook because its hard to reliably compare our benchmark against the paper due to a variety of system level differences. I think we can use any reasonable initialization and run the benchmarks for other implementations, so I tried to keep it simple. But if that is a particularly problematic initialization, let us just use that for the time being and later move to a simpler one when that problem is fixed.

numpyro/examples/covtype.py
fehiepsi: I see. The other way is to set `adapt_step_size=False` so step_size will be constant during init_kernel. (more precisely, find_initial_step_size won't trigger.)

neerajprad:
@fehiepsi - let me address all your comments. I think it would be best to just use the notebook's benchmark for both HMC and NUTS.

numpyro/examples/covtype.py
neerajprad: Done.

neerajprad:
Okay, I think this should address all your comments. I just had a couple of questions based on your last comment:

> The following timing in my system (cpu, fastmath disabled) shows that first step is important.

Are these numbers with the earlier initialization of 0s or the one that is in the notebook? I changed to the one in the notebook and the performance per leapfrog step remains the same. It is possible that the initial compilation time is large and is getting amortized over the larger number of leapfrog steps. But if that's the case, I think we probably should not overly focus our efforts on fixed costs (like compiling the `sample_kernel` twice) that are likely to get amortized away on larger models / larger number of samples. We can probably rely on JAX fixing that over the longer term. What would be nice is if we are very competitive on large models (I'm not sure if we are there yet), even if we are 10 seconds slower on smaller models (this is important too, but it is kind of optimizing for the tail so early in the project). What do you think?

numpyro/examples/covtype.py
```diff
@@ -0,0 +1,106 @@
...
+
+import numpyro.distributions as dist
+from numpyro.handlers import sample
+from numpyro.hmc_util import initialize_model
+from numpyro.mcmc import hmc
+from numpyro.util import tscan
+
+
+step_size = 0.00167132
+init_params = {"coefs": np.array(
```
fehiepsi: nit: It is better to use `onp.array` here so `jax.config.update("jax_platform_name", args.device)` can take effect.

numpyro/examples/covtype.py
```diff
@@ -0,0 +1,106 @@
...
+    N, dim = data.shape
+    coefs = sample('coefs', dist.norm(np.zeros(dim), np.ones(dim)))
+    logits = np.dot(data, coefs)
+    return sample('obs', dist.bernoulli(logits, is_logits=True), obs=labels)
+
+
+def benchmark_hmc(args, features, labels):
+    N, dim = features.shape
+    trajectory_length = step_size * args.num_steps
+    init_params = {'coefs': random.normal(key=random.PRNGKey(0), shape=(dim,))}
```
fehiepsi: nit: I guess you want to remove this line?

numpyro/examples/covtype.py
neerajprad: Interesting. Any idea what happens if the platform is changed later? Do we get an error or does it move the array to the gpu?

fehiepsi:
> Are these numbers with the earlier initialization of 0s or the one that is in the notebook?

Those are numbers from `init_params = {'coefs': random.normal(key=random.PRNGKey(0), shape=(dim,))}`.

> But if that's the case, I think we probably should not overly focus our efforts on fixed costs (like compiling the sample_kernel twice) that are likely to get amortized away on larger models / larger number of samples. We can probably rely on JAX fixing that over the longer term.

Yes, it would be great if this issue is fixed, so we just need to `jit` sample_kernel in `hmc` implementation and use it across `init` and `sample` stages. But right now, compiling time is a big problem:
+ when first step of NUTS takes hundreds of steps
+ when there are a bunch of sample statements in models

To make a fair benchmark (while waiting for the issue is fixed upstream), I think we can use non-prim version for a while with a trade-off that we have to concatenate samples without compiling. The other option is to go aggressive and make mcmc work as follows
```
mcmc = MCMC(sample_kernel, state, num_samples=1000)  # make a jit version of tscan here with num_samples is known, in other works, `bs` length is known
mcmc.compile()  # run tscan with 1 step; we can also trigger the hack for fast compiling here
mcmc.run()  # run tscan with 1000 steps
```
But I don't like this idea because it is just suitable for benchmarking. We have to recompile to draw another 1000 samples... So for the time being, I would stick with non-prim version and wait for the issue is fixed upstream. ^^

> What would be nice is if we are very competitive on large models (I'm not sure if we are there yet), even if we are 10 seconds slower on smaller models

Yes, I also think so.

Issue #113, opened 2019-04-17
Use upstream samplers


fehiepsi:
There are various sampler available upstream in the last version of JAX. Let's use these samplers instead.

neerajprad:
Can we close this issue now?

fehiepsi:
Yup :)

PR #116, opened 2019-04-19
Use upstream sampler

fehiepsi:
Address #113. I also update the notebook to address recent changes.


numpyro/mcmc.py
```diff
@@ -83,7 +83,7 @@ def init_kernel(init_samples,
             hmc_state, _ = fori_loop(0, num_warmup_steps, warmup_update, (hmc_state, wa_state))
             return hmc_state
         else:
-            return hmc_state, wa_state, warmup_update
+            return hmc_state, wa_state, wa_update
```
fehiepsi: I want to use wa_update instead because warmup_update uses non-jit version of sample_kernel.

Issue #77, opened 2019-04-02
Support float64 array


fehiepsi:
Currently, we only develop and test for float32 array. When the functionalities are ready, we should test for float64 array too.

neerajprad:
> When the functionalities are ready, we should test for float64 array too.

Note that float64 will be significantly slower than float32 for GPU, so I think it is a good practice to write our code assuming that we will run it on float32, i.e. ensure numerical stability whenever possible. e.g. by providing logit support in distributions. I think tests passing with float32 should almost always pass with float64 (the reverse isn't true)?

fehiepsi:
@neerajprad This is to prevent something like `a = np.array([1.])`, `np.arange(...)`, or `random.normal(...)` without `dtype` in the code.

PR #117, opened 2019-04-21
Benchmark numpyro on semisupervised hmm model

fehiepsi:
This PR is an attempt to do benchmark on a hmm model (which Stan is good at). Working on this kind of model will help us identify the edge limitations of our implementation.

Here are some of my observations:
+ ~Compiling `find_reasonable_step_size` is so slow on this model. This function requires a while_loop where body_fn needs to evaluate potential energy and its gradient. But there seems another source (which I haven't identified yet) for the slowness of this function.~
+ ~NUTS does not learn on CPU (for both fastmath enabled and disabled) due to [precision issue of `sum`](https://github.com/google/jax/issues/585)~
+ ~Comparing to Stan on CPU, NumPyro on GPU is extremely slow (30s vs 12 minutes). Maybe running on CPU will show a more comparable performance (when the above precision issue is fixed).~
+ ~Stan does inference better. This might come from the fact that Stan works in double precision.~ This is fixed, numpyro did a bit better job now.

hmm


fehiepsi:
Looking like using `lax.scan` (when its jvp rule is available) for forward_log_prob function will make things much faster.

neerajprad:
> NUTS does not learn on CPU (for both fastmath enabled and disabled) due to precision issue of sum

Does the precision issue persist with double arrays as well?

fehiepsi:
The precision issue is due to the algorithm of np.sum, so it happens for float64 array too. Currently, testing our hmc implementation on float64 is blocked by https://github.com/pyro-ppl/numpyro/issues/77.

neerajprad:
> Currently, testing our hmc implementation on float64 is blocked by #77.

I see. So the issue is that there is no way to set the default dtype to `np.float64` in jax?

neerajprad:
> Compiling find_reasonable_step_size is so slow on this model. 

So does this mean that the function runs fast but is slow under JIT? 

> This function requires a while_loop where body_fn needs to evaluate potential energy and its gradient. 

That should be true for all models, so I am not sure why that would be a contributing factor.

fehiepsi:
> So the issue is that there is no way to set the default dtype to np.float64 in jax?

To enable np.float64, we can export the environment `JAX_ENABLE_X64=1`. Currently, something like  `np.array([1.])`, `np.arange(...)`, or `random.normal(...)` without dtype kwarg would return np.float32 instead of np.float64.

> the function runs fast but is slow under JIT?

Yes, it is fast w.r.t. potential_fn evaluation time.

numpyro/mcmc.py
```diff
@@ -62,15 +63,19 @@ def init_kernel(init_samples,
         z_flat, unravel_fn = ravel_pytree(z)
         momentum_generator = partial(_sample_momentum, unravel_fn)
 
-        find_reasonable_ss = partial(find_reasonable_step_size,
-                                     potential_fn, kinetic_fn, momentum_generator)
+        wa_kwargs = {}
+        # FIXME: compiling find_reasonable_step_size is so slow
```
neerajprad: Could you elaborate on this issue? I thought `find_reasonable_step_size` will be called during `wa_init`, but since that is outside the JIT loop, why is it being compiled?

notebooks/semi_supervised_hmm.ipynb
```diff
@@ -0,0 +1,503 @@
...
+    "\n",
+    "    sample(\"supervised_categories\", dist.categorical(transition_prob[supervised_categories[:-1]]),\n",
+    "           obs=supervised_categories[1:])\n",
+    "    sample(\"supervised_words\", dist.categorical(emission_prob[supervised_categories]),\n",
+    "           obs=supervised_words)\n",
+    "\n",
+    "    transition_log_prob = np.log(transition_prob)\n",
+    "    emission_log_prob = np.log(emission_prob)\n",
+    "    log_prob = emission_log_prob[:, unsupervised_words[0]]\n",
+    "    # TODO: to be replaced by lax.scan when its jvp rule is implemented\n",
```
neerajprad: Unfortunately, I think this will be a blocker to getting reasonable performance on these kind of models.

numpyro/mcmc.py
fehiepsi: I mean the usage of `lax.while_loop` (which does compiling under the hood) in `find_reasonable_step_size` is so slow.

fehiepsi:
@neerajprad Now our nuts did a good job on this model (except for the slowness). The problem is injecting log_prob through `dist.Bernoulli(log_prob.exp())` is a bad way. It is better to use `dist.Multinomial(1, log_prob)`. :)

neerajprad:
> The problem is injecting log_prob through dist.Bernoulli(log_prob.exp()) is a bad way. It is better to use dist.Multinomial(1, log_prob).

Good to know. Based on our past experience, I think one place to check if some example is running too slow is whether the distributions involved have any numerical instability in their log density computation. I was similarly finding a very slow test in `test_mcmc.py` which turned out to be subtle bug in one of the distributions that I had ported.

fehiepsi:
@neerajprad lax.scan is working well and inference is really fast. So I convert the notebook to a script instead. The only blocker now is `np.sum` precision issue in CPU, which makes `accept_prob = 0` (despite that we spend 1000 steps for each tree). I think that this is good to review. Not hurry though, we are still waiting for the np.sum issue is resolved. :)

fehiepsi:
@neerajprad It turns out that the bug is from my math, not from `np.sum`. Thanks for your correction, the example runs well in CPU now.

neerajprad:
> the example runs well in CPU now. This gist shows that compiling time is 2.5x faster than Stan, and sampling time is 3x faster (despite that we spent 2x more leapfrog steps than Stan)

That's awesome news! I am looking at this example now.

test/test_examples.py
```diff
@@ -11,6 +11,7 @@
 EXAMPLES = [
     'baseball.py --num-samples 100 --num-warmup 100',
     'covtype.py --algo hmc --num-samples 10',
+    'hmm.py --num-samples 100 --num-warmup 100',
```
neerajprad: I think we'll need to wait for the next version of jax to enable this.

fehiepsi: Thanks, I add a mark to skip it

fehiepsi:
@neerajprad I guess funsor is good for hmm because the model is one of the core examples of funsor. But it is not clear to me how delay sampling will work with `scan`. It is an interesting problem though. :)

neerajprad:
> @neerajprad I guess funsor is good for hmm because the model is one of the core examples of funsor. But it is not clear to me how delay sampling will work with scan. It is an interesting problem though. :)

Funsor should be able to automatically construct the `log_prob` term that we are manually writing out here, and will generalize to all kinds of HMM models. We shouldn't even need `scan`, as the `log_prob` term will just be an einsum expression and we should be able to take grad through that. cc. @fritzo. Would be great to try it out, but we'll need to first build up the numpy/JAX backend in funsor. I started that, but there is more work to be done.

examples/hmm.py
```diff
@@ -0,0 +1,179 @@
...
+                        supervised_categories, supervised_words,
+                        unsupervised_words):
+    num_categories, num_words = transition_prior.shape[0], emission_prior.shape[0]
+    transition_prob = sample('transition_prob', dist.Dirichlet(
+        np.broadcast_to(transition_prior, (num_categories, num_categories))))
+    emission_prob = sample('emission_prob', dist.Dirichlet(
+        np.broadcast_to(emission_prior, (num_categories, num_words))))
+
+    # models supervised data
+    sample('supervised_categories', dist.Categorical(transition_prob[supervised_categories[:-1]]),
```
neerajprad: We are missing the first state. Should we also sample `init_prob` from some prior, and have `sample('init_state', dist.Categorical(init_prob), obs=supervised_categories[0])`?

examples/hmm.py
```diff
@@ -0,0 +1,179 @@
...
+                       supervised_categories, supervised_words, unsupervised_words,
+                       random.PRNGKey(2), args)
+    print_results(zs, transition_prob, emission_prob)
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(description='Semi-supervised Hidden Markov Model')
+    parser.add_argument('--num-categories', default=3, type=int)
+    parser.add_argument('--num-words', default=10, type=int)
+    parser.add_argument('--num-supervised', default=100, type=int)
```
neerajprad: Just for my understanding, how low can we go here before inference fails completely?

examples/hmm.py
```diff
@@ -0,0 +1,179 @@
...
+Model for a generative model with observations are words and latent variables
+are categories. Instead of automatically marginalizing all discrete latent
+variables (as in [2]), we will use the "forward algorithm" (which exploits the
+conditional independent of a Markov model - see [3]) to iteratively do this
+marginalization.
+
+The semi-supervised problem is chosen instead of an unsupervised one because it
+is hard to make the inference works for an unsupervised model (see the
+discussion [4]). On the other hand, this example also illustrates the usage of
+JAX's `lax.scan` primitive. The primitive will greatly improve compiling and
```
neerajprad: I'm curious where the benefit of using `scan` comes in - is it primarily in terms of a lower compile time, or is it also faster after compilation? My guess would be the former.

examples/hmm.py
neerajprad: I suppose the current formulation implicitly assumes that the initial state is uniformly distributed? If so, maybe we can just add a comment to that effect?

examples/hmm.py
fehiepsi: Yes, the first is the main benefit of lax.scan. I am not sure if it improves execution time so I'll remove it.

examples/hmm.py
fehiepsi: I think that if we don't use `init_prob` to anywhere else in the model then it is not necessary to add that uncertainty into the model. Given an observation categorical[0], we add 3 additional parameters. It is a bit nonsense to me.

fehiepsi: I don't think that we assume that in this model. Given a data X, y, you can model y ~ SomeDist(X) but need not to model X ~ SomePrior(...).

neerajprad: > then it is not necessary to add that uncertainty into the model.

The idea would be to encode our beliefs about the initial state into a prior (and it will likely affect inference depend on how strong this prior is), and get a posterior distribution over `init_prob`. If this prior is flat, then I guess what you are saying is correct, and it makes no difference in inference whether we include it or not (except that we won't end up recording the posterior in that case).

examples/hmm.py
fehiepsi: hmm20
hmm-10
FYI, the first one is `num_supervised=10`, the second one is `num_supervised=2`.

fehiepsi: So I guess it fails completely when n-sup between 2-10.

examples/hmm.py
neerajprad: > Given a data X, y, you can model y ~ SomeDist(X) but need not to model X ~ SomePrior(...).

Sure, I agree with that. If you don't model it, you are implicitly assuming it to be flat.

fehiepsi: Yes, you are right. I am assuming it to be flat. I don't know about that fact, so having a comment with `flat prior`, `uniform prior` makes me confused...

fehiepsi: will a comment "here we don't make any assumption about the first supervised category" make sense to you? It is my language for "here we make an assumption that the first supervised category comes from flat/uniform prior"

neerajprad: Sorry, my bad with the sloppy comment. :) Since I suppose we aren't interested in the distribution of the initial state, we can probably just add a comment "Assuming a flat/uniform prior over the simplex for the initial state".

neerajprad: Sorry, missed your earlier comment. Sure, feel free to write it in your own words. That makes sense to me!

examples/hmm.py
neerajprad: Interesting, thanks for the plots! The distributions don't look that much more different and the trends are kind of preserved which is good, but I wonder if we would end up with different results for other chains.

neerajprad:
Thanks for benchmarking this example, which has been helpful in so many ways. Given how slow HMM has been with Pyro NUTS, this is very promising, and hopefully will allow us to tackle more general models using Funsors in a not too distant future.

fehiepsi:
Thanks for reviewing this PR @neerajprad !

PR #118, opened 2019-04-23
Add fori_append construct in util

neerajprad:
This introduces a simpler version of `tscan` called `fori_append`, which as the name suggests, runs a for loop and gathers the results along the leading dimension of the pytree arg. This is simpler than the existing `tscan` version and slightly faster. 

Few things that I discovered:
 - Jitting the internal `fori_loop` is important. I have a version of this that uses a standard python `for` loop (and lax dynamic slice update) and only uses a jitted `f` arg, but it is slower (sometimes 2X so). Unfortunately, that also means that we cannot use the JITed version with tqdm / other updates. But we can probably use `jit=False` and print diagnostic information if needed.
 - After trying out many other versions, it seems unlikely that we can get further performance improvements from just the JIT alone. 
 - The `sample_kernel` if already compiled should be quite fast, i.e. there is no issue with double compilation.

### DIFF
--- a/numpyro/examples/covtype.py
+++ b/numpyro/examples/covtype.py
@@ -12,8 +12,7 @@ import numpyro.distributions as dist
 from numpyro.handlers import sample
 from numpyro.hmc_util import initialize_model
 from numpyro.mcmc import hmc
-from numpyro.util import tscan
-
+from numpyro.util import fori_append
 
 step_size = 0.00167132
 init_params = {"coefs": onp.array(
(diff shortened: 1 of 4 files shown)


neerajprad:
> Though I still think that the overhead of native for loop is small, it is better to merge this and compare in a later PR.

Sure! I had something like the version below with the native for loop. Feel free to play around with it (e.g. jitting `f`), and I think out tests should give you a reasonable idea about how the perf looks like with this implementation.

I think the current version should be able to support diagnostics with `jit=False` until we can implement something more elegant with https://github.com/google/jax/issues/364. So I didn't dig into this any deeper.

```python
def fori_append(f, a, n, transform):
    transform = jax.jit(transform)
    a_flat, trans_treedef = tree_flatten(transform(a))
    state = [lax.full((n,) + np.shape(i), 0, lax._dtype(i)) for i in a_flat]

    for i in range(n):
        a_out = f(a)
        state_out = transform(a_out)
        state_out = tree_flatten(state_out)[0]
        state = [lax.dynamic_update_index_in_dim(s, t[None, ...], i, axis=0)
                 for t, s in zip(state_out, state)]
        a = a_out
    return tree_unflatten(trans_treedef, state)
```

fehiepsi:
Thanks @neerajprad !

PR #119, opened 2019-04-24
Add local global trend notebook

fehiepsi:
This PR tries to replicate [local global trend](https://github.com/cbergmeir/Rlgt) time series model, so we can get an idea about how our hmc implementation applied to real problems. Some observations so far:
+ Similar to semi-supervised model, compiling time / sampling time (with unrolling for-loop involved in the model) is so slow comparing to Stan. We should replace the unrolling for-loop by lax.scan (when its jvp rule is available) to achieve a better performance.
+ Similar to semi-supervised notebook, lacking of precision of sum in CPU makes the inference in CPU so slow.
+ We still lack a predictive mechanism to forecast future values based on current samples.

Some enhancements which will be deferred to smaller PRs:
+ expose max_tree_depth
+ add trunccauchy, halfcauchy distributions


neerajprad:
@fehiepsi - Is this PR blocked due to numerical instability of `np.sum`?

fehiepsi:
@neerajprad Nope, I haven't taken a look at this for a while. It was blocked by the slowness without `lax.scan`. Incidentally, I am working on this again this morning. The forward is fast but the grad triggers a fixed point error. Maybe I need your help on isolating the error. :D

fehiepsi:
@neerajprad I just push the latest file. Could you help me double check that the models in Pylgt and Rlgt are equivalent?

neerajprad:
> @neerajprad I just push the latest files. Could you help me double check that the models in Pylgt and Rlgt are equivalent?

Thanks, @fehiepsi. I'll take a look at this shortly. This is really interesting!

fehiepsi:
@neerajprad It turns out that the bug is fixed upstream yesterday. You have to use the lastest jax version to pass the bug. :D

neerajprad:
I looked at the both the models and I couldn't see any disparities, but I'll take a closer look and play around with it.

I just had some questions / comments:
 - It seems that we have not yet implemented the `USE_REGRESSION` mode - are there any blockers that prevent us from doing that?
 - It would be nice to be able to record parameter values in the `transformed_data` block in the trace. One way might be to use `Delta` distribution on these parameters but we don't really want to compute gradients w.r.t. to them, or do any kind of inference. I think this will be an important use case to support, but I'm not sure what might be an elegant way to handle this.
 - It also seems that the chains haven't really converged for the Rlgt model (looking at the r_hat values), whereas they are all very close to 1. for your model. Any idea why there is such a huge difference?

fehiepsi:
> It seems that we have not yet implemented the USE_REGRESSION mode - are there any blockers that prevent us from doing that?

There is no blocker at all. It is just that the example code does not require that feature.

> It would be nice to be able to record parameter values in the transformed_data block in the trace. One way might be to use Delta distribution on these parameters but we don't really want to compute gradients w.r.t. to them, or do any kind of inference. I think this will be an important use case to support, but I'm not sure what might be an elegant way to handle this.

I use the following trick. It is pretty fast.
```
def model(...):
    return required_parameters

transformed_parameters = vmap(lambda asample: substitute(model, asample)(...))(samples)
```

> It also seems that the chains haven't really converged for the Rlgt model (looking at the r_hat values), whereas they are all very close to 1. for your model. Any idea why there is such a huge difference?

Yeah, I am curious about it too. I just follow the example at the [Rlgt](https://cran.r-project.org/web/packages/Rlgt/Rlgt.pdf) docs. The model does not converge, and the prediction is poor (cannot capture global trend). I am not sure why.

fehiepsi:
@neerajprad Just update the progress so far:
+ Forecasting is implemented, and it is pretty fast (thanks to `vmap`)
+ Range of prediction is good IMO but cannot capture global trend. We might ask Slawek if it is expected or not (together with Rlgt problem). Maybe he can suggest a better data for this model.

However, there still remains two problems:
+ tqdm is unusable due to [this issue](https://github.com/tqdm/tqdm/issues/249). I have tried to set `miniters=1` but no success.
+ `USE_SMOOTHED_ERROR=False` gives `nan` prediction. Maybe I forget to clip some values.

I'll ping you another time when the second issue is resolved, so we can discuss about the structure for this notebook. I might come up with a preliminary version with core functionalities / explanations. But to make it great, I believe you will have more ideas than me. :D

neerajprad:
> tqdm is unusable due to this issue. I have tried to set miniters=1 but no success.

I'll take a look at that. Hopefully, we can sort out why the Rlgt model isn't converging with Slawek soon, and also the issue with the predictions not capturing the global trend.

neerajprad:
> I use the following trick. It is pretty fast.

That's a neat trick. Maybe we should document it or make a utility function out of it.

fehiepsi:
> I'll take a look at that.

Thanks!

fehiepsi:
@neerajprad This PR is ready to review. I made a separate #169 for tqdm issue. Currently, the compiling time is pretty slow (especially during the first `find_initial_step_size` to get initial `step_size`) due to [this issue](https://github.com/google/jax/issues/772). But it won't block us from getting mcmc result. Btw, the inference is pretty fast: it took `14s` to draw 5000 samples. The inferred parameters are comparable to the result from Stan:
```
coefTrend     95.85   12.80  222.20  -71.33    8.62   35.99   98.23   676.17   301 1.03
powTrend      -0.08    0.01    0.24   -0.47   -0.27   -0.08    0.09     0.38   954 1.00
sigma          1.31    0.03    1.17    0.36    0.61    0.95    1.54     4.56  1133 1.00
offsetSigma   21.18    0.42   17.61    0.70    7.75   16.92   30.14    65.88  1793 1.00
levSm          0.88    0.00    0.07    0.72    0.84    0.89    0.93     0.98  1850 1.00
sSm            0.49    0.01    0.28    0.03    0.26    0.49    0.73     0.97  1871 1.00
nu            12.12    0.10    4.69    3.27    8.43   12.41   16.09    19.54  2055 1.00
powx           0.86    0.00    0.09    0.65    0.81    0.87    0.93     0.99   946 1.01
```
The only difference is at `coefTrend` where we get mean value `124.41` while Stan gives `95.85`. But our result has a much higher `n_eff` and smaller `Rhat` (note that we only use 1 chain while the Stan demo uses 4 chains), so I guess that our result is better?

Forecasting shows a decent result, and is quite similar to the result from R demo.

Because you have been playing around with LGT, if you have more ideas to complete the tutorial, feel free to push to this branch. Otherwise, we can improve the content in subsequential PRs, when more ideas come.

neerajprad:
It is great to see that the model has learnt the seasonal component well. My only high-level comment is that it seems to have also learnt a positive (significant at that) global trend which seems a bit unexpected  given the training data. Is this also what you see on Stan? Related to this, I think it might be worth running this on a dataset that actually did have a global trend that we wanted to catch; maybe we could experiment with one of the M3 datasets. Unfortunately, I am unable to run the notebook / example as it is crashing on Mac for some reason, so I am unable to play around with it.

fehiepsi:
> Is this also what you see on Stan?

Yes. As I said, the forecasting result is similar to Stan.

> maybe we could experiment with one of the M3 datasets

Yes, maybe the model performs well in other datasets. But it might have performed pretty well on this dataset, given the scale of the data. I just update the notebook with a bit change in the forecasting algorithm (removing noisy effect when updating `level`) and get a much better rmse/smape score. According [this thread](https://statmodeling.stat.columbia.edu/2017/03/01/facebooks-prophet-uses-stan/#comment-433298), the result outperforms prophet & a predator-prey model.

> Unfortunately, I am unable to run the notebook / example as it is crashing on Mac for some reason, so I am unable to play around with it.

I just run it fine (except slow - a few minutes - compiling time) in new env with lastest jax/jaxlib from pypi.

fehiepsi:
> this might just be a prior / hyperparam setting

nice suggestion! I will change prior of coef_trend to see if trending effect is reduced.

fehiepsi:
Oppss... the rmse formula in R demo is wrong, so the result in the notebook is also wrong. We get rmse=1500, so it is not as good as ARMA & predator-prey model.

neerajprad:
My suggestions for extensions / improvement are:
 - This model doesn't capture the second order seasonality effect with a periodicity of 4 (i.e. the fourth peak is larger). I think there are extensions that can capture this which will be interesting to include.
 - As I mentioned earlier, it will also be nice if we we could run this on a more complex dataset with a global trend as well as seasonality. 

Regarding the current notebook, the range of hpdi is too large, going to a minimum of 0. My guess is that this is at least partly because the model isn't able to capture the second order seasonality mentioned above, and this posterior mass would shrink if we did so. Anyways, this is worth investigating and at least adding a comment about later.

> I just run it fine (except slow - a few minutes - compiling time) in new env with lastest jax/jaxlib from pypi.

Are you on linux? Maybe its just an OS specific issue. My versions are as follows. Since the system crashes, I'm not even able to replicate this issue on a more minimal example.

```
jax==0.1.35
jaxlib==0.1.16
```

My other comments are more minor and I would rather just push to this branch, but I guess I'll have to first figure out why its crashing on my system. Please feel free to merge this in the meantime.

fehiepsi:
Yes, I am in linux. I'll try another seasonal model then pass it to you. :D my target is to win other tools in that discussion thread. Indeed, there are 6 varieties of SGT (2 season methods, 3 trend methods) for us to try.

fehiepsi:
@neerajprad  The new model has rmse 1243.49, which wins ARMA and predator-prey models in the above thread. In addition, it can capture second order seasonality effect. I am satisfied with the result now so I want to pass it to you to play more with this model or try it with other dataset if you want. Hope that the crashing issue is resolved in your system. I guess you can try to delete each part of the model to see which code causes the issue. ^^!

neerajprad:
Looks great! I'm taking a look, and will push to this branch directly. Feel free to make further edits / revert any changes that I make.

neerajprad:
I just had one additional comment regarding:

> Indeed, this value can be guessed by looking at autocorrelation of the training data:

I think this might be more confusing than illustrative for a reader, since just looking at only the autocorrelation for the training data it isn't obvious why we chose 38. I think it might be better to point out to the original dataset plot and briefly mention the second order seasonality effect that has a periodicity ~ 40.

We can elaborate on many other things, but we can punt some of it to the more basic tutorial and can keep updating this as we get more inputs. Great work!

fehiepsi:
Thanks @neerajprad ! All the changes look great! Regarding the number 38, yes, let's focus on the plot and just mention that 38 is one of the highest-autocorrelation lags.

I will also remove the mention "Note that all prior parameters are retrieved from [this file](https://github.com/cbergmeir/Rlgt/blob/master/Rlgt/R/rlgtcontrol.R) in the original source, ~except for `coef_trend`.~" because it is legacy. The current prior/algo of model and forecast_fn are identical to original implementation.

neerajprad:
LGTM, please feel free to merge this at your convenience.

PR #120, opened 2019-04-24
Use native for loop to collect samples

fehiepsi:
This PR implements `fori_collect` whose name is inspired by [lisp's loop do/collect macro](http://cl-cookbook.sourceforge.net/loop.html). The overhead of concatenation is pretty small. In bernoulli HMC test, this is a bit slower than fori_append (14.25s vs 11.29s) while timing is the same in NUTS test.

The only model which I observe a **big** slowdown is logistic_regression test (2x slower for HMC algo).

- [x] Investigate the slowdown in the logistic_regression test (*due to the slowness of jax.numpy.stack*)

### DIFF
--- a/numpyro/mcmc.py
+++ b/numpyro/mcmc.py
@@ -43,7 +43,6 @@ def hmc(potential_fn, kinetic_fn=None, algo='NUTS'):
     vv_init, vv_update = velocity_verlet(potential_fn, kinetic_fn)
     trajectory_len = None
     momentum_generator = None
-    _wa_update = None
 
     def init_kernel(init_samples,
                     num_warmup_steps,
@@ -56,7 +55,7 @@ def hmc(potential_fn, kinetic_fn=None, algo='NUTS'):
                     run_warmup=True,
                     rng=PRNGKey(0)):
         step_size = float(step_size)
-        nonlocal momentum_generator, _wa_update, trajectory_len
+        nonlocal momentum_generator, trajectory_len
         trajectory_len = float(trajectory_length)
         z = init_samples
         z_flat, unravel_fn = ravel_pytree(z)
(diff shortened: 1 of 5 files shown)


numpyro/mcmc.py
```diff
@@ -79,20 +78,16 @@ def init_kernel(init_samples,
         hmc_state = HMCState(vv_state.z, vv_state.z_grad, vv_state.potential_energy, 0, 0.,
                              wa_state.step_size, wa_state.inverse_mass_matrix, rng_hmc)
 
+        jwa_update = jit(wa_update)
         if run_warmup:
-            hmc_state, _ = jit(fori_loop, static_argnums=(2,))(0, num_warmup_steps, warmup_update,
-                                                               (hmc_state, wa_state))
+            for t in range(num_warmup_steps):
```
neerajprad: Does this change the timing? If so, I would suggest doing this in a separate PR so that it is easier to figure out how the timings are affected by each change separately.

test/test_mcmc.py
```diff
@@ -50,8 +50,8 @@ def model(labels):
     hmc_state = init_kernel(init_params,
                             trajectory_length=10,
                             num_warmup_steps=warmup_steps)
-    hmc_states = fori_append(sample_kernel, hmc_state, num_samples,
-                             transform=lambda x: transform_fn(x.z))
+    hmc_states = fori_collect(num_samples, sample_kernel, hmc_state,
```
neerajprad: I found this to be the case with the version of `fori_append` that I was working with as well and it seemed like this test is particularly impacted due to the large number of samples that we draw. To check this hypothesis, maybe we can increase the number of samples in other tests and see how the comparison holds up.

numpyro/mcmc.py
fehiepsi: I thought that it will help us avoid 2 time compiling. But I agree that it is better to do in a separate PR.

test/test_mcmc.py
fehiepsi: Yup, I think that some tests with 20000 samples will be a good benchmark for us. I'll make another pass with 20000 samples after investigating the slowness of logistic regression model.

fehiepsi:
@neerajprad Currently, in bernoulli and categorical tests, the non-prim method is 2x slower in HMC (10s vs 20s), and 1.5x slower in NUTS (20s vs 30s) due to the overhead cost (in average, each sample adds 0.5ms overhead). So I keep everything as is. When we have bigger models to benchmark against, let's consider using non-prim version again.

PR #121, opened 2019-04-25
Implement various distributions for benchmarking

fehiepsi:
`trunccauchy` follows the pattern of `truncnormal` which introduces shape parameters `a` and `b` to truncate the distribution.

In addition,

- [x] remove parse_args because it is unnecessary and not suitable for some shape parameters which we don't want to broadcast (e.g. truncate params)

- [x] add clearer messages for not yet implemented methods

TODO:

- [x] add pareto distribution (which is used in baseball)


fehiepsi:
This PR is blocked by jvp rule for atan2

test/test_distributions.py
```diff
@@ -71,17 +73,21 @@ def test_continuous_shape(jax_dist, loc, scale, prepend_shape):
...
     (dist.t, (-1,), np.inf),
     (dist.t, (np.array([-2., 3]),), np.array([1., np.nan])),
+    (dist.trunccauchy, (), -1),
+    (dist.trunccauchy, (), np.array([1., -2])),
     (dist.uniform, (), -1),
     (dist.uniform, (), np.array([0.5, -2])),
 ], ids=idfn)
 def test_continuous_validate_args(jax_dist, dist_args, sample):
-    valid_args = (1,) * len(dist_args)
+    valid_args = [i + 1 for i in range(jax_dist.numargs)]
```
fehiepsi: this is to make `a` < `b` for trunccauchy

neerajprad:
>  remove parse_args because it is unnecessary and not suitable for some shape parameters which we don't want to broadcast (e.g. truncate params)

Is this description outdated, because I do not see `parse_args` removed in the PR?

numpyro/distributions/discrete.py
```diff
@@ -74,7 +86,7 @@ def _rvs(self, n, p):
         return device_put(sample)
 
     def _logpmf(self, x, n, p):
-        n, p = _promote_dtypes(n, p)
+        x, n, p = _promote_dtypes(x, n, p)
```
neerajprad: Not relevant to this PR, but I just noticed that we aren't normalizing logits and assuming that the logits passed in are normalized i.e. `sum(np.exp(logits)) == 1`. This won't be true when we pass in logits from NNs.

fehiepsi:
> Is this description outdated, because I do not see parse_args removed in the PR?

Oppsss, I mean `promote_args` >"<

fehiepsi:
> because I want to prioritize removing this old API as soon as possible.
Could we move the current API to contrib instead of removing? Although adding multivariate distributions has trouble me a lot, it is much easier now. And it is easier for me to just copy scipy implementation rather rethinking the logic.

numpyro/distributions/discrete.py
fehiepsi: I think that we just need to normalize logits in multivariate distributions and I already did it.

neerajprad:
> Although adding multivariate distributions has trouble me a lot, it is much easier now. And it is easier for me to just copy scipy implementation rather rethinking the logic.

Sure, we should only make the switch once we have all the distributions covered, and we can swap contrib with what we have now. We can keep it as long as they are serving some purpose, but once the tests and examples start using the alternate API, it will be best to remove the old code.

PR #123, opened 2019-04-26
Implement halfnorm, truncnorm, poisson distributions

### LINKED ISSUE #1: Wrap distributions in scipy.stats for use with JAX
To make the internals of the `sampler` and `logpdf` methods visible to the tracer and compatible with jax, we need to:
 - Write the operations in terms of `jax.numpy` operations. 
 - We need to use jax's count based random number generator (via `PRNGKey`), and not the default used by scipy distributions which is numpy's globally mutable `mtrand`. Currently, the `PRNGKey` is being passed in through the `random_state` kwarg to the distribution's `_rvs` method.

Currently, only the normal distribution is wrapped. We can similarly wrap the following distributions (we may need to wrap over the samplers / rewrite the logic using jax operations):

**Continuous**
 - [x] beta @fehiepsi 
 - [x] cauchy @fehiepsi 
 - [x] dirichlet @fehiepsi 
 - [x] exponential @fehiepsi 
 - [x] uniform
 - [x] gamma  @fehiepsi 
 - [x] lognormal @fehiepsi 
 - [x] normal
 - [x] student-t  @fehiepsi 

**Discrete**
 - [x] bernoulli @neerajprad 
 - [x] categorical @neerajprad 
 - [x] binomial
 - [ ] poisson
 - [x] multinomial

fehiepsi:
This PR resolves #1. Requests for new distributions should be made in separate issues.


neerajprad:
If there are any further distributions you need, please do add them to the new API instead, so that we aren't doing double work, and instead more interesting stuff. :)

fehiepsi:
Pls feel free to do so.

Because I don’t have any appreciation of pytorch/tensorflow api beside it makes things easier for current pyro users, I would prefer adding new distributions to scipy API given than less code to be written down for me. For example, instead of finding what norm.inverse cdf is (useful for truncated normal), I just need to find the corresponding one in scipy and use the corresponding functions in jax.numpy/scipy. Mimicing pytorch implementation would require extra works to find the corresponding numpy/scipy functions. But the implementation of two APIs would be quite similar, so we can easily copy over. Double work is only available when there is no assignment on issues I guess.

But there are more advanced features which you might want to support in the future such as conjugate distributions, independent distribution,... so it would be easier to have Pytorch api at the first place. You don’t even have to think about supporting any new features when scipy api lives on contrib. If you need to change handlers/svi to support pytorch api, please do so. I think that I only need to add a few methods to jax_frozen such as sample, log_prob with corresponding args then I can do inference with scipy api using pytorch inference api.

neerajprad:
I think we can keep the existing scipy distribution wrappers in contrib, if you find it useful. I may be biased, but as you mentioned the distribution's log density evaluation methods have a 1:1 correspondence, so I think it shouldn't require any more effort to wrap them in the new API. I understand why you might feel that way, since you have already gotten used to working around scipy's API. 🙂 However, my concern here is not that we won't be able to work around the deficiencies of the API, but rather that it will be harder to extend in the longer term. I think it will also be harder for new users to understand and contribute to, or as you pointed out, for us to take advantage of new distributions functionality that is available in PyTorch or Tensorflow. 

My observation has been that the  reason for having multiple class hierarchies like `generic`, `multivariate`, `discrete`, `frozen` and resulting code complexity (like args parsing) in scipy stems from two design decisions in scipy - supporting frozen / non-frozen distributions, and supporting `loc`, `scale` as default parameters in many distributions. Neither of these are important for the use cases that we care about. In fact, our current implementations does not even have consistent behavior with regard to frozen / non-frozen distribution instances, and our inference methods cannot use non-frozen instances. The other use case can be easily handled by TransformedDistribution, which composes very nicely with the new API, and adds minimal code complexity to existing classes. So now with the current API, we are in a place where we have inherited all of this from an old code base when we do not have any legacy reasons to do so, with minimal benefits. 

If there are any features you need in the new API or have concerns around the design, let us surely discuss and address that! If you still believe that it is easier for you to copy over some scipy distribution that isn't currently implemented, you could use these wrappers. I am happy to port them over to the new API. I would however suggest that we could at least start using the new API for distributions that are not currently implemented in scipy, rather than adding more functionality on top of the existing wrappers.

fehiepsi:
As I said, I have not seen deficiencies of the scipy API yet and it is faster to add new distributions available in scipy so it will be more convenience for me. I do concern that it might cause confusion for users when there are 2 APIs available, so at one point in the future, we should remove scipy api. In addition, I don't see any missing features from PyTorch API w.r.t Scipy API yet.

> The other use case can be easily handled by TransformedDistribution, which composes very nicely with the new API, and adds minimal code complexity to existing classes

Incidentally, I have some thoughts about transformed distribution in the other PR. Transformed distributions, truncated distributions will play nicely with pytorch api but might add some performance issues because we have to pass through intermediate variables and no cancellation is involved. But it is easy to change I guess when we have scipy implementation to compare.

> If there are any features you need in the new API or have concerns around the design, let us surely discuss and address that!

There is nothing yet. Features like `cdf`, `icdf` are not necessary right now to me.

> I would however suggest that we could at least start using the new API for distributions that are not currently implemented in scipy, rather than adding more functionality on top of the existing wrappers.

Yup, I intended to do so for at least MVN and LKJ. There is no need to put it into contrib scipy.

neerajprad:
Thanks for clarifying. 

> Incidentally, I have some thoughts about transformed distribution in the other PR. Transformed distributions, truncated distributions will play nicely with pytorch api but might add some performance issues because we have to pass through intermediate variables and no cancellation is involved

I'm not sure if that would be a significant overhead, but we should profile and see. In JAX, I think there should be no overhead for this at least under JIT, but if you notice otherwise, let us follow up on this.

fehiepsi:
I just take a pass to pytorch distributions and catch [this bug](https://github.com/pytorch/pytorch/issues/15681) as an example for what I mentioned above. Looking like there is nothing wrong with the current implementations, but I'll take a closer look in your transition PR because I remember that I faced something similar to that bug (even before when I worked on LKJ). :)

PR #124, opened 2019-04-26
Add more distributions to contrib

neerajprad:
This adds most of the distributions that are being currently used to `contrib`. Once this merges, I'll add some more tests, and have our existing tests / examples start using this API. This has become a priority given that we have many more examples in the pipeline, and the migration will be tougher the more we keep adding to the scipy API. 

The reason why I believe that it is a good idea to not build on top of scipy's API is because it has many quirks that may be expected from an old distribution ([e.g.](https://github.com/scipy/scipy/blob/master/scipy/stats/_distn_infrastructure.py#L972)), and there is no reason for a new codebase to inherit these warts in the API. PyTorch's codebase or TensorFlow Probability is a much better template to follow in this regard, and will be easier to maintain as well as for new users to contribute to. This will also make it easier to plug this into `pyro.generic` with minimal boilerplate code.

Major changes:
 - changed `is_reparametrized` to `reparametrized_params` since a distribution may be reparametrizable w.r.t. some params and not others.
 - augmented multinomial sampler to take in batched `n`. The scalar `n` case is the same as earlier. Batched `n` while allowed is slow. I think when we have a faster, improved implementation, we can consider moving it upstream.
 - Simple poisson sampler based on Knuth's algorithm (more efficient sampler algorithms are available but this is probably good enough).

### DIFF
--- a/numpyro/contrib/distributions/continuous.py
+++ b/numpyro/contrib/distributions/continuous.py
@@ -66,7 +66,7 @@ class Beta(Distribution):
 
 
 class Cauchy(Distribution):
-    reparametrized_params = [(0, 'loc'), (1, 'scale')]
+    reparametrized_params = ['loc', 'scale']
     arg_constraints = {'loc': constraints.real, 'scale': constraints.positive}
     support = constraints.real
 
@@ -115,9 +115,10 @@ class Dirichlet(Distribution):
     def log_prob(self, value):
         if self._validate_args:
             self._validate_sample(value)
+        normalize_term = (gammaln(np.sum(self.concentration, axis=-1)) -
+                          np.sum(gammaln(self.concentration), axis=-1))
         return (np.sum(np.log(value) * (self.concentration - 1.), axis=-1) +
-                gammaln(np.sum(self.concentration, axis=-1)) -
-                np.sum(gammaln(self.concentration), axis=-1))
+                normalize_term)
 
     @property
     def mean(self):
@@ -150,8 +151,9 @@ class Gamma(Distribution):
     def log_prob(self, value):
         if self._validate_args:
             self._validate_sample(value)
-        return (self.concentration * np.log(self.rate) + (self.concentration - 1) * np.log(value) -
-                self.rate * value - gammaln(self.concentration))
+        normalize_term = (self.concentration * np.log(self.rate) -
+                          self.rate * value - gammaln(self.concentration))
+        return (self.concentration - 1) * np.log(value) + normalize_term
 
     @property
     def mean(self):
@@ -170,7 +172,7 @@ class Chi2(Gamma):
 
 
 class Exponential(Distribution):
-    reparametrized_params = [(0, 'rate')]
+    reparametrized_params = ['rate']
     arg_constraints = {'rate': constraints.positive}
     support = constraints.positive
 
@@ -197,7 +199,7 @@ class Exponential(Distribution):
 
 
 class HalfCauchy(TransformedDistribution):
-    reparametrized_params = [(0, 'scale')]
+    reparametrized_params = ['scale']
     arg_constraints = {'scale': constraints.positive}
     support = constraints.positive
 
@@ -226,7 +228,7 @@ class HalfCauchy(TransformedDistribution):
 class Normal(Distribution):
     arg_constraints = {'loc': constraints.real, 'scale': constraints.positive}
     support = constraints.real
-    reparametrized_params = [(0, 'loc'), (1, 'scale')]
+    reparametrized_params = ['loc', 'scale']
 
     def __init__(self, loc, scale, validate_args=None):
         self.loc, self.scale = promote_shapes(loc, scale)
@@ -255,7 +257,7 @@ class Normal(Distribution):
 class LogNormal(TransformedDistribution):
     arg_constraints = {'loc': constraints.real, 'scale': constraints.positive}
     support = constraints.positive
-    reparametrized_params = [(0, 'loc'), (1, 'scale')]
+    reparametrized_params = ['loc', 'scale']
 
     def __init__(self, loc, scale, validate_args=None):
         base_dist = Normal(loc, scale)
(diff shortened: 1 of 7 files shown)


neerajprad:
The failure in `test_distributions.py` seems to be unrelated.

numpyro/distributions/util.py
```diff
@@ -339,6 +340,26 @@ def categorical_rvs(key, p, shape=()):
...
+    def body_fn(val):
+        k, p, rng = val
+        k = np.where(p > L, k + 1, k)
+        rng, rng_u = random.split(rng)
+        u = random.uniform(rng_u, shape)
+        p = p * u
+        return k, p, rng
+
+    k, _, _ = lax.while_loop(lambda val: np.any(val[1] > L), body_fn, (k, p, key))
+    return k - 1
```
fehiepsi: +1 nice implementation

numpyro/distributions/util.py
```diff
@@ -348,16 +369,29 @@ def _scatter_add_one(operand, indices, updates):
...
+        excess = 0
     # NB: we transpose to move batch shape to the front
-    indices_2D = indices.reshape((n, -1)).T
+    indices_2D = (np.reshape(indices * mask, (n_max, -1,))).T
     samples_2D = vmap(_scatter_add_one, (0, 0, 0))(np.zeros((indices_2D.shape[0], p.shape[-1]),
                                                             dtype=indices.dtype),
                                                    np.expand_dims(indices_2D, axis=-1),
                                                    np.ones(indices_2D.shape, dtype=indices.dtype))
-    return samples_2D.reshape(shape + p.shape[-1:])
+    return samples_2D.reshape(shape + p.shape[-1:]) - excess
```
fehiepsi: I haven't taken a closer look here but your approach makes sense. I will make an issue later instead if I find any issue of this sampler.

numpyro/contrib/distributions/continuous.py
```diff
@@ -120,9 +252,56 @@ def variance(self):
         return np.broadcast_to(self.scale ** 2, self.batch_shape)
 
 
+class LogNormal(TransformedDistribution):
+    arg_constraints = {'loc': constraints.real, 'scale': constraints.positive}
+    support = constraints.positive
+    reparametrized_params = [(0, 'loc'), (1, 'scale')]
```
fehiepsi: just curious: why we need `0`, `1` here?

neerajprad:
Thanks for a thorough review, @fehiepsi! I'll address / respond to your comments shortly.

numpyro/contrib/distributions/discrete.py
```diff
@@ -1,35 +1,72 @@
...
-)
+    xlogy,
+    poisson)
 
 
-def _to_probs_bernoulli(logits):
-    return 1 / (1 + np.exp(-logits))
+def clamp_probs(probs):
+    finfo = np.finfo(get_dtypes(probs)[0])
+    return np.clip(probs, a_min=finfo.tiny, a_max=1. - finfo.eps)
```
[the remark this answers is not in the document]
neerajprad: @fehiepsi : That's the reason why I'm using `1. - finfo.eps` for `a_max`. `tiny` shouldn't be used for that - it will result in a `1.` and lead to issues.

numpyro/distributions/util.py
neerajprad: It is unfortunately a bit slow - but I don't think there is a straightforward way to get a fast batched multinomial sampler by just using jax primitives.

numpyro/contrib/distributions/continuous.py
neerajprad: I'm not too happy with this either. Will change it to something else, if there is a better way to do this. `0, 1` denote the position of the args, and this makes it easy to test. Another way might be to have an OrderedDict denoting whether the parameter is reparametrized `{'p1': False, 'p2': True}`, but that seems equally kludgey. Will think about it.

neerajprad: Okay, changed to list. I'm just using `inspect` in the test now, so this kludge doesn't need to get into the actual code.

numpyro/contrib/distributions/continuous.py
```diff
@@ -120,9 +252,56 @@ def variance(self):
...
+        self.scale, self.alpha = np.broadcast_to(scale, batch_shape), np.broadcast_to(alpha, batch_shape)
+        base_dist = Exponential(self.alpha)
+        transforms = [ExpTransform(), AffineTransform(loc=0, scale=self.scale)]
+        super(Pareto, self).__init__(base_dist, transforms, validate_args=validate_args)
+
+    @property
+    def mean(self):
+        # mean is inf for alpha <= 1
+        a = np.clip(self.alpha, a_min=1.)
+        return a * self.scale / (a - 1)
```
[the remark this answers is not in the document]
neerajprad: Done

neerajprad:
> Just a note that TransformedDistribution is nice but it can cause instability problems when computing log_prob, and the computation costs more time I guess (I remember that I face this issue while implementing LKJ previously)

Let us create an issue, and dive deeper into this, since it also possibly affects pytorch.

numpyro/contrib/distributions/discrete.py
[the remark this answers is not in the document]
fehiepsi: Sorry, I mean that the issue happens for some operators which requires `1 - p < 1`. But we have already addressed them at StickBreakingTransform (at least). I see no other issues in the log_prob of current distributions.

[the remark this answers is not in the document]
neerajprad: Your observation is right. Note that `tiny` is the smallest representable number with the given dtype. We cannot do `1. - tiny` and expect it to be less than 1. `eps` on the other hand gives the smallest number such that `1 + eps != 1`, and that works also with `1 - eps != 1` (as far as I can tell). That's the reason why `a_min` uses `tiny` but `a_max` uses `eps` so that the result is guaranteed to be in `(0, 1)` (exclusive).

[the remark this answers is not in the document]
fehiepsi: Understood!

PR #126, opened 2019-04-28
Add CorrCholesky transform

fehiepsi:
This PR adds CorrCholesky transform. Because each row of the triangular part of the cholesky of a correlation matrix has norm 1 (or diagonal part of correlation matrix is 1), we can apply StickBreaking process for the square of each row. Using that fact, we can vectorize this transform and avoid "for" loops in the transform.

### DIFF
--- a/numpyro/distributions/transforms.py
+++ b/numpyro/distributions/transforms.py
@@ -26,7 +26,7 @@ import jax.numpy as np
 from jax.scipy.special import expit, logit
 
 from numpyro.distributions import constraints
-from numpyro.distributions.util import cumprod, cumsum, sum_rightmost
+from numpyro.distributions.util import cumprod, cumsum, matrix_to_tril_vec, sum_rightmost, vec_to_tril_matrix
 
 
 def _clipped_expit(x):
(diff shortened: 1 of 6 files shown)


test/test_distributions.py
```diff
@@ -434,8 +439,9 @@ def test_constraints(constraint, x, expected):
     assert_array_equal(constraint(x), expected)
 
 
-@pytest.mark.parametrize('shape', [(), (1,), (3,), (5,), (3, 1), (1, 3), (5, 3)], ids=idfn)
+@pytest.mark.parametrize('shape', [(), (1,), (3,), (6,), (3, 1), (1, 3), (5, 3)], ids=idfn)
```
fehiepsi: We change this to 6 to make sure that it is the length of a lower tril vector (of 4 x 4 matrix)

test/test_distributions.py
```diff
@@ -467,8 +473,14 @@ def test_biject_to(constraint, shape):
     assert np.shape(actual) == batch_shape
     if len(shape) == transform.event_dim:
         if constraint is constraints.simplex:
-            expected = np.linalg.slogdet(jax.jacobian(transform)(x)[:-1, :])[1]
-            inv_expected = np.linalg.slogdet(jax.jacobian(transform.inv)(y)[:, :-1])[1]
+            expected = onp.linalg.slogdet(jax.jacobian(transform)(x)[:-1, :])[1]
```
fehiepsi: using `onp` because `slogdet` does not work in GPU

neerajprad:
Looks great! I will take a detailed look later tonight and merge this.

test/test_distributions.py
neerajprad: How is this handled on the gpu? Does jax send the array to the CPU for the numpy operation and transfer the result back?

numpyro/distributions/transforms.py
```diff
@@ -129,6 +129,91 @@ def log_abs_det_jacobian(self, x, y):
...
+        # inverse stick-breaking
+        z1m_cumprod = 1 - cumsum(y * y)
+        pad_width = [(0, 0)] * y.ndim
+        pad_width[-1] = (1, 0)
+        z1m_cumprod_shifted = np.pad(z1m_cumprod[..., :-1], pad_width,
+                                     mode="constant", constant_values=1.)
+        t = matrix_to_tril_vec(y, diagonal=-1) / np.sqrt(
+            matrix_to_tril_vec(z1m_cumprod_shifted, diagonal=-1))
+        # inverse of tanh
+        x = np.log((1 + t) / (1 - t)) / 2
```
neerajprad: Can we test how numerically stable this is, e.g. by applying the forward and inverse transforms? The LKJ implementation in Pyro brought up issues with numerical stability, but I do not recall what they were.

fehiepsi: We did this test for all transforms at [this line](https://github.com/pyro-ppl/numpyro/pull/126/files#diff-51c2d9ee9d114f18441e3579ec31fdbeR466).

fehiepsi: For some reason, I can't reply to the `slogdet` comment. In GPU, `onp.linalg.slogdet` returns a numpy array.

test/test_distributions.py
fehiepsi: In GPU, `onp.linalg.slogdet` returns a numpy array.

neerajprad: Wouldn't that be an issue when we try to use that with a GPU based array?

PR #127, opened 2019-04-28
Miscellaneous support for current blocked benchmarks

### LINKED ISSUE #88: Avoid compiling 2 times in HMC
Currently, we do compiling 2 times: 1 at init, 1 at sample, so it would be better if we only compile 1 time.

One way is to modify `tscan` with additional parameters: `skip` to skip first warmup_steps; put warmup_state to a lax.cond to decide if it will be updated or not (depending on we are in warmup phase or sampling phase).

fehiepsi:
This PR does not change previous behaviours but incorporates some lessons from current benchmark notebooks. In addition, this adds an option to resolves #88 by using `use_prims=False` in `init_kernel` and `fori_collect`.

- [x] Expose max_tree_depth 

- [x] Add an option `heuristic_step_size` to temporary disable `find_reasonable_step_size`. Will remove it when its slowness in large models is resolved. 

- [x] Add an option `use_prims` to fori_collect to make it achieve the same performance as `fori_append` (for small models, which has fast compiling)

- [x] Remove legacy tscan.

- [x] Allow using `tqdm.trange` in `fori_collect` (as demonstrated in logistic_regression notebook). This only takes effect when `use_prims=False`.

- [x] Update logistic regression notebook to use `fori_collect`.


notebooks/logistic_regression.ipynb
```diff
@@ -148,42 +123,41 @@
...
-   "execution_count": 6,
+   "execution_count": 5,
    "metadata": {},
    "outputs": [
     {
      "name": "stderr",
      "output_type": "stream",
      "text": [
-      "100%|██████████| 100/100 [00:19<00:00,  5.19it/s]\n"
+      "100%|██████████| 100/100 [00:03<00:00, 27.10it/s]\n"
```
fehiepsi: `fori_collect` is faster than the previous custom `mcmc`

numpyro/examples/covtype.py
```diff
@@ -68,17 +68,20 @@ def benchmark_hmc(args, features, labels):
...
+                                  trajectory_length=trajectory_length,
+                                  heuristic_step_size=False, run_warmup=False)
+    sample_kernel(hmc_state.update(step_size=1.))
     t1 = time.time()
     print("time for hmc_init: ", t1 - t0)
 
     def transform(state): return {'coefs': state.z['coefs'],
                                   'num_steps': state.num_steps}
 
-    hmc_state = hmc_state.update(step_size=step_size)
```
fehiepsi: we don't need to update `step_size` again when setting `heuristic_step_size=False`

numpyro/util.py
```diff
@@ -240,22 +142,33 @@ def fn_loop(n, x): return lax.fori_loop(0, n, body_fun, x)
...
+    if use_prims:
+        collection = np.zeros((n,) + init_val_flat.shape, dtype=init_val_flat.dtype)
+
+        def _body_fn(i, vals):
+            val, collection = vals
+            val = body_fun(val)
+            collection = ops.index_update(collection, i, ravel_fn(val))
+            return val, collection
+
+        _, collection = jit(lax.fori_loop, static_argnums=(2,))(0, n, _body_fn,
```
neerajprad: How does this version compare with `fori_append`?

numpyro/mcmc.py
```diff
@@ -79,12 +89,20 @@ def init_kernel(init_samples,
...
-            hmc_state, _ = jit(fori_loop, static_argnums=(2,))(0, num_warmup_steps, warmup_update,
-                                                               (hmc_state, wa_state))
+            if use_prims:
+                hmc_state, _ = jit(fori_loop, static_argnums=(2,))(0, num_warmup_steps,
+                                                                   warmup_update,
+                                                                   (hmc_state, wa_state))
+            else:
+                n_iter = (num_warmup_steps if isinstance(num_warmup_steps, Iterable)
+                          else range(num_warmup_steps))
+                for i in n_iter:
```
neerajprad: Since the only reason why we imagine this will be useful is for diagnostics / progress bar updates, I think we can just use tqdm directly instead of having to change the type of `num_warmup_steps` to support it. That way we can plug in other diagnostics too from `hmc_state`.

numpyro/mcmc.py
```diff
@@ -53,24 +55,32 @@ def init_kernel(init_samples,
...
+        max_treedepth = max_tree_depth
         z = init_samples
         z_flat, unravel_fn = ravel_pytree(z)
         momentum_generator = partial(_sample_momentum, unravel_fn)
 
-        find_reasonable_ss = partial(find_reasonable_step_size,
-                                     potential_fn, kinetic_fn, momentum_generator)
+        wa_kwargs = {}
+        # FIXME: compiling find_reasonable_step_size is so slow for some models
+        if heuristic_step_size:
```
neerajprad: Can you elaborate why we need this, and setting `adapt_step_size=False` is not sufficient?

numpyro/examples/covtype.py
```diff
@@ -68,17 +68,20 @@ def benchmark_hmc(args, features, labels):
     t0 = time.time()
     # TODO: Use init_params from `initialize_model` instead of fixed params.
     hmc_state, _, _ = init_kernel(init_params, num_warmup_steps=0, step_size=step_size,
-                                  trajectory_length=trajectory_length, run_warmup=False,
-                                  adapt_step_size=False)
+                                  trajectory_length=trajectory_length,
+                                  heuristic_step_size=False, run_warmup=False)
+    sample_kernel(hmc_state.update(step_size=1.))
```
neerajprad: Same comment here - I do not fully understand why this change is needed.

neerajprad:
>  In addition, this adds an option to resolves #88 by using use_prims=False in init_kernel and fori_collect.

How do the times compare on the `covtype.py` example with the two options?

numpyro/util.py
fehiepsi: The speed is the same as fori_append in test models.

neerajprad: In that case, let us just profile on `test_mcmc.py` and remove `fori_append`, since there shouldn't be a reason to have both of these?

numpyro/mcmc.py
fehiepsi: Do you want to add `tqdm` as a dependent? If that is the case, then I'll add a flag `prog_bar` as in Pyro for it.

neerajprad: I think we will provide some default diagnostics at some point, so we might as well add it. It's a light weight dependency!

numpyro/mcmc.py
fehiepsi: With this, we still adapt step_size but do not tuning step_size when mass_matrix is updated (at the end of each window). Indeed, tuning step_size is not important under dual averaging scheme.

fehiepsi: More context about this: there is a bug in Stan (even with latest version), where `find_reasonable_step_size` does not run at beginning of warmup phase; but step_size will change very fast based on `accept_prob` value. If given a step_size, accept_prob is small, then step_size will reduce a lot under dual averaging scheme.

However, tuning step_size is useful to help NUTS/HMC not spend so many steps at initial steps of each window. For example, if `inv_mm` changes from `1 -> 0.001` then `find_reasonable_ss` will help increase step_size to a much larger value than the current value.

numpyro/examples/covtype.py
fehiepsi: I just want to compile `sample_kernel` separating from running `sampling`. (assume that it is compiled from warmup phase).

fehiepsi:
> an end user probably shouldn't have to know or choose between fori_collect, fori_append, use_heuristic_step_size, use_prims, etc.

When I made the change in `hmc_init` to avoid recompiling, I have to make a change to `fori_append` too, but you seem don't think so. So it is better to make another version to avoid recompiling myself.

`use_heuristic_step_size` is temporary until we can resolve the slowness.

`use_prims` is important to keep the same performance as `fori_append`. I don't want to use prims but it is hard to persuade you drop it so it is better to keep it until more benchmark models are available. New models can also prove that I am wrong.

> a lot of benchmarking related constructs and hacks which were earlier confined to the notebook are leaking into the API with this change

I don't understand what do you mean by *leak* here. First, we have `tscan` which is indeed a hack because I don't understand how to use utilities in jax (so it is easier to mimic an available `scan` version). Now, we have simpler versions such as fori_append, fori_collect. If we detect something is not working as expected, then it is better to add an option to disable it until we can resolve the problem. As an example, do you think that the flag `_DISABLE_CONTROL_FLOW_PRIM` is a hack when it is hard for you to debug?

numpyro/mcmc.py
neerajprad: > With this, we still adapt step_size but do not tuning step_size when mass_matrix is updated (at the end of each window).

IIRC, we weren't doing this in Pyro initially either, but found that updating step size when mass matrix is updated helps prevent cases where we make very slow progress due to the mismatch that you mentioned above. So, in what situations, do you think turning this off is advisable? Is this mostly needed because compilation of `find_reasonable_step_size` is slow on larger models?

fehiepsi:
> How do the times compare on the covtype.py example with the two options?

`non_prims` would be faster because I pre-compile sample_kernel.

numpyro/examples/covtype.py
neerajprad: Since `sample_kernel` isn't jitted here, I would be very surprised if this helps.

How about we keep the notebook for profiling purposes (where we can have hacks to ensure faster compilation, etc.), but have the examples scripts reflect how we expect users to use the HMC interface?

numpyro/mcmc.py
fehiepsi: Yes, it is only there until this `FIXME` is resolved.

neerajprad:
> use_prims is important to keep the same performance as fori_append. I don't want to use prims but it is hard to persuade you drop it so it is better to keep it until more benchmark models are available. New models can also prove that I am wrong.

I do not have any issues with not using jax primitives and using native python loop, but if it is 100% slower on certain models, then it is nice to have the option of doing faster inference. We can still simplify the API as follows:
 - Remove `fori_append`, since `fori_collect` is competitive with the `use_prims` option.
 - Use `progbar=True` which uses the `fori_collect` version without jax primitives. 

> It is a good way to do IMO unless you have a better suggestion. As an example, do you think that the flag _DISABLE_CONTROL_FLOW_PRIM is a hack when it is hard for you to debug?

I do not actually understand the full extent of the problem, because I have only worked on smallish examples where compilation isn't as much of an issue to begin with, so I will defer to your judgment on how to make it more efficient. For the second question, I don't think that it impacts the API of the MCMC classes, so it is not a hack. It is a debugging utility.

fehiepsi:
> Remove fori_append, since fori_collect is competitive with the use_prims option.

I would defer this job for you. :)

> Use progbar=True which uses the fori_collect version without jax primitives.

Just did it in the last commit.

> For the second question, I don't think that it impacts the API of the MCMC classes, so it is not a hack. It is a debugging utility.

Hmm, that is not a good example from me. By a better suggestion, I mean a solution for the API, not a solution for performance. I guess we can add a context manager to disable `find_reasonable_step_size` so user interface will be the same but we can still have an option to disable it for running the benchmarks.

neerajprad:
> By a better suggestion, I mean a solution for the API, not a solution for performance. I guess we can add a context manager to disable find_reasonable_step_size so user interface will be the same but we can still have an option to disable it for running the benchmarks.

That sounds good. I think, based on this discussion, it might be best to clean up the API once the next JAX release is out and we have more clarity on all the performance related issues. I will create a task to revisit this before release.

PR #128, opened 2019-04-29
Use contrib.distributions in tests and examples

### LINKED ISSUE #82: Remove dependency on scipy.stats for distributions

neerajprad:
Fixes #82 
 - Adds some temporary utilities for compatibility with existing wrapper, which should be removed before release. For the time being, this allows us to switch between the two wrappers.
 - Moved tests / examples (not notebooks) to use the new wrappers.
 - Added constraint tests, and fixed bugs that were caught in existing implementations.

### DIFF
--- a/numpyro/contrib/distributions/__init__.py
+++ b/numpyro/contrib/distributions/__init__.py
@@ -1,3 +1,5 @@
+from jax import numpy as np
+
 from numpyro.contrib.distributions.continuous import (
     Beta,
     Cauchy,
@@ -47,3 +49,50 @@ __all__ = [
     'StudentT',
     'Uniform',
 ]
+
+# TODO: remove before release
+
+
+def get_bern(p, is_logits=False):
+    if is_logits:
+        return BernoulliWithLogits(p)
+    else:
+        return Bernoulli(p)
+
+
+def get_binom(n, p, is_logits=False):
+    if is_logits:
+        return BinomialWithLogits(p, n)
+    else:
+        return Binomial(p, n)
+
+
+def get_categorical(p, is_logits=False):
+    if is_logits:
+        return CategoricalWithLogits(p)
+    else:
+        return Categorical(p)
+
+
+def get_multinomial(n, p, is_logits=False):
+    if is_logits:
+        return MultinomialWithLogits(p, n)
+    else:
+        return MultinomialWithLogits(p, n)
+
+
+beta = Beta
+bernoulli = get_bern
+binom = get_binom
+cauchy = lambda loc=0., scale=1.: Cauchy(loc, scale)
+categorical = get_categorical
+dirichlet = lambda alpha: Dirichlet(concentration=alpha)
+expon = Exponential
+gamma = lambda conc, scale=1.: Gamma(conc, rate=1./scale)
+halfcauchy = lambda scale: HalfCauchy(scale)
+lognorm = lambda s, scale=1.: LogNormal(loc=np.log(scale), scale=s)
+multinomial = get_multinomial
+norm = lambda loc=0., scale=1.: Normal(loc, scale)
+pareto = lambda alpha, scale=1.: Pareto(scale, alpha)
+t = lambda df, loc=0., scale=1.: StudentT(df, loc, scale)
+uniform = Uniform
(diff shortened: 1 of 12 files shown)


numpyro/contrib/distributions/__init__.py
```diff
@@ -47,3 +51,50 @@
...
+dirichlet = lambda alpha: Dirichlet(concentration=alpha)
+expon = Exponential
+gamma = lambda conc, scale=1.: Gamma(conc, rate=1./scale)
+halfcauchy = lambda scale: HalfCauchy(scale)
+lognorm = lambda s, scale=1.: LogNormal(loc=np.log(scale), scale=s)
+multinomial = get_multinomial
+norm = lambda loc=0., scale=1.: Normal(loc, scale)
+pareto = lambda alpha, scale=1.: Pareto(scale, alpha)
+t = lambda df, loc=0., scale=1.: StudentT(df, loc, scale)
+uniform = Uniform
```
fehiepsi: Why don't we just copy the files over the new places instead of making wrappers like this?

numpyro/distributions/constraints.py
```diff
@@ -78,6 +78,14 @@ def __call__(self, value):
         return (value > self.lower_bound) & (value < self.upper_bound)
 
 
+class _Multinomial(Constraint):
+    def __init__(self, upper_bound):
+        self.upper_bound = upper_bound
+
+    def __call__(self, value):
+        return np.sum(value, -1) == self.upper_bound
```
fehiepsi: We miss a check for nonnegative here. And I guess you mean `total_count` instead of `upper_bound`?

numpyro/hmc_util.py
```diff
@@ -543,7 +544,12 @@ def _body_fn(state):
 
 def log_density(model, model_args, model_kwargs, params):
     def logp(d, val):
-        return d.logpdf(val) if isinstance(d.dist, jax_continuous) else d.logpmf(val)
+        if isinstance(d, Distribution):
+            return d.log_prob(val)
+        elif isinstance(d.dist, jax_continuous):
+            return d.logpdf(val)
+        else:
+            return d.logpmf(val)
```
fehiepsi: I think you can go ahead and drop supporting `jax_continuous/discrete`. We can add a `log_prob` method to `jax_frozen` and still have all versions work.

numpyro/distributions/constraints.py
neerajprad: `total_count` is stored as `upper_bound`. I just wanted to keep the same convention as the rest of the constraints class in terms of using `upper_bound` and `lower_bound` as attributes, but we can change that if it looks confusing. Will add the non-negativity constraint.

numpyro/contrib/distributions/__init__.py
neerajprad: I am just doing this in phases. Once this PR merges, I will have a simple PR that moves around the files and changes the names etc. Didn't want to put in too many changes in one PR.

numpyro/hmc_util.py
neerajprad: Sounds good. I'll do this separately, with the name changes and file reorg.

numpyro/distributions/constraints.py
```diff
@@ -88,6 +88,14 @@ def __call__(self, value):
         return (value > self.lower_bound) & (value < self.upper_bound)
 
 
+class _Multinomial(Constraint):
+    def __init__(self, upper_bound):
+        self.upper_bound = upper_bound
+
+    def __call__(self, value):
+        return (value >= 0) & (np.sum(value, -1) == self.upper_bound)
```
fehiepsi: I think we should use `np.all(value >=0, axis=-1)` instead of `value >= 0` to return a correct shape.

neerajprad: Oops..you are absolutely right, will change.

PR #129, opened 2019-04-30
Remove fori_append in favor of fori_collect

neerajprad:
This removes another control flow primitive that we had accumulated in our collection. :) This isn't needed since `fori_collect` with jax primitives seems to be better or at least competitive on tests and examples that I tried it on. This also remove the `use_prims` arg from `hmc`. Changes are made to the notebook, but I haven't run it so as not to pollute the existing output with runs from a different system.


fehiepsi:
Thanks for making this PR! I just want to check with you two things:
+ If `progbar=True`, we use slow version (for small models); and if we set it to False, then we use a fast version?
+ We now use slower version (progbar=True) in test?

notebooks/logistic_regression.ipynb
```diff
@@ -154,10 +153,10 @@
    ],
    "source": [
     "start = time.time()\n",
-    "hmc_states = fori_collect(tqdm.trange(num_samples), sample_kernel, hmc_state,\n",
+    "hmc_states = fori_collect(num_samples, sample_kernel, hmc_state,\n",
     "                          transform=lambda state: {\"coefs\": state.z[\"coefs\"],\n",
     "                                                   \"num_steps\": state.num_steps},\n",
-    "                          use_prims=False)\n",
+    "                          progbar=True)\n",
```
fehiepsi: this won't take effect unless you also make a similar change to `fori_collect` (removing `use_prims` arg)

neerajprad:
If progbar=True, we use slow version (for small models); and if we set it to False, then we use a fast version?

That's right - if the user asks for intermediate diagnostics, we use the slow version, otherwise we aggressively use JIT whenever possible.

We now use slower version (progbar=True) in test?
Why you still keep use_prims from fori_collect?

I completely missed that `fori_collect` also takes in a `progbar` argument in turn. I'll make the change to remove that and restore tests to the faster version.

fehiepsi:
> That's right - if the user asks for intermediate diagnostics, we use the slow version, otherwise we aggressively use JIT whenever possible.

Sound good, totally agree! :D

notebooks/logistic_regression.ipynb
neerajprad: Done.

PR #131, opened 2019-04-30
Test some edge examples from Pyro

fehiepsi:
This PR adds tests for:

- Stability of `beta binomial` model. It seems that our implementation for binomial's logprob is good.

- Change point detection from https://forum.pyro.ai/t/i-dont-understand-why-nuts-code-is-not-working-bayesian-hackers-mail/696 . In this example, by turning `prob_bar=True`, we can see it took a while to pass the first step (indeed, it took more than 10s to compile, and just take 1s to run warmup and 1s to get 1000 samples!), then everything will be very fast (much much faster than Pyro; in Pyro, I have to set `max_tree_depth=7` to make things run in a reasonable amount of time, otherwise it might took more than 20minutes to finish).

In addition, I remove unnecessary code from lkj transform. By moving the sum of `r` with identity to the end, I also make sure that `z1m_cumprod` will have all entries non-zero, so taking `sqrt` will be fine.

TODO:

- [x]  update tests to use new API


fehiepsi:
@neerajprad I mark `awaiting review` to mean that it is ready to take a look. I will update these tests when #130 is merged.

test/test_distributions.py
```diff
@@ -345,3 +347,94 @@ def test_distribution_constraints(jax_dist, sp_dist, params, prepend_shape):
...
+    (constraints.unit_interval, 0.1, True),
+    (constraints.unit_interval, np.array([-5, 0, 0.5, 1, 7]),
+     np.array([False, False, True, False, False])),
+])
+def test_constraints(constraint, x, expected):
+    assert_array_equal(constraint(x), expected)
+
+
+@pytest.mark.parametrize('shape', [(), (1,), (3,), (6,), (3, 1), (1, 3), (5, 3)])
+@pytest.mark.parametrize('constraint', [
```
neerajprad: Thanks for moving these over.

PR #132, opened 2019-05-01
Make utilities work for the newer version of JAX

### LINKED ISSUE #112: Some custom primitives does not work with the newest version of jax
The error happens because new version of jax requires a specific pattern for using `jit`'s static_argnums. See https://github.com/google/jax/pull/595 for more context. The failed functions include:

- [ ] xlogy
- [ ] xlog1py
- [ ] standard_gamma

In addition, I think that we can simplify the implementation of these functions by using decorator `custom_transform` as in `cumsum`, `cumprod`.

cc @neerajprad

fehiepsi:
Fix #112 by using `custom_transform` decorator.

In addition, this PR fixes tests with for fastmath mode (Peter (jax devs) mentioned that `np.inf`, `np.nan`, `np.exp` is fixed in the newer version of jaxlib so I would like to test it out and fixes errors during the way).

### DIFF
--- a/numpyro/distributions/util.py
+++ b/numpyro/distributions/util.py
@@ -7,9 +7,8 @@ import scipy.special as osp_special
 
 import jax.numpy as np
 from jax import canonicalize_dtype, custom_transforms, jit, lax, random, vmap
-from jax.core import Primitive
-from jax.interpreters import ad, partial_eval, xla
-from jax.numpy.lax_numpy import _promote_args_like, _promote_shapes
+from jax.interpreters import ad
+from jax.numpy.lax_numpy import _promote_args_like
 from jax.util import partial
 
 
@@ -168,21 +167,12 @@ def _standard_gamma_grad(sample, alpha):
     return grads.reshape(alpha.shape)
 
 
-def _standard_gamma_abstract_eval(key, alpha, jaxpr, aval, consts):
-    return lax.maybe_tracer_tuple_to_abstract_tuple(aval)
+@custom_transforms
+def standard_gamma_p(key, alpha):
+    return _standard_gamma_impl(key, alpha)
 
 
-def _standard_gamma_translate(c, key, alpha, jaxpr, aval, consts):
-    xla_computation = xla.jaxpr_computation(jaxpr, consts, (), c.GetShape(key), c.GetShape(alpha))
-    return c.Call(xla_computation, (key, alpha))
-
-
-# define primitive
-standard_gamma_p = Primitive('standard_gamma')
-standard_gamma_p.def_impl(partial(xla.apply_primitive, standard_gamma_p))
-standard_gamma_p.def_abstract_eval(_standard_gamma_abstract_eval)
-xla.translations[standard_gamma_p] = _standard_gamma_translate
-ad.defjvp2(standard_gamma_p, None,
+ad.defjvp2(standard_gamma_p.primitive, None,
            lambda tangent, sample, key, alpha, **kwargs: tangent * _standard_gamma_grad(sample, alpha))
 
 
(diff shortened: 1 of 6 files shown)


numpyro/distributions/util.py
```diff
@@ -395,6 +282,103 @@ def multinomial_rvs(key, n, p, shape=()):
     return np.reshape(samples_2D, shape + p.shape[-1:]) - excess
 
 
+def multinomial(key, p, n, shape=()):
+    return _multinomial(key, p, n, shape)
+
+
+def _xlogy_jvp_lhs(g, x, y):
```
fehiepsi: I group all samplers to one place, so other utilities are grouped into one place too.

numpyro/distributions/util.py
```diff
@@ -361,23 +244,27 @@ def body_fn(val):
...
     return lax.scatter_add(operand, indices, updates,
                            lax.ScatterDimensionNumbers(update_window_dims=(),
                                                        inserted_window_dims=(0,),
                                                        scatter_dims_to_operand_dims=(0,)))
 
 
-@partial(jit, static_argnums=(1, 3))
-def multinomial_rvs(key, n, p, shape=()):
+@partial(jit, static_argnums=(2, 3))
+def _multinomial(key, p, n, shape=()):
```
fehiepsi: the pattern
```
@partial(jit, static_argnums=(2, 3))
def _f(a,b,c,d):
    ...

def f(a,b,c,d):
    return _f(a,b,c,d)
```
is used in `jax.random` to support different ways to set `args`, `kwargs` to samplers. Without this pattern, calling `f(a,b,c=c,d=d)` will trigger an error that `c` and `d` must be args, not kwargs.

numpyro/distributions/util.py
```diff
@@ -395,6 +282,103 @@ def multinomial_rvs(key, n, p, shape=()):
...
 
 
+def multinomial(key, p, n, shape=()):
+    return _multinomial(key, p, n, shape)
+
+
+def _xlogy_jvp_lhs(g, x, y):
+    shape = lax.broadcast_shapes(np.shape(g), np.shape(y))
+    g = np.broadcast_to(g, shape)
+    y = np.broadcast_to(y, shape)
```
fehiepsi: `lax._brcast` is not exposed (changed to `lax.lax._brcast`) in the newer version of JAX, so I use `np.broadcast_to` instead

numpyro/distributions/util.py
neerajprad: Hmm..what's the reason for this? Is this related to the change that kwargs are not treated as static args by default in jit?

test/test_distributions.py
```diff
@@ -242,6 +242,14 @@ def test_log_prob(jax_dist, sp_dist, params, prepend_shape, jit):
         expected = sp_dist.logpdf(samples)
     except AttributeError:
         expected = sp_dist.logpmf(samples)
+    except ValueError as e:
+        # precision issue: np.sum(x / np.sum(x)) = 0.99999994 != 1
```
neerajprad: Maybe we should lower this precision value in the transform or have it be dependent on the data, since I think scipy would do the normalization behind the scenes anyways.

numpyro/distributions/util.py
fehiepsi: Yes, kwargs are not allowed in the positions of static args in the newer version (you can see that commit [here](https://github.com/google/jax/pull/595)).

test/test_distributions.py
fehiepsi: This is indeed scipy problem. Scipy works in float64 so their check is `< 1e-10` would be fine for the output of `np.sum(x / np.sum(x))`. Our check `1e-6` in float32 should be fine but our samples would not be valid samples for scipy.

PR #134, opened 2019-05-03
Revise some unnecessary broadcasting

fehiepsi:
It seems to me (not 100% sure) that it is not necessary to apply broadcasting in mean/variance for the following distributions.

Some other non-important changes are:
+ Correct the order of distributions
+ replace `lax.full` by `np.full`

@neerajprad I notice that the order of `total_count` and `probs` has been swapped in discrete distributions. This order is different from pytorch/tensorflow/scipy/r. Just wonder if we should swap it back?

### DIFF
--- a/numpyro/contrib/distributions/distribution.py
+++ b/numpyro/contrib/distributions/distribution.py
@@ -47,6 +47,12 @@ class jax_frozen(rv_frozen):
     def __call__(self, size=None, random_state=None):
         return self.rvs(size, random_state)
 
+    def log_prob(self, x):
+        if isinstance(self.dist, jax_continuous):
+            return self.logpdf(x)
+        else:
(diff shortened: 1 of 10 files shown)


neerajprad:
@neerajprad I notice that the order of total_count and probs has been swapped in discrete distributions. This order is different from pytorch/tensorflow/scipy/r. Just wonder if we should swap it back?

The reason for that is that `logits/probs` are not keyword arguments since they are now separated but `total_count=1` is a keyword argument which must be specified last. That said, I'm making a change to the distribution classes so that this separation is only on the backend, and the external wrapper takes in either logits/probs and delegates to the right implementation. I'll ensure then that this wrapper matches the interface in PyTorch.

fehiepsi:
Thanks for explaining @neerajprad !

PR #135, opened 2019-05-05
Add LKJ distribution and UCBadmit example

### LINKED ISSUE #122: Support multivariate normal distribution
These distributions will be useful for a [multilevel model](https://github.com/rmcelreath/rethinking/tree/Experimental#multilevel-model-formulas) which I am intending to make an example as hierarchy model.

fehiepsi:
This PR adds LKJ distribution (resolves #122) and a small example for it. The example requires HalfNormal distribution so I also add it in this PR.

Because we use non-centered parameterization technique, MVN is not necessary for this example.


numpyro/distributions/constraints.py
```diff
@@ -236,25 +236,6 @@ def log_abs_det_jacobian(self, x, y):
...
-    z = r ** 2
-    z1m_cumprod = cumprod(1 - z)
-    z1m_cumprod_sqrt = np.sqrt(z1m_cumprod)
-
-    pad_width = [(0, 0)] * z.ndim
-    pad_width[-1] = (1, 0)
-    z1m_cumprod_sqrt_shifted = np.pad(z1m_cumprod_sqrt[..., :-1], pad_width,
-                                      mode="constant", constant_values=1.)
-    y = (r + np.identity(r.shape[-1])) * z1m_cumprod_sqrt_shifted
-    return y
```
fehiepsi: this function is required in LKJ distribution so I move it to util

numpyro/distributions/constraints.py
```diff
@@ -282,9 +263,8 @@ class :class:`StickBreakingTransform` to transform :math:`X_i` into a
 
     def __call__(self, x):
         # we interchange step 1 and step 2.a for a better performance
-        eps = np.finfo(x.dtype).eps
-        t = np.clip(np.tanh(x), a_min=(-1 + eps), a_max=(1 - eps))
-        return _signed_stick_breaking_tril(t)
+        t = np.tanh(x)
+        return signed_stick_breaking_tril(t)
```
fehiepsi: I also move the `clip` logic to `signed_stick_breaking_tril` function because we also need to clip in the sampler.

numpyro/distributions/continuous.py
```diff
@@ -299,14 +497,17 @@ class StudentT(Distribution):
...
         self._chi2 = Chi2(self.df)
-        batch_shape = lax.broadcast_shapes(np.shape(self.df), np.shape(self.loc), np.shape(self.scale))
         super(StudentT, self).__init__(batch_shape, validate_args=validate_args)
 
     def sample(self, key, size=()):
-        std_normal = random.normal(key, shape=size + self.batch_shape)
-        z = self._chi2.sample(key, size)
+        key_normal, key_chi2 = random.split(key)
+        std_normal = random.normal(key_normal, shape=size + self.batch_shape)
+        z = self._chi2.sample(key_chi2, size)
```
fehiepsi: some bugs in StudentT distribution, where we didn't split the key and broadcasted `self.df`

numpyro/examples/baseball.py
```diff
@@ -195,8 +194,8 @@ def main(args):
                                partially_pooled,
                                partially_pooled_with_logit,
                                )):
-        rng, rng_predict = random.split(random.PRNGKey(i))
-        zs = run_inference(model, at_bats, hits, random.PRNGKey(i), args)
+        rng, rng_predict = random.split(random.PRNGKey(i + 1))
```
fehiepsi: I want to make sure that this PRNGKey is different from the one is used in running hmc_init

numpyro/examples/datasets.py
```diff
@@ -27,7 +27,11 @@
 ])
 
 BASEBALL = dset('baseball', [
-    "https://d2fefpcigoriu7.cloudfront.net/datasets/EfronMorrisBB.txt",
+    'https://d2fefpcigoriu7.cloudfront.net/datasets/EfronMorrisBB.txt',
+])
+
+UCBADMIT = dset('ucbadmit', [
+    'https://raw.githubusercontent.com/rmcelreath/rethinking/master/data/UCBadmit.csv',
```
fehiepsi: FIXME: should we upload the data to `cloudfront.net`?

neerajprad: @fehiepsi - I think that will be better. Let me upload that, and update the link here.

neerajprad: You should be able to fetch this from: `https://d2fefpcigoriu7.cloudfront.net/datasets/UCBadmit.csv`. Its such a tiny dataset that it doesn't really matter though. :)

neerajprad:
@fehiepsi - I want to take a more detailed look at LKJ and the example, after the #136 is merged.

neerajprad:
Could you merge with master, so as to remove the changes that are already in?

numpyro/examples/datasets.py
fehiepsi: Thanks @neerajprad !

fehiepsi:
Just done it! :)

numpyro/distributions/continuous.py
```diff
@@ -217,6 +224,197 @@ def variance(self):
...
+        # On the other hand, signed stick breaking process gives:
+        #   l_21 = p_21, l_31 = p_31, l_22 = sqrt(1 - p_21^2), l_32 = p_32 * sqrt(1 - p_31^2)
+        #   r_32 = l_21 * l_31 + l_22 * l_32
+        #        = p_21 * p_31 + p_32 * sqrt{(1 - p_21^2)*(1 - p_31^2)} = I
+        beta_sample = self._beta.sample(key, size)
+        partial_correlation = 2 * beta_sample - 1  # scale to domain to (-1, 1)
+        return signed_stick_breaking_tril(partial_correlation)
+
+    def _onion(self, key, size):
+        key_beta, key_normal = random.split(key)
```
neerajprad: This is nicely vectorized.

numpyro/distributions/continuous.py
```diff
@@ -217,6 +224,197 @@ def variance(self):
...
+    variables are not correlated.
+
+    When ``concentration < 1``, the distribution favors samples with small diagonal entries
+    (hence small determinent). This is useful when we know a priori that some underlying
+    variables are correlated.
+
+    :param int dimension: dimension of the matrices
+    :param ndarray concentration: concentration/shape parameter of the
+        distribution (often referred to as eta)
+    :param str sample_method: Either "cvine" or "onion". Both methods are proposed in [1] and
```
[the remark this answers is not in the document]
fehiepsi: "onion" method is much faster (as reported in the reference). The following benchmark is to generate 8000 80x80 matrices

C-compiled with full optimization (reported in Lewandowski et.'s paper):
+ C-vine: 13.4s
+ Onion: 4.33s

Matlab 2007b (reported in Lewandowski et.'s paper):

+ C-vine: 146.4s
+ Onion: 82.4s

Pyro
+ C-vine: 4.39 s

This PR
+ C-vine: 323 ms
+ Onion: 24.1 ms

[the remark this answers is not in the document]
fehiepsi: But "cvine" is easier to implement ^^ (LKJ can be implemented as a TransformedDistribution when using "cvine" method)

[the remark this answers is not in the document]
fehiepsi: On the other, they have the same performance for small dimension (but "onion" is also a little bit faster than "cvine")

fehiepsi:
Thanks for reviewing @neerajprad !

numpyro/distributions/continuous.py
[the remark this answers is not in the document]
neerajprad: > Onion: 24.1 ms

That's quite impressive! I'm guessing that the big speedup from Pyro is also due to vectorization. With regard to the paper, I think its a bit old, but a ~200 speedup is more than what you would get from just hardware improvements.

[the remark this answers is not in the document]
fehiepsi: Ah, sorry. Just recognize that it is just performance in GPU, where we get a lot of benefit from vectorization. In CPU, the performance is
```
Onion: 5.91 s
Cvine: 9.01 s
```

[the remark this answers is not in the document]
neerajprad: I see..I think this includes the compilation of the gamma sampler unfortunately. That's also the reason why `test_distributions` is slow - many test runs are dominated by the time to compile the gamma sampler for different shapes.

PR #136, opened 2019-05-06
Update to newer jax version

fehiepsi:
This PR makes our repo work for newer jax version (which has just been released). One of the new feature of jax is the unstable of fast-math has been resolved. So I guess it is safe to enable fastmath again in Travis now.

For some reasons, statistics mean_var test for StudentT fails (rtol is a bit higher than before) with new version so I changed parameters of that test to make tests pass.

`lax.scan` is removed in the newer version. In addition, I set `progbar=True` to make it consistence with `init_kernel`.

`optimizers` now returns 3 functions: init, update, and get_params (instead of `optimizers.get_params`) as previously.

### DIFF
--- a/numpyro/distributions/continuous.py
+++ b/numpyro/distributions/continuous.py
@@ -299,14 +299,18 @@ class StudentT(Distribution):
     reparametrized_params = ['loc', 'scale']
 
     def __init__(self, df, loc=0., scale=1., validate_args=None):
-        self.df, self.loc, self.scale = promote_shapes(df, loc, scale)
+        batch_shape = lax.broadcast_shapes(np.shape(df), np.shape(loc), np.shape(scale))
+        self.df = np.broadcast_to(df, batch_shape)
+        self.loc = np.broadcast_to(loc, batch_shape)
+        self.scale = np.broadcast_to(scale, batch_shape)
         self._chi2 = Chi2(self.df)
         batch_shape = lax.broadcast_shapes(np.shape(self.df), np.shape(self.loc), np.shape(self.scale))
         super(StudentT, self).__init__(batch_shape, validate_args=validate_args)
 
     def sample(self, key, size=()):
-        std_normal = random.normal(key, shape=size + self.batch_shape)
-        z = self._chi2.sample(key, size)
+        key_normal, key_chi2 = random.split(key)
+        std_normal = random.normal(key_normal, shape=size + self.batch_shape)
+        z = self._chi2.sample(key_chi2, size)
         y = std_normal * np.sqrt(self.df / z)
         return self.loc + self.scale * y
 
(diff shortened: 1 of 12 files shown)


numpyro/distributions/continuous.py
```diff
@@ -299,14 +299,18 @@ class StudentT(Distribution):
     reparametrized_params = ['loc', 'scale']
 
     def __init__(self, df, loc=0., scale=1., validate_args=None):
-        self.df, self.loc, self.scale = promote_shapes(df, loc, scale)
+        batch_shape = lax.broadcast_shapes(np.shape(df), np.shape(loc), np.shape(scale))
+        self.df = np.broadcast_to(df, batch_shape)
```
neerajprad: Do we need to do this explicit `broadcast_to`?

numpyro/distributions/continuous.py
```diff
@@ -299,14 +299,18 @@ class StudentT(Distribution):
...
+        self.scale = np.broadcast_to(scale, batch_shape)
         self._chi2 = Chi2(self.df)
         batch_shape = lax.broadcast_shapes(np.shape(self.df), np.shape(self.loc), np.shape(self.scale))
         super(StudentT, self).__init__(batch_shape, validate_args=validate_args)
 
     def sample(self, key, size=()):
-        std_normal = random.normal(key, shape=size + self.batch_shape)
-        z = self._chi2.sample(key, size)
+        key_normal, key_chi2 = random.split(key)
+        std_normal = random.normal(key_normal, shape=size + self.batch_shape)
```
neerajprad: Thanks for catching this, this is bad!

neerajprad:
> lax.scan is removed in the newer version.

I thought there was a plan to have a differentiable `lax.scan` in this version? Is that scrapped?

numpyro/distributions/continuous.py
fehiepsi: yes, it is required to have valid `chi2` samples (otherwise, the `chi2` samples will have a smaller shape than StudentT batch_shape

fehiepsi:
Yes, sadly it is not there yet >"<

numpyro/distributions/continuous.py
neerajprad: I see. I was wondering if we can `broadcast_to` only for `df` but use `promote_shapes` for `loc`, `scale`, but let us profile that separately and make a change, if needed.

neerajprad:
> Yes, sadly it is not there yet >"<

I was hoping your time series models would be unblocked now, but doesn't look like it.

numpyro/distributions/continuous.py
fehiepsi: yup, we don't need to broadcasting `loc`, `scale`. I'll fix that in LKJ pr. :)

PR #138, opened 2019-05-08
Add sphinx and rtd support for Numpyro

neerajprad:
This sets up the sphinx and readthedocs config for Numpyro. 

I have also tried to add documentation for the `mcmc` module to see if the docs are getting generated properly. One restriction is that closures aren't handled by sphinx. e.g. for the `hmc` function, the documentation for `init_kernel` and `sample_kernel` isn't generated. 

@jpchen - Let me know if anything else should be needed. 

TODOs after this PR is merged:
 - Finding work-arounds to make sphinx work with closures - easiest tweak would be to change these to classes, for namesake.
 - Adding documentation for SVI, distributions, and handlers.


numpyro/mcmc.py
```diff
@@ -40,6 +40,25 @@ def _euclidean_ke(inverse_mass_matrix, r):
...
+
+    [1] `MCMC Using Hamiltonian Dynamics`,
+    Radford M. Neal
+    [2] `The No-U-turn sampler: adaptively setting path lengths in Hamiltonian Monte Carlo`,
+    Matthew D. Hoffman, and Andrew Gelman.
+
+    :param potential_fn: Python callable that computes the potential energy
+        given input parameters. The input parameters to `potential_fn` can be
+        any python collection type, provided that ``init_samples`` argument to
+        ``init_kernel`` has the same type.
```
fehiepsi: The docs for `kinetic_fn`, `algo` are missing.

fehiepsi:
I think this is good to merge?

neerajprad:
> I think this is good to merge?

Yes, let us merge this. If we hit any blockers when putting this on readthedocs, we can bother @jpchen.

neerajprad:
> I guess we can add some simple signature to the docs of hmc and let users see details when they generate local functions in their IDE.

Good idea, I think the docs should appear fine in notebooks / IDE, but it would be nice if it was also a part of the official documentation online since `init_kernel`, `sample_kernel` are actually the more important pieces here.

docs/source/conf.py
```diff
@@ -0,0 +1,192 @@
...
+# is a PyTorch class.
+
+autodoc_inherit_docstrings = False
+
+autodoc_default_options = {
+    'member-order': 'bysource',
+    'show-inheritance': True,
+    'special-members': True,
+    'undoc-members': True,
+    'exclude-members': '__dict__,__module__,__weakref__',
```
jpchen: sorry, my email notifications were messed up. rtd has an outdated version of sphinx so you'll need to comment this out until they update.

PR #140, opened 2019-05-09
Modify HMC documentation to show docstrings for internal functions

neerajprad:
This makes a change to the documentation so that `init_kernel` and `sample_kernel` are documented within the docstring for `hmc` and can therefore be picked up by sphinx. To prevent having to rewrite the same docstring again for the enclosed functions (which would be useful in IDEs and in the notebook environment), the relevant parts from the enclosing function's docstring is taken to set the `__doc__` attribute of these functions. 

Addresses #29


fehiepsi:
Hah, nice trick!

PR #141, opened 2019-05-09
Unifying wrappers for distributions with logits/probs parametrization

neerajprad:
Based on @fritzo's suggestion in https://github.com/pyro-ppl/funsor/pull/135, this creates light wrappers for all distributions with logits / probs parametrization. The advantage of this is that the underlying implementations are separate which makes them easy to test, simplifies code and removes any risk of accessing lazy parameters and triggering unnecessary computation. At the same time, the wrapper gives an identical interface to `torch.distributions`. 

Testing: The wrappers are exercised in examples, the tests remain the same except for the renaming.


fehiepsi:
So this is what you have in mind for your previous comment. Got it now! :)

PR #143, opened 2019-05-10
 Add stochastic volatility model to examples

neerajprad:
This adds a stochastic volatility model taken from PyMC3 and originally discussed in this [forum post](https://forum.pyro.ai/t/problems-transforming-a-pymc3-model-to-pyro-mcmc/208/14). This also adds the `GaussianRandomWalk` distribution from @fehiepsi's post on the topic. I was interested in translating this, since this was particularly slow and problematic on Pyro, but this example now seems to work quite well, and is quite fast even for ~2400 datapoints.

Side note: ~Also updated jax dependency.~ ~Update: Reverted this, since this requires updating a few more tests.~ Pinning to `jax==0.1.27` which seems to be stable w.r.t. master. Moving to `0.1.28` leads to some unrelated failures in `test_distributions.py` that we should be separately resolved.

TODO:
 - [x] Add distribution tests.
 - [x] Fix distribution shape.


fehiepsi:
Really happy to see this implemented! Just a quick note that you can use `util.cumsum` to be able to take grad (or it might work now with the new version of JAX). I'll review in details later.

numpyro/examples/stochastic_volatility.py
```diff
@@ -0,0 +1,87 @@
...
+                              progbar=True)
+    print_results(hmc_states, dates)
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="Stochastic Volatility Model")
+    parser.add_argument('-n', '--num-samples', nargs='?', default=3000, type=int)
+    parser.add_argument('--num-warmup-steps', nargs='?', default=1500, type=int)
+    parser.add_argument('--device', default='cpu', type=str, help='use "cpu" or "gpu".')
+    parser.add_argument('--rng', default=0, type=int, help='random number generator seed')
```
fehiepsi: Can we use default `rng` different from the one in HMC init_kernel?

numpyro/examples/stochastic_volatility.py
```diff
@@ -0,0 +1,87 @@
...
+    jax_config.update('jax_platform_name', args.device)
+    _, fetch = load_dataset(SP500, shuffle=False)
+    dates, returns = fetch()
+    rng = random.PRNGKey(args.rng)
+    init_params, potential_fn, transform_fn = initialize_model(rng, model, (returns,), {})
+    init_kernel, sample_kernel = hmc(potential_fn, algo='NUTS')
+    hmc_state = init_kernel(init_params, args.num_warmup_steps)
+    hmc_states = fori_collect(args.num_samples, sample_kernel, hmc_state,
+                              transform=lambda hmc_state: transform_fn(hmc_state.z),
+                              progbar=True)
```
fehiepsi: nit: the `progbar` flag is not necessary now

numpyro/distributions/continuous.py
```diff
@@ -491,6 +492,42 @@ def support(self):
         return constraints.greater_than(self.scale)
 
 
+class GaussianRandomWalk(Distribution):
+    arg_constraints = {'num_steps': constraints.positive, 'scale': constraints.positive}
+    support = constraints.real
+    # FIXME: cannot take grad through random.normal with dynamic shape
```
fehiepsi: Opps, I forgot to ask about this. Could you pls elaborate the problem so I can take a look?

fehiepsi: I guess you mean `grad` w.r.t. num_steps?

neerajprad: That's correct. The reason why this isn't reparametrizable w.r.t. scale is because cumsum and (random.normal)'s shape depends on `num_steps` and we aren't able to take gradient through that. While I have marked it `FIXME`, this is only important if we want to use this for SVI.

PR #145, opened 2019-05-11
Respect unnormalized logits input in Categorical/Multinomial dist

fehiepsi:
Currently, logits of these distributions are normalized. The changes in this PR have the following folds
+ Removes the unexpected behaviour for users when `logits` input values are modified at __init__ 
+ Removes the unnecessary logsumexp op when only sampling is necessary
+ Matches with `tensorflow` definition of `logits` ([ref](https://www.tensorflow.org/probability/api_docs/python/tfp/distributions/Multinomial#args))
+ (personally) allow me inject a custom log_prob term to joint density (by setting value=0)

This PR also fixes the bug of missing `axis=-1` for `logsumexp` in previous PRs.

### DIFF
--- a/numpyro/contrib/distributions/multivariate.py
+++ b/numpyro/contrib/distributions/multivariate.py
@@ -10,7 +10,7 @@ import jax.numpy as np
 from jax import lax
 from jax.experimental.stax import softmax
 from jax.numpy.lax_numpy import _promote_dtypes
-from jax.scipy.special import digamma, gammaln
+from jax.scipy.special import digamma, gammaln, logsumexp
 
 from numpyro.contrib.distributions.discrete import binom
 from numpyro.contrib.distributions.distribution import jax_continuous, jax_discrete, jax_multivariate
@@ -54,6 +54,8 @@ class categorical_gen(jax_multivariate, jax_discrete):
         x = np.broadcast_to(x, batch_shape + (1,))
         p = np.broadcast_to(p, batch_shape + p.shape[-1:])
         if self.is_logits:
+            # normalize log prob
+            p = p - logsumexp(p, axis=-1, keepdims=True)
             # gather and remove the trailing dimension
             return np.take_along_axis(p, x, axis=-1)[..., 0]
         else:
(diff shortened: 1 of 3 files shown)


fritzo:
Interesting, could we have a deeper design discussion?

I don't fully understand this PR, but I do like the general idea of allowing the partition function `Z` to pass through distribution methods (by avoiding normalization). In Funsor we've made the decision to keep low-level funsors non-normalized but to keep the minipyro distributions interface normalized. Is that similar to what you're doing in this PR, to keep low-level functions non-normalized while ensuring `dist.Categorical(logits=logits).log_prob(-)` is indeed normalized?

fehiepsi:
When we have a `custom_log_prob`, with `total_count=1`, `value=0`, the normalize term `total_count * logsumexp(self.logits, axis=-1) - gammaln(total_count + 1)` will be `custom_log_prob` and the remaining term `np.sum(value * self.logits - gammaln(value + 1), axis=-1)` will be `0`. So the Multinomial log_prob will be `custom_log_prob`.

The BernoulliLogits only works for the case `custom_log_prob << 0`, where `binary_entropy_with_logits(logits, 1.) ~ logits`. Bernoulli trick does not work when `custom_log_prob ~0` or completely wrong when `custom_log_prob > 0` (in case we calculate unnormalized log_prob).

fehiepsi:
> Is that similar to what you're doing in this PR, to keep low-level functions non-normalized while ensuring dist.Categorical(logits=logits).log_prob(-) is indeed normalized?

I guess it is a bit different. The distributions `Categorical` and `Multinomial` are normalized. But we can inject an unnormalized term (calculated outside these distributions) by using the above trick. By using the **invalid** sample `value=0` (not belongs to the support), the custom log prob term will not be normalized. I think that the logic of `funsor` is correct, while here it is just a trick. ^^!

neerajprad:
@fritzo - This is mostly just a cleanup/bugfix PR. This allows for an additional hack, IIUC from what @fehiepsi mentioned, of injecting some custom term into the log joint computation by using `Multinomial(total_count=1, logits=-custom_term).log_prob(value=0) --> custom_term`. I'm calling it a hack because it will throw an error with validation check enabled. :) I think @fehiepsi has some models where this trick be useful, and it will be interesting to see if we can use funsors as the backend for such models.

neerajprad:
Merging this since this is just a bugfix PR, but we should have a separate design discussion on "allowing the partition function Z to pass through distribution methods (by avoiding normalization)" which will be useful for models that can exploit conjugacy, for instance.

fehiepsi:
> using Multinomial(total_count=1, logits=-custom_term).log_prob(value=0) --> custom_term

Opps... thanks for correcting me (about the sign of the custom term).

PR #147, opened 2019-05-12
Update to newer version of JAX

### LINKED ISSUE #146: binomial/multinomial tests fail with the new version of jax
Start with jax version 0.1.28, as @neerajprad observed, tests for Binomial/Multinomial failed. I am not sure what is the reason for it because except for the tests, these distributions work well in notebooks.

fehiepsi:
Fixes #146

Originally, this PR is to implement TruncatedCauchy (to port local global trend to new distribution API) but it seems that we can fix #146 by casting device array to onp.array. Given that it is more urgent (to make our tests faster), I push the fix instead.

In addition, I also add default args to our distribution interface.

### DIFF
--- a/numpyro/distributions/continuous.py
+++ b/numpyro/distributions/continuous.py
@@ -77,7 +77,7 @@ class Cauchy(Distribution):
     support = constraints.real
     reparametrized_params = ['loc', 'scale']
 
-    def __init__(self, loc, scale, validate_args=None):
+    def __init__(self, loc=0., scale=1., validate_args=None):
         self.loc, self.scale = promote_shapes(loc, scale)
         batch_shape = lax.broadcast_shapes(np.shape(loc), np.shape(scale))
         super(Cauchy, self).__init__(batch_shape=batch_shape, validate_args=validate_args)
@@ -140,7 +140,7 @@ class Exponential(Distribution):
     arg_constraints = {'rate': constraints.positive}
     support = constraints.positive
 
-    def __init__(self, rate, validate_args=None):
+    def __init__(self, rate=1., validate_args=None):
         self.rate = rate
         super(Exponential, self).__init__(batch_shape=np.shape(rate), validate_args=validate_args)
 
@@ -166,7 +166,7 @@ class Gamma(Distribution):
                        'rate': constraints.positive}
     support = constraints.positive
 
-    def __init__(self, concentration, rate, validate_args=None):
+    def __init__(self, concentration, rate=1., validate_args=None):
         self.concentration, self.rate = promote_shapes(concentration, rate)
         batch_shape = lax.broadcast_shapes(np.shape(concentration), np.shape(rate))
         super(Gamma, self).__init__(batch_shape=batch_shape,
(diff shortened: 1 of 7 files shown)


test/test_distributions.py
```diff
@@ -216,27 +216,30 @@ def test_sample_gradient(jax_dist, sp_dist, params):
         pytest.skip('{} not reparametrized.'.format(jax_dist.__name__))
 
     dist_args = [p.name for p in inspect.signature(jax_dist).parameters.values()]
+    params_dict = dict(zip(dist_args[:len(params)], params))
+    nonrepara_params_dict = {k: v for k, v in params_dict.items()
+                             if k not in jax_dist.reparametrized_params}
+    repara_params = tuple(v for k, v in params_dict.items()
+                          if k in jax_dist.reparametrized_params)
```
fehiepsi: This is to only take `grad` w.r.t. reparameterized params.

numpyro/distributions/continuous.py
```diff
@@ -468,6 +503,8 @@ class Pareto(TransformedDistribution):
     arg_constraints = {'alpha': constraints.positive, 'scale': constraints.positive}
     support = constraints.real
 
+    # FIXME: should we use `concentration` and `scale` for consistence with other distributions
+    # tensorflow uses `concentration` and `scale` though
```
fehiepsi: I think it is better if we can swap `alpha` and `scale` (which makes Pareto closer to scipy/tensorflow than pytorch). There we can also set default parameter `scale=1`.

numpyro/distributions/util.py
```diff
@@ -206,6 +206,7 @@ def _binomial(key, p, n=1, shape=()):
 
 
 def binomial(key, p, n=1, shape=()):
+    n = device_get(n)
```
fehiepsi: I think that `static_args` does not like DeviceArray anymore. Anyway this is no-op if n is numpy array so I guess things should be fine.

numpyro/distributions/continuous.py
neerajprad: Seems reasonable to do.

numpyro/distributions/util.py
neerajprad: Hmm..that's quite strange. Is this intended?

numpyro/distributions/continuous.py
```diff
@@ -200,13 +200,48 @@ def __init__(self, df, validate_args=None):
         super(Chi2, self).__init__(0.5 * df, 0.5, validate_args=validate_args)
 
 
+class GaussianRandomWalk(Distribution):
+    arg_constraints = {'num_steps': constraints.positive_integer, 'scale': constraints.positive}
+    support = constraints.real
+    reparametrized_params = ['scale']
```
neerajprad: Nice to see that this is working!

test/test_distributions.py
```diff
@@ -216,27 +216,30 @@ def test_sample_gradient(jax_dist, sp_dist, params):
...
     rng = random.PRNGKey(0)
 
     def fn(args):
-        return np.sum(jax_dist(*args).sample(key=rng))
+        args_dict = dict(zip(jax_dist.reparametrized_params, args))
+        return np.sum(jax_dist(**args_dict, **nonrepara_params_dict).sample(key=rng))
 
-    actual_grad = jax.grad(fn)(params)
-    assert len(actual_grad) == len(params)
+    actual_grad = jax.grad(fn)(repara_params)
```
neerajprad: +1.

numpyro/distributions/util.py
fehiepsi: I am not sure about that. We can file an issue to JAX devs if this makes regression in performance.

PR #150, opened 2019-05-13
Porting TruncatedNormal, TruncatedCauchy to new api

fehiepsi:
This PR adds TruncatedNormal, TruncatedCauchy distributions (these dists are popular to for strictly positive support). Currently, only `low` arg is support. `high` (upper bound) arg will be supported per user request.

@neerajprad After this, I don't need contrib scipy API any more. Please let me know if we should remove it so we can remove before alpha release. :)

### DIFF
--- a/test/test_distributions.py
+++ b/test/test_distributions.py
@@ -216,27 +216,30 @@ def test_sample_gradient(jax_dist, sp_dist, params):
         pytest.skip('{} not reparametrized.'.format(jax_dist.__name__))
 
     dist_args = [p.name for p in inspect.signature(jax_dist).parameters.values()]
+    params_dict = dict(zip(dist_args[:len(params)], params))
+    nonrepara_params_dict = {k: v for k, v in params_dict.items()
+                             if k not in jax_dist.reparametrized_params}
+    repara_params = tuple(v for k, v in params_dict.items()
+                          if k in jax_dist.reparametrized_params)
 
     rng = random.PRNGKey(0)
 
     def fn(args):
-        return np.sum(jax_dist(*args).sample(key=rng))
+        args_dict = dict(zip(jax_dist.reparametrized_params, args))
+        return np.sum(jax_dist(**args_dict, **nonrepara_params_dict).sample(key=rng))
 
-    actual_grad = jax.grad(fn)(params)
-    assert len(actual_grad) == len(params)
+    actual_grad = jax.grad(fn)(repara_params)
+    assert len(actual_grad) == len(repara_params)
 
     eps = 1e-5
-    for i in range(len(params)):
-        if np.result_type(params[i]) in (np.int32, np.int64) or \
-                dist_args[i] not in jax_dist.reparametrized_params:
-            continue
+    for i in range(len(repara_params)):
         args_lhs = [p if j != i else p - eps for j, p in enumerate(params)]
         args_rhs = [p if j != i else p + eps for j, p in enumerate(params)]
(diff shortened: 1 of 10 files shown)


test/test_distributions.py
```diff
@@ -231,7 +237,7 @@ def fn(args):
     actual_grad = jax.grad(fn)(repara_params)
     assert len(actual_grad) == len(repara_params)
 
-    eps = 1e-5
+    eps = 1e-3
```
fehiepsi: 1e-5 is so sensitive in float32. So I change to 1e-3 instead (follows [wikipedia](https://en.wikipedia.org/wiki/Numerical_differentiation))

test/test_distributions.py
```diff
@@ -100,6 +100,12 @@ def __new__(cls, jax_dist, *params):
     T(dist.StudentT, 1., 1., 0.5),
     T(dist.StudentT, 2., np.array([1., 2.]), 2.),
     T(dist.StudentT, np.array([3, 5]), np.array([[1.], [2.]]), 2.),
+    T(dist.TruncatedCauchy, -1., 0., 1.),
```
neerajprad: Since these distributions have corresponding scipy implementations available, can we add these to `_DIST_MAP` above? That will increase the test coverage for these distributions.

test/test_distributions.py
```diff
@@ -257,6 +263,14 @@ def test_log_prob(jax_dist, sp_dist, params, prepend_shape, jit):
     samples = jax_dist.sample(key=rng, size=prepend_shape)
     assert jax_dist.log_prob(samples).shape == prepend_shape + jax_dist.batch_shape
     if not sp_dist:
+        if isinstance(jax_dist, dist.TruncatedCauchy) or isinstance(jax_dist, dist.TruncatedNormal):
```
neerajprad: If we add this to `_DIST_MAP`, we shouldn't have to add this specially.

test/test_distributions.py
neerajprad: We are using 1e-3 for log_prob too.

neerajprad:
>  After this, I don't need contrib scipy API any more. Please let me know if we should remove it so we can remove before alpha release. :)

Thanks for adding these distributions! I would leave that to you. Depending on how useful you think this might be in the future, we could also just keep the wrappers around (maybe remove the ported distributions), in case there are other scipy distributions that can be quickly ported over using this wrapper.

test/test_distributions.py
fehiepsi: Unfortunately, there is no trunccauchy in scipy. In addition, scipy truncnorm does not support ndarray `low` arg. If I add `truncnorm`
```
sp_dist = osp_stats.truncnorm(a=onp.zeros(2), b=onp.ones(2), loc=onp.zeros(2), scale=onp.ones(2))
x = sp_dist.rvs(size=(1000, 2))
sp_dist.logpdf(x)  # scipy bug: ValueError
```
If we add `truncnorm` to DIST_MAP, then we can only test for scalars `low`, `loc`, and `scale` (because scipy `a` is `(low - loc) / scale`). I think that there is just a missing point here: missing the mean_var test w.r.t. scipy truncnorm implementation (with scalar parameters). We can add a separate test for it instead if you want.

neerajprad: I see, I didn't realize that `truncnorm` wasn't vectorized w.r.t. `low`, and there is no truncated cauchy at all. it is probably not worth adding a separate test for that.

PR #151, opened 2019-05-13
Test examples in travis CI

neerajprad:
Adds `test_example` to ensure that our examples are not outdated due to changes in the code.

Other minor clean-ups:
 - Changes initialize_model to take in `*args, **kwargs`.
 - Moved examples to the outermost directory, so that we can have separate dependency requirements for examples just like notebooks.


.gitignore
```diff
@@ -14,7 +14,7 @@ __pycache__/
 
 # data files
 numpyro/examples/.data
```
fehiepsi: nit: move this .data folder too

neerajprad: @fehiepsi - Note that all example utilities like `datasets.py` will still  be in `numpyro/examples` and the location of the `.data` will therefore not change. This is to make it easy for all examples to reuse these utilities (the `examples` folder isn't a python module), and also to share datasets.

fehiepsi: Ah, got it! Thank you for clarifying!

fehiepsi: Ah, got it! Thank you for clarifying!

fehiepsi:
@neerajprad feel free to merge this when tests pass. It is looking like some library is missing for running covtype and vae.

PR #153, opened 2019-05-15
Swap pareto alpha and scale position

fehiepsi:
This is to match the behaviour of scipy/tensorflow.


neerajprad:
> This is to match the behaviour of scipy/tensorflow.

I'm not sure why we decided to diverge in PyTorch, this seems more sensible.

Issue #154, opened 2019-05-16
NUTS doesn't converge on a stan model


neerajprad:
@stefanwebb pointed out that Pyro's NUTS on the [earnings latin square](https://github.com/stan-dev/example-models/blob/48dcab3e1b63c57948389602e9be469cb6702484/ARM/Ch.13/earnings_latin_square.stan) model gives extremely different results from Stan. cc. @jpchen 

To debug this, I have tried running the model on Pyro's NUTS and Numpyro's NUTS and both return results which are very far off from Stan with high r_hat values indicating that the procedure hasn't converged. Creating this issue to track progress on investigating this bug / discrepancy. 

Some notes:
 - It is possible that my translation of the model to Pyro/Numpyro is wrong. 
 - I have tried changing the default tensor type to Double in Pyro and changing the scale parameters to `dist.HalfCauchy(1.)` (instead of `dist.Uniform(0, 100)` which is more faithful to the Stan implementation) to see if that helps convergence. ~While it does seem to help somewhat, we still get very different results.~ This seems to help quite a lot on Numpyro (still checking on Pyro). 
 - Numpyro is much faster than Pyro (I think also faster than Stan), but seems to give incorrect results. Not surprising since the underlying issue, either in my code or in the inference algorithm, is likely the same for both the implementations.

Pyro code:
```python
import csv
from collections import defaultdict

import torch

import pyro
import pyro.distributions as dist
from pyro.infer.mcmc import NUTS, MCMC


torch.set_default_tensor_type('torch.DoubleTensor')
use_uniform = False


def scale():
    return dist.Uniform(0., 100.) if use_uniform else dist.HalfCauchy(1.)


def model(data):
    eth = data['eth']
    age = data['age']
    x = data['x']
    y = data['y']
    mu_a1 = pyro.sample('mu_a1', dist.Normal(0., 1.))
    mu_a2 = pyro.sample('mu_a2', dist.Normal(0., 1.))
    sigma_a1 = pyro.sample('sigma_a1', scale())
    sigma_a2 = pyro.sample('sigma_a2', scale())

    mu_b1 = pyro.sample('mu_b1', dist.Normal(0., 1.))
    mu_b2 = pyro.sample('mu_b2', dist.Normal(0., 1.))
    sigma_b1 = pyro.sample('sigma_b1', scale())
    sigma_b2 = pyro.sample('sigma_b2', scale())

    mu_c = pyro.sample('mu_c', dist.Normal(0., 1.))
[...]

PR #155, opened 2019-05-16
fix bug at constraint

fehiepsi:
Hopefully, this fixes the bug at #154. Looking like in PyTorch, AffineTransform does not work with various constraint domain, so the transforms of bounded parameters in Pyro will be wrong??? We should fix the issue in PyTorch.

### DIFF
--- a/numpyro/distributions/constraints.py
+++ b/numpyro/distributions/constraints.py
@@ -408,8 +408,9 @@ def _transform_to_greater_than(constraint):
 
 @biject_to.register(interval)
 def _transform_to_interval(constraint):
+    scale = constraint.upper_bound - constraint.lower_bound
     return ComposeTransform([SigmoidTransform(),
-                             AffineTransform(constraint.lower_bound, constraint.upper_bound,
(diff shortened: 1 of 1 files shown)


neerajprad:
Whoa..thanks for detecting this bug. I'm actually really surprised that we didn't notice this until now!

fehiepsi:
I guess no, I'll add test for it in the next PR.

neerajprad:
@fehiepsi - This seems to be correctly handled in [PyTorch](https://github.com/pytorch/pytorch/blob/master/torch/distributions/constraint_registry.py#L189). Do you think there is any other change that's needed?

fehiepsi:
I think that although codomain of ComposeTransform in PyTorch which involves AffineTransform is wrong, it just affects transformed distributions. So that issue does not affect your last model.

neerajprad:
I see the issue. Thanks for explaining, I'm looking into it! I'm not too concerned about that particular model, but more worried about using HMC / SVI with `TransformedDistribution`.

Issue #149, opened 2019-05-12
Make MCMC diagnostics available in Numpyro


neerajprad:
Right now we don't have any diagnostics to measure the quality of mixing. It will be nice to either implement diagnostics like we have done in Pyro, or perhaps just use the `arviz` library for this (preferable to offload this to arviz if possible).

fehiepsi:
I think we can have some diagnostics such as r_hat or n_eff to be able to use in examples and defer other diagnostics/information criterion to arviz.

fehiepsi:
Maybe I can take care of this issue because I have made a preliminary version `n_eff` last week.

PR #156, opened 2019-05-17
Implement summary stats/diagnostics

fehiepsi:
Addresses #149, this PR addes
+ effective sample size
+ split r hat
+ hpdi
+ a summary table
```
                           mean         sd       5.5%      94.5%      n_eff       Rhat
  emission_prob[0,0]       0.00       0.01       0.00       0.01     599.79       1.00
  emission_prob[0,1]       0.00       0.00       0.00       0.00     962.09       1.00
  emission_prob[0,2]       0.01       0.02       0.00       0.01     605.00       1.00
  emission_prob[0,3]       0.49       0.09       0.33       0.61     699.05       1.00
  emission_prob[0,4]       0.00       0.01       0.00       0.01     776.91       1.00
  emission_prob[0,5]       0.41       0.10       0.25       0.56     686.47       1.00
  emission_prob[0,6]       0.01       0.02       0.00       0.02     595.12       1.00
  emission_prob[0,7]       0.01       0.03       0.00 
```

The implementation is based on current Pyro implementations. Tests are compared against Pyro.

TODO:
- [x]  add tests

### DIFF
--- a/numpyro/diagnostics.py
+++ b/numpyro/diagnostics.py
@@ -161,16 +161,18 @@ def effective_sample_size(x):
     return n_eff
 
 
-def hpdi(x, prob=0.89):
+def hpdi(x, prob=0.89, axis=0):
     """
     Computes "highest posterior density interval" (HPDI) which is the narrowest
     interval with probability mass ``prob``.
 
     :param numpy.ndarray x: the input array.
     :param float prob: the probability mass of samples within the interval.
+    :param int axis: the dimension to calculate hpdi.
     :returns numpy.ndarray: quantiles of ``input`` at ``(1 - probs) / 2`` and
         ``(1 + probs) / 2``.
     """
+    x = onp.swapaxes(x, axis, 0)
     sorted_x = onp.sort(x, axis=0)
     mass = x.shape[0]
     index_length = int(prob * mass)
@@ -179,9 +181,11 @@ def hpdi(x, prob=0.89):
     intervals_length = intervals_right - intervals_left
     index_start = intervals_length.argmin(axis=0)
     index_end = index_start + index_length
-    hpd_left = onp.take_along_axis(sorted_x, index_start[None, ...], axis=0).squeeze(axis=0)
-    hpd_right = onp.take_along_axis(sorted_x, index_end[None, ...], axis=0).squeeze(axis=0)
-    return onp.stack([hpd_left, hpd_right])
+    hpd_left = onp.take_along_axis(sorted_x, index_start[None, ...], axis=0)
+    hpd_left = onp.swapaxes(hpd_left, axis, 0)
+    hpd_right = onp.take_along_axis(sorted_x, index_end[None, ...], axis=0)
+    hpd_right = onp.swapaxes(hpd_right, axis, 0)
+    return onp.concatenate([hpd_left, hpd_right], axis=axis)
 
 
 def summary(samples, prob=0.89):
--- a/numpyro/diagnostics.py
+++ b/numpyro/diagnostics.py
@@ -1,0 +1,216 @@
+from itertools import product
+
+import numpy as onp
+
+from jax import device_get
+
+
+def _compute_chain_variance_stats(x):
+    # compute within-chain variance and variance estimator
+    # input has shape C x N x sample_shape
+    C, N = x.shape[:2]
+    chain_var = x.var(axis=1) * N / (N - 1)
+    var_within = chain_var.mean(axis=0)
+    var_estimator = var_within * (N - 1) / N
+    if x.shape[0] > 1:
+        chain_mean = x.mean(axis=1)
+        var_between = chain_mean.var(axis=0) * C / (C - 1)
+        var_estimator = var_estimator + var_between
+    else:
+        var_within = var_estimator
+    return var_within, var_estimator
+
+
+def gelman_rubin(x):
+    """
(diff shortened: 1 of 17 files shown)


numpyro/distributions/constraints.py
```diff
@@ -361,8 +361,11 @@ def inv(self, y):
     def log_abs_det_jacobian(self, x, y):
         # Ref: https://mc-stan.org/docs/2_19/reference-manual/simplex-transform-section.html
         # |det|(J) = Product(y * (1 - z))
-        z = _clipped_expit(x - np.log(x.shape[-1] - np.arange(x.shape[-1])))
-        return np.sum(np.log(y[..., :-1] * (1 - z)), axis=-1)
+        z = x - np.log(x.shape[-1] - np.arange(x.shape[-1]))
+        z = np.clip(z, a_min=np.finfo(x.dtype).tiny)
+        # XXX we use the identity 1 - z = z * exp(-x) to not worry about
+        # the case z ~ 1
+        return np.sum(np.log(y[..., :-1]) + np.log(z) - x, axis=-1)
```
fehiepsi: a bit improvement in precision with the trade-off of additional log operator.

fehiepsi: Please ignore this! I'll revert this change when addresses reviews. There are some nice discussions about this in PyTorch.

numpyro/diagnostics.py
```diff
@@ -0,0 +1,217 @@
...
+    intervals_right = sorted_x[index_length:]
+    intervals_length = intervals_right - intervals_left
+    index_start = intervals_length.argmin(axis=0)
+    index_end = index_start + index_length
+    hpd_left = onp.take_along_axis(sorted_x, index_start[None, ...], axis=0).squeeze(axis=0)
+    hpd_right = onp.take_along_axis(sorted_x, index_end[None, ...], axis=0).squeeze(axis=0)
+    return onp.stack([hpd_left, hpd_right])
+
+
+def summary(samples, prob=0.89):
```
neerajprad: Nice! Can't wait to use this in some of our examples.

numpyro/diagnostics.py
```diff
@@ -0,0 +1,217 @@
...
+        if m == 1:
+            return target
+        target += 1
+
+
+def autocorrelation(x, axis=0):
+    """
+    Computes the autocorrelation of samples at dimension ``axis``.
+
+    :param torch.Tensor input: the input tensor.
```
neerajprad: `torch.Tensor` -> `np.ndarray`.

numpyro/diagnostics.py
```diff
@@ -0,0 +1,217 @@
...
+    :param int dim: the dimension to calculate autocorrelation.
+    :returns torch.Tensor: autocorrelation of ``input``.
```
neerajprad: same here

numpyro/diagnostics.py
```diff
@@ -0,0 +1,217 @@
...
+    [1] `Introduction to Markov Chain Monte Carlo`,
+        Charles J. Geyer
+    [2] `Stan Reference Manual version 2.18`,
+        Stan Development Team
+
+    :param numpy.ndarray x: the input array.
+    :returns numpy.ndarray: effective sample size of ``x``.
+    """
+    assert x.ndim >= 2
+    assert x.shape[1] >= 2
```
neerajprad: I think we are assuming here that `x.shape[1]` corresponds to the draw dimension which might not be correct with multiple chains?

fehiepsi: Yes, first dimension is chain dimension. It can be that `x.shape[0] == 1` (single chain).

numpyro/diagnostics.py
fehiepsi: Haha, thx! Will correct it soon.

test/test_diagnostics.py
```diff
@@ -0,0 +1,88 @@
...
+    autocovariance,
+    effective_sample_size,
+    gelman_rubin,
+    hpdi,
+    split_gelman_rubin
+)
+
+
+@pytest.mark.parametrize('statistics, input_shape, output_shape', [
+    (autocorrelation, (10,), (10,)),
```
neerajprad: Maybe we can add more chain dims to autocorrelation, autocovariance and hpdi to ensure that they are aggregating across the chain dim?

fehiepsi: Except for `hpdi` which there is no `axis` arg yet, the other two can use `axis` to specify draw dimension. Is that the thing you have in mind? I will add `axis` to hpdi.

numpyro/diagnostics.py
neerajprad: Hmm..sorry my earlier comment was wrong. I meant that the draw or batch dimension is either the first dim (0) or the second (1) with multiple chains. Just want to make sure that this will aggregate across all the chains when a chain dimension is present.

test/test_diagnostics.py
neerajprad: +1. That will be great. The reason why I'm thinking about this, is that it will be really straightforward to support serial chaining (I think you have probably already done some experiments with parallel chaining), and it will be nice if we could use these diagnostics out of the box.

numpyro/diagnostics.py
fehiepsi: Yup, split_gelman_rubin and effective_sample_size are chain diagnostics (but can work with num_chains=1), so the result is aggregated across all chains. In this PR, only `summary` does not support `chain` dimension yet.

PR #157, opened 2019-05-17
Adding some more documentation

neerajprad:
This adds docstring to the `svi` module (its pretty bare in terms of code actually).

In addition:
 - it undoes the docstring hack from #140 in favor of another more elegant hack. 😄This results in better rendered function signatures without compromising where the docstrings are placed.
 - adds an example code to HMC which can be run using doctest.

### DIFF
--- a/numpyro/hmc_util.py
+++ b/numpyro/hmc_util.py
@@ -580,9 +580,9 @@ def initialize_model(rng, model, *model_args, **model_kwargs):
     :param model: Python callable containing Pyro primitives.
     :param `*model_args`: args provided to the model.
     :param `**model_kwargs`: kwargs provided to the model.
-    :return: tuple of (`init_params`, `potential_fn`, `inv_transforms`)
+    :return: tuple of (`init_params`, `potential_fn`, `inv_transform_fn`)
         `init_params` are values from the prior used to initiate MCMC.
-        `inv_transforms` are inverse transform functions that are useful
+        `inv_transform_fn` is a callable that uses inverse transforms
         to convert unconstrained HMC samples to constrained values that
         lie within the site's support.
     """
--- a/numpyro/hmc_util.py
+++ b/numpyro/hmc_util.py
@@ -578,8 +578,8 @@ def initialize_model(rng, model, *model_args, **model_kwargs):
     :param jax.random.PRNGKey rng: random number generator seed to
         sample from the prior.
     :param model: Python callable containing Pyro primitives.
-    :param tuple model_args: args provided to the model.
-    :param dict model_kwargs: kwargs provided to the model.
+    :param `*model_args`: args provided to the model.
+    :param `**model_kwargs`: kwargs provided to the model.
     :return: tuple of (`init_params`, `potential_fn`, `inv_transforms`)
         `init_params` are values from the prior used to initiate MCMC.
         `inv_transforms` are inverse transform functions that are useful
--- a/numpyro/hmc_util.py
+++ b/numpyro/hmc_util.py
@@ -567,6 +567,24 @@ def transform_fn(transforms, params, invert=False):
 
 
 def initialize_model(rng, model, *model_args, **model_kwargs):
+    """
+    Given a model with Pyro primitives, returns a function which, given
+    unconstrained parameters, evaluates the potential energy (negative
+    joint density). In addition, this also returns initial parameters
+    sampled from the prior to initiate MCMC sampling and functions to
+    transform unconstrained values at sample sites to constrained values
+    within their respective support.
+
+    :param jax.random.PRNGKey rng: random number generator seed to
+        sample from the prior.
+    :param model: Python callable containing Pyro primitives.
+    :param tuple model_args: args provided to the model.
+    :param dict model_kwargs: kwargs provided to the model.
+    :return: tuple of (`init_params`, `potential_fn`, `transforms`)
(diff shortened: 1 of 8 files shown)


numpyro/hmc_util.py
```diff
@@ -567,6 +567,24 @@ def transform_fn(transforms, params, invert=False):
...
+    within their respective support.
+
+    :param jax.random.PRNGKey rng: random number generator seed to
+        sample from the prior.
+    :param model: Python callable containing Pyro primitives.
+    :param tuple model_args: args provided to the model.
+    :param dict model_kwargs: kwargs provided to the model.
+    :return: tuple of (`init_params`, `potential_fn`, `transforms`)
+        `init_params` are values from the prior used to initiate MCMC.
+        `transforms` are useful to convert unconstrained HMC samples
```
fehiepsi: typo: `transform_fn` instead of `transforms`. How about we rename it to `inv_transform_fn` to be aligned with Pyro NUTS's `transforms` and Stan reference manual?

numpyro/svi.py
```diff
@@ -12,7 +14,34 @@ def _seed(model, guide, rng):
...
+    :param model: Python callable with Pyro primitives for the model.
+    :param guide: Python callable with Pyro primitives for the guide
+        (recognition network).
+    :param loss: ELBo loss, i.e. negative Evidence Lower Bound, to minimize.
+    :param optim_init: initialization function returned by a JAX optimizer.
+        see: :mod:`jax.experimental.optimizers`.
+    :param optim_update: update function for the optimizer
+    :param get_params: function to get current parameters values given the
+        optimizer state.
+    :param kwargs: static arguments for the model / guide, i.e. arguments
```
fehiepsi: nit: missing `dict` for type

neerajprad: kwargs here is just a loose collection of python `**kwargs`, will change to `**kwargs` for clarity.

numpyro/hmc_util.py
neerajprad: Sounds good to me.

numpyro/hmc_util.py
```diff
@@ -580,15 +580,16 @@ def initialize_model(rng, model, *model_args, **model_kwargs):
...
     """
     model = seed(model, rng)
     model_trace = trace(model).get_trace(*model_args, **model_kwargs)
     sample_sites = {k: v for k, v in model_trace.items() if v['type'] == 'sample' and not v['is_observed']}
     transforms = {k: biject_to(v['fn'].support) for k, v in sample_sites.items()}
-    init_params = transform_fn(transforms, {k: v['value'] for k, v in sample_sites.items()}, invert=True)
+    init_params = transform_fn(transforms, {k: v['value'] for k, v in sample_sites.items()})
     return init_params, potential_energy(model, model_args, model_kwargs, transforms), \
-        jax.partial(transform_fn, transforms)
+        jax.partial(transform_fn, transforms, unconstrained=False)
```
fehiepsi: I think that this is `unconstrained=True`?

numpyro/hmc_util.py
```diff
@@ -561,8 +561,8 @@ def _potential_energy(params):
     return _potential_energy
 
 
-def transform_fn(transforms, params, invert=False):
-    return {k: transforms[k](v) if not invert else transforms[k].inv(v)
+def transform_fn(inv_transforms, params, unconstrained=True):
+    return {k: inv_transforms[k](v) if unconstrained else inv_transforms[k].inv(v)
```
fehiepsi: Maybe we can remove `unconstrained` kwargs, hidden this `transform_fn`, and make it only do inverse transform? Kind of
```
def _inv_transform_fn(inv_transforms, params):
    return {k: inv_transforms[k][v] for k, v in params.items()}
```

numpyro/hmc_util.py
```diff
@@ -580,15 +580,16 @@ def initialize_model(rng, model, *model_args, **model_kwargs):
     :param model: Python callable containing Pyro primitives.
     :param tuple model_args: args provided to the model.
     :param dict model_kwargs: kwargs provided to the model.
-    :return: tuple of (`init_params`, `potential_fn`, `transforms`)
+    :return: tuple of (`init_params`, `potential_fn`, `inv_transforms`)
```
fehiepsi: This is a callable, so I guess `inv_transform` is a better name?

fehiepsi:
Oppss, pls ignore my comments. They are outdated.

numpyro/hmc_util.py
```diff
@@ -561,16 +561,35 @@ def _potential_energy(params):
...
+    within their respective support.
+
+    :param jax.random.PRNGKey rng: random number generator seed to
+        sample from the prior.
+    :param model: Python callable containing Pyro primitives.
+    :param `*model_args`: args provided to the model.
+    :param `**model_kwargs`: kwargs provided to the model.
+    :return: tuple of (`init_params`, `potential_fn`, `inv_transforms`)
+        `init_params` are values from the prior used to initiate MCMC.
+        `inv_transforms` are inverse transform functions that are useful
```
fehiepsi: I think that `inv_transforms` is a callable which uses inverse transform functions to convert HMC ...

neerajprad: Sorry, my documentation is based on the outdated functionality when we were returning transforms as dicts. Will change.

PR #160, opened 2019-05-20
support float64

### LINKED ISSUE #77: Support float64 array

fehiepsi:
Fixes #77. Need to discuss about the API of `intialize_kernel` though.


numpyro/mcmc.py
```diff
@@ -17,8 +17,10 @@
 
 
 def _get_num_steps(step_size, trajectory_length):
-    num_steps = np.array(trajectory_length / step_size, dtype=np.int32)
-    return np.where(num_steps < 1, np.array(1, dtype=np.int32), num_steps)
+    num_steps = np.clip(trajectory_length / step_size, a_min=1)
+    # NB: casting to np.int64 does not take effect (returns np.int32 instead)
+    # if jax_enable_x64 is False
+    return num_steps.astype(np.int64)
```
fehiepsi: This is necessary to get HMC work in float64.

numpyro/hmc_util.py
```diff
@@ -586,10 +586,13 @@ def initialize_model(rng, model, *model_args, **model_kwargs):
         to convert unconstrained HMC samples to constrained values that
         lie within the site's support.
     """
+    dtype = model_kwargs.pop('dtype', np.float32)
     model = seed(model, rng)
     model_trace = trace(model).get_trace(*model_args, **model_kwargs)
     sample_sites = {k: v for k, v in model_trace.items() if v['type'] == 'sample' and not v['is_observed']}
     inv_transforms = {k: biject_to(v['fn'].support) for k, v in sample_sites.items()}
-    init_params = transform_fn(inv_transforms, {k: v['value'] for k, v in sample_sites.items()}, constrain=False)
+    init_params = transform_fn(inv_transforms,
```
neerajprad: Are these the only changes needed to run in `float64`?

fehiepsi: Yeah, we just need init values are float64 in HMC because samplers is not much involved (excet for bernoulli/uniform sampling to make transition/direction). I believe to make SVI run in float64, we have to seed dtype into samplers. So a better solition is to allow dtype keywork in distributions’ sample method. What do you think?

neerajprad: Hmm..if distribution parameters are `float64`, the samples drawn should automatically be float64 values, right? I think the most non-invasive way to support `float64` might be to have users explicitly use `np.float64` dtype in their models. We may still need to use the sample's dtype to construct float64 arrays when generating momentum etc, but it will be nice to be as less invasive as possible. Would also be good to convert an example/test to use `float64` and test.

fehiepsi: > if distribution parameters are float64, the samples drawn should automatically be float64 values

This is the case but e.g. uniform samples take max value is `1 - finfo(float32).eps` instead of `1 - finfo(float64).eps`. Random samplers in JAX has a `dtype` arg where default dtype is np.float32. If distribution params are float64, then operators like `loc + scale * eps` will return `float64`. But if distribution params are python scalar, then the returned value is `float32` (if eps comes from a random sampler with default dtype=np.float32). If we override the default behaviour of JAX samplers by creating wrappers in `util` with default dtype is `np.float64`, then we don't need to make this change in `initialize_model`. I think that this is also a good way to do because it can support both HMC and SVI.

> I think the most non-invasive way to support float64 might be to have users explicitly use np.float64 dtype in their models.

This actually makes things complicated, when users have to rewrite their models to run models in float64. Even if users make dist params to be float64, the samples are still in float32 and then converted to float64. If you don't like to have additional `dtype` arg then I propose to make wrappers as above. Then users/we don't have to touch anything. Just enable float64, the inference will run in float64. Otherwise, it will run in float32. Of course, there is no support for: enable float64 and want to run inference in float32.

> Would also be good to convert an example/test to use float64 and test.

Yes, because we have to enable float64 at beginning of the test, so I guess we need to add separated commands in travis. Something like
```
JAX_ENABLE_x64=True pytest test/test_mcmc.py
```
I will do it. If you have other suggestions, please let me know.

neerajprad: > If we override the default behaviour of JAX samplers by creating wrappers in util with default dtype is np.float64, then we don't need to make this change in initialize_model.

I think we should do that anyways (assuming that there isn't another pattern to change the default dtype), since even for HMC, we actually do want the sampler to use `float64` and not just do the computation in `float32` and then cast it to `float64`. I wish there was a simpler way to achieving this though - an analogue of `torch.set_default_tensor_type()` kind of a thing that did not require changing the default of every singe array constructor.

fehiepsi: Yeah, I wish so too. But jax random samplers do not follow that semantics. Other constructors such as `np.ones()`, `np.arange()`, `np.log(1)` works as expected, except random samplers. >"<

test/test_mcmc.py
```diff
@@ -25,13 +27,16 @@ def potential_fn(z):
     init_kernel, sample_kernel = hmc(potential_fn, algo=algo)
     init_params = np.array(0.)
     hmc_state = init_kernel(init_params,
-                            trajectory_length=10,
+                            trajectory_length=9,
```
fehiepsi: this is to pass precision test in float64

numpyro/distributions/random.py
```diff
@@ -12,27 +12,33 @@
 PRNGKey = random.PRNGKey
 
 
-def bernoulli(key, p=0.5, shape=(), dtype=np.float64):
+def bernoulli(key, p=0.5, shape=()):
+    dtype = np.zeros((0,)).dtype
     return random.bernoulli(key, lax.convert_element_type(p, dtype), shape)
```
neerajprad: Can we pass in `dtype=float64` here always? If `float64` is enabled that should initialize the sampler with the correct dtype, otherwise it should fall back to `float32`. Alternatively, we can do this just once and use the default type rather than creating a fresh numpy array each time a sampler is called to get the dtype (maybe we don't even need to do this if we can read the jax config instead).

```python
_DEFAULT_DTYPE = None  # depending on when the config is updated we can probably just assign the 
                       # dtype here or check from jax config.

def _get_default_dtype():
    global _DEFAULT_DTYPE
    if not _DEFAULT_DTYPE: 
        _DEFAULT_DTYPE = np.zeros((0,)).dtype
    return _DEFAULT_DTYPE
```

test/test_examples.py
```diff
@@ -11,8 +11,7 @@
 EXAMPLES = [
     'baseball.py --num-samples 100 --num-warmup 100',
     'covtype.py --algo hmc --num-samples 10',
-    pytest.param('hmm.py --num-samples 100 --num-warmup 100', marks=[pytest.mark.skip(
-        reason='lax.scan is not yet available in jax 0.1.28')]),
+    'hmm.py --num-samples 100 --num-warmup 100',
```
neerajprad: Nice!

numpyro/distributions/random.py
neerajprad: It also seems to me that this should be fixed in JAX to ensure that samplers behave consistently w.r.t. other array constructors in that if `float64` mode is enabled they do the same rather than always using `float32` as default.

fehiepsi: Yes, we can except for normal where finfo(dtype) is used in the implementation. Implementations which depends in `normal` also get this issue.

fehiepsi: I can follow your second suggestion instead. Thanks!

neerajprad: Created a jax issue as well: https://github.com/google/jax/issues/756.

fehiepsi: Thanks @neerajprad !

fehiepsi:
I'll remove most of the changes in this PR because most problems are addressed upstream. :)

fehiepsi:
Just revert the changes. I'll merge this when tests pass. :)

Issue #144, opened 2019-05-11
API issues to clean up before release


neerajprad:
This is to track minor API issues as we notice them, that will be nice to clean up. This is not super urgent - these are known things that can be cleaned up before release.

 - [x] It will be nice to see other diagnostic information like acceptance rate, and step size, since that's the easiest thing to make note of if the model is misspecified. @neerajprad 
 - [x] Support initialization from `Uniform(-2, 2)` directly on unconstrained parameters like in Stan.
 - [x] Renaming: Potential candidates for renaming (up for discussion)
   - heuristic_step_size: The name makes it appear as though it is a float arg and not a boolean arg. Don't have a better idea at the moment though. 
   - num_warmup_steps --> num_warmup or warmup_steps
 - [x] Certain functions need args, kwargs to be tuples, dicts so that the args / kwargs repacking does not result in jit recompilation. However, this style isn't very pythonic otherwise, and should be avoided when there is no need to JIT. So lets change `initialize_model(rng, model, model_args, model_kwargs)` to `initialize_model(rng, model, *args, **kwargs)`.
- [x] If the `model` or `potential_fn` is not jittable, I think we'll end up throwing an exception given than `sample_kernel` has a `jit` decorator. I think the right approach would be to pass a `jit_compile=True` flag to `hmc` so that the user does not need to modify the source code in case their model is not jittable.
- [x] Use default args such as loc=0, scale=1 in distributions
- [x] Rename args in Pareto for consistence.

neerajprad:
@fehiepsi - Feel free to add to this if you notice anything that appears off in our API.

fehiepsi:
> acceptance_rate

Would it be `accept_prob` instead? To me, accept_prob is a more meaningful term than acceptance_rate. It also aligns with `target_accept_prob` arg. In addition, num_steps is also a useful dynamic term which I usually print out to see how the model and inference works. Please consider it if the progbar has some spaces. :)

> heuristic_step_size: The name makes it appear as though it is a float arg and not a boolean arg. Don't have a better idea at the moment though.

[...]

PR #162, opened 2019-05-20
Change parameters initialization to Uniform(-2, 2)

neerajprad:
Addresses #144. Based on @fehiepsi's suggestion in #154.

### DIFF
--- a/numpyro/hmc_util.py
+++ b/numpyro/hmc_util.py
@@ -595,15 +595,15 @@ def initialize_model(rng, model, *model_args, initialize='uniform', **model_kwar
     model_trace = trace(model).get_trace(*model_args, **model_kwargs)
     sample_sites = {k: v for k, v in model_trace.items() if v['type'] == 'sample' and not v['is_observed']}
     inv_transforms = {k: biject_to(v['fn'].support) for k, v in sample_sites.items()}
-    prior_samples = constrain_fn(inv_transforms,
-                                 {k: v['value'] for k, v in sample_sites.items()}, invert=True)
+    prior_params = constrain_fn(inv_transforms,
+                                {k: v['value'] for k, v in sample_sites.items()}, invert=True)
     if initialize == 'uniform':
         init_params = {}
-        for k, v in prior_samples.items():
+        for k, v in prior_params.items():
             rng, = random.split(rng, 1)
             init_params[k] = random.uniform(rng, shape=np.shape(v), minval=-2, maxval=2)
     elif initialize == 'prior':
-        init_params = prior_samples
+        init_params = prior_params
     else:
         raise ValueError('initialize={} is not a valid initialization strategy.'.format(initialize))
     return init_params, potential_energy(model, model_args, model_kwargs, inv_transforms), \
--- a/numpyro/hmc_util.py
+++ b/numpyro/hmc_util.py
@@ -566,7 +566,7 @@ def constrain_fn(inv_transforms, params, invert=False):
             for k, v in params.items()}
 
 
-def initialize_model(rng, model, *model_args, initialize='uniform', **model_kwargs):
+def initialize_model(rng, model, *model_args, init_strategy='uniform', **model_kwargs):
     """
     Given a model with Pyro primitives, returns a function which, given
     unconstrained parameters, evaluates the potential energy (negative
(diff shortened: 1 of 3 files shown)


fehiepsi:
Thanks @neerajprad ! I think that this also fixes #154.

numpyro/hmc_util.py
```diff
@@ -590,6 +595,16 @@ def initialize_model(rng, model, *model_args, **model_kwargs):
     model_trace = trace(model).get_trace(*model_args, **model_kwargs)
```
fehiepsi: I think that we need to split rng into key and subkey, where key is used to seed model, and the other one to be used in `uniform` strategy.

numpyro/hmc_util.py
```diff
@@ -590,6 +595,16 @@ def initialize_model(rng, model, *model_args, **model_kwargs):
...
     sample_sites = {k: v for k, v in model_trace.items() if v['type'] == 'sample' and not v['is_observed']}
     inv_transforms = {k: biject_to(v['fn'].support) for k, v in sample_sites.items()}
-    init_params = constrain_fn(inv_transforms, {k: v['value'] for k, v in sample_sites.items()}, invert=True)
+    prior_samples = constrain_fn(inv_transforms,
+                                 {k: v['value'] for k, v in sample_sites.items()}, invert=True)
```
fehiepsi: While you address the above `rng` issue, could you change the name of `prior_samples` too. For example, `prior_params` or `prior_unconstrained...` will be less confusing.

numpyro/hmc_util.py
neerajprad: Does it matter? The random values from the seeded model aren't used (only the shapes are used), and we split the seed in the uniform initialization anyways.

fehiepsi: Kaka, you are absolutely right!

fehiepsi:
Looking like uniform strategy makes the changepoint test fails. @@ Do we discover another typical set of `tau` for this problem? :D

test/test_mcmc.py
```diff
@@ -126,7 +126,8 @@ def model(data):
         12,  35,  17,  23,  17,   4,   2,  31,  30,  13,  27,   0,  39,  37,
         5,  14,  13,  22,
     ])
-    init_params, potential_fn, constrain_fn = initialize_model(random.PRNGKey(2), model, count_data)
+    init_params, potential_fn, constrain_fn = initialize_model(random.PRNGKey(2), model, count_data,
+                                                               initialize='prior')
     init_kernel, sample_kernel = hmc(potential_fn)
```
neerajprad: @fehiepsi - I think there might be an issue with this test, specially with `assert max(tau_values) == 44`. No matter the number of warmup steps, I get ~ 70-71 with the uniform initialization strategy. I have changed it to `prior` above, but this is worth looking into.

fehiepsi: Looking at the plot at https://nbviewer.jupyter.org/github/CamDavidsonPilon/Probabilistic-Programming-and-Bayesian-Methods-for-Hackers/blob/master/Chapter1_Introduction/Ch1_Introduction_PyMC3.ipynb#Example:-Inferring-behaviour-from-text-message-data I think that we are just discovering another typical set of `tau`. The picture shows 2 highest modes: 1 is 44, 1 is 22. So 22 might be another solution of this problem. :D

neerajprad: The strange thing is that with uniform initialization, I consistently see 70-71 as max value for tau and for prior initialization the max is around 44-45. I would have thought that increasing the number of samples / warmup would uncover all the modes, but that doesn't seem to be the case.

test/test_mcmc.py
```diff
@@ -135,7 +136,6 @@ def model(data):
     tau_values, counts = onp.unique(tau_posterior, return_counts=True)
     mode_ind = np.argmax(counts)
     mode = tau_values[mode_ind]
-    assert max(tau_values) == 44
```
neerajprad: I verified that the mode remains the same when we increase the number of samples. This assertion however is violated and I don't think there is a reason why this should be true, so I'm removing it.

fehiepsi: Yup, you are right. I don't think so too. :D

test/test_mcmc.py
neerajprad: I have reverted to using the prior distribution because the sampling seems to get stuck at one of the modes and give biased results. This is a great model to run with multiple chains and check for convergence. :)

fehiepsi:
I think that with this random seed, the chain failed to converge. >"< I will test if this happens with Stan too.

neerajprad:
> I will test if this happens with Stan too.

With 4 parallel chains, it probably does a better job of exploring different regions of the posterior. Will increase the number of samples for this.

test/test_mcmc.py
fehiepsi: Yeah, this and your last regression model are the first HMC models which I see initialization plays an important role. :D :dancing_men:  It is good that we allow users provide custom initial values as their likes. Now, I understand why some users in forum and pyro ask for that feature.

PR #165, opened 2019-05-22
Use default support in transformed distributions

fehiepsi:
As title. This also fixes a small bug at support of Pareto.

### DIFF
--- a/numpyro/distributions/continuous.py
+++ b/numpyro/distributions/continuous.py
@@ -503,8 +503,7 @@ class Pareto(TransformedDistribution):
         batch_shape = lax.broadcast_shapes(np.shape(scale), np.shape(alpha))
         self.scale, self.alpha = np.broadcast_to(scale, batch_shape), np.broadcast_to(alpha, batch_shape)
         base_dist = Exponential(self.alpha)
-        transforms = [ExpTransform(domain=constraints.positive),
-                      AffineTransform(loc=0, scale=self.scale, domain=constraints.greater_than(1.))]
+        transforms = [ExpTransform(), AffineTransform(loc=0, scale=self.scale)]
         super(Pareto, self).__init__(base_dist, transforms, validate_args=validate_args)
 
     @property
@@ -518,6 +517,11 @@ class Pareto(TransformedDistribution):
         # var is inf for alpha <= 2
         a = lax.div((self.scale ** 2) * self.alpha, (self.alpha - 1) ** 2 * (self.alpha - 2))
         return np.where(self.alpha <= 2, np.inf, a)
+
+    # override the default behaviour to save computations
+    @property
+    def support(self):
+        return constraints.greater_than(self.scale)
 
 
 class StudentT(Distribution):
--- a/numpyro/distributions/continuous.py
+++ b/numpyro/distributions/continuous.py
@@ -238,7 +238,6 @@ class GaussianRandomWalk(Distribution):
 class HalfCauchy(TransformedDistribution):
     reparametrized_params = ['scale']
     arg_constraints = {'scale': constraints.positive}
-    support = constraints.positive
 
     def __init__(self, scale=1., validate_args=None):
         base_dist = Cauchy(0., scale)
@@ -263,7 +262,6 @@ class HalfCauchy(TransformedDistribution):
 class HalfNormal(TransformedDistribution):
     reparametrized_params = ['scale']
     arg_constraints = {'scale': constraints.positive}
-    support = constraints.positive
 
     def __init__(self, scale=1., validate_args=None):
         base_dist = Normal(0., scale)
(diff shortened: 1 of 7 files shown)


numpyro/distributions/continuous.py
```diff
@@ -501,13 +498,13 @@ def variance(self):
...
 class Pareto(TransformedDistribution):
     arg_constraints = {'alpha': constraints.positive, 'scale': constraints.positive}
-    support = constraints.real
 
     def __init__(self, alpha, scale=1., validate_args=None):
         batch_shape = lax.broadcast_shapes(np.shape(scale), np.shape(alpha))
         self.scale, self.alpha = np.broadcast_to(scale, batch_shape), np.broadcast_to(alpha, batch_shape)
         base_dist = Exponential(self.alpha)
-        transforms = [ExpTransform(), AffineTransform(loc=0, scale=self.scale)]
+        transforms = [ExpTransform(domain=constraints.positive),
```
neerajprad: It would be nice if `domain` could also be set later via a setter. Then we wouldn't have to pass domain to any of these transforms and we could just infer it in `TransformedDistribution`, but that can be delegated to another PR perhaps.

numpyro/distributions/continuous.py
```diff
@@ -482,7 +480,6 @@ def variance(self):
 
 class LogNormal(TransformedDistribution):
     arg_constraints = {'loc': constraints.real, 'scale': constraints.positive}
-    support = constraints.positive
```
neerajprad: This code will be run for every `sample("X", dist.X)` call. With these changes, we'll do some additional computation inside transforms and `TransformedDistribution.support`. Could you verify that this isn't leading to any significant perf regression when we call `dist.__init__()` many times?

fehiepsi: Hmm, I am confusing... Why we created TransformedDistribution because it is elegant, then worry for tiny performance regression due to the overhead of one or two python calls? If it is about performance, I would propose to worry more about something like:
+ the order of operators, when batch is involved
+ group operator together: for example, log(a) + log(b) = log(a * b)
+ cancellation in log_prob computation of Transformed distribution (did we benchmark against the native version where no TransformedDistribution is involved?)

I guess there is a signifiant perf regression for Pareto distribution, when support is calculated by `lower_bound = 0 + np.log(1.) * scale` instead of `lower_bound = np.log(scale)`. Additional + and * in jax.numpy calculation, although performed in singleton, will add overhead when it is not jitted because that's how JAX works. Each call will add hundred of microseconds. If that is what you worry about (where validation is enabled, no jit is involved, and don't want to spend hundreds of microsecond on these cheap calculations due to JAX +/* overheads), then I will change the support of Pareto to `np.log(scale)` instead.

numpyro/distributions/continuous.py
fehiepsi: Yeah, that would be ideal. This kind of pattern should be the default behaviour of `TransformedDistribution` (and `ComposedTransformation`), where `domain` is the support of base distribution and codomain is retrieved recursively. This will prevent errors when users try to use TransformedDistribution but forget to redefine a correct support. By using setter, we can also take care of the domain/codomain relation for `transform.inv`. We can override the default behaviour if we already know the support. I'll refine the default behaviour and defer the `setter` pattern to a later PR, when the idea is more consolidated. Will add a TODO for it too.

numpyro/distributions/continuous.py
neerajprad: The general pattern of inferring the codomain is very useful for user-defined distributions, and would prevent unanticipated bugs. Adding one or two Python calls to get more readable code is great, but I am not entirely convinced that removing the support attribute for the distribution implementations we provide is necessarily cleaner or more readable. Likewise, if you think there is a faster, elegant implementation for some of these distributions that doesn't use TransformedDistribution, I think it is reasonable to use that instead of relying on TransformedDistribution. 

The overhead from 1 or 2 Python calls adds up when we call `__init__` on these distributions many thousands of time (basically each time we evaluate potential energy). I agree with your assessment though that the only time we would be doing so is when we are running inference with JIT in which case these calls should be highly optimized and probably shouldn't add to the runtime. At least that is what I would hope and wanted to verify. Otherwise, the usage pattern would be to instantiate once and use the same distribution instance to sample from many times, so the initial cost will be amortized away. 

This is a very minor point though - I don't have a strong opinion on this, unless we profile and find this to add several seconds, which I highly doubt would be the case.

neerajprad:
Thanks for proposing and implementing this very useful change for TransformedDistribution!

numpyro/distributions/continuous.py
fehiepsi: I thought that the removal of support only affects `initialize_model` where we need to access `support` to get `transforms`, or when we turn on validate mode. Does a `@property` attribute of a class trigger when we call `__init__`? Maybe I understand your comment wrongly.

neerajprad: > @property attribute of a class trigger when we call __init__? 

No, it shouldn't be triggered. Sorry, that was really my misunderstanding - I thought that we are populating `.support` inside of `TransformedDistribution.__init__`. But you are absolutely right that this is only called once in `initialize_model` or when `validate_args` is True. Please disregard my comment above - this should have no effect on perf. :)

PR #166, opened 2019-05-24
add bayesian neural network example

martinjankowiak:
- [x] test examples logic?

anything else i might consider adding, @neerajprad? best practices i'm not following?

### DIFF
--- a/examples/bnn.py
+++ b/examples/bnn.py
@@ -1,0 +1,129 @@
+import matplotlib
+matplotlib.use('Agg')  # noqa: E402
+import matplotlib.pyplot as plt
+
+import argparse
+
+import numpy as onp
+import jax.numpy as np
+import jax.random as random
+
+import numpyro.distributions as dist
+from numpyro.handlers import sample, seed, substitute, trace
+from numpyro.hmc_util import initialize_model
+from numpyro.mcmc import mcmc
+
+
+"""
+We demonstrate how to use NUTS to do inference on a simple (small)
+Bayesian neural network with two hidden layers.
+"""
+
+
+# the non-linearity we use in our neural network
+def nonlin(x):
+    return np.tanh(x)
+
+
+# a two-layer bayesian neural network with computational flow
+# given by D_X => D_H => D_H => D_Y where D_H is the number of
+# hidden units. (note we indicate tensor dimensions in the comments)
+def model(X, Y, D_H):
+    D_X = X.shape[1]
+    D_Y = 1
+
+    # sample first layer (we put unit normal priors on all weights)
+    w1 = sample("w1", dist.Normal(np.zeros((D_X, D_H)), np.ones((D_X, D_H))))  # D_X D_H
+    z1 = nonlin(np.matmul(X, w1))   # N D_H  <= first layer of activations
+
+    # sample second layer
+    w2 = sample("w2", dist.Normal(np.zeros((D_H, D_H)), np.ones((D_H, D_H))))  # D_H D_H
+    z2 = nonlin(np.matmul(z1, w2))  # N D_H  <= second layer of activations
+
+    # sample final layer of weights and neural network output
+    w3 = sample("w3", dist.Normal(np.zeros((D_H, D_Y)), np.ones((D_H, D_Y))))  # D_H D_Y
+    z3 = np.matmul(z2, w3)  # N D_Y  <= output of the neural network
+
+    # we put a prior on the observation noise
+    prec_obs = sample("prec_obs", dist.Gamma(3.0, 1.0))
+    sigma_obs = 1.0 / np.sqrt(prec_obs)
+
+    # observe data
+    sample("Y", dist.Normal(z3, sigma_obs), obs=Y)
+
+
+# helper function for HMC inference
+def run_inference(model, args, rng, X, Y, D_H):
+    init_params, potential_fn, constrain_fn = initialize_model(rng, model, X, Y, D_H)
+    hmc_states = mcmc(args.num_warmup, args.num_samples, init_params,
+                      sampler='hmc', potential_fn=potential_fn, constrain_fn=constrain_fn)
+    return hmc_states
+
+
+# helper function for prediction
+def predict(model, rng, latents, X, D_H):
+    # we need to expand prec_obs so that the (hmc) samples dimension doesn't
+    # conflict with the data dimension in the final model sample statement
+    latents['prec_obs'] = latents['prec_obs'][:, None, None]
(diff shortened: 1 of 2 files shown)


examples/bnn.py
```diff
@@ -0,0 +1,129 @@
...
+    hmc_states = mcmc(args.num_warmup, args.num_samples, init_params,
+                      sampler='hmc', potential_fn=potential_fn, constrain_fn=constrain_fn)
+    return hmc_states
+
+
+# helper function for prediction
+def predict(model, rng, latents, X, D_H):
+    # we need to expand prec_obs so that the (hmc) samples dimension doesn't
+    # conflict with the data dimension in the final model sample statement
+    latents['prec_obs'] = latents['prec_obs'][:, None, None]
```
martinjankowiak: this is kind of gross :cry:

neerajprad: It's actually not as bad as some of the other examples, like baseball. :) What you could also do is remove this line, and just use vmap to batch over the latents (and rngs) instead in your main method. This is really useful when you need to do more complex manipulation for getting the batch shapes in order.

```python
predictions = vmap(lambda samples, rng: predict(model, rng, samples, X_test, D_H))\
     (latents, random.split(rng_predict, args.num_samples))[..., 0]
```

We can thank @fehiepsi for this really useful trick!

examples/bnn.py
```diff
@@ -0,0 +1,129 @@
...
+    mean_prediction = np.mean(predictions, axis=0)
+    percentiles = onp.percentile(predictions, [5.0, 95.0], axis=0)
+
+    # plot training data
+    ax.plot(X[:, 1], Y[:, 0], 'kx')
+    # plot 90% confidence level of predictions
+    ax.fill_between(X_test[:, 1], percentiles[0, :], percentiles[1, :], color='lightblue')
+    # plot mean prediction
+    ax.plot(X_test[:, 1], mean_prediction, 'blue', ls='solid', lw=2.0)
+
```
neerajprad: Can we also add labels and title to the plot?
```python
    ax.set(xlabel="X",
           ylabel="Y",
           title="Mean predictions with X% CI")
```

examples/bnn.py
fehiepsi: As an example to illustrate the trick, you can take a look at [this notebook](https://nbviewer.jupyter.org/github/pyro-ppl/numpyro/blob/e07e1b5470e2753c5b8073afccb4a8659eed1ee5/notebooks/Pylgt.ipynb), where I used `vmap` on `forecast` function to get `forecast_marginal`.

fehiepsi:
>  this example is actually maxing out the number of steps possible in NUTS which is 1023, so it seems to go on for that long without taking a U-turn

I faced it several times. `covtype` data is an example. And in early versions of semi-hmm too. :D

examples/bnn.py
martinjankowiak: ah yeah good point, `vmap` to the rescue!

examples/bnn.py
```diff
@@ -0,0 +1,129 @@
...
+    # note that Y will be sampled in the model because we pass Y=None here
+    model_trace = trace(model).get_trace(X=X, Y=None, D_H=D_H)
+    return model_trace['Y']['value']
+
+
+# create artificial regression dataset
+def get_data(N=50, D_X=3, sigma_obs=0.05, N_test=400):
+    D_Y = 1  # create 1d outputs
+    onp.random.seed(0)
+    X = 2.0 * np.arange(N) / N - 1.0
```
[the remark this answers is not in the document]
martinjankowiak: done

examples/bnn.py
```diff
@@ -0,0 +1,129 @@
...
+
+# a two-layer bayesian neural network with computational flow
+# given by D_X => D_H => D_H => D_Y where D_H is the number of
+# hidden units. (note we indicate tensor dimensions in the comments)
+def model(X, Y, D_H):
+    D_X = X.shape[1]
+    D_Y = 1
+
+    # sample first layer (we put unit normal priors on all weights)
+    w1 = sample("w1", dist.Normal(np.zeros((D_X, D_H)), np.ones((D_X, D_H))))  # D_X D_H
```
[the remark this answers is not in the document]
martinjankowiak: yeah i think `module` has the unfortunate consequence of making some people think bayesian inference works on neural network models with 10^8 parameters.....

examples/bnn.py
```diff
@@ -0,0 +1,129 @@
...
+# create artificial regression dataset
+def get_data(N=50, D_X=3, sigma_obs=0.05, N_test=400):
+    D_Y = 1  # create 1d outputs
+    onp.random.seed(0)
+    X = 2.0 * np.arange(N) / N - 1.0
+    X = np.power(X[:, None], np.arange(D_X))
+    W = 0.5 * onp.random.randn(D_X)
+    Y = np.dot(X, W) + 0.5 * np.power(0.5 + X[:, 1], 2.0) * np.sin(4.0 * X[:, 1])
+    Y += sigma_obs * onp.random.randn(N)
+    Y = Y[:, onp.newaxis]
```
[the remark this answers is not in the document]
martinjankowiak: done

examples/bnn.py
```diff
@@ -0,0 +1,129 @@
...
+    sigma_obs = 1.0 / np.sqrt(prec_obs)
+
+    # observe data
+    sample("Y", dist.Normal(z3, sigma_obs), obs=Y)
+
+
+# helper function for HMC inference
+def run_inference(model, args, rng, X, Y, D_H):
+    init_params, potential_fn, constrain_fn = initialize_model(rng, model, X, Y, D_H)
+    hmc_states = mcmc(args.num_warmup, args.num_samples, init_params,
```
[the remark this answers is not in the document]
martinjankowiak: done

neerajprad:
> I faced it several times. covtype data is an example.

Maybe, that is worth studying too then. :) This can potentially indicate problems with the model, or just that inference is slow, e.g. if the posterior is highly correlated, and will throw max treedepth warnings on Stan. You'll notice that the effective sample size is low too. Anyways, we are getting good predictions here, but it is worth studying in more detail.

examples/bnn.py
martinjankowiak: done

examples/bnn.py
martinjankowiak: although arguably the better solution would be for named tensor dimensions: then the subsitute-a-full-tensor version would have worked too : )

martinjankowiak:
@neerajprad thanks for the detailed review; all comments should now be addressed!

PR #167, opened 2019-05-24
add batching rule for standard_gamma

fehiepsi:
This is useful for using `vmap` on the distributions require `standard_gamma`. In particularly, this is required for LGT example. I also add some non-related fixes which I catch during the way.

Note that there might be some edge cases which are not covered in this batching rule (because my understanding of batching rule is limited). Until now I don't know any of them, so I would like to resolve related bugs (if any) in future PRs.

### DIFF
--- a/numpyro/distributions/continuous.py
+++ b/numpyro/distributions/continuous.py
@@ -121,9 +121,10 @@ class Dirichlet(Distribution):
     def log_prob(self, value):
         if self._validate_args:
             self._validate_sample(value)
-        normalize_term = (np.sum(gammaln(self.concentration), axis=-1) -
-                          gammaln(np.sum(self.concentration, axis=-1)))
-        return np.sum(np.log(value) * (self.concentration - 1.), axis=-1) - normalize_term
+        concentration = lax.convert_element_type(self.concentration, value.dtype)
+        normalize_term = (np.sum(gammaln(concentration), axis=-1) -
+                          gammaln(np.sum(concentration, axis=-1)))
+        return np.sum(np.log(value) * (concentration - 1.), axis=-1) - normalize_term
 
     @property
     def mean(self):
--- a/numpyro/distributions/util.py
+++ b/numpyro/distributions/util.py
@@ -7,7 +7,7 @@ import scipy.special as osp_special
 
 import jax.numpy as np
 from jax import canonicalize_dtype, custom_transforms, device_get, jit, lax, random, vmap
-from jax.interpreters import ad
+from jax.interpreters import ad, batching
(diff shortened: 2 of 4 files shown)


numpyro/distributions/continuous.py
```diff
@@ -121,9 +121,10 @@ def sample(self, key, size=()):
     def log_prob(self, value):
         if self._validate_args:
             self._validate_sample(value)
-        normalize_term = (np.sum(gammaln(self.concentration), axis=-1) -
-                          gammaln(np.sum(self.concentration, axis=-1)))
-        return np.sum(np.log(value) * (self.concentration - 1.), axis=-1) - normalize_term
+        concentration = lax.convert_element_type(self.concentration, value.dtype)
+        normalize_term = (np.sum(gammaln(concentration), axis=-1) -
+                          gammaln(np.sum(concentration, axis=-1)))
```
fehiepsi: gammaln requires `concentration` is float, so we need to use `convert_element_type`. Value is a simplex so it should be `float`.

neerajprad: I think we are stricter in terms of typing as compared to scipy distributions, and you'll probably find other distributions too which would throw an error if we passed in `np.array([1])` instead of `np.array([1.])` for some parameter (or value), when it expected a float. I think it would make sense to consistently handle this one way or the other. I prefer the stricter approach simply because its much easier to maintain than having to do this casting for all `sample` and `log_prob` methods. e.g. Another question might be whether we also want to support casting when a mix of float32 and float64 values are present, which would further complicate matters. 

If you think this will be nice to have, how about we place this logic in `__init__` and just convert it into the default float type (maybe we can disregard cases with mixed float32 and float64 types to begin with)? We can later handle this for other distributions.

numpyro/distributions/util.py
```diff
@@ -175,6 +179,7 @@ def _standard_gamma_p(key, alpha):
 
 ad.defjvp2(_standard_gamma_p.primitive, None,
            lambda tangent, sample, key, alpha, **kwargs: tangent * _standard_gamma_grad(sample, alpha))
+batching.defvectorized(_standard_gamma_p.primitive)
```
neerajprad: Nice!

numpyro/distributions/continuous.py
fehiepsi: Until now, I only observe that gammaln requires float as an argument, so we need casting before applying this operator. I think that it is jax issue (not casting to canonical float for this operator) but it might be not necessary to raise an issue upstream. If there are other operators which needs this casting, we might consider making things consistence.

Issue #48, opened 2019-03-15
Implement pathwise derivative for dirichlet distribution


neerajprad:
Refer to #44. We can implement dirichlet sample and grad methods that are similar to the ones we have in PyTorch.

fehiepsi:
More context about this issue:

Dirichlet sampler is available in #81. However, the derivative is based on the gamma sampler which generates dirichlet samples. It is not a correct pathwise derivative. The algorithm for Dirichlet pathwise derivate can be found in [this paper](https://arxiv.org/abs/1806.01851).

fehiepsi:
The current Dirichlet sampler has a reparameterized gradient but its variance would be higher than the version in PyTorch. Given that we mostly rely on the upstream jax samplers now, it is better to use the current high variance version and raise the request to jax team (if needed) instead.

adamgayoso:
Hello,

I'm trying to sample from a Dirichlet with 4 dimensions, but over a batch of thousands of elements. The gradients are very slow compared to PyTorch. Has there been any progress on speeding this up ?

fehiepsi:
Could you try TFP Dirichlet distribution? IIRC it has faster dirichlet sampler.

adamgayoso:
I'm getting nan's there related to dirichlet sampling where I'm not with jax.random.dirichlet/numpyro!...

adamgayoso:
I can confirm the sampling if faster in TFP, but I'll get NaN's during training, which I haven't narrowed down to a gradient issue or a sampling issue.

fehiepsi:
You might want to create a wrapper to [clip Dirichlet samples](https://github.com/pyro-ppl/numpyro/blob/f81ce325de402d9a9891497136a36e3b44f982cc/numpyro/distributions/continuous.py#L261). Something like
```
class RobustDirichlet(...):
    def sample(self, ...):
        samples = super().sample(...)
        return clip(samples)
```

martinjankowiak:
@adamgayoso it might help to make sure your parameter is bounded away from zero e.g. `concentration=0.01 + positive_param`

adamgayoso:
Thank you both for the suggestions, although I'm still getting nan's. I'd need more time to inspect if it's the gradients, but do you know if I can stitch the jax random dirichlet gradient fn with tfp gamma sampler?

martinjankowiak:
```
from tensorflow_probability.substrates.jax import distributions as tfd

def model():
[...]

PR #168, opened 2019-05-24
Add pathwise tests for some samplers

fehiepsi:
This PR adds some pathwise tests for some samplers. Currently, tests for Beta/StudentT are failed as expected. See #48

### DIFF
--- a/test/test_distributions.py
+++ b/test/test_distributions.py
@@ -9,7 +9,7 @@ from numpy.testing import assert_allclose, assert_array_equal
 import jax
 import jax.numpy as np
 import jax.random as random
-from jax import grad, lax, vmap
+from jax import grad, jacfwd, lax, vmap
 
 import numpyro.distributions as dist
 import numpyro.distributions.constraints as constraints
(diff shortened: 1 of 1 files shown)


test/test_distributions.py
```diff
@@ -249,6 +249,31 @@ def fn(args):
...
+    (dist.Gamma, osp.gamma, (0.1,)),
+    (dist.Gamma, osp.gamma, (10.,)),
+    # TODO: add more test cases for Beta/StudentT (and Dirichlet too) when
+    # their pathwise grad (independent of standard_gamma grad) is implemented.
+    pytest.param(dist.Beta, osp.beta, (1., 1.), marks=pytest.mark.xfail(
+        reason='currently, variance of grad of beta sampler is large')),
+    pytest.param(dist.StudentT, osp.t, (1.,), marks=pytest.mark.xfail(
+        reason='currently, variance of grad of t sampler is large')),
+])
+def test_pathwise_gradient(jax_dist, sp_dist, params):
```
neerajprad: From my understanding, `test_sample_gradient` is also a test for pathwise gradient, though I think this estimate might be more accurate (but I'm not sure since finite difference on the CDF can be numerically unstable too)?

fehiepsi: The CDF formula is used to compute pathwise derivative for those samplers, so it is better to test against that formula. I am not sure how precise it is when use `test_sample_gradient`, but it is better to pass these tests before passing `test_sample_gradient`.

PR #171, opened 2019-05-29
Add basic tutorial on bayesian regression

neerajprad:
This adds a basic bayesian regression tutorial for Numpyro. 

Reviewers - please feel free to push directly any nits / minor edits directly to this branch since that might be easier than adding comments to the notebook. 

Suggestions to improve the organization or clarifying certain points are welcome.


fehiepsi:
Look great overall! In addition to Martin's review, I just have some small nits:
+ why we need `%reset -s -f`?
+ use `from jax.config import config; config.update("jax_platform_name", "cpu")` to disable jax warning on falling back to CPU.
+ mention that `exp_log_likelihood` is expected log likelihood (instead of exponent ^^)

neerajprad:
@martinjankowiak, @fehiepsi - Thanks for reviewing! Will address all your comments shortly.

neerajprad:
> %reset -s -f

I like to do this to clear all the global variables when running the notebook. Its nice to be able to go back to that cell and run it to clear all variables in working memory, and run things afresh.

neerajprad:
> why 89%?

This is just the default returned by `HPDI`, same as in R (https://www.rdocumentation.org/packages/rethinking/versions/1.59/topics/HPDI). I'm sure there must be some reason for this, but I can specify any other value too in the call if 89% looks strange.

fehiepsi:
I just quote from **Rethinking** book

**Rethinking:** Why 95%? The most common interval mass in the natural and social sciences is the
95% interval. This interval leaves 5% of the probability outside, corresponding to a 5% chance of the
parameter not lying within the interval (although see below). This customary interval also reflects
the customary threshold for statistical significance, which is 5% or p < 0.05. It is not easy to defend
the choice of 95% (5%), outside of pleas to convention. Ronald Fisher is sometimes blamed for this
choice, but his widely cited 1925 invocation of it was not enthusiastic:

    “The [number of standard deviations] for which P = .05, or 1 in 20, is 1.96 or
    nearly 2; it is convenient to take this point as a limit in judging whether a devia-
    tion is to be considered significant or not.” 

Most people don’t think of convenience as a serious criterion. Later in his career, Fisher actively
advised against always using the same threshold for significance.

So what are you supposed to do then? There is no consensus, but thinking is always a good idea.
If you are trying to say that an interval doesn’t include some value, then you might use the widest
interval that excludes the value. Often, all confidence intervals do is communicate the shape of a
distribution. In that case, a series of nested intervals may be more useful than any one interval. For
example, why not present 67%, 89%, and 97% intervals, along with the median? Why these values?
No reason. They are prime numbers, which makes them easy to remember. But all that matters is
they be spaced enough to illustrate the shape of the posterior. And these values avoid 95%, since
conventional 95% intervals encourage many readers to conduct unconscious hypothesis tests.

 so my take away is: choose 89% to avoid 95% and to like prime numbers. :D

neerajprad:
I have no opinion on such issues :D, but here is Andrew Gelman's note on why he prefers 50% over 95% - https://statmodeling.stat.columbia.edu/2016/11/05/why-i-prefer-50-to-95-intervals/.

neerajprad:
> shouldn't 'constrain function' and the like be 'constraint function' everywhere (including in the code)

Hmm..the reason for choosing the name is that `contrain_fn` is just a function that constrains the input  to the site's support. `constraint_fn` on the other hand sounds like a function that would return a constraint object rather than doing the transformation itself. That's just my thought, and I am happy to change if `constraint_fn` sounds clear. Suggestions for other names are welcome - `transform_to_constrained` is probably better though so much more verbose.

fehiepsi:
@martinjankowiak At first, we go with `transform_fn` but it causes confusing in both code and discussions. The reason is this function applies *inverse* transforms (in the language of Stan and Pyro HMC) to get constrained values. The notion of `domain` or `codomain` also causes confusing because it does not distinguish a transform and its inverse. The name `constrain_fn` is used to say that this is a function to transform an unconstrained value to a constrained value, but it might not be a good term. Do you have other idea to avoid those confusions? How about `inv_transform` (we need to rename some args of this function if we use this name)? FYI, the current signature of `constrain_fn` is
```
def constrain_fn(inv_transforms, params, invert=False):
```

martinjankowiak:
@fehiepsi i don't feel strongly about `constrain(t)` so i defer to you two here

neerajprad:
Okay, I think all comments are addressed except for the naming of `constrain_fn` which I think can be deferred to another PR since that will need changes to the code too.

Regarding, `inv_transforms` - I think it is quite confusing as a name since it is not clear what inverse transform refers to. It seems that inverse transforms as denoting transforms from bounded domains to the real domain seems to just be a convention that we follow in Stan and PyTorch, but I suppose it could easily be the other way around? From that POV, `domain_transform` seems less confusing to me. Or maybe `support_transform`?

fehiepsi:
Yup, let's address that issue in another PR. I defer it to you if a change is needed. :)

fehiepsi:
Happy to merge this!

PR #173, opened 2019-05-30
Change README.md to reflect current status of the project

neerajprad:
Addresses #29.


neerajprad:
Reviewers - Please feel free to push to this branch directly.

neerajprad:
Given that it's `NumPy`, my vote is to go with `NumPyro`. :) I think there are inconsistencies in the remaining codebase too, and I'll fix them in a separate PR.

README.md
```diff
@@ -1,28 +1,60 @@
 # NumPyro 
 
 [![Build Status](https://travis-ci.com/pyro-ppl/numpyro.svg?branch=master)](https://travis-ci.com/pyro-ppl/numpyro)
+[![Documentation Status](https://readthedocs.org/projects/numpyro/badge/?version=latest)](https://numpyro.readthedocs.io/en/latest/?badge=latest)
 
-[Pyro](https://github.com/pyro-ppl/pyro) on Numpy. This uses 
-[JAX](https://github.com/google/jax) for autograd and JIT support. This is an 
-early stage experimental library that is under active development, and there are 
-likely to be  many changes to the API and internal classes, as the design evolves. 
+Probabilistic programming with Numpy on GPUs powered by [JAX](https://github.com/google/jax) for autograd and JIT support.
```
fehiepsi: nit: I guess you mean CPU/GPU? (and maybe TPU when JAX supports TPU on the cloud).

README.md
```diff
@@ -1,28 +1,60 @@
...
-   their models using the familiar Numpy API.
- - **Functional** - The API for the inference algorithms and other utility functions 
-   may deviate from Pyro in favor of a more *functional* style that works better 
-   with JAX. e.g. no global param store or random state.
- - **Fast** - Using JAX, we aim to aggressively JIT compile intermediate computations 
-   to XLA optimized kernels. We will evaluate JIT compilation, and benchmark runtime 
-   for Hamiltonian Monte Carlo.
-    
- ## Longer-term Plans
+NumPyro is a small probabilistic programming library built on [JAX](https://github.com/google/jax). It essentially provides a NumPy backend for [Pyro](https://github.com/pyro-ppl/pyro), with some minor changes to the inference API and syntax. Since we use JAX, we get autograd and JIT compilation to GPUs for free. This is an alpha release, and the API is subject to change as the design evolves.
```
fehiepsi: same here: I think that we need to rephrase `to GPUs`

README.md
```diff
@@ -1,28 +1,60 @@
...
+# install jax/jaxlib first for CUDA support
+pip install -e .[dev]
+```
+
+## Examples
+
+
+For some examples on specifying models and doing inference in NumPyro:
+
+ - [Bayesian Regression in Numpyro](https://github.com/pyro-ppl/numpyro/blob/master/notebooks/bayesian_regression.ipynb) - Start here to get acquainted with writing a simple model in NumPyro, MCMC inference API, effect handlers and writing custom inference utilities.
```
fehiepsi: I think that we might need to add a link using `nbviewer` because github is usually fail to render the notebook. WDYT?

README.md
```diff
@@ -1,28 +1,60 @@
...
+ 
+NumPyro is designed to be *lightweight* and focuses on providing a flexible substrate that users can build on:
+
+ - **Pyro Primitives:** NumPyro programs can contain regular Python and NumPy code, in addition to [Pyro primitives](http://pyro.ai/examples/intro_part_i.html) like `sample` and `param`. The model code should look very similar to Pyro except for some minor differences between PyTorch and Numpy's API. See [Examples](#Examples).
+ - **Inference algorithms:** NumPyro currently supports Hamiltonian Monte Carlo, including an implementation of the No U-Turn Sampler. One of the motivations for NumPyro was to speed up Hamiltonian Monte Carlo by JIT compiling the verlet integration step that includes multiple gradient computations. With JAX, we can compose `jit` and `grad` to compile the entire integration step into an XLA optimized kernel. We also eliminate Python overhead by JIT compiling the entire tree building stage in NUTS. There is also a basic Variational Inference implementation for reparameterized distributions.
```
fehiepsi: I would like to add a link to iterative NUTS at the end of "We also eliminate Python overhead by JIT compiling the entire tree building stage in NUTS (this is possible using [Iterative NUTS](link))". What do you think?

neerajprad: Sure, will add the link to the wiki. I think it will be best to publish that into a separate technical report / whitepaper, along with benchmarks soon!

README.md
neerajprad: Good idea, will change.

README.md
neerajprad: I thought it will be good to keep the summary concise, since running on CPUs is almost always supported, but I will clarify it below. I didn't want to get into TPUs because that's more a focus for the JAX team, but if we have success running NumPyro on TPUs, we should highlight that later. 😄

fehiepsi: yup, agree. I tried to use TPU in colab this morning but none of the installation methods in jax/README works. Then I realize that (also mentioned in the README) jax team is working on supporting cloud TPU. So running on TPU might not be possible at the current stage. :D

neerajprad: Clarified the summary, and I think all your comments should be addressed. :)

neerajprad:
> but you might want to make all the numpyro module references (e.g. numpyro.distributions) into links to the relevant readthedocs pages.

Good idea, just did. Documentations is a bit sparse, so I'll work on beefing that up.

neerajprad:
Should be good to merge, will work on consistent capitalization, additional docs separately.

PR #175, opened 2019-05-30
Update logistic regression notebook

fehiepsi:
This PR updates logistic regression notebook. Some changes and issues I get during the way:
+ `jit` is improved in new versions of jax/jaxlib. It is no longer necessary to use the hack `(hmc_state.update(step_size=1))` to force fast compiling. So I run the inference from beginning to end and report the result including a small compiling time.
+ update api using new distribution interface and inference utilities
+ increase num_samples of HMC to 1000 and num_samples of NUTS to 100 because it is fast in GPU
+ with new version of jaxlib, it is necessary to set `XLA_PYTHON_CLIENT_ALLOCATOR=platform`. Otherwise, after getting HMC results, running NUTS will trigger a memory error. More on this issue is discussed [upstream](https://github.com/google/jax/issues/788)
+ In CPU, for some reasons, I have to use `jaxlib=0.1.14` to make NUTS run. Otherwise, the notebook freezes. >"< I am isolating why this happens in this example, but not in other examples.


neerajprad:
> jit is improved in new versions of jax/jaxlib. It is no longer necessary to use the hack (hmc_state.update(step_size=1))

That's great to know. 

> In CPU, for some reasons, I have to use jaxlib=0.1.14 to make NUTS run.

That's a bummer. I just hope this doesn't affect too many other models too.

fehiepsi:
@neerajprad I just want to update that the above GPU and CPU issues have been resolved with the latest version of jaxlib. :)

PR #178, opened 2019-05-31
Add batching rule for some custom primitives

### LINKED ISSUE #176: Batching for custom transforms
Batching is currently not implemented for `custom_transforms` like `xlogy` and `xlog1py` so functions using these primitives will fail with `vmap`.

fehiepsi:
Resolves #176. The idea is to add singleton dims to the front of each operand, then adjust vmap dim and move it to the front.

### DIFF
--- a/test/test_distributions_util.py
+++ b/test/test_distributions_util.py
@@ -170,6 +170,42 @@ def test_standard_gamma_batch():
         assert_allclose(samples[i], standard_gamma(rngs[i], alphas[i]))
 
 
+@pytest.mark.parametrize('prim', [
+    xlogy,
+    xlog1py,
+])
+def test_binop_batch_rule(prim):
+    bx = np.array([1., 2., 3.])
+    by = np.array([2., 3., 4.])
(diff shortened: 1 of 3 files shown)


test/test_distributions_util.py
```diff
@@ -170,6 +170,42 @@ def test_standard_gamma_batch():
...
 
 
+@pytest.mark.parametrize('prim', [
+    xlogy,
+    xlog1py,
+])
+def test_binop_batch_rule(prim):
+    bx = np.array([1., 2., 3.])
+    by = np.array([2., 3., 4.])
+    x = np.array(1.)
```
neerajprad: What if this was just a python numeric type - would the test work fine?

fehiepsi: Yup, it works fine for `x = 1, y=2`. To not worry about python scalar, we just need to use `np.shape(x)` or `np.reshape(x, ...)` instead of `x.shape`, `x.reshape(...)`. :D

PR #180, opened 2019-05-31
Minor pending fixes 

neerajprad:
- Changes capitalization to `NumPyro` at all places in the code.
 - More obvious disclaimers in README.
 - Fixed a mistake in the `bayesian_regression` on instructions for running vanilla hmc.

### DIFF
--- a/notebooks/bayesian_regression.ipynb
+++ b/notebooks/bayesian_regression.ipynb
@@ -4,13 +4,13 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "# Bayesian Regression Using Numpyro\n",
+    "# Bayesian Regression Using NumPyro\n",
     "\n",
-    "In this tutorial, we will explore how to do bayesian regression in Numpyro, using a simple example adapted from Statistical Rethinking [[1](#References)]. In particular, we would like to explore the following:\n",
-    " - Write a simple model using the `sample` Numpyro primitive.\n",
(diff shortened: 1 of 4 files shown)


notebooks/bayesian_regression.ipynb
```diff
@@ -1216,14 +1216,14 @@
...
-    " - Finally, we use the [mcmc](https://numpyro.readthedocs.io/en/latest/mcmc.html#numpyro.mcmc.mcmc) function to run inference using the default `NUTS` sampler. Note that to run vanilla HMC, all you need to do is to pass `sampler='hmc'` as argument to `mcmc` instead. This is a convenience utility that does all of the following:\n",
+    " - Finally, we use the [mcmc](https://numpyro.readthedocs.io/en/latest/mcmc.html#numpyro.mcmc.mcmc) function to run inference using the default `NUTS` sampler. Note that to run vanilla HMC, all you need to do is to pass `algo='hmc'` as argument to `mcmc` instead. This is a convenience utility that does all of the following:\n",
```
fehiepsi: nit: is it `HMC` or `hmc`? I think we are missing a check to make `algo` consistent.

fehiepsi: But it is not important. We might use `algo.lower() == 'NUTS'` to make it flexible, but not related to this PR though.. :D

neerajprad: Hmm..yeah, I think we should add some validation logic actually. Will do that.

PR #186, opened 2019-06-03
Implement MVN distribution

fehiepsi:
Address #122

TODO
- [x] Add tests for mvn distribution
- [x] Add tests for lower_cholesky/positive_definite constraints
- [x] Implement biject_to(lower_cholesky) (low-priority, only useful for SVI)


setup.cfg
```diff
@@ -16,6 +16,8 @@ multi_line_output = 3
 filterwarnings = error
     ignore:numpy.ufunc size changed,:RuntimeWarning
     ignore:Using a non-tuple sequence:FutureWarning
+    ignore:numpy.linalg support is experimental:UserWarning
+    ignore:scipy.linalg support is experimental:UserWarning
```
fehiepsi: mvn triangular_solve / cholesky / eigh require numpy.linalg, so we need to filter those warnings.

numpyro/distributions/continuous.py
```diff
@@ -462,52 +465,156 @@ def log_prob(self, value):
...
+    for i in range(bL.ndim - 2):
+        permute_inv_dims += (sample_ndim + i, len(out_shape) + i)
+    M = np.transpose(M, permute_inv_dims)
+    return np.reshape(M, out_shape)
+
+
+@copy_docs_from(Distribution)
+class MultivariateNormal(Distribution):
+    arg_constraints = {'loc': constraints.real,
+                       'covariance_matrix': constraints.positive_definite,
```
neerajprad: Since this is adding lazy parameters to `arg_constraints` we should also make a change in `Distribution.__init__` to only check parameters that are there in the instances `__dict__` 

```python
if param not in self.__dict__ and isinstance(getattr(type(self), param), lazy_property):
    continue 
```

test/test_distributions.py
```diff
@@ -35,6 +35,13 @@ def __new__(cls, jax_dist, *params):
         return super(cls, T).__new__(cls, jax_dist, sp_dist, params)
 
 
+def _mvn_to_scipy(loc, cov, prec, tril):
+    jax_dist = dist.MultivariateNormal(loc, cov, prec, tril)
+    mean = device_get(jax_dist.mean)
+    cov = device_get(jax_dist.covariance_matrix)
```
neerajprad: Why do we need `device_get` before passing to `osp.multivariate_normal`. Would a `DeviceArray` not work here?

setup.cfg
neerajprad: Could you remove `@pytest.mark.filterwarnings` from `test_mcmc` since we don't need that now?

test/test_distributions.py
```diff
@@ -82,15 +90,19 @@ def __new__(cls, jax_dist, *params):
...
     T(dist.LKJCholesky, 2, 0.5, "cvine"),
     T(dist.LKJCholesky, 5, np.array([0.5, 1., 2.]), "onion"),
     T(dist.LKJCholesky, 5, np.array([0.5, 1., 2.]), "cvine"),
     T(dist.LKJCholesky, 3, np.array([[3., 0.6], [0.2, 5.]]), "onion"),
     T(dist.LKJCholesky, 3, np.array([[3., 0.6], [0.2, 5.]]), "cvine"),
+    T(dist.LogNormal, 1., 0.2),
+    T(dist.LogNormal, -1., np.array([0.5, 1.3])),
+    T(dist.LogNormal, np.array([0.5, -0.7]), np.array([[0.1, 0.4], [0.5, 0.1]])),
+    T(dist.MultivariateNormal, 0., np.array([[1., 0.5], [0.5, 1.]]), None, None),
+    T(dist.MultivariateNormal, np.array([1., 3.]), None, np.array([[1., 0.5], [0.5, 1.]]), None),
```
neerajprad: The addition of optional parameters makes these fixtures somewhat wonky - maybe it will be nice to support kwargs so that we do not need to filter for `None` in tests. Should be a separate PR though.

numpyro/distributions/continuous.py
fehiepsi: Thanks for providing an explicit code! It is very helpful! :D

test/test_distributions.py
fehiepsi: Oppss, it is legacy for debugging. It seems not necessary now.

test/test_distributions.py
fehiepsi: Yeah, it might be more explicit if we define subclasses in the test:
```
def MultivariateNormalCov(dist.MultivariateNormal):
    def __init__(self, loc, cov):
        super(dist.MultivariateNormal, self).__init__(loc, covarariance_matrix=cov)
```
Do you want me to make that change?

fehiepsi: Hmm, I made an attempt but it is a bit complicated to handle `inspect` code. Will defer to another PR where we meet this `None` pattern again.

neerajprad: Don't worry about it. This is already a big PR, and it might actually not be worth adding more complexity just for a single test case. :)

neerajprad:
Looks great! There's just one shape failure in constraints test that you might want to look at.

test/test_distributions.py
neerajprad: I think the subclassing idea might be nice along with a wrapper function like our logit / prob parametrization. Let us worry about it in a different PR though.

PR #190, opened 2019-06-04
Remove x64 in gp example

fehiepsi:
In my local machine, `x64` is not necessary. Indeed, the result is stable even when I remove `jitter`. Maybe using `LogNormal` instead of hand code transforms (so skipping logdet_jacobian terms), the inference is better? @martinjankowiak could you double check that x64 is not necessary now?

In addition, I revise `cholesky_inverse` a bit using the suggestion by @fritzo. (using `solve_triangular` with `identity` matrix instead of `inv`).

### DIFF
--- a/examples/gp.py
+++ b/examples/gp.py
@@ -1,23 +1,20 @@
-import matplotlib
-matplotlib.use('Agg')  # noqa: E402
-import matplotlib.pyplot as plt
-import jax
-
 import argparse
 
+import matplotlib
+import matplotlib.pyplot as plt
 import numpy as onp
-from jax import vmap
+
+import jax
 import jax.numpy as np
 import jax.random as random
+from jax import vmap
 
 import numpyro.distributions as dist
 from numpyro.handlers import sample
 from numpyro.hmc_util import initialize_model
 from numpyro.mcmc import mcmc
 
-from jax.config import config
-# we use double precision to minimize any possible numerical instabilities in jax linear algebra
-config.update('jax_enable_x64', True)
+matplotlib.use('Agg')  # noqa: E402
 
 """
 In this example we show how to use NUTS to sample from the posterior
--- a/examples/gp.py
+++ b/examples/gp.py
@@ -23,7 +23,7 @@ over the hyperparameters of a gaussian process.
 
 
 # squared exponential kernel with diagonal noise term
-def kernel(X, Z, var, length, noise, jitter=1.0e-5, include_noise=True):
+def kernel(X, Z, var, length, noise, jitter=1.0e-6, include_noise=True):
     deltaXsq = np.power((X[:, None] - Z) / length, 2.0)
     k = var * np.exp(-0.5 * deltaXsq)
     if include_noise:
--- a/examples/gp.py
+++ b/examples/gp.py
@@ -62,7 +62,7 @@ def predict(rng, X, Y, X_test, var, length, noise):
     k_XX = kernel(X, X, var, length, noise, include_noise=True)
     K_xx_inv = np.linalg.inv(k_XX)
     K = k_pp - np.matmul(k_pX, np.matmul(K_xx_inv, np.transpose(k_pX)))
-    sigma_noise = np.sqrt(np.diag(K)) * jax.random.normal(rng, (X_test.shape[0],))
+    sigma_noise = np.sqrt(np.clip(np.diag(K), a_min=0.)) * jax.random.normal(rng, X_test.shape[:1])
     mean = np.matmul(k_pX, np.matmul(K_xx_inv, Y))
     # we return both the mean function and a sample from the posterior predictive for the
     # given set of hyperparameters
@@ -100,7 +100,7 @@ def main(args):
                               predict(rng, X, Y, X_test, var, length, noise))(*vmap_args)
 
     mean_prediction = onp.mean(means, axis=0)
-    percentiles = onp.nanpercentile(predictions, [5.0, 95.0], axis=0)
+    percentiles = onp.percentile(predictions, [5.0, 95.0], axis=0)
 
     # make plots
     fig, ax = plt.subplots(1, 1)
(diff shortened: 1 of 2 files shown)


numpyro/distributions/util.py
```diff
@@ -391,7 +392,9 @@ def cholesky_inverse(matrix):
     # which is more numerically stable.
     # Refer to:
     # https://nbviewer.jupyter.org/gist/fehiepsi/5ef8e09e61604f10607380467eb82006#Precision-to-scale_tril
-    return np.swapaxes(np.linalg.inv(np.linalg.cholesky(matrix[..., ::-1, ::-1])[..., ::-1, ::-1]), -2, -1)
+    tril_inv = np.swapaxes(np.linalg.cholesky(matrix[..., ::-1, ::-1])[..., ::-1, ::-1], -2, -1)
+    identity = np.broadcast_to(np.identity(matrix.shape[-1]), tril_inv.shape)
+    return solve_triangular(identity, tril_inv, lower=True)
```
neerajprad: Ohh nice, didn't know this was available. Does it make things faster on the gp example?

examples/gp.py
```diff
@@ -1,23 +1,20 @@
...
+from jax import vmap
 
 import numpyro.distributions as dist
 from numpyro.handlers import sample
 from numpyro.hmc_util import initialize_model
 from numpyro.mcmc import mcmc
 
-from jax.config import config
-# we use double precision to minimize any possible numerical instabilities in jax linear algebra
-config.update('jax_enable_x64', True)
```
neerajprad: On my system too, while the posterior estimates look similar, the predictions are very off and are likely affected by lack of precision (attached image). I haven't really investigated further, but we should revert this until we understand what the issue is.

[gp_plot (2).pdf](https://github.com/pyro-ppl/numpyro/files/3254975/gp_plot.2.pdf)

numpyro/distributions/util.py
fehiepsi: this does not affect `gp` example because MVN there uses `covariance_matrix`, not `precision_matrix`.

examples/gp.py
fehiepsi: I just discover that I use an old version of `jaxlib` (0.1.14), and get a nice plot. Then when I update it to `0.1.17`, I get a plot similar to you. This might be a regression of `jaxlib`.

fehiepsi: And when I update `jax` to `0.1.36`, I get a nice plot again. >"<

fehiepsi: Maybe this [PR](https://github.com/google/jax/pull/785) is helpful for MVN because during warmup phase, potential_energy will be `nan` when something is wrong (instead of some random stuffs). It is just my assumption though. Will test on GPU too.

martinjankowiak: no idea what's going on but i think it may actually be a problem in prediction not in inference

fehiepsi: Yup, you are right. It seems that the issue is from getting some `nan` over some of the posterior samples. I just replace `percentage` with `nanpercentage` and get good plot. :D

martinjankowiak: another "solution" for single precision seems to be to increase the jitter a lot (e.g to 0.01)

fehiepsi: hah, so we have 3 solutions so far:
+ update jax
+ increase jitter
+ use `nanpercentage`

I would go for `nanpercentage` to be safe for the case posterior contains some unstable samples.

martinjankowiak: actually i think the best solution is the make the following change:

``` sigma_noise = np.sqrt(np.clip(np.diag(K), 0.0, None)) * jax.random.normal(rng, (X_test.shape[0],))```

then you can use a small jitter no problem

martinjankowiak: (the nan's are from taking the sqrt of negative reals)

fehiepsi: Yup, you are right. I can recall that phenomenon in Pyro. Despite that theoretically, `k_pp - np.matmul(k_pX, np.matmul(K_xx_inv, np.transpose(k_pX)))` will have positive diagonal part, there is no such guarantee for numerical computers. I will use your solution instead.

neerajprad:
LGTM. Is this good to merge, @martinjankowiak?

neerajprad:
The GPU gives a high speedup on this for larger N, which is good to see! For N=400, it is like 7X of CPU, but not sure how that compares with other frameworks.

fehiepsi:
Thanks @martinjankowiak pointing out the source of issue! I always have a good feeling when looking at GP results and recalling its beautiful math.

> The GPU gives a high speedup on this for larger N

It is good to know. I will make an example from Stan to compare the performance. For that N, it might be faster than GPyTorch (based on my little experience for NUTS in GPyTorch). But for larger `N`, GPyTorch will be faster (indeed, the strength of GPyTorch is for exact GP with large N and sparse GP with large number of inducing points).

fehiepsi:
I make a [gist](https://gist.github.com/fehiepsi/48e59df49cb431a5724eda8d93f3ce70) to show the benchmark in Stan. It took more than 30s for compiling and more than 5m for warmup and sampling in Stan. While total time of NumPyro is just 2m. So for such N, in my system, NumPyro GPU is 4x faster than NumPyro CPU, which is 2.5x faster than Stan (so NumPyro GPU is 10x faster than Stan CPU). That is with that fact that our kernel implementation is not optimized (in computing `deltaXsq` term).

To my surprised, the inference result is pretty well, even with this large N. I believe that doing inference for this simple model in Pyro (with N=400) will always trigger Cholesky errors.
[gp_plot.pdf](https://github.com/pyro-ppl/numpyro/files/3259996/gp_plot.pdf)

fehiepsi:
Actually, looking at the result of `N=400` again, I still can't believe that it works well without Cholesky issue. I just guess Cholesky implementation in JAX is more stable than the one in PyTorch.

neerajprad:
Great to know, I think we already have a few models as benchmarks now.

PR #191, opened 2019-06-05
correct url of the project

fehiepsi:
Currently, [PyPI](https://pypi.org/project/numpyro/) gives wrong statistics due to the url pointing to the old address. This PR fixes that issue.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -26,7 +26,7 @@ setup(
     version='0.1.0',
     description='Pyro PPL on Numpy',
     packages=find_packages(include=['numpyro', 'numpyro.*']),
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -26,7 +26,7 @@
     version='0.1.0',
     description='Pyro PPL on Numpy',
     packages=find_packages(include=['numpyro', 'numpyro.*']),
-    url='https://github.com/neerajprad/numpyro',
+    url='https://github.com/pyro-ppl/numpyro',
```
neerajprad: My bad, completely missed this!

PR #194, opened 2019-06-07
Initial support for chains

fehiepsi:
This PR makes some changes to make sure `pmap` work. In addition, I'll add a toy example so that we can discuss about a required interface in `mcmc` method.

TODO
- [x] Add a test in Travis

### DIFF
--- a/examples/hmm.py
+++ b/examples/hmm.py
@@ -1,4 +1,5 @@
 import argparse
+import time
 
 import numpy as onp
 
@@ -155,8 +156,11 @@ def main(args):
         transition_prior, emission_prior, supervised_categories,
         supervised_words, unsupervised_words,
     )
+    start = time.time()
     samples = mcmc(args.num_warmup, args.num_samples, init_params,
-                   potential_fn=potential_fn, constrain_fn=constrain_fn)
+                   potential_fn=potential_fn, constrain_fn=constrain_fn,
+                   progbar=True)
+    print('\nMCMC elapsed time:', time.time() - start)
     print_results(samples, transition_prob, emission_prob)
 
 
--- a/numpyro/distributions/constraints.py
+++ b/numpyro/distributions/constraints.py
@@ -122,8 +122,11 @@ class _Multinomial(Constraint):
 
 class _PositiveDefinite(Constraint):
     def __call__(self, x):
+        # check for symmetric
+        symmetric = np.all(np.all(x == np.swapaxes(x, -2, -1), axis=-1), axis=-1)
         # check for the smallest eigenvalue is positive
-        return np.linalg.eigh(x)[0][..., 0] > 0
+        positive = np.linalg.eigh(x)[0][..., 0] > 0
+        return symmetric & positive
 
 
 class _Real(Constraint):
--- a/numpyro/util.py
+++ b/numpyro/util.py
@@ -128,7 +128,7 @@ def fori_collect(n, body_fun, init_val, transform=identity, progbar=True, **prog
     ravel_fn = lambda x: ravel_pytree(transform(x))[0]  # noqa: E731
 
     if not progbar:
-        collection = np.zeros((n,) + init_val_flat.shape, dtype=init_val_flat.dtype)
+        collection = np.zeros((n,) + init_val_flat.shape)
 
         def _body_fn(i, vals):
             val, collection = vals
@@ -136,7 +136,14 @@ def fori_collect(n, body_fun, init_val, transform=identity, progbar=True, **prog
             collection = ops.index_update(collection, i, ravel_fn(val))
             return val, collection
 
-        _, collection = fori_loop(0, n, _body_fn, (init_val, collection))
+        # PERF: jitting the for loop may be faster on certain models or high
+        # number of samples.
+        # TODO: remove this condition when the issue is resolved
+        if progbar is None:  # NB: if progbar=None, we jit fori_loop
+            _, collection = jit(fori_loop, static_argnums=(2,))(0, n, _body_fn,
+                                                                (init_val, collection))
+        else:
+            _, collection = fori_loop(0, n, _body_fn, (init_val, collection))
     else:
         diagnostics_fn = progbar_opts.pop('diagnostics_fn', None)
         progbar_desc = progbar_opts.pop('progbar_desc', '')
@@ -150,7 +157,8 @@ def fori_collect(n, body_fun, init_val, transform=identity, progbar=True, **prog
                 if diagnostics_fn:
                     t.set_postfix_str(diagnostics_fn(val), refresh=False)
 
-        collection = np.stack(collection)
+        # XXX: jax.numpy.stack/concatenate is currently slow
+        collection = onp.stack(collection)
 
     return vmap(unravel_fn)(collection)
 
--- a/numpyro/util.py
+++ b/numpyro/util.py
@@ -136,8 +136,7 @@ def fori_collect(n, body_fun, init_val, transform=identity, progbar=True, **prog
             collection = ops.index_update(collection, i, ravel_fn(val))
             return val, collection
 
-        _, collection = jit(lax.fori_loop, static_argnums=(2,))(0, n, _body_fn,
-                                                                (init_val, collection))
+        _, collection = fori_loop(0, n, _body_fn, (init_val, collection))
     else:
         diagnostics_fn = progbar_opts.pop('diagnostics_fn', None)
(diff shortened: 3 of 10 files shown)


test/test_mcmc.py
```diff
@@ -201,3 +202,30 @@ def model(data):
...
+
+    init_kernel, sample_kernel = hmc(potential_fn, algo=algo)
+    init_params = np.array([0., -1.])
+    rngs = random.split(random.PRNGKey(0), 2)
+
+    init_kernel_pmap = pmap(lambda init_param, rng: init_kernel(
+        init_param, trajectory_length=9, num_warmup=warmup_steps, progbar=False, rng=rng))
+    init_states = init_kernel_pmap(init_params, rngs)
+
+    fori_collect_pmap = pmap(lambda hmc_state: fori_collect(num_samples, sample_kernel, hmc_state,
```
neerajprad: This is pretty cool! Based on your observation, how much slower is this as compared to a single chain?

numpyro/util.py
```diff
@@ -149,11 +148,9 @@ def _body_fn(i, vals):
                 val = body_fun(val)
                 collection.append(jit(ravel_fn)(val))
                 if diagnostics_fn:
-                    # TODO: set refresh=True when its performance issue is resolved
                     t.set_postfix_str(diagnostics_fn(val), refresh=False)
 
-        # XXX: jax.numpy.stack/concatenate is currently so slow
-        collection = onp.stack(collection)
+        collection = np.stack(collection)
```
neerajprad: +1. Is the perf disparity completely resolved?

numpyro/util.py
```diff
@@ -136,8 +136,7 @@ def _body_fn(i, vals):
             collection = ops.index_update(collection, i, ravel_fn(val))
             return val, collection
 
-        _, collection = jit(lax.fori_loop, static_argnums=(2,))(0, n, _body_fn,
-                                                                (init_val, collection))
+        _, collection = fori_loop(0, n, _body_fn, (init_val, collection))
```
neerajprad: Is this required for `pmap` or is this because there is no difference between jit and no-jit now - I remember there was a significant difference in some tests in `test_mcmc` earlier.

numpyro/util.py
fehiepsi: It is not completely resolved. It is still slow for large num_samples, but not so slow as before. So I'll revert this change. :)

test/test_mcmc.py
fehiepsi: It is a bit faster than a single chain for this test (17s vs 19s) (I guess by the same reason as we jit lax.fori_loop). It is hard to tell how the change will affect bigger models because intra-op parallelism in CPU is not useful for this simple model. We need to do benchmark for bigger models while support multi-chain in mcmc (in later PRs though).

numpyro/util.py
fehiepsi: This is required by `pmap`. I'll test if we really need 10s/8000 samples of jit(lax.fori_loop).

test/test_mcmc.py
neerajprad: > It is a bit faster than a single chain for this test (17s vs 19s)

That's strange - you mean using pmap to draw 16000 samples is faster than using the default single chain to draw 8000 samples?

numpyro/util.py
neerajprad: It will be sad if we have to take a significant hit on performance just to be compatible with pmap. I would suggest adding a comment here and in `init_kernel` like `PERF: jitting the for loop may be faster on certain models / high number of samples`, so that we can easily track any perf issues we are leaving on the table and go back and undo it if needed (while benchmarking for instance).

test/test_mcmc.py
fehiepsi: Yes. pmap just replicates jaxpr across device, so its effect is similar to jit(fori_loop) I guess. FYI, I don't know why `jit(fori_loop)` is faster than `fori_loop`.

neerajprad: Ahh..I see. So that's the thing - maybe we need to understand why `jit(fori_loop)` is faster than `fori_loop`. This doesn't exactly seem very reasonable, maybe we should ask JAX devs if this is expected.

numpyro/util.py
fehiepsi: I don't know if which test makes it hits **significant**. Dropping jit in fori_collect costs about 1s for bernoulli, categorical tests (with 20000 samples). Testing on 10000 samples of hmm gives 106s vs 101s. But the costs are larger for warmup_update. I am seeing how much it costs.

test/test_mcmc.py
fehiepsi: Haha, incidently, I just ask on Slack. :D

numpyro/util.py
fehiepsi: Hmm, actually I'm seeing different results of jit fori_loop and non-jit. So I'll spend more effort to debug what is the source of this problem.

examples/hmm.py
```diff
@@ -155,8 +156,11 @@ def main(args):
         transition_prior, emission_prior, supervised_categories,
         supervised_words, unsupervised_words,
     )
+    start = time.time()
     samples = mcmc(args.num_warmup, args.num_samples, init_params,
-                   potential_fn=potential_fn, constrain_fn=constrain_fn)
+                   potential_fn=potential_fn, constrain_fn=constrain_fn,
+                   progbar=True)
+    print('\nMCMC elapsed time:', time.time() - start)
```
fehiepsi: @neerajprad I think that `hmm` is a good example to track down performance issue because the costs in tests are just at most 2 seconds. To make it easy to address this issue, I added a condition `progbar=None` to keep the old behaviour (meaning: jit fori_loop). To my observation, with `num_warmup=500`, `num_samples=1000`, the elapsed time is 50s vs 63s. While with `num_warmup=5000`, `num_samples=10000`, the elapsed time is 115s vs 128s. This 13s is magical and consistent regardless num_warmup and num_samples (I did try `num_warmup=5/50`, `num_samples=10/100`). So I believe that something is cached with `jit(fori_loop)` but don't know which one.
(note that to make a fair benchmark, we need to disable fast math mode: `XLA_FLAGS=--xla_cpu_enable_fast_math=false`)

On the other hand, I reopen https://github.com/pyro-ppl/numpyro/issues/88 because we need to address 2-time compiling issue for `pmap`. I have a feeling that the above 13s is related to that issue so I will make an attempt to address that issue in the follow-up PR. Anyway, it is not related and things should be ready for us to discuss about the interface of `mcmc` class. Let's chat about that when you have time. :)

numpyro/distributions/constraints.py
```diff
@@ -122,8 +122,11 @@ def __call__(self, x):
 
 class _PositiveDefinite(Constraint):
     def __call__(self, x):
+        # check for symmetric
+        symmetric = np.all(np.all(x == np.swapaxes(x, -2, -1), axis=-1), axis=-1)
         # check for the smallest eigenvalue is positive
-        return np.linalg.eigh(x)[0][..., 0] > 0
+        positive = np.linalg.eigh(x)[0][..., 0] > 0
+        return symmetric & positive
```
fehiepsi: I saw some tests are passed for the newest version so I remove the marks. By default, `np.linalg.eigh` uses the lower triangular part to compute eigen values, so we need to add a check for symmetric.

examples/hmm.py
fehiepsi: It turns out (#196) that there is a huge benefit of using 1-time compiling. #196 makes the above benchmark results obsolete and only be good for reference. :)

neerajprad: @fehiepsi - If this 13 s. is mostly constant and doesn't scale with number of samples, we can remove `jit`. 

> ready for us to discuss about the interface of mcmc class.

What's the change to the interface that you were thinking of?

fehiepsi: @neerajprad I was worried that I might make some wrong adjustment, so I keep the old behaviour so you can easily try on other examples/benchmarks. :)

> What's the change to the interface that you were thinking of?

Let's discuss some of them in #196. For `mcmc`, I think that we just need 2 flags `num_chains` and `parallel=True/False`. But we need some changes in `summary` and `fori_collect`.

fehiepsi: Btw, unable to use `jit(fori_loop)` is a bug and has been fixed upstream. So we can use `jit(fori_loop)` now. Sorry for driving the implementation back and forth! Anyway, I think this PR is still good for having `pmap` tested before the next jax release. :)

PR #196, opened 2019-06-09
Merge warmup update and sample kernel

fehiepsi:
This PR addresses #88 by merging warmup update to sample_kernel function. Here are benefits and trade-offs for following this way.

### benefit
+ **main**: allow running mcmc from warmup to sample in 1 single fori_loop, so we can avoid 2 time compiling issue.
+ **minor**: allow using `fori_collect` to collect samples from warmup phase with the same code

### trade-off
+ because we merge warmup_update and sample_kernel, the jaxpr for `sample_kernel` will be longer, hence affect compiling time

### benchmark
I use `hmm` example for benchmark, with `num_warmup=10000` and `num_samples=10000`. The following report is for current vs previous
+ `progbar=True`: 136s vs 139.5s
+ `progbar=None` (jit fori_loop): 131s vs 138s
+ `progbar=False`: 145s vs 153s
+ use `_mcmc2` (1 time compiling): 117s

So there is a regression in this PR when `progbar=True` (due to the above trade-off, or maybe because HMCState is more complicated so it takes a bit of efforts for unravel work or get_diagnostic_str work).

TODO
- [x] see why mean_accept_prob of sampling phase is different from previous version.

Other related points to discuss:
+ should we move the logic of `mcmc2` to `fori_collect` by introduction a `start_to_collect_idx` arg so we can keep the interface clean?
+ how `summary` handles 1 chain and multi-chains? should we always enforce first dimension is chain dimension, even for 1 chain?

### DIFF
--- a/numpyro/mcmc.py
+++ b/numpyro/mcmc.py
@@ -239,10 +239,11 @@ def hmc(potential_fn, kinetic_fn=None, algo='NUTS'):
                 # number of samples.
                 # TODO: remove this condition when the issue is resolved
                 if progbar is None:  # NB: if progbar=None, we jit fori_loop
-                    hmc_state = jit(fori_loop, static_argnums=(2,))(0, num_warmup,
-                                                                    sample_kernel, hmc_state)
+                    hmc_state = jit(fori_loop, static_argnums=(2,))(
+                        0, num_warmup, lambda *args: sample_kernel(args[1]), hmc_state)
                 else:
-                    hmc_state = fori_loop(0, num_warmup, sample_kernel, hmc_state)
+                    hmc_state = fori_loop(
+                        0, num_warmup, lambda *args: sample_kernel(args[1]), hmc_state)
             else:
                 with tqdm.trange(num_warmup, desc='warmup') as t:
                     for i in t:
--- a/numpyro/mcmc.py
+++ b/numpyro/mcmc.py
@@ -237,7 +237,7 @@ def hmc(potential_fn, kinetic_fn=None, algo='NUTS'):
         hmc_state = HMCState(0, vv_state.z, vv_state.z_grad, vv_state.potential_energy, 0, 0., 0.,
                              wa_state, rng_hmc)
 
-        if run_warmup:
+        if run_warmup and num_warmup > 0:
             # JIT if progress bar updates not required
             if not progbar:
                 # TODO: keep jit version and remove non-jit version for the next jax release
--- a/numpyro/mcmc.py
+++ b/numpyro/mcmc.py
@@ -236,7 +236,14 @@ def hmc(potential_fn, kinetic_fn=None, algo='NUTS'):
         if run_warmup:
             # JIT if progress bar updates not required
             if not progbar:
-                hmc_state, _ = fori_loop(0, num_warmup, warmup_update, (hmc_state, wa_state))
+                # PERF: jitting the for loop may be faster on certain models or high
+                # number of samples.
+                # TODO: remove this condition when the issue is resolved
+                if progbar is None:  # NB: if progbar=None, we jit fori_loop
+                    hmc_state, _ = jit(fori_loop, static_argnums=(2,))(0, num_warmup, warmup_update,
+                                                                       (hmc_state, wa_state))
+                else:
+                    hmc_state, _ = fori_loop(0, num_warmup, warmup_update, (hmc_state, wa_state))
             else:
                 with tqdm.trange(num_warmup, desc='warmup') as t:
                     for i in t:
--- a/numpyro/mcmc.py
+++ b/numpyro/mcmc.py
@@ -5,7 +5,7 @@ from collections import namedtuple
 import tqdm
 
 import jax.numpy as np
-from jax import jit, ops, partial, random, vmap
+from jax import jit, partial, random
 from jax.flatten_util import ravel_pytree
 from jax.random import PRNGKey
 from jax.tree_util import register_pytree_node
@@ -29,10 +29,15 @@ A :func:`~collections.namedtuple` consisting of the following fields:
    does not correspond to the proposal if it is rejected.
  - **mean_accept_prob** - Mean acceptance probability until current iteration
    during warmup adaptation or sampling (for diagnostics).
- - **step_size** - Step size to be used by the integrator in the next iteration.
-   This is adapted during warmup.
- - **inverse_mass_matrix** - The inverse mass matrix to be be used for the next
-   iteration. This is adapted during warmup.
+ - **adapt_state** - A ``AdaptState`` namedtuple which contains adaptation information
+   during warmup:
+
+   + **step_size** - Step size to be used by the integrator in the next iteration.
+   + **inverse_mass_matrix** - The inverse mass matrix to be used for the next
+     iteration.
+   + **mass_matrix_sqrt** - The square root of mass matrix to be used for the next
+     iteration.
+
  - **rng** - random number generator seed used for the iteration.
 """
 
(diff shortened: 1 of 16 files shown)


numpyro/mcmc.py
```diff
@@ -229,36 +230,26 @@ def init_kernel(init_params,
...
+                        0, num_warmup, lambda *args: sample_kernel(args[1]), hmc_state)
             else:
                 with tqdm.trange(num_warmup, desc='warmup') as t:
                     for i in t:
-                        hmc_state, wa_state = warmup_update(i, (hmc_state, wa_state))
-                        # TODO: set refresh=True when its performance issue is resolved
+                        hmc_state = sample_kernel(hmc_state)
                         t.set_postfix_str(get_diagnostics_str(hmc_state), refresh=False)
-            # Reset `i` and `mean_accept_prob` for fresh diagnostics.
-            hmc_state.update(i=0, mean_accept_prob=0)
```
fehiepsi: this is a bug of previous code, where update takes no effect.

neerajprad:
> So there is a regression in this PR when progbar=True (due to the above trade-off, or maybe because HMCState is more complicated so it takes a bit of efforts for unravel work or get_diagnostic_str work).

The numbers that you posted above don't show a regression - 136 (current) vs 139 (previous).

Is this PR related to the `pmap` issue or is this purely meant to handle #88?

neerajprad:
Okay, just saw your comment on #194.

fehiepsi:
> 136 (current) vs 139 (previous)

Oppss... my mistake! The order is reversed. Basically, 2-time compiling or using `progbar=True` is a bit slower. But it turns out that 1 time compiling is good. :)

numpyro/mcmc.py
```diff
@@ -15,8 +15,7 @@
 from numpyro.util import cond, fori_collect, fori_loop, identity
 
 HMCState = namedtuple('HMCState', ['i', 'z', 'z_grad', 'potential_energy', 'num_steps', 'accept_prob',
-                                   'mean_accept_prob', 'step_size', 'inverse_mass_matrix', 'mass_matrix_sqrt',
-                                   'rng'])
+                                   'mean_accept_prob', 'adapt_state', 'rng'])
```
fehiepsi: TODO: change docs for `adapt_state`

numpyro/mcmc.py
```diff
@@ -409,3 +406,39 @@ def mcmc(num_warmup, num_samples, init_params, sampler='hmc',
...
+
+    def _body_fn(i, vals):
+        val, collection = vals
+        val = body_fun(val)
+        i = np.where(i >= num_warmup, i - num_warmup, 0)
+        collection = ops.index_update(collection, i, ravel_fn(val))
+        return val, collection
+
+    n = num_warmup + num_samples
+    _, collection = jit(fori_loop, static_argnums=(2,))(0, n, _body_fn,
```
fehiepsi: TODO: add a note to remove `jit` for `pmap` and add it again with the next jax release.

neerajprad:
> Is this PR related to the pmap issue or is this purely meant to handle #88?

You mentioned that the pmap jit issue is resolved, so IIUC, this PR is now only meant to take care of the perf issue highlighted in #88?

neerajprad:
> should we move the logic of mcmc2 to fori_collect by introduction a start_to_collect_idx arg so we can keep the interface clean?

Are there any blockers to merging the mcmc2 logic into the mcmc class itself?

> how summary handles 1 chain and multi-chains? should we always enforce first dimension is chain dimension, even for 1 chain?

I think it might be reasonable to do that for the intermediate data dict, but we should return the samples where the leading dim is the batch shape for the default case of `num_chains=1` at least. Maybe we can add another default argument `flatten_chains=True` so that we use the non-flattened one for computing summary statistics but return the flattened chains by default (i.e. unless `flatten_chains=False` which the user might want if they would like to do something with the chain information).

fehiepsi:
> this PR is now only meant to take care of the perf issue highlighted in #88

Yup, we won't have to worry about the jit fori_loop issue with the next release of jax.

> Are there any blockers to merging the mcmc2 logic into the mcmc class itself?

It would be nice to use `fori_collect` in `mcmc`, but an additional arg is required. Howabout we use the same signature as [lax.fori_loop](https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.fori_loop.html#jax.lax.fori_loop) which `lower` mean and `upper` means for the interval to collect? It can be useful for other mcmc algorithms too (to not collect burn-in samples).

> I think it might be reasonable to do that for the intermediate data dict, but we should return the samples where the leading dim is the batch shape for the default case of num_chains=1 at least.

Thanks! That's a reasonable way to go.

> Maybe we can add another default argument flatten_chains=True so that we use the non-flattened one for computing summary statistics but return the flattened chains by default (i.e. unless flatten_chains=False which the user might want if they would like to do something with the chain information).

It is up to you to have this flag. I also think that it is good to have, so we don't need flatten code for using samples to make prediction. :)

neerajprad:
> Howabout we use the same signature as lax.fori_loop which lower and upper are the interval to collect (in other words, we skip the first lower samples?

Yes, I think that's a nice interface to have for `fori_collect` with lower (inclusive) and upper (exclusive).

test/test_mcmc.py
```diff
@@ -125,19 +125,15 @@ def model(data):
...
-                            dense_mass=dense_mass)
-    hmc_states = fori_collect(num_samples, sample_kernel, hmc_state,
-                              transform=lambda x: constrain_fn(x.z),
-                              progbar=False)
-    assert_allclose(np.mean(hmc_states['p_latent'], 0), true_probs, atol=0.02)
+    # TODO: having progbar=None just to test if that value works,
+    # change it to False when jit fori_loop issue is resolved
+    samples = mcmc(warmup_steps, num_samples, init_params, constrain_fn=constrain_fn, progbar=None,
+                   print_summary=False, potential_fn=potential_fn, algo=algo, trajectory_length=1.,
+                   dense_mass=dense_mass)
```
fehiepsi: @neerajprad This test exercises the new change for `mcmc` with `progbar=None` (or `False`). In my system, with dense_mass=False, algo=NUTS, this gives us 15s while the fastest speed in some recent versions are 20s. So I think the change in this PR is good to go because it works well for both of this fast-to-compile test and slow-to-compile hmm example. But we need a bit trade-off when `progbar=True`.

numpyro/mcmc.py
```diff
@@ -30,10 +29,15 @@
...
-   This is adapted during warmup.
- - **inverse_mass_matrix** - The inverse mass matrix to be be used for the next
-   iteration. This is adapted during warmup.
+ - **adapt_state** - A ``AdaptState`` namedtuple which contains adaptation information
+   during warmup:
+
+   + **step_size** - Step size to be used by the integrator in the next iteration.
+   + **inverse_mass_matrix** - The inverse mass matrix to be used for the next
+     iteration.
+   + **mass_matrix_sqrt** - The square root of mass matrix to be used for the next
```
neerajprad: Thanks for adding this argument - it might be worth mentioning that this refers to the cholesky factorization for dense mass matrix.

numpyro/mcmc.py
neerajprad: I think this is done now.

neerajprad:
This was a really nice refactor!

numpyro/mcmc.py
```diff
@@ -304,16 +294,26 @@ def sample_kernel(hmc_state):
...
-        r = momentum_generator(hmc_state.mass_matrix_sqrt, rng_momentum)
+        r = momentum_generator(hmc_state.adapt_state.mass_matrix_sqrt, rng_momentum)
         vv_state = IntegratorState(hmc_state.z, r, hmc_state.potential_energy, hmc_state.z_grad)
-        vv_state, num_steps, accept_prob = _next(hmc_state.step_size,
-                                                 hmc_state.inverse_mass_matrix,
+        vv_state, num_steps, accept_prob = _next(hmc_state.adapt_state.step_size,
+                                                 hmc_state.adapt_state.inverse_mass_matrix,
                                                  vv_state, rng_transition)
+        # not update adapt_state after warmup phase
+        adapt_state = cond(hmc_state.i < wa_steps,
```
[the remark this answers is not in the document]
fehiepsi: Yes, it evaluate 2 branches at compiling time but not when running.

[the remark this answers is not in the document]
neerajprad: That's great, works for us.

PR #198, opened 2019-06-10
Support chains to mcmc

fehiepsi:
I am excited to see that `pmap` works natively with our mcmc interface. This PR supports `num_chains` arg to `mcmc` and add a test for it.

I have not tested if `num_chains` works for our examples yet. Let's defer it to later PRs (add `num_chains` to parse args) where we can report or fix related issues (e.g. one thing I catch is batching rule for `standard_gamma` does not work in case `key` is batched but `alpha` is not).


numpyro/hmc_util.py
```diff
@@ -741,20 +742,33 @@ def initialize_model(rng, model, *model_args, init_strategy='uniform', **model_k
...
-    prior_params = constrain_fn(inv_transforms,
-                                {k: v['value'] for k, v in sample_sites.items()}, invert=True)
-    if init_strategy == 'uniform':
-        init_params = {}
-        for k, v in prior_params.items():
-            rng, = random.split(rng, 1)
-            init_params[k] = random.uniform(rng, shape=np.shape(v), minval=-2, maxval=2)
-    elif init_strategy == 'prior':
-        init_params = prior_params
+    def initialize_one(key, only_params=False):
```
neerajprad: nit: would suggest renaming this to `single_chain_init` or `initialize_single_chain` which will be clearer at first glance.

numpyro/hmc_util.py
```diff
@@ -726,7 +726,8 @@ def initialize_model(rng, model, *model_args, init_strategy='uniform', **model_k
     within their respective support.
 
     :param jax.random.PRNGKey rng: random number generator seed to
-        sample from the prior.
+        sample from the prior. The returned `init_params` will have the
```
neerajprad: Would suggest adding a test for this parametrized on `init_strategy` to make sure that we end up returning distinct initial points for either strategy.

examples/gp.py
```diff
@@ -123,6 +125,7 @@ def main(args):
     parser = argparse.ArgumentParser(description="Gaussian Process example")
     parser.add_argument("-n", "--num-samples", nargs="?", default=1000, type=int)
     parser.add_argument("--num-warmup", nargs='?', default=1000, type=int)
+    parser.add_argument("--num-chains", nargs='?', default=1, type=int)
```
neerajprad: +1. I'm kind of surprised that all of these examples run with multiple chains with relatively little change to the codebase. How does the runtime compare now for 1 vs 4 chains for instance?

numpyro/mcmc.py
```diff
@@ -390,29 +392,61 @@ def mcmc(num_warmup, num_samples, init_params, sampler='hmc',
...
 
         init_kernel, sample_kernel = hmc(potential_fn, kinetic_fn, algo)
         if progbar:
             hmc_state = init_kernel(init_params, num_warmup, progbar=progbar, **sampler_kwargs)
-            samples = fori_collect(0, num_samples, sample_kernel, hmc_state,
-                                   transform=lambda x: constrain_fn(x.z),
-                                   progbar=progbar,
-                                   diagnostics_fn=get_diagnostics_str,
-                                   progbar_desc='sample')
+            flatten_samples = fori_collect(0, num_samples, sample_kernel, hmc_state,
```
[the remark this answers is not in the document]
fehiepsi: yes, I'll make the change

numpyro/hmc_util.py
fehiepsi: Nice suggestion! We want to make sure that we get a fixed init_params for each random seed.

examples/gp.py
fehiepsi: FYI, I compare 1 (with the fastest variant progbar=None) vs 4 chains vs 12 chains

+ BNN example: 52s vs 56s vs 58s

+ GP example: 17s vs 20s vs 29s

To be fair, I should test with the slowest single chain and remove the `print_summary` process. But I hope these numbers are enough to get an overall idea.

fehiepsi: In GP example, maybe intra-op parallelism play an important role. :)

neerajprad: The numbers look great to me. By forcing the device count to 2, are we disabling intra-op parallelism? If that's the case, maybe we are getting good results on these examples due to small data size, and something like covtype will be much slower, in comparison.

fehiepsi: Yeah, let me test that example too.

fehiepsi: + Covtype example (HMC, num_steps=10, num_samples=100): 30s vs 34s vs 44s.

Somehow, the pattern is similar to GP. They are 1.5-2x slower in 12 cores, but consider the benefit that we get 12x more samples, that is a good trade-off.

Testing in covtype example, I'm observing that we get different summary results for 1 chain vs 4 chains. That may be just a random seed issue or a fast-math-mode issue. I'll investigate why but it does not related to this PR.

neerajprad: That looks great, actually! I am quite surprised that there's that much headroom by including 12 cores since I would have thought that on covtype all the cpu cores would be utilized already. Will play around with it to see what's happening later.

PR #200, opened 2019-06-10
Support batching rule for gamma

fehiepsi:
This PR covers other usage cases of gamma batching rule:
+ key is batched, alpha is not batched
+ key is not batched, alpha is batched

Note that the previous version only works for batched key and batched alpha.

In addition, because `gammaln` supports int dtype, we don't need to use `lax.convert_element_type` at various places.

### DIFF
--- a/numpyro/distributions/util.py
+++ b/numpyro/distributions/util.py
@@ -174,6 +174,16 @@ def _standard_gamma_grad(sample, alpha):
     return grads.reshape(alpha.shape)
 
 
+def _standard_gamma_batching_rule(batched_args, batch_dims):
+    x, y = batched_args
+    bx, by = batch_dims
+    size = next(t.shape[i] for t, i in zip(batched_args, batch_dims)
+                if i is not None)
+    x = batching.bdim_at_front(x, bx, size, force_broadcast=True)
+    y = batching.bdim_at_front(y, by, size, force_broadcast=True)
+    return _standard_gamma_p(x, y), 0
+
+
 @custom_transforms
 def _standard_gamma_p(key, alpha):
     return _standard_gamma_impl(key, alpha)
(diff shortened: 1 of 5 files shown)


numpyro/distributions/util.py
```diff
@@ -174,14 +174,24 @@ def _standard_gamma_grad(sample, alpha):
     return grads.reshape(alpha.shape)
 
 
+def _standard_gamma_batching_rule(batched_args, batch_dims):
+    x, y = batched_args
+    bx, by = batch_dims
+    size = next(t.shape[i] for t, i in zip(batched_args, batch_dims)
```
neerajprad: This seems to only use information from `x` to decide the size. Do we not need to broadcast with `y[i]`?

numpyro/distributions/util.py
```diff
@@ -174,14 +174,24 @@ def _standard_gamma_grad(sample, alpha):
...
+
 @custom_transforms
 def _standard_gamma_p(key, alpha):
     return _standard_gamma_impl(key, alpha)
 
 
 ad.defjvp2(_standard_gamma_p.primitive, None,
            lambda tangent, sample, key, alpha, **kwargs: tangent * _standard_gamma_grad(sample, alpha))
-batching.defvectorized(_standard_gamma_p.primitive)
+batching.primitive_batchers[_standard_gamma_p.primitive] = _standard_gamma_batching_rule
```
neerajprad: I am guessing this is also an issue with other samplers like `binomial`, `poisson` etc.? Are there JAX samplers which require custom batching implementations too?

numpyro/distributions/util.py
fehiepsi: It is required by vmap that batching sizes must be equal for vmap args, so we dont need to check the batch size of y if bx is not None.

numpyro/distributions/util.py
fehiepsi: I have not tested through those samplers. But vmap works for all of our examples. So I think we don’t need to do so for non-custom-transform samplers.

fehiepsi:
> why we don't need to do this for other samplers

I think that we just need to do this for custom primitives which we need to define things such as jvprule. Previously, we do it through `def_primitive, jaxpr, partial_eval...` stuffs but recently, we use `custom_transform` wrapper. Currently, only gamma sampler, xlogy, xlog1py, cumsum, cumprod requires this wrapper so we only need to define batching rule for these utilities.

Though it works for other samplers, there might be some bugs available. Let's try to fix if we observe any of them. :)

Issue #201, opened 2019-06-11
Address API changes in jax 0.1.37


fehiepsi:
JAX 0.1.37 comes up with a revised `custom_transform` which breaks our current implementation. More on this soon.

PR #202, opened 2019-06-12
Fix issues with new version of JAX

fehiepsi:
Addresses #201. New version of custom_transforms simplifies our implementation in that we don't need to define batching_rule for various operator. But it introduced an issue at `cumprod` because batching rule for 'reduce_window' (used by np.cumprod) is not implemented.

In addition, the bug `jit(jit(fori_loop))` is fixed with the new version, so this PR drops the case `progbar=None` and always `jit(fori_loop)`.

Other PRs will be blocked until this is merged.

**Note:** What I have not tested is if `jit(fori_loop)` affects `pmap`. And will address it in  #198.

### DIFF
--- a/numpyro/distributions/util.py
+++ b/numpyro/distributions/util.py
@@ -417,7 +417,7 @@ def cumsum(x):
     return np.cumsum(x, axis=-1)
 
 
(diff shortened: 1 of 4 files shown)


fehiepsi:
Yeah, I need to fix the bug at `cumsum` too. >"<

PR #203, opened 2019-06-13
masking code for autoregressive nn

martinjankowiak:
as a first step towards building normalizing flows, we add logic for building masks needed for MADE-like NNs.

it's probably not necessary to check the details of the mask building logic, since this code is lightly adapted from (highly tested) pyro code. however, i'm not really aware of good practices for dtype stuff, so please pay attention to that. these masks need to be real-valued so that they can be used as they are [here](https://github.com/pyro-ppl/pyro/blob/8d017d8b935fb8a83dec9261806d35606c259bcc/pyro/nn/auto_reg_nn.py#L75).

(the next step will be to build such a `MaskedLinear` object in stax so any advice there is appreciated)

### DIFF
--- a/numpyro/contrib/nn/auto_reg_nn.py
+++ b/numpyro/contrib/nn/auto_reg_nn.py
@@ -1,0 +1,58 @@
+# lightly adapted from https://github.com/pyro-ppl/pyro/blob/dev/pyro/nn/auto_reg_nn.py
+
+from __future__ import absolute_import, division, print_function
+
+import numpy as onp
+
+
+def sample_mask_indices(input_dim, hidden_dim):
+    """
+    Samples the indices assigned to hidden units during the construction of MADE masks
+
+    :param input_dim: the dimensionality of the input variable
+    :type input_dim: int
+    :param hidden_dim: the dimensionality of the hidden layer
+    :type hidden_dim: int
+    """
+    indices = onp.linspace(1, input_dim, num=hidden_dim)
+    # Simple procedure tries to space fractional indices evenly by rounding to nearest int
+    return onp.round(indices)
+
+
+def create_mask(input_dim, hidden_dims, permutation, output_dim_multiplier):
+    """
+    Creates (non-conditional) MADE masks
+
+    :param input_dim: the dimensionality of the input variable
+    :type input_dim: int
+    :param hidden_dims: the dimensionality of the hidden layers(s)
+    :type hidden_dims: list[int]
+    :param permutation: the order of the input variables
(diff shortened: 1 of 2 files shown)


fehiepsi:
Hi @martinjankowiak, thanks for making this first step! It is going to be really cool stuffs. :) I'll take this chance to learn about normalizing flow too. My only concern is you are using `onp` instead of jax numpy. If the masks are constant for each inference step then it would be fine. However, if the masks are changed randomly, then we have to make sure that `jitting` works. I think that masks will be changed based on `permutation` so I need to verify that jitting works (operators which depending on input_dim and hidden_dims would be fine btw). I'll review carefully and comment about related issue soon.

About dtype/device, with JAX, you mostly don't have to worry them. If x64 is enabled, dtypes will be int64 and float64. If x32, then they will be int32 and float32.

I'll take a look at `MaskedLinear` too after making sure that jitting works.

martinjankowiak:
all the masks/permutations are fixed once at the beginning and then stay constant throughout. that's why i used `onp`

fehiepsi:
About MaskedLinear, how about just mimic `stax.Dense`? You can add a logic for `bias=False` if you want.

```
from jax.experimental.stax import glorot, randn

# Each layer constructor function returns an (init_fun, apply_fun) pair, where
#   init_fun: takes an rng key and an input shape and returns an
#     (output_shape, params) pair,
#   apply_fun: takes params, inputs, and an rng key and applies the layer.

def MaskLinear(out_dim, mask, W_init=glorot(), b_init=randn()):
    def init_fun(rng, input_shape):
        output_shape = input_shape[:-1] + (out_dim,)
        k1, k2 = random.split(rng)
        W, b = W_init(k1, (input_shape[-1], out_dim)), b_init(k2, (out_dim,))
        return output_shape, (W, b)

    def apply_fun(params, inputs, **kwargs):
        W, b = params
        return np.dot(inputs, W * mask) + b

    return init_fun, apply_fun
```

martinjankowiak:
sampling masks and the like can make sense in certain contexts but not in the HMC context where we want a flow whose density is precisely computable pointwise

martinjankowiak:
yeah, thanks, i think `MaskedLinear` should be that easy

martinjankowiak:
actually this mask creation stuff is the trickiest code of the whole business

fehiepsi:
> sampling masks and the like can make sense in certain contexts but not in the HMC context where we want a flow whose density is precisely computable pointwise

That's the idea! Thanks!!

PR #205, opened 2019-06-14
Pass in split RNGs to multiple chains

neerajprad:
Minor fix to ensure that chains are getting distinct random number generator seeds.

### DIFF
--- a/examples/gp.py
+++ b/examples/gp.py
@@ -54,7 +54,7 @@ def run_inference(model, args, rng, X, Y):
     init_params, potential_fn, constrain_fn = initialize_model(rng, model, X, Y)
     start = time.time()
     samples = mcmc(args.num_warmup, args.num_samples, init_params, num_chains=args.num_chains,
-                   sampler='hmc', potential_fn=potential_fn, constrain_fn=constrain_fn, progbar=None)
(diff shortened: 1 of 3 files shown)


examples/gp.py
```diff
@@ -54,7 +54,7 @@ def run_inference(model, args, rng, X, Y):
     init_params, potential_fn, constrain_fn = initialize_model(rng, model, X, Y)
     start = time.time()
     samples = mcmc(args.num_warmup, args.num_samples, init_params, num_chains=args.num_chains,
-                   sampler='hmc', potential_fn=potential_fn, constrain_fn=constrain_fn, progbar=None)
+                   sampler='hmc', potential_fn=potential_fn, constrain_fn=constrain_fn)
```
neerajprad: @fehiepsi - Since your change disables progress bar with pmap anyways, I have kept the progress bar for the default case so that it is easy to track progress and users aren't confused. e.g. on the GPU, this will be slow for now, and it will be nice to see progress, otherwise users might think that it isn't making any progress.

fehiepsi: Thanks @neerajprad ! It is a legacy flag which I used to benchmark result.

PR #208, opened 2019-06-15
implement scale messenger

fehiepsi:
This messenger is used to modify log density of sub-posteriors (rather reimplementing `potential_energy`, `initialize_model` for them),... One thing I concerned about is whether the logic at transform log_det is right, so I need your opinions about it.

### DIFF
--- a/numpyro/hmc_util.py
+++ b/numpyro/hmc_util.py
@@ -696,16 +696,22 @@ def log_density(model, model_args, model_kwargs, params):
     log_joint = 0.
     for site in model_trace.values():
         if site['type'] == 'sample':
-            log_joint = log_joint + np.sum(site['fn'].log_prob(site['value']))
+            log_prob = np.sum(site['fn'].log_prob(site['value']))
+            if 'scale' in site:
+                log_prob = site['scale'] * log_prob
+            log_joint = log_joint + log_prob
     return log_joint, model_trace
 
 
 def potential_energy(model, model_args, model_kwargs, inv_transforms):
     def _potential_energy(params):
         params_constrained = constrain_fn(inv_transforms, params)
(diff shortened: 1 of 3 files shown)


numpyro/hmc_util.py
```diff
@@ -696,16 +696,22 @@ def log_density(model, model_args, model_kwargs, params):
...
 def potential_energy(model, model_args, model_kwargs, inv_transforms):
     def _potential_energy(params):
         params_constrained = constrain_fn(inv_transforms, params)
-        log_joint = jax.partial(log_density, model, model_args, model_kwargs)(params_constrained)[0]
+        log_joint, model_trace = log_density(model, model_args, model_kwargs, params_constrained)
         for name, t in inv_transforms.items():
-            log_joint = log_joint + np.sum(t.log_abs_det_jacobian(params[name], params_constrained[name]))
+            t_log_det = np.sum(t.log_abs_det_jacobian(params[name], params_constrained[name]))
+            if 'scale' in model_trace[name]:
+                t_log_det = model_trace[name]['scale'] * t_log_det
```
fehiepsi: do we need to scale `log_abs_det_jacobian`?

martinjankowiak: this looks right. although i guess you're assuming scale is a float?

fehiepsi: Yes. Do you know other usage cases, so I can address it or add a note for future reconsideration?

martinjankowiak: this is fine. i think in the main repo we allow tensorial scaling through mask. you should probably explicitly say that scale is positive real-valued in the docstring and check that it's positive in the code

neerajprad:
Feel free to merge pending any further comments from @martinjankowiak.

PR #210, opened 2019-06-18
add made-like autoregressive nn

martinjankowiak:


numpyro/contrib/nn/auto_reg_nn.py
```diff
@@ -57,3 +60,117 @@ def create_mask(input_dim, hidden_dims, permutation, output_dim_multiplier):
...
+    :param nonlinearity: The nonlinearity to use in the feedforward network such as ReLU. Note that no
+        nonlinearity is applied to the final network output, so the output is an unbounded real number.
+    :type nonlinearity: callable. defaults to ReLU.
+
+    Reference:
+
+    MADE: Masked Autoencoder for Distribution Estimation [arXiv:1502.03509]
+    Mathieu Germain, Karol Gregor, Iain Murray, Hugo Larochelle
+    """
+    def __init__(self, input_dim, hidden_dims, param_dims=[1, 1], nonlinearity=relu):
```
martinjankowiak: doing everything in init that doesn't depend on the rng

numpyro/contrib/nn/auto_reg_nn.py
```diff
@@ -57,3 +60,117 @@ def create_mask(input_dim, hidden_dims, permutation, output_dim_multiplier):
...
+        :param rng: rng used to initialize parameters
+        :param input_shape: input shape
+        :param permutation: an optional permutation that is applied to the inputs and controls the order of the
+            autoregressive factorization. in particular for the identity permutation the autoregressive structure
+            is such that the Jacobian is upper triangular. By default this is chosen at random.
+        :type permutation: array of ints
+        """
+        if permutation is None:
+            # By default set a random permutation of variables, which is important for performance with multiple steps
+            rng, rng_perm = random.split(rng)
```
martinjankowiak: please check my rng usage

numpyro/contrib/nn/auto_reg_nn.py
```diff
@@ -57,3 +60,117 @@ def create_mask(input_dim, hidden_dims, permutation, output_dim_multiplier):
...
+            self.permutation = onp.array(random.shuffle(rng_perm, onp.arange(self.input_dim)))
```
neerajprad: Is there a reason to use `onp` instead of `np`? It also looks like we are converting the DeviceArray to `onp.array` here.

numpyro/contrib/nn/auto_reg_nn.py
```diff
@@ -57,3 +60,117 @@ def create_mask(input_dim, hidden_dims, permutation, output_dim_multiplier):
...
+            mask_apply = mask[1]
+            out = mask_apply(param, out)
+            if k < len(self.masked_layers) - 1:
+                out = self.nonlinearity(out)
+
+        # reshape output as necessary
+        if self.output_multiplier == 1:
+            return out
+        else:
+            out = out.reshape(list(inputs.shape[:-1]) + [self.output_multiplier, self.input_dim])
```
neerajprad: Would suggest using `np.reshape(out, ..)` instead which will be jittable.

numpyro/contrib/nn/auto_reg_nn.py
neerajprad: Looks good to me.

numpyro/contrib/nn/auto_reg_nn.py
```diff
@@ -57,3 +60,117 @@ def create_mask(input_dim, hidden_dims, permutation, output_dim_multiplier):
...
+        return input_shape, init_params
+
+    def apply_fun(self, params, inputs, **kwargs):
+        """
+        :param params: layer parameters
+        :param inputs: layer inputs
+        """
+        out = inputs
+
+        for k, (mask, param) in enumerate(zip(self.masked_layers, params)):
```
neerajprad: I think this might need a few modifications to make it jittable, like using jax control flow primitives - using `fori_loop` instead of native for loop which will be faster. I think the if-else blocks may work fine because they are static, i.e. won't rely on traced values. It might be worth adding a `jit` arg to the test to see if this works. I'm also okay with addressing this in a separate PR instead.

numpyro/contrib/nn/auto_reg_nn.py
martinjankowiak: if the outer `onp.array` is not there i get errors like 
```
>       var_index[permutation] = onp.arange(input_dim)
E       IndexError: too many indices for array

numpyro/contrib/nn/auto_reg_nn.py:40: IndexError
```
not sure what's going on

numpyro/contrib/nn/auto_reg_nn.py
martinjankowiak: let's worry about jit elsewhere

numpyro/contrib/nn/auto_reg_nn.py
martinjankowiak: done

numpyro/contrib/nn/auto_reg_nn.py
neerajprad: Hmm..I think the reason might be that `var_index` itself is `onp.array` which doesn't work well with a JAX DeviceArray for some reason. Anyways, I think fixing this might require a few other changes outside this PR, so best to handle this separately.

PR #211, opened 2019-06-18
Automatically transform constrained parameters in SVI

### LINKED ISSUE #187: Optimize parameters with constraints in SVI
Currently, our very basic SVI implementation would require users to transform the parameters to unconstrained space before handing it to SVI. Like HMC, let us do this automatically behind the scenes or provide utilities so as to avoid having to write boiler plate code for this.

neerajprad:
Fixes #187.

### DIFF
--- a/examples/vae.py
+++ b/examples/vae.py
@@ -113,7 +113,7 @@ def main(args):
             loss_sum, rng = val
             rng, = random.split(rng, 1)
             rng, batch = binarize(rng, test_fetch(i, test_idx)[0])
-            loss = jit(svi_eval)(rng, opt_state, (batch,), (batch,)) / len(batch)
+            loss = svi_eval(rng, opt_state, (batch,), (batch,)) / len(batch)
             loss_sum += loss
             return loss_sum, rng
 
--- a/examples/vae.py
+++ b/examples/vae.py
@@ -83,7 +83,7 @@ def main(args):
     opt_init, opt_update, get_params = optimizers.adam(args.learning_rate)
     svi_init, svi_update, svi_eval = svi(model, guide, elbo, opt_init, opt_update, get_params,
                                          encode=encode, decode=decode, z_dim=args.z_dim)
-    svi_update = jit(svi_update)
+    svi_update = jit(svi_update, static_argnums=(3,))
     rng = PRNGKey(0)
     train_init, train_fetch = load_dataset(MNIST, batch_size=args.batch_size, split='train')
     test_init, test_fetch = load_dataset(MNIST, batch_size=args.batch_size, split='test')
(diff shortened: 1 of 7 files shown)


numpyro/svi.py
```diff
@@ -52,32 +58,43 @@ def init_fn(rng, model_args=(), guide_args=(), params=None):
...
+                inv_transforms[site['name']] = transform
+                params[site['name']] = transform.inv(site['value'])
 
-    def update_fn(i, opt_state, rng, model_args=(), guide_args=()):
+        def constrain_fn(inv_transforms, params):
+            return {k: inv_transforms[k](v) for k, v in params.items()}
+
+        return optim_init(params), jax.partial(constrain_fn, inv_transforms)
+
+    def update_fn(i, rng, opt_state, constrain_fn, model_args=(), guide_args=()):
```
neerajprad: The one disadvantage to this interface is that `constrain_fn` is the only static argument to `update_fn` and we will need to mark this as a static arg for the jit to work.

fehiepsi: I think we can simplify this by making `constrain_fn` a local variable of `svi`, and define it in `svi_init`.

neerajprad: Yes, I think it is better to remove `constrain_fn` as args here. Will make the change.

numpyro/svi.py
```diff
@@ -52,37 +60,52 @@ def init_fn(rng, model_args=(), guide_args=(), params=None):
...
             if site['type'] == 'param':
-                params[site['name']] = site['value']
-        return optim_init(params)
+                constraint = site['kwargs'].pop('constraint', constraints.real)
+                transform = biject_to(constraint)
+                inv_transforms[site['name']] = transform
+                params[site['name']] = transform.inv(site['value'])
+
+        def transform_constrained(inv_transforms, params):
+            return {k: inv_transforms[k](v) for k, v in params.items()}
```
fehiepsi: can we reuse `constrain_fn` logic in `hmc_util`?

examples/vae.py
```diff
@@ -113,7 +113,7 @@ def body_fun(i, val):
             loss_sum, rng = val
             rng, = random.split(rng, 1)
             rng, batch = binarize(rng, test_fetch(i, test_idx)[0])
-            loss = svi_eval(opt_state, rng, (batch,), (batch,)) / len(batch)
+            loss = jit(svi_eval)(rng, opt_state, (batch,), (batch,)) / len(batch)
```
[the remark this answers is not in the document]
neerajprad: Sorry, I was testing something. This shouldn't be needed.

numpyro/svi.py
neerajprad: I have moved this to `infer_util` now, and we can put more common utilities there for prediction, etc in the future.

PR #212, opened 2019-06-18
Make isort independent of import type within sections

neerajprad:
This makes a config change to isort to ensure that packages are not sorted by import type which can result in confusing layout, i.e. all modules from a package are not grouped together (see below):

Current:
```
import numpy as onp
import pytest
import scipy.stats as osp_stats
from numpy.testing import assert_allclose
```

Proposed:
```
import numpy as onp
from numpy.testing import assert_allclose
import pytest
import scipy.stats as osp_stats
```

### DIFF
--- a/examples/minipyro.py
+++ b/examples/minipyro.py
@@ -1,8 +1,8 @@
 import argparse
 
-import jax.numpy as np
 from jax import lax, random
 from jax.experimental import optimizers
+import jax.numpy as np
 from jax.random import PRNGKey
 
 import numpyro.distributions as dist
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -3,6 +3,10 @@ import sys
 
 import sphinx_rtd_theme
 
+# HACK: This is to ensure that local functions are documented by sphinx.
+from numpyro.mcmc import hmc  # noqa: E402
+from numpyro.svi import svi  # noqa: E402
+
 # import pkg_resources
 
 # -*- coding: utf-8 -*-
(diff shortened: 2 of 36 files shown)


examples/minipyro.py
```diff
@@ -1,8 +1,8 @@
 import argparse
 
-import jax.numpy as np
 from jax import lax, random
 from jax.experimental import optimizers
+import jax.numpy as np
```
fehiepsi: The import order is a bit strange to me. Do you recommend to use this pattern?

neerajprad: I think it is a matter of taste. I recommend this because it is easier to see which modules are imported when they are sorted by module names and then add to it if needed, rather than duplicating because you missed the import.

A simple example like the one above based on the existing rules is:

```
import numpy.random as random
import pytest
from numpy.random import randint
```

If this was instead organized so that `numpy.random` was together, a user could instead choose to use `random.randint` in the code instead, instead of reimporting specific functions from `numpy.random` simply because it is easier to not miss that both have been imported. This also seems much more intuitive in terms of organization, and if I didn't have isort I would simply organize it this way by hand. That said, this is only my preference. I have just found the current behavior a bit unintuitive, but don't have strong opinions on this. I would like to understand why you find the above organization more natural?

fehiepsi: Thanks for explaining! It is just more familiar for me to use
```
import numpyro
from numpyro.contrib.distributions import cauchy, norm
```
than
```
import numpyro
from numpyro.contrib.distributions import cauchy, norm
```

neerajprad: I wanted to discuss before we merged this. I'll be happy to revert this if mixing different import types seems unnatural - my IDE is configured differently so I may be biased here, but I can see your point too.

fehiepsi: I think that the change looks fine. At first I find it unnatural but after working with some examples, except from/import, module/submodule names are sorted as expected. :)

PR #213, opened 2019-06-19
add inverse autoregressive transform

martinjankowiak:


numpyro/distributions/iaf.py
```diff
@@ -0,0 +1,84 @@
...
+        """
+        Calculates the elementwise determinant of the log jacobian
+        :param x: the input to the transform
+        :type x: numpy array
+        :param y: the output of the transform
+        :type y: numpy array
+        """
+        log_scale = self.arn.apply_fun(self.params, x)[..., 1, :]  # TODO: this shouldn't be recomputed here
+        log_scale = clamp_preserve_gradients(log_scale, self.log_scale_min_clip, self.log_scale_max_clip)
+        return log_scale.sum(-1)
```
martinjankowiak: what shape should this return?

numpyro/distributions/iaf.py
```diff
@@ -0,0 +1,84 @@
...
+from numpyro.distributions.constraints import Transform, real
+from jax.lax import stop_gradient
+import jax.numpy as np
+
+
+def clamp_preserve_gradients(x, min, max):
+    return x + stop_gradient(np.clip(x, a_min=min, a_max=max) - x)
+
+
+class InverseAutoRegressiveTransform(Transform):
```
martinjankowiak: do i need any `event_dim` logic?

fehiepsi: I don't think that it is necessary unless you want to use ComposeTransform. It is safe to define `event_dim=0/1` if your `domain` and `codomain` have the same dimension. In the other cases (e.g. domain is 1D, codomain is 2D), the current logic of ComposeTransform logdet will be wrong (something like `coevent_dim` can remedy it). Anyway, you can just left a TODO mark so we can look at it again later. :)

numpyro/distributions/iaf.py
fehiepsi: It is the shape of common batch shape of `x` and `y`.

numpyro/distributions/__init__.py
```diff
@@ -55,6 +56,7 @@
     'GaussianRandomWalk',
     'HalfCauchy',
     'HalfNormal',
+    'InverseAutoRegressiveTransform',
```
fehiepsi: nit: I think that it is better to lower case `Regressive` to match literature. Btw, should it be `InverseAutoregressiveFlow` to match pyro name?

numpyro/distributions/iaf.py
fehiepsi: I think that the shape here is correct.

numpyro/distributions/iaf.py
```diff
@@ -0,0 +1,84 @@
...
+
+from __future__ import absolute_import, division, print_function
+
+from numpyro.distributions.constraints import Transform, real
+from jax.lax import stop_gradient
+import jax.numpy as np
+
+
+def clamp_preserve_gradients(x, min, max):
+    return x + stop_gradient(np.clip(x, a_min=min, a_max=max) - x)
```
fehiepsi: good to know about `lax.stop_gradient`

numpyro/distributions/iaf.py
```diff
@@ -0,0 +1,84 @@
+# adapted from https://github.com/pyro-ppl/pyro/blob/dev/pyro/distributions/transforms/iaf.py
+
+from __future__ import absolute_import, division, print_function
+
+from numpyro.distributions.constraints import Transform, real
+from jax.lax import stop_gradient
+import jax.numpy as np
+
+
+def clamp_preserve_gradients(x, min, max):
```
neerajprad: Could you rename this to `"_clamp..."` so as to signify private usage? Otherwise, we can add it to `distributions.util` in case this will be more generally useful.

numpyro/distributions/iaf.py
```diff
@@ -0,0 +1,84 @@
...
+        :param params: the parameters for the autoregressive neural network
+        :type list:
+        """
+        self.arn = autoregressive_nn
+        self.params = params
+        self.log_scale_min_clip = log_scale_min_clip
+        self.log_scale_max_clip = log_scale_max_clip
+
+    @property
+    def codomain(self):
```
neerajprad: nit: Can be simplified to `codomain = real`, or we can also just leave it out and it will be correctly inherited from `Transform`.

numpyro/distributions/iaf.py
martinjankowiak: ok

numpyro/distributions/__init__.py
martinjankowiak: ok

numpyro/distributions/iaf.py
```diff
@@ -0,0 +1,84 @@
...
+
+    where :math:`\\mathbf{x}` are the inputs, :math:`\\mathbf{y}` are the outputs, :math:`\\mu_t,\\sigma_t`
+    are calculated from an autoregressive network on :math:`\\mathbf{x}`, and :math:`\\sigma_t>0`.
+
+    References
+
+    1. Improving Variational Inference with Inverse Autoregressive Flow [arXiv:1606.04934]
+    Diederik P. Kingma, Tim Salimans, Rafal Jozefowicz, Xi Chen, Ilya Sutskever, Max Welling
+    """
+    def __init__(self, autoregressive_nn, params, log_scale_min_clip=-5., log_scale_max_clip=3.):
```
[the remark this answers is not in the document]
martinjankowiak: i don't think there's any harm to exposing this

numpyro/distributions/iaf.py
```diff
@@ -0,0 +1,84 @@
...
+        """
+        :param autoregressive_nn: an autoregressive neural network whose forward call returns a real-valued
+            mean and logit-scale as a tuple
```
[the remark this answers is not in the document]
martinjankowiak: yup

numpyro/distributions/iaf.py
martinjankowiak: ok

numpyro/distributions/iaf.py
martinjankowiak: ok

martinjankowiak:
thanks for the review. addressed all comments. probably good to merge once tests pass

PR #215, opened 2019-06-20
Fix a small bug at `rng` arg of mcmc

fehiepsi:
If users provide `rng` to `mcmc` with single chain, the current code will throw error because it expected that `rng` will have `batch_size = 1`. This bug comes from my last rng PR and this PR fixes that.

In addition, I
+ move `log_density` to `infer_util` because it is not specific to `hmc`,
+ move euclidean ke to `hmc_util`
+ fix [tutorial](http://pyro.ai/numpyro/bayesian_regression.html#Bayesian-Regression-Using-NumPyro) not render list of items. We need an empty line before each markdown list.

### DIFF
--- a/notebooks/source/bayesian_regression.ipynb
+++ b/notebooks/source/bayesian_regression.ipynb
@@ -7,6 +7,7 @@
     "# Bayesian Regression Using NumPyro\n",
     "\n",
     "In this tutorial, we will explore how to do bayesian regression in NumPyro, using a simple example adapted from Statistical Rethinking [[1](#References)]. In particular, we would like to explore the following:\n",
+    "\n",
     " - Write a simple model using the `sample` NumPyro primitive.\n",
     " - Run inference using MCMC in NumPyro, in particular, using the No U-Turn Sampler (NUTS) to get a posterior distribution over our regression parameters of interest.\n",
     " - Learn about utilities such as `initialize_model` that are useful for running HMC.\n",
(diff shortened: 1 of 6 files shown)


numpyro/mcmc.py
```diff
@@ -405,11 +401,14 @@ def mcmc(num_warmup, num_samples, init_params, num_chains=1, sampler='hmc',
...
+            rngs = sampler_kwargs.pop('rng', vmap(PRNGKey)(np.arange(num_chains)))
+        else:
+            rng = sampler_kwargs.pop('rng', PRNGKey(0))
 
         init_kernel, sample_kernel = hmc(potential_fn, kinetic_fn, algo)
         if progbar:
-            hmc_state = init_kernel(init_params, num_warmup, progbar=progbar, rng=rngs[0],
+            hmc_state = init_kernel(init_params, num_warmup, progbar=progbar, rng=rng,
                                     **sampler_kwargs)
             samples_flat = fori_collect(0, num_samples, sample_kernel, hmc_state,
```
neerajprad: Not related to this PR, but could you add `progbar_desc='sample'` in `single_chain_mcmc`?

notebooks/source/bayesian_regression.ipynb
```diff
@@ -1342,7 +1347,8 @@
     "\n",
     "Let us now look at the posterior predictive distribution to see how our predictive distribution looks with respect to the observed divorce rates. To get samples from the posterior predictive distribution, we need to run the model by substituting the latent parameters with samples from the posterior. This sounds complicated, but this can be easily achieved by using effect handlers from the [handlers module](https://numpyro.readthedocs.io/en/latest/handlers.html).\n",
     "\n",
-    "In particular, note the use of the `substitute`, `seed` and `trace` effect handlers in the `predict` function. \n",
+    "In particular, note the use of the `substitute`, `seed` and `trace` effect handlers in the `predict` function.\n",
+    "\n",
```
neerajprad: Thanks for fixing this!

numpyro/mcmc.py
fehiepsi: I just check: that `single_chain_mcmc` does not use `progbar` so I think that adding `progbar_desc='sample'` does not take effect.

neerajprad: Sorry, my bad. I saw this in some test, but that uses fori_collect directly, so it isn't related.

PR #217, opened 2019-06-21
Add delta distribution

fehiepsi:
Delta distribution is useful for populating a joint latent variable into site values. We'll mostly do inference in unconstrained space so `log_density` arg is useful to incorporate log Jacobian of transforms if site's support is not real.


numpyro/distributions/discrete.py
```diff
@@ -301,6 +303,43 @@ def Categorical(probs=None, logits=None, validate_args=None):
...
+    def sample(self, key, sample_shape=()):
+        shape = sample_shape + self.batch_shape + self.event_shape
+        return np.broadcast_to(self.value, shape)
+
+    def log_prob(self, x):
+        if self._validate_args:
+            self._validate_sample(x)
+        log_prob = np.log(x == self.value)
+        log_prob = sum_rightmost(log_prob, len(self.event_shape))
+        return log_prob + self.log_density
```
neerajprad: ~Shouldn't `self.log_density` be only added to elements where `x == self.value`?~ I see that we are relying on inf addition here. Does that give any issues with gradients?

numpyro/distributions/discrete.py
```diff
@@ -301,6 +303,43 @@ def Categorical(probs=None, logits=None, validate_args=None):
         raise ValueError('One of `probs` or `logits` must be specified.')
 
 
+@copy_docs_from(Distribution)
+class Delta(Distribution):
```
neerajprad: Could you add this to docs?

numpyro/distributions/discrete.py
```diff
@@ -301,6 +303,43 @@ def Categorical(probs=None, logits=None, validate_args=None):
...
+    arg_constraints = {'value': constraints.real, 'log_density': constraints.real}
+    support = constraints.real
+
+    def __init__(self, value=0., log_density=0., event_dim=0, validate_args=None):
```
neerajprad: nit: the argument name `event_dim` is misleading, since this refers to the number of event dims from the right and not a single dimension. A better name might be `event_ndims` or `event_dims`.

numpyro/distributions/discrete.py
fehiepsi: We don't have tests for grad of invalid samples (for other distributions too). I am not sure how to properly test it.

numpyro/distributions/discrete.py
fehiepsi: Yup, at first I used `event_ndim` but later decide that just keep the same interface as in Pyro. Will change back to ndim.

numpyro/distributions/discrete.py
fehiepsi: The following script shows that `grad` does not affect by `inf`, but I am not sure if that is what you want to test?
```
def f(x):
    return np.log(x == x + 1) + 2 * x

grad(f)(1.)  # 2
```

neerajprad: My initial thought was that the gradient should be nan for values that don't match. i.e. suppose the initial value is `[1, 2]` (assuming event_ndim=0) and the value to be scored is `[1, 3]`, then the `log_prob` term would be `[0, -inf]`, and hence the gradient would be `[.., nan]`. However, this isn't correct either, and it probably doesn't matter because autograd will give the wrong value for the gradient.

fehiepsi: Got it. Hopefully we don't have usage cases which have invalid values for Delta distribution.

PR #219, opened 2019-06-23
add sparse linear regression example

martinjankowiak:


examples/sparse_regression.py
```diff
@@ -0,0 +1,202 @@
...
+           obs=Y)
+
+
+# Compute the mean and variance of coefficient theta_i (where i = dimension) for a
+# MCMC sample (eta1, xi, ...). Compare to theorem 5.1 in reference [1].
+def compute_mean_variance(X, Y, dimension, msq_inv, lam, eta1, xisq_inv, c, prec_obs):
+    P, N = X.shape[1], X.shape[0]
+
+    probe = np.zeros((2, P))
+    probe = jax.ops.index_update(probe, jax.ops.index[:, dimension], np.array([0.5, -0.5]))
```
martinjankowiak: this requires jax 0.1.38

examples/sparse_regression.py
```diff
@@ -0,0 +1,201 @@
...
+    lam = sample("lambda", dist.HalfCauchy(np.ones(P)))
+    kappa = np.sqrt(msq) * lam / np.sqrt(msq + np.power(eta1 * lam, 2.0))
+
+    # sample observation noise
+    prec_obs = sample("prec_obs", dist.Gamma(3.0, 1.0))
+
+    # compute kernel
+    kX = kappa * X
+    k = kernel(kX, kX, eta1, eta2, hypers['c']) + np.eye(N) / prec_obs
+    assert k.shape == (N, N)
```
martinjankowiak: does this sort of thing slow down jax?

fehiepsi: It does not affect afaik. I think we don't need to worry about operators which involves only constants. :)

examples/sparse_regression.py
```diff
@@ -0,0 +1,201 @@
...
+# Most of the model code is concerned with constructing the sparsity inducing prior.
+def model(X, Y, hypers):
+    S, P, N = hypers['expected_sparsity'], X.shape[1], X.shape[0]
+
+    sigma = sample("sigma", dist.HalfNormal(hypers['alpha3']))
+    phi = sigma * (S / np.sqrt(N)) / (P - S)
+    eta1 = sample("eta1", dist.HalfCauchy(phi))
+
+    msq_inv = sample("msq_inv", dist.Gamma(hypers['alpha1'], hypers['beta1']))
+    msq = 1.0 / msq_inv
```
fehiepsi: Do you need InvGamma to match the same model? We can have one if you think that it is necessary (e.g. scales of Gamma and InvGamma are different so it might affect the result).

martinjankowiak: i don't think it's necessary but it would be nice to have (just to make things less verbose)

examples/sparse_regression.py
```diff
@@ -0,0 +1,202 @@
...
+    k = kernel(kX, kX, eta1, eta2, hypers['c']) + np.eye(N) / prec_obs
+    assert k.shape == (N, N)
+
+    # sample Y according to the standard gaussian process formula
+    sample("Y", dist.MultivariateNormal(loc=np.zeros(X.shape[0]), covariance_matrix=k),
+           obs=Y)
+
+
+# Compute the mean and variance of coefficient theta_i (where i = dimension) for a
+# MCMC sample (eta1, xi, ...). Compare to theorem 5.1 in reference [1].
```
fehiepsi: +1

examples/sparse_regression.py
fehiepsi: If I understand correctly, then the probe value at `dimension` should be `[1., -1.]` (in other words, `e_i = 1`). Because the kernel is not linear, I don't see the equivalence of your derivation and Theorem 5.1. Could you pls elaborate this point?

examples/sparse_regression.py
```diff
@@ -0,0 +1,202 @@
...
+              'alpha2': 3.0, 'beta2': 1.0,
+              'alpha3': 1.0, 'c': 1.0,
+              'alpha_obs': 3.0, 'beta_obs': 1.0}
+
+    # do inference
+    rng = random.PRNGKey(0)
+    samples = run_inference(model, args, rng, X, Y, hypers)
+
+    # compute the mean and square root variance of each coefficient theta_i
+    means, stds = vmap(lambda dim: analyze_dimension(samples, X, Y, dim, hypers))(np.arange(args.num_dimensions))
```
fehiepsi: `vmap` is so useful :D

examples/sparse_regression.py
```diff
@@ -0,0 +1,202 @@
...
+
+
+# Helper function for analyzing the posterior statistics for coefficient theta_i
+def analyze_dimension(samples, X, Y, dimension, hypers):
+    vmap_args = (samples['msq_inv'], samples['lambda'], samples['eta1'],
+                 samples['xisq_inv'], samples['prec_obs'])
+    mus, variances = vmap(lambda msq_inv, lam, eta1, xisq_inv, prec_obs:
+                          compute_mean_variance(X, Y, dimension, msq_inv, lam,
+                                                eta1, xisq_inv, hypers['c'], prec_obs))(*vmap_args)
+    mean, variance = gaussian_mixture_stats(mus, variances)
```
fehiepsi: :+1:  looking like a useful trick

examples/sparse_regression.py
martinjankowiak: i think it's fine because the quadratic piece cancels out

examples/sparse_regression.py
```diff
@@ -0,0 +1,202 @@
...
+
+
+def main(args):
+    jax_config.update('jax_platform_name', args.device)
+    X, Y, expected_thetas = get_data(N=args.num_data, P=args.num_dimensions, S=args.active_dimensions)
+
+    # setup hyperparameters
+    hypers = {'expected_sparsity': max(1.0, args.num_dimensions / 10),
+              'alpha1': 3.0, 'beta1': 1.0,
+              'alpha2': 3.0, 'beta2': 1.0,
```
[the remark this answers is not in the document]
martinjankowiak: i can change the noise dist but i don't want to stray from the paper too much

examples/sparse_regression.py
```diff
@@ -0,0 +1,202 @@
...
+    msq = 1.0 / msq_inv
+    xisq_inv = sample("xisq_inv", dist.Gamma(hypers['alpha2'], hypers['beta2']))
+
+    eta2 = np.square(eta1) * msq_inv / np.sqrt(xisq_inv)
+
+    lam = sample("lambda", dist.HalfCauchy(np.ones(P)))
+    kappa = np.sqrt(msq) * lam / np.sqrt(msq + np.square(eta1 * lam))
+
+    # sample observation noise
+    prec_obs = sample("prec_obs", dist.Gamma(hypers['alpha_obs'], hypers['beta_obs']))
```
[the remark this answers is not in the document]
martinjankowiak: where do you see this in the paper?

examples/sparse_regression.py
```diff
@@ -0,0 +1,202 @@
...
+def dot(X, Z):
+    return np.dot(X, Z[..., None])[..., 0]
+
+
+# The kernel that corresponds to our quadratic regressor.
+def kernel(X, Z, eta1, eta2, c):
+    eta1sq = np.square(eta1)
+    eta2sq = np.square(eta2)
+    k1 = 0.5 * eta2sq * np.square(1.0 + dot(X, Z))
+    k2 = 0.5 * eta2sq * dot(np.square(X), np.square(Z))
```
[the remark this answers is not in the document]
martinjankowiak: ok

examples/sparse_regression.py
```diff
@@ -0,0 +1,202 @@
...
+    k3 = (eta1sq - eta2sq) * dot(X, Z)
+    k4 = np.square(c) - 0.5 * eta2sq
```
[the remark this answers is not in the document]
martinjankowiak: the arxiv paper has typos which were fixed in the icml version

examples/sparse_regression.py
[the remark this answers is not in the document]
martinjankowiak: hmm i tried this and got weird results. i think it puts too much mass at zero noise. is there really any need to make this change? the important point is the sparsity inducing prior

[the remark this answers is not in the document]
fehiepsi: Sorry, I thought I have deleted this comment (in favor of hyperparameter comment). It is better to reduce the number of hyperparameters. But this point is not important if you already find good ones for the synthesis dataset.

examples/sparse_regression.py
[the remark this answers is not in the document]
fehiepsi: Yup, it depends on you. Just my suggestion if it helps reducing the effort of tuning hyperparameters. These choices of alpha1, beta1, alpha2, beta2 are just to have StudentT priors over slopes but the authors of the papers skip mentioning about it (instead, they use refs). Nevermind about this point if you already find good hyperparameters.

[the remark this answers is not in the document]
martinjankowiak: i'm fine with the current settings. i may revisit this as i play around with the model more

fehiepsi:
Just curious, do you intend to extend the code to discover pairwise interaction? Looking like all ingredients are available. :smiley:

martinjankowiak:
yes i'll probably eventually add some pairwise bits

Issue #214, opened 2019-06-20
Implement IAF autoguide


fehiepsi:
Given that we will have IAF implemented soon, it is time to make some auto guide using it. So far, I guess that we just need to mimic the current [implementation](https://github.com/pyro-ppl/pyro/blob/dev/pyro/contrib/autoguide/__init__.py#L618) in Pyro.

- [x] Implement AutoDiagonalNormal (mainly to sketch out necessary ingredients for IAF one)
- [x] Implement AutoIAFNormal

neerajprad:
@fehiepsi - are you working on this?

fehiepsi:
I am making Delta distribution. Do you intend to work on auto normal?

neerajprad:
Sure, I can try to sketch out auto normal in the meantime.

PR #224, opened 2019-06-26
add IAF autoguide

fehiepsi:
Addresses #214. After implementing, for some reasons, the logistic regression test just passed magically. I'll need to replicate a simple example from IAF paper to confirm that the class actually works.

TODO:
- ~[ ] Add an example with spherical Gaussian prior (figure 1 of https://arxiv.org/abs/1606.04934)~
- ~[ ] fix jitting bug~ this is JAX bug and can avoid by running the update under the prims `fori_loop` or `fori_collect`


test/contrib/test_autoguide.py
```diff
@@ -39,6 +39,7 @@ def body_fn(i, val):
```
neerajprad: Could you test this on `beta_bernoulli`? I think the diag normal guide gives a somewhat biased estimate for `beta`, and I'm curious to see if this does better.

fehiepsi: The IAF guide only works with latent_size > 1. I'll see if I can remove that restriction. If not, I'll add another similar test. :)

neerajprad: I see. You could also just increase the dimension in that test (for latent and observed data).

martinjankowiak:
a single iaf will always leave one dimension marginally gaussian. you need
to compose multiple iafs in >1 dimensions to fully escape gaussianity

test/contrib/test_autoguide.py
fehiepsi: Looking like bias issue still happens with IAF. I'll try to add multiple IATransform to see if it helps.

fehiepsi:
> a single iaf will always leave one dimension marginally gaussian. you need to compose multiple iafs in >1 dimensions to fully escape gaussianity

Yup, that makes sense. I'll make default implementation as in Section 4.1.1 of Neutra paper (3 flows, 2 hidden layer / flow, reverse permutation between layers) and in IAF paper. Later, we need to update AutoIAFNormal in Pyro too.

test/contrib/test_autoguide.py
neerajprad: I haven't looked into this in detail, and that was just my hypothesis. It is also possible that the data size is small. If NUTS gives answers close to the true value, then we know it is a bias issue.

fehiepsi: It turns out that we should fix the values of true coefs. :D

test/contrib/test_autoguide.py
```diff
@@ -32,13 +34,16 @@ def body_fn(i, val):
         loss, opt_state_, rng_ = svi_update(i, rng_, opt_state_, model_args=(data,), guide_args=(data,))
         return opt_state_, rng_
 
-    opt_state, _ = lax.fori_loop(0, 300, body_fn, (opt_state, rng_train))
-    median = guide.median(opt_state)
-    assert_allclose(median['beta'], 0.8, rtol=rtol)
+    opt_state, _ = lax.fori_loop(0, 1000, body_fn, (opt_state, rng_train))
+    true_coefs = (np.sum(data, axis=0) + 1) / (data.shape[0] + 2)
```
fehiepsi: The posterior with prior Beta(1, 1) has mean `(num_success + 1) / (num_data + 2)`

test/contrib/test_autoguide.py
```diff
@@ -32,13 +34,16 @@ def body_fn(i, val):
         loss, opt_state_, rng_ = svi_update(i, rng_, opt_state_, model_args=(data,), guide_args=(data,))
         return opt_state_, rng_
 
-    opt_state, _ = lax.fori_loop(0, 300, body_fn, (opt_state, rng_train))
-    median = guide.median(opt_state)
-    assert_allclose(median['beta'], 0.8, rtol=rtol)
```
fehiepsi: the median of Beta distribution is a bit complicated to compute so it is better to skip it and use empirical test from `sample_posterior` instead.

test/contrib/test_autoguide.py
neerajprad: I completely missed that the data has only 10 elements. Makes sense!

fehiepsi:
@martinjankowiak @neerajprad I am getting stuck at using IAF for a non-trivial example. Could you take a look at https://gist.github.com/fehiepsi/c7ca886fc941a6385cdc27c1de60c044 and give me some suggestions? Basically, the notebook shows the following issues:
+ I am not sure if the approach in the notebook to approximate log_density mathematically correct. That approach is just my understanding.
+ If we jit svi_update, we'll get `nan` loss after first step
+ If we don't jit svi_update, the inference will be very slow. In addition, the result is very poor.

martinjankowiak:
i think for the model you just want to do something like 

`pyro.sample('obs', dist.Bernoulli(logits=my_custom_log_density), obs=np.array([1.]))`

i'd also maybe start out with something easier than moons. like a mixture of two gaussians.

fehiepsi:
@martinjankowiak Yes, I use Delta for that purpose. (nb: the trick `pyro.sample('obs', dist.Bernoulli(logits=my_custom_log_density), obs=np.array([1.]))` is not quite correct because it will give us `log(sigmoid(custom_log_density))` instead of `custom_log_density`.) I'll try mixture of two gaussians after fixing the jit issue. :)

fehiepsi:
I think that I have isolated the bug with dual moon pe and reported at https://github.com/google/jax/issues/939 (not sure if there are other issues). In the mean time, I'll play with mixture model instead.

fehiepsi:
I just update the [gist](https://gist.github.com/fehiepsi/c7ca886fc941a6385cdc27c1de60c044) with both dual moon and mixture of Gaussians. None of them gives good result but at least we can see that the model can learn. For some reason, the unwrapped samples have Gaussian density (rather than target density). I think that this PR is ready to review. In the mean time, I will try to go around other github repos to see how people uses IAF for density estimation (I believe I am missing a basic thing here).

martinjankowiak:
how about try an actual bayesian model? e.g. exponential likelihood with gamma prior in two dimensions?

then you can at least be sure the problem isn't with the dist.Uniform/dist.Delta stuff

fehiepsi:
@martinjankowiak It seems that the masks are a bit weird to me. There are a lot of zeros in the masks, which is quite different from MADE paper. Is it expected?
```
import pprint
import numpy as onp
from numpyro.contrib.nn.auto_reg_nn import create_mask

pprint.pprint([x.T for x in create_mask(2, [10, 10], onp.array([0, 1]), 2)[0]])
```
which yields
```
[array([[0., 0.],
       [0., 0.],
       [0., 0.],
       [0., 0.],
       [0., 0.],
       [1., 0.],
       [1., 0.],
       [1., 0.],
       [1., 0.],
       [1., 0.]]),
 array([[1., 1., 1., 1., 1., 0., 0., 0., 0., 0.],
       [1., 1., 1., 1., 1., 0., 0., 0., 0., 0.],
       [1., 1., 1., 1., 1., 0., 0., 0., 0., 0.],
       [1., 1., 1., 1., 1., 0., 0., 0., 0., 0.],
       [1., 1., 1., 1., 1., 0., 0., 0., 0., 0.],
       [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],
       [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],
       [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],
       [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],
       [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.]]),
 array([[1., 1., 1., 1., 1., 0., 0., 0., 0., 0.],
       [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],
       [1., 1., 1., 1., 1., 0., 0., 0., 0., 0.],
       [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.]])]
```
I expect the first mask should be
```
[array([[1., 0.],
       [1., 0.],
       [1., 0.],
       [1., 0.],
       [1., 0.],
       [1., 0.],
       [1., 0.],
       [1., 0.],
       [1., 0.],
       [1., 0.]]),
```
and the second mask should me `np.ones((10, 10))` e.g.

martinjankowiak:
yes the masks are a bit different. this is directly translated from pyro.
it should work fine. all that matters is that it's bijective and has a
computable jacobian. both of which are established in the tests

Am Fr., 28. Juni 2019 um 08:20 Uhr schrieb Du Phan <notifications@github.com

neerajprad:
Taking a look at this now; sorry about the delay in reviewing this!

neerajprad:
@fehiepsi - I was looking at your half-moon example, and tried to replicate the same in Pyro. I was initially facing the same issue as you did with a single mode in between the two moons. I changed the following in the code:
 - changed the prior to not be quite as wide (Uniform(-3, 3)).
 - changed `hidden_dim` to 20 and number of flows to 5 (I had to modify the current AutoIAFNormal guide to change the number of flows).
 - Adam's learning rate to 1e-3 and ran for 8k steps.

I was able to get better results with that. Not claiming that these settings are optimal, and these problems do seem to require fair amount of hyperparameter twiddling. Your implementation looks correct otherwise, and might just tuning some of these hyperparams.

image

martinjankowiak:
yeah, these things can be a bit tricky to tune. also i think two dimensions can be particularly tricky

fehiepsi:
Thanks @neerajprad ! Change prior did help a lot for getting stable results. I have played around with those parameters but cannot get a result as good as you. :( Only one thing which makes me feel good is it is fast to run inference in NumPyro, even 100000 steps just took a few seconds, so I can try out many assumptions quickly. ^^

So I decided to read literature about flows instead. It was a really good journey, so I would like to share what I learn here too
+ Single mode issue of IAF can be observed at Figure 6 of [neural autoregressive flow](https://arxiv.org/abs/1804.00779.pdf) paper. As Martin suggested, two dimensions can be tricky for *autoregressive* flows. Maybe a residual connection can help (I will explore a bit more of this idea to confirm).
+ Some new flows such as [block neural autoregressive flow](https://arxiv.org/abs/1904.04676) (NBAF) can do pretty good job for this kind of job. So I think that it is worth to try it out. If it works, then we might not have to use multi-chains in HMC to detect multi-modes of posterior for mixture problems (I hope so :D). Good news is that @stefanwebb has implemented a lot of flows in Pyro (:tada:). I'll try to convert NBAF (mainly because I like the simplicity idea of this flow) to numpyro to see if it helps (despite that its inverse can not be found analytically, we don't need to calculate inverse at all).

martinjankowiak:
@fehiepsi while it's true that you don't need an inverse for SVI, it is useful in other contexts (e.g. in HMC). if you to implement a more recent flow i suggest you try [this one](https://arxiv.org/pdf/1906.04032.pdf), which has analytic inverses

stefanwebb:
If I can throw my two cents in here, I would start off with Neural Autoregressive Flow rather than Block-NAF. I don't think normalizing flows for variational inference is a very well understood territory at the moment... Most of the literature seems to focus on density estimation for images and I think there may be additional issues that haven't been addressed when using them for VI...

Neural Spline Flows seem promising in that they are flexible and have an easy inverse, but one thing to keep in mind is that you have the set the end-points of the splines. If you don't know where the interesting part of your posterior space is (and the support not bounded) then this may be very limiting. For image models, on the other hand, it isn't a problem

stefanwebb:
Why would autoregressive flows not work well in two-dimensions?

stefanwebb:
@fehiepsi To get IAF to work well I would use two affine transformation layers plus a permutation layer in between. Make sure there the identity permutation is being used in the affine layers. In general though, the transformation of IAF is very weak and I recommend switching over to NAF

fehiepsi:
@martinjankowiak I think that we don't need to compute the inverse for HMC, so any transform which is fast to forward and compute logdet would be fine for us. Neural spline flow looks a bit complicated to me to understand, which the figure 1 of block NAF paper is quite intuitive. :D 

@stefanwebb Because I want to use the flow in HMC, I would favour a smaller and faster model (to forward and compute logdet). I just play with flows a few days ago and from BNAF paper, BNAF is comparable to NAF but requires less parameters than NAF. Could you give me more of your opinion about these two flows?

About IAF, I did use 3 flows (or more) with reversed permutation for flows 2 and 3. Does that align with your suggestion? I mean, what I use is
```
x = autoregressive1(x)
x = autoregressive2(x, permutation=::-1)
```
Is it equivalent to
```
x = autoregressive1(x)
x = x[::-1]
x = autoregressive2(x)
```
?

martinjankowiak:
@fehiepsi you don't need the inverse during HMC but you need it at the end to transform samples back to the original space. since you only need to do it once, it doesn't need to be particularly fast, but you need to be able to do it. so if there is no analytic inverse a fully numerical procedure is fine as long as it's sufficiently stable. this is why i implemented IAF first as a simplest option. it has limited expressivity but it has an easy to compute analytic inverse. block-naf would potentially be great for HMC but one would need to implement a numerical inverse.

fehiepsi:
@martinjankowiak Assuming we want to find the posterior p(z | x), we use IAF/NAF to reparametrize z = f(u). Then first we learn `f` using SVI such that q(z=f(u)) ~ p(z | x), where u ~ N(0, 1). After that, we use HMC to get samples `u`s. Then we apply `f` to get samples `z`s. Did I understand correctly? If I am wrong and we need to use inverse, then I agree that NAF would be a bad choice due to its numerical inverse.

martinjankowiak:
@fehiepsi yeah you're right we don't need it to be invertible. for some reason i had misremembered that that was required. ok that's good. onward to more flexible flows!

fehiepsi:
@martinjankowiak Great to hear your confirmation!

I just did some more experiments based on @stefanwebb suggestion and seem have detected several issues:
+ It seems that the transformed distribution log_prob is wrong because I used the base dist Normal with `event_shape=()` and we need to specify that InverseAutoregressive transform also has event_dim=1. After fixing the above bug (by temporarily creating a wrapper for Normal with correct event_shape), the result looks much better and closer to @neerajprad result.
+ I think we need to cache `x` and `log_scale` in InverseAutoregressive transform because we use inverse operator to calculate log_prob of transformed distribution. For 2D, the inverse is not expensive but it might cause some numerical issues.
+ The way I specify the model might not be good. I think it is better to generate data from those potential_fn and fit the guide and model on these data.

fehiepsi:
Finally, I get a working version by combining both @neerajprad and @stefanwebb suggestions:
+ use a single hidden layer `hidden_dims=[20]`
+ use skip connections and identity permutation in each affine flow, and reverse dimensions between flows

dual_moon

martinjankowiak:
@fehiepsi awesome nice work! yeah, unfortunately these things can be a bit tricky to get to work.

fehiepsi:
@martinjankowiak @neerajprad this is ready to review. I'll make an example for flow hmc in a follow-up PR. :)

numpyro/contrib/autoguide/__init__.py
```diff
@@ -253,3 +255,83 @@ def quantiles(self, opt_state, quantiles):
...
+        for i in range(self.num_flows):
+            arn_params = params['{}_arn__{}'.format(self.prefix, i)]
+            if i > 0:
+                flows.append(PermuteTransform(np.arange(latent_size)[::-1]))
+            flows.append(InverseAutoregressiveTransform(self.arns[i], arn_params))
+        iaf_dist = dist.TransformedDistribution(_Normal(np.zeros(latent_size), 1.), flows)
+        latent_sample = iaf_dist.sample(rng, sample_shape)
+        return self._unpack_latent(latent_sample)
+
+    def _sample_latent(self, *args, **kwargs):
```
neerajprad: I am beginning to think that we will be better served by avoiding `*args, **kwargs` in these concrete classes, and just using fully qualified names like `sample_posterior(self, rng, opt_state, sample_shape=())`. That will both simplify the code and be self-documenting. Since I started with this, it is better to do it in a separate PR for both the classes.

numpyro/contrib/autoguide/__init__.py
```diff
@@ -253,3 +255,83 @@ def quantiles(self, opt_state, quantiles):
...
+        latent_size = np.size(self._init_latent)
+        if latent_size == 1:
+            raise ValueError('latent dim = 1. Consider using AutoDiagonalNormal instead')
+        flows = []
+        if not self.arns:
```
neerajprad: I don't yet have a solution for this, but I think we might need to rethink the API to make `__call__` jittable.

numpyro/distributions/util.py
```diff
@@ -361,6 +361,10 @@ def relu(x):
     return np.maximum(x, 0.)
 
 
+def elu(x):
```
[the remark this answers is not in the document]
fehiepsi: `elu` is the recommended nonlinearity in both IAF paper and Neutralization (hmc+iaf) paper so I add it to exercise alternatives. In addition, its smoothness can help HMC (according to @martinjankowiak :D). Probably we'll make it the default nonlinearlity in iaf.

Btw, with functional style ARN in #246, we'll use stax.Relu, stax.Elu instead. So these custom non-linearity implementation will be removed.

numpyro/contrib/autoguide/__init__.py
fehiepsi: Yup, agree. :D

numpyro/contrib/autoguide/__init__.py
fehiepsi: It is somehow jittable I guess. The test and my experiments exercise __call__ with `lax.fori_loop` without problem. These constructions happen at svi_init.

numpyro/contrib/autoguide/__init__.py
```diff
@@ -253,3 +255,83 @@ def quantiles(self, opt_state, quantiles):
...
+
+class AutoIAFNormal(AutoContinuous):
+    """
+    This implementation of :class:`AutoContinuous` uses a Diagonal Normal
+    distribution transformed via a
+    :class:`~numpyro.distributions.iaf.InverseAutoregressiveTransform`
+    to construct a guide over the entire latent space. The guide does not
+    depend on the model's ``*args, **kwargs``.
+
+    Usage::
```
martinjankowiak: this docstring is missing some entries

numpyro/contrib/autoguide/__init__.py
neerajprad: I see; it might be worth having a separate method for initialization, since it will be easy to miss this and end up jitting the initialization code (if used outside the svi context). I'll try to refactor this in a separate PR.

Issue #226, opened 2019-06-27
Unit tests failing on jax==0.1.39


neerajprad:
Distribution tests are failing with:

TypeError: Primal inputs to reverse-mode differentiation must be of float or complex type, got type int32

fehiepsi:
Thanks for addressing this, @neerajprad !

PR #227, opened 2019-06-27
Fix tests with jax==0.1.39

neerajprad:
Addresses #226


numpyro/distributions/constraints.py
```diff
@@ -428,6 +428,7 @@ def __call__(self, x):
         pad_width = [(0, 0)] * x.ndim
         pad_width[-1] = (0, 1)
         z_padded = np.pad(z, pad_width, mode="constant", constant_values=1.)
+        pad_width = [(0, 0)] * x.ndim
```
neerajprad: This is needed because reusing the same `pad_width` results in a cache hit (the id for the arg doesn't change) and the older value of the function call is used.

fehiepsi:
@neerajprad Could you disable tests for `distributions.contrib`? I'll delete (maybe just keep the bare bone) that to reduce maintenance overhead.

neerajprad:
> @neerajprad Could you disable tests for distributions.contrib? I'll delete (maybe just keep the bare bone) that to reduce maintenance overhead.

I think those tests should be fixed now; we can keep whatever we need to and remove the rest in a later PR.

fehiepsi:
Thanks for addressing them! ^^

PR #228, opened 2019-06-28
Remove laxtuple

fehiepsi:
The new version of JAX supports namedtuple out of box, so we can drop laxtuple for now. :)

The slowness of `np.stack/np.concatenate` is also solved. So we can use it in `fori_collect`.

### DIFF
--- a/examples/vae.py
+++ b/examples/vae.py
@@ -83,7 +83,6 @@ def main(args):
     opt_init, opt_update, get_params = optimizers.adam(args.learning_rate)
     svi_init, svi_update, svi_eval = svi(model, guide, elbo, opt_init, opt_update, get_params,
                                          encode=encode, decode=decode, z_dim=args.z_dim)
-    svi_update = jit(svi_update)
     rng = PRNGKey(0)
     train_init, train_fetch = load_dataset(MNIST, batch_size=args.batch_size, split='train')
     test_init, test_fetch = load_dataset(MNIST, batch_size=args.batch_size, split='test')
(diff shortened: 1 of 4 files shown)


examples/vae.py
```diff
@@ -83,7 +83,6 @@ def main(args):
     opt_init, opt_update, get_params = optimizers.adam(args.learning_rate)
     svi_init, svi_update, svi_eval = svi(model, guide, elbo, opt_init, opt_update, get_params,
                                          encode=encode, decode=decode, z_dim=args.z_dim)
-    svi_update = jit(svi_update)
```
fehiepsi: this is not necessary because we use `svi_update` inside a jitted function.

numpyro/mcmc.py
```diff
@@ -270,7 +260,8 @@ def _nuts_next(step_size, inverse_mass_matrix, vv_state, rng):
                                  max_tree_depth=max_treedepth)
         accept_prob = binary_tree.sum_accept_probs / binary_tree.num_proposals
         num_steps = binary_tree.num_proposals
-        vv_state = vv_state.update(z=binary_tree.z_proposal,
+        vv_state = IntegratorState(z=binary_tree.z_proposal,
+                                   r=vv_state.r,
```
neerajprad: Is this added for logging purposes?

neerajprad:
Feel free to merge this at your convenience!

numpyro/mcmc.py
fehiepsi: This is added because namedtuple does not have `update` method. Here, `r` field keeps its old value.

neerajprad: Ahh..sorry, I missed that. Makes sense.

PR #229, opened 2019-07-01
add pairwise interaction analysis to sparse_regression

martinjankowiak:


examples/sparse_regression.py
```diff
@@ -95,12 +96,44 @@ def compute_mean_variance(X, Y, dimension, msq, lam, eta1, xisq, c, var_obs):
...
+
+    kX = kappa * X
+    kprobe = kappa * probe
+
+    k_xx = kernel(kX, kX, eta1, eta2, c) + var_obs * np.eye(N)
+    k_xx_inv = np.linalg.inv(k_xx)
+    k_probeX = kernel(kprobe, kX, eta1, eta2, c)
+    k_prbprb = kernel(kprobe, kprobe, eta1, eta2, c)
+
+    vec = np.array([0.25, -0.25, -0.25, 0.25])
```
fehiepsi: should `probe[:, dim1] = np.array([1.0, -1.0, 0, 1.0])`, `probe[:, dim2] = np.array([0.0, 0.0, 1.0, 1.0])` and `vec = np.array([-0.5, 0.5, -1, 1])` as in the paper?

martinjankowiak: the paper isn't quite right. it assumes there are no theta_ii terms but i do the more general thing and keep those. you can see if you run it that it computes the expected means

fehiepsi: yeah, you are right. I don't know what's going on with the notations in formula (16) in the paper but I can compute the required term from your `probe` and `vec`. Well done! Just for my understanding: it seems to me that we can use a 3xP probe `probe[dim1] = [1, 0, 1]`, `probe[dim2] = [0, 1, 1]` and `vec = [-1, -1, 1]` because `phi_ij = phi_i + phi_j + phi_ij + phi_ii + phi_jj - (phi_i + phi_ii) - (phi_j + phi_jj)) = g(e_i + e_j) - g(e_i) - g(e_j)`?

martinjankowiak: i think if you do that the constant term won't cancel

martinjankowiak: if you're ok with this let's merge as it. i will probably revisit the sparsity inducing prior and remove quadratic terms of the form x_ii and then add code for sampling from the posterior in a subsequent PR

fehiepsi: Bingo! There is a constant term also. Thanks!

fehiepsi: This PR is already well done, the above question is just for my understanding. :)

Issue #164, opened 2019-05-21
Add support for InverseTransform


neerajprad:
This will allow us to create new transform classes out of existing ones. e.g. `LogitTransform` as `SigmoidTransform().inv`. 

It will also be nice if the client code can just take in a list of transforms and call `.inv` on it without having any knowledge of whether the original or the inverted transform was passed in.

e.g. 
**Numpyro**

```sh
>>> t = SigmoidTransform()

>>> t.inv
 <bound method SigmoidTransform.inv of <numpyro.distributions.constraints.SigmoidTransform object at 0x7f92680ce710>>

>>> t.inv.inv
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-10-15abeba100ea> in <module>
----> 1 t.inv.inv

AttributeError: 'function' object has no attribute 'inv'
```

**Pytorch**
```sh
>>> t = SigmoidTransform()

>>> t.inv
 _InverseTransform()

>>> t.inv.inv
 SigmoidTransform()
```

fehiepsi:
I have thought about this and found that it just makes things more confusing. Could you be more explicit about `the client code can just take in a list of transforms and call .inv on it without having any knowledge of whether the original or the inverted transform was passed in`? I guess there are some usage cases where both `SigmoidTransform()` and `SigmoidTransform().inv` are used in client code?

neerajprad:
> Could you be more explicit about the client code can just take in a list of transforms and call .inv on it without having any knowledge of whether the original or the inverted transform was passed in? 

The reason would just be to have a wider class of transforms available by not having to write a separate implementation for the inverse transform if a transform is bijective. e.g. Instead of a `LogitTransform`, we should just be able to use `SigmoidTransform().inv` as a regular transform. Right now transforms from a constraint to real are not first class transforms (but simply functions) because, amongst other things, we cannot invert them. Just as we can compose transforms, we should also be able to invert them when they are invertible. Also, you could imagine a use case where we need to compose with the inverse transform.

neerajprad:
Changed the title/description to better reflect the motivation. This is a nice to have, but isn't blocking anything that we are working on right now.

fehiepsi:
Hmm, this might be nice to have. But I am curious because I don't know an explicit distribution which requires this feature.

neerajprad:
[...]

PR #234, opened 2019-07-03
Revise IAF transform

fehiepsi:
This PR revises IAF a bit to support doing experiments in #224. Basically, I want to incorporate suggestions from @stefanwebb to make it easier to test out various settings so we can make IAF autoguide more stable.

Changes include:
+ I move `InverseAutoregressiveTransform` to `flows.py`, which is the place we can put more flows. My motivation is to avoid using `dist.InverseAutoregressiveTransform` or `dist.iaf.InverseAutoregressiveTransform`, in favour of the pattern
```
from numpyro.distributions.flows import IAFTransform, NSFTransform, ...
```
So I also revise `iaf_test` to make it easier to add tests for other flows.
+ Add `event_dim`, `domain`, `codomain` to IAF transform
+ Add `caching` kwargs to IAF transform so we can use cached result in auto guide.
+ Using only `jax.numpy` throughout `contrib.nn`.
+ About API, I move `permutation` of `AutoregressiveNN` to `__init__` to bring `init_fun` and `apply_fun` closer to Stax module.
+ Support `skip_connections`. I need to add `bias` arg to `MaskedDense` to support this.
+ Match mask of autoregressive_nn with MADE paper.

This PR is blocked by #233 where `real_vector` constraint is added. In the mean time, I am happy to discuss more about the above changes so we can have better solutions for them.


numpyro/contrib/nn/auto_reg_nn.py
```diff
@@ -38,35 +35,29 @@ def create_mask(input_dim, hidden_dims, permutation, output_dim_multiplier):
...
-    var_index[permutation] = onp.arange(input_dim)
-    dtype = var_index.dtype
+    var_index = np.zeros(permutation.shape[0])
+    var_index = ops.index_update(var_index, permutation, np.arange(input_dim))
 
     # Create the indices that are assigned to the neurons
     input_indices = 1 + var_index
-    hidden_indices = [sample_mask_indices(input_dim, h) for h in hidden_dims]
-    output_indices = onp.tile(var_index + 1, output_dim_multiplier)
+    hidden_indices = [sample_mask_indices(input_dim - 1, h) for h in hidden_dims]
```
martinjankowiak: why `- 1`?

numpyro/distributions/flows.py
```diff
@@ -26,20 +23,25 @@ class InverseAutoregressiveTransform(Transform):
...
+    event_dim = 1
+
+    def __init__(self, autoregressive_nn, params, caching=False,
+                 log_scale_min_clip=-5., log_scale_max_clip=3.):
         """
         :param autoregressive_nn: an autoregressive neural network whose forward call returns a real-valued
             mean and log scale as a tuple
         :param params: the parameters for the autoregressive neural network
         :type list:
+        :param bool caching: where to cache results during forward pass
```
martinjankowiak: where => whether

numpyro/distributions/flows.py
```diff
@@ -49,25 +51,34 @@ def __call__(self, x):
         mean, log_scale = out[..., 0, :], out[..., 1, :]
```
martinjankowiak: note that this `0/1` stuff i wrote is kinda gross. it'd be better if shapes were such that we could just do
`mean, log_scale = self.arn.apply_fun(self.params, x)`
of course, you needn't do that in this PR. but maybe put a TODO?

numpyro/contrib/nn/auto_reg_nn.py
fehiepsi: In section 4 of [MADE](https://arxiv.org/pdf/1502.03509.pdf) paper, these numbers `m(k)` lie between `1` and `D - 1` (the long explanation there is a unit with `m(k) = D` will depend on all inputs, hence could not be used to model `p(x_k | x_{i < k})`

numpyro/distributions/flows.py
fehiepsi: nice suggestion! I'll do it shortly in a follow-up PR.

numpyro/contrib/nn/auto_reg_nn.py
martinjankowiak: ok but please help me understand what is the exact claim you're making. there are three kinds of mask code you can imagine:

i) one that is correct but has too many zeros
ii) one that is correct and is somehow optimal in that it has the fewest possible zeros
iii) one that is incorrect because it's not autoregressive (in which case the inversion tests etc should fail)

in making this change are you moving from i) to ii) or what?

is @stefanwebb's implementation of masking in pyro somehow suboptimal? or is it just that his code is organized differently from the description in the MADE paper (but is ultimately equivalent) and you're making a change that makes it look more like the MADE paper (but doesn't functionally change anything).

fehiepsi:
Yeah, let's make it clearer about `caching` stuff so we can discuss about better alternatives.
+ In [pytorch](https://github.com/pytorch/pytorch/blob/master/torch/distributions/transformed_distribution.py#L111) and numpyro, the way we calculate a transformed distribution log_prob is
```
x = transform.inv(value)
transformed_dist.log_prob(value) = transform.logdet(x, value) + base_dist.log_prob(x)
```
+ When we use this distribution in `guide`, then the first job is to get a sample `value`. To get `value`, we get a sample `x` from base_dict and set `value = transform(x)`. After that, when we compute `log_prob`, we don't need to recompute the inverse of `value`. That is the reason for having `caching=True`.
+ When we use this distribution in `model`, then the sample `value` is substituted from `guide` (which we don't know which process generates it). To compute log_prob, we need to compute the inverse explicitly (because there is no cached value). That is the scenario for having `caching=False`.

But I am not sure if a `caching` flag is needed. Things might be trickier when InverseTransform is supported in #164.

numpyro/contrib/nn/auto_reg_nn.py
fehiepsi: @martinjankowiak I'll try to make it clearer but please let me know if there are any vague point to you. :)

The change is just for i) to ii). As long as the masks created by `output_indices = (1, ..., D) > hidden_indices >= input_indices = (1, ..., D)`, autoregressive property is guaranteed. In the previous version, hidden indices are created equally from `1 to D` (in `sample_mask_indices`), then -1 (indeed, `-1` is involved by changing the above inequality to `output_idx >= hidden_idx > input_idx`). Equivalent, the previous PR samples hidden indices from 0 -> D - 1. According the the paper, these hidden nodes with `index = 0` will be constant (because that index will always be smaller than input_indices, the hidden node will be masked out). I think that also explains for the phrase *The output first in the order (e.g. x_2 in the figure) is connected to hidden units rather than being unattached. Tracing a path back through the network, however, this variable will still be unconnected to any input variables* because any output indices will be able to connect to that `idx=0` hidden node (`[1, ..., D] > 0`), but that hidden node cannot connect to any input node (`0 < [1, ..., D]`).

This PR samples hidden indices from (1 to D - 1) instead. So that phenomenon (hidden idx = 0) won't happen.

neerajprad:
@fehiepsi - Sorry about the delay in reviewing this. Let us target getting this merged today.

fehiepsi:
No worries, pls take your time.

neerajprad:
> Yeah, let's make it clearer about caching stuff so we can discuss about better alternatives.

My first thought here would be to avoid this as far as possible. Using cache will make it really hard to JIT such classes. In general, I think this is one thing we need to be careful when we use classes with JAX, i.e. controlling for mutable state. e.g. we don't reset any distribution state after construction and it will be nice if we could maintain this for transformed distributions.

We can probably wrap this within our handler classes so that for transformed distribution, we pass along both `x, y` transparently, without having to resort to caching. Let me see if I can put something up for discussion.

fehiepsi:
> We can probably wrap this within our handler classes

Yup, I have thought about using handler but couldn't come up with a solution. Waiting to see your resolution for it. :)

numpyro/distributions/flows.py
```diff
@@ -78,6 +89,9 @@ def log_abs_det_jacobian(self, x, y):
         :param y: the output of the transform
         :type y: numpy array
         """
-        log_scale = self.arn.apply_fun(self.params, x)[..., 1, :]  # TODO: this shouldn't be recomputed here
-        log_scale = _clamp_preserve_gradients(log_scale, self.log_scale_min_clip, self.log_scale_max_clip)
+        if self.caching:
+            log_scale = self._cached_log_scale
```
neerajprad: Okay, I see the issue that you mentioned. Yes, we'll need to pass in `intermediates` to `log_abs_det_jacobian`. Basically, `TransformedDistribution.sample_with_intermediates` returns a list of `(x, transform_intermediates)` and `transform_intermediates` is passed down to `log_abs_det_jacobian`. We don't need to enforce what `transform_intermediates` is - it could be an array, list or None (by default). WDYT?

fehiepsi: That makes sense to me. I'll follow that and make a separate PR which can resolve caches for log_det method.

neerajprad: The changes look good to me otherwise! I'll take another pass once this is resolved.

fehiepsi: @neerajprad I just remove caching stuff and incorporate intermediates stuff. For a test of IAF guide, I'll do it at #224. The changes here only exercise the test `test_transformed_distribution_intermediates`, which confirms the adding of `call_with_intermediates/log_abs_det_jacobian_with_intermediates` method do not affect current behaviours of transformed distributions.

neerajprad: Great, taking a look at this.

numpyro/distributions/constraints.py
```diff
@@ -189,6 +189,15 @@ def inv(self, y):
     def log_abs_det_jacobian(self, x, y):
         raise NotImplementedError
 
+    def call_with_intermediates(self, x):
+        return self(x), None
+
+    def log_abs_det_jacobian_with_intermediates(self, x, y, intermediates=None):
```
neerajprad: Since this function has the same return type as `log_abs_det_jacobian`, my suggestion would be to just have intermediates as an optional kwarg there `log_abs_det_jacobian(x, y, intermediates=None)`, instead of defining a new method. If the transforms has a default `call_with_intermediates`, the `intermeadiates=None` kwarg will work just fine.

fehiepsi: Thanks! I just update it.

Issue #216, opened 2019-06-21
Timeseries model does not work with new version of `lax.scan`


fehiepsi:
With jax 0.1.35, the model works well but it no longer works with newer versions of JAX. In particular, these models are blocked by https://github.com/google/jax/issues/790.

fehiepsi:
We can use the trick at https://github.com/google/jax/issues/790#issuecomment-506867835 to resolve this problem. However, I am noticing that NUTS is spending way more number of steps than before (so it is a few time slower than before). So we should identify the source of this problem.

PR #235, opened 2019-07-08
Work around lax.scan bug in ts forecasting tutorial

fehiepsi:
This PR adds a workaround for the issue https://github.com/pyro-ppl/numpyro/issues/216. It seems that inital values of `lax.scan` should not be scalars.

In addition, I remove setting `target_accept_prob=0.9` and `max_tree_depth=12`. For some reasons, I can't replicate exactly the old result (despite that I have gone back to the old version of jax/jaxlib with numpyro 0.1.0) and it is a bit slow (takes more than 6 minutes) to run inference with these settings (num_steps usually get the value ~4000). Use the default setting, we get a little bit better Rhat/sMAPE/rmse, so I made the change.

TODO:
- [x] add num_chains. Currently, blocked by #236.


fehiepsi:
The change here is ready to review now. Due to [this issue](https://github.com/google/jax/issues/1172) (not related to the change in this PR), the model is about 4x slower than before. :(

neerajprad:
> Due to this issue (not related to the change in this PR), it took a few minutes just to find initial params. :(

The JAX issue doesn't seem to be a regression. Any idea why this is taking longer now than with the earlier versions of JAX?

fehiepsi:
Now we run a while_loop to check the validity of initial params and a vmap over while_loop for multi-chains. These ops incur compiling the process of generating random latent values, which is slow for this model.

fehiepsi:
@neerajprad Matt suggested to use `random.split(key, 10)` for 10 latent sites to avoid the issue of compiling many random.split ops. That would require us revise `seed` handler. Maybe we can allow an additional arg `rng_map` for `seed`, when `rng_map` is a dictionary with keys are site names and values are random states (each for each site). In find initial params function, we will do something like
```
rng_seed, rng_prototype = random.split(rng)
prototype_trace = trace(seed(model, rng_prototype))

    # in a jitted function
    rngs = random.split(rng_seed, len(prototype_trace))
    rng_map = dict(zip(prototype_trace.keys(), rngs))
    model_trace = trace(seed(model, rng_map=rng_map)).get_trace(...)
```
However, currently we do the following
```
seeded_model = substitute(model, substitute_fn=block(seed(init_strategy, subkey)))
```
Do you have any idea on how we can use `rng_map` to get `seeded_model` with `init_strategy`?

neerajprad:
> Maybe we can allow an additional arg rng_map for seed

It is possible that we end up calling split again for the same sample site, but this will reduce the number of split calls. If there aren't any better alternatives, I think a nice solution would just be to hold a cache of rng seeds in the handler and use it. 

If that's the only source of slowdown currently, we can handle this in a separate PR.

fehiepsi:
@neerajprad Matt already made a quick enhancement for `random.split` so the time to find 4 sets of initial params is around 48s, which is still slow but much faster than the previous version (4x faster). With that fix, I think we don't need to revise `seed` handler anymore.

In finding initial valid params, I change `vmap(single_chain_init)(rng)` to `lax.map(single_chain_init, rng)` to make it faster to compile `find_valid_initial_params` method. With that change, time to find 4 sets of initial params is reduced from 48s -> 35s. I think I can reduce that number to 20s with some changes in `find_valid_initial_params` method.

fehiepsi:
With the latest changes, time to find 4 sets of initial params is reduced to 19s (it already took 15s to find initial params in case num_chains=1). This is already good to me. :D

neerajprad:
You mentioned that its taking 6 mins to run inference now, and I'm guessing that initialization only accounts for a part of this slowdown. Are there other undiscovered issues still?

fehiepsi:
> its taking 6 mins to run inference now

hmm, this is with max_tree_depth=12, target_accept_prob=0.9. I have not checked it again but the run_inference now just takes more than 1 minute (faster than before), even with num_chains=4. I'll change the setting back.

fehiepsi:
@neerajprad Yes, the issue is still there with `target_accept_prob=0.9`. With that target accept prob, step size is smaller than before, trees are larger than before, and n_eff is pretty small.

neerajprad:
>  Yes, the issue is still there with target_accept_prob=0.9. With that target accept prob, step size is smaller than before, trees are larger than before, and n_eff is pretty small.

Thanks for explaining. I'll also take a look at this when I have some time.

Issue #236, opened 2019-07-08
initialize_model should return valid initial params


fehiepsi:
Currently, we don't have a mechanism as in Pyro to detect this. So it is nice to detect this.

I face this issue while working with ts forecasting tutorial, where I want to support multi-chains for that model. At first, mcmc just run forever with `num_chains=4`. So I start to debug which chain causes trouble. I discover that for some random seed, `potential_fn(init_params)` can be `nan`. But that is not the end of the story. Assume that we find initial params such that its potential energy is finite, there can be a chance that its grad gets `nan` or `inf`. This in turn will cause `find_reasonable_stepsize` loop forever. The reason is with whatever step_size, after taking a leapfrog move, the new position's energy will always be `nan` (e.g. `grad(pe) = nan` -> `r_next = r + grad(pe) = nan`).

NB: the non-prim versions of control flows in `numpyro.util` are very handy. Without them, it is pretty hard to debug this kind of issue in JAX.

PR #237, opened 2019-07-08
Try to find valid inital parameters

fehiepsi:
Addresses #236.

Tests:
+ Current examples give the same result.
+ The time series forecasting tutorial works well with this change.


neerajprad:
We are getting a new exception - "NotImplementedError: Batching rule for 'staged_expit' not implemented"

fehiepsi:
Yup, the exception is strange because it is not triggered previously under `vmap`. I'll investigate more about this issue. Looking like those custom transforms do not play well inside a `while_loop` + `vmap`.

fehiepsi:
@neerajprad The issue is fixed in the master version of jax.

numpyro/hmc_util.py
```diff
@@ -775,9 +775,30 @@ def single_chain_init(key, only_params=False):
                     potential_energy(seeded_model, model_args, model_kwargs, inv_transforms),
                     jax.partial(transform_fn, inv_transforms))
 
+    _, potential_fn, constrain_fn = single_chain_init(rng if rng.ndim == 1 else rng[0])
+
+    def single_chain_init_validated(key):
```
neerajprad: Is it possible to make this a separate external function `find_valid_initial_params(key, potential_fn, strategy='uniform', prior_params=None)` which can be called from within `single_chain_init` by default? That way, we can also repurpose it for autoguides.

fehiepsi: For `uniform` strategy, it is quite straightforward. But for `prior` strategy, we have to get traces from the model. I think that I can revise the logic here a bit to match your suggestion.

fehiepsi: @neerajprad I have implemented `find_valid_initial_params` in `infer_util` together with various init_strategy we have so far: `feasible`, `uniform`, `prior`, `median`. Looking like it works as expected (modulo issue #241). Does this align with your suggestion?

neerajprad: This is looking great!

numpyro/contrib/autoguide/__init__.py
```diff
@@ -142,31 +108,31 @@ class AutoContinuous(AutoGuide):
...
         super(AutoContinuous, self)._setup_prototype(*args, **kwargs)
+        # FIXME: without block statement, get AssertionError: all sites must have unique names
+        init_params, is_valid = block(find_valid_initial_params)(self.rng, self.model, *args,
+                                                                 init_strategy=self.init_strategy,
+                                                                 **kwargs)
         self._inv_transforms = {}
-        unconstrained_sites = {}
         for name, site in self.prototype_trace.items():
             if site['type'] == 'sample' and not site['is_observed']:
                 # TODO: handle dynamic support
```
fehiepsi: @neerajprad I intended to handle dynamic support here but the situation is more complicated, which requires we change the logic of SVI. SVI also have constrain_fn, which is static after `svi_init` method. So I would like to defer it to another PR. Maybe we should do the same thing as in HMC for SVI (which is to use `log_density` with `skip_dist_transforms=True`). Here inv_transforms still refers to transform to `site['fn'].support`, rather than `site['fn'].base_dist.support`.

fehiepsi: Hmm, in SVI, we initialize `param` handler by its init value, rather than from `init_strategy`, so I think we need to handle dynamic for SVI first, then look into this PR later.

fehiepsi:
@neerajprad I think that this PR is good to review now. The changes in this PR are
+ expose `latent_size` in autoguide.
+ make `substitute(..., substitute_fn)` substitute base values (instead of final values) and init_strategy generates base values (generating base values and generating final values are somehow equivalent because e.g. transform of median of base values is median of final values - if num_samples in `init_to_median` is an odd number). The reason is to make it easier to get unconstrained parameters for potential_energy. If we substitute final values, no intermediate values will be available in `model_trace`, hence it will be more complicated to get unconstrained values.
+ implement `find_valid_initial_params` in `infer_util` (mostly based on the old code in `initialize_model`). Because this function uses `potential_energy`, `potential_energy` is moved to `infer_util` too (to avoid cyclical import).

numpyro/handlers.py
```diff
@@ -391,9 +391,18 @@ def process_message(self, msg):
                     else:
                         msg['value'] = self.base_param_map[msg['name']]
         elif self.substitute_fn is not None:
-            value = self.substitute_fn(msg)
-            if value is not None:
-                msg['value'] = value
+            base_value = self.substitute_fn(msg)
+            if base_value is not None:
+                if msg['type'] == 'sample':
+                    msg['value'], msg['intermediates'] = msg['fn'].transform_with_intermediates(
```
neerajprad: What happens when the given site is not a `TransformedDistribution`? Then `msg['fn'].trans..` will throw an error, it seems.

neerajprad: Maybe we should add a check to that effect, i.e. only do this if `msg['fn']` is a `TransformedDistribution`?

fehiepsi: If it is not a `TransformedDistribution`, then this will return `base_value ` and empty list. I added it [here](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/distributions/distribution.py#L127) in previous PR for `base_param_map`. :D

numpyro/handlers.py
```diff
@@ -391,9 +391,18 @@ def process_message(self, msg):
...
+                        base_value)
+                else:
+                    constraint = msg['kwargs'].pop('constraint', real)
+                    transform = biject_to(constraint)
+                    if isinstance(transform, ComposeTransform):
+                        msg['value'] = ComposeTransform(transform.parts[1:])(base_value)
```
neerajprad: I'm also beginning to think that there was value in our earlier design of also storing `msg['base_value']`, since we are finding more places where we should be working with base values instead of actual values. Nothing related to this PR though. :)

numpyro/handlers.py
neerajprad: Oh..interesting, I forgot that the base Distribution class supports this.

numpyro/handlers.py
fehiepsi: Yes, I think so too. Except for computing KL(p||q) in SVI, the current inference algorithms (HMC and autoguide) works in unconstrained space. Hence base value is preferred.

Actually, our approach is similar to [Automatic Reparameterisation of Probabilistic Programs](https://arxiv.org/pdf/1906.03028.pdf) paper. We don't use additional handler for this job and don't rewrite all of our distributions in term of TransformedDistribution. Users can easily trigger non-center reparameterization by defining their priors as a TransformedDistribution (e.g. TransformedDistribution(Normal(0, 1), AffineTransform(loc, scale))`. I just recognize this recently and like our approach better. :)

fehiepsi: I think that we are having many cool features since the last release. :D

neerajprad: Yes, lets start planning a release after jax's next release.

Issue #238, opened 2019-07-10
Efficient computation of TransformedDistribution.log_prob without caching


neerajprad:
_Originally posted by @fehiepsi in https://github.com/pyro-ppl/numpyro/pull/234#issuecomment-508241392_

Yeah, let's make it clearer about `caching` stuff so we can discuss about better alternatives.
+ In [pytorch](https://github.com/pytorch/pytorch/blob/master/torch/distributions/transformed_distribution.py#L111) and numpyro, the way we calculate a transformed distribution log_prob is
```
x = transform.inv(value)
transformed_dist.log_prob(value) = transform.logdet(x, value) + base_dist.log_prob(x)
```
+ When we use this distribution in `guide`, then the first job is to get a sample `value`. To get `value`, we get a sample `x` from base_dict and set `value = transform(x)`. After that, when we compute `log_prob`, we don't need to recompute the inverse of `value`. That is the reason for having `caching=True`.
+ When we use this distribution in `model`, then the sample `value` is substituted from `guide` (which we don't know which process generates it). To compute log_prob, we need to compute the inverse explicitly (because there is no cached value). That is the scenario for having `caching=False`.

But I am not sure if a `caching` flag is needed. Things might be trickier when InverseTransform is supported in #164.

PR #239, opened 2019-07-10
Use intermediate values for TransformedDistribution.log_prob

neerajprad:
This tries to offer one possible solution to #238, i.e. reusing intermediate values from the forward computation in `TransformedDistribution.log_prob`. I am not completely sure if this will help with the caching issue in #234, but opening this up for discussion.


neerajprad:
> pending your work on log_density

Sorry, I didn't intend to write a new function; changed our existing `log_density` function to handle this.

PR #246, opened 2019-07-16
Remove some gross in arn and iaf

fehiepsi:
Per requested by @martinjankowiak, I revise these classes.

Beside cleaning the code, here are some usage changes:
+ AutoregressiveNN is implemented as a function, instead of a class. This is to match stax module.
+ `param_dims` are moved to the leading dimension, instead of second to last dimension.
+ In case `param_dims=[1, 1]`, we return a tuple of arrays which leading dims are 1 (instead of an array with leading dim is 2). Users can still get the later behaviour by setting `param_dims=[2]`.

There is one regression:
+ When computing the inverse, the update `x = ops.index_update(x, (..., idx), (y[..., idx] - mean) * inverse_scale)` operated at index `idx` will be replaced by `x = (y - mean) * inverse_scale`, hence might be a bit slower. But this regression is pretty small comparing to the forward pass of arn `self.arn(self.params, x)`. The reason is with functional style, we don't have access to `arn`'s `permutation`.

### DIFF
--- a/numpyro/contrib/nn/auto_reg_nn.py
+++ b/numpyro/contrib/nn/auto_reg_nn.py
@@ -1,10 +1,10 @@
 # lightly adapted from https://github.com/pyro-ppl/pyro/blob/dev/pyro/nn/auto_reg_nn.py
 
 from jax import ops, random
+from jax.experimental import stax
 import jax.numpy as np
 
 from numpyro.contrib.nn.masked_dense import MaskedDense
-from numpyro.distributions.util import relu
 
 
 def sample_mask_indices(input_dim, hidden_dim):
@@ -59,7 +59,7 @@ def create_mask(input_dim, hidden_dims, permutation, output_dim_multiplier):
 
 
 def AutoregressiveNN(input_dim, hidden_dims, param_dims=[1, 1], permutation=None,
-                     skip_connections=False, nonlinearity=relu):
+                     skip_connections=False, nonlinearity=stax.Relu):
     """
     An implementation of a MADE-like auto-regressive neural network.
 
@@ -108,15 +108,16 @@ def AutoregressiveNN(input_dim, hidden_dims, param_dims=[1, 1], permutation=None
         if h < input_dim:
             raise ValueError('Hidden dimension must not be less than input dimension.')
 
-    skip_layer = None
-    masked_layers = []
+    net = None
 
     def init_fun(rng, input_shape):
         """
         :param rng: rng used to initialize parameters
         :param input_shape: input shape
         """
-        nonlocal permutation, skip_layer, masked_layers
+        # TODO: consider removing permutation so we can move those layer constructions outside
+        # init_fun. It seems that we can add a PermuteTransform layer to achieve the same effect.
+        nonlocal permutation, net
 
         if permutation is None:
             # By default set a random permutation of variables, which is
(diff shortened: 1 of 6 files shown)


numpyro/contrib/nn/auto_reg_nn.py
```diff
@@ -86,107 +87,86 @@ class AutoregressiveNN(object):
...
+
+    def init_fun(rng, input_shape):
         """
         :param rng: rng used to initialize parameters
         :param input_shape: input shape
         """
-        if self.permutation is None:
+        # TODO: consider removing permutation so we can move those layer constructions outside
+        # init_fun. It seems that we can add a PermuteTransform layer to achieve the same effect.
+        nonlocal permutation, net
```
martinjankowiak: is there no better solution than this `nonlocal` business?

numpyro/distributions/flows.py
```diff
@@ -48,8 +49,7 @@ def __call__(self, x):
         return self.call_with_intermediates(x)[0]
 
     def call_with_intermediates(self, x):
-        out = self.arn.apply_fun(self.params, x)
-        mean, log_scale = out[..., 0, :], out[..., 1, :]
+        mean, log_scale = self.arn(self.params, x)
```
martinjankowiak: 💯

numpyro/contrib/nn/auto_reg_nn.py
fehiepsi: It is tricky I guess. We can do it if we remove permutation=None behavior (so users have to specify permutation unless using the default one). I think it is a good idea to remove permutation=None behavior. What do you think?

neerajprad: The other solution would be to pass them back as return values in `init` and pass them back in in `apply`. Btw, there is little difference in behavior between using classes with fixed attributes (i.e. no change of state outside `__init__`), and using nonlocal in an initialization function to store and set variable values defined at the function scope level. The advantage of doing it this way is that it is easy to see where state setting happens and hard to accidentally mutate something outside of the initialization function. The advantage of using classes is that they are better supported, e.g. sphinx understands it well; likewise method signatures can be introspected easily by IDEs, pickling support for methods, etc.

numpyro/contrib/nn/auto_reg_nn.py
```diff
@@ -58,7 +58,8 @@ def create_mask(input_dim, hidden_dims, permutation, output_dim_multiplier):
     return masks, mask_skip
 
 
-class AutoregressiveNN(object):
+def AutoregressiveNN(input_dim, hidden_dims, param_dims=[1, 1], permutation=None,
+                     skip_connections=False, nonlinearity=stax.Relu):
```
fehiepsi: todo: remove relu in distributions.util
todo: update autoiaf to incorporate changes in arn

numpyro/contrib/nn/auto_reg_nn.py
```diff
@@ -58,7 +58,8 @@ def create_mask(input_dim, hidden_dims, permutation, output_dim_multiplier):
     return masks, mask_skip
 
 
-class AutoregressiveNN(object):
+def AutoregressiveNN(input_dim, hidden_dims, param_dims=[1, 1], permutation=None,
```
neerajprad: nit: Since this isn't a class, can we rename this to `autoregressive_nn` as per convention? It will be really odd otherwise to expect the initialization to give you back two callables.

fehiepsi: I think that `AutoregressiveNN` matches the current convention of [stax](https://github.com/google/jax/blob/master/jax/experimental/stax.py) module: we used stax.Dense, stax.Relu,... despite that they are functions.

neerajprad: I see; it seems that they are using caps for all layer objects. We can just follow their convention then.

neerajprad:
@martinjankowiak - should I merge this?

martinjankowiak:
@neerajprad go ahead

PR #247, opened 2019-07-17
Implement consensus MC and parametric MC algorithms

fehiepsi:
This PR adds a few simple but fast subposterior algorithms.

TODO:
- [x] add tests for these algorithms

### DIFF
--- a/numpyro/hmc_util.py
+++ b/numpyro/hmc_util.py
@@ -792,7 +792,7 @@ def _cov(samples):
     return wc_final(state)[0]
 
 
-def consensus(subposteriors, num_draws, rng=None):
+def consensus(subposteriors, num_draws, diagonal=False, rng=None):
     """
     Merges subposteriors following consensus Monte Carlo algorithm.
 
@@ -802,36 +802,43 @@ def consensus(subposteriors, num_draws, rng=None):
        Steven L. Scott, Alexander W. Blocker, Fernando V. Bonassi, Hugh A. Chipman,
        Edward I. George, Robert E. McCulloch
 
-    :param list subposteriors: a list which each element is a collection of samples.
+    :param list subposteriors: a list in which each element is a collection of samples.
     :param int num_draws: number of draws from the merged posterior.
+    :param bool diagonal: whether to compute weights using variance or covariance, defaults to
+        `False` (using covariance).
     :param jax.random.PRNGKey rng: source of the randomness, defaults to `jax.random.PRNGKey(0)`.
     :return: a collection of `num_draws` samples with the same data structure as each subposterior.
     """
     rng = random.PRNGKey(0) if rng is None else rng
     # stack subposteriors
     joined_subposteriors = tree_multimap(lambda *args: np.stack(args), *subposteriors)
-    # shape = num_subposteriors x num_samples x sample_shape
+    # shape of joined_subposteriors: n_subs x n_samples x sample_shape
     joined_subposteriors = vmap(vmap(lambda sample: ravel_pytree(sample)[0]))(joined_subposteriors)
 
     # randomly gets num_draws from subposteriors
     n_subs = len(subposteriors)
     n_samples = tree_flatten(subposteriors[0])[0][0].shape[0]
+    # shape of draw_idxs: n_subs x num_draws x sample_shape
     draw_idxs = random.randint(rng, shape=(n_subs, num_draws), minval=0, maxval=n_samples)
     joined_subposteriors = vmap(lambda x, idx: x[idx])(joined_subposteriors, draw_idxs)
 
-    # compute weights for each subposterior (ref: Section 3.1 of [1])
-    weights = vmap(lambda x: np.linalg.inv(_cov(x)))(joined_subposteriors)
-    normalized_weights = np.matmul(np.linalg.inv(np.sum(weights, axis=0)), weights)
-
-    # get weighted samples
-    samples_flat = np.einsum('ijk,ilk->lj', normalized_weights, joined_subposteriors)
+    if diagonal:
+        # compute weights for each subposterior (ref: Section 3.1 of [1])
+        weights = vmap(lambda x: (1 - 1 / n_samples) / np.var(x, axis=0))(joined_subposteriors)
+        normalized_weights = weights / np.sum(weights, axis=0)
+        # get weighted samples
+        samples_flat = np.einsum('ij,ikj->kj', normalized_weights, joined_subposteriors)
+    else:
+        weights = vmap(lambda x: np.linalg.inv(_cov(x)))(joined_subposteriors)
+        normalized_weights = np.matmul(np.linalg.inv(np.sum(weights, axis=0)), weights)
+        samples_flat = np.einsum('ijk,ilk->lj', normalized_weights, joined_subposteriors)
 
     # unravel_fn acts on 1 sample of a subposterior
     _, unravel_fn = ravel_pytree(tree_map(lambda x: x[0], subposteriors[0]))
     return vmap(lambda x: unravel_fn(x))(samples_flat)
 
 
-def parametric(subposteriors, num_draws, rng=None):
+def parametric(subposteriors, num_draws, diagonal=False, rng=None, return_params=False):
     """
     Merges subposteriors following (embarrassingly parallel) parametric Monte Carlo algorithm.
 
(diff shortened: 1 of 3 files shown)


fehiepsi:
The errors in Travis are unrelated to this PR. I disable the float64 tests and make a separated issue to address it.

numpyro/hmc_util.py
```diff
@@ -781,3 +782,81 @@ def single_chain_init(key, only_params=False):
...
+    cov = np.linalg.inv(np.sum(weights, axis=0))
+    normalized_weights = np.matmul(cov, weights)
+
+    # comparing to consensus implementation, we compute weighted mean here
+    submeans = np.mean(joined_subposteriors, axis=1)
+    mean = np.einsum('ijk,ik->j', normalized_weights, submeans)
+    samples_flat = dist.MultivariateNormal(mean, cov).sample(rng, (num_draws,))
+
+    _, unravel_fn = ravel_pytree(tree_map(lambda x: x[0], subposteriors[0]))
+    return vmap(lambda x: unravel_fn(x))(samples_flat)
```
martinjankowiak: why return only samples and not instead return the mean/cov as well?

numpyro/hmc_util.py
```diff
@@ -781,3 +782,81 @@ def single_chain_init(key, only_params=False):
...
+    """
+    Merges subposteriors following consensus Monte Carlo algorithm.
+
+    **References:**
+
+    1. *Bayes and big data: The consensus Monte Carlo algorithm*,
+       Steven L. Scott, Alexander W. Blocker, Fernando V. Bonassi, Hugh A. Chipman,
+       Edward I. George, Robert E. McCulloch
+
+    :param list subposteriors: a list which each element is a collection of samples.
```
martinjankowiak: nit: `a list in which...`

numpyro/hmc_util.py
```diff
@@ -781,3 +782,81 @@ def single_chain_init(key, only_params=False):
...
+    rng = random.PRNGKey(0) if rng is None else rng
+    # stack subposteriors
+    joined_subposteriors = tree_multimap(lambda *args: np.stack(args), *subposteriors)
+    # shape = num_subposteriors x num_samples x sample_shape
+    joined_subposteriors = vmap(vmap(lambda sample: ravel_pytree(sample)[0]))(joined_subposteriors)
+
+    # randomly gets num_draws from subposteriors
+    n_subs = len(subposteriors)
+    n_samples = tree_flatten(subposteriors[0])[0][0].shape[0]
+    draw_idxs = random.randint(rng, shape=(n_subs, num_draws), minval=0, maxval=n_samples)
```
martinjankowiak: what's going on here can you please add more comments?

numpyro/hmc_util.py
```diff
@@ -781,3 +782,81 @@ def single_chain_init(key, only_params=False):
...
+    joined_subposteriors = vmap(lambda x, idx: x[idx])(joined_subposteriors, draw_idxs)
+
+    # compute weights for each subposterior (ref: Section 3.1 of [1])
+    weights = vmap(lambda x: np.linalg.inv(_cov(x)))(joined_subposteriors)
```
martinjankowiak: this is a full covariance, right? might we want to use a diagonal cov in some cases?

numpyro/hmc_util.py
fehiepsi: I think we can add an arg `return_draws` to get `mean`/`cov`. The reason to return draws is draws will be used to be transformed back (in flow hmc) to the constrained domain. Or to be used in predictions. But `mean`/`cov` can be useful when users change their mind and want to get more draws without having to rerun this method. So I'll add an option to achieve it.

numpyro/hmc_util.py
fehiepsi: This is to "randomly gets num_draws from subposteriors" (with replacement). For each subposterior, we get `num_draws` draws from it.

numpyro/hmc_util.py
fehiepsi: I just add an optional arg `return_params` for your convenience. If you have other suggestion, please let me know.

numpyro/hmc_util.py
fehiepsi: done, thanks!

numpyro/hmc_util.py
fehiepsi: I just add a comment `# shape of draw_idxs: n_subs x num_draws x sample_shape`. Hope that it can explain the code.

numpyro/hmc_util.py
fehiepsi: done

numpyro/hmc_util.py
```diff
@@ -781,3 +782,106 @@ def single_chain_init(key, only_params=False):
...
+    joined_subposteriors = tree_multimap(lambda *args: np.stack(args), *subposteriors)
+    # shape of joined_subposteriors: n_subs x n_samples x sample_shape
+    joined_subposteriors = vmap(vmap(lambda sample: ravel_pytree(sample)[0]))(joined_subposteriors)
+
+    # randomly gets num_draws from subposteriors
+    n_subs = len(subposteriors)
+    n_samples = tree_flatten(subposteriors[0])[0][0].shape[0]
+    # shape of draw_idxs: n_subs x num_draws x sample_shape
+    draw_idxs = random.randint(rng, shape=(n_subs, num_draws), minval=0, maxval=n_samples)
+    joined_subposteriors = vmap(lambda x, idx: x[idx])(joined_subposteriors, draw_idxs)
```
martinjankowiak: why should `joined_subposteriors` depend on `draw_idxs` instead of all the samples present? is this what they do in the reference?

numpyro/hmc_util.py
```diff
@@ -781,3 +782,106 @@ def single_chain_init(key, only_params=False):
...
+        normalized_weights = np.matmul(cov, weights)
+
+        # comparing to consensus implementation, we compute weighted mean here
+        mean = np.einsum('ijk,ik->j', normalized_weights, submeans)
+        samples_flat = dist.MultivariateNormal(mean, cov).sample(rng, (num_draws,))
+
+    _, unravel_fn = ravel_pytree(tree_map(lambda x: x[0], subposteriors[0]))
+    samples = vmap(lambda x: unravel_fn(x))(samples_flat)
+    if return_params:
+        return samples, mean, cov
```
martinjankowiak: probably better to just return mean and cov by default? (or return a MultivariateNormal)?

martinjankowiak: so users can do `samples, _, _ =...` if they don't care about it

numpyro/hmc_util.py
fehiepsi: The paper only mentioned that we draw num_draws from each subposterior. I guess you wonder about the case these samples are already resampled, so we don’t need to do resample here? That makes sense to me. Currently, numpyro just returns all samples. I don’t need extra codes for this resampling job. So I’ll make num_draws=None by default, which only merges these subposteriors. If num_draws not None, it will do resampling job for each subposterior to get num_draws samples for my convenience.

numpyro/hmc_util.py
fehiepsi: Yup, I’ll make return mean, cov by default if num_draws=None. Otherwise, it will return num_draws samples.

numpyro/hmc_util.py
fehiepsi: done!

numpyro/hmc_util.py
fehiepsi: done, the interface looks more reasonable now, thanks!

fehiepsi:
@martinjankowiak Thanks so much for your thorough review!

> why would you ever want to resample your (expensively obtained) MCMC samples?

This procedure might not be necessary if it is fast to use MCMC samples for predictions. I saw that other frameworks do that way too. For example,
+ [Rlgt package](https://github.com/cbergmeir/Rlgt/blob/master/Rlgt/R/forecast.rlgtfit.R#L45), `NUM_OF_TRIALS` plays resampling (with replacement) role.
+ [PyMC3 sample_posterior_predictive](https://docs.pymc.io/api/inference.html?highlight=sample_ppc#pymc3.sampling.sample_posterior_predictive) which has `samples` parameter to play that role.
+ [Stan posterior_predict](http://mc-stan.org/rstanarm/reference/posterior_predict.stanreg.html) which has `draws` parameter to play that role (but Stan uses resampling without replacements).

In addition, for `consensus`, I want to take advantage of its shuffling effect to not propagate correlation of contiguous samples in each subposterior to the joined posterior (though you might think that it is not necessary).

numpyro/hmc_util.py
```diff
@@ -781,3 +782,112 @@ def single_chain_init(key, only_params=False):
...
+        mean = np.einsum('ij,ij->j', normalized_weights, submeans)
+    else:
+        weights = vmap(lambda x: np.linalg.inv(_cov(x)))(joined_subposteriors)
+        cov = np.linalg.inv(np.sum(weights, axis=0))
+        normalized_weights = np.matmul(cov, weights)
+
+        # comparing to consensus implementation, we compute weighted mean here
+        mean = np.einsum('ijk,ik->j', normalized_weights, submeans)
+
+    if num_draws is not None:
```
neerajprad: nit: It will be nice if we can keep the function return type independent of the arguments. In this case, what do you think about refactoring so that `parametric` returned mean, cov, but we had another helper function called `parametric_draws` that calls parametric and returns `num_draws` samples? This latter function could be used in the tests then. This will avoid boiler-plate, but also keep function types consistent.

numpyro/hmc_util.py
```diff
@@ -781,3 +782,112 @@ def single_chain_init(key, only_params=False):
...
+        samples_flat = np.einsum('ijk,ilk->lj', normalized_weights, joined_subposteriors)
+
+    # unravel_fn acts on 1 sample of a subposterior
+    _, unravel_fn = ravel_pytree(tree_map(lambda x: x[0], subposteriors[0]))
+    return vmap(lambda x: unravel_fn(x))(samples_flat)
+
+
+def parametric(subposteriors, num_draws=None, diagonal=False, rng=None):
+    """
+    Merges subposteriors following (embarrassingly parallel) parametric Monte Carlo algorithm.
```
neerajprad: I think we should expose this in the `contrib` docs, but feel free to defer this to a later PR.

neerajprad:
@fehiepsi - Feel free to merge this at your convenience. My comments aren't blockers, and we can discuss / refactor the interface in separate PRs.

fehiepsi:
Thanks @neerajprad ! I'll incorporate your suggestions and will merge this.

Issue #242, opened 2019-07-12
Handle intermediate values in composed transform


fehiepsi:
Currently, with #239 and #234, we can use intermediate values to avoid recomputing some expensive terms for log_det. We can also revise ComposeTransform to take advantage of the intermediate values. This is especially useful when we take composed transform of flows and use that transform for other inferences such as HMC.

Issue #241, opened 2019-07-11
Handle dynamic support


fehiepsi:
Just discover some edge case which we have not thought about it before. Consider the model
```
def model():
    x = sample('x', dist.Normal(0, 1))
    y = sample('y', dist.Uniform(x, x + 1))
```
The support of `y` site does depend on `x` so `initialize_model` will return different transforms for different initial `rng`.

neerajprad:
Hmm..interesting. This will also be an issue with Pyro, and does seem serious because we will silently give wrong results. Maybe we should just have `transform_fn` take in both constraints and params.

fehiepsi:
Yup, this affects both autoguide and hmc inferences because both of them do inference in unconstrained space. As you said, this is indeed a serious problem. I think that we can resolve it by using non-centered reparametrization technique, that is to define each distribution as a transformed distribution with an affine transform. We do inference in the unconstrained domain of `base_dist`'s support. It seems also have an advantage that the inference brings all variables to a unit scale (hence it is easier to adapt mass matrix). The disadvantage is we have to take 2 steps to get the samples (but it is cheap): first transform the unconstrained value back to base dist support domain, then trace the model again with these in-base_dist-support-domain values. I have not put more thoughts on this but the above solution seems doable. We might leverage handlers to avoid modifying our distribution implementation. And we only need to do it for interval/half-interval domains with constant dist args.

neerajprad:
> I think that we can resolve it by using non-centered reparametrization technique, that is to define each distribution as a transformed distribution with an affine transform. 

You mean the user has to use `TransformedDistribution` in their model or we do it behind the scenes? It seems to me that if it is the latter, that will have the same issue with having to detect the correct loc, scale for distributions with dynamic support? Maybe an example will help clarify.

The alternative seems even trickier because that would involve finding the new transform for each single step of the integrator, and also storing the transform for each sample so that the unconstrained samples can later be transformed into constrained samples.

fehiepsi:
I think we can just do it behind the scene. What I thought is to have log_density computed as follows
```
def potential_energy(params, base_dist_inv_transforms):
    log_joint = 0.

    # step 1:
    base_dist_samples = {}
    for name, value in params.items():
         base_dist_samples[name] = base_dist_inv_transforms(value)

    # step 2:
    # we can modify behaviour of handlers to make this job easy
    model_trace = get_trace(...)  # substitute base_dist sample and get trace

    log_joint = 0.
[...]

PR #248, opened 2019-07-17
Add an example for NeuTra HMC

fehiepsi:
Now, we have AutoIAFNormal merged. It is time to add an example for NeuTra HMC. This PR serves as a boiler plate for further enhancement in `autoguide`, `transforms` modules, so I tried to make it as general (in the sense that the same procedure can be applied for models with many latent sites) as possible. Some of them are already discussed in #241, #242.

TODO:
- [x] make potential_fn with flow transform log det adjustment
- [x] train NUTS
- [x] plot result

Result:
[neutra.pdf](https://github.com/pyro-ppl/numpyro/files/3402580/neutra.pdf)

### DIFF
--- a/examples/neutra.py
+++ b/examples/neutra.py
@@ -57,12 +57,12 @@ def main(args):
 
     print("Start vanilla HMC...")
     # TODO: set progbar=True when https://github.com/google/jax/issues/939 is resolved
-    vanilla_samples = mcmc(10000, 10000, init_params=np.array([2., 0.]), potential_fn=dual_moon_pe,
-                           progbar=False)
+    vanilla_samples = mcmc(args.num_warmup, args.num_samples, init_params=np.array([2., 0.]),
+                           potential_fn=dual_moon_pe, progbar=False)
 
     opt_init, opt_update, get_params = optimizers.adam(0.001)
     rng_guide, rng_init, rng_train = random.split(random.PRNGKey(1), 3)
-    guide = AutoIAFNormal(rng_guide, dual_moon_model, get_params, hidden_dims=[15])
+    guide = AutoIAFNormal(rng_guide, dual_moon_model, get_params, hidden_dims=[args.num_hidden])
     svi_init, svi_update, _ = svi(dual_moon_model, guide, elbo, opt_init, opt_update, get_params)
     opt_state, _ = svi_init(rng_init)
 
@@ -72,11 +72,13 @@ def main(args):
         return i + 1, loss, opt_state_, rng_
 
     print("Start training guide...")
-    losses, opt_states = fori_collect(0, 10000, jit(body_fn), (0, 0., opt_state, rng_train),
+    losses, opt_states = fori_collect(0, args.num_iters, jit(body_fn),
+                                      (0, 0., opt_state, rng_train),
                                       transform=lambda x: (x[1], x[2]), progbar=False)
     last_state = tree_map(lambda x: x[-1], opt_states)
     print("Finish training guide. Extract samples...")
-    guide_samples = guide.sample_posterior(random.PRNGKey(0), last_state, sample_shape=(10000,))
+    guide_samples = guide.sample_posterior(random.PRNGKey(0), last_state,
+                                           sample_shape=(args.num_samples,))
 
     transform = guide.get_transform(last_state)
     unpack_fn = lambda u: guide.unpack_latent(u, transform={})  # noqa: E731
@@ -88,7 +90,7 @@ def main(args):
     # TODO: expose latent_size in autoguide
     init_params = np.zeros(np.size(guide._init_latent))
     print("\nStart NeuTra HMC...")
-    zs = mcmc(10000, 10000, init_params, potential_fn=transformed_potential_fn)
+    zs = mcmc(args.num_warmup, args.num_samples, init_params, potential_fn=transformed_potential_fn)
     print("Transform samples into unwarped space...")
     samples = vmap(transformed_constrain_fn)(zs)
     summary(tree_map(lambda x: x[None, ...], samples))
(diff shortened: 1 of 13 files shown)


martinjankowiak:
can you plot the learned guide as well?

martinjankowiak:
`          mean       std    median      5.0%     95.0%     n_eff     r_hat`
`x[0]      0.05      1.79      0.74     -2.27      2.26      2.84      3.96`
are the bad n_eff r_hat numbers here just indicative of the fact that these are not foolproof metrics?

martinjankowiak:
also can we make this more of a direct comparison between vanilla hmc and neutra? e.g. does the step size get larger? are the trees less deep in nuts?

fehiepsi:
Yup, I'll add additional plots for them shortly. About `n_eff`, I guess the guide is not flexible enough? Let's see what is the result if running mcmc alone without guide.

martinjankowiak:
you might also just make the bimodality of the moons a bit less extreme (if you have that under your control)

fehiepsi:
> also can we make this more of a direct comparison between vanilla hmc and neutra?

@martinjankowiak I just can't run mcmc on this potential_fn (step_size -> nan). I am not sure what's going on here. Even when I set initial params near one of two modes [-2, 0], [2, 0], mcmc is still freeze in the middle of warmup phase.

examples/neutra.py
```diff
@@ -0,0 +1,111 @@
...
+
+    return transformed_potential_fn
+
+
+def main(args):
+    jax_config.update('jax_platform_name', args.device)
+
+    # FIXME: I cann't run mcmc for this potential function
+    # print("Start vanilla HMC...")
+    # vanilla_samples = mcmc(100, 100, init_params=np.array([2., 0.]), potential_fn=dual_moon_pe)
```
fehiepsi: I added a FIXME here for further investigate. If this is a hard-to-sample potential_fn, then the current approach is doing well. Otherwise, there should be a problem in our implementation.

neerajprad: I think this is more likely due to some assumption we are making during adaptation. I would expect HMC to at least explore one of the modes depending on initialization.

fehiepsi: Yup, that is likely the case. I’ll make a separate issue for this.

neerajprad:
> are the bad n_eff r_hat numbers here just indicative of the fact that these are not foolproof metrics?

It will be interesting to see what these metrics look like in the warped space. A high autocorrelation along dim 0 seems expected for vanilla hmc on this data, but does this mean that even when we generate samples from the warped space, we are mostly drawing consecutively from one of the modes?

fehiepsi:
> It will be interesting to see what these metrics look like in the warped space.

Good idea, that would give more hints for us!

> visualizing a few trajectories overlaid on top of your density plot.

I think that this is doable for both HMC and NUTS. It would be an excellent application of functional pseudo-random number generator in term of reproducing results (a tree doubles itself at which leaves is already determined by initial random seed, so we don't have to go over transition logic,... - only velocity verlet is enough). I will come up with a simple utility to do this job.

neerajprad:
>  I will come up with a simple utility to do this job.

I think even before we look at the leapfrog trajectory, it will be instructive to plot just a sequence of samples produced by NUTS to see how sampling in the unwarped space looks like. e.g. do we jump between the two modes frequently or explore one first and then the other (which seems to be the case looking at the neff values), etc.

fehiepsi:
@martinjankowiak @neerajprad I think that I have addressed most of your request. Currently, I have to disable progbar for vanilla due to some JAX bug, so you can't observe its logging infos such as step_size (I guess this is not important because we are doing inferences in different spaces) or num_steps. The NeuTra trajectory takes less iterations to jump back and forth between 2 modes. It should be noted that the random generators to initialize params of arn network are changed in my last PR, so `hidden_dims=[20]` does not work any more. I change it to `hidden_dims=[15]` and it looks fine again. Here are the current results
[neutra.pdf](https://github.com/pyro-ppl/numpyro/files/3409283/neutra.pdf)
```
Start vanilla HMC...


                mean       std    median      5.0%     95.0%     n_eff     r_hat
Param:0[0]     -0.32      1.77     -1.32     -2.34      2.20      8.95      1.01
Param:0[1]      0.04      1.23      0.06     -1.86      1.94    436.51      1.01


Start training guide...
Finish training guide. Extract samples...

Start NeuTra HMC...
warmup: 100%|████████████████████| 4000/4000 [00:13<00:00, 307.66it/s, 1 steps of size 8.55e-02. acc. prob=0.79]
sample: 100%|████████████████████| 4000/4000 [00:04<00:00, 871.45it/s, 3 steps of size 8.55e-02. acc. prob=0.85]


                mean       std    median      5.0%     95.0%     n_eff     r_hat
Param:0[0]     -2.53      0.65     -2.85     -3.37     -1.57      3.09      1.97
Param:0[1]      6.59      1.22      7.19      4.80      7.99      2.99      2.05


Transform samples into unwarped space...


                mean       std    median      5.0%     95.0%     n_eff     r_hat
      x[0]     -0.42      1.82     -1.47     -2.44      2.15      3.02      2.11
      x[1]     -0.37      1.07     -0.47     -2.03      1.35    394.05      1.00
```

martinjankowiak:
hmm, the neutra contour lines seem of lower quality than the pure hmc ones?

fehiepsi:
Yes, it seems so, and the result depends on the guide. This is not an example to show that NeuTra is better than vanilla. Rather than that, it is just a sketch for how to implement NeuTra HMC in NumPyro, so we can adapt it for other models. Anyway, I added some args so we can try on these numbers (num_warmup, num_hidden,...) later. I've also set `num_warmup=0` by default because it seems that we don't need to tune adapt mass matrix (assuming that posterior in warped space is unit Gaussian). Anyway, that's just a fun thing to do. We need to do more experiments to evaluate effects of this approach.

[neutra.pdf](https://github.com/pyro-ppl/numpyro/files/3412052/neutra.pdf)

numpyro/contrib/autoguide/__init__.py
```diff
@@ -193,6 +195,14 @@ def __call__(self, *args, **kwargs):
 
         return result
 
+    def sample_posterior(self, rng, opt_state, *args, **kwargs):
+        sample_shape = kwargs.pop('sample_shape', ())
+        latent_size = np.size(self._init_latent)
+        transform = self.get_transform(opt_state)
```
neerajprad: I'm not sure if this pattern will be general enough for the `AutoContinuous` family, e.g. in the case that we have data dependent guides. We may want to consider subclassing this into `AutoTransformed` or something. On the other hand, I have mixed feelings about having deeply nested class hierarchies. 

I'm fine leaving this in for now, but the interface issues will be worth clearing once we have a few more examples.

fehiepsi: You are right. I have not considered data dependent guide. Let's resort this when we have a concrete example.

PR #250, opened 2019-07-17
Fix nan during step size adaptation

### LINKED ISSUE #249: HMC adaptation stuck at warmup phase due to step_size -> nan
A script to replicate this issue
```
import jax.numpy as np
from jax.scipy.special import logsumexp
from numpyro.mcmc import mcmc

def dual_moon_pe(x):
    term1 = 0.5 * ((np.linalg.norm(x, axis=-1) - 2) / 0.4) ** 2
    term2 = -0.5 * ((x[..., :1] + np.array([-2., 2.])) / 0.6) ** 2
    return term1 - logsumexp(term2, axis=-1)

mcmc(1000, 1000, init_params=np.array([2., 0.]), potential_fn=dual_moon_pe)
```

neerajprad:
Fixes #249 by thresholding on minimum value for step size.

### DIFF
--- a/numpyro/hmc_util.py
+++ b/numpyro/hmc_util.py
@@ -248,6 +248,7 @@ def find_reasonable_step_size(potential_fn, kinetic_fn, momentum_generator, inve
         # case for a diverging trajectory (e.g. in the case of evaluating log prob
         # of a value simulated using a large step size for a constrained sample site).
         step_size = (2.0 ** direction) * step_size
+        step_size = np.clip(step_size, a_min=np.finfo(step_size.dtype).tiny)
         r = momentum_generator(inverse_mass_matrix, rng_momentum)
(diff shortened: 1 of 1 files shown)


neerajprad:
While this fixes the particular nan issue, HMC seems to completely fail on the half moons dataset and doesn't even sample from one of the modes. I'm not sure if that's to be expected, or there are other issues with the code.

fehiepsi:
Thanks @neerajprad ! This fixes the issue in my system. :)

fehiepsi:
> HMC seems to completely fail on the half moons dataset and doesn't even sample from one of the modes

I'll try PyMC3 to see if it shows the same phenomenon. I'm not sure if Stan supports general potential fn. If it does, I'll try it too. :)

Issue #256, opened 2019-07-25
Sampling issues with NeutraHMC


neerajprad:
I noticed a couple of issues when refactoring some of the autoguide code:

 - [x] It seems to me that there is a bug in the transformed PE since the `unpack_fn` currently also applies the transform to constrain the samples. As such the original PE function gets constrained instead of unconstrained samples. ~While this can be fixed with the following changes, I am unable to get a good acceptance rate with these changes. @fehiepsi - any idea what's happening here?~

```diff
--- a/numpyro/contrib/autoguide/__init__.py
+++ b/numpyro/contrib/autoguide/__init__.py
@@ -165,7 +165,7 @@ class AutoContinuous(AutoGuide):
                                     unpacked_samples)

         transform = self._inv_transforms if transform is None else {}
-        return transform_fn(self._inv_transforms, unpacked_samples)
+        return transform_fn(transform, unpacked_samples)
```

```diff
--- a/numpyro/infer_util.py
+++ b/numpyro/infer_util.py
@@ -40,5 +40,7 @@ def transform_fn(transforms, params, invert=False):
     :param invert: Whether to apply the inverse of the transforms.
     :return: `dict` of transformed params.
     """
-    return {k: transforms[k](v) if not invert else transforms[k].inv(v)
+    if invert:
+        transforms = {k: v.inv for k, v in transforms.items()}
+    return {k: transforms[k](v) if k in transforms else v
             for k, v in params.items()}
```

 - [x] The autoguide is itself not doing a very good job of covering both the modes. See figure below. I tried out a few seeds but the results don't look too good. Fix/debug.

- [ ] Neutra HMC performs not good with a trained guide.

[neutra.pdf](https://github.com/pyro-ppl/numpyro/files/3433506/neutra.pdf)

fehiepsi:
> The autoguide is itself not doing a very good job of covering both the modes.

If you change the default `num-iters` from 10000 to 100000, you'll get better fit for guide. It seems that your fix also helps improve n_eff and r_hat a lot. The only issue is neutra posterior is not as good as before, but I doubt that it is a side effect of using autoguide transform.

neerajprad:
Good to know, I'll play around with it some more.

fehiepsi:
[...]

PR #258, opened 2019-07-25
Fix transformed PE function in NeutraHMC

neerajprad:
Fix for #256, but doesn't lead to any qualitative improvement in inference due to bad autoguide fit.

### DIFF
--- a/numpyro/infer_util.py
+++ b/numpyro/infer_util.py
@@ -40,5 +40,7 @@ def transform_fn(transforms, params, invert=False):
     :param invert: Whether to apply the inverse of the transforms.
     :return: `dict` of transformed params.
     """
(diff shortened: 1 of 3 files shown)


numpyro/infer_util.py
```diff
@@ -40,5 +40,7 @@ def transform_fn(transforms, params, invert=False):
     :param invert: Whether to apply the inverse of the transforms.
     :return: `dict` of transformed params.
     """
-    return {k: transforms[k](v) if not invert else transforms[k].inv(v)
+    if invert:
+        transforms = {k: v.inv for k, v in transforms.items()}
+    return {k: transforms[k](v) if k in transforms else v
```
fehiepsi: +1

PR #260, opened 2019-07-26
Simplify and refactor autoguide code for additional flexibility

neerajprad:
This is a straight refactor of the existing code: 
 - Code duplication in `_sample_latent`, `sample_posterior` is resolved by using handlers. All internal methods use param/sample primitives,  and external methods for accessing post-inference properties takes in `opt_state` and can wrap these internal methods within an effect handler if needed.
 - An additional `setup` method is introduced that is called the first time `__call__` is run and sets up any necessary state and returns the initial parameters for the guide.
 - A `base_dist` is added to `__call__` and `sample_posterior` which can be used to sample an initial latent, and provide the most basic facility for a data-dependent guide. Currently all our guides assume that this is `Normal(0, 1)` but this can be more general. e.g. in the case of a VAE where it is a normal distribution whose parameters are encoded by a NN. I think some patterns discussed by @fritzo and @stefanwebb in https://github.com/pyro-ppl/pyro/pull/1937 may be useful here later.

### DIFF
--- a/test/contrib/test_autoguide.py
+++ b/test/contrib/test_autoguide.py
@@ -1,7 +1,7 @@
 from numpy.testing import assert_allclose
 import pytest
 
-from jax import lax, random
+from jax import lax, random, jit
 from jax.experimental import optimizers
 import jax.numpy as np
 
@@ -9,6 +9,7 @@ from numpyro.contrib.autoguide import AutoDiagonalNormal, AutoIAFNormal
 import numpyro.distributions as dist
 from numpyro.handlers import sample
 from numpyro.svi import elbo, svi
+from numpyro.util import control_flow_prims_disabled, fori_loop
 
 
 @pytest.mark.parametrize('auto_class', [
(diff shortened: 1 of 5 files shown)


test/contrib/test_autoguide.py
```diff
@@ -65,14 +66,17 @@ def model(data, labels):
...
+    @jit
     def body_fn(i, val):
         opt_state_, rng_ = val
         loss, opt_state_, rng_ = svi_update(i, rng_, opt_state_,
                                             model_args=(data, labels),
                                             guide_args=(data, labels))
         return opt_state_, rng_
 
-    opt_state, _ = lax.fori_loop(0, 1000, body_fn, (opt_state, rng_train))
+    # TODO: Remove once https://github.com/pyro-ppl/numpyro/issues/259 is fixed.
```
neerajprad: The while loop compilation here takes more than 300 seconds, but this is quite fast if we run this with a native for loop.

fehiepsi: Are you using dev version of `jax`? In the new version, we either have to jit body_fn or jit fori_loop.

neerajprad: No, I'm using the release version. The issue is that `while_loop` triggers an xla compilation by default and that takes forever (more than 300 seconds). However, if we replace that with a python for loop, it takes only 11 seconds to run.

PR #263, opened 2019-07-26
Fix some issues related to reuse rngs

fehiepsi:
This PR prevents reusing random keys, which might cause unexpected behaviour. I think we'd better avoid using the pattern `random.split(key, 1)`.

### DIFF
--- a/numpyro/handlers.py
+++ b/numpyro/handlers.py
@@ -325,8 +325,8 @@ class seed(Messenger):
 
     def process_message(self, msg):
         if msg['type'] == 'sample':
-            msg['kwargs']['random_state'] = self.rng
-            self.rng, = random.split(self.rng, 1)
+            self.rng, rng_sample = random.split(self.rng)
+            msg['kwargs']['random_state'] = rng_sample
 
 
 class substitute(Messenger):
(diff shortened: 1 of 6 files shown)


numpyro/handlers.py
```diff
@@ -325,8 +325,8 @@ def __init__(self, fn, rng):
 
     def process_message(self, msg):
         if msg['type'] == 'sample':
-            msg['kwargs']['random_state'] = self.rng
-            self.rng, = random.split(self.rng, 1)
+            self.rng, rng_sample = random.split(self.rng)
+            msg['kwargs']['random_state'] = rng_sample
```
fehiepsi: the issue here is when `msg['fn']` uses `self.rng` key, it might split the key (e.g. in StudentT distribution)

numpyro/svi.py
```diff
@@ -90,12 +90,12 @@ def update_fn(i, rng, opt_state, model_args=(), guide_args=()):
...
         :return: tuple of `(loss_val, opt_state, rng)`.
         """
-        model_init, guide_init = _seed(model, guide, rng)
+        rng, rng_seed = random.split(rng)
+        model_init, guide_init = _seed(model, guide, rng_seed)
         params = get_params(opt_state)
         loss_val, grads = value_and_grad(loss)(params, model_init, guide_init, model_args,
                                                guide_args, kwargs, constrain_fn=constrain_fn)
         opt_state = optim_update(i, grads, opt_state)
-        rng, = random.split(rng, 1)
```
fehiepsi: the issue here is if `rng` is splitted in `_seed`, it might be reused here; we should prevent this

neerajprad:
I think the test failure is just due to noise.

PR #265, opened 2019-07-28
Handling dynamic support

fehiepsi:
Addresses #241. The implementation is based on the discussions at #241.

TODO
- [x] add test to make sure that potential_fn & constrain_fn are the same as before. 
- [x] add test to compare with previous version

I will do convert Uniform/TruncatedNormal/TruncatedCauchy to a transformed distribution in a follow-up PR.

@neerajprad I would like to need your help to revise the logic which I made at `base_param_map`, `transform_with_intermediates`,... Do you have other suggestions?

### DIFF
--- a/numpyro/distributions/distribution.py
+++ b/numpyro/distributions/distribution.py
@@ -158,13 +158,9 @@ class Distribution(object):
 
     def __call__(self, *args, **kwargs):
         key = kwargs.pop('random_state')
-        base_value = kwargs.pop('base_value', None)
         sample_intermediates = kwargs.pop('sample_intermediates', False)
         if sample_intermediates:
-            if base_value is None:
-                return self.sample_with_intermediates(key, *args, **kwargs)
-            else:
-                return self.transform_with_intermediates(base_value)
+            return self.sample_with_intermediates(key, *args, **kwargs)
         return self.sample(key, *args, **kwargs)
 
 
--- a/numpyro/distributions/distribution.py
+++ b/numpyro/distributions/distribution.py
@@ -124,6 +124,9 @@ class Distribution(object):
         """
         return self.sample(key, sample_shape=sample_shape), []
 
+    def transform_with_intermediates(self, base_value):
+        return base_value, []
+
     def log_prob(self, value):
         """
         Evaluates the log probability density for a batch of samples given by
@@ -155,9 +158,13 @@ class Distribution(object):
 
     def __call__(self, *args, **kwargs):
         key = kwargs.pop('random_state')
+        base_value = kwargs.pop('base_value', None)
         sample_intermediates = kwargs.pop('sample_intermediates', False)
         if sample_intermediates:
-            return self.sample_with_intermediates(key, *args, **kwargs)
+            if base_value is None:
+                return self.sample_with_intermediates(key, *args, **kwargs)
+            else:
+                return self.transform_with_intermediates(base_value)
         return self.sample(key, *args, **kwargs)
 
 
@@ -176,17 +183,25 @@ class TransformedDistribution(Distribution):
     arg_constraints = {}
 
     def __init__(self, base_distribution, transforms, validate_args=None):
-        self.base_dist = base_distribution
         if isinstance(transforms, Transform):
-            self.transforms = [transforms, ]
+            transforms = [transforms, ]
         elif isinstance(transforms, list):
             if not all(isinstance(t, Transform) for t in transforms):
                 raise ValueError("transforms must be a Transform or a list of Transforms")
-            self.transforms = transforms
         else:
             raise ValueError("transforms must be a Transform or list, but was {}".format(transforms))
-        shape = self.base_dist.batch_shape + self.base_dist.event_shape
-        event_dim = max([len(self.base_dist.event_shape)] + [t.event_dim for t in self.transforms])
+        # XXX: this logic will not be valid when IndependentDistribution is support;
+        # in that case, it is more involved to support Transform(Indep(Transform));
+        # however, we might not need to support such kind of distribution
+        # and should raise an error if base_distribution is an Indep one
+        if isinstance(base_distribution, TransformedDistribution):
+            self.base_dist = base_distribution.base_dist
+            self.transforms = base_distribution.transforms + transforms
+        else:
+            self.base_dist = base_distribution
+            self.transforms = transforms
+        shape = base_distribution.batch_shape + base_distribution.event_shape
+        event_dim = max([len(base_distribution.event_shape)] + [t.event_dim for t in transforms])
         batch_shape = shape[:len(shape) - event_dim]
         event_shape = shape[len(shape) - event_dim:]
         super(TransformedDistribution, self).__init__(batch_shape, event_shape, validate_args=validate_args)
@@ -206,7 +221,11 @@ class TransformedDistribution(Distribution):
         return x
 
     def sample_with_intermediates(self, key, sample_shape=()):
-        x = self.base_dist.sample(key, sample_shape)
+        base_value = self.base_dist.sample(key, sample_shape)
+        return self.transform_with_intermediates(base_value)
+
+    def transform_with_intermediates(self, base_value):
+        x = base_value
         intermediates = []
         for transform in self.transforms:
             x_tmp = x
--- a/numpyro/handlers.py
+++ b/numpyro/handlers.py
@@ -378,9 +378,8 @@ class substitute(Messenger):
         elif self.base_param_map is not None:
             if msg['name'] in self.base_param_map:
                 if msg['type'] == 'sample':
-                    msg['base_value'] = self.base_param_map[msg['name']]
                     msg['value'], msg['intermediates'] = msg['fn'].transform_with_intermediates(
-                        msg['base_value'])
+                        self.base_param_map[msg['name']])
                 else:
                     msg['value'] = self.base_param_map[msg['name']]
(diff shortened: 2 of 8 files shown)


test/test_mcmc.py
```diff
@@ -84,6 +84,21 @@ def model(labels):
...
+    def model(data):
+        alpha = sample('alpha', dist.Uniform(0, 1))
+        loc = sample('loc', dist.TransformedDistribution(
+            dist.Uniform(0, 1), constraints.AffineTransform(0, alpha)))
+        sample('obs', dist.Normal(loc, 0.1), obs=data)
+
+    data = true_coef + random.normal(random.PRNGKey(0), (1000,))
+    init_params, potential_fn, constrain_fn = initialize_model(random.PRNGKey(2), model, data)
+    samples = mcmc(1000, 1000, init_params, potential_fn=potential_fn, constrain_fn=constrain_fn)
+    assert_allclose(np.mean(samples['loc'], 0), true_coef, atol=0.05)
```
fehiepsi: This test will fail with a large margin if dynamic support is not supported. Depending on the initial value of `alpha`, initial support of `loc` will be (0, alpha). So if the initial `alpha` < 0.8, we can not make `loc` near the true value: 0.9.

numpyro/handlers.py
```diff
@@ -365,21 +365,29 @@ class substitute(Messenger):
...
+        self.base_param_map = base_param_map
         super(substitute, self).__init__(fn)
 
     def process_message(self, msg):
         if self.param_map is not None:
             if msg['name'] in self.param_map:
                 msg['value'] = self.param_map[msg['name']]
+        elif self.base_param_map is not None:
+            if msg['name'] in self.base_param_map:
+                if msg['type'] == 'sample':
```
neerajprad: Do we need to distinguish the `type` of the site? I would imagine that even if its a param statement, we would want this to only modify the `base_value`.

numpyro/handlers.py
```diff
@@ -365,21 +365,29 @@ class substitute(Messenger):
...
+                    msg['base_value'] = self.base_param_map[msg['name']]
```
neerajprad: Rather than modify `__call__` by putting in a `base_value`, how about we restrict the scope of the change to just this handler?

```
msg['value'], msg['intermediates'] = msg['fn'].transform_with_intermediates(msg['base_value'])
```

numpyro/distributions/distribution.py
```diff
@@ -155,9 +158,13 @@ def _validate_sample(self, value):
 
     def __call__(self, *args, **kwargs):
```
neerajprad: Refer to my comment below regarding keeping this method as is and instead using `transform_with_intermediates` in the handler directly.

numpyro/infer_util.py
```diff
@@ -12,17 +12,29 @@ def log_density(model, model_args, model_kwargs, params):
...
     log_joint = 0.
     for site in model_trace.values():
         if site['type'] == 'sample':
             value = site['value']
             intermediates = site['intermediates']
-            log_prob = np.sum(site['fn'].log_prob(value, intermediates) if intermediates
-                              else site['fn'].log_prob(value))
+            if intermediates:
+                if skip_dist_transforms:
+                    log_prob = site['fn'].base_dist.log_prob(intermediates[0][0])
```
neerajprad: This is assuming a certain form for the intermediates, namely, `(val, log_abs_det_jacobian)`. Are we assuming that this is the interface that all `TransformedDistribution` classes will follow?

numpyro/handlers.py
fehiepsi: This seems a bit tricky. We'll need to address `base_value` for `param` site in `apply_stack`. I'll think more about this later. My mind is currently down with `dual_moon_pe` issue. :D

numpyro/handlers.py
fehiepsi: can we do it here, rather that at `apply_stack` level?

numpyro/infer_util.py
fehiepsi: I think subclasses of `TransformedDistribution` should not redefine `transform_with_intermediates` method. Then the pattern `[[val1, transform_inter1], [val2, transform_inter2], ...]` will be consistent.

numpyro/handlers.py
neerajprad: I think this should work as is. Since `msg['value']` is already populated, `apply_stack` will just return without calling `msg['fn']`.

fehiepsi: Thanks, I'll try it. I still do not fully understand how composed handlers work so this might be a good chance to learn. :D

numpyro/infer_util.py
```diff
@@ -12,17 +12,29 @@ def log_density(model, model_args, model_kwargs, params):
     :param dict model_kwargs`: kwargs provided to the model.
     :param dict params: dictionary of current parameter values keyed by site
         name.
+    :param bool skip_transforms: whether to compute log probability of a site
```
neerajprad: This should be renamed to `skip_dist_transforms`.

numpyro/infer_util.py
neerajprad: Hmm..how about checking for `base_value` instead? Then you should be able to use that directly here. `site['fn'].base_dist.log_prob(base_value)`.

numpyro/handlers.py
neerajprad: Note that this behavior is the same as what `replay` would do otherwise with `param_map`, that is it will directly modify `msg['value']`. For `base_param_map`, once we are fixing the base value, it seems to me that we should be able to populate `msg['value']` without any issues since all the parameters should be known by this time (though I might have missed some edge cases). Let me know if you face any issues.

numpyro/infer_util.py
fehiepsi: understood! I will make a change

numpyro/handlers.py
fehiepsi: @neerajprad It is tricky to support `improper prior with dynamic support`. The reason is: for a site with that type, we are assuming a uniform distribution (in constrained space) on its domain. When the support change, the notion of uniform distribution on the support seems invalid to me.

numpyro/handlers.py
fehiepsi: Thanks! Your suggestion is working well! :D

numpyro/hmc_util.py
```diff
@@ -755,10 +766,16 @@ def single_chain_init(key, only_params=False):
         seeded_model = seed(model, key)
         model_trace = trace(seeded_model).get_trace(*model_args, **model_kwargs)
         constrained_values, inv_transforms = {}, {}
+        has_transformed_dist = False
         for k, v in model_trace.items():
             if v['type'] == 'sample' and not v['is_observed']:
-                constrained_values[k] = v['value']
-                inv_transforms[k] = biject_to(v['fn'].support)
+                if v['intermediates']:
+                    constrained_values[k] = v['intermediates'][0][0]
```
neerajprad: nit: use `v['base_value']` instead of `v['intermediates'][0][0]`.

numpyro/handlers.py
neerajprad: Yes, I think our improper priors with dynamic support wouldn't work (for similar reasons that this PR is trying to solve for). e.g. `pyro.param(2., constraint=constraints.greater_than(mu)`. However, I think there may be a way to specify this by making our param constraint handling more intelligent. Right now its just an identity function, but we can imagine that if there was a base value provided along with a constraint, it would do the appropriate transformation. e.g. in the case above we would have a base value in `[0, 1)`, then apply a transform to convert this to `[0, inf)` and add a location factor of `mu` here. We can probably try this separately.

fehiepsi: That seems reasonable to me. Currently, there are only two constraints which can have dynamic support: `greater_than` and `interval`. Both of them are defined as a ComposeTransform of a base transform and a affine transform (which we can exploit). We can put (improper) uniform prior on the codomain of these base transforms. If the affine transform's parameters are constant, then jacobian adjustment is the same as the current version (up to a constant, which is `log_scale` of affine transform), so we are good with it. Do I understand correctly?

numpyro/hmc_util.py
fehiepsi: The `base_value` key is only available when we substitute using `base_param_map`. How about we make `sample_with_intermediates` return a sample together with a dict with two keys:
`base_value` which holds the base value, and `transform_intermediates` which holds transform intermediates, where each transform intermediate is a list of `[output, logdet_intermediate]`? That way, we can set
```
constrained_values[k] = v['intermediates']['base_value']
```
and in the handler, we don't need to add the key `base_value` for our nodes.

numpyro/handlers.py
neerajprad: Yes, that sounds right.

numpyro/hmc_util.py
neerajprad: Sorry, I missed that we wouldn't have `base_value` populated here. I think your current design is reasonable in that case and we can stick with this. 

> and in the handler, we don't need to add the key base_value for our nodes.

I think we can do this even right now by using `intermediates[0][0]` as you are doing here (since `value` is populated inside `substitute` now and not split between `substitute` and `apply_stack`), but I am fine with keeping a redundant `base_value` in the trace dict if you think that's useful.

numpyro/handlers.py
fehiepsi: This feature will be nice to have. I'll address it in a followup PR.

PR #266, opened 2019-07-29
Make fori_collect work with int type.

### LINKED ISSUE #252: fori_collect does not work with non-float arrays
The issue is ravel_pytree which calls `vjp` and does not work with int32 or bool arrays. As such, `fori_collect` will fail if the default identity transform is used.

fehiepsi:
This is an attempt to fix #252 by patching JAX.

### DIFF
--- a/numpyro/patch.py
+++ b/numpyro/patch.py
@@ -1,3 +1,6 @@
+import jax
+
+
 def patch_dependency(target, root_module):
     parts = target.split('.')
     assert parts[0] == root_module.__name__
@@ -15,3 +18,9 @@ def patch_dependency(target, root_module):
         return new_fn
 
     return decorator
+
+
+@patch_dependency('jax.api._check_inexact_input_vjp', jax)
+def _check_inexact_input_vjp(x):
+    # allow inexact inputs (int/bool) to reverse-mode differentiation
+    pass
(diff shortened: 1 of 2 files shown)


numpyro/patch.py
```diff
@@ -15,3 +18,9 @@ def decorator(new_fn):
         return new_fn
 
     return decorator
+
+
+@patch_dependency('jax.api._check_inexact_input_vjp', jax)
```
neerajprad: I am a bit hesitant to patch over and bypass this check, as it might lead to hard to diagnose bugs in the future. I also think that the JAX team probably didn't intend to expose `ravel_pytree` for external usage. How about we just define our own `ravel_pytree` in `numpyro.util` and not use `vjp`? 

```python
def ravel_list(*leaves):
    ravel_leaves = [np.ravel(elt) for elt in leaves] if leaves else [np.array([])]
    len_leaves = [len(l) for l in ravel_leaves]
    unravel_list = lambda arr: [arr[i: i+l] for i, l in enumerate(len_leaves)]
    return np.concatenate(ravel_leaves), unravel_list


def ravel_pytree(pytree):
    leaves, treedef = tree_flatten(pytree)
    flat, unravel_list = ravel_list(*leaves)
    unravel_pytree = lambda flat: tree_unflatten(treedef, unravel_list(flat))
    return flat, unravel_pytree
```

We can also cache that if it is being called frequently, though I'm not sure if we need to do that.

fehiepsi: @neerajprad Could you make a PR for your solution? Your code sounds reasonable to me and I agree that we should go for a solution like yours.

neerajprad: Sure, I'll send in a PR shortly.

neerajprad: Well my snippet above isn't entirely correct, and this turned out to be more complex, e.g. when we have mixed types - bools, ints, etc. I should have something up soon though.

PR #267, opened 2019-07-29
 Implement ravel_pytree for internal usage

### LINKED ISSUE #252: fori_collect does not work with non-float arrays

neerajprad:
Fixes #252, based on discussion in #266.

### DIFF
--- a/numpyro/util.py
+++ b/numpyro/util.py
@@ -1,3 +1,4 @@
+from collections import namedtuple
 from contextlib import contextmanager
 import random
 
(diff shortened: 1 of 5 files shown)


numpyro/util.py
```diff
@@ -171,3 +172,27 @@ def decorator(destin_class):
...
+
+
+def _ravel_list(*leaves):
+    leaves_metadata = tree_map(lambda l: pytree_metadata(np.ravel(l), np.shape(l), np.size(l), lax.dtype(l)),
+                               leaves)
+    leaves_idx = np.cumsum(np.array((0,) + tuple(d.size for d in leaves_metadata)))
+
+    def unravel_list(arr): return [np.reshape(lax.dynamic_slice_in_dim(arr, leaves_idx[i], m.size),
+                                              m.shape).astype(m.dtype)
+                                   for i, m in enumerate(leaves_metadata)]
```
fehiepsi: beautiful!!

numpyro/util.py
```diff
@@ -171,3 +172,27 @@ def decorator(destin_class):
...
+
+pytree_metadata = namedtuple('pytree_metadata', ['flat', 'shape', 'size', 'dtype'])
+
+
+def _ravel_list(*leaves):
+    leaves_metadata = tree_map(lambda l: pytree_metadata(np.ravel(l), np.shape(l), np.size(l), lax.dtype(l)),
+                               leaves)
+    leaves_idx = np.cumsum(np.array((0,) + tuple(d.size for d in leaves_metadata)))
+
+    def unravel_list(arr): return [np.reshape(lax.dynamic_slice_in_dim(arr, leaves_idx[i], m.size),
```
fehiepsi: nit: I think it is easier to read if we start with new line (or use `unravel_list = lambda arr: ...`)

numpyro/util.py
```diff
@@ -171,3 +172,27 @@ def decorator(destin_class):
...
+                                   for i, m in enumerate(leaves_metadata)]
+
+    return np.concatenate([m.flat for m in leaves_metadata]), unravel_list
+
+
+def ravel_pytree(pytree):
+    leaves, treedef = tree_flatten(pytree)
+    flat, unravel_list = _ravel_list(*leaves)
+
+    def unravel_pytree(arr): return tree_unflatten(treedef, unravel_list(arr))
```
fehiepsi: nit: having return statement in newline

neerajprad:
> Could you also update fori_collect using this?

`fori_collect` will just pick it up since its defined in the same file, and we aren't importing `ravel_pytree` from jax in that file.

numpyro/util.py
neerajprad: Done.

numpyro/util.py
neerajprad: Done

Issue #251, opened 2019-07-17
HMC performs poorly for dual moon potential_fn when progbar=True


fehiepsi:
This issue is noticed by @neerajprad . The script
```
import jax.numpy as np
from jax.scipy.special import logsumexp
from numpyro.mcmc import mcmc

def dual_moon_pe(x):
    term1 = 0.5 * ((np.linalg.norm(x, axis=-1) - 2) / 0.4) ** 2
    term2 = -0.5 * ((x[..., :1] + np.array([-2., 2.])) / 0.6) ** 2
    return term1 - logsumexp(term2, axis=-1)

samples = mcmc(4000, 4000, init_params=np.array([2., 0.]), potential_fn=dual_moon_pe)
```
gives
```
                mean       std    median      5.0%     95.0%     n_eff     r_hat
Param:0[0]      1.59      0.00      1.59      1.59      1.59      0.50      1.00
Param:0[1]     -0.58      0.00     -0.58     -0.58     -0.58      4.49      1.10
```

fehiepsi:
Initial observation shows that the parameters stuck at the mean point.

neerajprad:
I don't necessarily see parameters stuck at the mean point, but they are mostly confined to various regions where  I am guessing that the leapfrog trajectory diverges. By fixing the step size in HMC, I can see greater progress, but it is still very bad. 

```python
samples = mcmc(1000, 10000, init_params=np.array([1., -2.]), potential_fn=dual_moon_pe,
               algo='HMC', trajectory_length=0.3, step_size=1e-5, adapt_step_size=False)
```

[vanilla.pdf](https://github.com/pyro-ppl/numpyro/files/3404317/vanilla.pdf)

fehiepsi:
@neerajprad I made a Stan code to see how it performs. It is surprised that Stan did a good job for sampling this distribution.
```
from pystan import StanModel

model_code = """
parameters {
  real x1;
  real x2;
}
model {
  target += -0.5 * pow((sqrt(pow(x1, 2) + pow(x2, 2)) - 2) / 0.4, 2);
  target += log_sum_exp(-0.5 * pow((x1 - 2) / 0.6, 2), -0.5 * pow((x1 + 2) / 0.6, 2));
}
"""

m = StanModel(model_code=model_code)
fit = m.sampling(chains=1, iter=10000)
print(fit.stansummary())
x1 = fit.extract()['x1']
x2 = fit.extract()['x2']
```
[...]

PR #268, opened 2019-07-30
Update travis to use JAX master

### LINKED ISSUE #241: Handle dynamic support

fehiepsi:
This is to fix and unblock various issues and PRs: fixes #241, #251, unblock #235, #237

The new jaxlib version `0.1.22` makes it impossible for me to train AutoIAFNormal (or AutoDiagonalNormal) on dual moon model under fastmath mode. So I add a warning for it.

I have tuned NeutraHMC args a bit so we get a better guide (see image below). However, the NeuTra samples are not good, so I added this issue to #256.

[neutra.pdf](https://github.com/pyro-ppl/numpyro/files/3445521/neutra.pdf)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ dist: xenial
 install:
     - pip install -U pip
     - pip install jaxlib
-    - pip install jax
+    - pip install -e git+https://github.com/google/jax.git@5861e37edd82ec5bd737fdb33b1ac6f2684e4827#egg=jax
     - pip install .[examples,test]
     - pip freeze
 
--- a/setup.py
+++ b/setup.py
@@ -32,7 +32,7 @@ setup(
     install_requires=[
         # TODO: pin to a specific version for the next release (unless JAX's API becomes stable)
         'jax @ git+https://github.com/google/jax.git@5861e37edd82ec5bd737fdb33b1ac6f2684e4827#egg=jax',
-        'jaxlib>=0.1.18',
+        'jaxlib>=0.1.22',
         'tqdm',
     ],
     extras_require={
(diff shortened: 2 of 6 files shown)


.travis.yml
```diff
@@ -15,7 +15,7 @@ dist: xenial
 install:
     - pip install -U pip
     - pip install jaxlib
-    - pip install jax
+    - pip install -e git+https://github.com/google/jax.git@5861e37edd82ec5bd737fdb33b1ac6f2684e4827#egg=jax
```
neerajprad: Are the changes here backwards compatible w.r.t. the latest JAX release? If not, we should probably also change `setup.py` so that users working on NumPyro's master branch are not left with a broken setup.

fehiepsi: It is not backward compatible, so I just update the setup file.

setup.py
```diff
@@ -31,7 +31,7 @@
     author_email='npradhan@uber.com',
     install_requires=[
         # TODO: pin to a specific version for the next release (unless JAX's API becomes stable)
-        'jax>=0.1.37',
+        'jax @ git+https://github.com/google/jax.git@5861e37edd82ec5bd737fdb33b1ac6f2684e4827#egg=jax',
         'jaxlib>=0.1.18',
```
neerajprad: Do we also need to update jaxlib?

fehiepsi: yup :D

PR #269, opened 2019-07-31
implement call with intermediates for compose transform

fehiepsi:
Address #242.

### DIFF
--- a/numpyro/distributions/constraints.py
+++ b/numpyro/distributions/constraints.py
@@ -283,10 +283,7 @@ class ComposeTransform(Transform):
             x = y_tmp
         # account the the last transform, where y is available
         inter = None if intermediates is None else intermediates[-1]
-        if inter is None:
-            logdet = self.parts[-1].log_abs_det_jacobian(x, y)
-        else:
-            logdet = self.parts[-1].log_abs_det_jacobian(x, y, intermediates=inter)
+        logdet = self.parts[-1].log_abs_det_jacobian(x, y, intermediates=inter)
         result = result + sum_rightmost(logdet, self.event_dim - self.parts[-1].event_dim)
         return result
 
--- a/numpyro/distributions/constraints.py
+++ b/numpyro/distributions/constraints.py
@@ -269,15 +269,39 @@ class ComposeTransform(Transform):
         return y
 
     def log_abs_det_jacobian(self, x, y, intermediates=None):
+        if intermediates is not None:
+            if len(intermediates) != len(self.parts):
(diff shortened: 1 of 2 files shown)


numpyro/distributions/constraints.py
```diff
@@ -269,16 +269,40 @@ def inv(self, y):
...
-            y_tmp = part(x)
-            result = result + sum_rightmost(part.log_abs_det_jacobian(x, y_tmp),
-                                            self.event_dim - part.event_dim)
+        for i, part in enumerate(self.parts[:-1]):
+            y_tmp = part(x) if intermediates is None else intermediates[i][0]
+            inter = None if intermediates is None else intermediates[i][1]
+            if inter is None:
+                logdet = part.log_abs_det_jacobian(x, y_tmp)
+            else:
+                logdet = part.log_abs_det_jacobian(x, y_tmp, intermediates=inter)
```
neerajprad: It would be nice if we could call `.log_abs_det_jacobian()` with `intermediates=None` for any transform. Then we wouldn't need this if / else branch. 

```diff
-            if inter is None:
-                logdet = part.log_abs_det_jacobian(x, y_tmp)
-            else:
-                logdet = part.log_abs_det_jacobian(x, y_tmp, intermediates=inter)
+            logdet = part.log_abs_det_jacobian(x, y_tmp, intermediates=inter)
```

We could add `intermediates=None` kwarg to `Transform.log_abs_det_jacobian` method (like `TransformedDistribution.log_prob`). That will require carrying a redundant `intermediates=None` kwarg for transforms that don't cache intermediates. Alternatively, we could have `log_abs_det_jacobian_(x, y, intermediates)` defined in `Transform` which defaults to `log_abs_det_jacobian(x, y)`. 

This is just a nit, and not a blocker.

fehiepsi: @neerajprad Can we have a separate `log_abs_det_jacobian_with_intermediates`? This pattern is appeared in `TransformedDistribution` too.

neerajprad: I'm still not sure what would be better - whether to have `intermediates=None` as a default kwarg to `log_abs_det_jacobian` which makes it easy to share code (like `TransformedDistribution.log_prob`) or to have a separate method which can have a default implementation in  `Transform`. How about just adding a `transform=None` like we do for flows? I'm fine with having a separate method too if you think that'll be cleaner.

fehiepsi: I think that having `intermediates=None` as a default kwarg would be fine. I'll go with it. :D

neerajprad:
Looks great!

PR #270, opened 2019-07-31
Convert TruncatedCauchy, TruncatedNormal, Uniform to transformed distribution

fehiepsi:
As a follow-up PR of #265, currently, only these distributions have dynamic support.

Because Abs is not a bijective transform, I have converted HalfNormal, HalfCauchy to a subclass of Distribution, not TransformedDistribution.

### DIFF
--- a/numpyro/distributions/continuous.py
+++ b/numpyro/distributions/continuous.py
@@ -30,7 +30,7 @@ from jax.scipy.linalg import solve_triangular
 from jax.scipy.special import gammaln, log_ndtr, ndtr, ndtri
 
 from numpyro.distributions import constraints
-from numpyro.distributions.constraints import AbsTransform, AffineTransform, ExpTransform, PowerTransform
+from numpyro.distributions.constraints import AffineTransform, ExpTransform, PowerTransform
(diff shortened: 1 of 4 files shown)


numpyro/distributions/continuous.py
```diff
@@ -715,35 +715,44 @@ def variance(self):
...
 
     def log_prob(self, value):
         if self._validate_args:
             self._validate_sample(value)
-        low = (self.low - self.loc) / self.scale
         # pi / 2 is arctan of self.high when that arg is supported
-        normalize_term = np.log(np.pi / 2 - np.arctan(low)) + np.log(self.scale)
-        return - np.log1p(((value - self.loc) / self.scale) ** 2) - normalize_term
+        normalize_term = np.log(np.pi / 2 + np.arctan(self.base_loc))
+        return - np.log1p((value - self.base_loc) ** 2) - normalize_term
```
fehiepsi: @neerajprad to verify the equivalent of these formulas, we only need to replace low -> 0, scale -> 1, loc -> base_loc. Anyway, the current tests already covert it.

neerajprad: Thanks, that makes sense.

PR #271, opened 2019-07-31
dynamic support for improper prior

fehiepsi:
As discussed in #265, we can have dynamic support for improper prior.

### DIFF
--- a/numpyro/handlers.py
+++ b/numpyro/handlers.py
@@ -81,6 +81,8 @@ from collections import OrderedDict
 
 from jax import random
 
+from numpyro.distributions.constraints import biject_to, real, ComposeTransform
+
 _PYRO_STACK = []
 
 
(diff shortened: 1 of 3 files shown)


numpyro/handlers.py
```diff
@@ -381,7 +383,13 @@ def process_message(self, msg):
                     msg['value'], msg['intermediates'] = msg['fn'].transform_with_intermediates(
                         self.base_param_map[msg['name']])
                 else:
-                    msg['value'] = self.base_param_map[msg['name']]
+                    base_value = self.base_param_map[msg['name']]
```
neerajprad: This is pretty neat. I'm not sure if this decomposition of `ComposeTransform` into an invariant first transform and a model dependent second transform will always be valid in the future, but works great for all constraints that we have for now!

test/test_mcmc.py
```diff
@@ -100,6 +100,20 @@ def model(data):
...
+    true_coef = 0.9
+
+    def model(data):
+        alpha = sample('alpha', dist.Uniform(0, 1))
+        loc = param('loc', 0., constraint=constraints.interval(0., alpha))
+        sample('obs', dist.Normal(loc, 0.1), obs=data)
+
+    data = true_coef + random.normal(random.PRNGKey(0), (1000,))
+    init_params, potential_fn, constrain_fn = initialize_model(random.PRNGKey(2), model, data)
+    samples = mcmc(1000, 1000, init_params, potential_fn=potential_fn, constrain_fn=constrain_fn)
```
neerajprad: I know this should work, but it still seems surprising that it does. :)

PR #272, opened 2019-08-01
Add some more plots to the NeutraHMC example

neerajprad:
This makes a few changes and adds some plots for debugging the half moons example:
 - Changed the prior to `Unif(-4, 4)` from `Unif(-10, 10)` which seems easier to sample for NeutraHMC. I also tried other priors like Normal and Cauchy, but none were notably better as to warrant the change.
 - Also changed the number of samples to 20k, since we were getting biased results for NeutraHMC with 10k samples.
 - Added some more plots to debug - inc. a plot of the warped posterior (see below).

[neutra.pdf](https://github.com/pyro-ppl/numpyro/files/3458679/neutra.pdf)

**Some observations:**

 - Even though the above measures help NeutraHMC, Vanilla HMC still seems a better choice on this example in terms of the quality of the generated posterior.
 - Not sure how much we can generalize from this example, but the posterior in the warped space isn't very gaussian, definitely more so than the unwarped posterior distribution but it has more of a butterfly shape. Changing the prior has an effect on the shape and can be very far from gaussian.
 - For some reason, NeutraHMC is placing a bulk of the posterior mass lower than where you would expect. Not sure why that's the case.

### DIFF
--- a/examples/neutra.py
+++ b/examples/neutra.py
@@ -37,7 +37,7 @@ def dual_moon_pe(x):
 
 
 def dual_moon_model():
-    x = sample('x', dist.Uniform(-10 * np.ones(2), 10 * np.ones(2)))
+    x = sample('x', dist.Uniform(-4 * np.ones(2), 4 * np.ones(2)))
     pe = dual_moon_pe(x)
     sample('log_density', dist.Delta(log_density=-pe), obs=0.)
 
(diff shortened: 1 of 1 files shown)


martinjankowiak:
hmm, i'm not sure if we should necessarily expect NeutraHMC to do better here. even if the learned guide brings the modes closer together, it also inevitably produces wrinkles and what not. i think it'd be more telling to do a quantitative test in more dimensions where we can compare to results in the paper.

fehiepsi:
Here is what I thought (which is a highly personal view).

> vanilla HMC still seems a better choice on this example

I guess because two modes are near to each other (compare to the variance at each mode), so trajectories can move from one mode to another mode

> the posterior in the warped space isn't very gaussian

This is fine I guess. We tried to minimize KL with the gaussian one q(z) but it does not guarantee that after transforming, p(z) will be gaussian.

> NeutraHMC is placing a bulk of the posterior mass lower than where you would expect

This one has been bugging me. To me, it is highly a side effect of IAF flow but can't understand why. I have not had time to work on BNAF flow to confirm it yet.

> i'm not sure if we should necessarily expect NeutraHMC to do better here

I agree with @martinjankowiak here. It is better to try with more examples, especially with high dimensional problems. I am working on covtype dataset for this purpose.

What I hope from this example is to detect bugs in our approach/code. So we can justify the situation in more realistic problems.

examples/neutra.py
```diff
@@ -100,36 +100,51 @@ def body_fn(val):
...
     ax4.contourf(X1, X2, P, cmap='OrRd')
-    sns.kdeplot(samples['x'][:, 0].copy(), samples['x'][:, 1].copy(), ax=ax4)
-    ax4.plot(samples['x'][-50:, 0], samples['x'][-50:, 1], 'bo-', alpha=0.5)
-    ax4.set(xlim=[-3, 3], ylim=[-3, 3], aspect='equal',
+    sns.kdeplot(vanilla_samples[:, 0].copy(), vanilla_samples[:, 1].copy(), n_levels=30, ax=ax4)
+    ax4.plot(vanilla_samples[-50:, 0], vanilla_samples[-50:, 1], 'bo-', alpha=0.5)
+    ax4.set(xlim=[-3, 3], ylim=[-3, 3],
+            xlabel='x0', ylabel='x1', title='Posterior using vanilla HMC sampler')
+
+    sns.scatterplot(zs[:, 0], zs[:, 1], ax=ax5, hue=samples['x'][:, 0] < 0.)
```
fehiepsi: nit: it might be nice if you can control the transparency of samples e.g. `s=80, alpha=0.5, edgecolor="none"` (where s is size of scatter point, alpha is transparency, edgecolor="none" to remove white border). Please customize them as you like and feel free to merge this afterwards. :)

PR #273, opened 2019-08-02
Dynamic support for svi and autoguide

fehiepsi:
This might be the last piece for handling dynamic supports. However, I found it a bit complicated to handle all usage cases.

@neerajprad Could you pls take a careful look at the logic of this PR? I am scared that I did serious mistakes here and there.

TODO:
- [x] Test autoguide, svi with dynamic support models.
- [x] currently, autoguide test failed, so there should be a bug here.

### DIFF
--- a/numpyro/svi.py
+++ b/numpyro/svi.py
@@ -6,7 +6,7 @@ from jax import random, value_and_grad
 from numpyro.contrib.autoguide import AutoContinuous
 from numpyro.distributions import constraints
 from numpyro.distributions.constraints import biject_to, ComposeTransform
-from numpyro.handlers import replay, seed, substitute, trace, Messenger
+from numpyro.handlers import replay, seed, substitute, trace
 from numpyro.infer_util import log_density, transform_fn
 
 
@@ -35,6 +35,9 @@ def svi(model, guide, loss, optim_init, optim_update, get_params, **kwargs):
     :return: tuple of `(init_fn, update_fn, evaluate)`.
     """
     constrain_fn = None
+    # NB: only skip transforms for AutoContinuous guide
+    loss_fn = jax.partial(loss, skip_dist_transforms=True) if isinstance(guide, AutoContinuous) \
+        else loss
 
     def init_fn(rng, model_args=(), guide_args=(), params=None):
         """
@@ -95,8 +98,8 @@ def svi(model, guide, loss, optim_init, optim_update, get_params, **kwargs):
         rng, rng_seed = random.split(rng)
         model_init, guide_init = _seed(model, guide, rng_seed)
         params = get_params(opt_state)
-        loss_val, grads = value_and_grad(loss)(params, model_init, guide_init, model_args,
-                                               guide_args, kwargs, constrain_fn=constrain_fn)
+        loss_val, grads = value_and_grad(loss_fn)(params, model_init, guide_init, model_args,
+                                                  guide_args, kwargs, constrain_fn=constrain_fn)
         opt_state = optim_update(i, grads, opt_state)
         return loss_val, opt_state, rng
 
@@ -115,7 +118,7 @@ def svi(model, guide, loss, optim_init, optim_update, get_params, **kwargs):
         """
         model_init, guide_init = _seed(model, guide, rng)
         params = get_params(opt_state)
-        return loss(params, model_init, guide_init, model_args, guide_args, kwargs, constrain_fn=constrain_fn)
+        return loss_fn(params, model_init, guide_init, model_args, guide_args, kwargs, constrain_fn=constrain_fn)
 
     # Make local functions visible from the global scope once
     # `svi` is called for sphinx doc generation.
@@ -148,7 +151,8 @@ def get_param(opt_state, model, guide, get_params, constrain_fn, rng,
     return params
 
 
-def elbo(param_map, model, guide, model_args, guide_args, kwargs, constrain_fn):
+def elbo(param_map, model, guide, model_args, guide_args, kwargs, constrain_fn,
+         skip_dist_transforms=False):
     """
     This is the most basic implementation of the Evidence Lower Bound, which is the
     fundamental objective in Variational Inference. This implementation has various
@@ -176,11 +180,6 @@ def elbo(param_map, model, guide, model_args, guide_args, kwargs, constrain_fn):
     guide_log_density, guide_trace = log_density(guide, guide_args, kwargs, param_map)
     # NB: we only want to substitute params not available in guide_trace
     param_map = {k: v for k, v in param_map.items() if k not in guide_trace}
-    # NB: only skip transforms for AutoContinuous guide
-    skip_dist_transforms = False
-    guide_fn = guide.fn if isinstance(guide, Messenger) else guide
-    if isinstance(guide_fn, AutoContinuous):
-        skip_dist_transforms = True
     model_log_density, _ = log_density(replay(model, guide_trace), model_args, kwargs, param_map,
                                        skip_dist_transforms=skip_dist_transforms)
     # log p(z) - log q(z)
--- a/numpyro/svi.py
+++ b/numpyro/svi.py
@@ -18,8 +18,6 @@ def _seed(model, guide, rng):
 
 
 def svi(model, guide, loss, optim_init, optim_update, get_params, **kwargs):
-    constrain_fn = None
-
     """
     Stochastic Variational Inference given an ELBo loss objective.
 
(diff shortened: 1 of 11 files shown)


numpyro/svi.py
```diff
@@ -151,7 +155,15 @@ def elbo(param_map, model, guide, model_args, guide_args, kwargs, constrain_fn):
...
-    model_log_density, _ = log_density(replay(model, guide_trace), model_args, kwargs, param_map)
+    # NB: we only want to substitute params not available in guide_trace
+    param_map = {k: v for k, v in param_map if k not in guide_trace}
+    # NB: only skip transforms for AutoContinuous guide
+    skip_dist_transforms = False
+    guide_fn = guide.fn if isinstance(guide, Messenger) else guide
+    if isinstance(guide_fn, AutoContinuous):
+        skip_dist_transforms = True
+    model_log_density, _ = log_density(replay(model, guide_trace), model_args, kwargs, param_map,
+                                       skip_dist_transforms=skip_dist_transforms)
```
fehiepsi: This is where I found the trickiest. I might be making a mistake here. In autocontinuous, we only adjust jacobian from unconstrained_value -> base_value, hence I think that for log_density, we only need to compute log_prob of base distributions.

neerajprad: This seems right to me reg. the jacobian adjustment term. The other issue that we discussed reg. having `param` statements with dynamic parameters should just be handled by your changes to `svi.init_fn` and the `substitute` handler. We can probably just test this with SVI on `test_improper_prior` by moving the param statements to a guide and sampling from the Normal distribution and changing the param statements in the model to sample statements with suitable priors.

fehiepsi: Thanks @neerajprad ! I just add a test for SVI, it works well. Currently, the test for autoguide fails so I'll keep debugging for it.

fehiepsi:
The changes here are non-trivial so I would like to summary them here. Let's consider the triple (`inv_transforms`, `substitute`, `log_density`) for each situation separately
### HMC
+ sample sites with dynamic support:
    - `inv_transforms`: unconstrained values -> base values
+ param sites with dynamic constraint:
    - `inv_transforms`: unconstrained params -> base params
+ substitute: `base_param_map = base_values + base_params`
+ log_density: log prob of base distributions
+ API changes: #265 for sample sites (handling TransformedDistribution at substitute) and #271 for param sites (handling ComposeTransform at substitute)

### SVI
+ sample sites with dynamic support:
    - `inv_transforms`: none
+ param sites with dynamic constraint:
    - `inv_transforms`: unconstrained params -> `base params=base model params + base guide params`
+ guide substitute: `base_param_map = base_params`
+ guide log_density: log prob of full distributions
+ model replay: all `final values` from guide trace, all `final params` from guide trace
+ model substitute: `base_param_map = base_model_params \ base_guide_params`
+ model log_density: log prob of full distributions
+ API changes: `constrain_fn` only transforms unconstrained params to base params. To transform unconstrained params to final params, we use `get_param` function (which will be moved to `svi` in future PR).

### Auto guide
The most complicated one, many API changes here.
+ sample sites with dynamic support:
   - `inv_transforms`: unconstrained values -> base values
+ param sites with dynamic constraint:
   - svi will handle them
+ guide substitute: `base_param_map = loc + scale` (base value of scale is still scale)
+ guide log_density: log prob of base distribution
...
to be filled

fehiepsi:
@neerajprad It seems that things work as expected now. Unless it is important to have docs for some tricky methods (for clarity), I'll make sure that I'll fill all the remaining docs before the next release. We are going to refactor the API soon so I would like to prioritize refactoring first.

neerajprad:
Great! I'll take a look at this shortly.

numpyro/svi.py
```diff
@@ -151,7 +177,17 @@ def elbo(param_map, model, guide, model_args, guide_args, kwargs, constrain_fn):
     """
     param_map = constrain_fn(param_map)
     guide_log_density, guide_trace = log_density(guide, guide_args, kwargs, param_map)
-    model_log_density, _ = log_density(replay(model, guide_trace), model_args, kwargs, param_map)
+    if is_autoguide:
```
neerajprad: I am trying to understand why we need this. Could you elaborate on what goes wrong if we remove this `if` block?

fehiepsi: I have tried to elaborate the difference of autoguide/svi/hmc at [this comment](https://github.com/pyro-ppl/numpyro/pull/273#issuecomment-518078197). Here if we remove `if` block, final values of model will be filled by values from guide (which corresponds to base values of model), which is incorrect. We want to substitute these values from guide to base values of model. In particular (see [this test](https://github.com/pyro-ppl/numpyro/pull/273/files#diff-21afc849214aedbc4266eafd7e3f1483R117))
```
def model():
    x = sample('x', dist.TransformedDistribution(base_dist, transform))
```
will be interpreted as the following model in autoguide
```
def model():
    x_base = sample('x', base_dist)
    x = transform(x_base)
```

neerajprad: That makes sense. I am concerned about leakage of abstraction here though. I think it is important that when we call `autoguide()` we generate samples that are within the model's support and not base values. It seems to me then that the right place to handle this is within the autoguide class themselves, but I will need to run this code myself to have a more concrete proposal. One way would be to use something like [CatTransform](https://github.com/pytorch/pytorch/blob/master/torch/distributions/transforms.py#L562) in `autoguide.__call__` to do this (my initial thought, not sure if it would handle all the cases).

It may also turn out that what you are doing in this PR is indeed the cleanest way of supporting autoguides, but I'll need to play around with the code to convince myself. Would it be possible to separate out the SVI related changes (your earlier PR) from the autoguide related changes and put up 2 PRs? We can get the former merged before discussing the latter set of changes. If its all tangled up, don't worry about it. I'll just need a bit longer to review this, and offer more concrete suggestions.

fehiepsi: Separating into 2 PRs is pretty straightforward so I'll do it to clarify things. The reason I made this PR is while refactoring init strategy, I realize that it is complicated to have dynamic support in autoguide. Change it leads to a change in SVI too. Maybe the test for SVI passed while test for autoguide failed made you think that I made a PR for SVI first, then added some commits to support autoguide later. It is indeed the other way around. :D

neerajprad: Haha..thanks for clarifying!

test/contrib/test_autoguide.py
```diff
@@ -83,3 +84,71 @@ def body_fn(i, val):
...
+
+    opt_state, _ = fori_loop(0, 1000, body_fn, (opt_state, rng_train))
+    median = guide.median(opt_state)
+    assert_allclose(median['loc'], true_coef, rtol=0.05)
+    # test .quantile method
+    median = guide.quantiles(opt_state, [0.2, 0.5])
+    assert_allclose(median['loc'][1], true_coef, rtol=0.1)
+
+
+def test_dynamic_supports():
```
neerajprad: This test seems to pass without the `if is_autoguide` block in `svi.py`. Could you modify this so that it fails? It will be nice to be able to refactor that part, and have a unit test tell us that the refactoring handles the case of dynamic support for autoguides.

neerajprad:
@fehiepsi - Thanks for splitting the PR, this makes it much easier for me to understand the changes involved. I am still of the opinion that it will be nice to maintain an interface where `guide(*args, **kwargs)` generates samples from the model's latent support, e.g. for predictive without having to pass in a boolean flag whether the guide is an autoguide. The change might be a little involved though, and it is best to punt it to a follow-up PR (I can take a look at that). 

Except for having a test that actually catches this issue of incorrect sampling in autoguide dynamic support, the changes look good to me.

fehiepsi:
I guess we can make it failed with `alpha` in (2, 3) instead. Let me try it.

test/contrib/test_autoguide.py
fehiepsi: @neerajprad do you mean by replacing `if is_autoguide` by `if False`, the test will pass?

neerajprad: > @neerajprad do you mean by replacing if is_autoguide by if False, the test will pass? 

Yes, the test passes without this block.

> Do you modify skip_dist_transforms=is_autoguide too?

I didn't modify this. I was hoping that changing either would result in a failing test.

fehiepsi: In my system, setting `if False` will throw. Maybe it is precision issue, I'll check it again.
```
>       assert_allclose(actual_loss, expected_loss)
E       AssertionError: 
E       Not equal to tolerance rtol=1e-07, atol=0
E       
E       Mismatch: 100%
E       Max absolute difference: 277.75
E       Max relative difference: 0.00472562
E        x: array(58497.617, dtype=float32)
E        y: array(58775.367, dtype=float32)
```

neerajprad: Oh..sorry my bad! I have been running the test in `test_svi.py` instead of `test_autoguide.py`, which will obviously pass. Sorry for wasting your time.

neerajprad: I am able to replicate the failure now; thanks for checking.

fehiepsi: Hmm, I run `JAX_ENABLE_X64=1 XLA_FLAGS=--xla_cpu_enable_fast_math=false pytest -vs test/contrib/test_autoguide.py -k dynamic`, the test failed with a large margin
```
>       assert_allclose(actual_loss, expected_loss)
E       AssertionError: 
E       Not equal to tolerance rtol=1e-07, atol=0
E       
E       Mismatch: 100%
E       Max absolute difference: 37648.07914654
E       Max relative difference: 0.44394272
E        x: array(47155.831707)
E        y: array(84803.910853)
```
I think it is just lucky to pass in your system. :D I have revised the test to have a larger relative difference (~0.01 in my system), hope that it will fail in your case.

fehiepsi: Haha, np!

PR #275, opened 2019-08-05
dynamic support for svi

fehiepsi:
This PR is a part of #273, only including SVI related changes.

### DIFF
--- a/numpyro/hmc_util.py
+++ b/numpyro/hmc_util.py
@@ -803,7 +803,6 @@ def initialize_model(rng, model, *model_args, init_strategy='uniform', **model_k
             raise ValueError('initialize={} is not a valid initialization strategy.'.format(init_strategy))
 
         if has_transformed_dist:
-            # we might want to replay the trace here
             constrain_fun = jax.partial(constrain_fn, seeded_model, model_args, model_kwargs, inv_transforms)
         else:
             constrain_fun = jax.partial(transform_fn, inv_transforms)
--- a/numpyro/hmc_util.py
+++ b/numpyro/hmc_util.py
@@ -712,6 +712,19 @@ def euclidean_kinetic_energy(inverse_mass_matrix, r):
 
 
 def potential_energy(model, model_args, model_kwargs, inv_transforms):
+    """
+    Makes a callable which computes potential energy of a model given unconstrained params.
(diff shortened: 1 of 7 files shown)


numpyro/hmc_util.py
```diff
@@ -711,11 +711,11 @@ def euclidean_kinetic_energy(inverse_mass_matrix, r):
     return 0.5 * np.dot(v, r)
 
 
-def potential_energy(model, model_args, model_kwargs, inv_transforms, skip_dist_transforms=True):
+def potential_energy(model, model_args, model_kwargs, inv_transforms):
```
neerajprad: I think, at some point, we should document the requirements on params - namely they should be unconstrained, and be from the base distribution.

numpyro/hmc_util.py
```diff
@@ -801,16 +791,16 @@ def single_chain_init(key, only_params=False):
 
         if has_transformed_dist:
             # we might want to replay the trace here
-            constrain_fn = make_constrain_fn(seeded_model, model_args, model_kwargs, inv_transforms)
+            constrain_fun = jax.partial(constrain_fn, seeded_model, model_args, model_kwargs, inv_transforms)
```
neerajprad: nit: remove comment above.

numpyro/infer_util.py
```diff
@@ -56,3 +53,10 @@ def transform_fn(transforms, params, invert=False):
         transforms = {k: v.inv for k, v in transforms.items()}
     return {k: transforms[k](v) if k in transforms else v
             for k, v in params.items()}
+
+
+def constrain_fn(model, model_args, model_kwargs, transforms, params):
```
neerajprad: Same as above - we should document that this constrains to base values, and perhaps change the name or add an optional kwarg like `use_base_dist_params=True`. We can punt this to later PRs after the refactoring is done.

numpyro/svi.py
```diff
@@ -125,6 +126,27 @@ def evaluate(rng, opt_state, model_args=(), guide_args=()):
     return init_fn, update_fn, evaluate
 
 
+# TODO: move this function to svi, rename it to get_params
```
neerajprad: +1, that will be great.

Issue #276, opened 2019-08-05
cleanup code


fehiepsi:
Just make a list here for some changes which have been put upstream to JAX to simplify our code base.
- np.var has `ddof` support
- np.median is implemented and has batching rule
- np.cov is available
- _elemwise_no_params functionality is available in jax
- various nonlinearity at stax is available: stax.Sigmoid (so we can revise vae example with this), slax.Elu
- entr/multigammaln is available at jax.scipy.special
- np.tril_indices is available
- lax.map can be used in sequential mcmc

PR #280, opened 2019-08-09
Updates for jax 0.1.41

### LINKED ISSUE #279: HMM example failing on travis with jax=0.1.41
With the lates jax, the hmm example throws an error in travis with the following message:

```
test/test_examples.py::test_cpu[hmm.py --num-samples 100 --num-warmup 100 --num-chains 2] Running:
python examples/hmm.py --num-samples 100 --num-warmup 100 --num-chains 2
Simulating data...
Starting inference...
  File "/home/travis/build/pyro-ppl/numpyro/examples/hmm.py", line 180, in <module>
    main(args)
  File "/home/travis/build/pyro-ppl/numpyro/examples/hmm.py", line 159, in main
    supervised_words, unsupervised_words,
  File "/home/travis/build/pyro-ppl/numpyro/numpyro/hmc_util.py", line 775, in initialize_model
    raise RuntimeError("Cannot find valid initial parameters. Please check your model again.")
RuntimeError: Cannot find valid initial parameters. Please check your model again.
FAILED
```

I am unable to reproduce this locally on my mac.

fehiepsi:
This PR fixes some issues at the examples involving `logsumexp`. It seems that there is some issues of the upstream `logsumexp` under fastmath mode.

In addition, this incorporates various cleanup at #276.

Fix #279.

### DIFF
--- a/examples/neutra.py
+++ b/examples/neutra.py
@@ -1,11 +1,11 @@
 import argparse
-import warnings
+import os
 
 from matplotlib.gridspec import GridSpec
 import matplotlib.pyplot as plt
 import seaborn as sns
 
-from jax import jit, random, vmap
+from jax import lax, random, vmap
 from jax.config import config as jax_config
 from jax.experimental import optimizers
 import jax.numpy as np
@@ -19,7 +19,10 @@ from numpyro.handlers import sample
 from numpyro.hmc_util import initialize_model
 from numpyro.mcmc import mcmc
 from numpyro.svi import elbo, svi
-from numpyro.util import fori_collect
+
+# TODO: remove when the issue https://github.com/google/jax/issues/939 is fixed upstream
+# The behaviour when training guide under fast math mode is unstable.
+os.environ["XLA_FLAGS"] = "--xla_cpu_enable_fast_math=false"
 
 """
 This example illustrates how to use a trained AutoIAFNormal autoguide to transform a posterior to a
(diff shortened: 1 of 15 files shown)


numpyro/distributions/util.py
```diff
@@ -453,6 +441,12 @@ def signed_stick_breaking_tril(t):
     return y
 
 
+def logsumexp(x, axis=0, keepdims=False):
+    x_max = np.max(x, axis=axis, keepdims=True)
```
neerajprad: Nice! Just curious - is the upstream `logsumexp` implementation doing something else that is causing issues with fastmath mode?

numpyro/util.py
```diff
@@ -69,7 +68,7 @@ def fori_loop(lower, upper, body_fun, init_val):
             val = body_fun(i, val)
         return val
     else:
-        return jit(lax.fori_loop, static_argnums=(2,))(lower, upper, body_fun, init_val)
+        return lax.fori_loop(lower, upper, body_fun, init_val)
```
neerajprad: Is the issue of slow compilation resolved with the latest release?

examples/neutra.py
```diff
@@ -64,26 +67,19 @@ def main(args):
     svi_init, svi_update, _ = svi(dual_moon_model, guide, elbo, opt_init, opt_update, get_params)
     opt_state, _ = svi_init(rng_init)
 
-    def body_fn(val):
-        i, loss, opt_state_, rng_ = val
+    def body_fn(val, i):
```
neerajprad: nit: can we keep the order (i, val) instead to match the ordering in `lax.fori_loop`'s `body_fn`?

numpyro/util.py
fehiepsi: Yes, it is fixed the last minute. :)

numpyro/distributions/util.py
fehiepsi: I think this is the only issue of NumPyro with JAX which have not been isolated yet. It is related to fastmath mode. I will ping you when I have more information. Here is the [upstream implementation](https://github.com/google/jax/blob/master/jax/scipy/special.py#L77) which does something else with -inf...

examples/neutra.py
fehiepsi: Here I want to use `lax.scan`, so i is just a dummy variable (though it is currently used in `svi_update`).

numpyro/distributions/util.py
neerajprad: Ahh..interesting. Feel free to create a NumPyro issue with your comment, for future reference.

fehiepsi: Actually, we have a TODO in neutra example which points to a similar issue.

PR #282, opened 2019-08-12
Add an example to illustrate automatic reparameterization in NumPyro

fehiepsi:
This PR adds a simple example to illustrate the recent feature added in NumPyro.

[funnel_plot.pdf](https://github.com/pyro-ppl/numpyro/files/3494572/funnel_plot.pdf)


neerajprad:
Nice! That's a beautiful illustration of HMC with non-centered reparameterization.

Issue #274, opened 2019-08-04
Refactor SVI api


fehiepsi:
I think we can simplify the common procedure of svi by
- [x]  group `(opt_init, opt_update, get_params)` into one input

```
opt_init, opt_update, get_params = optimizers.adam(0.05)
svi_init, svi_update, svi_eval = svi(model, guide, elbo, opt_init, opt_update,
                                           get_params)
```
by
```
optim = optimizers.adam(0.05)
svi_init, svi_update, svi_eval, get_params = svi(model, guide, elbo, optim)
```
The reason is we usually don't work with opt_init, opt_update.
- [ ]  Merge the input `i, rng, opt_state` into `svi_state`, so that
    - svi_init(...) returns init state
    - svi_update(state, model_data, guide_data) returns new state
This way, users usually don't need to write a custom `body_fn` to train SVI. In addition, using `fori_collect` will be easier (similar to `sample_kernel` in hmc).
- [x]  ~Allow `fori_collect` to return `collection` and last `val`. The reason is we don't want to collect of svi_states during the way (for memory reason). Only last state and losses are enough.~ **Solution:** use `lax.scan` instead.
- [x]  Use `get_params(svi_state)` in place of `constrain_fn(get_param(opt_state))`. We might have some args such as `get_params(svi_state, model_data, guide_data, )`. So users don't need to know what constrain_fn is when using `svi`.
- [x] We don't need to have `get_params` arg in `autoguide.__init__(...)`. We only need to feed in the params got from `get_params(svi_state)`.
- [x] Maybe we can move `rng` of autoguide to `setup`
- [x] It would be nice to have something similar to `pyro.module(...)`. Currently, the logic to handle nn parameters is a bit complicated - (@neerajprad)
```
def guide(batch, **kwargs):
    encode = kwargs['encode']
    encoder_params = param('encoder', None)
    ...
    z_loc, z_std = encode(encoder_params, batch)

encoder_init, encode = encoder(args.hidden_dim, args.z_dim)
decoder_init, decode = decoder(args.hidden_dim, 28 * 28)
_, encoder_params = encoder_init(rng_enc, (args.batch_size, 28 * 28))
_, decoder_params = decoder_init(rng_dec, (args.batch_size, args.z_dim))
params = {'encoder': encoder_params, 'decoder': decoder_params}

opt_state, constrain_fn = svi_init(rng_init, (sample_batch,), (sample_batch,), params)
```
I think we can replace it by
```
[...]

PR #283, opened 2019-08-14
Merge optim tuples

fehiepsi:
Address part of #274

### DIFF
--- a/docs/source/mcmc.rst
+++ b/docs/source/mcmc.rst
@@ -22,8 +22,6 @@ MCMC Utilities
 
 .. autofunction:: numpyro.util.fori_collect
 
-.. autofunction:: numpyro.diagnostics.summary
-
 .. autofunction:: numpyro.hmc_util.consensus
 
 .. autofunction:: numpyro.hmc_util.parametric
--- a/test/test_svi.py
+++ b/test/test_svi.py
@@ -25,8 +25,8 @@ def test_beta_bernoulli():
                        constraint=constraints.positive)
         sample("beta", dist.Beta(alpha_q, beta_q))
 
-    opt_init, opt_update, get_params = optimizers.adam(0.05)
-    svi_init, svi_update, _ = svi(model, guide, elbo, opt_init, opt_update, get_params)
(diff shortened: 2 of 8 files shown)


docs/source/mcmc.rst
```diff
@@ -22,8 +22,6 @@ MCMC Utilities
 
 .. autofunction:: numpyro.util.fori_collect
 
-.. autofunction:: numpyro.diagnostics.summary
```
fehiepsi: this is available in diagnostic docs

test/test_svi.py
```diff
@@ -25,8 +25,8 @@ def guide():
                        constraint=constraints.positive)
         sample("beta", dist.Beta(alpha_q, beta_q))
 
-    opt_init, opt_update, get_params = optimizers.adam(0.05)
-    svi_init, svi_update, _ = svi(model, guide, elbo, opt_init, opt_update, get_params)
+    _, _, get_params = optim = optimizers.adam(0.05)
```
neerajprad: I have a suggestion. How about we just wrap the `optimizers` module in jax, and just make these available as class methods? Kind of what we do in Pyro. Then users can just call `opt.get_params()`, or `opt.update()`, etc. That will also help us pass `opt` directly to `svi` without having to do any packing, and pass the same to autoguides. I think the stax and optimizers modules are designed to be used as substrates for more complex functionality anyways.

neerajprad: I think this is a good intermediate solution anyways, and am happy to merge this and handle this in a separate PR.

fehiepsi: I vote for it. Though currently, it seems not necessary, I think it is potentially useful later on (e.g. if we need to create something similar to ClippedAdam, users won't get confused about where to import this optimizer: from jax.experimental.optimizers or numpyro.optim...).

fehiepsi: There are some merge conflicts so I'll try to make a class version for this while addressing them.

fehiepsi:
Let's close this because there are many merge conflicts. It is better to rewrite the change with the new optim class.

PR #284, opened 2019-08-22
Refactoring to replay from base values by default

neerajprad:
Changes:
 - [x] `replay` messenger replays using base values directly. This allows us to remove any autoguide related logic from svi.
 - [x] moved `get_params` to within `svi`. Unfortunately, we can't pass in the optimizer directly to svi, since we need access to the `optim.get_params` function in autoguides, and its not easy (or even desirable) to replace that with our `svi.get_params` function.


examples/vae.py
```diff
@@ -63,8 +63,8 @@ def binarize(rng, batch):
 def main(args):
     encoder_init, encode = encoder(args.hidden_dim, args.z_dim)
     decoder_init, decode = decoder(args.hidden_dim, 28 * 28)
-    opt_init, opt_update, get_params = optimizers.adam(args.learning_rate)
-    svi_init, svi_update, svi_eval = svi(model, guide, elbo, opt_init, opt_update, get_params,
+    opt_init, opt_update, get_opt_params = optimizers.adam(args.learning_rate)
```
neerajprad: We need to initialize the optimizer outside the svi function so that we can pass `get_params` to autoguides.

fehiepsi:
> Unfortunately, we can't pass in the optimizer directly to svi, since we need access to the optim.get_params function in autoguides

You are right. I believe that there are also some bugs which are not detected in our test cases. I have a branch with some fixes but it became a bit complicated that hold me off from further refactoring. I'll push it to a PR to catch them.

examples/vae.py
fehiepsi: How about creating an arg to get_params method to achieve that purpose?

numpyro/handlers.py
```diff
@@ -195,7 +195,13 @@ def __init__(self, fn, guide_trace):
...
     def process_message(self, msg):
         if msg['name'] in self.guide_trace:
-            msg['value'] = self.guide_trace[msg['name']]['value']
+            guide_site = self.guide_trace[msg['name']]
+            if msg['type'] == 'sample':
+                    value, intermediates = guide_site['value'], guide_site['intermediates']
+                    base_value = intermediates[0][0] if intermediates else value
+                    msg['value'], msg['intermediates'] = msg['fn'].transform_with_intermediates(base_value)
+            elif msg['type'] == 'param':
+                msg['value'] = guide_site['value']
```
fehiepsi: This is a great solution!

examples/vae.py
neerajprad: Hmm..I'm not sure if that would help, specially if `get_params` is an internal function of `svi`, since we have to declare the autoguide before `svi`. Or maybe I'm misunderstanding your suggestion?

fehiepsi: My idea is to make AutoGuide separated from `opt_state`. We just need to provide `opt_get_params(opt_state)` to AutoGuide methods to obtain the results. So I think we can achieve it by `params = svi_get_params(opt_state, unconstrained=True)` and use this `params` in autoguide methods.

neerajprad: Hopefully, this shouldn't matter if we convert it into a class. Its fine to only send constrained parameters to autoguide methods directly, but then the onus of getting the constrained parameters falls on the user.

numpyro/svi.py
```diff
@@ -177,17 +171,9 @@ def elbo(param_map, model, guide, model_args, guide_args, kwargs, constrain_fn,
...
-                param_map[name] = site['value']
-    else:
-        # NB: we only want to substitute params not available in guide_trace
-        param_map = {k: v for k, v in param_map.items() if k not in guide_trace}
-        model = replay(model, guide_trace)
-    model_log_density, _ = log_density(model, model_args, kwargs, param_map,
-                                       skip_dist_transforms=is_autoguide)
+    param_map = {k: v for k, v in param_map.items() if k not in guide_trace}
+    model = replay(model, guide_trace)
+    model_log_density, _ = log_density(model, model_args, kwargs, param_map)
```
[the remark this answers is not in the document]
neerajprad: My understanding was that if the same sample site is present in both, they both will have a constant log det term for scaling which should cancel out when we do `model_log_density - guide_log_density`. We can probably also include it for both. Let me know if I got that wrong.

examples/vae.py
fehiepsi: >  the onus of getting the constrained parameters falls on the user

I want the simplicity too. Assume that we use classes and have `optimizer = optim.Adam(...)`, then users need to either
```
guide = AutoDiagonalNormal(optimizer)
guide.quantile(svi_state)
```
or
```
guide = AutoDiagonalNormal()
guide.quantile(optimizer.get_params(svi_state.opt_state))
```
It seems that the first way (which is more aligned with current interface) is simpler.

Another way is to have an additional field in `svi_state`, says `svi_state.opt_params`, so users only need to provide
```
guide.quantile(svi_state.opt_params)
```
WDYT?

numpyro/svi.py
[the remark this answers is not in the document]
fehiepsi: I think that we do inference for SVI in constrained space, so we have to compute `log_prob` of constrained values. The issue will happen when prior is `dist.Normal` but guide is `dist.Uniform`, or when prior and guide are transformed distributions with different transforms. In SVI, we compute KL of 2 constrained variables, while in autoguide, we compute KL of 2 unconstrained variables.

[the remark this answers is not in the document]
neerajprad: > The issue will happen when prior is dist.Normal but guide is dist.Uniform

This will not work if the guide has a different support for a sample site than the model. I have been assuming that any well designed guide should retain the sample site's support (like we do in autoguides). So even if we use a uniform distribution in the guide, we would probably want to transform it back to the Real domain so as to map to the support of the site in the model (a Normal distribution). We can add a check for this, if needed.

> or when prior and guide are transformed distributions with different transforms. 

Could you give an example? Suppose we have a uniform distribution from [0, 5] in the model and we have a transformed distribution in the guide that transforms from a Normal distribution to [0, 5]. Then I think that in both the cases we can just compute the log density for the base values (ignoring the scaling term from [0, 1] -> [0, 5]).

Maybe we can add some tests to compute the elbo for all of these cases, but just want to make sure I really understand the issues involved, and am not oversimplifying.

examples/vae.py
neerajprad: Either (1) or (2) looks fine to me right now, though (2) looks better since that frees us to declare the optimizer wherever we can.

numpyro/svi.py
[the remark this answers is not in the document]
fehiepsi: My assumption is for the case `model` and `guide` have different transforms, such as
```
def model():
    sample('x', dist.TransformedDistribution(dist.Normal(), AffineTransform(0, 1))

def guide():
    sample('x', dist.TransformedDistribution(dist.Normal(), AffineTransform(0, 2))
```

Because `AffineTransform(0, 1)` logdet is different from `AffineTransform(0, 2)` logdet, `log_prob_guide(x) - log_prob_model(x)` will be different from `log_prob_guide(base_x) - log_prob_model(base_x)`.

Similarly, for bounded support, logdet of `Sigmoid` is different from logdet of `Compose([Tanh, Affine(0.5, 0.5))` though both of them have the same codomain `[0, 1]`.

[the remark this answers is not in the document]
neerajprad: Thanks for explaining, that makes sense. I am still unclear about why we need to do anything differently for autoguides though. What goes wrong if we instead have `skip_dist_transforms=False` for both guide and model under all cases?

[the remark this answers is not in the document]
neerajprad: That passes all our tests btw - the computation of the elbo loss term is different in `test_autoguides.py::test_dynamic_supports` but that's expected since the log density term will be different for `Uniform(0, alpha)` and `Uniform(0, 1)` but it shouldn't affect the optimal parameter values.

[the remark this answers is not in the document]
neerajprad: I have reverted the autoguide check. While it would be nice to not have to make a distinction in svi, I don't see any easy way to accomplish this currently.

neerajprad:
For the time being, I have pinned jax to version 0.1.41, so that we can handle all update related changes in follow-up PRs.

numpyro/svi.py
```diff
@@ -177,17 +171,9 @@ def elbo(param_map, model, guide, model_args, guide_args, kwargs, constrain_fn,
     """
     param_map = constrain_fn(param_map)
     guide_log_density, guide_trace = log_density(guide, guide_args, kwargs, param_map)
```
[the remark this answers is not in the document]
neerajprad: Done.

numpyro/svi.py
```diff
@@ -177,15 +173,8 @@ def elbo(param_map, model, guide, model_args, guide_args, kwargs, constrain_fn,
...
-        # in autoguide, a site's value holds intermediate value
-        for name, site in guide_trace.items():
-            if site['type'] == 'sample':
-                param_map[name] = site['value']
-    else:
-        # NB: we only want to substitute params not available in guide_trace
-        param_map = {k: v for k, v in param_map.items() if k not in guide_trace}
-        model = replay(model, guide_trace)
+    param_map = {k: v for k, v in param_map.items() if k not in guide_trace}
+    model = replay(model, guide_trace)
```
fehiepsi: @neerajprad I missed this issue: in SVI (not autoguide), we don't want to replay intermediate values. We only want to replay intermediate values in autoguide. So I guess we need to revert to the old logic too. Sorry for missing this!

neerajprad: Sorry, that was my bad. This feature of supporting dynamic constraints has added more complexity to the code base than I had not anticipated earlier. I'll revert this in another PR.

Issue #290, opened 2019-08-23
Add cheap tests to assert correct computation of ELBO


neerajprad:
With dynamic constraints and changes in the autoguide class, it is easy to introduce regression for a class of models while having our current test suite pass. We should have really cheap unit tests (not doing full inference) that validate ELBO computation for different classes of models (with autoguides and custom guides) so as to avoid such regressions.

fehiepsi:
Agree, this is important to have.

Issue #287, opened 2019-08-22
substitute logic in autoguide is inconsistence


fehiepsi:
Currently, e.g., in AutoDiagonalNormal class, we substitute unconstrained scale value in `median`, `quantiles`, `sample_posterior`. I think that we should use `substitute(self._loc_scale, base_param_map=params)()` there.

neerajprad:
That's definitely wrong. Also, I think substitute should substitute the constrained value for `param` sites by default (exactly as in `base_param_map` but in this case the domain of the values is Real and we apply the full transform to constrained).

PR #291, opened 2019-08-23
add svi dynamic support test

fehiepsi:
Addresses #290.

@neerajprad Is this test enough to guarantee regressions?

### DIFF
--- a/numpyro/distributions/constraints.py
+++ b/numpyro/distributions/constraints.py
@@ -239,7 +239,7 @@ class AffineTransform(Transform):
         return (y - self.loc) / self.scale
 
     def log_abs_det_jacobian(self, x, y, intermediates=None):
-        return sum_rightmost(np.broadcast_to(np.log(np.abs(self.scale)), x.shape), self.event_dim)
+        return sum_rightmost(np.broadcast_to(np.log(np.abs(self.scale)), np.shape(x)), self.event_dim)
 
 
 class ComposeTransform(Transform):
--- a/test/test_svi.py
+++ b/test/test_svi.py
@@ -6,7 +6,8 @@ import jax.numpy as np
 
 import numpyro.distributions as dist
 from numpyro.distributions import constraints
-from numpyro.handlers import param, sample
+from numpyro.distributions.constraints import SigmoidTransform, AffineTransform
+from numpyro.handlers import param, sample, substitute
 from numpyro.svi import elbo, svi
 from numpyro.util import fori_loop
 
@@ -67,3 +68,26 @@ def test_dynamic_constraints():
     opt_state, rng = fori_loop(0, 300, body_fn, (opt_state, rng_train))
     params = get_params(opt_state, rng, guide_args=())
     assert_allclose(params['loc'], true_coef, atol=0.05)
+
+
+def test_dynamic_supports():
+    x_prior = dist.TransformedDistribution(
+        dist.Normal(), [AffineTransform(0, 2), SigmoidTransform(), AffineTransform(0, 3)])
+    x_guide = dist.Uniform(0, 3)
+
+    def model():
+        sample('x', x_prior)
+
+    def guide():
+        sample('x', x_guide)
+
+    opt_init, opt_update, get_opt_params = optimizers.adam(0.01)
+    # set base value of x_guide is 0.9
+    x_base = 10
+    guide = substitute(guide, base_param_map={'x': x_base})
+    svi_init, _, svi_eval = svi(model, guide, elbo, opt_init, opt_update, get_opt_params)
+    opt_state, get_params = svi_init(random.PRNGKey(0), (), ())
+    actual_loss = svi_eval(random.PRNGKey(1), opt_state)
+    x, _ = x_guide.transform_with_intermediates(x_base)
+    expected_loss = x_guide.log_prob(x) - x_prior.log_prob(x)
+    assert_allclose(actual_loss, expected_loss)


fehiepsi:
hmm, I guess I need to add a similar test for `autoguide` too. But it seems not trivial to access latent values of autoguide's guide.

fehiepsi:
@neerajprad While thinking about this, I think we can add some signals to `autoguide`'s latent `sample` primitives to handle the differences in elbo computations of svi and autoguide. For example,
we can have
```
sample('x', dist.Delta(value, log_density), skip_transforms=True)
```
Then we make `replay(model, guide_trace)` only replay `base_value` if `skip_transforms==True`. In addition, in log_density computation, if we see this flag from a latent site, then we just need to compute log_prob of base_value. This way resolves the gross caused by autoguide in elbo implementation.

test/test_svi.py
```diff
@@ -67,3 +68,26 @@ def body_fn(i, val):
...
+    def guide():
+        sample('x', x_guide)
+
+    opt_init, opt_update, get_opt_params = optimizers.adam(0.01)
+    # set base value of x_guide is 10
+    x_base = 10
+    guide = substitute(guide, base_param_map={'x': x_base})
+    svi_init, _, svi_eval = svi(model, guide, elbo, opt_init, opt_update, get_opt_params)
+    opt_state, get_params = svi_init(random.PRNGKey(0), (), ())
+    actual_loss = svi_eval(random.PRNGKey(1), opt_state)
```
neerajprad: Hmm..this seems to be giving nan values.

fehiepsi: ...

fehiepsi: My bad, 10 is out of support of Uniform(0, 1).

numpyro/distributions/constraints.py
```diff
@@ -172,7 +173,8 @@ def __call__(self, x):
 ##########################################################
 
 def _clipped_expit(x):
-    return np.clip(expit(x), a_min=np.finfo(x.dtype).tiny, a_max=1.-np.finfo(x.dtype).eps)
+    dtype = get_dtype(x)
```
fehiepsi: this deals with the case `x` is python scalar.

numpyro/distributions/util.py
```diff
@@ -389,8 +389,8 @@ def promote_shapes(*args, shape=()):
                 if len(s) < num_dims else arg for arg, s in zip(args, shapes)]
 
 
-def get_dtypes(*args):
-    return [canonicalize_dtype(lax.dtype(arg)) for arg in args]
+def get_dtype(x):
+    return canonicalize_dtype(lax.dtype(x))
```
fehiepsi: I find no place we want to get multiple dtypes so I make this simplification.

numpyro/svi.py
```diff
@@ -70,12 +70,10 @@ def init_fn(rng, model_args=(), guide_args=(), params=None):
...
                 transform = biject_to(constraint)
                 if isinstance(transform, ComposeTransform):
-                    base_transform = transform.parts[0]
-                    inv_transforms[site['name']] = base_transform
-                    params[site['name']] = base_transform(transform.inv(site['value']))
+                    inv_transforms[site['name']] = transform.parts[0]
                 else:
                     inv_transforms[site['name']] = transform
-                    params[site['name']] = site['value']
+                params[site['name']] = transform.inv(site['value'])
```
fehiepsi: this is a bug, which incorrectly defines init params. I added an assertation at `test_beta_bernoulli` in `test_svi` to prevent this regression.

test/test_svi.py
fehiepsi: I changed `10` to `0.9` and added an assert statement that `actual_loss` is finite.

neerajprad:
> Then we make replay(model, guide_trace) only replay base_value if skip_transforms==True. In addition, in log_density computation, if we see this flag from a latent site, then we just need to compute log_prob of base_value. This way resolves the gross caused by autoguide in elbo implementation. WDYT?

That sounds reasonable. My thought after making an attempt at simplifying this is that what we have right now is adding to the internal complexity of these classes which makes refactoring harder but doesn't affect the public interface. This proposal will shift the complexity to the handler class which is more public. e.g. its not immediately obvious why `sample` should keep track of `skip_dist_transforms` since it doesn't affect the sampling but only other handlers and utility functions. Maybe we can have a dictionary called `site_config` that can be used to pass such settings. In any case, I would suggest creating an issue and handling this in a separate PR where we can discuss the merits of doing it this way. For the next release, this isn't too important since this is internal and can be refactored any time. I do think though that it is a promising solution to clean up the code!

test/test_svi.py
```diff
@@ -67,3 +69,27 @@ def body_fn(i, val):
...
+
+    def model():
+        sample('x', x_prior)
+
+    def guide():
+        sample('x', x_guide)
+
+    opt_init, opt_update, get_opt_params = optimizers.adam(0.01)
+    # set base value of x_guide is 0.9
+    x_base = 0.9
```
neerajprad: Rather than setting this to 0.9, what do you think about validating the elbo computation vmapped over rng and comparing this estimate to the Monte Carlo estimate by drawing n samples from x_guide and computing `np.mean(x_prior.log_prob(samples) - x_guide.log_prob(samples))`? Or maybe we can have these tests separately for different kinds of model/guide pairs.

fehiepsi: I think we can add another separated test, like `test_kl_dynamic_support()`. Let me try it.

fehiepsi:
> its not immediately obvious why sample should keep track of skip_dist_transforms

That makes sense, I haven't thought about it. How about having a handler to inject this kwarg in autoguide?

fehiepsi:
@neerajprad I am seeing some old tests are failing with the fix at `svi_init`. So I'll address them in this PR too (possibly fixing the issue at autoguide will solve #287 too).

test/test_svi.py
neerajprad: I'm happy to merge this and have kl tests in a separate PR as long as existing tests pass.

fehiepsi:
@neerajprad It is a bit complicated to fix issues in autoguide, so I marked the places which might causes issue and will address them in a later PR.

neerajprad:
> @neerajprad It is a bit complicated to fix issues in autoguide, so I marked the places which might causes issue and will address them in a later PR.

Sounds good! Btw..if needed, we can also first try to simplify these classes as much as possible. If we need to remove certain features to facilitate that, we should do that. e.g. removing dynamic support for autoguides, or even just ignoring the autoguide module for now.

neerajprad:
> That makes sense, I haven't thought about it. How about having a handler to inject this kwarg in autoguide?

I think we will still need to make changes to `replay` and `log_density` etc. If it can all be subsumed within the contrib module, then that's good. But if it requires changes to the core code, we should probably debate whether its worth adding additional complexity.

numpyro/distributions/constraints.py
```diff
@@ -550,13 +550,17 @@ def _transform_to_corr_cholesky(constraint):
 
 @biject_to.register(greater_than)
 def _transform_to_greater_than(constraint):
+    if constraint is positive:
```
neerajprad: Is this just for efficiency, or does it lead to some corner cases in dynamic support?

numpyro/svi.py
```diff
@@ -35,6 +35,10 @@ def svi(model, guide, loss, optim_init, optim_update, get_optim_params, **kwargs
     :return: tuple of `(init_fn, update_fn, evaluate)`.
     """
     constrain_fn = None
+    # NB: these will become properties of SVI classes
+    has_dynamic_constraints = False
+    _model_args = None
```
neerajprad: `guide_args` and `model_args` might change per mini-batch. Do we need to store this? I think that supporting sample statements with dynamic support is probably more important than supporting dynamic `param` and users will typically have static constraints for these. If so, what do you think about deferring adding this for now? We can mark any failing tests as so and revisit this after the refactor.

fehiepsi: I think it is particularly unnecessary to support param statements with dynamic constraints. So it is better to remove these stuffs in SVI.

numpyro/distributions/constraints.py
fehiepsi: Yes, it is just for efficiency.

numpyro/svi.py
```diff
@@ -147,7 +132,7 @@ def evaluate(rng, opt_state, model_args=(), guide_args=()):
     return init_fn, update_fn, evaluate
 
 
-def elbo(param_map, model, guide, model_args, guide_args, kwargs, constrain_fn, is_autoguide=False):
+def elbo(param_map, model, guide, model_args, guide_args, kwargs):
```
fehiepsi: @neerajprad I removed `constrain_fn` for clarity. We no longer expose `constrain_fn` for users so we'd better remove it. In addition, I removed `is_autoguide` arg and move the check `isinstance(guide, AutoContinuous)` from svi to inside of elbo to clean up the interface. We'll address issues of autoguide in another PR. I think we should set default values to args, kwargs here but let's defer it to another PR.

test/test_svi.py
fehiepsi: @neerajprad There is nothing change in SVI implementation to handle dynamic support, so having KL test seems not necessary to me. The current test is just to make sure that future changes won't break the logic of elbo.

Issue #298, opened 2019-08-26
Remove params arg from SVI


neerajprad:
With the `module` primitive we can remove the params arg to SVI and also simplify the AutIAFNormal class.

Issue #148, opened 2019-05-12
Make inference utilities available for prediction


neerajprad:
Related issue - #41. 

It will be nice to provide an easy interface for doing predictions (ideally a vectorized version of `TracePredictive`), so that users do not need to write code to do this by themselves. e.g. [baseball](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/examples/baseball.py#L150). More detailed discussion on the design is needed, and hopefully, we'll have some more insights on this after attempting this in Pyro https://github.com/pyro-ppl/pyro/issues/1725.

fehiepsi:
I think that for many models, we can also make a vectorized version in numpyro by using `vmap` (not sure how to do it in Pyro though).

fehiepsi:
More tasks to complete this issue:

- [ ] Implement predictive for log likelihood. (I wondered if we can store `args`, `kwargs` in mcmc object so that these model-related information can be accessed when we pass `mcmc` object to other library (e.g. arviz))
- [ ] Implement predictive for prior. For this purpose, we can have `num_samples` and allow empty `posterior_samples` as in NumPyro.

neerajprad:
@fehiepsi - I think we can close this issue now?

fehiepsi:
Sure! I'll go for a pass tomorrow to see if we miss any feature for arviz, then will update #349.

PR #305, opened 2019-08-28
Remove param arg from SVI

fehiepsi:
Addresses #298

~Blocked by #302~


fehiepsi:
@neerajprad The change with `PRNGIdentity` (pretty cool!) is small and works beautifully, so I merge that change in this PR too. You can see the diff at latest commit.

neerajprad:
> maybe the prediction utility #148 or add docs to our current transforms, which is currently blank). 

I think the predictive utility can wait for `0.2.1`, but feel free to put out PRs adding documentation for `0.2`. This isn't the most urgent thing, and I see that you already have something in the other PR. :)

> Edit: or maybe I can start taking a look at vmap of build_tree. :)

I think that will be very useful to investigate, and benchmark w.r.t. pmap. Even if it doesn't work in its current form, it will be good to know what the blockers are.

Issue #285, opened 2019-08-22
Use classes as interface for SVI and MCMC


neerajprad:
**Tasks** (to be expanded): 
 - [x] Create `SVI` class; mark existing `svi` interface as deprecated.
 - [x] Create `MCMC` class; mark existing `mcmc` interface as deprecated.
 - [x] Create `Optim` class to wrap over jax optimizers.

**Motivation:**

Related: #274 

As I am looking at the codebase with fresh eyes, it seems to me that we have to write a lot of boiler-plate code in NumPyro due to the functional style of the API. 

e.g. for SVI:

```python
    opt_init, opt_update, opt_params = optimizers.adam(0.01)
    rng_guide, rng_init, rng_train = random.split(random.PRNGKey(1), 3)
    svi_init, svi_update, _ = svi(model, guide, elbo, opt_init, opt_update, opt_params)
    opt_state, get_params = svi_init(rng_init, model_args=(data,), guide_args=(data,))

    def body_fn(i, val):
        opt_state_, rng_ = val
        loss, opt_state_, rng_ = svi_update(i, rng_, opt_state_, model_args=(data,), guide_args=(data,))
        return opt_state_, rng_

    opt_state, _ = fori_loop(0, 1000, body_fn, (opt_state, rng_train))
    print(get_params(opt_state))
```

While this can be simplified somewhat with #274, in an ideal world, this should be as simple and easy to remember as:

```python
svi = svi(model, guide, optim, loss)
svi.run(1000, rng)  # also provide svi.update for users to loop
print(svi.get_params())
```

**Alternative:** Another option would be to provide another functional API like `svi_run(..)` in the same spirit as `mcmc`. The disadvantage of that is that it caters to a very simple use case with a single return type. If the users want to get the optimizer `get_state` and the `constrain_fn` and also the final `parameters`, it becomes cumbersome. We'll have to offer two options (like with `mcmc`) - a simple interface which returns the learned parameters, or else write all the boiler plate code and get access to all the intermediate variables.

I am trying to think through the benefits / drawbacks of using classes vs. our current functional API:

**Using classes:**
Benefits:
 - We can have a minimal API that offers the same functionality with much less boilerplate code, since we don't have to cycle arguments back and forth, and can abstract whatever functional logic we need to deal with the RNG or optimizer inside the class.
[...]

PR #310, opened 2019-08-28
Add high level class based interface for MCMC

neerajprad:
Addresses #285.

 - [x] Update documentation.
 - [x] Verify exact equality w.r.t. the `mcmc` interface.
 - [x] Figure out why the current interface is 1.5 times slower than the `mcmc` interface, and if the two can be brought to parity. Update: The class interface is only slow with `progbar=True` but I don't see anything fundamentally different between this and what we are doing now except that we are running the warmup within `init_kernel`.
 - [x] Add unit tests.

Future PRs:
 - Add deprecation warning for `mcmc`.
 - Update tutorials/examples.

### DIFF
--- a/numpyro/mcmc.py
+++ b/numpyro/mcmc.py
@@ -382,7 +382,7 @@ def mcmc(num_warmup, num_samples, init_params, num_chains=1, sampler='hmc',
            intercept      -0.03       0.02      -0.06       0.00     402.53       1.00
     """
     warnings.warn("This interface to MCMC is deprecated and will be removed in the "
-                  "next version. Please use `numpyrio.mcmc.MCMC` instead.",
+                  "next version. Please use `numpyro.mcmc.MCMC` instead.",
                   DeprecationWarning)
     sequential_chain = False
     if xla_bridge.device_count() < num_chains:
@@ -479,8 +479,8 @@ class MCMCKernel(ABC):
 
 class HMC(MCMCKernel):
     """
-    Hamiltonian Monte Carlo inference, using fixed number of steps with
-    step size and mass matrix adaptation.
+    Hamiltonian Monte Carlo inference, using fixed trajectory length, with
+    provision for step size and mass matrix adaptation.
 
     **References:**
 
@@ -540,7 +540,8 @@ class HMC(MCMCKernel):
     def init(self, rng, num_warmup, init_params=None, model_args=(), model_kwargs={}):
         constrain_fn = None
         if self.model is not None:
-            init_params_, self.potential_fn, constrain_fn = initialize_model(rng, self.model,
+            rng, rng_init_model = random.split(rng)
+            init_params_, self.potential_fn, constrain_fn = initialize_model(rng_init_model, self.model,
                                                                              *model_args, **model_kwargs)
             if init_params is None:
                 init_params = init_params_
@@ -559,7 +560,8 @@ class HMC(MCMCKernel):
                                  target_accept_prob=self.target_accept_prob,
                                  trajectory_length=self.trajectory_length,
                                  max_tree_depth=self.max_tree_depth,
-                                 run_warmup=False)
+                                 run_warmup=False,
+                                 rng=rng)
         return init_state, constrain_fn
 
     def sample(self, state):
@@ -679,11 +681,7 @@ class MCMC(object):
         self._samples_flat = None
 
     def _single_chain_mcmc(self, init, args=(), kwargs={}):
-        init_params = None
-        if len(init) == 1:
-            rng, = init
-        else:
-            rng, init_params = init
+        rng, init_params = init
         hmc_state, constrain_fn = self.sampler.init(rng, self.num_warmup, init_params,
                                                     model_args=args, model_kwargs=kwargs)
         if self.constrain_fn is None:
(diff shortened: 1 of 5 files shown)


numpyro/mcmc.py
```diff
@@ -77,6 +79,12 @@ def get_diagnostics_str(hmc_state):
                                                               hmc_state.mean_accept_prob)
 
 
+def get_progbar_desc_str(num_warmup, hmc_state):
+    if hmc_state.i < num_warmup:
```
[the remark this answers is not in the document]
neerajprad: I verified that the performance is unimpacted, there was a problem with the logistic_regression test where we weren't passing in the `algo` argument so it was actually running NUTS instead of HMC and appeared to be faster.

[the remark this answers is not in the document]
fehiepsi: haha, this is truly tricky to detect.

numpyro/mcmc.py
```diff
@@ -428,3 +437,159 @@ def single_chain_mcmc(rng, init_params):
...
+
+    def _single_chain_mcmc(self, rng, init_params=None, args=(), kwargs={}):
+        hmc_state, constrain_fn = self.sampler.init(rng, self.num_warmup, init_params,
+                                                    model_args=args, model_kwargs=kwargs)
+        if self.constrain_fn is None:
+            self.constrain_fn = identity if constrain_fn is None else constrain_fn
+        samples = fori_collect(self.num_warmup, self.num_warmup + self.num_samples,
+                               self.sampler.sample,
+                               hmc_state,
+                               transform=lambda x: self.constrain_fn(x.z),
```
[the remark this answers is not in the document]
neerajprad: Collecting warmup samples should be a pretty straightforward addition. Reg. 

>Maybe we can have diagnostic_fn to kwargs of run method, that works like lambda x: (x.num_steps, x.step_size) and set transform=lambda x: (self.constrain_fn(x.z), self.diagnostic_fn())

Since we can collect all the HMC states (not just `z`), would it work if we specify the list of fields to collect through an arg like `collect_fields=('z', 'num_steps')`. We are storing `num_steps` in HMCState so that should be possible, but let me know if I am overlooking something.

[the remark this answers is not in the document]
neerajprad: I would like to address both of these in a follow-up PR.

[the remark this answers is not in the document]
fehiepsi: > address both of these in a follow-up PR.
Definitely!

> collect through an arg like collect_fields=('z', 'num_steps')

This is much better than the thing I wanted, and easier to remember too. (using `transform=lambda x: x.adapt_state.step_size` is a bit unconventional, especially for users).

numpyro/mcmc.py
```diff
@@ -428,3 +441,303 @@ def single_chain_mcmc(rng, init_params):
...
+        :param state: Arbitrary data structure representing the state for the
+            kernel. For HMC, this is given by :data:`~numpyro.mcmc.HMCState`.
+        :return: Next `state`.
+        """
+        raise NotImplementedError
+
+
+class HMC(MCMCKernel):
+    """
+    Hamiltonian Monte Carlo inference, using fixed number of steps with
```
fehiepsi: nit: maybe `number of steps` -> `trajectory length` (which is more correct, especially with adaptation scheme)?

numpyro/mcmc.py
```diff
@@ -428,3 +441,303 @@ def single_chain_mcmc(rng, init_params):
...
+    def __init__(self,
+                 model=None,
+                 potential_fn=None,
+                 kinetic_fn=None,
+                 step_size=1.0,
+                 adapt_step_size=True,
+                 adapt_mass_matrix=True,
+                 dense_mass=False,
+                 target_accept_prob=0.8,
+                 trajectory_length=2*math.pi):
```
fehiepsi: could you also expose `max_tree_depth`?

numpyro/mcmc.py
```diff
@@ -428,3 +441,303 @@ def single_chain_mcmc(rng, init_params):
...
+
+        self._samples = None
+        self._samples_flat = None
+
+    def _single_chain_mcmc(self, init, args=(), kwargs={}):
+        init_params = None
+        if len(init) == 1:
+            rng, = init
+        else:
+            rng, init_params = init
```
fehiepsi: Hmm, `pmap`, `lax.map` supports batching over `None` and it is clear that if `init_params` is not specified, then until here it is still None. I think we can simplify this with
```
rng, init_params = init

# and at `run` method
samples = map_fn((rngs, init_params))
```

numpyro/mcmc.py
```diff
@@ -428,3 +441,303 @@ def single_chain_mcmc(rng, init_params):
...
+        self.trajectory_length = trajectory_length
+        self.sample_fn = None
+        self.algo = 'HMC'
+        self.max_tree_depth = 10
+
+    @copy_docs_from(MCMCKernel.init)
+    def init(self, rng, num_warmup, init_params=None, model_args=(), model_kwargs={}):
+        constrain_fn = None
+        if self.model is not None:
+            init_params_, self.potential_fn, constrain_fn = initialize_model(rng, self.model,
```
fehiepsi: A `rng` is missing at `hmc_init_fn`. I think we can split rng here.
```
rng, rng_init_model = random.split(rng)
init_params_, self.potential_fn, constrain_fn = initialize_model(rng_init_model, ...
```

numpyro/mcmc.py
neerajprad: Oh..I didn't realize batching over None was allowed. Let me simplify that.

numpyro/mcmc.py
neerajprad: That's exposed in `NUTS`, which is basically subclassing off of this. I wanted to separate out the kernels so that we don't end up with bugs like `sampler='hmc', algo='nuts'`.

numpyro/mcmc.py
neerajprad: Good catch! Since none of our tests in `test_mcmc` pass `rng` directly to the `mcmc` function and instead only use it in `intialize_model`, I was getting full agreement. Made the change now.

PR #314, opened 2019-08-29
Merge opt_state and rng to svi_state

### LINKED ISSUE #274: Refactor SVI api

fehiepsi:
Fixes #274.

### DIFF
--- a/examples/vae.py
+++ b/examples/vae.py
@@ -72,31 +72,31 @@ def main(args):
     num_train, train_idx = train_init()
     rng, rng_binarize, rng_init = random.split(rng, 3)
     sample_batch = binarize(rng_binarize, train_fetch(0, train_idx)[0])
-    opt_state, get_params = svi_init(rng_init, (sample_batch,), (sample_batch,))
+    svi_state, get_params = svi_init(rng_init, (sample_batch,), (sample_batch,))
 
     @jit
-    def epoch_train(opt_state, rng):
+    def epoch_train(svi_state, rng):
         def body_fn(i, val):
-            loss_sum, opt_state, rng = val
-            rng, rng_binarize = random.split(rng)
+            loss_sum, svi_state = val
+            rng_binarize = random.fold_in(rng, i)
             batch = binarize(rng_binarize, train_fetch(i, train_idx)[0])
-            loss, opt_state, rng = svi_update(rng, opt_state, (batch,), (batch,),)
+            svi_state, loss = svi_update(svi_state, (batch,), (batch,))
             loss_sum += loss
-            return loss_sum, opt_state, rng
+            return loss_sum, svi_state
 
-        return lax.fori_loop(0, num_train, body_fn, (0., opt_state, rng))
+        return lax.fori_loop(0, num_train, body_fn, (0., svi_state))
 
     @jit
-    def eval_test(opt_state, rng):
-        def body_fun(i, val):
-            loss_sum, rng = val
-            rng, rng_binarize, rng_eval = random.split(rng, 3)
+    def eval_test(svi_state, rng):
+        def body_fun(i, loss_sum):
+            rng_binarize = random.fold_in(rng, i)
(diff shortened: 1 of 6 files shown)


examples/vae.py
```diff
@@ -72,31 +72,31 @@ def main(args):
...
-        def body_fun(i, val):
-            loss_sum, rng = val
-            rng, rng_binarize, rng_eval = random.split(rng, 3)
+    def eval_test(svi_state, rng):
+        def body_fun(i, loss_sum):
+            rng_binarize = random.fold_in(rng, i)
             batch = binarize(rng_binarize, test_fetch(i, test_idx)[0])
-            loss = svi_eval(rng_eval, opt_state, (batch,), (batch,)) / len(batch)
+            # FIXME: does this lead to a requirement for an rng arg in svi_eval?
+            loss = svi_eval(svi_state, (batch,), (batch,)) / len(batch)
```
fehiepsi: there is only an issue with the merge here, I am not sure if we need to use different rng for svi_state when eval.

neerajprad: Since this is used on a hold out set, it seems harmless to me to use a stream which will be used again for training from that state onwards, but I could be missing some edge cases.

Issue #312, opened 2019-08-29
Add tests and allow users use vmap instead of pmap in MCMC


fehiepsi:
Probably this is the last piece to make our internal composable with jax api. Fixes #312

Working
- [x] sample_kernel
- [x] nuts_next
- [x] hmc_next
- [x] warmup adapter

PR #315, opened 2019-08-30
Add test for vmap

fehiepsi:
Addresses #312

Blocked by the next JAX release.

### DIFF
--- a/numpyro/contrib/autoguide/__init__.py
+++ b/numpyro/contrib/autoguide/__init__.py
@@ -326,11 +326,11 @@ class AutoIAFNormal(AutoContinuous):
     :param `**arn_kwargs`: keywords for constructing autoregressive neural networks, which includes:
 
         * **hidden_dims** (``list[int]``) - the dimensionality of the hidden units per layer.
-            Defaults to ``[latent_size, latent_size]``.
+          Defaults to ``[latent_size, latent_size]``.
         * **skip_connections** (``bool``) - whether to add skip connections from the input to the
-            output of each flow. Defaults to False.
+          output of each flow. Defaults to False.
         * **nonlinearity** (``callable``) - the nonlinearity to use in the feedforward network.
-            Defaults to :func:`jax.experimental.stax.Relu`.
+          Defaults to :func:`jax.experimental.stax.Relu`.
     """
     def __init__(self, model, prefix="auto", init_strategy=init_to_median,
                  num_flows=3, **arn_kwargs):
--- a/numpyro/mcmc.py
+++ b/numpyro/mcmc.py
@@ -211,7 +211,7 @@ def hmc(potential_fn, kinetic_fn=None, algo='NUTS'):
         step_size = lax.convert_element_type(step_size, xla_bridge.canonicalize_dtype(np.float64))
         nonlocal momentum_generator, wa_update, trajectory_len, max_treedepth, wa_steps
         wa_steps = num_warmup
-        trajectory_len = float(trajectory_length)
+        trajectory_len = trajectory_length
         max_treedepth = max_tree_depth
         z = init_params
         z_flat, unravel_fn = ravel_pytree(z)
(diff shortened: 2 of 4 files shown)


setup.py
```diff
@@ -31,7 +31,7 @@
     author_email='npradhan@uber.com',
     install_requires=[
         # TODO: pin to a specific version for the next release (unless JAX's API becomes stable)
-        'jax==0.1.43',
+        'jax==0.1.44',
```
fehiepsi: I pumped jax version to support this feature.

numpyro/mcmc.py
```diff
@@ -212,7 +212,7 @@ def init_kernel(init_params,
         step_size = lax.convert_element_type(step_size, xla_bridge.canonicalize_dtype(np.float64))
         nonlocal momentum_generator, wa_update, trajectory_len, max_treedepth, wa_steps
         wa_steps = num_warmup
-        trajectory_len = float(trajectory_length)
+        trajectory_len = trajectory_length
```
fehiepsi: step_size is float, so we don't need to cast trajectory_len to float. Historically, I guess these casting are necessary because we use float division at `num_steps = trajectory_length / step_size`

neerajprad: Yes, I think we needed this earlier if a user passed ints for both `step_size` and `trajectory_length` but with the `convert_element_type` in the first line, we don't need it anymore.

numpyro/mcmc.py
```diff
@@ -650,6 +650,11 @@ class MCMC(object):
     :param constrain_fn: Callable that converts a collection of unconstrained
         sample values returned from the sampler to constrained values that
         lie within the support of the sample sites.
+    :param str chain_method: One of 'parallel' (default), 'sequential', 'vectorized'. The method
```
neerajprad: Could you add a note saying that this is an experimental argument that may be removed in a future version without warning? 

nit: `hence allows us collecting samples` --> `hence allowing us to collect samples`.

numpyro/contrib/autoguide/__init__.py
```diff
@@ -326,11 +326,11 @@ class AutoIAFNormal(AutoContinuous):
     :param `**arn_kwargs`: keywords for constructing autoregressive neural networks, which includes:
 
         * **hidden_dims** (``list[int]``) - the dimensionality of the hidden units per layer.
-            Defaults to ``[latent_size, latent_size]``.
+          Defaults to ``[latent_size, latent_size]``.
```
fehiepsi: 4 indent here leads to [unintended format](https://numpyro.readthedocs.io/en/latest/autoguide.html#autoiafnormal)

numpyro/mcmc.py
fehiepsi: done

test/test_mcmc_interface.py
```diff
@@ -260,9 +263,56 @@ def model(labels):
...
+    if chain_method == 'parallel' and xla_bridge.device_count() == 1:
+        pytest.skip('parallel method requires device_count greater than 1.')
+    warmup_steps, num_samples = 100, 2000
+    # Here are settings which is currently supported.
+    rng = random.PRNGKey(2)
+    step_size = 1.
+    target_accept_prob = 0.8
+    trajectory_length = 1.
+    # Not supported yet:
+    #   + adapt_step_size
```
neerajprad: Why are these arguments not supported?

fehiepsi: For the first two ones, we can support easily by changing if else to lax.cond. It is also doable for max tree depth. However, I dont think that we need to support these features. Jit mcmc is valuable if we want to replicate mcmc with many different settings. To me, the most valuable ones are: step_size, trajectory length as done in many tfp papers (when users want to tune these hyperparameters), or data as in consensus Hmc. For num_warmup, num_samples, supporting them might be complicated because the sample shapes depend on num_samples. I also don’t know if users need to tune these values, so depending on user request, we will consider supporting them.

PR #316, opened 2019-08-30
Add additional options for MCMC sampling

neerajprad:
Adds 2 additional features as requested by @fehiepsi:
 - Ability to collect warmup samples via `collect_warmup` arg to `.run`.
 - Collect other fields via the `collect_fields` arg to `MCMC.run` - returns either the samples by default or a tuple corresponding to each of the collected fields.

### DIFF
--- a/numpyro/mcmc.py
+++ b/numpyro/mcmc.py
@@ -710,8 +710,9 @@ class MCMC(object):
         :param random.PRNGKey rng: Random number generator key to be used for the sampling.
         :param args: Arguments to be provided to the :meth:`numpyro.mcmc.MCMCKernel.init` method.
             These are typically the arguments needed by the `model`.
-        :param tuple, list collect_fields: Fields from :data:`numpyro.mcmc.HMCState` to collect
+        :param collect_fields: Fields from :data:`numpyro.mcmc.HMCState` to collect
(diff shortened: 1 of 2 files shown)


numpyro/mcmc.py
```diff
@@ -677,31 +678,42 @@ def __init__(self,
...
-    def run(self, rng, *args, init_params=None, **kwargs):
+    def run(self, rng, *args, collect_fields=('z',), collect_warmup=False, init_params=None, **kwargs):
         """
         Run the MCMC samplers and collect samples.
 
         :param random.PRNGKey rng: Random number generator key to be used for the sampling.
         :param args: Arguments to be provided to the :meth:`numpyro.mcmc.MCMCKernel.init` method.
             These are typically the arguments needed by the `model`.
+        :param tuple, list collect_fields: Fields from :data:`numpyro.mcmc.HMCState` to collect
+            during the MCMC run. By default, only the latent sample sites `z` is collected.
```
fehiepsi: nit: I think that this won't be rendered correctly with sphinx. the better way is
```
:param collect_fields:
:type collect_fields: tuple or list
```

neerajprad: Done.

PR #321, opened 2019-09-06
Update examples to new interface

fehiepsi:
TODO:
- [x] update tutorials


numpyro/hmc_util.py
```diff
@@ -769,9 +769,12 @@ def initialize_model(rng, model, *model_args, init_strategy=init_to_uniform, **m
...
                                     invert=True)
 
-    potential_fn = jax.partial(potential_energy, seeded_model, model_args, model_kwargs, inv_transforms)
+    # NB: we use model instead of seeded_model to prevent unexpected behaviours (if any)
+    potential_fn = jax.partial(potential_energy, model, model_args, model_kwargs, inv_transforms)
     if has_transformed_dist:
-        constrain_fun = jax.partial(constrain_fn, seeded_model, model_args, model_kwargs, inv_transforms)
+        # FIXME: why using seeded_model here triggers an error for funnel reparam example
+        # if we use MCMC class (mcmc function works fine)
+        constrain_fun = jax.partial(constrain_fn, model, model_args, model_kwargs, inv_transforms)
```
fehiepsi: I keep getting the error
```
Exception: Can't lift sublevels 2 to 1 
```
when using `seeded_model` with MCMC class in funnel reparam model.
```
  File "examples/funnel.py", line 46, in reparam_model                                                  
    y = numpyro.sample('y', dist.Normal(0, 3))                                                          
  File "/home/fehiepsi/numpyro/numpyro/primitives.py", line 62, in sample                               
  File "/home/fehiepsi/numpyro/numpyro/primitives.py", line 12, in apply_stack                          
    handler.process_message(msg)                                                                        
  File "/home/fehiepsi/numpyro/numpyro/handlers.py", line 334, in process_message                       
    self.rng, rng_sample = random.split(self.rng)                                                       
  File "/home/fehiepsi/miniconda3/envs/pydata/lib/python3.6/site-packages/jax/random.py", line 193, in s
plit                                                                                                    
    return _split(key, num)      
```
Anyway, I think it is safe to use `model` here because there should not have any stochastic behaviour at potential_fn and constrain_fn.

Because I was feeling that this is a dangerous issue so I mark this as `FIXME`. Will remove it when I understand the reason behind.

numpyro/mcmc.py
```diff
@@ -701,7 +701,7 @@ def _single_chain_mcmc(self, init, collect_fields=('z',), collect_warmup=False,
         if len(collect_fields) == 1:
             samples = (samples,)
         samples = dict(zip(collect_fields, samples))
-        samples['z'] = self.constrain_fn(samples['z'])
+        samples['z'] = vmap(self.constrain_fn)(samples['z'])
```
fehiepsi: Without vmap, we rely on broadcasting behaviour of transforms. In some examples, broadcasting does not work.

numpyro/hmc_util.py
fehiepsi: I believe that using `model` instead of `seeded_model` also makes things faster (we don't have to split keys when evaluating `potential_fn`.

fehiepsi:
In bayesian tutorial, most contents are kept as-is. I just rephrase the following paragraph

Thankfully, all of this is handled on the backend for us. Let us go through the steps one by one.

- JAX uses functional PRNGs. Unlike other languages / frameworks which maintain a global random state, in JAX, every call to a sampler requires an explicit PRNGKey. We will split our initial random seed for subsequent operations, so that we do not accidentally reuse the same seed.
- The function initialize_model takes a model along with model arguments (and keyword arguments), and returns a tuple of initial parameters, potential energy function, and constrain function. The initial parameters are used to initiate the MCMC chain, the potential energy function is a callable that when given unconstrained sample values returns the potential energy at these sample values. This is used by the verlet integrator in HMC. Lastly, constrain_fn is a callable that transforms the unconstrained samples returned by HMC/NUTS to sample values that lie within the constrained support.
- Finally, we use the mcmc function to run inference using the default NUTS sampler. Note that to run vanilla HMC, all you need to do is to pass algo='HMC' as argument to mcmc instead. This is a convenience utility that does all of the following: - Runs warmup - adapts steps size and mass matrix.
- Uses the sample from the warmup phase to start MCMC. - Return samples from the posterior distribution and print diagnostic information.

---
to

In NumPyro, there is a utility function initialize_model, which takes a model along with model arguments (and keyword arguments), and returns a tuple of initial parameters, potential energy function, and constrain function. The initial parameters are used to initiate the MCMC chain, the potential energy function is a callable that when given unconstrained sample values returns the potential energy at these sample values. This is used by the verlet integrator in HMC. Lastly, constrain function is a callable that transforms the unconstrained samples returned by HMC/NUTS to sample values that lie within the constrained support. For a concrete use case of the initialize_model utility, we refer to stochastic volativity example.

Thankfully, all of these are handled on the backend for us. Let us go through the steps one by one.

   - JAX uses functional PRNGs. Unlike other languages / frameworks which maintain a global random state, in JAX, every call to a sampler requires an explicit PRNGKey. We will split our initial random seed for subsequent operations, so that we do not accidentally reuse the same seed.
   - Finally, we use the MCMC class to run inference with NUTS sampler. Note that to run vanilla HMC, we will use HMC class. MCMC is a convenience class that has the following methods:
       + run(...): runs warmup, adapts steps size and mass matrix, and does sampling using the sample from the warmup phase.
       + print_summary(): print diagnostic information.
       + get_samples(): gets samples from the posterior distribution.

numpyro/mcmc.py
```diff
@@ -688,7 +688,9 @@ def _single_chain_mcmc(self, init, collect_fields=('z',), collect_warmup=False,
         hmc_state, constrain_fn = self.sampler.init(rng, self.num_warmup, init_params,
                                                     model_args=args, model_kwargs=kwargs)
         if self.constrain_fn is None:
-            self.constrain_fn = identity if constrain_fn is None else constrain_fn
+            constrain_fn = identity if constrain_fn is None else constrain_fn
+        else:
+            constrain_fn = self.constrain_fn
```
fehiepsi: We won't use `constrain_fn` elsewhere, so it is better to not set this property here. The advantage, as shown in bayesian regression tutorial, is that users don't have to redefine `kernel`, `mcmc` to get samples if args in `run()` method changes.

numpyro/mcmc.py
```diff
@@ -533,7 +533,7 @@ def __init__(self,
         self.dense_mass = dense_mass
         self.target_accept_prob = target_accept_prob
         self.trajectory_length = trajectory_length
-        self.sample_fn = None
+        self._sample_fn = None
```
fehiepsi: nit: this is to avoid having the same purpose as `self.sample`

numpyro/hmc_util.py
neerajprad: While this change looks okay to me, I would like to better understand the failure when using `seeded_model` since this might be indicative of other underlying issues.

neerajprad:
I just saw the tutorial, and the changes look good to me. I don't think we need to talk about `initialize_model` in the tutorials anymore, so good to see it go.

PR #324, opened 2019-09-07
Do not call random.split when sample site is observed

neerajprad:
This makes a minor optimization to seed to not call `random.split` if the site is already observed. This requires the following changes:
 - Using `condition` instead of `substitute` whenever possible, e.g. in the `predictive` utility.
 - Using `seed(condition(model, ..))` instead of `condition(seed(model, ..))`, the former being more efficient in that it will not call `random.split` for conditioned sites.

### DIFF
--- a/numpyro/handlers.py
+++ b/numpyro/handlers.py
@@ -330,7 +330,7 @@ class seed(Messenger):
         super(seed, self).__init__(fn)
 
     def process_message(self, msg):
-        if msg['type'] == 'sample':
+        if msg['type'] == 'sample' and not msg['is_observed']:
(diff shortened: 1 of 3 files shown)


numpyro/handlers.py
```diff
@@ -330,7 +330,7 @@ def __init__(self, fn, rng):
         super(seed, self).__init__(fn)
 
     def process_message(self, msg):
-        if msg['type'] == 'sample':
+        if msg['type'] == 'sample' and not msg['is_observed']:
```
fehiepsi: haha, I intended to do this too but worried about site effects.

PR #326, opened 2019-09-07
update benchmark notebook

fehiepsi:
This update logistic regression notebook with new api. Also includes benchmark results from Edward2 and Pyro.


fehiepsi:
@neerajprad I have just merged master and updated the interface: `collect_fields` -> `extra_fields`.

Issue #281, opened 2019-08-12
Pin jaxlib version in next release


kalradivyanshu:
Hey,

I was following: http://pyro.ai/numpyro/bayesian_regression.html, and tried to run the following code:

```
import jax
import jax.numpy as np
from jax import random, vmap
from jax.config import config; config.update("jax_platform_name", "cpu")
from jax.scipy.special import logsumexp
import matplotlib
import matplotlib.pyplot as plt
import numpy as onp
import pandas as pd
import seaborn as sns

from numpyro.diagnostics import hpdi
import numpyro.distributions as dist
from numpyro.handlers import sample, seed, substitute, trace
from numpyro.hmc_util import initialize_model
from numpyro.mcmc import mcmc

%matplotlib inline
plt.style.use('bmh')
plt.rcParams.update({'font.size': 16,
                     'xtick.labelsize': 14,
                     'ytick.labelsize': 14,
                     'axes.titlesize': 'large',
                     'axes.labelsize': 'medium'})
```

and it gave the error:

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-4-2f9c08a220e0> in <module>
     12 
     13 from numpyro.diagnostics import hpdi
---> 14 import numpyro.distributions as dist
     15 from numpyro.handlers import sample, seed, substitute, trace
     16 from numpyro.hmc_util import initialize_model

/usr/local/lib/python3.6/site-packages/numpyro/distributions/__init__.py in <module>
----> 1 from numpyro.distributions.continuous import (
      2     Beta,
      3     Cauchy,
      4     Chi2,
      5     Dirichlet,

/usr/local/lib/python3.6/site-packages/numpyro/distributions/continuous.py in <module>
     29 from jax.scipy.special import gammaln, log_ndtr, ndtr, ndtri
     30 
---> 31 from numpyro.distributions import constraints
     32 from numpyro.distributions.constraints import AbsTransform, AffineTransform, ExpTransform
     33 from numpyro.distributions.distribution import Distribution, TransformedDistribution

/usr/local/lib/python3.6/site-packages/numpyro/distributions/constraints.py in <module>
     26 from jax.scipy.special import expit, logit
     27 
---> 28 from numpyro.distributions.util import cumprod, cumsum, matrix_to_tril_vec, signed_stick_breaking_tril, sum_rightmost
     29 
     30 ##########################################################

[...]

Issue #184, opened 2019-06-01
pypi does not render README


fehiepsi:
Current [pypi description](https://pypi.org/project/numpyro/) does not show README properly. Looking like we get the same issue at [this thread](https://stackoverflow.com/questions/26737222/pypi-description-markdown-doesnt-work). This happens in Pyro too. I think that we need to update `setuptools`, `twine`, `wheel` (or pinning to some working versions).

neerajprad:
I realized this after uploading that the legacy command `python setup.py sdist upload` does not seem to render markdown, and we need to use twine to upload. Unfortunately, updating the description requires uploading a new release, so we will have to wait for the next one to fix this.

Pyro's README is actually fixed by @fritzo which works even with the legacy commands because it converts markdown to rst (and removes badges). Unfortunately, I didn't have `pypandoc` installed in my particular environment when I uploaded the last release and it didn't convert it into `rst`, but this should be fixed too with the next release. 

If markdown doesn't work for some reason, I think we can just use the tried and tested way of converting to rst as in Pyro. Let us keep the issue open until that is fixed.

neerajprad:
This should be fixed in the current release - https://pypi.org/project/numpyro/0.2.0/.

PR #328, opened 2019-09-08
Bump to version 0.2

neerajprad:


fehiepsi:
@neerajprad Will we need to update README file? Do we need to address #184 (PyPI rendering) and #281 (check version in examples/tutorials) too?

neerajprad:
> Do we need to address #184 (PyPI rendering) 

We shouldn't need any changes for this. We just need to use `twine` instead of `setup sdist`, I'll check again.

> and #281 (check version in examples/tutorials, sorry I just recall this - I did close that issue without noticing this task) too?

I think given all the breaking changes in JAX, this might be a good thing to do. I'll add it to the examples in this PR.

fehiepsi:
@neerajprad While editing release note, I found that MVN/PRNGIdentity is not exposed to the docs, could you pls add them in this PR too?

fehiepsi:
@neerajprad I have taken my pass at the release note. :)

PR #330, opened 2019-09-08
Add diverging diagnostics

fehiepsi:
This PR adds diverging field to HMC so we can know which transition is diverging. Surprisingly, the change to support this feature is small.


fehiepsi:
@neerajprad I think that this is a nice feature to have for 0.2 release. :)

PR #331, opened 2019-09-08
bump version in notebooks

jpchen:
missed one in #328, also use the canonical url for docs


neerajprad:
My bad, I always end up missing the bumping the version for the docs. Thanks for the fix, @jpchen!

PR #332, opened 2019-09-09
Fix issue of finding_initial_params when prior has sample shape kwarg

fehiepsi:
As title. I missed this situation when refactoring init_strategy.

### DIFF
--- a/test/test_mcmc_interface.py
+++ b/test/test_mcmc_interface.py
@@ -263,23 +263,21 @@ def test_diverging(kernel_cls, adapt_step_size):
         assert_allclose(num_divergences, num_warmup + num_samples)
 
 
-def test_schools_model():
+def test_prior_with_sample_shape():
     data = {
         "J": 8,
         "y": np.array([28.0, 8.0, -3.0, 7.0, -1.0, 1.0, 18.0, 12.0]),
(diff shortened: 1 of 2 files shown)


test/test_mcmc_interface.py
```diff
@@ -263,6 +263,25 @@ def model(data):
...
+        "sigma": np.array([15.0, 10.0, 16.0, 11.0, 9.0, 11.0, 10.0, 18.0]),
+    }
+
+    def schools_model():
+        mu = numpyro.sample('mu', dist.Normal(0, 5))
+        tau = numpyro.sample('tau', dist.HalfCauchy(5))
+        theta = numpyro.sample('theta', dist.Normal(mu, tau), sample_shape=(data['J'],))
+        numpyro.sample('obs', dist.Normal(theta, data['sigma']), obs=data['y'])
+
+    num_samples = 500
```
neerajprad: Thanks for adding this test. I think `sample_shape` isn't quite as heavily tested as it should be.

fehiepsi: Yeah, I just paid attention to this arg lately (when implementing `predictive` utility). In Pyro, I will use `plate` instead. >"< Incidently, I also discovered that this arg is popular among PyMC3 models so I think it is good to support it.

PR #334, opened 2019-09-14
move rng_seed out of svi

fehiepsi:
@neerajprad I think you might need this separation to vectorize elbo in order to support num_particles? With this, I think we can use `vectorized_elbo` with SVI interface.

```
def vectorized_elbo(rng, param, model, guide, ..., num_particles=1):
    # if num_particles > 1
    return np.mean(vmap(lambda rng: elbo(rng, param, model, ...))(random.split(rng, num_particles)))
```


neerajprad:
Thanks for sending this PR! I had made the changes locally, but didn't send out a PR for this.

Issue #66, opened 2019-03-25
Support compatibility module in numpyro to support Pyro API


neerajprad:
As discussed in https://github.com/pyro-ppl/pyro/pull/1790, to support models written in Pyro, we can have a `compatibility` wrapper around distributions and inference algorithms. 

I think we may have to think about some of the interface issues here, because the efficient way of doing things in Pyro (using classes, storing state) may not be ideal in Numpyro. Some of this can be abstracted within this compatibility module, but other things like user `for` loops may need to be re-written in jax to make them more efficient.

Some of the questions that we will need to think about (from https://github.com/pyro-ppl/pyro/pull/1790), and my initial thoughts. I might expand this list as I work more on this, but feel free to add your thoughts:

 - There is no global param store, so to get something like `pyro.get_param_store` to work, we'll need to simply cache the optimizer state somewhere towards the end of svi training. A more faithful implementation will unfortunately need to copy arrays every time the optimizer state is updated, and that will be too expensive. It might be best to not provide a compatibility wrapper for this, at least to begin with.
 - the loop `for ..: svi.step()` will be suboptimal, and should ideally be replaced by `lax.fori_loop` when users are on the JAX backend. This could be an order of magnitude faster! - I think the solution here might be to absorb the for loop within SVI's API so that Numpyro and Pyro can handle it differently.
 - There is no `pyro.module`, because the stax NNs don't have any parameters that need to be registered. All parameters need to be cycled back and forth with the update function. I think it should be okay to skip `pyro.module` as well since we don't need to register any NN parameters.

neerajprad:
Closing this - we can open more specific issues later.

PR #336, opened 2019-09-17
Add a simplified plate context manager

### LINKED ISSUE #41: Add support for plate context manager
We can start with a basic implementation that just does distributions broadcasting under the hood and stores the plate information at sample sites for any inference algorithms.

neerajprad:
Fixes #41. Additionally, this is a blocker for #66. 

This adds a simplified version of the context manager used in Pyro models. This has the same broadcasting and scaling semantics as Pyro's plate context manager. 

There are a few simplifications - `size` argument is mandatory, there is no `subsample` argument and the context manager does not return indices like `arange` or samples from a categorical distribution. If needed, we can support these arguments later, but it is unclear if our simple interface for SVI/HMC without enumeration would require a more complex implementation.

### DIFF
--- a/test/test_handlers.py
+++ b/test/test_handlers.py
@@ -1,7 +1,10 @@
+import functools
+from operator import mul
+
 from numpy.testing import assert_allclose
 import pytest
 
-from jax import random
+from jax import random, jit
 
 import numpyro
 from numpyro import handlers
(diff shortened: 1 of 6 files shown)


test/test_handlers.py
```diff
@@ -58,3 +58,61 @@ def model():
...
+
+
+@pytest.mark.parametrize('model', [
+    model_nested_plates_0,
+    model_nested_plates_1,
+    model_nested_plates_2,
+    model_subsample_1,
+])
+def test_plate(model):
+    jit(handlers.seed(model, random.PRNGKey(1)))
```
fehiepsi: nit: I think you need to call this function.

fehiepsi: To be safe, could you return all sample sites in all models and add a statement to check if the output of jitted model is the same as non-jitted one?

neerajprad: Done; just checking the trace values of sample sites.

PR #340, opened 2019-09-18
Implement AutoBNAFNormal autoguide

### LINKED ISSUE #256: Sampling issues with NeutraHMC

fehiepsi:
Based on #230, we can simply create a BNAF autoguide. Fix #256.

#### Tasks

- [x] add tests
- [x] try on neutra example
- [x] debug the instability issue


examples/neutra.py
```diff
@@ -49,7 +44,7 @@ def make_transformed_pe(potential_fn, transform, unpack_fn):
     def transformed_potential_fn(z):
         u, intermediates = transform.call_with_intermediates(z)
         logdet = transform.log_abs_det_jacobian(z, u, intermediates=intermediates)
-        return potential_fn(unpack_fn(u)) + logdet
+        return potential_fn(unpack_fn(u)) - logdet
```
fehiepsi: this is a bug :(

examples/neutra.py
```diff
@@ -11,32 +14,30 @@
...
 from numpyro.infer import MCMC, NUTS, SVI
 from numpyro.infer.util import initialize_model, transformed_potential_energy
 
-# TODO: remove when the issue https://github.com/google/jax/issues/939 is fixed upstream
-# The behaviour when training guide under fast math mode is unstable.
-os.environ["XLA_FLAGS"] = "--xla_cpu_enable_fast_math=false"
+
+# XXX: upstream logsumexp throws NaN under fast-math mode + MCMC's progress_bar=True
+def logsumexp(x, axis=0):
+    return np.log(np.sum(np.exp(x), axis=axis))
```
fehiepsi: It took me a day to debug the issue: NeuTra throws nan when using fast-math mode and enable progress_bar. We have noticed this issue previously: https://github.com/pyro-ppl/numpyro/pull/280

It is pretty hard to isolate the issue: I trained AutoBNAF with fast_math disabled, store params in a file, then enable fast-math, load params, create transform_potential_fn, run MCMC, get NaN... NaN also happens for HMC with only 1 leapfrog step. However, when I run the utility velocity_verlet alone (out of MCMC scope), NaN doesn't happen. All I can tell (after diving deeper into the source code): somehow transform_pe has NaN grad under fast-math mode.

fehiepsi: Interestingly, using `np.log(np.sum(np.exp(x - np.max(x)), axis=axis)) + np.max(x)` here also throws NaN. I have no idea why... I believe 0 x inf is happening in grad somewhere here. If we face this NaN, it is better to disable fast-math mode, until JAX fast-math is more stable.

setup.py
```diff
@@ -46,7 +46,7 @@
             'pyro-api>=0.1.1'
         ],
         'dev': ['ipython', 'isort'],
-        'examples': ['matplotlib'],
+        'examples': ['matplotlib', 'seaborn'],
```
fehiepsi: I added seaborn here because matplotlib does not have a similar function for kdeplot.

examples/neutra.py
neerajprad: That's some serious sleuthing! Any ideas on what kind of values does it throw `nan`?

numpyro/contrib/autoguide/__init__.py
```diff
@@ -543,6 +544,54 @@ def _get_transform(self):
         return ComposeTransform(flows)
 
 
+class AutoBNAFNormal(AutoContinuous):
+    """
+    This implementation of :class:`AutoContinuous` uses a Diagonal Normal
+    distribution transformed via a
+    :class:`~numpyro.distributions.iaf.InverseAutoregressiveTransform`
```
neerajprad: Update docstring?

examples/neutra.py
```diff
@@ -2,6 +2,9 @@
 # SPDX-License-Identifier: Apache-2.0
 
 """
+Neural Transport
+================
+
 This example illustrates how to use a trained AutoIAFNormal autoguide to transform a posterior to a
```
neerajprad: We should update this description too.

examples/neutra.py
fehiepsi: The input to DualMoon log_prob is just a usual value ~ `np.array([-4., 0.])`. Then, the input to `logsumexp` will be around `np.array([-50., -5.])`. Grad at this value will be `(3e-20, 1)`.

fehiepsi:
@neerajprad I copy the image from the gallery here
sphx_glr_neutra_001

numpyro/contrib/autoguide/__init__.py
fehiepsi: opps... will fix!

examples/neutra.py
fehiepsi: Thanks!

examples/neutra.py
neerajprad: I thought this was the function throwing nan, but it seems that the nan is being thrown somewhere else even though the logsumexp function gives finite gradients.

fehiepsi: This is a bit mysterious. Given the `params` which throw `nan` in grad (I know `grad` is `nan` by changing transition prob logic to capture NaN grad in hmc_state), `value_and_grad(potential_fn)(params)` gives finite results when I run it outside MCMC.

PR #342, opened 2019-09-19
Add a note for sample primitive

### LINKED ISSUE #337: Document behavior of primitive statements in README
Hello! I'll preface this issue by stating that I'm new to numpyro, so there's a significant chance that I'm not using `numpyro.sample` correctly.

When building models, I want to test draws from my random variables to make sure I'm passing parameters correctly, getting the kwargs correct, etc.; I'll lift a line out of a model function, [like this one](https://github.com/pyro-ppl/numpyro/blob/master/examples/baseball.py#L111), and run it in an IPython window to see if I get an error. It seems like I'm unable to do this by itself in an IPython console.

## Minimal example

```
In [1]: import numpyro

In [2]: import numpyro.distributions as d

In [3]: numpyro.__version__
Out[3]: '0.2.0'

In [4]: numpyro.sample("x", d.Normal(0, 1))
KeyError                                  Traceback (most recent call last)
<ipython-input-4-21ce96e72ad0> in <module>
----> 1 numpyro.sample("x", d.Normal(0, 1))

~/miniconda3/envs/fusion/lib/python3.7/site-packages/numpyro/primitives.py in sample(name, fn, obs, sample_shape)
     45     # if there are no active Messengers, we just draw a sample and return it as expected:
     46     if not _PYRO_STACK:
---> 47         return fn(sample_shape=sample_shape)
     48
     49     # Otherwise, we initialize a message...

~/miniconda3/envs/fusion/lib/python3.7/site-packages/numpyro/distributions/distribution.py in __call__(self, *args, **kwargs)
    161
    162     def __call__(self, *args, **kwargs):
--> 163         key = kwargs.pop('random_state')
    164         sample_intermediates = kwargs.pop('sample_intermediates', False)
    165         if sample_intermediates:

KeyError: 'random_state'
```

## Thoughts
I can see that in [numpyro.sample](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/primitives.py#L47), we return `fn(sample_shape=sample_shape)`, which in this case attempts to call `dist.Normal(0, 1)(sample_shape=sample_shape)`. However, looking at [distributions.py](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/distributions/distribution.py#L163), it seems that `Distribution.sample` expects a `random_state` kwarg that isn't getting passed.

When I do the following, everything is fine:

```
In [5]: import jax.random as random

In [6]: d.Normal(0, 1).sample(random.PRNGKey(12))
/Users/chtu8001/miniconda3/envs/fusion/lib/python3.7/site-packages/jax/lib/xla_bridge.py:114: UserWarning: No GPU/TPU found, falling back to CPU.
  warnings.warn('No GPU/TPU found, falling back to CPU.')
Out[6]: DeviceArray(-0.5135873, dtype=float32)

In [7]: d.Normal(0, 1).sample(random.PRNGKey(12), sample_shape=(4, ))
Out[7]:
DeviceArray([-1.31179953, -0.70821768,  0.18907626, -1.09385514],
            dtype=float32)
```

so I expect that we can't use `numpyro.sample` outside a model (although it's not totally clear to me how that is defined, something to do with Messengers and PYRO_STACK). I'm wondering if this is by design and I should just use the second, working example, or if I'm misunderstanding how `numpyro.sample` should be used.

Thanks!

fehiepsi:
Fixes #337. Hopefully, this makes the docs clearer.

### DIFF
--- a/numpyro/primitives.py
+++ b/numpyro/primitives.py
@@ -89,7 +89,7 @@ def sample(name, fn, obs=None, random_state=None, sample_shape=()):
         'name': name,
         'fn': fn,
         'args': (),
-        'kwargs': {'sample_shape': sample_shape},
+        'kwargs': {'random_state': random_state, 'sample_shape': sample_shape},
         'value': obs,
         'scale': 1.0,
         'is_observed': obs is not None,
--- a/numpyro/primitives.py
+++ b/numpyro/primitives.py
@@ -60,7 +60,7 @@ class Messenger(object):
             return self.fn(*args, **kwargs)
 
 
-def sample(name, fn, obs=None, sample_shape=()):
+def sample(name, fn, obs=None, random_state=None, sample_shape=()):
     """
     Returns a random sample from the stochastic function `fn`. This can have
     additional side effects when wrapped inside effect handlers like
@@ -68,22 +68,20 @@ def sample(name, fn, obs=None, sample_shape=()):
 
     .. note::
         By design, `sample` primitive is meant to be used inside a NumPyro model.
-        Then `seed` handler is used to inject a random state to `fn`. Because
-        there is no global random state in NumPyro, NumPyro distributions'
-        `__call__` method requires an explicit keywork `random_state`. Hence, to
-        use `sample` outside of NumPyro model, we can use the following pattern
-
-        >>> numpyro.sample('x', partial(dist.Normal(0, 1), random_state=random.PRNGKey(0)))  # doctest: +SKIP
+        Then :class:`~numpyro.handlers.seed` handler is used to inject a random
+        state to `fn`. In those situations, `random_state` keyword will take no
+        effect.
 
     :param str name: name of the sample site
     :param fn: Python callable
     :param numpy.ndarray obs: observed value
+    :param jax.random.PRNGKey random_state: an optional random key for `fn`.
     :param sample_shape: Shape of samples to be drawn.
     :return: sample from the stochastic `fn`.
     """
     # if there are no active Messengers, we just draw a sample and return it as expected:
     if not _PYRO_STACK:
-        return fn(sample_shape=sample_shape)
+        return fn(random_state=random_state, sample_shape=sample_shape)
 
     # Otherwise, we initialize a message...
     initial_msg = {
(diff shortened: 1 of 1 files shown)


numpyro/primitives.py
```diff
@@ -66,6 +66,15 @@ def sample(name, fn, obs=None, sample_shape=()):
     additional side effects when wrapped inside effect handlers like
     :class:`~numpyro.handlers.substitute`.
 
+    .. note::
+        By design, `sample` primitive is meant to be used inside a NumPyro model.
+        Then `seed` handler is used to inject a random state to `fn`. Because
+        there is no global random state in NumPyro, NumPyro distributions'
+        `__call__` method requires an explicit keywork `random_state`. Hence, to
```
neerajprad: keywork -> keyword

> use `sample` outside of NumPyro model, we can use the following pattern

I think its best to not advertise this convoluted pattern and instead recommend users to just call the distribution's `sample` method with the rng key.

neerajprad:
Also, as you suggested, I think it will be nice to be able to include `random_state` directly as argument to the `sample` primitive. Otherwise the first couple of lines are useless without that.

```
    if not _PYRO_STACK:
        return fn(sample_shape=sample_shape)
```

fehiepsi:
Yeah, including `random_state` directly seems reasonable to me. I think those lines in Pyro supports `sample` from any stochastic function given its input *args, **kwargs. However, I don't know about any usage case beyond `fn` is a distribution. Do you have any idea about it?

neerajprad:
> However, I don't know about any usage case beyond fn is a distribution. Do you have any idea about it?

None really, since all distributions will need RNG, and it should be safe to do that. This will also help us have a way to sample from distributions in the generic interface without having to pass rng keys in Pyro.

numpyro/primitives.py
```diff
@@ -60,21 +60,28 @@ def __call__(self, *args, **kwargs):
...
     :param fn: Python callable
     :param numpy.ndarray obs: observed value
+    :param jax.random.PRNGKey random_state: an optional random key for `fn`.
     :param sample_shape: Shape of samples to be drawn.
     :return: sample from the stochastic `fn`.
     """
     # if there are no active Messengers, we just draw a sample and return it as expected:
     if not _PYRO_STACK:
-        return fn(sample_shape=sample_shape)
+        return fn(random_state=random_state, sample_shape=sample_shape)
```
neerajprad: If `random_state` is not None, we should also include it in kwargs in the msg below.

fehiepsi: I agree, but it is best to not use this pattern in our examples/tutorials. >"<

PR #344, opened 2019-09-19
Add compatibility module for pyro.generic

neerajprad:
This seeds the `numpyro.compat` module with some basic wrappers for MCMC in preparation for #66. This will allow us to start testing the generic module in Pyro, and prepare for a more involved design discussion. Tests with the generic module will be in Pyro.

### DIFF
--- a/numpyro/compat/handlers.py
+++ b/numpyro/compat/handlers.py
@@ -1,0 +1,11 @@
+from jax import random
+
+from numpyro.handlers import *  # noqa: F401, F403
+from numpyro.handlers import seed as _seed
+
+
+# This is so that users do not have to import PRNGKey from jax.random.
+# XXX: Should we make this the default?
+class seed(_seed):
+    def __init__(self, fn, rng):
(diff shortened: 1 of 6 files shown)


numpyro/compat/handlers.py
```diff
@@ -0,0 +1,11 @@
+from jax import random
+
+from numpyro.handlers import *  # noqa: F401, F403
+from numpyro.handlers import seed as _seed
+
+
+# This is so that users do not have to import PRNGKey from jax.random.
+# XXX: Should we make this the default?
```
fehiepsi: I think we can hold a global `seed` (similar to _PYRO_STACK), which will change its value when `pyro.set_rng_seed(...)` is called. We will update this seed before feeding it into SVI.init or MCMC.run.

> Should we make this the default?

Sure. That would be convenient. Our current API does not require users use `random.split(...)`.

fehiepsi:
Sorry, I miss this

> Tests with the generic module will be in Pyro.

Issue #341, opened 2019-09-19
Benchmark MCMC vectorized method


fehiepsi:
We need to benchmark `vmap + fori_collect` vs `fori_collect + vmap`. Using the latter will allow us add progress bar for multi-chain (issue #309). In addition, the latter one will help us benchmark the effect of vectorizing chains having different trajectory lengths (because we can easily ignore compiling time).

Issue #309, opened 2019-08-28
Add progress bar for HMC num_chains > 1


neerajprad:
Right now, it is very hard to ascertain progress when running MCMC with num_chains > 1. Let us figure out ways to show progress bar updates. Even a progress bar without diagnostic updates will be useful.

Currently, chain methods that support progress bar:

- [x] parallel
- [x] sequential
- [x] vectorized

With this, we can make the following enhancement.

- [x] Output other metadata like when the model is compiling and when it is sampling. This is especially useful when we run multiple chains and it is not clear whether all the time is being taken to compile the model or if it is the sampling part that is taking time.

fehiepsi:
Hmm, is this possible? AFAIK, the progressbar will only show the progress of compiling the code, while the real execution is hidden. Hopefully, there will be some tools in JAX allowing us measure this progress.

neerajprad:
> is this possible?

Maybe not yet; it will likely depend on solutions for https://github.com/google/jax/issues/364. For sequential chains, it might be possible (though I am not fully sure).

fehiepsi:
JAX now supports an experimental [host_callback](https://github.com/google/jax/blob/master/jax/experimental/host_callback.py). It seems that this is doable now, but we need to study how that module works.

jeremiecoullon:
If it's helpful: this [answered question](https://github.com/google/jax/discussions/4763#discussioncomment-121452) in the JAX Discussion shows how to print the progress of an MCMC chain within a compiled function (using the `host_callback` module mentioned above).

fehiepsi:
That's a great news @jeremiecoullon! Have you come up with a solution for MCMC chain yet? This would be one of the most requested feature from NumPyro users. :)

jeremiecoullon:
So I did a quick test which seems to work. 

### Suggested solution:

In `numpyro/util.py` I added the following functions:

```python
from jax.experimental import host_callback

def _print_consumer(arg, transrorm):
    i, n_iter = arg
    print(f"Iteration {i}/{n_iter}")

@jit
def progress_bar(arg, result):
    """
    Print progress of loop only if iteration number is a multiple of the print_rate

    Usage: carry = progress_bar((iter_num, n_iter, print_rate), carry)
    """
    i, n_iter, print_rate = arg
    result = lax.cond(
        i%print_rate==0,
        lambda _: host_callback.id_tap(_print_consumer, (i, n_iter), result=result),
        lambda _: result,
        operand=None)
[...]

PR #345, opened 2019-09-21
use fori + vmap instead of vmap + fori for vectorized method

fehiepsi:
Addresses #341 and #309. We can have progress_bar with sequential and vectorized now. I also added a note that sampling will be faster when progress_bar=False.

Test already covers the change. An illustration can be found in
https://nbviewer.jupyter.org/gist/fehiepsi/b9ff5cb6df8b110172471084df1cbe99

### DIFF
--- a/numpyro/util.py
+++ b/numpyro/util.py
@@ -128,7 +128,7 @@ def fori_collect(lower, upper, body_fun, init_val, transform=identity, progbar=T
                 val = jit(body_fun)(val)
                 if i >= lower:
                     collection.append(jit(ravel_fn)(val))
-                t.set_description(progbar_desc(val), refresh=False)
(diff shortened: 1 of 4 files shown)


numpyro/util.py
```diff
@@ -72,6 +72,20 @@ def fori_loop(lower, upper, body_fun, init_val):
         return lax.fori_loop(lower, upper, body_fun, init_val)
 
 
+def lax_map(f, xs):
```
neerajprad: Since this is a very particular use case, let us keep it private to mcmc class, or add some docstring so that this isn't used accidentally instead of `lax.map`.

fehiepsi: yeah, that makes sense

Issue #343, opened 2019-09-19
Adapt Unit distribution / pyro.factor from Pyro


fehiepsi:
We have some examples which uses Delta distribution to inject log_prob (hmm, neutra), so it is good to have a utility for those patterns:
+ inject a log_prob term
```
def model():
    log_prob = ...
    pyro.sample('log_prob', dist.Delta(0., log_prob), obs=0.)
```
+ convert a log_prob_fn to a distribution (without having to define a custom distribution). Currently we use the following way (which is not quite accurate): the log_prob at `x` site `dist.Normal(0, 100000)` will add a large negative *constant* to model's log_prob, which makes it harder to interprete the `log_density` output.
```
def model():
    x = pyro.sample('x', dist.Normal(0, 100000))
    log_prob = log_prob_fn(x)
    pyro.sample('log_prob', dist.Delta(0., log_prob), obs=0.)
```

**Edit:** It is better to define a custom distribution for this purpose.

A related issue is to have a utility for [transformed_pe](https://github.com/pyro-ppl/numpyro/blob/master/examples/neutra.py#L49) so users can avoid bugs related to the signs of log_det terms. It can be useful for 2 usage cases:
+ **density matching** (to sample from a known density function): When the density p(x) is known, we approximate it by a flow `flow(z)` where z ~ q(z) ~ N(0, 1) (which is what AutoContinuous does). Then instead of sampling from p(x) directly, we sample from q(z) and transform those samples using `flow`. To get those `z`, we can sample from N(0, 1) directly or using MCMC (neutra) with the objective: ``p(flow(z)).|det(flow)|``.
+ **density estimation** (to compute density from samples): When data x~p(x) is known but `p` is unknown, we use a flow `flow` to convert `p` to an easy-to-evaluate density `q=N(0,1)`, that is we are using SVI with a `None` guide and `model` is ``q(flow(x))|det(flow)|``.

In both cases, our objective is to compute base_dist's log_prob, given dist's log_prob. So it is good to have support for InverseTransform #164. In NumPyro modeling language, they are equivalent to
+ the first case
```
def model():
    # NB: we need to have a mechanism to disable automatic reparameterization
    sample('z', dist.TransformedDistribution(p_dist, flow.inv))
```
+ the second case
```
def model():
    sample('x', dist.TransformedDistribution(q_dist, flow.inv), obs=X)
```

neerajprad:
[...]

PR #348, opened 2019-09-24
Fix the bug of computing transformed potential energy in neutra example

fehiepsi:
This PR fixes the error at #256 due to . In addition, I moved transformed_pe to `infer_util` because it is useful for various purposes (as explained at #343).

A transform UnpackTransform is introduced to reflect better the purpose of `get_transform` in autoguide.

### issue

it is hard to fit AuroIAFNormal to this potential function when we remove x's prior. :( I'll use BNAF instead when #230 is merged.

### DIFF
--- a/numpyro/contrib/autoguide/__init__.py
+++ b/numpyro/contrib/autoguide/__init__.py
@@ -12,7 +12,7 @@ from numpyro import handlers
 from numpyro.contrib.nn.auto_reg_nn import AutoregressiveNN
 import numpyro.distributions as dist
 from numpyro.distributions import constraints
-from numpyro.distributions.constraints import AffineTransform, ComposeTransform, PermuteTransform, biject_to
+from numpyro.distributions.constraints import AffineTransform, ComposeTransform, PermuteTransform, UnpackTransform, biject_to
 from numpyro.distributions.flows import InverseAutoregressiveTransform
 from numpyro.distributions.util import sum_rightmost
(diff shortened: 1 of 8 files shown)


numpyro/infer_util.py
```diff
@@ -120,6 +121,23 @@ def potential_energy(model, model_args, model_kwargs, inv_transforms, params):
     return - log_joint
 
 
+def transformed_potential_energy(potential_energy, inv_transform, z):
+    """
+    Given a potential energy `p(x)`, compute potential energy of `p(z)`
```
neerajprad: +1 on the choice of variable names and this docstring. Its surprising how much can keeping documentation and variable naming straight help notice bugs like this.

numpyro/contrib/autoguide/__init__.py
```diff
@@ -225,7 +230,8 @@ def get_transform(self, params):
         :return: the transform of posterior distribution
         :rtype: :class:`~numpyro.distributions.constraints.Transform`
         """
-        return handlers.substitute(self._get_transform, params)()
+        return ComposeTransform([handlers.substitute(self._get_transform, params)(),
```
neerajprad: Are there any places where we are using `guide.unpack_latent` anymore if `get_transform` does the unpacking by default? If that is not the case, can we make this private `_unpack_latent` instead?

fehiepsi: Yes, we don't need those `unpack_latent` anymore! Nice catch!

PR #350, opened 2019-09-24
Allow seed handler to be used as a context manager

neerajprad:
This allows for usage like the following:

```python
with handlers.seed(rng=1):
  numpyro.sample('x', dist.Normal(0., 1.))
```

**Changes:**
 - Allows for using `seed` as a contextmanager, so that we don't need to wrap things in a callable.
 - Changes the seed handler to take in Python ints so that we don't have to do `seed(random.PRNGKey(1)` and can instead just write `seed(1)`.

cc. @martinjankowiak

### DIFF
--- a/numpyro/handlers.py
+++ b/numpyro/handlers.py
@@ -312,7 +312,9 @@ class seed(Messenger):
     so that we use a fresh seed for each subsequent call without having to
     explicitly pass in a `PRNGKey` to each `sample` call.
     """
-    def __init__(self, fn, rng):
+    def __init__(self, fn=None, rng=None):
+        if isinstance(rng, int):
+            rng = random.PRNGKey(rng)
         self.rng = rng
         super(seed, self).__init__(fn)
 
--- a/numpyro/handlers.py
+++ b/numpyro/handlers.py
@@ -311,6 +311,28 @@ class seed(Messenger):
     primitive inside the function results in a splitting of this initial seed
     so that we use a fresh seed for each subsequent call without having to
     explicitly pass in a `PRNGKey` to each `sample` call.
+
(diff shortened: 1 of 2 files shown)


martinjankowiak:
should this go into some docstring/example to highlight usage?

neerajprad:
> should this go into some docstring/example to highlight usage?

Good point; I'll add it to the docstring. I'm also making some changes to the readme so we can highlight this usage pattern while addressing #337.

neerajprad:
> is discouraged because the keys might be reused (this is just a minor point to me)

I don't think the keys should get reused. If we have another sample inside it will sample with the split subkey, so this should be safe.

> works under vmap so I think we should test it. 

This should work with vmap, since in that case we will be passing a PRNGKey. I'll add a test.

neerajprad:
> not sure if if isinstance(rng, int) works under vmap so I think we should test it.

Oh..you mean vmapping over something like `np.arange(100)`. Hmm..yes, that will surely not work, let me try your suggestion.

fehiepsi:
Re reusing keys: I mean that if users want to draw a lot of samples (e.g. to test if some samplers are biased), then keys derived from some initial seeds might collide (but I am not sure if that will happen ^^).

neerajprad:
I think the splitting model is such that if we start from 0, 1, 2.., we will get a very different stream of random numbers by calling `random.split` so this shouldn't be an issue.

fehiepsi:
Yeah, I also think it might not be an issue. But I don't know why I usually have a bad feeling about reusing key stuff. :))

neerajprad:
I added a test with vmapping over an arange.

Issue #351, opened 2019-09-25
Trying to reproduce the textmessage-example from bayesian-methods-for-hackers


timnon:
I had some problems recreating the textmessage-example from bayesian methods for hackers (https://nbviewer.jupyter.org/github/CamDavidsonPilon/Probabilistic-Programming-and-Bayesian-Methods-for-Hackers/blob/master/Chapter1_Introduction/Ch1_Introduction_PyMC2.ipynb).

This example requires a `DiscreteUniform`-distribution, which can be created via a `Categorical`. However, the following model stalled when sampling from the posterior:

```
import numpyro
import numpyro.distributions as dist
from numpyro.mcmc import MCMC, NUTS
import jax
import numpy as np

n_count_data = 10

def model():
    tau_ = numpyro.sample("tau_", dist.Categorical(np.repeat(1./n_count_data, n_count_data)), obs=np.arange(n_count_data))

key = jax.random.PRNGKey(0)
key, key_ = jax.random.split(key)
num_warmup, num_samples = 100, 1000

kernel = NUTS(model)
mcmc = MCMC(kernel, num_warmup, num_samples)
mcmc.run(key_)
mcmc.print_summary()
```

I would expect this to terminate.

fehiepsi:
Hi @timnon, the issue is your model does not have any latent variable for MCMC to run over. Your `tau_` is an observed variable. You can take a look at this [gist](https://nbviewer.jupyter.org/gist/fehiepsi/aa6ad7ff128ba57ffc46ca1101bb35fc) to see how to make a corresponding model in NumPyro.

fehiepsi:
@timnon Thanks for raising this issue. I am going to close this issue because the site effect that MCMC stuck when the model has no latent site has been addressed in #352. For a repro code of the textmessage example, you can take a look at the gist in the above comment. :)

timnon:
Thanks for the feedback, would be indeed good to have some error message in case there is no variable to sample from. Also thanks for the gist, looks more sane than my attempt. In general, is there already some attempt to adapt bayesian-methods-for-hackers to numpyro? Or is it too early due to changing interfaces, e.g. missing `DiscreteUniform`, but which can be easily created like in the gist.

fehiepsi:
[...]

PR #352, opened 2019-09-25
let mcmc run with empty latent sites

fehiepsi:
Address #351. This PR fixes a problem which happens in an empty model: step_size keeps increasing forever in `find_reasonable_step_size` function.

@neerajprad probably we don't want to run mcmc at all if the model has no latent site, but it is not addressed in this PR. I let you justify the correct behavior in that case.

### DIFF
--- a/numpyro/mcmc.py
+++ b/numpyro/mcmc.py
@@ -220,7 +220,6 @@ def hmc(potential_fn, kinetic_fn=None, algo='NUTS'):
         z = init_params
         z_flat, unravel_fn = ravel_pytree(z)
         momentum_generator = partial(_sample_momentum, unravel_fn)
-        adapt_step_size = adapt_step_size if init_params else False
 
         find_reasonable_ss = partial(find_reasonable_step_size,
                                      potential_fn, kinetic_fn,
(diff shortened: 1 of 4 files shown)


numpyro/mcmc.py
```diff
@@ -738,7 +738,7 @@ def _single_chain_mcmc(self, init, collect_fields=('z',), collect_warmup=False,
         if len(collect_fields) == 1:
             samples = (samples,)
         samples = dict(zip(collect_fields, samples))
-        samples['z'] = vmap(constrain_fn)(samples['z'])
+        samples['z'] = vmap(constrain_fn)(samples['z']) if len(tree_flatten(samples)[0]) > 0 else samples['z']
```
neerajprad: if `samples['z']` is empty does vmap complain?

neerajprad:
> @neerajprad probably we don't want to run mcmc at all if the model has no latent site, but it is not addressed in this PR. I let you justify the correct behavior in that case.

I think it is better to just let it run through the code rather than short-circuiting it, since it should be quite fast anyhow.

numpyro/mcmc.py
fehiepsi: Yes, `vmap` will complain about it: `import jax; jax.vmap(lambda x: x)({})` triggers
```
ValueError: not enough values to unpack (expected 1, got 0)
```

fehiepsi:
Agreed! It is quite fast! :D

Issue #349, opened 2019-09-24
Higher level interface for predictions and arviz integration


neerajprad:
### ArviZ integration

In https://github.com/arviz-devs/arviz/pull/811, I already made an attempt for this integration. As demonstrated in [InferenceData introduction](https://arviz-devs.github.io/arviz/notebooks/XarrayforArviZ.html), there are several groups of an inference data (prior/posterior/predictive/log_lik/inferece_stats/observed_data/...). Currently, users can do something like [NumPyro section in ArviZ cookbook](https://arviz-devs.github.io/arviz/notebooks/InferenceDataCookbook.html#From-NumPyro). The `az.from_numpyro` requires 3 args `posterior` (which is `mcmc`), `prior` (i.e. prior_predictive), `posterior_predictive`. Currently, we lack a handy utility for prior predictive.

Internally, we lack `observed_data` and `log_lik` groups. The reason is given a `mcmc` instance, we do not store `args`, `kwargs`, so even that we can access to model, we still unable to run it from `arviz`. `observed_data` is required for [ppcplot](https://arviz-devs.github.io/arviz/examples/plot_ppc.html) or [plot_loo_pit](https://arviz-devs.github.io/arviz/examples/plot_loo_pit_overlay.html). `log_lik` is useful for many stuffs, including some bayesian criterions.

**Solutions:**

+ store args, kwargs of `run` method to MCMC state.
+ make prior_predictive, log_like_predictive utilities.

### SVI

I think that current `predictive` utility is enough to get samples from `guide` (similar to the sample_posterior method in autoguide).

### What predictive likes in other frameworks

**Stan** has [generated quantities](https://mc-stan.org/docs/2_18/reference-manual/program-block-generated-quantities.html) block, which is used to
+ forward sampling to generate simulated data for model testing,
+ generating predictions for new data,
+ calculating posterior event probabilities, including multiple comparisons, sign tests, etc.,
+ calculating posterior expectations,
+ transforming parameters for reporting,
+ applying full Bayesian decision theory,
+ calculating log likelihoods, deviances, etc. for model comparison.

**PyMC3** has [deterministic nodes](https://docs.pymc.io/notebooks/api_quickstart.html#Deterministic-transforms) to keep track of transformed variables and [posterior_predictive](https://docs.pymc.io/notebooks/api_quickstart.html#4.-Posterior-Predictive-Sampling) to get values at observed nodes.

Typically, for prediction/forecasting, I imagine we will do something like
```
[...]

PR #354, opened 2019-09-25
Implement log likelihood and update baseball example

fehiepsi:
Addresses #349.

### TODO
- [x] add test for log_likelihood

### DIFF
--- a/examples/baseball.py
+++ b/examples/baseball.py
@@ -9,9 +9,9 @@ import jax.random as random
 from jax.scipy.special import logsumexp
 
 import numpyro
-from numpyro import handlers
 import numpyro.distributions as dist
 from numpyro.examples.datasets import BASEBALL, load_dataset
+from numpyro.infer_util import log_likelihood, predictive
 from numpyro.mcmc import MCMC, NUTS
 
 """
@@ -54,16 +54,6 @@ hyper-parameters) of running HMC on different problems.
 """
 
 
-# TODO: Remove broadcasting logic when support for `pyro.plate` is
-# available.
-#
-# Note that the current manual broadcasting logic is designed to
-# also work when the latent variables are batched along the leading
-# axis. This is useful for doing vectorized predictions, i.e. getting
-# the entire empirical distribution in one pass through the model.
-# This broadcasting logic can be removed when support for plates is
-# available.
-
 def fully_pooled(at_bats, hits=None):
     r"""
     Number of hits in $K$ at bats for each player has a Binomial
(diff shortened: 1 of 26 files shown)


examples/baseball.py
```diff
@@ -142,28 +132,20 @@ def run_inference(model, at_bats, hits, rng, args):
...
-            site_log_prob = site['fn'].log_prob(site['value'])
-            log_joint = log_joint + np.sum(site_log_prob.reshape(site_log_prob.shape[:1] + (-1,)),
-                                           -1)
-        log_post_density = logsumexp(log_joint) - np.log(np.shape(log_joint)[0])
-        print('\nPosterior log density: {:.2f}\n'.format(log_post_density))
+        post_loglik = log_likelihood(model, z, at_bats, hits)['obs']
+        # computes expected log predictive density at each data point
+        exp_log_density = logsumexp(post_loglik, axis=0) - np.log(np.shape(post_loglik)[0])
+        # reports log predictive density of all test points
+        print('\nLog pointwise predictive density: {:.2f}\n'.format(exp_log_density.sum()))
```
fehiepsi: @neerajprad I changed the log density formula which we need to report. Comparing joint_density of these models is not meaningful because each one has its own latent variables and priors. FYI, log pointwise predictive density can be found at https://arxiv.org/pdf/1307.5928.pdf formula (4).

neerajprad: Yes, this sounds correct to me! Thanks for fixing this, I think this was wrong in Pyro and just migrated from there.

examples/baseball.py
```diff
@@ -126,13 +117,12 @@ def partially_pooled_with_logit(at_bats, hits=None):
...
+    loc = numpyro.sample("loc", dist.Normal(-1, 1))
+    scale = numpyro.sample("scale", dist.HalfCauchy(1))
     num_players = at_bats.shape[0]
-    loc = numpyro.sample("loc", dist.Normal(np.array([-1.]), np.array([1.])))
-    scale = numpyro.sample("scale", dist.HalfCauchy(np.array([1.])))
-    shape = np.shape(loc)[:np.ndim(loc) - 1] + (num_players,)
-    alpha = numpyro.sample("alpha", dist.Normal(np.broadcast_to(loc, shape),
-                                                np.broadcast_to(scale, shape)))
-    return numpyro.sample("obs", dist.Binomial(at_bats, logits=alpha), obs=hits)
+    with numpyro.plate("num_players", num_players):
```
neerajprad: +1 on cleaning this up!

PR #356, opened 2019-09-25
Fix plate behavior for distributions with batch shape

### LINKED ISSUE #355: `plate` has unexpected behaviour when dist has batch_shape
I faced this issue while revising baseball example using `plate`. The repro code
```
import jax.numpy as np
import numpyro
from numpyro.handlers import trace, seed

def model():
    with numpyro.plate('plate', 10):
        numpyro.sample('x', dist.Normal(np.zeros(10), 1))

trace(seed(model, 0)).get_trace()['x']['value'].shape
```
should return `(10,)` but currently, it returns `(10, 10)`.

cc @neerajprad

neerajprad:
Fixes #355.

### DIFF
--- a/numpyro/primitives.py
+++ b/numpyro/primitives.py
@@ -240,9 +240,12 @@ class plate(Messenger):
             raise ValueError('Expected dimensions within plate = {}, which is less than the '
                              'distribution\'s batch shape = {}.'.format(len(expected_shape), len(dist_batch_shape)))
         trailing_shape = expected_shape[overlap_idx:]
-        if sum(map(sub, lax.broadcast_shapes(trailing_shape, dist_batch_shape), dist_batch_shape)) > 0:
-            raise ValueError('Distribution batch shape = {} cannot be broadcast to {}.'
-                             .format(trailing_shape, dist_batch_shape))
+        # e.g. distribution with batch shape (1, 5) cannot be broadcast to (5, 5)
+        broadcast_shape = lax.broadcast_shapes(trailing_shape, dist_batch_shape)
+        if sum(map(sub, broadcast_shape, dist_batch_shape)) > 0:
+            raise ValueError('Distribution batch shape = {} cannot be broadcast up to {}. '
+                             'Consider using unbatched distributions.'
+                             .format(dist_batch_shape, broadcast_shape))
         batch_shape = expected_shape[:overlap_idx]
         if 'sample_shape' in msg['kwargs']:
             batch_shape = lax.broadcast_shapes(msg['kwargs']['sample_shape'], batch_shape)
--- a/numpyro/primitives.py
+++ b/numpyro/primitives.py
@@ -241,7 +241,7 @@ class plate(Messenger):
         trailing_shape = expected_shape[overlap_idx:]
         # e.g. distribution with batch shape (1, 5) cannot be broadcast to (5, 5)
         broadcast_shape = lax.broadcast_shapes(trailing_shape, dist_batch_shape)
-        if sum(map(sub, broadcast_shape, dist_batch_shape)) > 0:
+        if broadcast_shape != dist_batch_shape:
             raise ValueError('Distribution batch shape = {} cannot be broadcast up to {}. '
                              'Consider using unbatched distributions.'
                              .format(dist_batch_shape, broadcast_shape))
--- a/numpyro/primitives.py
+++ b/numpyro/primitives.py
@@ -1,5 +1,6 @@
 from collections import namedtuple
 import functools
+from operator import sub
 
 import jax
 from jax import lax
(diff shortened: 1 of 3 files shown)


numpyro/primitives.py
```diff
@@ -232,7 +233,17 @@ def process_message(self, msg):
...
         cond_indep_stack.append(frame)
-        batch_shape = self._get_batch_shape(cond_indep_stack)
+        expected_shape = self._get_batch_shape(cond_indep_stack)
+        dist_batch_shape = msg['fn'].batch_shape if msg['type'] == 'sample' else ()
+        overlap_idx = len(expected_shape) - len(dist_batch_shape)
+        if overlap_idx < 0:
+            raise ValueError('Expected dimensions within plate = {}, which is less than the '
+                             'distribution\'s batch shape = {}.'.format(len(expected_shape), len(dist_batch_shape)))
+        trailing_shape = expected_shape[overlap_idx:]
+        if sum(map(sub, lax.broadcast_shapes(trailing_shape, dist_batch_shape), dist_batch_shape)) > 0:
```
fehiepsi: nit: can we just compare `lax.broadcast_shapes(trailing_shape, dist_batch_shape)` and `dist_batch_shape`?

numpyro/primitives.py
```diff
@@ -232,7 +233,17 @@ def process_message(self, msg):
...
+            raise ValueError('Distribution batch shape = {} cannot be broadcast to {}.'
+                             .format(trailing_shape, dist_batch_shape))
```
fehiepsi: nit: I guess you mean `"Distribution batch shape = {} cannot be broadcasted from the trailing plate shape {}.".format(dist_batch_shape, trailing_shape)`?

numpyro/primitives.py
fehiepsi: Hmm, this seems related to https://github.com/pyro-ppl/numpyro/issues/355#issuecomment-535246983: to deal with the case `trailing_shape=(2, 1, 5)`, `dist_batch_shape=(1, 3, 5)`. I think that it is better to add a note here to reflect your idea.

numpyro/primitives.py
neerajprad: I meant to rather than from, since this should fail precisely when we try to broadcast up an existing distribution batch dim. Let me add the detail about the trailing plate dim.

neerajprad: Sorry, arguments here are reversed, will fix.

numpyro/primitives.py
```diff
@@ -232,7 +233,20 @@ def process_message(self, msg):
...
+        expected_shape = self._get_batch_shape(cond_indep_stack)
+        dist_batch_shape = msg['fn'].batch_shape if msg['type'] == 'sample' else ()
+        overlap_idx = len(expected_shape) - len(dist_batch_shape)
+        if overlap_idx < 0:
+            raise ValueError('Expected dimensions within plate = {}, which is less than the '
+                             'distribution\'s batch shape = {}.'.format(len(expected_shape), len(dist_batch_shape)))
+        trailing_shape = expected_shape[overlap_idx:]
+        # e.g. distribution with batch shape (1, 5) cannot be broadcast to (5, 5)
+        broadcast_shape = lax.broadcast_shapes(trailing_shape, dist_batch_shape)
+        if sum(map(sub, broadcast_shape, dist_batch_shape)) > 0:
```
fehiepsi: As I understand, broadcast_shape[i] >= dist_batch_shape[i] for each index `i`. So this sum greater than 0 is equivalent to the equality of two shapes. I would prefer to compare `broadcast_shape == dist_batch_shape` directly.

fehiepsi:
@neerajprad Looking likes the way I used `plate` at [infer_test](https://github.com/pyro-ppl/numpyro/blob/master/test/test_infer_util.py#L20) is wrong. Could you help me correct it? I am not sure what is the best way to fix it. Would adding `with numpyro.plate("plate", 1000, dim=-2):` fix the issue?

neerajprad:
> Would adding with numpyro.plate("plate", 1000, dim=-2): fix the issue?

Yes, just fixing that.

numpyro/primitives.py
neerajprad: Note that broadcasting up in particular cannot be done, but the reverse is allowed:

```python
with plate('plate', 10, dim=-2):
  pyro.sample('b', dist.Bernoulli(np.array([0.5, 0.5]))) # batch shape = (2,) ; trailing plate = (1,)
```

This is what we are doing in the infer test below actually. Directly comparing will also fail one of the tests in `test_handlers`.

fehiepsi: But `trailing_shape=(1,)` will be broadcasted to `dist_batch_shape=(2,)` in that case, so we have
`broadcast_shape=(2,)`, `dist_batch_shape=(2,)`. Do I understand it correctly?

neerajprad: My bad, I got confused with my own broadcasting manipulations! 😄 You are right, we can directly compare and it will fail the other case where we are trying to broadcast up.

neerajprad:
@fehiepsi - I think all comments should be addressed. 'm merging this to unblock other PRs, let me know if there are any follow-up comments.

fehiepsi:
Thanks @neerajprad !

Issue #357, opened 2019-09-26
Declare dependent dimensions


fehiepsi:
In Pyro, we can declare dependent dimensions through `.to_event()`. Currently, this is useful for correctly computing log_prob of a transformed distribution where `transform.event_dim>0` but `base_dist.event_shape=()`. In particular, we need this to declare `base_dist` in autoguide.

In Pyro, this is achieved by using Independent class. It seems unnecessary to me. Can we just manipulate `batch_shape`, `event_shape` and correctly compute the returned value for `log_prob`? I think we just need 1 line of code for each distribution
```
def log_prob(...):
    ...
    # event_ndim minus 1 for multivariate distributions
    return sum_rightmost(log_prob, len(self.event_shape))
```

If Independent class turns out to be useful, then we will implement it instead.

neerajprad:
I think following the Pyro pattern here will be more modular. Adjusting the `batch_shape`, `event_shape` inside each distribution will lead to issues like `batch_shape` not matching the shape of distribution parameters which will necessitate all kinds of checks that are avoidable. Using `.to_event` will confine the change to a single class / method. I am happy to implement a version of this.

fehiepsi:
Sure, I didn't follow up with Distribution development from the start so I didn't appreciate much the current design decision. I imagine the method `to_event` can be implemented as
```
def to_event(self, n):
    self.event_shape = self.batch_shape[-n:] + self.event_shape
    self.batch_shape = self.batch_shape[:-n]
```
For some situations such as `loc.shape=(10, 1)`, `scale.shape=(1, 10)`, this still works.

While writing this reply, I do agree that using Independent will be beneficial for functions which utilizes this dependent information. It will take more effort to check how many dependent dims a distribution has (because some are univariate, some are multivariate).

neerajprad:
Np, it is always good to question existing design patterns! Regarding the solution above, it is not that it cannot be done, but this will leak into individual distribution classes (methods like `log_prob`; additionally, if any methods use batch shape in the constructor to do something like set some internal attribute, and we later change batch shape, we will need to also change what was done internally in the constructor). 

[...]

PR #359, opened 2019-09-26
Add mv affine transform

fehiepsi:
In preparation for having mvn guide, this PR adds multivariate affine transform.

This transform is useful to create transformed distribution with univariate base_dist, as long as we can declare dependent dimension (issue #357) for base_dist. In particular, `TransformedDistribution(Normal(0, 1).to_event(1), MultivariateAffineTransform(loc, scale_tril))` is equivalent to `MVN(loc, scale_tril)`.

### DIFF
--- a/numpyro/distributions/constraints.py
+++ b/numpyro/distributions/constraints.py
@@ -29,6 +29,7 @@ from jax.flatten_util import ravel_pytree
 from jax.lib.xla_bridge import canonicalize_dtype
 from jax.nn import softplus
 import jax.numpy as np
+from jax.scipy.linalg import solve_triangular
 from jax.scipy.special import expit, logit
 
 from numpyro.distributions.util import (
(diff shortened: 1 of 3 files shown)


numpyro/distributions/constraints.py
```diff
@@ -432,6 +433,39 @@ def log_abs_det_jacobian(self, x, y, intermediates=None):
...
+
+    def __init__(self, loc, scale_tril):
+        # TODO: relax this condition per user request
+        if np.ndim(scale_tril) != 2:
+            raise ValueError("Only support 2-dimensional scale_tril matrix.")
+        self.loc = loc
+        self.scale_tril = scale_tril
+
+    def __call__(self, x):
+        return self.loc + np.squeeze(np.matmul(self.scale_tril, x[..., np.newaxis]), axis=-1)
```
neerajprad: nit: why not just `self.scale_tril @ x`?

fehiepsi: `x` here is a vector, so we have to use matrix-vector product here. We can replace matmul by @ but we still need to unsqueeze and squeeze.

neerajprad: I think @ should support matrix-vector product:

```
In [3]: np.ones((4, 3)) @ np.ones(3)
Out[3]: DeviceArray([3., 3., 3., 3.], dtype=float32)
```

fehiepsi: But it will fail when `x` has batch dimension, so we can't compose this transform with distributions having non-trivial batch_shape.

neerajprad: Oh, now I understand. Just for my own notes, this will fail on:

```python
>>> np.ones((4, 3)) @ np.ones([2, 3])
TypeError: dot_general requires contracting dimensions to have the same shape, got [3] and [2].

>>> np.matmul(np.ones((4, 3)), np.ones([2, 3])[..., np.newaxis])
DeviceArray([[[3.],
              [3.],
              [3.],
              [3.]],

             [[3.],
              [3.],
              [3.],
              [3.]]], dtype=float32)

```

PR #360, opened 2019-09-27
Add to_event to Distribution to change batch-event dim boundary

### LINKED ISSUE #357: Declare dependent dimensions

neerajprad:
Fixes #357. 

I think we had a work-arounds in the code for the absence of this facility, that can be addressed as follow-up PRs once this merges.

### DIFF
--- a/numpyro/distributions/__init__.py
+++ b/numpyro/distributions/__init__.py
@@ -36,7 +36,7 @@ from numpyro.distributions.discrete import (
     Poisson,
     PRNGIdentity
 )
-from numpyro.distributions.distribution import Distribution, TransformedDistribution
+from numpyro.distributions.distribution import Distribution, Independent, TransformedDistribution
 
 __all__ = [
     'Bernoulli',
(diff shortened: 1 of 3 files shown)


fehiepsi:
This is great! This passed all the checkboxes in my head. Happy to merge!

PR #361, opened 2019-09-27
implement auto mvn guide

fehiepsi:
This PR
+ adds MVN guide
+ move median/quantiles methods to AutoContinuous (as in Pyro)
+ clean up some legacy docs and _Normal.


numpyro/contrib/autoguide/__init__.py
```diff
@@ -247,27 +248,8 @@ def sample_posterior(self, rng, params, sample_shape=()):
...
-    def _get_transform(self):
-        loc, scale = self._loc_scale()
-        return AffineTransform(loc, scale, domain=constraints.real_vector)
-
     def _loc_scale(self):
-        loc = numpyro.param('{}_loc'.format(self.prefix), self._init_latent)
-        scale = numpyro.param('{}_scale'.format(self.prefix), np.ones(self.latent_size),
-                              constraint=constraints.positive)
-        return loc, scale
+        raise NotImplementedError
```
neerajprad: Can we remove this from AutoContinuous? I don't think it makes sense to have this for a continuous autoguide in general. If subclasses need to abstract some internal computation within `_loc_sacle`, they are free to do so.

fehiepsi: Hmm, `median` and `quantiles` from `continuous` require this function. Should we move them out also?

> I don't think it makes sense to have this for a continuous autoguide in general

Agreed! It does not make sense for flow style autoguide. I think we'd better remove `loc_scale` and implement `median`, `quantiles` directly for Normal, MVN guide. FYI LaplaceApproximation does not require `loc_scale` to compute `median` (only `loc` is enough, so no need to compute hessian... for `scale`). WDYT?

neerajprad: > I think we'd better remove loc_scale and implement median, quantiles directly for Normal, MVN guide.

Yes, I think it sounds reasonable to remove loc_scale and mark median, quantiles etc. as NotImplemented in AutoContinuous. They can be implemented by the subclasses as needed. Even if there's a small amount of code duplication, its fine.

fehiepsi: >  I think it sounds reasonable to remove loc_scale and mark median, quantiles etc. as NotImplemented in AutoContinuous. They can be implemented by the subclasses as needed. Even if there's a small amount of code duplication, its fine.

Yeah, I will go with this.

fehiepsi: Done

PR #363, opened 2019-09-27
Implement laplace approximation

fehiepsi:
Laplace Approximation and LowRankMVN have different behavior to the remaining continuous guides:
+ LA optimizes MAP point during inference. At the end of inference, we compute hessian of the loss w.r.t. MAP point and obtain the transform.
+ LowRankMVN optimizes low rank form during inference. At the end of inference, we convert low rank form to full rank form and obtain the transform.

Currently, to get transform, we `substitute` params to `_get_transform` method. Then we use that transform to train the guide. This won't work for LA & LowRank because we only know transform at the end of the SVI loop (i.e. when we have access to `params`). So I guess we should rewrite public methods for these two classes.

---

LA in NumPyro will be a bit different from LA in Pyro:
+ In Pyro, we keep parameters in global store, so after getting MAP point, we call a method `laplace_approximation` to register new params `loc`, `scale_tril` to global store.
+ In NumPyro, all we need is the transform and it can be obtained by using `get_transform()` method.
So the API will be a bit different:
+ In Pyro
```
delta_guide = AutoLaplaceApproximation(model)
svi = SVI(model, delta_guide, ...)
... # run svi
mvn_guide = delta_guide.laplace_approximation()
posterior_dist = mvn_guide.get_posterior()
unconstrained_samples = posterior_dist.sample(sample_shape)
# then constrain samples
```
+ In NumPyro
```
guide = AutoLaplaceApproximation(model)
svi = SVI(model, guide, ...)
... # run svi and get params
transform = guide.get_transform(params)  # this does laplace approximation
samples = guide.sample_posterior(rng, params, sample_shape)
```

---
### TODO
- [x] compute hessian from elbo loss

Blocked by #361.


numpyro/contrib/autoguide/__init__.py
```diff
@@ -21,8 +21,9 @@
     UnpackTransform,
     biject_to
 )
-from numpyro.distributions.util import sum_rightmost
+from numpyro.distributions.util import cholesky_inverse, sum_rightmost
 from numpyro.infer_util import constrain_fn, find_valid_initial_params, init_to_median, transform_fn
+import numpyro.infer.svi
```
fehiepsi: The second time I faced a circular import issue in NumPyro: we call `elbo` which belongs to `svi.py` file, in that file, we call `AutoContinuous`, which is this file. >"<

neerajprad: Yeah ideally the SVI class shouldn't need to know about AutoContinuous, but I don't have a better solution to that issue.

neerajprad: My suggestion would be to do the following:
 - Refactor ELBO to make it a class.
 - Subclass ELBO in Autoguide to use `base_param_map`, we can call this method `Autoguide_ELBO`, or something.

That way we can inherit any other functionality from the base ELBO class later. That way, also all contrib code also remains in contrib and isn't imported by core NumPyro.

fehiepsi: Agreed! We should distinguish the name when the meaning of ELBO in the usual guide and the auto guide are different (one is in constrained space, the other is in unconstrained space). I remember that you already suggested this previously.

neerajprad:
Looks great!

Issue #362, opened 2019-09-27
Optimize test suite to run faster on Travis


fehiepsi:
TODOs:
 - [ ] Investigate why individual distribution tests take in the order of seconds. Maybe its just the JIT compilation of the gamma sampler? In which case, maybe we should disable JIT in some of these tests.
Du: JIT compilation for gamma sampler is slow, but disable it would be much slower.
 - [x] Remove duplicate MCMC tests for the older functional api. Only a few can be tested.
 - [x] Clean up distributions contrib tests. 
 - [x] Make test_initialize_model in MCMC faster
 - [x] Test only a subset of models with double precision.

neerajprad:
I think now is also the time when it might be worth going through the test suite and seeing what tests take up most of the time. It is usually the case that there are only a few tests that take up a majority of the time, and we can explore if that can be optimized. Some of the distribution tests took longer than 1s to generate samples (mostly to do with the gamma sampler) earlier, but I am not sure if that problem is still as bad. 

We also have other new repositories coming up which need to share the same set of travis instances, so it will be good to be conservative in how many instances a single project uses.

fehiepsi:
> It is usually the case that there are only a few tests that take up a majority of the time, and we can explore if that can be optimized.

let me do quick profiling for this

> We also have other new repositories coming up which need to share the same set of travis instances

I see, let's see what profiling result is first.

fehiepsi:
### test_distribution: 310s
```
26.90s call     test/test_distributions.py::test_mean_var[LKJCholesky-None-params26]                      [50/8123]
15.35s call     test/test_distributions.py::test_mean_var[LKJCholesky-None-params25]                               
10.94s call     test/test_distributions.py::test_mean_var[LKJCholesky-None-params28]                               
10.38s call     test/test_distributions.py::test_mean_var[LKJCholesky-None-params27]                               
8.44s call     test/test_distributions.py::test_pathwise_gradient[Gamma-sp_dist0-params0]                          
7.12s call     test/test_distributions.py::test_pathwise_gradient[Beta-sp_dist3-params3]                           
5.75s call     test/test_distributions.py::test_mean_var[Beta-<lambda>-params2]
5.04s call     test/test_distributions.py::test_dist_shape[prepend_shape0-LKJCholesky-None-params26]
4.62s call     test/test_distributions.py::test_dist_shape[prepend_shape1-LKJCholesky-None-params27]
[...]

PR #365, opened 2019-09-28
Improving testing time 

fehiepsi:
Addresses #362 

### Done
- [x] contrib/test_distributions: 60s -> 0s
- [x] other contrib: 110s -> 55s
- [x] revise test_infer_util + distributions_util, remove tests for standard_gamma in util (this is tested in JAX and in test_distributions): 30s -> 0
- [x] test_distribution: relax mean_var LKJ tests 310s -> 230s
- [x] test_hmc_util 217s -> 67s

### Future PRs
@neerajprad  could you make an attempt at those tests first for which one we need to keep and which one we need to remove? I think we don't need to (e.g. x64 for all tests in test_mcmc) but there are some other strategies to do: e.g. mark skip if in Travis,...
- test_mcmc
- test_mcmc_interface


neerajprad:
> @neerajprad could you make an attempt at those tests first for which one we need to keep and which one we need to remove? I think we don't need to (e.g. x64 for all tests in test_mcmc) but there are some other strategies to do: e.g. mark skip if in Travis,...

I think we can simply remove many of those tests which are actually just duplicates.

test/contrib/test_autoguide.py
```diff
@@ -15,10 +15,14 @@
 from numpyro.distributions import constraints
 from numpyro.distributions.flows import InverseAutoregressiveTransform
 from numpyro.handlers import substitute
+from numpyro.infer_util import init_to_median
 from numpyro.svi import SVI, elbo
 from numpyro.util import fori_loop
 
 
+init_strategy = partial(init_to_median, num_samples=2)
```
fehiepsi: This reduces the time spent to find initial_params a bit. It took a few seconds to call `trace(model).get_trace(*args, **kwargs)` the first time.

test/contrib/test_nn.py
```diff
@@ -97,3 +99,18 @@ def test_masks(input_dim, n_layers, output_dim_multiplier):
                 if mask_skip[idx + jdx * input_dim, kdx]:
                     skip_connections.add(kdx)
             assert_array_equal(list(sorted(skip_connections)), correct)
+
+
+@pytest.mark.parametrize('input_dim', [5, 7])
+def test_masked_dense(input_dim):
```
fehiepsi: I moved the content of `test_masked_dense` to here.

test/contrib/test_nn.py
```diff
@@ -51,9 +53,9 @@ def test_auto_reg_nn(input_dim, hidden_dims, param_dims, skip_connections):
     assert onp.sum(onp.abs(onp.triu(permuted_jac))) == 0.0
 
 
-@pytest.mark.parametrize('input_dim', [5, 8])
-@pytest.mark.parametrize('n_layers', [1, 3])
-@pytest.mark.parametrize('output_dim_multiplier', [1, 4])
+@pytest.mark.parametrize('input_dim', [2, 6])
```
fehiepsi: This test uses many for-loops, which is slow. So I reduced those numbers a bit.

test/test_distributions.py
```diff
@@ -472,7 +475,7 @@ def fn(*args):
 
 @pytest.mark.parametrize('jax_dist, sp_dist, params', CONTINUOUS + DISCRETE)
 def test_mean_var(jax_dist, sp_dist, params):
-    n = 200000
+    n = 20000 if jax_dist is dist.LKJCholesky else 200000
```
fehiepsi: we don't need too many LKJ samples to get good precision.

test/test_distributions_util.py
```diff
@@ -123,60 +122,6 @@ def test_fn(x):
     assert_allclose(jacobian(cumprod)(x), jacobian(test_fn)(x), atol=1e-7)
 
 
-@pytest.mark.parametrize('alpha, shape', [
-    (1., ()),
-    (1., (2,)),
-    (np.array([1., 2.]), ()),
-    (np.array([1., 2.]), (3, 2)),
-])
-def test_standard_gamma_shape(alpha, shape):
```
fehiepsi: those tests are covered in JAX and test_distributions.py file.

test/test_hmc_util.py
```diff
@@ -454,10 +458,10 @@ def model(data):
         5,  14,  13,  22,
     ])
 
-    rngs = random.split(random.PRNGKey(1), 12)
+    rngs = random.split(random.PRNGKey(1), 2)
```
fehiepsi: it is unnecessary to test the correctness of batched `rngs` with batch=12. I think 2 is enough.

fehiepsi:
@neerajprad Testing time is reduced from 38m to 23m. I am satisfied with the result now so would like to do further tuning when total time goes over 30m again.

neerajprad:
> 19m with this PR.

That's great! Yes, it is best to keep revisiting this periodically.

PR #366, opened 2019-09-28
Add documentation for Independent

neerajprad:
Adds Independent to doc and addresses todo in autoguide.

### DIFF
--- a/docs/source/distributions.rst
+++ b/docs/source/distributions.rst
@@ -9,6 +9,14 @@ Distribution
     :show-inheritance:
     :member-order: bysource
 
+Independent
+-----------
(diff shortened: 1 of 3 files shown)


neerajprad:
> In the last PR, you mentioned that you are going to update distribution classes. 

I don't think we need to update any distribution classes, I probably meant updating the parts of the code which were using work-arounds in the absence of Independent, and the only place I found was autoguide. Are there other places that you can think of?

> Is it related to the current behaviour that we don't broadcast mean, variance in Independent?

Do we need to broadcast means, I think delegating to `base_dist.mean` should be sufficient since this won't add any batch dimensions?

fehiepsi:
Oppss... my bad, you are right!

Issue #243, opened 2019-07-13
Support GramianTransform


fehiepsi:
This is a transform from A -> A @ A.T, which if
+ composed with LowerCholeskyTransform will create PositiveDefiniteTransform, which transforms a real vector to a positive definite matrix (though I am not interested in Wishart/Inverse-Wishart distributions, this transform will be helpful for other contributors who are interested in implementing those distributions)
+ composed with CorrCholeskyTransform will create CorrelationTransform, which transforms a real vector to a correlation matrix.

Then we can define LKJ distribution as TransformedDistribution(LKJCholesky, GramianTransform), so that users don't have to work with the scale tril version of a correlation matrix. It is better to for users see directly correlations of variables, rather than getting Cholesky and manually apply Grammian operator.

Inference will happen in LKJCholesky support when the approach to transformed distribution in #241 is available. Then it would be fast and not suffered from Cholesky numerical issue because no Cholesky operator is involved when computing log_prob terms.

PR #368, opened 2019-09-28
Implement inverse Cholesky transform

fehiepsi:
Addresses #243. This PR adds InvCholeskyTransform, which creates Gramian matrix: y = x @ x.T.

Some consequences of this transform are:
+ bijectors for correlation matrices, positive_definite matrices: `biject_to(constraints.corr_matrix)`, `biject_to(constraints.positive_matrix)`
+ LKJ distribution, which is the transformed distribution of LKJCholesky and the above InvCholeskyTransform.
(and if users want to implement Wishart/InverseWishart distribution over positive definite matrices, the positive_definite bijector `biject_to(constraints.positive_matrix)` would be helpful for them: it allows them to run MCMC over those priors)


numpyro/distributions/constraints.py
```diff
@@ -412,10 +424,47 @@ def log_abs_det_jacobian(self, x, y, intermediates=None):
         return np.full(np.shape(x) if self.event_dim == 0 else np.shape(x)[:-1], 0.)
 
 
+class InvCholeskyTransform(Transform):
```
fehiepsi: @neerajprad any suggestion for a better name? ^^

numpyro/distributions/constraints.py
```diff
@@ -412,10 +424,47 @@ def log_abs_det_jacobian(self, x, y, intermediates=None):
...
+            n = np.shape(x)[-1]
+            order = np.arange(n - 1, -1, -1)
+            return np.sum(order * np.log(np.diagonal(x, axis1=-2, axis2=-1)), axis=-1)
+
+
 class LowerCholeskyTransform(Transform):
     domain = real_vector
     codomain = lower_cholesky
-    event_dim = 1
+    event_dim = 2
```
fehiepsi: I changed `event_dim` to 2 to:
+ match Pyro implementation
+ make log_abs_det_transform of ComposeTransform([LowerCholesky, InvCholesky]) correct: we need to keep both transforms have the same event_dim
+ make [event_shape](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/distributions/distribution.py#L272) of TransformedDistribution(dist.Normal(np.zeros(3), 1), LowerCholeskyTransform) correct.

Side effect:
+ ComposedTransformed([AffineTransform(domain=real_vector), LowerCholeskyTransform]) will have wrong log_det due to [this line](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/distributions/constraints.py#L288) (the former part has event_dim=1, while the later part has event_dim)

To fix the above site effect, we can set `event_dim=1` for both LowerCholeskyTransform and InvCholeskyTransform. In that case, shape sematic of TransformedDistribution(dist.Normal(np.zeros(3), 1), LowerCholeskyTransform) will be wrong. However, I think that no user wants to construct those kinds of ComposeTransform, TransformedDistribution so either way (setting `event_dim=1` or `event_dim=2` is fine to me). Personally, I prefer to set `event_dim=1` because it is helpful to think the codomain lower_cholesky as a vector, rather than a matrix (when computing jacobian), but I don't want to break the convention in Pyro/PyTorch or create confusion for users. So I set `event_dim=2` here.

numpyro/distributions/constraints.py
neerajprad: I think this name is only confusing because of the presence of `LowerCholeskyTransform` which probably should have been named `LowerTrilTransform` or something.

numpyro/distributions/constraints.py
neerajprad: I think this is fine. I will need to think about it more to see if there is a more general way to conceptualize event dimensions in transforms, but let us spend time on this when we have more uses in hand.

PR #369, opened 2019-09-29
Expose energy for bfmi diagnostic

### LINKED ISSUE #349: Higher level interface for predictions and arviz integration

fehiepsi:
Resolves #349.

This PR exposes energy, which is required for [bfmi diagnostic](https://mc-stan.org/misc/warnings.html#bfmi-low), and [energy plot](https://nbviewer.jupyter.org/gist/fehiepsi/f65519d34526cce4f14b381c6a9a98f8#Energy-Plot). Because this is just a scalar, I think that storing this value to TreeInfo won't affect the performance.

@neerajprad This [gist](https://nbviewer.jupyter.org/gist/fehiepsi/f65519d34526cce4f14b381c6a9a98f8) illustrates how numpyro and arviz are integrated. With this PR, we fully support arviz (I need to make a PR to arviz with those updates after 0.2.1 is released)! Cheers, :D

I think it is good to make a tutorial which shows how simple it is to run MCMC in numpyro, and showcase the support for arviz. WDYT?

### DIFF
--- a/numpyro/hmc_util.py
+++ b/numpyro/hmc_util.py
@@ -22,7 +22,7 @@ IntegratorState = namedtuple('IntegratorState', ['z', 'r', 'potential_energy', '
 
 TreeInfo = namedtuple('TreeInfo', ['z_left', 'r_left', 'z_left_grad',
                                    'z_right', 'r_right', 'z_right_grad',
-                                   'z_proposal', 'z_proposal_pe', 'z_proposal_grad',
+                                   'z_proposal', 'z_proposal_pe', 'z_proposal_grad', 'z_proposal_energy',
                                    'depth', 'weight', 'r_sum', 'turning', 'diverging',
                                    'sum_accept_probs', 'num_proposals'])
 
@@ -506,10 +506,10 @@ def _combine_tree(current_tree, new_tree, inverse_mass_matrix, going_right, rng,
         turning = current_tree.turning
 
     transition = random.bernoulli(rng, transition_prob)
-    z_proposal, z_proposal_pe, z_proposal_grad = cond(
+    z_proposal, z_proposal_pe, z_proposal_grad, z_proposal_energy = cond(
         transition,
-        new_tree, lambda tree: (tree.z_proposal, tree.z_proposal_pe, tree.z_proposal_grad),
-        current_tree, lambda tree: (tree.z_proposal, tree.z_proposal_pe, tree.z_proposal_grad)
+        new_tree, lambda tree: (tree.z_proposal, tree.z_proposal_pe, tree.z_proposal_grad, tree.z_proposal_energy),
+        current_tree, lambda tree: (tree.z_proposal, tree.z_proposal_pe, tree.z_proposal_grad, tree.z_proposal_energy)
     )
 
     tree_depth = current_tree.depth + 1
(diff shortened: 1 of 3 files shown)


neerajprad:
> @neerajprad This gist illustrates how numpyro and arviz are integrated. With this PR, we fully support arviz (I need to make a PR to arviz with those updates after 0.2.1 is released)! Cheers, :D

That's great! @null-a - I think we should have these utilities available within the next release for numpyro and arviz. 

I think it is good to make a tutorial which shows how simple it is to run MCMC in numpyro, and showcase the support for arviz. WDYT?

Sure! We can modify the existing examples too. I think one good example is the existing one on non-centered parametrization which can be converted into a tutorial that demonstrates this integration and shows how to use diagnostic information to detect issues with the model.

Issue #364, opened 2019-09-28
Run MCMC with multiple chains without modifying any environment flags


neerajprad:
Users would like to be able to run MCMC without having to set the `--xla_force_host_platform_device_count` flag in advance. In certain cases, it is not possible to do this, e.g. in a repl or notebook environment. 

cc. @null-a who will need this for brmp.

Related JAX issue - https://github.com/google/jax/issues/1408.

fehiepsi:
As a temporary solution, I think we can do this automatically when users import `numpyro` (by setting `os.environ` as in our examples).

neerajprad:
Hmm..I was under the impression that we had to set the environment flag before loading JAX, but seems like this should work even after the fact by setting this through `os.environ`. Is that true? In that case, we can just set/unset it as needed.

fehiepsi:
In addition, I also think that it is good to use `cpu` by default. So my proposal is:
1. set cpu + max device_count by default
2. make some utilities: `numpyro.util.set_backend('gpu')` to change backend, `numpyro.util.set_num_devices(...)` to change the number of required devices.

fehiepsi:
> Is that true? In that case, we can just set/unset it as needed.

yes, it is true

fehiepsi:
I am not sure what intra-op is useful for, because I didn't observe much different in covtype data (in cpu).

neerajprad:
I think it will be safer to retain the XLA defaults since that's what would be most rigorously tested by the XLA and JAX devs. My concern was that this wasn't even possible to do once we imported JAX. But if it is, we can just do this in the MCMC class itself without changing anything else.

neerajprad:
Oh sorry, I missed this "The only requirement is we set it at the top of the script (i.e. no np.ones(3) is called until we set these flags". That we certainly can't guarantee.

fehiepsi:
Even if we don't change the defaults, having those utilities is helpful (rather than having to remember the name of those flags - I find myself always go to examples folder to copy those flags to notebooks :D).

neerajprad:
I agree we should have those utilities. I'm also fine changing the default to CPU and setting num devices (though hopefully the num device argument won't carry over when we change the backend to GPU) to CPU count, but it will be good to confirm with the JAX devs just so that we don't end up with a default config that causes other hard-to-debug issues.

PR #370, opened 2019-09-29
make some utils to set JAX flags

fehiepsi:
Addresses #364. I have only changed the default platform to `cpu` (instead of `gpu`). I think it is better to not change the default cpu count = 1. If users want to run 4 chains parallelly in CPU, they should use `numpyro.util.set_host_devices(4)`.

Some facts which I collected:
+ By default, JAX uses `gpu`.
+ If we set XLA_FLAGS host device_count to a number, then CPU platform will be used by default.
+ We can change platform multi times using `jax.config.update('jax_platform_name', platform)`. The last one will take effect as long as no `np.ones(3)` is called.
+ Changing XLA_FLAGS host device_count will not affect the number of GPU device counts. No matter which value host (CPU) device_count takes, if we change platform to `gpu`, then `xla_bridge.device_count()` will list the number of GPUs which we have.

So `set_host_devices` only affects the number of CPUs available to us.

TODO (after this PR):
+ update examples to use these utils

### DIFF
--- a/numpyro/util.py
+++ b/numpyro/util.py
@@ -58,8 +58,7 @@ def set_host_devices(n):
     :param int n: number of CPU devices to use.
     """
     xla_flags = os.getenv('XLA_FLAGS', '').lstrip('--')
-    xla_flags = re.sub(re.compile(r'xla_force_host_platform_device_count=.+\s'),
-                       '', xla_flags).split()
+    xla_flags = re.sub(r'xla_force_host_platform_device_count=.+\s', '', xla_flags).split()
     os.environ['XLA_FLAGS'] = ' '.join(['--xla_force_host_platform_device_count={}'.format(n)]
                                        + xla_flags)
 
--- a/numpyro/util.py
+++ b/numpyro/util.py
@@ -23,7 +23,10 @@ def set_rng_seed(rng_seed):
 
 def set_platform(platform=None):
     """
-    :param str platform: either 'cpu', 'gpu', 'tpu'.
+    Changes platform to CPU, GPU, or TPU. This utility only takes
+    effect at the beginning of your program.
+
+    :param str platform: either 'cpu', 'gpu', or 'tpu'.
     """
     if platform is None:
         platform = os.getenv('JAX_PLATFORM_NAME', 'cpu')
@@ -32,6 +35,25 @@ def set_platform(platform=None):
 
 def set_host_devices(n):
     """
+    By default, XLA considers all CPU cores as one device. This utility tells XLA
+    that there are `n` host (CPU) devices available to use. As a consequence, this
+    allows parallel mapping in JAX :func:`jax.pmap` to work in CPU platform.
+
+    .. note:: This utility only takes effect at the beginning of your program.
+        Under the sence, this sets the environment variable
+        `XLA_FLAGS=--xla_force_host_platform_device_count=[num_devices]`, where
+        `[num_device]` is the desired number of CPU devices `n`.
+
+    .. warning:: We do not understand much the side effects when using
+        `xla_force_host_platform_device_count` flag. If you observe some strange
+        phenomenon when using this utility, please let us know through our issue
+        or forum page. Here we quote from XLA source code the meaning of this flag:
+        "Force the host platform to pretend that there are these many host
+        'devices'. All of these host devices are backed by the same threadpool.
+        Setting this to anything other than 1 can increase overhead from context
+        switching but we let the user override this behavior to help run tests
+        on the host that run models in parallel across multiple devices."
+
     :param int n: number of CPU devices to use.
     """
     xla_flags = os.getenv('XLA_FLAGS', '--xla_force_host_platform_device_count={}'.format(n))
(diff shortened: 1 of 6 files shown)


fehiepsi:
@mattjj We totally rely on `XLA_FLAGS=--xla_force_host_platform_device_count=...' to run chains parallelly in CPU. But we are not sure about its side effect. Previously you suggested to also change `--xla_cpu_multi_thread_eigen=False` and `--intra_op_parallelism_threads=1`, but I have been using only `xla_force_host_platform_device_count=12` for a few months without recognizing any problem. So I wonder if those flags are necessary. In addition, recently I observed that I can set `XLA_FLAGS=--xla_force_host_platform_device_count=100`, and `pmap(f)(np.arange(100))` still works. So I believe that those CPU devices are different from CPU cores. This makes me confused. Could you suggest us what is the best practice we should do?

neerajprad:
I think you might just need to loosen some of the thresholds on the failing tests, though I am not sure why it is failing on this PR.

numpyro/util.py
```diff
@@ -19,6 +21,23 @@ def set_rng_seed(rng_seed):
...
+    :param str platform: either 'cpu', 'gpu', 'tpu'.
+    """
+    if platform is None:
+        platform = os.getenv('JAX_PLATFORM_NAME', 'cpu')
+    jax.config.update('jax_platform_name', platform)
+
+
+def set_host_devices(n):
+    """
+    :param int n: number of CPU devices to use.
```
neerajprad: You mentioned that setting this changes the default to platform to 'cpu'. Will be good to add that as a warning in the docstring, and also expose these in the docs.

neerajprad: Also what happens when we set this to some number which doesn't match the number of actual CPUs we have?

neerajprad: Depending on your answer to 1, it might also be worth having a default option based on `os.cpu_count()`.

fehiepsi: > Will be good to add that as a warning in the docstring

It is not necessary if we set `cpu` as default platform.

> what happens when we set this to some number which doesn't match the number of actual CPUs we have?

I have no idea about this. It just works. You can set it to 1000 even that you just have 10-core CPUs. Because I don't understand it, I don't want to set those as default.

numpyro/util.py
```diff
@@ -19,6 +21,45 @@ def set_rng_seed(rng_seed):
...
+        "Force the host platform to pretend that there are these many host
+        'devices'. All of these host devices are backed by the same threadpool.
+        Setting this to anything other than 1 can increase overhead from context
+        switching but we let the user override this behavior to help run tests
+        on the host that run models in parallel across multiple devices."
+
+    :param int n: number of CPU devices to use.
+    """
+    xla_flags = os.getenv('XLA_FLAGS', '--xla_force_host_platform_device_count={}'.format(n))
+    os.environ['XLA_FLAGS'] = xla_flags
```
neerajprad: Sorry, I didn't notice this earlier - this will overwrite `XLA_FLAGS` in the environment, e.g. if there are other options to disable multithreading, fast math mode etc. We should check that and add to it instead. 

```python
xla_flags = os.getenv('XLA_FLAGS, '').split()
os.environ['XLA_FLAGS'] = ' '.join(['--xla_force_host_platform_device_count={}'.format(n)] + xla_flags)
```

numpyro/util.py
```diff
@@ -19,6 +21,45 @@ def set_rng_seed(rng_seed):
...
+
+
+def set_host_devices(n):
+    """
+    By default, XLA considers all CPU cores as one device. This utility tells XLA
+    that there are `n` host (CPU) devices available to use. As a consequence, this
+    allows parallel mapping in JAX :func:`jax.pmap` to work in CPU platform.
+
+    .. note:: This utility only takes effect at the beginning of your program.
+        Under the sence, this sets the environment variable
```
neerajprad: nit: maybe you meant "under the hood"?

numpyro/util.py
neerajprad: > You can set it to 1000 even that you just have 10-core CPUs.

Well, that's strange for sure.

numpyro/util.py
fehiepsi: This won’t take effect if XLA flags available in the environment but your solution is better. I think we also need to check if this flag is available to override it.

numpyro/util.py
fehiepsi: Yup

numpyro/util.py
neerajprad: Yes, this won't actually take an effect at all. Maybe something like the following:

```python
import re 

xla_flags = os.getenv('XLA_FLAGS, '')
xla_flags = re.sub('--xla_force_host_platform_device_count=.+\s', '', xla_flags).split()
os.environ['XLA_FLAGS'] = ...
```

fehiepsi: Thank @neerajprad , it works like a charm! :D I just make a little adjusting regarding `--` only allowed to be at the beginning of the flags. I have checked locally that XLA accepts this pattern with multi flags.

neerajprad: That looks right!

PR #375, opened 2019-10-01
Use device utils in examples/tutorials

### LINKED ISSUE #364: Run MCMC with multiple chains without modifying any environment flags

fehiepsi:
Fixes #364


examples/gp.py
```diff
@@ -132,8 +129,7 @@ def main(args):
     parser.add_argument("--device", default='cpu', type=str, help='use "cpu" or "gpu".')
     args = parser.parse_args()
 
-    if args.device == 'cpu' and args.num_chains <= os.cpu_count():
-        os.environ['XLA_FLAGS'] = '--xla_force_host_platform_device_count={}'.format(
-            args.num_chains)
+    numpyro.util.set_platform(args.device)
+    numpyro.util.set_host_devices(args.num_chains)
```
neerajprad: If we set the platform as GPU in the first line, does the second line interfere with that? (IIRC you mentioned that it did.)

fehiepsi: Good question! The second line just changes the default behavior (which is `platform=''` in JAX) from GPU to CPU. If the platform has been set to `platform='gpu'`, the second line will take no effect.

neerajprad: I see..we should also consider naming `set_host_devices` to `set_host_device_count`.

PR #376, opened 2019-10-01
improve binom/multinomial sampling

fehiepsi:
For the following code
```
from numpyro.distributions.util import binomial
from jax.random import PRNGKey
import jax.numpy as np

%timeit binomial(PRNGKey(0), 0.5, np.array(10), shape=(1000,)).copy()
```
+ before this PR, it returns 129 ms
+ with this PR, it returns 1.04 ms

In addition, I fixed a bug when `n` is an `int`: Previously, `binomial(PRNGKey(0), 0.5, 10, shape=(1000,))` will trigger the error `AttributeError: 'int' object has no attribute 'copy'` when we tried to apply `device_get` to an `int` (not np.ndarray) number.

### DIFF
--- a/numpyro/distributions/util.py
+++ b/numpyro/distributions/util.py
@@ -3,7 +3,7 @@ import math
 
 import scipy.special as osp_special
 
-from jax import custom_transforms, defjvp, jit, lax, random, vmap, device_get
+from jax import custom_transforms, defjvp, jit, lax, random, vmap
 from jax.lib.xla_bridge import canonicalize_dtype
 import jax.numpy as np
 from jax.numpy.lax_numpy import _promote_args_like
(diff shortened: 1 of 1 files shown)


numpyro/distributions/util.py
```diff
@@ -1,24 +1,20 @@
...
 from jax.scipy.linalg import solve_triangular
 from jax.util import partial
 
 
-# TODO: inefficient implementation; jit currently fails due to
-# dynamic size of random.uniform.
-@partial(jit, static_argnums=(2, 3))
-def _binomial(key, p, n, shape):
+@partial(jit, static_argnums=(3, 4))
+def _binomial(key, p, n, n_max, shape):
```
neerajprad: Good idea to jit n_max instead of n!

fehiepsi: Yes, it helps a lot! But I still can't find a way to make this works for a traced array `n`. I believe that we should revise the algorithm a bit for that purpose. But it is not worth to think about this now.

Issue #371, opened 2019-09-30
expose submodules to the main namespace


fehiepsi:
NumPyro is lightweight, so submodules live in their own files. Currently, we don't expose those submodules to `__init__`, so only a few submodules are exposed to the main namespace `numpyro`. This in turn leads to a lot of import statements for a numpyro program.

I would propose to expose all submodules to the main namespace for convenience. 

TODOs:
 - [x] Separate out `transforms` into its own module `numpryo.distributions.transforms`. This will align it with Pyro and help keep the generic interface alike.
 - [x] Move SVI and MCMC into `infer` module but make all inference algorithms available from the `infer` namespace.
 - [x] Add infer_util
 - [x] Add diagnostics

neerajprad:
@fehiepsi - I have added some subtasks here. I think it will help to keep distributions, inference algorithms and handlers in their own namespace so as to avoid name collisions. e.g. it seems reasonable to be able to do:
```
from numpyro import distributions as dist, infer, handlers

dist.Normal(..
infer.NUTS(..
handlers.replay(..
```

rather than make each of these classes available for import directly from `numpyro`. WDYT? The above structure also matches the generic API.

fehiepsi:
Sure, that would be very helpful! My proposal is to add
```
import numpyro.infer_util
import numpyro.util
import numpyro.diagnostics
```
to `__init__.py` file so that when I type `numpyro.`, the IDE (& autocompletion) will display out these submodules (together with their docstrings) for me to choose. Currently, typing `numpyro.` only displays some primitives, `patch`, `version`, and `distributions`.

PR #379, opened 2019-10-03
Refactor namespace for generic compatibility

neerajprad:
Addresses #371.


numpyro/__init__.py
```diff
@@ -1,6 +1,21 @@
 import numpyro.patch  # noqa: F401
-from numpyro.primitives import module, param, plate, sample  # noqa: F401
-from numpyro.version import __version__  # noqa: F401
-import numpyro.util as util
+from numpyro import compat, diagnostics, distributions, infer, infer_util, util
```
fehiepsi: can we add `handlers` here?

numpyro/__init__.py
```diff
@@ -1,6 +1,21 @@
...
 util.set_platform('cpu')
+
+
+__all__ = [
+    '__version__',
+    'compat',
+    'diagnostics',
+    'distributions',
+    'infer',
+    'infer_util',
```
fehiepsi: should we move `infer_util` (and rename it to `util`) and `hmc_util` to `infer`?

neerajprad: It is nice to be able to distinguish between utilities for setting random number generator seed and control flow primitives, from inference utilities for prediction and log density computation. I don't think most users of the library should care about anything in `hmc_util`, and if they do, it is probably worth exposing it directly in `infer_util`! We should also think about what goes into `util` and `infer_util`. For instance, since we are exposing our version of control flow primitives, it is worth documenting why we have that and how to use it for debugging. Otherwise, it is better to not expose it as part of the API.

numpyro/__init__.py
neerajprad: Sorry, I missed that. Good catch.

numpyro/__init__.py
neerajprad: Another reason for the separation is to just avoid circular imports - `util` shouldn't depend on any other `numpyro` modules, as it currently stands. `infer_util` on the other hand has many other dependencies like `handlers`, `infer`, etc. Right now, it is very easy for internal code to use anything from `util` without worrying about circular imports. On the other hand, utilities in `infer_util` should be mostly for end users (/tutorials/examples), and not as much for internal usage. We have a few that are used internally (e.g. potential_energy) which I think are best moved from `infer_util` to some other module for internal usage. 

Anyways, this PR is the first one. We should think more closely about our organization based on these principles and change it soon, since such changes are backwards incompatible and we will be stuck with them.

fehiepsi: I just mean to move these files to `infer` folder.

> most users of the library should care about anything in hmc_util, and if they do, it is probably worth exposing it directly in infer_util

agreed!

> For instance, since we are exposing our version of control flow primitives, it is worth documenting why we have that and how to use it for debugging. Otherwise, it is better to not expose it as part of the API.

You are right. It is better to not expose them. Only `control_flow_prims_disabled` is useful for us and for users (to debug).

PR #381, opened 2019-10-05
Add init_to_value strategy

fehiepsi:
This adds another initial strategy, which is useful for setting initial values to (some of) latent variables.

I also expose `init_strategy` to HMC class and change the initial strategy to `init_to_uniform` so that it is
+ consistent with MCMC
+ faster to find a valid initial point

I don't think that we need to follow Pyro to set default init strategy to `init_to_median`.

Blocked by #382.


fehiepsi:
Let me collect some pros and cons of these initialize strategies through some examples so we can decide what is the trade-off between those methods.

neerajprad:
Any ideas why svi.init takes 12 seconds with `init_to_median` on this simple model and how does that compare with Pyro? I am happy to change the initialization, but if it is dictated by speed concerns, it will be good to know what is the reason why this is slow and if it can be made faster?

fehiepsi:
> Any ideas why svi.init takes 12 seconds with init_to_median on this simple model

Yeah, the thing I didn't notice is the prior is Beta, which is slow to compile. A further investigation (I think that I have mentioned about it before but not sure when) shows that we always trace the model/guide 2 times to find initial params:
+ first trace gives a prototype params (for init_state of the below while_loop).
+ second trace is for the while_loop to find a valid trace.

I think we can skip the second trace if the prototype is valid. This can be done outside jit context: e.g. in svi and in single-chain mcmc. I will make a change to reduce this overhead. 

> how does that compare with Pyro?

Pyro does not face this problem.

fehiepsi:
With the latest changes, the time for `svi.init` is reduced to 7.1s.

neerajprad:
> With the latest changes, the time for svi.init is reduced to 7.1s.

I think the gamma sampler compilation makes numpyro seem slower than it is on small models, but there's little that we can do about it, I think.

fehiepsi:
> the gamma sampler compilation makes numpyro seem slower than it is on small models, but there's little that we can do about it, I think.

Yup, recent changes in JAX make it slower than before. I can't see a solution for it in the near future. When I have time, I'll look at https://github.com/google/jax/issues/552 again to see if `warp-based parallelism` helps for https://github.com/google/jax/issues/1188.

numpyro/infer/util.py
```diff
@@ -322,18 +328,25 @@ def body_fn(state):
...
+        # XXX assume we already got `prototype` with corresponding inv_transforms,
+        # we can compute the validity pf those params to avoid the below
+        # while_loop, which might be slow to compile.
+        prototype_params, inv_transforms = prototype
+        is_valid = validate_params(prototype_params, inv_transforms)
+        init_state = (0, rng, prototype_params, is_valid)
+    else:
+        _, _, prototype_params, is_valid = init_state = body_fn((0, rng, None, None))
+
+    if not_jax_tracer(is_valid):
```
neerajprad: Could you explain why we need `not_jax_tracer`?

numpyro/infer/util.py
```diff
@@ -269,7 +269,7 @@ def init_to_value(values):
 
 
 def find_valid_initial_params(rng, model, *model_args, init_strategy=init_to_uniform(),
-                              param_as_improper=False, prototype_params=None, **model_kwargs):
+                              param_as_improper=False, prototype=None, **model_kwargs):
```
neerajprad: How about also taking `inv_transforms` (maybe calling it `support_transforms` or `constraint_transforms`) as input instead of having a tuple for `prototype`?

neerajprad:
Please disregard my comments above, seems like I have been reviewing an older version of the PR. 😿

numpyro/infer/util.py
```diff
@@ -286,12 +331,16 @@ def body_fn(state):
     if prototype_params is not None:
         init_state = (0, rng, prototype_params, False)
     else:
-        init_state = body_fn((0, rng, None, None))
+        _, _, prototype_params, is_valid = init_state = body_fn((0, rng, None, None))
+        if not_jax_tracer(is_valid):
```
neerajprad: Just for my understanding, could you explain why we need this?

fehiepsi: This is not necessary for the usual case of SVI. I just want to make sure that if someone runs `svi` inside vmap/jit (e.g. vmap over different learning rate/initial seed/shards of data), the check `if is_valid` will trigger an error.

PR #382, opened 2019-10-05
A pass for 0.2.1 release

fehiepsi:
This is my attempt to reorganize the repo for the next release:
+ move `hmc_util` to `infer`
+ not expose `util` file
+ expose `optim`
+ expose `biject_to` to `distributions` namespace (like PyTorch)
+ expose `set_platform` and `set_host_device_count` to `numpyro` namespace (mimic the way we do for `pyro.set_rng_seed`)
+ move `infer_util` to `infer`, expose `predictive`, `log likelihood`, init strategies.
+ move `initialize_model` out of `hmc_util`: we need so many imports from `infer_util` for this function; after this, `hmc_util` is pretty standalone.


neerajprad:
Do you think we should also move `infer_util` to `infer`, and expose `predictive`, `log_likelihood` etc. directly from the `infer` module?

fehiepsi:
That makes sense to me. I can't find a reason for not doing so. That way also helps us decide what should be exposed to users.

docs/source/mcmc.rst
```diff
@@ -34,12 +34,12 @@ Hamiltonian Monte Carlo
...
 .. autofunction:: numpyro.util.fori_collect
 
-.. autofunction:: numpyro.hmc_util.consensus
+.. autofunction:: numpyro.infer.hmc_util.consensus
 
-.. autofunction:: numpyro.hmc_util.parametric
+.. autofunction:: numpyro.infer.hmc_util.parametric
 
-.. autofunction:: numpyro.hmc_util.parametric_draws
+.. autofunction:: numpyro.infer.hmc_util.parametric_draws
```
neerajprad: Not for this PR, but it will be nice if we had some example / tutorials to demonstrate how these are useful.

neerajprad:
The organization looks really good, and likely something that we don't need to touch in later releases. I'll take another look at the docs and the modules before release.

docs/source/mcmc.rst
fehiepsi: Sure, parallel hmc is a cool feature to introduce. I'll write a tutorial when flowhmc experiment is done (which I am targeting for the next week). :)

Issue #378, opened 2019-10-02
Revise the pattern of loss_fn and init_strategy  


fehiepsi:
Currently, we implemented elbo / init strategies as functions. This makes it a bit tricky to change the default parameters of these utilities (e.g. num_particles in "future" `elbo`, num_samples in `init_to_median`, `radius` in `init_to_uniform`). I would propose the following pattern:
```
def _init_to_median(site, num_samples=15):
    ...

def init_to_median(num_samples=15):
    return partial(_init_to_median, num_samples=num_samples)
```
and similarly,
```
def _elbo(..., num_particles=1):
    return ...

def elbo(num_particles=1):
    return partial(_elbo, num_particles=num_particles)
```
My motivation is:
+ I am intending to add a new init strategy, which set init params to my desired values (in Pyro, we can do it with `pyro.param("auto_loc", desired_value)` but there is no easy way to do it in NumPyro). So I can use something like `init_strategy=init_to_desired_values(desired_value)`
(rather than `init_strategy=partial(init_to_desired_values, desired_values=desired_values)`) when constructing an autoguide.
+ Some loss function such as `renyi_elbo` will require user set `num_particles`.

All of those can be achieved by requiring users to use `partial(fn, settings)` but it is so inconvenient. The proposal pattern has been used in various places in JAX: e.g. to initialize stax parameters, to create optimizer.

**What will change if the above proposal is accepted:**
+ To create an SVI instance, the pattern will be changed a little bit
```
svi = SVI(model, guide, elbo(), optim.Adam(0.1))
```

fehiepsi:
Woohoo, making PRs for this proposal is fun. Happy to close!

PR #383, opened 2019-10-06
Convert elbo into a class for compatibility with Pyro

fehiepsi:
As discussed in #378, it is better to use a generator to create loss_fn because
+ Users don't care about elbo's args, kwargs.
+ This allows us to set hyperparameters of these loss functions.


neerajprad:
How about just using an`ELBO` class like in Pyro, which should probably address all of the issues you raised in #378 and be much more extensible?

fehiepsi:
Yup, we can make elbo a class to match Pyro. It will be a very simple class, that works just like a function (no stuff like differentiable_loss, loss_and_grad, loss,...). But again, matching Pyro interface is a plus for users. :)

docs/source/svi.rst
```diff
@@ -9,4 +9,15 @@ Stochastic Variational Inference (SVI)
 
 ELBo
 ----
-.. autofunction:: numpyro.infer.svi.elbo
+
```
neerajprad: Not for this PR, but we should rename this `ELBO` too to match the rest of the codebase.

examples/minipyro.py
```diff
@@ -1,14 +1,13 @@
...
-import jax
 from jax import random
 import jax.numpy as np
 from jax.random import PRNGKey
 
 import numpyro
 from numpyro import optim
 import numpyro.distributions as dist
-from numpyro.infer import SVI, elbo
+from numpyro.infer import SVI, Trace_ELBO
```
neerajprad: We are stuck with the `Trace_ELBO` name in Pyro due to backwards compatibility reasons. Let us put this in compat if we need to do this for Pyro compatibility, and use the `ELBO` name in the main codebase.

numpyro/infer/elbo.py
```diff
@@ -0,0 +1,89 @@
...
+        :param args: arguments to the model / guide (these can possibly vary during
+            the course of fitting).
+        :param kwargs: keyword arguments to the model / guide (these can possibly vary
+            during the course of fitting).
+        :return: negative of the Evidence Lower Bound (ELBo) to be minimized.
+        """
+        raise NotImplementedError
+
+
+class Trace_ELBO(ELBO):
```
neerajprad: I would suggest merging this with `ELBO` and removing `Trace_ELBO`. Other classes are free to replace the loss method as they see fit.

examples/minipyro.py
fehiepsi: done

docs/source/svi.rst
fehiepsi: done

numpyro/contrib/autoguide/__init__.py
```diff
@@ -383,3 +388,42 @@ def _get_transform(self):
             arnn = numpyro.module('{}_arn__{}'.format(self.prefix, i), arn, (self.latent_size,))
             flows.append(InverseAutoregressiveTransform(arnn))
         return ComposeTransform(flows)
+
+
+class AutoContinuousELBO(ELBO):
```
neerajprad: So happy to see this refactored out!

PR #384, opened 2019-10-06
use args kwargs in svi

fehiepsi:
This PR makes SVI init/update/evaluate signature consistent with Pyro, MCMC, predictive. As shown in autoguide tests, the code is simpler and does not require us to repeat args and put data to a tuple.

As shown in VAE example, we don't need to distinguish static/dynamic args.


numpyro/infer/svi.py
```diff
@@ -243,37 +230,37 @@ def get_params(self, svi_state):
...
         """
         Take a single step of SVI (possibly on a batch / minibatch of data),
         using the optimizer.
 
         :param svi_state: current state of SVI.
-        :param tuple model_args: dynamic arguments to the model.
-        :param tuple guide_args: dynamic arguments to the guide.
+        :param args: arguments to the model / guide (these can possibly vary during
+            the course of fitting).
+        :param kwargs: keyword arguments to the model / guide.
```
neerajprad: Hmm..I think it is better to store kwargs in `__init__` since those are static args. Otherwise, using kwargs with JIT will result in recompilation of `update_fn` every time we call `update_fn`, like `update_fn(svi_state, .., a=x, b=y)`.

fehiepsi: Nice catch!!! I'll think more about this issue.

fehiepsi: @neerajprad I just learned that jit works for kwargs too: [gist](https://gist.github.com/fehiepsi/7f70265780aaabfa32c5819b479399fd)

fehiepsi: This might not be a problem for us because svi.update is mostly used in a loop, where its signatures are consistent.

neerajprad: I think that due to some recent changes, this works, but it will only work for array arguments. Suppose we pass in a static argument like `encoder=..` which is a python function for a VAE, then with this, we cannot simply @jit the `update` step. 

Actually, I am not even sure what the current behavior of the JIT is - it doesn't seem possible to mark the encoder as a static arg either, which means that users will need to write a wrapper for `update_fn` just to be able to JIT it when they have a non-array static argument. Why not just store it instead?

fehiepsi: The following code works
```
def f(x):
    return x

def g(x, *args, **kwargs):
    return x + (args[0](x) if args else 0) + (list(kwargs.values())[0](x) if kwargs else 0)

jit(g, static_argnums=(1,))(1, f)
```
so as long as the `encoder` is passed as a static `arg`, not `kwargs`, then you can jit `update_fn`.

> Why not just store it instead?

I think it is reasonable to do so. Not-valid-JAX type should be considered as `static_args`. It will be harder to decide if a keyword is static or dynamic.

> we cannot simply @jit the update step.

This seems like a JAX knowledge. I think we can make a gotcha notebook (similar to [JAX one](https://jax.readthedocs.io/en/latest/notebooks/Common_Gotchas_in_JAX.html)) which collects what we have learned to use JAX correctly in NumPyro. WDYT?

neerajprad: Let me try to clarify what I meant. In Pyro's JITTrace_ELBO, keyword arguments are used to denote static arguments. A change in these will retrigger compilation. This was also the case earlier, where keyword arguments in JAX would be treated as static arguments and we would have been able to pass `encoder=some_python_callable` and jitting the update step would have worked as is. 

This was changed in https://github.com/google/jax/pull/595, and it is now not possible to treat keyword arguments as static arguments (see https://github.com/google/jax/issues/1159, for example). As such, for the same model to work in Pyro and NumPyro identically, the user will now need to write a wrapper over `update_fn` so that static arguments are instead positional arguments. I think it is therefore better for `update` (and, `init`) to not take kwargs at all. Then we can just treat keyword arguments as static arguments as they are documented and treated in Pyro too. The only benefit to passing keyword arguments to `update` (that I can see) is a surface level similarity with Pyro, but will require users to actually write a wrapper over `update` to make it work in practice. Maybe we can put this in `numpyro.compat.infer` if it is really important to pass kwargs to `update`.

> This seems like a JAX knowledge. I think we can make a gotcha notebook (similar to JAX one) which collects what we have learned to use JAX correctly in NumPyro. WDYT?

It doesn't have to be. I think if we just treat args as we were doing earlier, we don't need to do anything else. The documentation is self-explanatory.

numpyro/infer/svi.py
```diff
@@ -44,26 +44,23 @@ def svi(model, guide, loss, optim, **kwargs):
...
+    def init_fn(rng, *args, **kwargs):
         """
 
         :param jax.random.PRNGKey rng: random number generator seed.
-        :param tuple model_args: arguments to the model (these can possibly vary during
-            the course of fitting).
-        :param tuple guide_args: arguments to the guide (these can possibly vary during
+        :param args: arguments to the model / guide (these can possibly vary during
             the course of fitting).
+        :param kwargs: keyword arguments to the model / guide.
```
neerajprad: Could you add that: (a) these should have array values, and (b) they can also vary during the course of fitting (like `args`, just to clarify that there is nothing special about these kwargs)?

numpyro/infer/svi.py
```diff
@@ -191,36 +187,33 @@ class SVI(object):
...
+    def init(self, rng, *args, **kwargs):
         """
 
         :param jax.random.PRNGKey rng: random number generator seed.
-        :param tuple model_args: arguments to the model (these can possibly vary during
-            the course of fitting).
-        :param tuple guide_args: arguments to the guide (these can possibly vary during
+        :param args: arguments to the model / guide (these can possibly vary during
             the course of fitting).
+        :param kwargs: keyword arguments to the model / guide.
```
neerajprad: Same comment as earlier for `svi`.

numpyro/infer/svi.py
neerajprad: Actually, just (b) is enough. They are free to use non-array values as they like, provided that they treat it as such when using JIT.

fehiepsi: Thanks, @neerajprad ! Just done it.

Issue #373, opened 2019-09-30
Masking out of support log_prob


fehiepsi:
Several months ago, we decided to not have "masking" pattern when computing log_prob for out-of-support samples. I thought that it was good because we already have a check for invalid samples.

We did so because masking is not necessary for our inference algorithms. Now, given that users will want to use our distributions for something else: e.g. to compute custom log_prob, to plot log_prob, we should have a masking pattern for popular distribution (maybe not for all distribution, to avoid expensive checks such as positive_definite). Currently, Uniform(0, 1).log_prob(-1) will be `0` (instead of `-inf`), which is not desirable.

fehiepsi:
This [gist](https://nbviewer.jupyter.org/gist/fehiepsi/eb3053aad5ece9ee4feba452d22e2740) shows that the distributions have this issue are:
+ Exponential
+ HalfCauchy/HalfNormal
+ TruncatedCauchy/TruncatedNormal
+ Uniform
Masking for these distributions is easy so I think that it is worth to add masks for these distributions.

neerajprad:
> Now, given that users will want to use our distributions for something else: e.g. to compute custom log_prob, to plot log_prob, we should have a masking pattern for popular distribution (maybe not for all distribution, to avoid expensive checks such as positive_definite). Currently, Uniform(0, 1).log_prob(-1) will be 0 (instead of -inf), which is not desirable.

I think the correct solution here should be for us to turn on `validate_args` by default, so that users can catch bugs when they are wrong. Otherwise, this will turn on some validation checks on large tensors by default with no way of turning it off. If a user is confident of their model and wants to run it as fast as possible, it should be possible to not have to do these checks which will be possible by just turning distribution validation check off.

PR #385, opened 2019-10-06
mask log prob of univariate distribution

fehiepsi:
Addresses #373.

At first, I thought that we need a general approach to this issue. However, when plotting out the log_prob of all distributions, we can see that there are just a few distributions that have this issue. So I go ahead and do the masking job for each of them, instead of trying to think of a general solution.

Affected distributions:
+ Exponential
+ HalfNormal/HalfCauchy/TruncatedNormal/TruncatedCauchy
+ Uniform
+ Bernoulli
+ Categorical

Not tested distributions:
+ all multivariate ones

### DIFF
--- a/numpyro/distributions/continuous.py
+++ b/numpyro/distributions/continuous.py
@@ -156,7 +156,7 @@ class Exponential(Distribution):
     def log_prob(self, value):
         if self._validate_args:
             self._validate_sample(value)
-        return np.log(self.rate) - self.rate * value
+        return np.where(value >= 0, np.log(self.rate) - self.rate * value, -np.inf)
 
     @property
     def mean(self):
@@ -262,7 +262,7 @@ class HalfCauchy(Distribution):
     def log_prob(self, value):
         if self._validate_args:
             self._validate_sample(value)
-        return self._cauchy.log_prob(value) + np.log(2)
+        return np.where(value >= 0, self._cauchy.log_prob(value) + np.log(2), -np.inf)
 
     @property
     def mean(self):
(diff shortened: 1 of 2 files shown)


neerajprad:
Refer to my comment in #373. I think we shouldn't turn this on by default. We call distribution methods many many times and for large models, this will have non-negligible overhead. How about we instead turn on distributions validation check by default, and provide a class attribute to turn it off like in PyTorch. That should be sufficient for the use cases you mention.

fehiepsi:
Though I don't think we should care about masking op, this solution might not be good.

> turn on distributions validation check by default

I think that we can have the check by default (though it will be slower for support-checking and we need a workaround to disable the check inside jit/vmap/pmap). A context manager `numpyro.enable_validation()` would be helpful (somehow we forgot about it during refactorings?). Let's assume that we keep the default behavior (faster, no masking) and discuss what we should do if we turn on `checking`. I don't think that raising an error for out-of-support value is a good solution because those values should have probability `0`. A better solution to me is to raise a warning and return a masked result.

```
log_prob = ...
if self._validate_args:
    mask = self._validate_sample(value)
    log_prob = np.where(mask, log_prob, -np.inf)
```
and `self._validate_sample` is implemented as
```
mask = support(value)
if check is outside of jit/vmap/pmap:
    if np.all(mask):  # together with 
        warning("value is outside of support"
    return mask
```
WDYT? This is a more reasonable approach to me.

neerajprad:
> A better solution to me is to raise a warning and return a masked result.

That sounds reasonable to me. Our current design forces the user to do this masking instead, but I like your idea of doing this masking and returning `-inf` for out of support values, if we can do this consistently for all distributions.

fehiepsi:
Yeah, let me create another PR for this.

PR #388, opened 2019-10-07
Out of support mask

fehiepsi:
Follow up the discussion at #385. This PR adds `validate_sample` decorator to mask out-of-support values as `-np.inf`. @neerajprad What do you think about this way?

### DIFF
--- a/numpyro/distributions/distribution.py
+++ b/numpyro/distributions/distribution.py
@@ -28,7 +28,8 @@ import jax.numpy as np
 
 from numpyro.distributions.constraints import is_dependent
 from numpyro.distributions.transforms import Transform
-from numpyro.distributions.util import lazy_property, sum_rightmost
+from numpyro.distributions.util import lazy_property, sum_rightmost, validate_sample
+from numpyro.util import not_jax_tracer
 
 
 class Distribution(object):
@@ -78,8 +79,10 @@ class Distribution(object):
                     continue
                 if is_dependent(constraint):
                     continue  # skip constraints that cannot be checked
-                if not np.all(constraint(getattr(self, param))):
-                    raise ValueError("The parameter {} has invalid values".format(param))
+                is_valid = np.all(constraint(getattr(self, param)))
+                if not_jax_tracer(is_valid):
+                    if not is_valid:
+                        raise ValueError("The parameter {} has invalid values".format(param))
         super(Distribution, self).__init__()
 
     @property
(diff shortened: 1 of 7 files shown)


neerajprad:
> @neerajprad What do you think about this way?

This looks pretty neat. Can we toggle `Distribution._validate_args` to True now?

fehiepsi:
@neerajprad I have copied Pyro implementation of `validation_enabled` and `enable_validation` utilities to here and also add a note that those utilities do not take effect under JIT context.

numpyro/distributions/distribution.py
```diff
@@ -22,11 +22,48 @@
...
 from numpyro.distributions.constraints import is_dependent
 from numpyro.distributions.transforms import Transform
-from numpyro.distributions.util import lazy_property, sum_rightmost
+from numpyro.distributions.util import lazy_property, sum_rightmost, validate_sample
+from numpyro.util import not_jax_tracer
+
+_VALIDATION_ENABLED = False
+
+
+def enable_validation(is_validate=True):
```
neerajprad: This can be a more general utility that can be used for validation outside the context of distributions. We should move this to `util` and import it in `numpyro.__init__` instead.

numpyro/distributions/distribution.py
```diff
@@ -65,6 +102,12 @@ class Distribution(object):
     reparametrized_params = []
     _validate_args = False
 
+    @staticmethod
```
neerajprad: Would just renaming this to `Distribution.set_args_validation(...)`, the default is kind of implicit at this level. Also, have we turned on distribution validation check? I think we should do that by default now.

numpyro/distributions/distribution.py
fehiepsi: I exposed this to numpyro namespace. The reason for this to be here is to avoid circular import: these utils require access to Distribution class.

numpyro/distributions/distribution.py
fehiepsi: This is just to match PyTorch name, not important I guess?

> turn on validation

We can turn on by default but we should turn it off in SVI and MCMC because it will add overhead while give no effect. WDYT?

numpyro/distributions/distribution.py
neerajprad: hmm..yeah, I think lets keep it here, in that case. If we have more validation checks we can do provide a utility method to do it.

numpyro/distributions/distribution.py
neerajprad: This is not particularly important, since we don't use that `staticmethod` except from the utility methods in Pyro. I will leave it up to you in that case, I don't have a strong opinion. 

> We can turn on by default but we should turn it off in SVI and MCMC because it will add overhead while give no effect. WDYT?

Yeah, let us just keep it as is and advertise `numpyro.enable_validation` in our tutorials/examples instead.

PR #390, opened 2019-10-08
Add a compat version for SVI, Trace_ELBO.

fehiepsi:
This PR adds SVI, Trace_ELBO to `compat`. So that we can test SVI in pyro.generic with NumPyro backend (ref. https://github.com/pyro-ppl/pyro/issues/2053).

Of course, we can't achieve the full potential of NumPyro SVI, where the overhead of each step can be optimized with `lax.fori_loop` or `lax.scan`. But we already jit each step in the implementation, so this overhead (about 1s/10000 steps) can only be observed when we run a large amount of steps in small models. In addition, JAX has [asynchronous dispatch](https://jax.readthedocs.io/en/latest/async_dispatch.html) mechanism, which can help blend this overhead with data loader time.

Tested in minipyro example, using Pyro style.

Blocked by #383.


neerajprad:
Could you merge the latest changes from master?

numpyro/compat/infer.py
```diff
@@ -95,3 +97,56 @@ def get_samples(self, num_samples=None, group_by_chain=False):
...
+            self.svi_state, loss = jit(self.update)(self.svi_state, *args, **kwargs)
+        except TypeError as e:
+            if 'not a valid JAX type' in str(e):
+                raise TypeError('NumPyro backend requires args, kwargs to be arrays or tuples, '
+                                'dicts of arrays.')
+            else:
+                raise e
+        return loss
+
+    def get_param_store(self):
```
neerajprad: Can we name this `get_params` instead, so as not to mislead users that there is really a param store?

fehiepsi: Oppss... I have no idea why I thought that this is going to override `get_params` method of numpyro svi. Python can do it with `super(...)`... I think I need to refresh my mind. :))

PR #394, opened 2019-10-09
fix bug at vectorized method

fehiepsi:
The shape of the vectorized method is num_samples x num_chains. We need to swap it. I added a test to make sure that we get the correct shape.

### DIFF
--- a/numpyro/infer/mcmc.py
+++ b/numpyro/infer/mcmc.py
@@ -2,7 +2,7 @@ from abc import ABC, abstractmethod
 from collections import namedtuple
 import functools
 import math
-from operator import attrgetter, itemgetter
+from operator import attrgetter
 import os
 import warnings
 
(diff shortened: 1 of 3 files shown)


numpyro/infer/mcmc.py
```diff
@@ -722,7 +722,7 @@ def __init__(self,
                 "CI" in os.environ or "PYTEST_XDIST_WORKER" in os.environ):
             self.progress_bar = False
 
-        self._collect_fields = ('z',)
+        self.collect_fields = ('z',)
```
fehiepsi: I exposed collect_fields so that we can `dict(zip(mcmc.collect_fields, mcmc.get_samples()))` to get the dict which maps field name to its value.

fehiepsi:
Oppsss, I mistakenly pushed the change to this PR...

numpyro/infer/mcmc.py
fehiepsi: This is helpful for [arviz PR](https://github.com/arviz-devs/arviz/blob/master/arviz/data/io_numpyro.py#L41) too. Currently, I have to use `mcmc._samples` to access the name of these fields. With this change, we won't need to use any hidden property for the PR in arviz.

neerajprad: How about we do this by default then in `get_samples`, since presumably if you asked for multiple fields it will help to return a dict instead of a tuple? Also, this should only affect the `get_samples` method.

```
samples = get_items(self._samples) if group_by_chain else get_items(self._samples_flat)
return dict(zip(self._collect_fields, samples)) if isinstance(samples, tuple) else samples
```

fehiepsi: It is equivalent good to me. Both are sufficient for my usage. :)

neerajprad: Let us return the dict instead then. I think that will be more straightforward than having to query for collect_fields from MCMC.

fehiepsi: While changing the code for this, I found that the change might lead to confusion. E.g. what if users don’t specify collect_fields, will we return a dict with z field in that case? Why sometime we need to call get_samples()[“z”], while sometime we don’t need to specify [“z”]... I think it is better to:
+ always collect z field
+ only requires users specify not-z fields in collect_fields
+ have an additional method such as get_fields() to return those additional fields
This is more explicit to me and does not require extra thoughts to use mcmc correctly. Users also don’t jump into the issue that they forget to specify z in collect_fields. WDYT?

neerajprad: > E.g. what if users don’t specify collect_fields, will we return a dict with z field in that case? 

No, this change will only affect the return type when there are multiple fields. This is also the behavior currently, an `np.ndarray` is returned when only a single field is present, otherwise a tuple is returned. All this change does is to return a dict instead of a tuple which is always well-defined.

> Why sometime we need to call get_samples()[“z”], while sometime we don’t need to specify [“z”].

That's also the behavior right now - with multiple fields, users will need to index into the correct position to get "z", i.e. they will need to write `get_samples()[0]` for instance. If they specify only "z" in collect fields or for that matter any other single field, they'll get a numpy array instance and not a dict.

> always collect z field
> only requires users specify not-z fields in collect_fields

That seems reasonable to me. We should specify this in the docstring if we make this change, and maybe rename `collect_fields` to `extra_fields` or something.

> have an additional method such as get_fields() to return those additional fields
This is more explicit to me and does not require extra thoughts to use mcmc correctly. 

My thoughts, based on the comment above, is that its not needed. When a user says, they want extra fields, it is reasonable to return the dict by default. I don't see why they should have an expectation that a tuple will be returned instead.

neerajprad: If you think this needs more thought / discussion, we can just go with your proposed change of using `.collect_fields` attribute, and revisit this API decision later down the line.

fehiepsi: > rename collect_fields to extra_fields or something

Sounds reasonable to me.

> When a user says, they want extra fields, it is reasonable to return the dict by default.

To me, returning a dict will better for users (at least for me, I took extra codes to merge collect_fields with `get_samples()` into a dict, then use summary to get an overview of some stats e.g. potential_energy when I want to compare with lp__ field in Stan; summary of stats is also adopted in [rlgt](https://github.com/pyro-ppl/numpyro/blob/10594c8864d3ed140f8d1927561c08d074a3742d/notebooks/Rlgt.ipynb) package)

It would be tricky to decide what to do for third-party library like arviz: what will be the result of mcmc.get_samples() if arviz doesn't know the method that its user has collected.

When we know what we need to do, this extra method is not needed. Even though I know, previously I have to merge the fields with the result of get_samples; now I have to filter out `z` field from that result to summarize those stats...

neerajprad: > It would be tricky to decide what to do for a third-party library like arviz: what will be the result of mcmc.get_samples() if arviz doesn't know the method that its user has collected.

I understand your suggestion much better now. I misread your earlier comment as saying that we should add another argument to `get_samples`, sorry about that! Instead what you are saying is that even our current interface isn't ideal (which I agree with), and having `get_extra_fields` sounds like an improvement because our return types will be consistent no matter the arguments provided to this class. Yes, please go for it! 👍

PR #396, opened 2019-10-10
Add tests from pyroapi

neerajprad:
Addresses https://github.com/pyro-ppl/pyro-api/issues/2.


numpyro/compat/pyro.py
```diff
@@ -1 +1,10 @@
+import warnings
+
+from numpyro.compat.util import UnsupportedAPIWarning
 from numpyro.primitives import module, param, plate, sample  # noqa: F401
+
+
+def get_param_store():
+    warnings.warn('NumPyro does not have a parameter store.', category=UnsupportedAPIWarning)
```
fehiepsi: nit: I think it will be helpful to add: "If you are doing SVI, values of `param` primitives can be obtained by using `svi.get_params()` method."

setup.py
```diff
@@ -37,7 +37,11 @@
     ],
     extras_require={
         'doc': ['sphinx', 'sphinx_rtd_theme'],
-        'test': ['flake8', 'pytest>=4.1'],
+        'test': [
+            'flake8',
+            'pytest>=4.1',
+            'pyro-api@https://api.github.com/repos/pyro-ppl/pyro-api/tarball/master'
```
fritzo: :+1: I'll use this method in funsor as well.

test/conftest.py
```diff
@@ -9,3 +11,10 @@ def pytest_runtest_setup(item):
     if 'JAX_ENABLE_x64' in os.environ:
         config.update('jax_enable_x64', True)
     set_rng_seed(0)
+
+
+def pytest_runtest_call(item):
```
fritzo: I would suggest limiting this hook's scope by creating a directory structure
```
numpyro
+-- test/
    +-- conftest.py  # unmodified
    +-- pyroapi/
        +-- conftest.py  # xfail if not implemented only applies to this subdir
        +-- test_pyroapi.py
```
I have used this pattern in funsor and am refactoring pyroapi to use this pattern in https://github.com/pyro-ppl/pyro-api/pull/5

numpyro/compat/util.py
```diff
@@ -0,0 +1,7 @@
+class UnsupportedAPIWarning(Warning):
```
fritzo: Looks great! If this class proves useful, we may want to move it to `pyroapi`, maybe calling it a `NotImplementedWarning` to match `NotImplementedError`.

test/test_pyroapi.py
```diff
@@ -0,0 +1,11 @@
+from pyroapi import handlers, pyro_backend
+from pyroapi.tests import *  # noqa F401
+
+pytestmark = pytest.mark.filterwarnings("ignore::numpyro.compat.util.UnsupportedAPIWarning")
+
+
+@pytest.yield_fixture
+def backend():
+    with pyro_backend('numpy'):
+        with handlers.seed(rng_seed=1):
```
fritzo: Hmm, I think we should try to avoid any backend-specific code here. I have a few questions:
1. Why is this needed here, even though we don't have this in pyroapi/test/test_tests.py?
2. Can we add a hook to the pyro_backend context manager so that, for each backend a set of contexts is entered in pyro_backend? That way we can set a default seed when entering a `with pyro_backend('numpy')` context?

test/conftest.py
neerajprad: Yes, I was a little worried about it affecting other tests. This sounds like a good way to handle it.

test/test_pyroapi.py
fritzo: @neerajprad does this solution seem to hacky?
https://github.com/pyro-ppl/pyro-api/pull/new/set-seed-by-default

neerajprad: 1. All inference algorithms require an explicit `rng_seed` in NumPyro, which is used for sampling from the models, etc. The tests in `test_tests.py` are only currently testing functionality and therefore don't need to fix the rng seed (for other backends). If we ever fix the seed, this would be a backend-agnostic way of doing it. An alternative would be to sample a `random.randint` in the compatibility wrapper as a seed if we aren't inside the `seed` handler context. WDYT?
2. We can do that, but we can only do so for the known backends, or provide extra arguments so that new backends can similarly register context managers to be entered into. Also, this will have a different behavior since the tests will be deterministic for NumPyro (also a problem with using handlers like above) unlike other backends, so using a global non-deterministic seed seems like a better solution.

> @neerajprad does this solution seem to hacky?
https://github.com/pyro-ppl/pyro-api/pull/new/set-seed-by-default

The only issue that I see there is that it will fix the seed by default for all tests? Does that seem fine to you?

neerajprad: The good thing with your proposal is that is guaranteed to work for other cases too, e.g. when sampling from a model etc. outside an inference context.

neerajprad: I commented on your proposal. I think that is doable if we change the behavior of nested seed handlers in NumPyro.

fritzo:
Should we merge this before https://github.com/pyro-ppl/pyro-api/pull/6, or do you want to wait and remove the seed logic in this PR?

test/pyroapi/test_pyroapi.py
```diff
@@ -0,0 +1,12 @@
...
+from pyroapi.tests import *  # noqa F401
+import pytest
+
+pytestmark = pytest.mark.filterwarnings("ignore::numpyro.compat.util.UnsupportedAPIWarning")
+
+
+@pytest.yield_fixture
+def backend():
+    with pyro_backend('numpy'):
+        with handlers.seed(rng_seed=1):
```
fritzo: no longer needed, right?

neerajprad: Yes, let me resolve that.

fritzo:
Can we merge this to unblock https://github.com/pyro-ppl/pyro-api/pull/5 ?

neerajprad:
> Can we merge this to unblock pyro-ppl/pyro-api#5 ?

Done!

PR #397, opened 2019-10-11
Make behavior of nested seed handler consistent with Pyro

neerajprad:
Refer to https://github.com/pyro-ppl/pyro-api/pull/6 for details.

The current behavior of the seed handler is that the outermost context determines the rng seed, i.e. in `seed(seed(fn, ..), 1)`, the seed used is always 1. This is because of the order in which the effect handlers are applied. This is in contrast to Pyro where seeding happens in `__enter__` and therefore the innermost context determines the seed used. This changes the behavior to match that of Pyro. More importantly, this will let us use the pattern in https://github.com/pyro-ppl/pyro-api/pull/6, so that we can have default seeding in an external context manager which will can be overridden by individual tests. 

cc. @fritzo.

### DIFF
--- a/test/test_handlers.py
+++ b/test/test_handlers.py
@@ -1,4 +1,4 @@
-from numpy.testing import assert_allclose
+from numpy.testing import assert_allclose, assert_raises
 import pytest
 
 from jax import jit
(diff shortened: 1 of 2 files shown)


test/test_handlers.py
```diff
@@ -53,6 +53,21 @@ def _sample():
     assert_allclose(xs, ys, atol=1e-6)
 
 
+def test_nested_seeding():
+    def fn(rng_1, rng_2, rng_3, sample_x=False):
```
fritzo: what is the `sample_x` for?

neerajprad: My bad, there was a case I wanted to test against, but I am not changing sample_x. Will figure out a better way to test it, and fix.

test/test_handlers.py
```diff
@@ -53,6 +53,23 @@ def _sample():
...
+                xs.append(numpyro.sample('x', dist.Normal(0., 1.)))
+                with handlers.seed(rng=rng_3):
+                    xs.append(numpyro.sample('y', dist.Normal(0., 1.)))
+        return np.stack(xs)
+
+    s1, s2 = fn(0, 1, 2), fn(3, 1, 2)
+    assert_allclose(s1, s2)
+    s1, s2 = fn(0, 1, 2), fn(3, 1, 4)
+    assert_allclose(s1[0], s2[0])
+    assert_raises(AssertionError, assert_allclose, s1[1], s2[1])
```
fehiepsi: interesting

PR #399, opened 2019-10-14
Defer to autodelta if hessian of laplace approximation is singular

fehiepsi:
When hessian is singular, we will get NaN samples because the transform `scale_tril` is NaN. But there are problems where there is no (or little) uncertainty about MAP point. This PR masks out `scale_tril` by zero if hessian is singular, hence gives us MAP point as posterior samples.

In addition, I wrap `scale_factor` positive check in `scale` handler by `not_jax_tracer` to be able to jit models with scale handler.

### DIFF
--- a/numpyro/contrib/autoguide/__init__.py
+++ b/numpyro/contrib/autoguide/__init__.py
@@ -1,5 +1,6 @@
 # Adapted from pyro.contrib.autoguide
 from abc import ABC, abstractmethod
+import warnings
 
 from jax import hessian, random, vmap
 from jax.experimental import stax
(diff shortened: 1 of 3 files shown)


numpyro/contrib/autoguide/__init__.py
```diff
@@ -366,6 +366,7 @@ def loss_fn(z):
         loc = params['{}_loc'.format(self.prefix)]
         precision = hessian(loss_fn)(loc)
         scale_tril = cholesky_inverse(precision)
+        scale_tril = np.where(np.isnan(scale_tril), 0., scale_tril)
```
neerajprad: Do you think it is worth raising a warning here too?

fehiepsi: Sounds reasonable to me. I'll add it.

neerajprad:
Please feel free to merge at your convenience.

PR #401, opened 2019-10-17
More compatibility tweaks for pyro-api

neerajprad:
Follow up on https://github.com/pyro-ppl/pyro-api/pull/8 in pyro-api, this increases the coverage of tests with the following changes to the compatibility wrapper:

 - Added support for `ClippedAdam` optimizer.
 - Wrappers to allow for torch style zeros/ones array initializers. i.e. in pytorch, `torch.ones(2, 3)`, `torch.ones((2, 3))` and `torch.ones([2, 3])` can all be used to initialize a tensor of size (2, 3). In NumPy, the first invocation isn't allowed.
 - Allow accessing SVI parameters via the global param store when `pyro.param` is used outside an inference context. This doesn't support the full param store API e.g. initializing using constraints etc., which are handled by inference algorithms directly in NumPyro. This simply updates the global param store when `svi.step` is called. 

With these fixes, all tests pass expect `test_mean_field ` which isn't implemented.

### DIFF
--- a/numpyro/compat/infer.py
+++ b/numpyro/compat/infer.py
@@ -132,7 +132,7 @@ class SVI(svi.SVI):
                                 'dicts of arrays.')
             else:
                 raise e
-        params = super(SVI, self).get_params(self.svi_state)
+        params = jit(super(SVI, self).get_params)(self.svi_state)
         get_param_store().update(params)
         return loss
 
--- a/numpyro/compat/infer.py
+++ b/numpyro/compat/infer.py
@@ -4,6 +4,7 @@ from jax import jit
 
 import numpyro
 import numpyro.distributions as dist
+from numpyro.compat.pyro import get_param_store
 from numpyro.infer import elbo, mcmc, svi
 
 
@@ -131,6 +132,8 @@ class SVI(svi.SVI):
                                 'dicts of arrays.')
             else:
                 raise e
+        params = super(SVI, self).get_params(self.svi_state)
+        get_param_store().update(params)
         return loss
 
     def get_params(self):
--- a/numpyro/compat/ops.py
+++ b/numpyro/compat/ops.py
@@ -1,3 +1,34 @@
+import numpy as onp
+
+import jax.numpy as np
 from jax.numpy import *  # noqa: F401, F403
 
 tensor = array  # noqa: F405
+
+randn = onp.random.randn
+
+
+# Provide wrappers to initialize ones/zeros using the pytorch convention
+# of using *sizes. e.g. ops.ones(2, 3) as well as ops.ones((2, 3)) can
+# be used to initialize an array of ones with shape (2, 3).
+
+def ones(*sizes, **kwargs):
+    if len(sizes) == 0:
+        raise ValueError('Positional `size` argument not provided.')
+    elif len(sizes) == 1:
+        if isinstance(sizes[0], (tuple, list)):
+            sizes = sizes[0]
+    if not all([isinstance(s, int) for s in sizes]):
+        raise ValueError('Invalid data type for `size` provided.')
+    return np.ones(sizes, **kwargs)
+
+
+def zeros(*sizes, **kwargs):
+    if len(sizes) == 0:
+        raise ValueError('Positional `size` argument not provided.')
+    elif len(sizes) == 1:
+        if isinstance(sizes[0], (tuple, list)):
+            sizes = sizes[0]
+    if not all([isinstance(s, int) for s in sizes]):
+        raise ValueError('Invalid data type for `size` provided.')
+    return np.ones(sizes, **kwargs)
--- a/numpyro/compat/pyro.py
+++ b/numpyro/compat/pyro.py
@@ -1,13 +1,30 @@
 import warnings
 
 from numpyro.compat.util import UnsupportedAPIWarning
-from numpyro.primitives import module, param, plate, sample  # noqa: F401
+from numpyro.primitives import module, param as _param, plate, sample  # noqa: F401
+
+
+_PARAM_STORE = {}
 
 
 def get_param_store():
-    warnings.warn('NumPyro does not have a parameter store. '
-                  'Value of SVI parameters can be obtained via '
-                  'SVI.get_params() method.',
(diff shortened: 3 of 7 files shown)


numpyro/compat/pyro.py
```diff
@@ -1,13 +1,33 @@
...
+    return _PARAM_STORE
+
+
+def clear_param_store():
+    return _PARAM_STORE.clear()
+
+
+def param(name, *args, **kwargs):
+    # Get value assuming statement is wrapped in a substitute handler.
+    val = _param(name, *args, **kwargs)
```
fehiepsi: Look reasonable to me!

numpyro/compat/infer.py
```diff
@@ -131,6 +132,8 @@ def step(self, *args, rng=None, **kwargs):
                                 'dicts of arrays.')
             else:
                 raise e
+        params = super(SVI, self).get_params(self.svi_state)
```
fehiepsi: I think we should `jit` this.

numpyro/compat/ops.py
```diff
@@ -1,3 +1,34 @@
...
+
+def ones(*sizes, **kwargs):
+    if len(sizes) == 0:
+        raise ValueError('Positional `size` argument not provided.')
+    elif len(sizes) == 1:
+        if isinstance(sizes[0], (tuple, list)):
+            sizes = sizes[0]
+    if not all([isinstance(s, int) for s in sizes]):
+        raise ValueError('Invalid data type for `size` provided.')
+    return np.ones(sizes, **kwargs)
```
fehiepsi: lgtm

numpyro/optim.py
```diff
@@ -75,6 +77,27 @@ def __init__(self, *args, **kwargs):
...
+class ClippedAdam(_NumpyroOptim):
+    """
+    :class:`~numpyro.optim.Adam` optimizer with gradient clipping.
+
+    **Reference:**
+
+    `A Method for Stochastic Optimization`, Diederik P. Kingma, Jimmy Ba
+    https://arxiv.org/abs/1412.6980
+    """
+    def __init__(self, *args, **kwargs):
```
fehiepsi: I think we should expose `clip_norm` here and add docs for it.

numpyro/compat/ops.py
```diff
@@ -1,3 +1,34 @@
...
+
+import jax.numpy as np
 from jax.numpy import *  # noqa: F401, F403
 
 tensor = array  # noqa: F405
+
+randn = onp.random.randn
+
+
+# Provide wrappers to initialize ones/zeros using the pytorch convention
```
fritzo: I think it's fine to restrict to numpy-style usage in pyro-api, but if you want to maintain these that's fine too. Let me know either way, and we can switch the pyroapi.tests to use numpy-style invocations.

numpyro/optim.py
```diff
@@ -75,6 +77,27 @@ def __init__(self, *args, **kwargs):
         super(Adam, self).__init__(optimizers.adam, *args, **kwargs)
 
 
+class ClippedAdam(_NumpyroOptim):
```
fritzo: Note Martin's `ClippedAdam` also includes learning rate decay. We don't need to implement that now, but can you make sure this errors if an `lrd` kwarg is found?

neerajprad: I think we can support lrd too, because we can provide a schedule for the learning rate to the jax optimizer. I will try to include that or error if it turns out to be more complicated.

numpyro/compat/infer.py
neerajprad: Sure, will do that. Do you think this will have significant overhead?

numpyro/compat/ops.py
neerajprad: I don't mind maintaining these wrappers, my hope is that this will be relatively hassle free, lets see. :smile: There is a bit of an inconsistency in the numpy API too, because `torch.random.randn()` doesn't accept a tuple size arg, so this provides a more consistent interface.

numpyro/compat/infer.py
fehiepsi: Yes, especially for models with many constrained supports.

numpyro/optim.py
neerajprad: done.

numpyro/optim.py
neerajprad: Done.

numpyro/compat/infer.py
neerajprad: Yeah, that's a good point. I wish there was a way to signal the end of learning with Pyro's interface. Pyro's param store will do this lazily, but we'll be inefficiently doing this per `svi.step`. :(

fritzo: Can you explain for me how this is less efficient than Pyro? Even in Pyro constrained params must be transformed once per svi step.

fehiepsi: I meant this will have significant overhead when we didn't jit it. In NumPyro, params are already transformed to compute elbo. The transform here is not necessary unless we want to access param values from the param store. But I think this overhead is small because we already jit it.

neerajprad: Actually, I think the only difference is that pyro will compute and cache this when it calls `pyro.param` whereas we will be doing this at least twice (once in inference and then again to put the parameters in the param store). But hopefully, with JIT, this shouldn't be as much of an issue.

fehiepsi: @neerajprad How about we resolve this by changing [this line](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/infer/svi.py#L205) to
```
def loss_fn(x):
     constrained_params = self.constrain_fn(x)
     return self.loss.loss(rng_seed, constrained_params, ...), constrained_params

(loss, constrained_params), grad = value_and_grad(loss_fn, has_aux=True)(params)
```
Basically, we use `has_aux=True` for constrained_params and move it to `svi_state`.

neerajprad: That sounds reasonable. You mean, add another field called `constrained_params` to `SVIState`?

neerajprad:
@fehiepsi - I will address your suggestion regarding the optimization in a separate PR, if that's okay?

Issue #387, opened 2019-10-06
Low-rank Multivariate Normal distribution


fehiepsi:
This is a multivariate normal distribution with covariance matrix is parameterized by: `cov = W @ W.T + D` where `W` is a low-rank matrix, `D` is a diagonal matrix. Taking advantage of low-rank form and some algebra tricks, it is much faster to sample from or compute log probability of this distribution (comparing to MultivariateNormal distribution). If you love matrices, you might love to solve this issue. ;)

### Applications
+ This can replace MultivariateNormal priors in your model.
+ This can be used to implement a fast multivariate autoguide, which is very useful for models high dimensional latent variables.
+ Under the hood, Spare regression Gaussian Process models use this distribution as prior.

### How to implement?
+ [PyTorch implementation](https://pytorch.org/docs/stable/_modules/torch/distributions/lowrank_multivariate_normal.html) has detailed comments on the math of each method, so the best way is to mimic PyTorch implementation.
+ https://en.wikipedia.org/wiki/Woodbury_matrix_identity
+ https://en.wikipedia.org/wiki/Matrix_determinant_lemma

### What I can take away when solving this issue?
+ Embrace your linear algebra knowledge.
+ Familiarize yourself with JAX [numpy.linalg](https://jax.readthedocs.io/en/latest/jax.numpy.html#module-jax.numpy.linalg), [scipy.linalg](https://jax.readthedocs.io/en/latest/jax.scipy.html#module-jax.scipy.linalg) operators.
+ Know how to implement a distribution in NumPyro.

As usual, please comment below any related question that you have. We are happy to help you go through this issue. :)

TuanNguyen27:
I would like to implement this! Is this the right place to put the code ? https://github.com/pyro-ppl/numpyro/blob/master/numpyro/distributions/continuous.py

fehiepsi:
@TuanNguyen27 Yes, that's the place to put the code.

It's great to hear that you want to implement this distribution! Please let me know if you have any questions. FYI (in case you follow PyTorch implementation), we do not support the `.expand` method so you don't need to worry about `_unbroadcasted` stuff. :)

PR #403, opened 2019-10-18
Implement Low-rank MVN

TuanNguyen27:
This PR implements Low-rank Multivariate Normal Distribution, following [PyTorch implementation](https://pytorch.org/docs/stable/_modules/torch/distributions/lowrank_multivariate_normal.html) ref issue #387


TuanNguyen27:
@fehiepsi please review! I expect a lot of rough edges, as this is my first attempt using jax :)

fehiepsi:
@TuanNguyen27 Awesome!! I will review this soon.

numpyro/distributions/continuous.py
```diff
@@ -651,6 +651,156 @@ def variance(self):
...
+    return np.squeeze(np.matmul(bmat, np.expand_dims(bvec, axis=-1)), axis=-1)
+
+
+def _batch_capacitance_tril(W, D):
+    r"""
+    Computes Cholesky of :math:`I + W.T @ inv(D) @ W` for a batch of matrices :math:`W`
+    and a batch of vectors :math:`D`.
+    """
+    # is this the right way to specify np.transpose ?
+    Wt_Dinv = np.transpose(W, axes=[-1, -2]) / np.expand_dims(D, -2)
```
fehiepsi: I usually use [swapaxes](https://docs.scipy.org/doc/numpy/reference/generated/numpy.swapaxes.html) for transposing in numpy.

numpyro/distributions/continuous.py
```diff
@@ -651,6 +651,156 @@ def variance(self):
...
+
+
+def _batch_lowrank_logdet(W, D, capacitance_tril):
+    r"""
+    Uses "matrix determinant lemma"::
+        log|W @ W.T + D| = log|C| + log|D|,
+    where :math:`C` is the capacitance matrix :math:`I + W.T @ inv(D) @ W`, to compute
+    the log determinant.
+    """
+    sign_det_D, log_det_D = np.linalg.slogdet(D)
```
fehiepsi: `D` is a diagonal vector, so we can compute `logdet` by `np.log(D).sum(-1)`

numpyro/distributions/continuous.py
```diff
@@ -651,6 +651,156 @@ def variance(self):
...
+    Wt_Dinv_x = _batch_mv(Wt_Dinv, x)
+    mahalanobis_term1 = np.sum(np.square(x) / D, axis=-1)
+    mahalanobis_term2 = _batch_mahalanobis(capacitance_tril, Wt_Dinv_x)
+    return mahalanobis_term1 - mahalanobis_term2
+
+
+@copy_docs_from(Distribution)
+class LowRankMultivariateNormal(Distribution):
+    arg_constraints = {"loc": constraints.real, "cov_factor": constraints.real, "cov_diag": constraints.positive}
+    support = constraints.real
```
fehiepsi: `constraints.real` -> `constraints.real_vector`

numpyro/distributions/continuous.py
```diff
@@ -651,6 +651,156 @@ def variance(self):
...
+    Wt_Dinv = np.transpose(W, axes=[-1, -2]) / np.expand_dims(D, -2)
+    Wt_Dinv_x = _batch_mv(Wt_Dinv, x)
+    mahalanobis_term1 = np.sum(np.square(x) / D, axis=-1)
+    mahalanobis_term2 = _batch_mahalanobis(capacitance_tril, Wt_Dinv_x)
+    return mahalanobis_term1 - mahalanobis_term2
+
+
+@copy_docs_from(Distribution)
+class LowRankMultivariateNormal(Distribution):
+    arg_constraints = {"loc": constraints.real, "cov_factor": constraints.real, "cov_diag": constraints.positive}
```
fehiepsi: not important: "loc": constraints.real -> "loc": constraints.real_vector

numpyro/distributions/continuous.py
```diff
@@ -651,6 +651,156 @@ def variance(self):
...
+        try:
+            loc_, self.cov_factor, cov_diag_ = np.broadcast_arrays(loc_, cov_factor, cov_diag_)
+        except RuntimeError:
+            raise ValueError("Incompatible batch shapes: loc {}, cov_factor {}, cov_diag {}"
+                             .format(loc.shape, cov_factor.shape, cov_diag.shape))
+        self.loc = loc_[..., 0]
+        self.cov_diag = cov_diag_[..., 0]
+        batch_shape = self.loc.shape[:-1]
+        self._unbroadcasted_cov_factor = cov_factor
+        self._unbroadcasted_cov_diag = cov_diag
```
fehiepsi: I think you can just work with `cov_factor` and `cov_diag` directly. This `unbroadcasted` stuff is only useful for `.expand` method, which we don't use in NumPyro.

numpyro/distributions/continuous.py
```diff
@@ -651,6 +651,156 @@ def variance(self):
...
+        if cov_factor.shape[-2:-1] != event_shape:
+            raise ValueError("`cov_factor` must be a batch of matrices with shape {} x m"
+                             .format(event_shape[0]))
+        if cov_diag.shape[-1:] != event_shape:
+            raise ValueError("`cov_diag` must be a batch of vectors with shape {}".format(event_shape))
+
+        loc_ = np.expand_dims(loc, -1)
+        cov_diag_ = np.expand_dims(cov_diag, -1)
+        try:
+            loc_, self.cov_factor, cov_diag_ = np.broadcast_arrays(loc_, cov_factor, cov_diag_)
```
fehiepsi: I think it is better to not broadcast here. You can use [promote_shapes](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/distributions/util.py#L195), which append/prepend `1` to the correct position.
```
loc, cov_factor, cov_diag = promote_shapes(loc[..., np.new_axis], cov_factor, cov_diag[..., np.new_axis])
batch_shape = lax.broadcast_shapes(np.shape(loc), np.shape(cov_factor), np.shape(cov_diag))[:-2]
self.loc = np.broadcast_to(loc[..., 0], batch_shape + event_shape)
self.cov_factor = cov_factor
self.cov_diag = cov_diag
```

numpyro/distributions/continuous.py
```diff
@@ -651,6 +651,156 @@ def variance(self):
...
+        super(LowRankMultivariateNormal, self).__init__(batch_shape, event_shape, validate_args=validate_args)
+
+    @property
+    def mean(self):
+        return self.loc
+
+    @lazy_property
+    def variance(self):
+        # no expand needed here ??
+        return np.sum(np.square(self._unbroadcasted_cov_factor), axis=-1) + self._unbroadcasted_cov_diag
```
fehiepsi: I think we need to broadcast the result to `self.batch_shape + self.event_shape`, use `np.broadcast_to(..., self.batch_shape + self.event_shape)`

numpyro/distributions/continuous.py
```diff
@@ -651,6 +651,156 @@ def variance(self):
...
+        # for Cholesky decomposition (see http://www.gaussianprocess.org/gpml/, Section 3.4.3):
+        #     W @ W.T + D = D1/2 @ (I + D-1/2 @ W @ W.T @ D-1/2) @ D1/2
+        # The matrix "I + D-1/2 @ W @ W.T @ D-1/2" has eigenvalues bounded from below by 1,
+        # hence it is well-conditioned and safe to take Cholesky decomposition.
+        cov_diag_sqrt_unsqueeze = np.expand_dims(np.sqrt(self._unbroadcasted_cov_diag), axis=-1)
+        Dinvsqrt_W = self._unbroadcasted_cov_factor / cov_diag_sqrt_unsqueeze
+        K = np.matmul(Dinvsqrt_W, np.transpose(Dinvsqrt_W, axes=[-1, -2]))
+        K = np.add(K, np.identity(K.shape(-1)))
+        scale_tril = cov_diag_sqrt_unsqueeze * np.linalg.cholesky(K)
+        # again no expand function here ?
```
fehiepsi: yes, I think we don't need to broadcast for distribution-specific properties. But we need to broadcast for general properties such as: `variance`, `mean`, `entropy`.

numpyro/distributions/continuous.py
```diff
@@ -651,6 +651,156 @@ def variance(self):
...
+        covariance_matrix = self._unbroadcasted_cov_diag + np.matmul(self._unbroadcasted_cov_factor,
+                                    np.transpose(self._unbroadcasted_cov_factor, axes=[-1, -2]))
+        return covariance_matrix
+
+    @lazy_property
+    def precision_matrix(self):
+        # We use "Woodbury matrix identity" to take advantage of low rank form::
+        #     inv(W @ W.T + D) = inv(D) - inv(D) @ W @ inv(C) @ W.T @ inv(D)
+        # where :math:`C` is the capacitance matrix.
+        Wt_Dinv = (np.transpose(self._unbroadcasted_cov_factor, axes=[-1, -2])
```
fehiepsi: ditto: use `np.swap_axes` here is better I guess. NumPy transpose behavior is different from PyTorch.

numpyro/distributions/continuous.py
```diff
@@ -651,6 +651,156 @@ def variance(self):
...
+            self._validate_sample(value)
+        diff = value - self.loc
+        M = _batch_lowrank_mahalanobis(self._unbroadcasted_cov_factor,
+                                       self._unbroadcasted_cov_diag,
+                                       diff,
+                                       self._capacitance_tril)
+        log_det = _batch_lowrank_logdet(self._unbroadcasted_cov_factor,
+                                        self._unbroadcasted_cov_diag,
+                                        self._capacitance_tril)
+        return -0.5 * (self._event_shape[0] * np.log(2 * np.pi) + log_det + M)
```
fehiepsi: nit: it is better to use `self.loc.shape[-1]` for event size. We don't use `self._event_shape` in NumPyro.

numpyro/distributions/continuous.py
```diff
@@ -651,6 +651,156 @@ def variance(self):
...
+
+    def entropy(self):
+        log_det = _batch_lowrank_logdet(self._unbroadcasted_cov_factor,
+                                        self._unbroadcasted_cov_diag,
+                                        self._capacitance_tril)
+        H = 0.5 * (self._event_shape[0] * (1.0 + np.log(2 * np.pi)) + log_det)
```
fehiepsi: nit: `self._event_shape[0]` -> `self.loc.shape[-1]`

numpyro/distributions/continuous.py
```diff
@@ -651,6 +651,156 @@ def variance(self):
...
+        return H
```
fehiepsi: We need to broadcast here: `np.broadcast_to(H, self.batch_shape)`

fehiepsi:
@TuanNguyen27 Would you mind exposing this distribution to [__init__](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/distributions/__init__.py) file, [docs](https://github.com/pyro-ppl/numpyro/blob/master/docs/source/distributions.rst#multivariatenormal) file, and add tests for this distribution? You would need to make a function similar to [_mvn_to_scipy](https://github.com/pyro-ppl/numpyro/blob/master/test/test_distributions.py#L42):
```
def _lowrank_mvn_to_scipy(loc, cov_factor, cov_diag):
    jax_dist = dist.LowRankMultivariateNormal(loc, cov_factor, cov_diag)
    mean = jax_dist.mean
    cov = jax_dist.covariance_matrix
    return osp.multivariate_normal(mean=mean, cov=cov)
```
, add `dist.LowRankMultivariateNormal: _lowrank_mvn_to_scipy,` to [DIST_MAP](https://github.com/pyro-ppl/numpyro/blob/master/test/test_distributions.py#L49), and add some examples as your choice similar to [CONTINUOUS](https://github.com/pyro-ppl/numpyro/blob/master/test/test_distributions.py#L118) list. I believe the tests should pass. You can test them locally with
```
pytest -vs test/test_distributions.py -k LowRank
```

TuanNguyen27:
@fehiepsi thanks for the detailed reviews! Will address comments soon :)

numpyro/distributions/continuous.py
TuanNguyen27: @fehiepsi curious question for my own understanding: why do we broadcast `self.batch_shape + self.event_shape` here but only broadcast `self.batch_shape` for entropy calculation ?

fehiepsi: [entropy](https://en.wikipedia.org/wiki/Differential_entropy) is computed from the probability density function `log_prob` (which have shape `batch_shape`), hence it will have shape `batch_shape`. `variance` (similar to `mean`) is computed from `sample`s, hence it will have shape `batch_shape + event_shape`. I think this is true for all distributions.

numpyro/distributions/continuous.py
fehiepsi: @TuanNguyen27 you might want to change `support` to `constraints.real_vector`, instead of `cov_factor` (which should have constraint `real_matrix` - but we don't have that constraint yet).

numpyro/distributions/continuous.py
```diff
@@ -651,6 +651,151 @@ def variance(self):
...
+        "loc": constraints.real_vector,
+        "cov_factor": constraints.real,
+        "cov_diag": constraints.positive
+        }
+    support = constraints.real_vector
+
+    def __init__(self, loc, cov_factor, cov_diag, validate_args=None):
+        if loc.ndim < 1:
+            raise ValueError("`loc` must be at least one-dimensional.")
+        self.event_shape = loc.shape[-1:]
```
TuanNguyen27: @fehiepsi i just realized `loc.shape[-1:]` returns a tuple instead of an int (i was copying pytorch's code too closely). I think it should be `loc.shape[-1]` instead, like how we use it later on in the code, is this correct ?

TuanNguyen27:
> @TuanNguyen27 Would you mind exposing this distribution to [**init**](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/distributions/__init__.py) file, [docs](https://github.com/pyro-ppl/numpyro/blob/master/docs/source/distributions.rst#multivariatenormal) file, and add tests for this distribution? You would need to make a function similar to [_mvn_to_scipy](https://github.com/pyro-ppl/numpyro/blob/master/test/test_distributions.py#L42):
> 
> ```
> def _lowrank_mvn_to_scipy(loc, cov_factor, cov_diag):
>     jax_dist = dist.LowRankMultivariateNormal(loc, cov_factor, cov_diag)
>     mean = jax_dist.mean
>     cov = jax_dist.covariance_matrix
>     return osp.multivariate_normal(mean=mean, cov=cov)
> ```
> 
> , add `dist.LowRankMultivariateNormal: _lowrank_mvn_to_scipy,` to [DIST_MAP](https://github.com/pyro-ppl/numpyro/blob/master/test/test_distributions.py#L49), and add some examples as your choice similar to [CONTINUOUS](https://github.com/pyro-ppl/numpyro/blob/master/test/test_distributions.py#L118) list. I believe the tests should pass. You can test them locally with
> 
> ```
> pytest -vs test/test_distributions.py -k LowRank
> ```

I'm having trouble running the tests locally: 
```
E   ImportError: cannot import name 'transforms' from 'numpyro.distributions' (/Users/nguyen.tuan/anaconda3/lib/python3.7/site-packages/numpyro/distributions/__init__.py)
```

numpyro/distributions/continuous.py
```diff
@@ -651,6 +651,151 @@ def variance(self):
...
+class LowRankMultivariateNormal(Distribution):
+    arg_constraints = {
+        "loc": constraints.real_vector,
+        "cov_factor": constraints.real,
+        "cov_diag": constraints.positive
+        }
+    support = constraints.real_vector
+
+    def __init__(self, loc, cov_factor, cov_diag, validate_args=None):
+        if loc.ndim < 1:
```
fehiepsi: nit: in `init` method, using np.ndim(loc) and np.shape(...) is better (they work for scalar numbers, which does not have attributes `.ndim`, `.shape`).

numpyro/distributions/continuous.py
fehiepsi: @TuanNguyen27  `loc.shape[-1:]` is the correct event_shape. The current convention is to define
```
event_shape = loc.shape[-1:]
...
batch_shape = lax.broadcast_shapes(...)
...
super(LowRankMultivariateNormal, self).__init__(batch_shape, event_shape, validate_args=validate_args)
```

> loc.shape[-1] instead, like how we use it later on in the code

Yes, later in the code, we use `loc.shape[-1]` to get integer values (e.g. to compute log_prob).

test/test_distributions.py
```diff
@@ -116,6 +124,9 @@ def _mvn_to_scipy(loc, cov, prec, tril):
     T(dist.MultivariateNormal, np.arange(6, dtype=np.float32).reshape((3, 2)), None, None,
       np.array([[1., 0.], [0., 1.]])),
     T(dist.MultivariateNormal, 0., None, np.broadcast_to(np.identity(3), (2, 3, 3)), None),
+    T(dist.LowRankMultivariateNormal, np.zeros(2), np.array([1, 0]), np.diag(np.array([1, 1]))),
```
fehiepsi: `cov_diag` is a vector, so I think `np.array([1, 1])` is enough.

`cov_factor` is a matrix, so I guess you will want to define it as: `np.array([[1], [0]])`

test/test_distributions.py
```diff
@@ -116,6 +124,9 @@ def _mvn_to_scipy(loc, cov, prec, tril):
...
+    T(dist.LowRankMultivariateNormal, np.arange(6, dtype=np.float32).reshape((2, 3)),
+      np.arange(12, dtype=np.float32).reshape((3, 4)), np.diag(np.array([1, 2, 3]))),
```
fehiepsi: Could you change `cov_factor` to something like `np.arange(6, dtype=np.float32).reshape((3, 2))` so it is indeed "low rank" (though full rank still works, it won't give us the benefit of low-rank algebra tricks). :D

fehiepsi:
> I'm having trouble running the tests locally:

Could you try to uninstall numpyro and install again with `pip install -e .` instead?

numpyro/distributions/continuous.py
```diff
@@ -651,6 +651,151 @@ def variance(self):
...
+
+def _batch_capacitance_tril(W, D):
+    r"""
+    Computes Cholesky of :math:`I + W.T @ inv(D) @ W` for a batch of matrices :math:`W`
+    and a batch of vectors :math:`D`.
+    """
+    Wt_Dinv = np.swapaxes(W, -1, -2) / np.expand_dims(D, -2)
+    K = np.matmul(Wt_Dinv, W)
+    # could be inefficient
+    return np.linalg.cholesky(np.add(K, np.identity(K.shape(-1))))
```
fehiepsi: typo: K.shape(-1) -> K.shape[-1]

numpyro/distributions/continuous.py
```diff
@@ -651,6 +651,151 @@ def variance(self):
...
+            raise ValueError("`cov_factor` must be at least two-dimensional, "
+                             "with optional leading batch dimensions")
+        if np.shape(cov_factor)[-2:-1] != event_shape:
+            raise ValueError("`cov_factor` must be a batch of matrices with shape {} x m"
+                             .format(event_shape[0]))
+        if np.shape(cov_diag)[-1:] != event_shape:
+            raise ValueError("`cov_diag` must be a batch of vectors with shape {}".format(self.event_shape))
+
+        loc, cov_factor, cov_diag = promote_shapes(loc[..., np.newaxis], cov_factor, cov_diag[..., np.newaxis])
+        batch_shape = lax.broadcast_shapes(np.shape(loc), np.shape(cov_factor), np.shape(cov_diag))[:-2]
```
fehiepsi: sorry, my last code snippet missed a line after this
```
cov_diag = cov_diag[..., 0]
```

numpyro/distributions/continuous.py
```diff
@@ -651,6 +651,151 @@ def variance(self):
...
+        self.loc = np.broadcast_to(loc[..., 0], batch_shape + event_shape)
+        self.cov_factor = cov_factor
+        self.cov_diag = cov_diag
+        self._capacitance_tril = _batch_capacitance_tril(cov_factor, cov_diag)
+        super(LowRankMultivariateNormal, self).__init__(batch_shape = batch_shape, event_shape = event_shape, validate_args=validate_args)
```
fehiepsi: nit: no space around "=". I think you can install `pip install flake8` and use `flake8` to catch lint errors locally.

numpyro/distributions/continuous.py
```diff
@@ -651,6 +651,151 @@ def variance(self):
...
+        cov_diag_sqrt_unsqueeze = np.expand_dims(np.sqrt(self.cov_diag), axis=-1)
+        Dinvsqrt_W = self.cov_factor / cov_diag_sqrt_unsqueeze
+        K = np.matmul(Dinvsqrt_W, np.swapaxes(Dinvsqrt_W, -1, -2))
+        K = np.add(K, np.identity(K.shape(-1)))
+        scale_tril = cov_diag_sqrt_unsqueeze * np.linalg.cholesky(K)
+        return scale_tril
+
+    @lazy_property
+    def covariance_matrix(self):
+        covariance_matrix = self.cov_diag + np.matmul(
```
fehiepsi: Sorry, I remember that you asked how to do diag_embed but I forgot to comment. For now, you can use
```
# TODO: find a better solution to create a diagonal matrix
self.cov_diag[..., np.newaxis] * np.identity(self.loc.shape[-1])
```

numpyro/distributions/continuous.py
```diff
@@ -651,6 +651,151 @@ def variance(self):
...
+        # We use "Woodbury matrix identity" to take advantage of low rank form::
+        #     inv(W @ W.T + D) = inv(D) - inv(D) @ W @ inv(C) @ W.T @ inv(D)
+        # where :math:`C` is the capacitance matrix.
+        Wt_Dinv = (np.swapaxes(self.cov_factor, -1, -2)
+                   / np.expand_dims(self.cov_diag, axis=-2))
+        A = solve_triangular(Wt_Dinv, self._capacitance_tril, lower=True)
+        return np.reciprocal(self.cov_diag) - np.matmul(np.swapaxes(A, -1, -2), A)
+
+    def sample(self, key, sample_shape=()):
+        W_shape = sample_shape + self.cov_factor.shape[-1:]
```
fehiepsi: I think to get a correct out shape, we need to use
```
batch_shape = sample_shape + self.batch_shape
W_shape = batch_shape + self.cov_factor.shape[-1:]
D_shape = batch_shape + self.cov_diag.shape[-1:]
eps_W = random.normal(key, W_shape)
eps_D = random.normal(key, D_shape)
```

numpyro/distributions/continuous.py
```diff
@@ -651,6 +651,151 @@ def variance(self):
...
+        eps_W = random.normal(key, W_shape)
+        eps_D = random.normal(key, sample_shape)
+        return (self.loc + _batch_mv(self.cov_factor, eps_W)
+                + np.sqrt(self.cov_diag) * eps_D)
+
+    def log_prob(self, value):
+        if self._validate_args:
+            self._validate_sample(value)
```
fehiepsi: nit: it is better to use
```
@validate_sample
def log_prob(self, value):
    diff = ...
```
so the output log_prob is correctly masked for out-of-support samples.

numpyro/distributions/continuous.py
TuanNguyen27: @fehiepsi There's another `diag_embed` for  `precision_matrix`, i suppose i'll need to do the same thing there?

numpyro/distributions/continuous.py
TuanNguyen27: Do you mean, `self.cov_diag = cov_diag[..., 0]` ? But i think you mentioned earlier that cov_diag is usually a vector (since it represents a diagonal matrix) -- so what's the intention behind slicing for the first column here ? (just out of curiosity)

fehiepsi: Ah, because we added a new axis `cov_diag[..., np.newaxis]` above (to promote_shapes), we drop it here. ^^

numpyro/distributions/continuous.py
```diff
@@ -651,6 +651,160 @@ def variance(self):
...
+        inverse_cov_diag = np.reciprocal(self.cov_diag)
+        diag_embed = inverse_cov_diag[..., np.newaxis] * np.identity(self.loc.shape[-1])
+        return diag_embed - np.matmul(np.swapaxes(A, -1, -2), A)
+
+    def sample(self, key, sample_shape=()):
+        batch_shape = sample_shape + self.batch_shape
+        W_shape = batch_shape + self.cov_factor.shape[-1:]
+        D_shape = batch_shape + self.cov_diag.shape[-1:]
+        eps_W = random.normal(key, W_shape)
+        eps_D = random.normal(key, sample_shape)
```
fehiepsi: typo: this is `D_shape`.

numpyro/distributions/continuous.py
fehiepsi: @TuanNguyen27 You don't need those two lines `if self._validate...` when we use the decorator `validate_sample`.

numpyro/distributions/continuous.py
fehiepsi: Yes, you are right!

numpyro/distributions/continuous.py
fehiepsi: Probably with this change and the lint issues are fixed, all tests will pass.

Cheers!

TuanNguyen27:
> LGTM if tests pass. Thanks for working on this great PR @TuanNguyen27 !

@fehiepsi actually the tests are failing due to some shape issues :(

numpyro/distributions/continuous.py
fehiepsi: @TuanNguyen27 I think the issue happens at this line. It is better to do `cov_diag = cov_diag[..., 0]` here, rather than in `self.cov_diag = cov_diag[..., 0]` because we use `cov_diag` to compute _batch_capacitance_tril.

numpyro/distributions/continuous.py
```diff
@@ -651,6 +651,159 @@ def variance(self):
...
+                             .format(event_shape[0]))
+        if np.shape(cov_diag)[-1:] != event_shape:
+            raise ValueError("`cov_diag` must be a batch of vectors with shape {}".format(self.event_shape))
+
+        loc, cov_factor, cov_diag = promote_shapes(loc[..., np.newaxis], cov_factor, cov_diag[..., np.newaxis])
+        batch_shape = lax.broadcast_shapes(np.shape(loc), np.shape(cov_factor), np.shape(cov_diag))[:-2]
+        self.loc = np.broadcast_to(loc[..., 0], batch_shape + event_shape)
+        self.cov_factor = cov_factor
+        self.cov_diag = cov_diag[..., 0]
+        self._capacitance_tril = _batch_capacitance_tril(cov_factor, cov_diag)
```
fehiepsi: Otherwise, you can use `self.cov_diag` here instead of `cov_diag`.

numpyro/distributions/continuous.py
```diff
@@ -651,6 +651,159 @@ def variance(self):
...
+        inverse_cov_diag = np.reciprocal(self.cov_diag)
+        diag_embed = inverse_cov_diag[..., np.newaxis] * np.identity(self.loc.shape[-1])
+        return diag_embed - np.matmul(np.swapaxes(A, -1, -2), A)
+
+    def sample(self, key, sample_shape=()):
+        batch_shape = sample_shape + self.batch_shape
+        W_shape = batch_shape + self.cov_factor.shape[-1:]
+        D_shape = batch_shape + self.cov_diag.shape[-1:]
+        eps_W = random.normal(key, W_shape)
+        eps_D = random.normal(key, D_shape)
```
fehiepsi: Sorry, I missed this duplicated-random-state issue. Could you fix this with
```
key_W, key_D = random.split(key)
eps_W = random.normal(key_W, W_shape)
eps_D = random.normal(key_D, D_shape)
```

TuanNguyen27:
@fehiepsi thanks for the fix! i think tests are passing now :) Another quick q: i didn't add test cases where `loc` or `cov_diag` or `cov_factor` could be `None`. Is that necessary?

fehiepsi:
Re None arg: It is not necessary because we didn’t set those args to be None by default.

Issue #402, opened 2019-10-18
Implement several important discrete distributions


fehiepsi:
To name a few:

- [x] [BetaBinomial](https://github.com/pyro-ppl/pyro/blob/dev/pyro/distributions/conjugate.py)
- [x] [GammaPoisson](https://github.com/pyro-ppl/pyro/blob/dev/pyro/distributions/conjugate.py)
- [x] [ZeroInflatedPoisson](https://github.com/pyro-ppl/pyro/blob/dev/pyro/distributions/zero_inflated_poisson.py)
- [x] [OrderedLogit](https://en.wikipedia.org/wiki/Ordered_logit) which is required for ordinal regression

PR #406, opened 2019-10-19
add ordered logistic distribution

fehiepsi:
This together with #405 solves #402.

This distribution is useful for problems with ordered categorical outcomes: https://en.wikipedia.org/wiki/Ordinal_regression

References:

+ https://mc-stan.org/docs/2_20/functions-reference/ordered-logistic-distribution.html
+ https://mc-stan.org/docs/2_20/reference-manual/ordered-vector.html

### DIFF
--- a/docs/source/distributions.rst
+++ b/docs/source/distributions.rst
@@ -281,6 +281,14 @@ MultinomialProbs
     :show-inheritance:
     :member-order: bysource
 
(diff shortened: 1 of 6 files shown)


neerajprad:
It will be nice to support ordinal regression in brmp. cc. @null-a

PR #407, opened 2019-10-23
Remove test_example marker from tests

neerajprad:
I was getting a PytestUnknownMarkerWarning locally, which was getting caught as an error. If we want to use this as a marker, we will need to register it as such in `conftest.py`. I think using a marker is currently overkill, and we can just use the `-k` option instead. If we have multiple stages, then we might need to have a marker. 

I am still not sure why our travis builds were not failing. 

```
    PytestUnknownMarkWarning,
E   pytest.PytestUnknownMarkWarning: Unknown pytest.mark.test_examples - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/latest/mark.html
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,12 +36,12 @@ jobs:
           name: "unit tests"
           python: 3.6
           script:
-              - pytest -vs -m "not test_examples" --deselect test/contrib/test_distributions_contrib.py --durations=100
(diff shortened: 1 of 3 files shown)


fehiepsi:
> we can just use the -k option instead

Agreed! :D

PR #408, opened 2019-10-24
Rename rng argument throughout the codebase

### LINKED ISSUE #377: Rename rng -> rng_key
I think this was a case of sloppy naming that we began with, and has percolated throughout the code base. Since we only ever pass the `PRNGKey`, the argument name should reflect that. Thanks to @fritzo for calling this out while reviewing the generic PR in Pyro.

neerajprad:
Fixes #377. Due to how widely this is used I have deprecated the `rng` argument name currently, and we can remove it in the following release. This was an opportunity to remove our compatibility wrapper and use `rng_seed` instead.


numpyro/util.py
```diff
@@ -19,14 +19,14 @@
 _DISABLE_CONTROL_FLOW_PRIM = False
 
 
-def set_rng_seed(rng_seed):
+def set_rng_key_seed(rng_key_seed):
```
fehiepsi: This utility does not relevant to `rng_key` so it is better to keep the old version.

examples/stochastic_volatility.py
```diff
@@ -80,7 +80,7 @@ def main(args):
     parser.add_argument('-n', '--num-samples', nargs='?', default=3000, type=int)
     parser.add_argument('--num-warmup', nargs='?', default=1500, type=int)
     parser.add_argument('--device', default='cpu', type=str, help='use "cpu" or "gpu".')
-    parser.add_argument('--rng', default=21, type=int, help='random number generator seed')
+    parser.add_argument('--rng_key', default=21, type=int, help='random number generator seed')
```
fehiepsi: nit: `rng_key` -> `rng-seed`

numpyro/handlers.py
```diff
@@ -333,27 +334,30 @@ class seed(Messenger):
...
+       >>> y = seed(model, rng_key=1)()
        >>> assert x == y
     """
-    def __init__(self, fn=None, rng=None):
-        if isinstance(rng, int) or np.size(rng) == 1:
-            rng = random.PRNGKey(rng)
-        self.rng = rng
+    def __init__(self, fn=None, rng_key=None, rng=None):
+        if rng is not None:
+            warnings.warn('`rng` argument is deprecated and renamed to `rng_key_key` instead.', DeprecationWarning)
```
fehiepsi: typo: `rng_key_key` -> `rng_key`

numpyro/handlers.py
```diff
@@ -333,27 +334,30 @@ class seed(Messenger):
     .. doctest::
 
        >>> # as context manager
-       >>> with handlers.seed(rng=1):
+       >>> with handlers.seed(rng_key=1):
```
fehiepsi: This will probably make users confused. Probably we want to keep `rng` as int and `rng_key` as PRNGKey?

neerajprad: My plan was removing `rng` argument entirely, and not keep two versions. Maybe we should actually keep this `rng_seed` in that case, and the remaining arguments where we use a `PRNGKey` as `rng_key`?

numpyro/util.py
neerajprad: good catch, will fix.

numpyro/handlers.py
neerajprad: Changed this; `rng` argument is only there for backwards compatibility reasons but is deprecated.

numpyro/handlers.py
```diff
@@ -334,23 +338,25 @@ class seed(Messenger):
...
     """
-    def __init__(self, fn=None, rng_key=None, rng=None):
+    def __init__(self, fn=None, rng_seed=None, rng=None):
         if rng is not None:
-            warnings.warn('`rng` argument is deprecated and renamed to `rng_key_key` instead.', DeprecationWarning)
-            rng_key = rng
-        if isinstance(rng_key, int) or np.size(rng_key) == 1:
-            rng_key = random.PRNGKey(rng_key)
-        self.rng_key = rng_key
+            warnings.warn('`rng` argument is deprecated and renamed to `rng_key` instead.', DeprecationWarning)
```
fehiepsi: typo: `rng_key` -> `rng_seed`.

numpyro/handlers.py
fehiepsi: That's fine with me.

neerajprad:
Should be good to merge.

Issue #398, opened 2019-10-12
Support `parallel` argument to predictive


fehiepsi:
In Pyro MCMC predictive, we have a `parallel` keyword to resolve the plate issue https://github.com/pyro-ppl/pyro/issues/1995. JAX supports batching out of the box. However,
using vmap for a large num_samples and a large model will consume [much more memory](https://github.com/google/jax/issues/1113). I faced memory issue in training bnaf for covtype dataset and using `lax.map` solves the problem.

Probably we should change the keyword to `sequential` to avoid confusion.

neerajprad:
That's a good idea, we should have this.

PR #409, opened 2019-10-24
add parallel arg to predictive

fehiepsi:
Addresses #398.

TODO:
- [x] support predictive with guide/params.


numpyro/infer/util.py
```diff
@@ -416,7 +416,7 @@ def single_chain_init(key):
     return init_params, potential_fn, constrain_fun
 
 
-def predictive(rng, model, posterior_samples, *args, num_samples=None, return_sites=None, **kwargs):
+def predictive(rng, model, posterior_samples, *args, num_samples=None, return_sites=None, parallel=True, **kwargs):
```
neerajprad: What do you think about changing this to a class interface like Pyro? I think we should try to get as much of the interface related changes as possible since these are the most difficult to change in later releases.

fehiepsi: Hmm, I will do it. It will be a good chance to make `predictive` work for `SVI` too.

numpyro/infer/util.py
```diff
@@ -416,49 +415,82 @@ def single_chain_init(key):
...
+        self.posterior_samples = {} if posterior_samples is None else posterior_samples
+        self.num_samples = num_samples
+        self.guide = guide
+        self.params = params
+        self.return_sites = return_sites
+        self.parallel = parallel
+
+    def _single_prediction(self, val, model_args=(), model_kwargs={}):
+        rng_key, samples = val
+        model_trace = trace(seed(condition(self.model, samples), rng_key)).get_trace(
```
fehiepsi: @neerajprad I would like to make this work for improper prior so probably `substitute` is better used here. WDYT?

neerajprad: Sure, could you clarify what will be the difference in behavior?

fehiepsi: [condition](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/handlers.py#L274) only affects sample sites but I would like to condition on num_samples values (obtained from MCMC) of a `param` site too, so `substitute` would be better used here.

numpyro/infer/util.py
```diff
@@ -416,49 +415,102 @@ def single_chain_init(key):
...
-    :param args: model arguments.
+    :param callable guide: optional guide to get posterior samples of sites not present
+        in `posterior_samples`.
+    :param dict params: dictionary of values for param sites of model/guide.
+    :param int num_samples: number of samples
     :param list return_sites: sites to return; by default only sample sites not present
         in `posterior_samples` are returned.
-    :param int num_samples: number of samples
-    :param kwargs: model kwargs.
+    :param bool parallel: whether to predict in parallel using JAX vectorized map :func:`jax.vmap`.
```
neerajprad: Do you think we should make this `True` by default? How often have you run into memory issues with `vmap` used for doing predictions?

test/test_infer_util.py
```diff
@@ -40,26 +41,74 @@ def model(data=None):
...
+    predictive.return_sites = ["beta", "obs"]
+    predictive_samples = predictive.get_samples(random.PRNGKey(1))
     # check shapes
     assert predictive_samples["beta"].shape == (100,) + true_probs.shape
     assert predictive_samples["obs"].shape == (100,) + data.shape
     # check sample mean
     assert_allclose(predictive_samples["obs"].reshape((-1,) + true_probs.shape).mean(0), true_probs, rtol=0.1)
 
 
+def test_predictive_with_guide():
```
neerajprad: We should add a test with an autoguide. I am not fully sure if this will work correctly with autoguides, and we may need something like `Predictive(model, poutine.block(guide, hide=[list of obs sites])`.

numpyro/infer/util.py
fehiepsi: This is already True by default. :D

neerajprad: Sorry, I meant `False` (I think the corresponding argument in Pyro is `sequential` which is the opposite of what we have here). Sequential will be slow but will almost always return correct results instead of running out of memory.

test/test_infer_util.py
fehiepsi: For autoguide, I used its `sample_posterior` method without any problem yet. This interface (Predictive with `guide`) might not generate correct sample sites (because AutoGuide's `guide` generates samples in base distribution's domain). >"<

numpyro/infer/util.py
fehiepsi: I only face this problem in covtype example. It would be fine to change the default value to False because either way is fast. :D

test/test_infer_util.py
neerajprad: Hmm..that's not ideal. We should at the very least document that, or try to make the utility function work in either case (a hacky way would be to check if the guide is an Autoguide and call `sample_posterior` but its annoying to have utility function take on a contrib dependency).

numpyro/infer/util.py
neerajprad: If its fast, let us change it to `False` then to preserve the same behavior as in Pyro.

fehiepsi: Yeah, I will do it before the release. :)

test/test_infer_util.py
fehiepsi: I think we should add a note to AutoContinuous docs.

PR #412, opened 2019-10-25
skip numpy warning error for constant values

fehiepsi:
For some priors such as LKJCholesky, the upper diagonal matrix will always be constant. When we print the summary, a numpy warning will be triggered due to (0 / 0) issue. This PR skips those warnings (this should be fine because we still have `NaN` displayed).

A quick test for this issue is:
```
summary(np.ones((2, 10))`
```


fehiepsi:
@neerajprad This should be good to merge (tested with a bunch of models with constant params locally).

PR #421, opened 2019-10-30
Update bayesian regression tutorial

neerajprad:
Tutorial cleanup and updates:
 - Highlight use of `Predictive` class.
 - Remove details about `initialize_model` which is confusing since we aren't using it anywhere.
 - Other minor changes.


notebooks/source/_static/css/pyro.css
```diff
@@ -26,4 +26,8 @@
 
 .wy-side-nav-search>a img.logo, .wy-side-nav-search .wy-dropdown>a img.logo {
     max-width: 60%;
-}
\ No newline at end of file
+}
+
+.wy-nav-content {
+    max-width: 1000px !important;
```
neerajprad: @jpchen - I am increasing the max width since I see many figures / tables which require horizontal scrolling. Should we also do this for Pyro?

notebooks/source/conf.py
```diff
@@ -109,7 +109,7 @@
 # documentation.
 #
 html_theme_options = {
-    'logo_only': True
+    # 'logo_only': True
```
neerajprad: I removed this since without `NumPyro Tutorials` in the nav bar, the Pyro logo looks confusing with a version number of `0.2.0` for NumPyro. Ideally `NumPyro` in the font below the logo and above the version would be much cleaner, but that can wait for later. I also am not sure if there's a way to do this in conf.

notebooks/source/_static/css/pyro.css
jpchen: yeah that sounds reasonable

fehiepsi:
LGTM overall! I just have some comments:
+ I think you can replace
```
dset['AgeScaled'] = (dset.MedianAgeMarriage - onp.mean(dset.MedianAgeMarriage)) / onp.std(dset.MedianAgeMarriage)
```
with
```
standarize = lambda x: (x - x.mean()) / x.std()
dset['AgeScaled'] = dset.MedianAgeMarriage.pipe(standarize)
```
+ In `expected_log_likelihood` section, should we use the [log pointwise predictive density](https://github.com/pyro-ppl/numpyro/blob/master/examples/baseball.py#L146) instead. Basically, it computes `log_mean_exp` first, then take sum over data points. While in the tutorial, we take sum over data points first, then computes `log_mean_exp`.
+ In "Effect of Incorporating Measurement Noise on Residuals" section, the plot is hard to distinguish:
residual vs sd.
+ The gray color and pink color in posterior predictive plots look similar to me. How about pink and black?
+ I think after we explained `predict_fn`, it is better to use `Predictive` again. :)
+ It would be nice to add a sentence mentioning about `log_likelihood` utility in `Model Log Likelihood ` section.

neerajprad:
> In expected_log_likelihood section, should we use the log pointwise predictive density instead. Basically, it computes log_mean_exp first, then take sum over data points. While in the tutorial, we take sum over data points first, then computes log_mean_exp.

It makes sense to use the pointwise predictive density, but I think it will require a bit more explanation perhaps. I'll try to address that. The scale will also be quite different from what we have here which is log of an expectation instead of expectation of a log that the pointwise predictive density give us. 

> I think after we explained predict_fn, it is better to use Predictive again. :)

What do we use Predictive again for though? We have already output the mean predictions above, and I initially had a plot too but its the same plot as the one below which is repetitive. 

Will address other comments too. Thanks for reviewing!

fehiepsi:
> What do we use Predictive again for though?

It is just for convenience. So we don't need to redefine `predict_fn`. But it would be fine if it is your purpose. :)

> log of an expectation instead of expectation of a log

I think that the scale will be similar, it is just different on the order which we compute empirical result.
+ In log likelihood, we use the formula: log p(X) ~ log[(1/n) sum_i [p(X, z_i)]] = log_mean_exp(sum_x(logp(x, z_i))
+ In log predictive, we use the formula: log p(X) ~ sum_x (logp(x)) = sum_x log_mean_exp(logp(x, z_i))

If we only have 1 single sample `z`, then both are equivalent. When we have many posterior samples, using log predictive is better IMO because information criterions use it. In linear regression, log predictive formula is more aligned to the sum squared estimates of errors to me: that is for each data point, we give prediction (using all posterior samples), then we compute the squared error (which is log_prob of Normal) and take the sum. There, the log probability of each prediction from posterior samples is `log_mean_exp(logp(x, z_i))`.

neerajprad:
> It is just for convenience. So we don't need to redefine predict_fn.

Oh..you mean in the later models, use Predictive directly. Yes, I agree. I think once we have shown how to do it using effect handlers, we don't need to keep doing it that way.

To me, it seems that the formula currently is `log(E(y|\theta))`, and the point-wise formula is `E[log p(y|\theta)]`, where we take the monte carlo expectation over different parameter values sampled from the posterior. Then the former will upper bound the latter (by Jensen's inequality). Is my understanding correct? I might be missing some important detail. The scale might still be pretty similar, but it does seem like a different quantity.

neerajprad:
@fehiepsi - I think I should have address all your comments in the last commit. Let me know if there is anything else.

PR #423, opened 2019-10-30
Tweak docs and change the default value of parallel arg in Predictive

fehiepsi:
This PR exposes `AutoContinuous` to docs and adds a note about the best way to use it in SVI and get posterior samples.

In addition, per discussion, I changed `parallel` arg of Predictive class to False, which is still fast and has less memory requirement.

### DIFF
--- a/docs/source/utilities.rst
+++ b/docs/source/utilities.rst
@@ -19,6 +19,8 @@ set_host_device_count
 ---------------------
 .. autofunction:: numpyro.util.set_host_device_count
 
+.. _inference_utilities:
+
 Inference Utilities
 ===================
 
--- a/docs/source/utilities.rst
+++ b/docs/source/utilities.rst
@@ -19,8 +19,6 @@ set_host_device_count
 ---------------------
 .. autofunction:: numpyro.util.set_host_device_count
 
-.. _inference_utilities:
-
 Inference Utilities
 ===================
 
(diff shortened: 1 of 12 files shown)


docs/source/utilities.rst
```diff
@@ -19,6 +19,8 @@ set_host_device_count
 ---------------------
 .. autofunction:: numpyro.util.set_host_device_count
 
+.. _inference_utilities:
```
neerajprad: What does this do?

numpyro/contrib/autoguide/__init__.py
```diff
@@ -99,23 +99,27 @@ class AutoContinuous(AutoGuide):
...
 
     Assumes model structure and latent dimension are fixed, and all latent
     variables are continuous.
 
-    Reference:
+    .. note::
+        We recommend to use :class:`AutoContinuousELBO` as the
+        objective function `loss` in :class:`~numpyro.infer.svi.SVI`.
+        In addition, to get posterior samples from this autoguide, it
+        is better to use the :meth:`sample_posterior` method.
```
neerajprad: nit: consider adding the reason why we recommend doing that. 

"We recommend using `:meth:`sample_posterior method for drawing posterior samples from the autoguide as it will automatically do any unpacking and transformations required to constrain the samples to the domain of the latent sites."

docs/source/utilities.rst
fehiepsi: I want to point to the docs of init strategies in AutoContinuous docs, so I create this label.

neerajprad: I see. It would be nice to have initialization strategies within a separate header actually. Can we do that by using?

```
Initialization Strategies
^^^^^^^^^^^^^^^^^^^^^^^^
```

Then you could just link to that directly.

numpyro/contrib/autoguide/__init__.py
fehiepsi: done

docs/source/utilities.rst
fehiepsi: Thanks! I just made some changes. Is it what you suggested? It looks good to me. :D

neerajprad: Looks great!

Issue #426, opened 2019-10-31
Updates to README for next release


neerajprad:
- [x] Small example to showcase MCMC inference in NumPyro.

 - FAQs:
   - [x] Sampling outside an inference context. 
   - [x] Any other differences from Pyro.
      + improper prior
      + noncenter reparameterization
      + svi step
   - [x] Writing backend agnostic code.

PR #427, opened 2019-10-31
Update README.md

neerajprad:
Partially addresses #426. 

View (rich diff) - https://github.com/pyro-ppl/numpyro/pull/427/files?short_path=04c6e90#diff-04c6e90faac2675aa89e2176d2eec7d8


README.md
```diff
@@ -19,6 +19,113 @@ NumPyro is designed to be *lightweight* and focuses on providing a flexible subs
...
+
+```python
+# Eight Schools example - Non-centered Reparametrization
```
fehiepsi: nit: missing `>>> `

README.md
```diff
@@ -19,6 +19,113 @@ NumPyro is designed to be *lightweight* and focuses on providing a flexible subs
...
+, where `y` are the treatment effects and `sigma` the standard error. We build a hierarchical model for the study where we assume that the group-level parameters `theta` for each school are sampled from a Normal distribution with unknown mean `mu` and standard deviation `tau`, while the observed data are in turn generated from a Normal distribution with mean and standard deviation given by `theta` (true effect) and `sigma`, respectively. This allows us to estimate the population-level parameters `mu` and `tau` by pooling from all the observations, while still allowing for individual variation amongst the schools using the group-level `theta` parameters.
+
+```python
+>>> # Eight Schools example
+    def eight_schools(J, sigma, y=None):
+        mu = numpyro.sample('mu', dist.Normal(0, 5))
+        tau = numpyro.sample('tau', dist.HalfCauchy(5))
+        with numpyro.plate('J', J):
+            theta = numpyro.sample('theta', dist.Normal(mu, tau))
+            numpyro.sample('obs', dist.Normal(theta, sigma), obs=y)
```
fehiepsi: nit: `...` is missing at the beginning of each lines

README.md
```diff
@@ -19,6 +19,113 @@ NumPyro is designed to be *lightweight* and focuses on providing a flexible subs
...
+>>> print("Number of divergences: {}".format(sum(mcmc.get_extra_fields()['diverging'])))
+
+Number of divergences: 0
+```
+
+Now, let us assume that we have a new school for which we have not observed any test scores, but we would like to generate predictions. NumPyro provides a [Predictive](http://num.pyro.ai/en/latest/utilities.html#numpyro.infer.util.Predictive) class for such a purpose. Note that in the absence of any observed data, we simply use the population-level parameters to generate predictions. The `Predictive` utility conditions the unobserved `mu` and `tau` sites to values drawn from the posterior distribution fron our previous MCMC run, and runs the model forward to generate predictions. 
+
+```python
+>>> # New School
+def new_school():
```
fehiepsi: ditto: missing `...` at the beginning of each line

README.md
```diff
@@ -19,6 +19,113 @@ NumPyro is designed to be *lightweight* and focuses on providing a flexible subs
...
+
+```python
+>>> # Eight Schools example - Non-centered Reparametrization
+... def eight_schools_noncentered(J, sigma, y=None):
+...     mu = numpyro.sample('mu', dist.Normal(0, 5))
+...     tau = numpyro.sample('tau', dist.HalfCauchy(5))
+...     with numpyro.plate('J', J):
+...         theta = numpyro.sample('theta', 
+...                                dist.TransformedDistribution(dist.Normal(0., 1.),
```
fehiepsi: Could you confirm if we get the same result with `dist.Normal(np.zeros(J), 1.)`? :D I am a bit curious about it.

neerajprad: We get similar results (good), but it is not exactly the same. I don't know off the top of my head why that's the case though - it seems to me that we should draw the same sequence of random numbers in either case.

neerajprad:
@fehiepsi - Let us merge this, I'm happy to address further comments in follow-up PRs. There are a few more things to add to the readme.

README.md
fehiepsi: Think about this again, I guess it comes from [init_strategy](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/infer/util.py#L202). Let me go deeper into this issue to make the behavior consistent.

Btw, I want to confirm that hmm/covtype benchmark is still good (indeed we get better results) after the change at the terminate condition. :D But the flowhmc work is 5x slower (trees are bigger). Do you think that it is a blocker that we need to look into before the release?

neerajprad: > But the flowhmc work is 5x slower (trees are bigger). Do you think that it is a blocker that we need to look into before the release?

I think flowhmc is mostly for internal research, so I'm not too concerned about that, but we should look into why this results in 5X larger trees. We can release a patch shortly after if you think there are other kinds of models that might be affected.

fehiepsi: I can confirm that the difference is due to precision. The following are the values of 4th element of `theta` in two versions. We can see that after 5 steps, the values in two versions are a bit different.
```
[(-2.2974488735198975, -2.2974488735198975),
 (-2.2974488735198975, -2.2974488735198975),
 (-2.320910930633545, -2.320910930633545),
 (-1.7636312246322632, -1.7636312246322632),
 (3.898336887359619, 3.898348093032837),
 (-2.118842363357544, -2.118828773498535),
 (-12.201387405395508, -12.20230484008789),
 (-9.108250617980957, -9.103801727294922),
 (-9.108250617980957, -9.103801727294922),
 (7.62028694152832, 7.620795249938965),
```

neerajprad: > Probably, (x - 0) / 1 will not be equal to (x - np.zeros(8)) / np.ones(8), or how gradients are numerically accumulated...

Hmm..yeah I think the code generated for the backward pass may be different or the sequencing of broadcasting operations is different. The code generated by the tracer for the two versions is not identical for some reason.

PR #432, opened 2019-11-01
Rename random_state to rng_key in numpyro.sample

### LINKED ISSUE #429: Change random_state to rng_key in numpyro.sample
We used the `random_state` argument name for compatibility with Scipy distributions but we don't need to do that anymore. `rgn_key` will be a more accurate name and match what JAX uses internally.

neerajprad:
Fixes #429.

### DIFF
--- a/numpyro/primitives.py
+++ b/numpyro/primitives.py
@@ -60,7 +60,7 @@ class Messenger(object):
             return self.fn(*args, **kwargs)
 
 
-def sample(name, fn, obs=None, random_state=None, sample_shape=()):
+def sample(name, fn, obs=None, rng_key=None, sample_shape=()):
     """
     Returns a random sample from the stochastic function `fn`. This can have
     additional side effects when wrapped inside effect handlers like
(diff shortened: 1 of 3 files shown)


numpyro/primitives.py
```diff
@@ -60,7 +60,7 @@ def __call__(self, *args, **kwargs):
             return self.fn(*args, **kwargs)
 
 
-def sample(name, fn, obs=None, random_state=None, sample_shape=()):
+def sample(name, fn, obs=None, rng_key=None, sample_shape=()):
```
fehiepsi: Do you think that having `rng_key` here is important when this primitive is meant to be run under handler.seed? If it is, then we can just keep this api. But if it not, we can reorder sample into
```
def sample(name, fn, sample_shape=(), obs=None):
```
which is useful for users who are familiar with NumPyro modelling.
```
x = sample('x', dist.Normal(), (10,))
```
I don't have much opinion about this. Just want to ask for your opinion.

fehiepsi:
@neerajprad feel free to merge for your convenience.

numpyro/primitives.py
neerajprad: I think it is better to provide it, since it keeps things transparent. For instance, it is like the `sample_shape` argument even if users mostly just use the `plate` handler. So users should be able to call `numpyro.sample('x', dist.Normal(0, 1), rng_key=PRNGKey(1))` without an effect handler. Your snippet is still possible in an effect handler, they just need to use `sample_shape=(10,)` which seems clearer. I don't have any particular opinion on the position of the keyword arguments though.

fehiepsi: >  users mostly just use the plate handler

Yeah, I think I should start using this again. Not sure from which time I used `sample_shape` everywhere. :D

Issue #424, opened 2019-10-30
Suggested changes to diagnostics / summary


neerajprad:
- [x] As in Pyro, we should have the summary function split into two components - one that returns a dict object `d[site_name] = {'mean': .., 'std', ..}` and a `print_summary` function which formats this into a table. The idea is that other examples can make use of the summary function and add to the diagnostics if needed.
 - [ ] Output other metadata like when the model is compiling and when it is sampling. This is specially useful when we run multiple chains and it is not clear whether all the time is being taken to compile the model or if it is the sampling part that is taking time. Related #309. 
 - [x] Divergence is useful information and we should record it by default and have it be emitted as part of the `summary` information.

fehiepsi:
Probably we should rename `summary` to `print_summary` before the release?

About the second point, until JAX/XLA supports doing it, we can recommend users to set `num_warmup=1, `num_samples=0` if they are curious.

About divergence information, we can easily support it before the release. Are you working on this?

neerajprad:
None of these are release blockers and can make it into the next minor release. Just the documentation / making sure that the interface is reasonable is important for this release.

fehiepsi:
I moved the description for the remaining issue to #309, which is quite related. Let's close this for now.

PR #433, opened 2019-11-01
add diverging to `extra_fields` by default

fehiepsi:
I think that this release is stable so let's address part of #424 which might lead to interface change.
+ set `extra_fields=('diverging',)` by default
+ rename `summary` to `print_summary`, which should have different meaning when #424 is addressed.

This PR also fixes a bug when model is empty but extra_fields is not empty.

### DIFF
--- a/numpyro/diagnostics.py
+++ b/numpyro/diagnostics.py
@@ -16,7 +16,7 @@ __all__ = [
     'gelman_rubin',
     'hpdi',
     'split_gelman_rubin',
-    'summary',
+    'print_summary',
 ]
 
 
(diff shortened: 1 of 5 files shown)


numpyro/diagnostics.py
```diff
@@ -241,7 +241,7 @@ def summary(samples, prob=0.90, group_by_chain=True):
     header_format = name_format + ' {:>9} {:>9} {:>9} {:>9} {:>9} {:>9} {:>9}'
     columns = ['', 'mean', 'std', 'median', '{:.1f}%'.format(50 * (1 - prob)),
                '{:.1f}%'.format(50 * (1 + prob)), 'n_eff', 'r_hat']
-    print('\n')
+    print()
```
fehiepsi: This already adds a newline. `print('\n')` adds 2 new lines.

PR #435, opened 2019-11-01
Bump up version to 0.2.1

neerajprad:
- Bump up version to 0.2.1
 - Many small edits to the README.

Lets discuss offline before merging this.


README.md
```diff
@@ -85,10 +90,10 @@ The values above 1 for the split Gelman Rubin diagnostic (`r_hat`) indicates tha
 ...                                                             dist.transforms.AffineTransform(mu, tau)))
 ...         numpyro.sample('obs', dist.Normal(theta, sigma), obs=y)
 
->>> nuts_kernel = NUTS(eight_schools)
+>>> nuts_kernel = NUTS(eight_schools_noncentered)
 >>> mcmc = MCMC(nuts_kernel, num_warmup=500, num_samples=1000)
 >>> rng_key = random.PRNGKey(0)
->>> mcmc.run(rng_key, J, sigma, y=y, extra_fields=('diverging',))
+>>> mcmc.run(rng_key, J, sigma, y=y, extra_fields=('potential_energy',))
```
fehiepsi: In my PR, I set the arg `extra_fields=('diverging',)` by default. So could you make a corresponding change in MCMC.run method, which does
```
extra_fields = set(['diverging']).union(set(extra_fields))
```
to match the behavior here?

neerajprad: Oh..I see, I'll do that. 

> Could you also update versions in examples/tutorials too?

Forgot about that; will do.

numpyro/infer/mcmc.py
```diff
@@ -635,7 +635,7 @@ def run(self, rng_key, *args, extra_fields=('diverging',), collect_warmup=False,
                 raise ValueError('`init_params` must have the same leading dimension'
                                  ' as `num_chains`.')
         assert isinstance(extra_fields, (tuple, list))
-        collect_fields = ('z',) + tuple(extra_fields) if 'z' not in extra_fields else extra_fields
+        collect_fields = set(('z', 'diverging') + tuple(extra_fields))
```
fehiepsi: Do we need another wrapper `tuple(...)` to convert this variable to a tuple? I am not sure if the order of `states` in `_single_chain_mcmc` method is aligned well if we use `set`.
```
collect_fn = attrgetter(*collect_fields)
...
states = dict(zip(collect_fields, states))
```

neerajprad: Good point; fixed!

neerajprad:
Ready to merge.

PR #442, opened 2019-11-11
make sure to use new sample_fn for new data

### LINKED ISSUE #441: Avoid recompilation in MCMC.run with new data
This can be important for saving on compilation cost. 

There seems to be some bug in in HMC/NUTS kernels or the MCMC class due to which this doesn't work currently. See. @null-a's [snippet](https://gist.github.com/null-a/fa9e2038630daf104a7ed3821799db57) for details.

fehiepsi:
Addresses the bug in #441.

### DIFF
--- a/numpyro/infer/mcmc.py
+++ b/numpyro/infer/mcmc.py
@@ -431,14 +431,14 @@ class HMC(MCMCKernel):
         )
         if rng_key.ndim == 1:
             init_state = hmc_init_fn(init_params, rng_key)
-            self._sample_fn = sample_fn
         else:
             # XXX it is safe to run hmc_init_fn under vmap despite that hmc_init_fn changes some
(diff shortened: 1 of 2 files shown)


neerajprad:
IIUC..this will still cause the sample function to be recompiled, which takes any benefits from reusing `.run`. Correct me if I'm mistaken? We should definitely merge this as a bug fix, but fixing #441 will require us to solve this without having to recompile in `collect_fn`.

fehiepsi:
Yes, it will recompile. I think it is better to let users decide when they want to recompile, rather than handling it automatically.

fehiepsi:
I think we can use @null-a test. :D

neerajprad:
@fehiepsi - Feel free to merge this at your convenience.

Issue #440, opened 2019-11-11
Error in bnn.py for num-chains > 2


ziatdinovmax:
When I try running the bnn.py example (without any modifications) with num-chains > 2 it throws the following error:
```
jupyter@pytorch-highmem1:~$ python3 bnn.py --num-chains=3
Traceback (most recent call last):
  File "bnn.py", line 145, in <module>
    main(args)
  File "bnn.py", line 105, in main
    samples = run_inference(model, args, rng_key, X, Y, D_H)
  File "bnn.py", line 64, in run_inference
    mcmc.run(rng_key, X, Y, D_H)
  File "/opt/anaconda3/lib/python3.7/site-packages/numpyro/infer/mcmc.py", line 644, in run
    rng_keys = random.split(rng_key, self.num_chains)
  File "/opt/anaconda3/lib/python3.7/site-packages/jax/random.py", line 194, in split
    return _split(key, num)
  File "/opt/anaconda3/lib/python3.7/site-packages/jax/api.py", line 148, in f_jitted
    out = xla.xla_call(flat_fun, *args_flat, device=device, backend=backend)
  File "/opt/anaconda3/lib/python3.7/site-packages/jax/core.py", line 591, in call_bind
    outs = primitive.impl(f, *args, **params)
  File "/opt/anaconda3/lib/python3.7/site-packages/jax/interpreters/xla.py", line 418, in _xla_call_impl
    compiled_fun = _xla_callable(fun, device, backend, *map(abstractify, args))
  File "/opt/anaconda3/lib/python3.7/site-packages/jax/linear_util.py", line 217, in cached_fun
    ans, f_prev = cached_fun_body(f, args)
  File "/opt/anaconda3/lib/python3.7/site-packages/jax/linear_util.py", line 214, in cached_fun_body
    return call(f, *args), f
  File "/opt/anaconda3/lib/python3.7/site-packages/jax/interpreters/xla.py", line 434, in _xla_callable
    jaxpr, (pvals, consts, env) = pe.trace_to_subjaxpr(fun, master, False).call_wrapped(pvals)
  File "/opt/anaconda3/lib/python3.7/site-packages/jax/linear_util.py", line 165, in call_wrapped
    ans = self.f(*args, **dict(self.params, **kwargs))
  File "/opt/anaconda3/lib/python3.7/site-packages/jax/random.py", line 199, in _split
    return lax.reshape(threefry_2x32(key, counts), (num, 2))
[...]

PR #443, opened 2019-11-11
fix bnn example

fehiepsi:
We missed this issue during refactoring I guess. I am not sure why [Travis does not detect `bnn` test](https://travis-ci.com/pyro-ppl/numpyro/jobs/255246963#L814)...

Addresses #440.


neerajprad:
>  I am not sure why Travis does not detect bnn test...

It is probably not running it in parallel since it doesn't have 2 cores?

Issue #438, opened 2019-11-06
More runtime flexibility in the MCMC class


neerajprad:
- [ ] Being able to freeze/start the chain from the last state without doing warmup adaptation.
 - [ ] In some cases, users may want to reuse the adapted `step_size` and `mass_matrix` parameters from the first run and do subsequent runs (say on slightly different datasets) without doing `warmup`.

An alternative would be to keep the `MCMC` class simple, and use the `hmc` function directly to get more control over the runs. I think both the above use cases can be supported by directly using the `hmc` class at the cost of writing some additional boilerplate code.

fehiepsi:
I think that we can solve both issues by introducing an additional arg `inverse_mass_matrix` to HMC.__init__() and store the last state of MCMC.

Having `inverse_mass_matrix` arg will also be helpful for the case users to use AutoDiagonalNormal to learn inverse_mass_matrix (which is much faster than using MCMC warmup).

Storing the last state of MCMC is also useful. We can inspect `step_size`, `inverse_mass_matrix` from it without having to use `extra_fields` to store corresponding values from all transitions.

With the last state, we can start MCMC again with
```
mcmc = MCMC(NUTS(..., step_size=last_state.adapt_state.step_size, inverse_mass_matrix=last_state.adapt_state.inverse_mass_matrix))
mcmc.run(..., init_params=last_state.z)
```

neerajprad:
> WDYT about the above proposal?

This will need us to expose those arguments in `HMC.init` and also the `inverse_mass_matrix` argument in `hmc`. I think doing so for the latter function definitely makes sense since that should have the super-set of all functionality we would like HMC to support. 

For the MCMC class (and HMC/NUTS kernels), I am not fully convinced if it is worth introducing too much complexity at this point, if users can directly use the `hmc` function for more control if they so desire (we should definitely document this better for visibility). For instance, they may want to use the step size and mass matrix from the adaptation phase and the z values from the last run. This can be supported by `MCMC` too, but users will need to set `mcmc.num_warmup=0`, store these parameters from the MCMC run, initialize a new kernel with these parameters which will result in recompilation. With the hmc interface, it is as simple as replacing those values in `HMCState` and just using the sample function.

This may be worth supporting within `MCMC` eventually. Currently, I think that the internals of that class are already quite complex due to having to support chaining using `vmap` etc., and it might be worth refactoring before we add too many other responsibilities on to that class.

fehiepsi:
[...]

PR #444, opened 2019-11-12
expose inverse_mass_matrix to hmc interface

fehiepsi:
Partly addresses #438.

### DIFF
--- a/numpyro/infer/mcmc.py
+++ b/numpyro/infer/mcmc.py
@@ -170,6 +170,7 @@ def hmc(potential_fn, kinetic_fn=None, algo='NUTS'):
     def init_kernel(init_params,
                     num_warmup,
                     step_size=1.0,
+                    inverse_mass_matrix=None,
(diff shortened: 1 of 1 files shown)


numpyro/infer/mcmc.py
```diff
@@ -187,6 +188,8 @@ def init_kernel(init_params,
         :param float step_size: Determines the size of a single step taken by the
             verlet integrator while computing the trajectory using Hamiltonian
             dynamics. If not specified, it will be set to 1.
+        :param numpy.ndarray inverse_mass_matrix: Inverse of the mass matrix.
```
neerajprad: nit: How about rewording this to - "Initial value for inverse mass matrix. This may be adapted during warmup if `adapt_mass_matrix = True`. If no value is specified, then it is initialized to the identity matrix."?

fehiepsi: Thanks, Neeraj!

Issue #449, opened 2019-11-20
[FR] supporting `pyro.factor` primitive


hzyjerry:
Hi, I'm currently doing research on agent modeling, and I'm wondering if the functionalities of `pyro.factor` is in the roadmap of numpyro?

The pyro repo has some [great examples](https://github.com/pyro-ppl/pyro/blob/dev/examples/rsa/generics.py) covering the use case I have in mind.

fehiepsi:
Hi @hzyjerry, we will support that feature soon.

hzyjerry:
Thanks! The latest PR looks great

PR #450, opened 2019-11-20
add factor primitive

fehiepsi:
Addresses #449. Ported from Pyro to here.

The main purpose to have this is to match Pyro API. As in Pyro, this is a little bit cheaper than using `dist.Delta`.

TODO:
- [x] add tests
- [x] update some examples which are using dist.Delta for this purpose.

### DIFF
--- a/numpyro/distributions/distribution.py
+++ b/numpyro/distributions/distribution.py
@@ -26,6 +26,7 @@ from contextlib import contextmanager
 import warnings
 
 import jax.numpy as np
+from jax import lax
 
 from numpyro.distributions.constraints import is_dependent
 from numpyro.distributions.transforms import Transform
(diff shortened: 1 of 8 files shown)


numpyro/distributions/distribution.py
```diff
@@ -386,3 +387,28 @@ def mean(self):
...
+    The unit type has a single value with no data, i.e. ``value.size == 0``.
+
+    This is used for :func:`numpyro.factor` statements.
+    """
+    arg_constraints = {'log_factor': real}
+    support = real
+
+    def __init__(self, log_factor, validate_args=None):
+        batch_shape = np.shape(log_factor)
+        event_shape = (0,)  # This satisfies .size == 0.
```
neerajprad: Could you elaborate on why this 0 sized dim is needed?

fehiepsi: This is to make sure that the sample does not hold any data (only shape is available).

PR #451, opened 2019-11-20
Make it possible to JIT compile model args/kwargs

neerajprad:
Partially addresses #441, by allowing for a `potential_fn_gen` argument to `hmc`, which returns a `potential_fn` when provided model args/kwargs. This does not change the higher-level MCMC/NUTS APIs, but simply provides the functionality to exist in the lower level API. This can be seen in `test_reuse_mcmc_pe_gen`, where the second run is almost free. This requires corresponding changes in `initialize_model`. 

The code can be slightly simplified if we make this the default (i.e. `hmc` taking in only `potential_fn_gen`), but we can do that in the future, if needed. In later PRs, we can start using this in MCMC/NUTS so that once the `sample_kernel` is compiled we can use `MCMC.run` multiple times with other data.

### DIFF
--- a/test/test_mcmc.py
+++ b/test/test_mcmc.py
@@ -445,7 +445,8 @@ def test_functional_map(algo, map_fn):
 
 
 def test_reuse_mcmc_run():
-    y_obs = onp.random.normal(3, 1, (100,))
+    y1 = onp.random.normal(3, 0.1, (100,))
+    y2 = onp.random.normal(-3, 0.1, (100,))
 
     def model(y_obs):
         mu = numpyro.sample('mu', dist.Normal(0., 1.))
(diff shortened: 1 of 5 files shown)


fehiepsi:
Oh, I have thought it would be a bit complicated to change the codebase... Nicely done! I'll review this PR carefully.

test/test_mcmc.py
```diff
@@ -454,9 +455,39 @@ def model(y_obs):
...
+
+    samples = fori_collect(0, 500, _sample, (init_state, y1),
+                           transform=lambda state: constrain_fn(y1)(state[0].z))
+    assert_allclose(samples['mu'].mean(), 3., atol=0.1)
+
+    # Run on data, re-using `mcmc` - this should be much faster.
+    init_state = init_kernel(init_params, num_warmup=300, model_args=(y2,))
+    samples = fori_collect(0, 500, _sample, (init_state, y2),
+                           transform=lambda state: constrain_fn(y2)(state[0].z))
+    assert_allclose(samples['mu'].mean(), -3., atol=0.1)
```
fehiepsi: nice test! this test resolves my wonder!

neerajprad:
Thanks for reviewing, @fehiepsi. 

>  I guess in a single chain, we can also avoid recompiling when changing num_warmup, num_samples with this.

Yes, I think this will be useful whenever users have a single model and they are running it multiple times to draw more samples, do inference on new data, etc. I am not sure, but I think Stan also reuses the compiled model whenever it can.

PR #455, opened 2019-11-21
Misc refactoring for mcmc util functions

neerajprad:
- Factors out the `potential_fn` / `constrain_fn` setup into a distinct function. This is needed for #441. 
 - Removes the `prototype_params` parameter from `find_valid_initial_params`. I think we put this in initially to get some minor speedup, but I don't think it should be needed now. Happy to revert it / explore other alternatives if this seems important.

### DIFF
--- a/test/test_mcmc.py
+++ b/test/test_mcmc.py
@@ -433,7 +433,7 @@ def test_functional_map(algo, map_fn):
     rng_keys = random.split(random.PRNGKey(0), 2)
 
     init_kernel_map = map_fn(lambda init_param, rng_key: init_kernel(
-        init_param, trajectory_length=9, num_warmup=warmup_steps, progbar=False, rng_key=rng_key))
+        init_param, trajectory_length=9, num_warmup=warmup_steps, rng_key=rng_key))
     init_states = init_kernel_map(init_params, rng_keys)
 
     fori_collect_map = map_fn(lambda hmc_state: fori_collect(0, num_samples, sample_kernel, hmc_state,
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,8 @@ jobs:
           script:
               - pytest -vs -k "not test_example" --deselect test/contrib/test_distributions_contrib.py --durations=100
               - JAX_ENABLE_x64=1 pytest -vs test/test_mcmc.py -k x64
-              - XLA_FLAGS="--xla_force_host_platform_device_count=2" pytest -vs test/test_mcmc.py -k chain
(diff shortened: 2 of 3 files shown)


test/test_mcmc.py
```diff
@@ -433,7 +433,7 @@ def potential_fn(z):
     rng_keys = random.split(random.PRNGKey(0), 2)
 
     init_kernel_map = map_fn(lambda init_param, rng_key: init_kernel(
-        init_param, trajectory_length=9, num_warmup=warmup_steps, progbar=False, rng_key=rng_key))
+        init_param, trajectory_length=9, num_warmup=warmup_steps, rng_key=rng_key))
```
neerajprad: It is strange that this wasn't throwing an error in travis.

fehiepsi: I think we missed this test when merging two test_mcmc files. Currently, we only test for chain: `pytest -vs test/test_mcmc.py -k chain`.

neerajprad: Thanks, @fehiepsi. That was my mistake, fixed.

neerajprad:
>  Do you want to do benchmark again forthe time series forecasting tutorial, which has forced us to do such optimization: #235 (comment)? I think you can benchmark with the following code cell in that tutorial

I'll benchmark and post an update. Thanks for reviewing, @fehiepsi.

neerajprad:
@fehiepsi - So I checked that the benchmark is still slower but nowhere as bad as it was. Currently, it takes 12s. on master and 24 s. with this branch. However, if you `jit` the `sgt` model itself, it takes less than 3s and we don't need to do anything else. So for cases like these with `scan` loop in the models, maybe its a better idea to just `jit` the model to begin with. Have you noticed this on other models as well?

fehiepsi:
> jit the model to begin with

Do you mean to `jit` scan_exp_val? That makes sense to me!

> Have you noticed this on other models as well?

I think that this only affects long compiling models. I haven't noticed the issue on other models. But I can recall that the change in this PR will make the single chain faster.

This PR is good to merge IMO. Should I merge this now?

neerajprad:
> Do you mean to jit scan_exp_val?

I just jiited to the `sgt` model itself. I can update the notebook in a separate PR.

Yes this should be good to merge. If there are any follow-up perf issues, I will open separate issues and address these.

fehiepsi:
> I just jiited to the sgt model itself.

This is great!!!!!!!!!! I never thought that we can do it. Thank you a lot for benchmarking and pointing out this solution!

PR #456, opened 2019-11-22
Ability to reuse compiled assets in MCMC

### LINKED ISSUE #441: Avoid recompilation in MCMC.run with new data

neerajprad:
Fixes #441. 

This introduces a `jit_model_args` argument to `MCMC`. This argument is needed because otherwise we are making a strong assumption that models can be JITed wrt to its arguments. This isn't true, e.g. for UCBAdmit or bnn example. 

As such, the current behavior is:
 - When this argument is `True`, it is the user's responsibility to ensure that their models can be JITed, i.e. they should not be using `onp` on model arguments etc. Subsequent `MCMC.run` will be much faster when this is set to True if the model is being run on a different but same sized dataset.
 - When this is `False`, the model will still give the correct results. Additionally, it will still not recompile if the model args/kwargs remain the same. e.g. if we would like to gather more samples keeping the same model arguments.

**TODOs:**
 - [x] Vectorized tests are failing because `args, kwargs` is not vectorized in `MCMC._sample`. ~Is there a better way of JITing that is compatible with vectorization, but equally fast on non-vectorized code (I have not investigated much)? An easy but hacky fix would be to copy and stack the `args, kwargs` to match the number of chains.~ I think there is not much to be done here but handling `args, kwargs` the way we handle `rng_keys`. Right now we are just tiling it so that these arguments remain the same for all chains, but it is possible to pass in different arguments and run parallel chains on different PEs. :) 
 - [x] Add documentation for `MCMC.warmup`. 
 - [x] Support `jit_model_args` argument to preserve old behavior (needed for `bnn` example).
 - [x] vectorized tests fail with `jit_model_args=False` much later when running NUTS. 

I have taken the liberty to change some of the details (and, non user-facing interface) for the `MCMCKernel` class but these are unlikely to be used outside of the `MCMC` interface which remains the same. For instance, all tests pass without issues.

### DIFF
--- a/numpyro/infer/mcmc.py
+++ b/numpyro/infer/mcmc.py
@@ -6,9 +6,11 @@ from operator import attrgetter
 import os
 import warnings
 
-from jax import jit, lax, partial, pmap, random, vmap, device_get
+from jax import jit, lax, partial, pmap, random, vmap, device_get, ShapedArray
+from jax.core import Tracer
 from jax.dtypes import canonicalize_dtype
 from jax.flatten_util import ravel_pytree
+from jax.interpreters.xla import DeviceArray
 from jax.lib import xla_bridge
 import jax.numpy as np
 from jax.random import PRNGKey
@@ -694,9 +696,18 @@ class MCMC(object):
         if self._jit_model_args:
             args, kwargs = (None,), (None,)
         else:
-            args = tree_map(lambda x: x.tobytes() if isinstance(x, np.ndarray) else x, self._args)
-            kwargs = tree_map(lambda x: x.tobytes() if isinstance(x, np.ndarray) else x,
-                              tuple(sorted(self._kwargs.items())))
+            # XXX: Is there a better hash key that we can use?
+            def _hashable(x):
+                # When the arguments are JITed, ShapedArray is hashable.
+                if isinstance(x, Tracer):
+                    return x
+                elif isinstance(x, DeviceArray):
+                    return x.copy().tobytes()
+                elif isinstance(x, np.ndarray):
+                    return x.tobytes()
+                return x
+            args = tree_map(lambda x: _hashable(x), self._args)
+            kwargs = tree_map(lambda x: _hashable(x), tuple(sorted(self._kwargs.items())))
         key = args + kwargs
         try:
             fn = self._cache.get(key, None)
@@ -785,16 +796,10 @@ class MCMC(object):
         else:
             rng_keys = random.split(rng_key, self.num_chains)
             if self._jit_model_args:
-                args = tree_map(lambda x: np.broadcast_to(x, (self.num_chains, 1)), args)
-                kwargs = tree_map(lambda x: np.broadcast_to(x, (self.num_chains, 1)), kwargs)
                 partial_map_fn = partial(self._single_chain_jit_args,
                                          collect_fields=collect_fields,
                                          collect_warmup=collect_warmup)
             else:
-                # XXX: This is unfortunately still needed. Is there a better pattern?
-                # if chain_method == 'vectorized':
-                #     args = tree_map(lambda x: np.tile(x, (self.num_chains, 1)), args)
-                #     kwargs = tree_map(lambda x: np.tile(x, (self.num_chains, 1)), kwargs)
                 partial_map_fn = partial(self._single_chain_nojit_args,
                                          model_args=args,
                                          model_kwargs=kwargs,
--- a/numpyro/infer/mcmc.py
+++ b/numpyro/infer/mcmc.py
@@ -677,8 +677,8 @@ class MCMC(object):
         hmc_state, args, kwargs = state
         return self.sampler.sample(hmc_state, args, kwargs), args, kwargs
 
-    def _single_chain_mcmc(self, init, collect_fields=('z',), collect_warmup=False, args=(), kwargs={}):
-        rng_key, init_params = init
+    def _single_chain_mcmc(self, init, collect_fields=('z',), collect_warmup=False):
+        rng_key, init_params, args, kwargs = init
         init_state = self.sampler.init(rng_key, self.num_warmup, init_params,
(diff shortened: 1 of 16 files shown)


test/test_mcmc.py
```diff
@@ -455,39 +455,9 @@ def model(y_obs):
 
     # Run MCMC on zero observations.
     kernel = NUTS(model)
-    mcmc = MCMC(kernel, 200, 200)
+    mcmc = MCMC(kernel, 300, 500)
     mcmc.run(random.PRNGKey(32), y1)
 
-    # Run on data, re-using `mcmc`.
+    # Re-run on new data - should be much faster.
```
neerajprad: This should work as expected without recompilation. cc @null-a.

neerajprad:
Thanks for reviewing, @fehiepsi! I will address all your comments shortly. I am changing this back to WIP because I am facing issues with the latest jax release where these are still recompiled. Once I figure that out, I'll update this, and address all your comments.

numpyro/infer/mcmc.py
```diff
@@ -677,16 +755,26 @@ def run(self, rng_key, *args, extra_fields=(), collect_warmup=False, init_params
...
-                                     args=args,
-                                     kwargs=kwargs)
+            if self._jit_model_args:
+                args = tree_map(lambda x: np.tile(x, (self.num_chains, 1)), args)
+                kwargs = tree_map(lambda x: np.tile(x, (self.num_chains, 1)), kwargs)
+                partial_map_fn = partial(self._single_chain_jit_args,
+                                         collect_fields=collect_fields,
+                                         collect_warmup=collect_warmup)
+            else:
+                # XXX: This is unfortunately still needed. Is there a better pattern?
```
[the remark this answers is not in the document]
neerajprad: Thanks for the suggestion! I think with this, we probably don't even need to care about this broadcasting stuff since these parameters remain the same for all the chains.

numpyro/infer/mcmc.py
```diff
@@ -677,16 +755,26 @@ def run(self, rng_key, *args, extra_fields=(), collect_warmup=False, init_params
...
             states = tree_map(lambda x: x[np.newaxis, ...], states_flat)
         else:
             rng_keys = random.split(rng_key, self.num_chains)
-            partial_map_fn = partial(self._single_chain_mcmc,
-                                     collect_fields=collect_fields,
-                                     collect_warmup=collect_warmup,
-                                     args=args,
-                                     kwargs=kwargs)
+            if self._jit_model_args:
+                args = tree_map(lambda x: np.tile(x, (self.num_chains, 1)), args)
```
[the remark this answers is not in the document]
neerajprad: We shouldn't need it if we specify the batching dimensions in vmap.

test/test_mcmc.py
```diff
@@ -455,39 +478,9 @@ def model(y_obs):
 
     # Run MCMC on zero observations.
     kernel = NUTS(model)
-    mcmc = MCMC(kernel, 200, 200)
+    mcmc = MCMC(kernel, 300, 500, jit_model_args=True)
```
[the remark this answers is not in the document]
neerajprad: Good idea. I have been trying to restrict pytest params since a lot of the test time is taken up by jax compilation. For instance, adding these params makes this test run in 1min from the earlier 10s. But I think its a good idea to aim for better coverage and then think about making tests faster. Btw..on jax 0.1.49, this was taking around 6-7 seconds, but my guess is that #463 has increased this time to 10s now.

neerajprad:
@fehiepsi - This should work with the latest jax release. The current version of JAX will remove any entries from the cache marked for collection (i.e. there is no reference being stored by the client code), which wasn't the case earlier when it was using fastcache. As such, our earlier logic did not work. This has been resolved by using `get_cached_fn`. This should work well, and I think will be much more robust. There is another issue (#463) that might be trickier to resolve - actually I am not even sure if its worth spending too much time on that, unless we find that to be an issue on a larger class of models.

numpyro/infer/mcmc.py
```diff
@@ -615,33 +687,74 @@ def __init__(self,
...
+                    return x.copy().tobytes()
+                elif isinstance(x, np.ndarray):
+                    return x.tobytes()
+                return x
+            args = tree_map(lambda x: _hashable(x), self._args)
+            kwargs = tree_map(lambda x: _hashable(x), tuple(sorted(self._kwargs.items())))
+        key = args + kwargs
+        try:
+            fn = self._cache.get(key, None)
+        except TypeError:
```
fehiepsi: Could you add a comment for this logic? I don't understand why `TypeError` happens here.

neerajprad: Yes, that might happen when args/kwargs contains an unhashable type. In that case, we can't do any caching, but this shouldn't fail. I'll add a comment.

fehiepsi: Oh, I get it now. Thanks for explaining!

fehiepsi:
I can't wait to test this on the parallel covtype code. :D

neerajprad:
Thanks for a thorough review, @fehiepsi! If you notice any issues after this PR merges, let me know. There are quite a few corner cases, which I may have overlooked.

VincentBt:
@neerajprad Great feature! Is there a way to do the same if the 2 datasets don't have the same size?

neerajprad:
> @neerajprad Great feature! Is there a way to do the same if the 2 datasets don't have the same size?

XLA will recompile if the tensor shape changes, so the only way to support this with different sized datasets is to pre-allocate the input / output tensor and then use `handlers.mask` to only consider the relevant number of entries. Let me put out a PR for `handlers.mask` and then we can investigate this particular use case.

PR #457, opened 2019-11-22
Fix docs build in travis

neerajprad:
~For some reason the previous builds seem to have passed.~ This only raises a warning, turns out that there was a lint error with #456.


fehiepsi:
Thanks for the fix!

PR #458, opened 2019-11-23
Update jax to the latest release

neerajprad:


neerajprad:
@fehiepsi - let us not merge this yet. I am seeing a perf regression with this release on a few examples. Also, it looks like the logic in #456 breaks with the release. I'm looking into that.

neerajprad:
@fehiepsi - This should be good to merge. There are a few changes to to compilation caching in JAX, but we can address that later once this PR merges.

Issue #415, opened 2019-10-28
Add tutorial for models with missing data


fehiepsi:
We should able to run MCMC for models with missing data in NumPyro because NumPyro supports improper priors through `param` primitive. This is also a good chance to illustrate that NumPyro supports improper priors.

Related issues in other repos:
+ Pyro: https://github.com/pyro-ppl/pyro/issues/1106, https://github.com/pyro-ppl/pyro/issues/1676
+ brmp: https://github.com/pyro-ppl/brmp/issues/43, https://github.com/pyro-ppl/brmp/issues/44

neerajprad:
I think the issue in Pyro refers to allowing a way to support models with sample statements of the kind `pyro.sample('obs', dist(..), obs=[some partially observed tensor])`, where we condition on the observed data and impute the missing entries. Could you clarify how having `param` statements in MCMC help this use case?

fehiepsi:
I will use the most naive way to impute those missing data. For example,
```
def model(x, y):
    loc = sample('loc', Normal(0, 1))
    scale = sample('scale', LogNormal(0, 1))
    isnan = onp.isnan(x)
    x_impute = param('x_impute', np.zeros(isnan.sum()))
    x = ops.index_update(x, onp.nonzero(isnan), x_impute)  # update x
    sample('x', dist.Normal(loc, scale), obs=x)
    ...
```
I am not sure how to make this possible with `pyro.sample`...

neerajprad:
I see - I think your method will give us the ML estimate for the missing values.

> I am not sure how to make this possible with pyro.sample.

I am not quite sure, but maybe sampling from the `dist.Normal(..)` might work:

```
    x_impute = sample('x_impute', dist.Normal(loc, scale))
    x = ops.index_update(x, onp.nonzero(isnan), x_impute)  # update x
    sample('x', dist.Normal(loc, scale), obs=x)
```

In any case, I think this will make for a good tutorial.

fehiepsi:
> the ML estimate for the missing values

Actually, those imputed values will have priors `dist.Normal(loc, scale)` because the statement
```
sample('x', dist.Normal(loc, scale), obs=x)
```
acts on the "merged" x (x_notnan & x_impute).

The script in your comment will double log_prob of `x_impute`.

neerajprad:
[...]

PR #460, opened 2019-11-24
Bayesian imputation tutorial

fehiepsi:
Addresses #415. This tutorial deals with missing age in titanic dataset.


neerajprad:
@fehiepsi - I would like to cut a release now, but I will review this tutorial soon and update our tutorials page. I think this deserves a solid review, and I don't want to block the release on this.

fehiepsi:
Sure, please take your time. This is not important for the release.

neerajprad:
@fehiepsi - I had completely lost track of this tutorial, so thanks for bringing it to my notice again. 🙂 

Some broad questions on the tutorial:
 - Since you start with "the most popular way to perform imputation is to fill a missing value with the mean, median, or mode of its corresponding feature", it will be nice to have some comparisons with a baseline model. While just taking the mean of the column is likely to perform poorly and since we are assuming a generative model `age ~ f(title)`, a more reasonable baseline would be the means for each of the title. Is there any advantage to using an improper distribution over this baseline? I was also not sure what using the Improper distribution gave us instead of sampling from `Normal(age_mu, age_sigma)` as we do for prediction (or is it the same thing?).
 - For the model validation part, it will be nice to see the full confusion matrix. Also, since the class labels weren't balanced in training, I'm not sure if 0.5 is the best threshold. It will be nice if the last graph showed the actual labels too, or we had plots with actuals and predicted together. I think if the test dataset doesn't have the survival label, we can split the training set itself into a train + test set with the survival label intact, otherwise it isn't clear what the test is giving us, and we can probably leave it out. 

Once I understand this better, I had some more minor nits that I can directly push upstream.

fehiepsi:
> if the test dataset doesn't have the survival label, we can split the training set itself into a train + test set with the survival label intact

I think the main point I wanted to deliver there is to marginalize out `age_impute` when it comes to testing. Things will go wrong if users use the full `posterior_samples` to test. I will remove that part and revise the argument a bit to focus more on that point.

> since the class labels weren't balanced in training, I'm not sure if 0.5 is the best threshold

What do you think about get predictions for `survived` and then report the proportion? I guess they are equivalent but the latter seems to be more reasonable.

> see the full confusion matrix that plot

Sure, I'll add.

> have some comparisons with a baseline model

I think using Bayesian imputation will be better. Let me add the baseline result. Hope that I am not wrong. :D

> I was also not sure what using the Improper distribution gave us instead of sampling from Normal(age_mu, age_sigma) as we do for prediction

Good point!!! Originally, I used `param` statement for those imputation values but recently changed to ImproperedUniform. You are right that is better to use `dist.Normal(...).mask(False)` so we can merge two branches: `if survived is not None`. I hope that after https://github.com/pyro-ppl/pyro/issues/1106, those imputation logics can be moved to a handler or something... so users just need to say "oh, here I want a partially-observed site, `nan` elements should be sampled" - something like
```
with handlers.partial_observe():  # age can contain `nan` values
    numpyro.sample("age", dist.Normal(age_mu, age_sigma), obs=age)
```

martinjankowiak:
:tada:
i'll review by tonight or tomorrow

martinjankowiak:
here are some initial comments i'll take another look tomorrow:
- misbalanced parentheses () in the first paragraph
- "in below" -> below
- typo "datset"
- "apriori" -> "a priori"
- "in average" -> "on average"
- can you compute test set accuracy?

fehiepsi:
Thanks for reviewing @neerajprad @martinjankowiak! I have addressed your comments and made the following changes:
+ print out confusion matrix (instead of plotting)
+ remove the prediction in test data (because we have to submit to kaggle to evaluate it), and all I wanted to deliver there is to mention about the fact that we need to marginalize out imputing samples for train data
+ revise the model using `mask` instead of ImproperUniform.
+ add an keyword `bayesian_impute=True/False` to the model to switch between 2 options.

fehiepsi:
> would it make sense to split train -> (train', test)

Yes, I think for evaluation, it is better to do train/test split or even do cross validation because hyperparameters might be chosen only for a specific split. I don't focus on evaluation part of this tutorial though. To me, it is better to have a section discuss about those "impute" values from posterior. But I don't have ideas for that, mainly because we don't know the actual values of those missing data in titanic dataset. A better tutorial would be to take a full dataset, remove some observations, then construct the models and do interpretations...

neerajprad:
Looks good! Just one thing: In model2b: L3  `numpyro.sample("x", dist.Normal(0, 1).expand([6]), obs=x_imputed)`. Should it be .expand([10])?

fehiepsi:
Thanks, I just made the fix!

Issue #467, opened 2019-11-27
Improvements to help benchmark numpyro code


neerajprad:
- [x] Ability to reuse the compiled model + adapted state to rerun sampling from different seeds.
 - [x] Separating `summary` statistics from `print_summary`
 - [x] Add support for using x64. 
 - [ ] Fix perf regression when progress bar is disabled.
cc. @fehiepsi - feel free to add to this.

PR #469, opened 2019-11-27
Ability to start MCMC sampling from the same warmup state

neerajprad:
Addresses #467. 

This introduces a `return_init_state` argument to `fori_collect`, which if `True` returns the `lower-1` state. This is used by the `reuse_warmup` argument to `MCMC.run` as follows:

```python
# run for 1 sample; num_samples=0 will throw an error
mcmc = MCMC(num_warmup, num_samples=1)
mcmc.run(rng=.., *args, **kwargs)
# change number of samples
mcmc.num_samples = 1000
# runs from the last warmup state avoiding compilation
mcmc.run(rng=..., *args, reuse_warmup=True, **kwargs)
```

This pattern can probably be improved - in particular `num_samples=1` seems a bit odd. Ideally, we should make `num_samples=0` by default in which case running mcmc without num_samples argument will simply compile the model and store the initial state from warmup. This however requires much more refactoring, whose value is questionable at the moment. If this turns out to be a useful pattern, we can make this prettier.

### DIFF
--- a/numpyro/util.py
+++ b/numpyro/util.py
@@ -130,7 +130,8 @@ def identity(x):
     return x
 
 
-def fori_collect(lower, upper, body_fun, init_val, transform=identity, progbar=True, **progbar_opts):
+def fori_collect(lower, upper, body_fun, init_val, transform=identity,
+                 progbar=True, return_init_state=False, **progbar_opts):
     """
     This looping construct works like :func:`~jax.lax.fori_loop` but with the additional
     effect of collecting values from the loop body. In addition, this allows for
@@ -164,23 +165,29 @@ def fori_collect(lower, upper, body_fun, init_val, transform=identity, progbar=T
         collection = np.zeros((upper - lower,) + init_val_flat.shape)
 
         def _body_fn(i, vals):
-            val, collection = vals
+            val, collection, start_state = vals
             val = body_fun(val)
             i = np.where(i >= lower, i - lower, 0)
+            start_state = lax.cond(i == lower-1,
+                                   start_state, lambda _: val,
+                                   start_state, lambda x: x)
             collection = ops.index_update(collection, i, ravel_fn(val))
-            return val, collection
+            return val, collection, start_state
 
-        _, collection = fori_loop(0, upper, _body_fn, (init_val, collection))
+        _, collection, start_state = fori_loop(0, upper, _body_fn, (init_val, collection, init_val))
     else:
         diagnostics_fn = progbar_opts.pop('diagnostics_fn', None)
         progbar_desc = progbar_opts.pop('progbar_desc', lambda x: '')
         collection = []
 
         val = init_val
+        start_state = None
         with tqdm.trange(upper) as t:
             for i in t:
                 val = jit(body_fun)(val)
-                if i >= lower:
+                if i == lower - 1:
+                    start_state = val
+                elif i >= lower:
                     collection.append(jit(ravel_fn)(val))
                 t.set_description(progbar_desc(i), refresh=False)
                 if diagnostics_fn:
@@ -188,7 +195,8 @@ def fori_collect(lower, upper, body_fun, init_val, transform=identity, progbar=T
 
         collection = np.stack(collection)
 
-    return vmap(unravel_fn)(collection)
+    unravel_collection = vmap(unravel_fn)(collection)
+    return unravel_collection, start_state if return_init_state else start_state
 
 
 def copy_docs_from(source_class, full_text=False):
--- a/numpyro/util.py
+++ b/numpyro/util.py
@@ -183,8 +183,7 @@ def fori_collect(lower, upper, body_fun, init_val, transform=identity,
         progbar_desc = progbar_opts.pop('progbar_desc', lambda x: '')
         collection = []
 
-        val = init_val
-        start_state = None
+        val, start_state = init_val, init_val
         with tqdm.trange(upper) as t:
             for i in t:
                 val = jit(body_fun)(val)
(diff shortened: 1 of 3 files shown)


neerajprad:
@fehiepsi - let me know if you have other suggestions on how to achieve this.

numpyro/util.py
```diff
@@ -164,31 +168,38 @@ def fori_collect(lower, upper, body_fun, init_val, transform=identity, progbar=T
...
                     collection.append(jit(ravel_fn)(val))
                 t.set_description(progbar_desc(i), refresh=False)
                 if diagnostics_fn:
                     t.set_postfix_str(diagnostics_fn(val), refresh=False)
 
         collection = np.stack(collection)
 
-    return vmap(unravel_fn)(collection)
+    unravel_collection = vmap(unravel_fn)(collection)
+    return unravel_collection, start_state if return_init_state else start_state
```
fehiepsi: typo: `else start_state` -> `else unravel_collection`. I guess we also need parentheses for `(unravel_collection, start_state)`

neerajprad: yes, just fixing this.

neerajprad:
> With that, we don't need to have reuse_warmup, so the way to achieve a similar result will be

I think you'll need to also set `mcmc.num_warmup = 0` in that example? So what you are saying is that internally if `num_warmup=0` and we have some `init_state`, we will use that instead?

fehiepsi:
@neerajprad To bypass the restriction `num_samples=0`, I think you can change the collection line in `fori_collect` to
```
collection = np.stack(collection) if len(collection) > 0 else np.zeros((upper - lower,) + init_val_flat.shape)
```
and in MCMC._single_chain_mcmc
```
        # Apply constraints if number of samples is non-zero
        site_values = tree_flatten(states['z'])[0]
        if len(site_values) > 0 and site_values[0].size > 0:
            states['z'] = lax.map(self.constrain_fn, states['z'])
```
I tested that this works with funnel example.

neerajprad:
Thanks for your suggestion, @fehiepsi. I'm also noticing another issue. I'll address these, and ping you again for a review. 😄

fehiepsi:
Re setting num_warmup=0: I believe we can use `init_state.i` as a marker for lower/upper logic of fori_collect. Let me think a bit more about it.

neerajprad:
> Re setting num_warmup=0: I believe we can use init_state.i as a marker for lower/upper logic of fori_collect. Let me think a bit more about it.

This isn't too important though; I think `num_samples = 0` makes sense. Beyond that whether we have to set `num_warmup=0` or use `reuse_warmup` doesn't really matter. I have a slight preference for the latter, so that things are explicit and there is no magical behavior for `num_warmup=0`  (e.g. is someone is expecting `hmc.init` to be triggered even with `num_warmup=0`, that won't happen with this design).

fehiepsi:
I meant we don't need to set `warmup=0`. `init_state/current_state` already holds the current step. I think we can make the internal of MCMC to be:
```
mcmc = MCMC(NUTS(model), num_warmup, 0)
mcmc.run()
assert mcmc._init_state.i == num_warmup + 0
mcmc.num_samples = 1000
mcmc.run()
assert mcmc._init_state.i == num_warmup + num_samples
```
To achieve that, we can change the lower/upper of fori_collect in MCMC to be
```
if self._init_state.i >= num_warmup:
    lower = 0
    upper = self.num_warmup + self.num_samples - self._init_state.i
else:
    lower = num_warmup - self._init_state.i
    upper = lower + self.num_samples
```

neerajprad:
Okay, let me refactor based on your suggestions first, and post an update.

neerajprad:
@fehiepsi - Regarding your last point, I think we still need the `reuse_warmup` flag, otherwise we will wrongly use the `_init_state` from last time, even if the user want to do a fresh warmup run (e.g. on new data).

fehiepsi:
Yeah, that's tricky. With that, `run`  has two meaning: keep running and "reset and run again"... :D 

May I suggest another name for the flag: `use_current_state`?

neerajprad:
>  With that, run has two meaning: keep running and "reset and run again"... :D

I suppose `.run` can be used in many ways, where the underlying requirement is that you want to share something with the previous run - e.g. share compiled code but run on fresh data, or share everything but run with a fresh seed, or share warmup, etc.

> May I suggest another name for the flag: use_current_state?

You mean change `reuse_warmup` --> `use_current_state`? The problem with that name is that it isn't really using the most current state from sampling. It is actually just using the last state returned by warmup. How about `use_warmup_state`?

fehiepsi:
> How about reuse_warmup_state?

Yeah, this is a better name for me. Please forget about `use_current_state`, it requires `fori_collect` to return the last state (instead of `start_state`). I don't intend to "run with 1000 samples", then "keep running with 9000 samples" (though I believe people usually do this way to collect samples from MCMC hubs - as I understand from parallel HMC papers which I read). Let's revisit it when necessary.

Btw, I think you need to deal with multi-chains ;).

neerajprad:
> (though I believe people usually do this way to collect samples from MCMC hubs - as I understand from parallel HMC papers which I read). Let's revisit it when necessary.

Yes, I do realize that it might be important for certain use cases. And as you mention, we would rather have some users request it first before adding it in. 

> Btw, I think you need to deal with multi-chains ;).

I'll take a look. 😅

neerajprad:
@fehiepsi - I didn't see anything immediately wrong with multiple chains, and I enabled this in the smoke test, which should pass. Let me know if I missed something.

fehiepsi:
@neerajprad Can I merge this now?

neerajprad:
Yes, lets merge this. If there are any follow-up issues, I'll put up a separate PR.

PR #470, opened 2019-11-28
Create a benchmarks module

neerajprad:
This creates a benchmarks module in the main repo. Currently this has changes from #469, which will be merged in once that PR lands. This branch should only contain changes to the benchmarks module.

We should ensure the following:
 - [x] Use float64 for CPU if possible and float32 for CUDA, specially when comparing with Stan.
 - [x] Report both time per leapfrog and time per effective sample.
 - [x] Exclude compilation + warmup from the times.
 - [x] Add `requirements.txt` with pinned dependencies so that the benchmarks are reproducible.


neerajprad:
@fehiepsi - we have made a number of changes to the interface, and jax's caching has changed too since the last version, so if you notice any benchmarks look off, let me know.

fehiepsi:
Yup, so far there is a bit regression in hmm benchmark: `0.09` -> `0.12` ms/leapfrog because we use progress_bar=True here. But that is not important IMO. The benefit of enabling `probbar=True` is we don't need to set num_samples=100,000 (to reduce the contribution of compiling time).

FYI, Pyro is 1.5x faster than before (with PyTorch 1.3.1), but it is still very slow... In 32bit, NumPyro has a bit smaller n_eff than Stan (566 vs 603), a much higher number of divergences than Stan (282 vs 42). But in 64 bits, n_eff of NumPyro is higher than Stan (752 vs 603), no divergences, 1.5x slower than 32 bits (0.17 vs 0.12 ms/leapfrog).

neerajprad:
> Yup, so far there is a bit regression in hmm benchmark: 0.09 -> 0.12 ms/leapfrog because we use progress_bar=True here. 

I think its fine to use `progress_bar=False` for benchmarks. One issue that I notice is that `progress_bar=False` does not give identical results currently (it is likely wrong). I'll try to debug and push a fix for that.

benchmarks/sparsereg_run.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+set -xe
+
+device=cpu
+N=100
+benchmark_dir=$( cd $(dirname "$0") ; pwd -P )
```
neerajprad: @fehiepsi - here's a driver script to run the benchmarks..could you run this with the same system config as the remaining ones? Let me fix the issue with the progress bar first.

benchmarks/sparsereg_run.sh
```diff
@@ -0,0 +1,20 @@
...
+
+for backend in numpyro stan; do
+    for seed in 1 2 3 4 5; do
+        for P in 10 20 30 40 50 60 70 80; do
+            if [[ ${device} = "cpu" ]]; then
+                python ${benchmark_dir}/sparse_regression.py --backend ${backend} --num-data ${N} \
+                    --num-dimensions ${P} --device ${device} --seed ${seed} --x64 #--disable-progbar
```
neerajprad: @fehiepsi - I have commented out `disable-progbar` because currently `fori_collect` incurs significant compilation cost and is slower. I'm finding that results are faster and better with progress bar enabled. I wrote a test to check if we get the same results with / without progbar, which seems to pass for a simple example. However for this example (sparse regression), the warmup state and the overall results are different. I'm not sure why that should be the case.

```python
def test_progbar():
    true_mean, true_std = 1., 2.
    num_warmup, num_samples = 10, 80

    def model(data):
        mean = numpyro.param('mean', 0.)
        std = numpyro.param('std', 1., constraint=constraints.positive)
        return numpyro.sample('obs', dist.Normal(mean, std), obs=data)

    data = dist.Normal(true_mean, true_std).sample(random.PRNGKey(1), (2000,))
    kernel = NUTS(model=model)
    # num_samples must be >= 1; otherwise we'll get an error
    mcmc = MCMC(kernel, num_warmup)
    mcmc.run(random.PRNGKey(2), data)
    mcmc.num_samples = num_samples
    mcmc.run(random.PRNGKey(3), data, reuse_warmup_state=True)
    mcmc1 = MCMC(kernel, num_warmup, progress_bar=False)
    t1 = time.time()
    mcmc1.run(random.PRNGKey(2), data)
    t2 = time.time()
    print(t2 - t1)
    mcmc1.num_samples = num_samples
    mcmc1.run(random.PRNGKey(3), data, reuse_warmup_state=True)
    t3 = time.time()
    print(t3 - t2)
    check_close(mcmc1.get_samples(), mcmc.get_samples(), atol=1e-3, rtol=0.05)
    check_close(mcmc1._warmup_state, mcmc._warmup_state, atol=1e-3, rtol=0.05)
```

neerajprad: I have committed the test, in case you want to play with it, but I'll remove it eventually.

fehiepsi: Yup, I expected that the second run will compile everything again. For example,
```python
import time
import jax; jax.config.update("jax_platform_name", "cpu")

def body_fn(i, val):
    key, x = val
    key, subkey = jax.random.split(key)
    return key, x + jax.random.gamma(subkey, 1.)

tic = time.time()
body_fn(0, (jax.random.PRNGKey(0), 1.))[0].copy()
print(time.time() - tic)

tic = time.time()
body_fn(0, (jax.random.PRNGKey(0), 1.))[0].copy()
print(time.time() - tic)

tic = time.time()
jax.lax.fori_loop(0, 10, body_fn, (jax.random.PRNGKey(0), 1.))[0].copy()
print(time.time() - tic)

tic = time.time()
jax.lax.fori_loop(0, 20, body_fn, (jax.random.PRNGKey(0), 1.))[0].copy()
print(time.time() - tic)
```
would print
```
1.6252799034118652
0.0012903213500976562
1.8120296001434326
0.0021545886993408203
```

Re test_progbar: it passes in my system (jax 0.1.53 and jaxlib 0.1.36)...

I'll try to see if we can trigger the cache when progbar=False and `collection` has the same shape.

neerajprad: In your benchmark the only thing that I find surprising  is that even if a part (or most) of the `body_fn` is pre-compiled, e.g. 

```python
@jax.jit
def sample_gamma(key, x):
    key, subkey = jax.random.split(key)
    return key, jax.random.gamma(subkey, 1.) + x


def body_fn(i, val):
    key, x = val
    return sample_gamma(key, x)
```

it will still take a good amount of time to compile the `fori_loop` - precompiling significant portion of the `body_fn` doesn't mean that the compilation of the `fori_loop` will be significantly faster. Once we realize that, its probably not surprising what we observe.

> I'll try to see if we can trigger the cache when progbar=False and collection has the same shape.

That shouldn't be hard to do (I can put up something if that seems important), but I'm not sure if it is worth doing that since this will only matter on multiple runs. I am much more concerned about correctness, that is the fact that we get quite different answers on this example seems fishy.

fehiepsi: I think this has been observed in neutra example too: fori_loop/scan, for loop with jit(svi.update), fast_math enable/disable gives different answers. I guessed the issue happens at `cholesky` at that time but haven't looked back for it...

fehiepsi: > Once we realize that

I meant so (`jax.random.gamma` is jitted by default). :)

> this will only matter on multiple runs

I think we supported multiple run with different data for progress_bar=True, so it is worth to support for progress_bar=False too. Actually, besides improving the benchmark, I also need this optimization for parallel hmc, where multi-chain is used (current optimization only applies for 1 chain MCMC). I just learned that `pmap` can use the cached
```
import time
import jax
import numpyro; numpyro.set_host_device_count(10)

def sample_gamma(key):
    return jax.random.gamma(key, 1.)

tic = time.time()
jax.pmap(sample_gamma)(jax.random.split(jax.random.PRNGKey(0), 10))
print(time.time() - tic)

tic = time.time()
jax.pmap(sample_gamma)(jax.random.split(jax.random.PRNGKey(1), 10))
print(time.time() - tic)
```
I have been trying to store `_body_fn` in `fori_collect`, WDYT about doing so? If you have a different idea, please let me know.

> I can put up something if that seems important

This would be great! :D

fehiepsi: To support this, I think we need to make a couple of changes:
- Not use fori_collect in MCMC, rewrite it such that _body_fn is cached inside MCMC.
- Drop the pattern: num_samples=0... num_warmup and num_samples are fixed values.
- Create a (hidden) method: run_warmup to only run the warmup phase, store the last value in a hidden warmup_state.

WDYT?

neerajprad: That sounds reasonable, but I think we can support the full pattern with some small changes. I do agree that we shouldn't advertise this pattern and maybe introduce a `_` before these parameters, so as to not encourage using it, but it will be nice to be able to support it.

fehiepsi: Yeah, let me try to make hidden methods `_compile` (to separate compiling time), `_warmup` (to separate warmup time),... to test out if caching works (this is not possible without your recent PRs). But feel free to make your changes, so we can have different versions to discuss/choose from.

neerajprad: Could you take a look at https://github.com/pyro-ppl/numpyro/pull/479? I think that's simpler than other alternatives? Feel free to push to it directly, I'll take a look at it again in the afternoon.

neerajprad:
I am seeing a slight reduction in `n_eff` after merging in #480 when progress bar is enabled. Most likely it is just noise, but let me know if you notice anything in the other examples.

fehiepsi:
> a slight reduction in n_eff

looking into this...

fehiepsi:
Re different result with/without progbar: Peter has a great explanation at https://github.com/google/jax/issues/829#issuecomment-502671723. I think that the situation here is similar, especially when cholesky is involved (a slight difference in precision can lead to large change in the decomposed result). Here are results of neutra svi train with/without fori loop and with/without fast-math-mode (of course, IAF guide is not a good guide here but I just want to illustrate the issue). All of them are completely different...
[neutra.pdf](https://github.com/pyro-ppl/numpyro/files/3914495/neutra.pdf)
[neutra_fast.pdf](https://github.com/pyro-ppl/numpyro/files/3914496/neutra_fast.pdf)
[neutra_fori.pdf](https://github.com/pyro-ppl/numpyro/files/3914497/neutra_fori.pdf)
[neutra_fori_fast.pdf](https://github.com/pyro-ppl/numpyro/files/3914498/neutra_fori_fast.pdf)

I think that the difference at `n_eff` is due to the combination of `body_fn` and `transform_fn` in `fori_collect`. It is related to the above precision issue.

fehiepsi:
@neerajprad I am making the following changes for consistency:
+ separating NumPyro compiling time
+ make NumPyro elapsed agree with Stan elapsed time (including warmup + sampling)
+ make sure that NumPyro uses a correct key in `.run` from the `warmup_state`. Otherwise, we need to split `rng_key` and provide different keys for `.warmup` and `.run`.
+ not use TransformedDistribution for eta1: transformed distribution is useful to get `n_eff` of `eta1` but we record `n_eff` of `eta_1_base` in Stan. So I make this change for consistency.
+ use `multi_normal_cholesky` in Stan because our MVN distribution uses Cholesky. Otherwise, Stan will use `LDLt` decomposition.
+ change scale parameters `beta_1` and `beta_2` to somehow match [authors' implementation](https://github.com/agrawalraj/skim/blob/master/skim.py) (there, they use the degree of freedom 25, while we use 6; and slope scale 3, while we use 2).  The [horseshoe prior](https://betanalpha.github.io/assets/case_studies/bayes_sparse_regression.html#21_sparsity-inducing_shapes) induces a student-t prior over slopes. They use the prior StudentT(25, 0, 3), while with the last change, we use prior StudentT(6, 0, 2). If we set `beta_1=1`, then StudentT(6, 0, 1/sqrt(3)) is a quite strict prior for the slopes. IMO, InvGamma(3, 1) is only useful for the obs, where we know that the output has a small noise. `beta_1=1` is an informative prior, which forces `m_sq` to have small values. The change in the last commit relaxes this prior (similar for `xi_sq` with `beta_2` hyperparameter). @martinjankowiak do you have any opinion about this point? I am happy to change it back if you guys found it unnecessary. Empirically, I don't find the change is much helpful for `n_eff`, but the variance of `lambda` looks better for most runs.

martinjankowiak:
i'm afraid i haven't much intuition about hyperparameters....

btw in pyro why don't you merge half-steps when you run multiple verlet steps?

(i.e. combine [line 46](https://github.com/pyro-ppl/pyro/blob/0ce93af91b7129f065676e61548a927d4e0d7878/pyro/ops/integrator.py#L46) with line 54)

fehiepsi:
> merge half-steps when you run multiple verlet steps

I recall that we merged them previously (for HMC) but separating it out during refactoring. Updating those `r` sites given a know `z_grad` is cheap IMO. But we can reinstate if it is necessary for you.

martinjankowiak:
oh i'm not making a suggestion. i just would have imagined that merging could give a noticeable perf bump but maybe not

junpenglao:
Instead of copying the old experimental Edward nuts implementation to your repository, why not importing the TFP NUTS implementation (https://github.com/tensorflow/probability/blob/master/tensorflow_probability/python/mcmc/nuts.py)? It is also a iterative implementation and you can compile it to XLA+GPU.

neerajprad:
@junpenglao - Thanks for the pointer. We will definitely add the tfp NUTS implementation to the benchmarks. This branch is basically resurrecting some of our old code from a few months back (I think the tfp one was experimental back then, but I could be wrong).

fehiepsi:
@junpenglao Could you give me an example of how to use TFP nuts so I can give it a try (I am just curious, so please take your time)? As @neerajprad pointed out, these are codes from a few months ago and we put it here for reproducible purposes. You know, it is not easy to keep up with new release packages for benchmarking. FYI, this `benchmarks` branch does not serve for state-of-the-art purposes.

junpenglao:
Thank you both! I think the best place to start is the [test case for tfp nuts](https://github.com/tensorflow/probability/blob/master/tensorflow_probability/python/mcmc/nuts_test.py) and [my introductory colab](https://colab.research.google.com/gist/junpenglao/51cd25c6372f8d2ab3490d4af8f97401/tfp_nuts_demo.ipynb).

I am more than happy to review and help in subsequent PRs if you are interested in substituting the Edward_nuts with the TFP_nuts.

fehiepsi:
Awesome! Thanks, I’ll give it a try tonight!

fehiepsi:
@neerajprad @junpenglao It seems that tfp does a bit better job (1.3x faster) in this GPU example (see [the gist](https://gist.github.com/fehiepsi/251c4c29b94e41b27cddbf76b9f04d62)).

It would be nice to see if we can improve the speed in GPU. I don't have an answer for this difference. Probably the small operators add up? @junpenglao given your expertise of tfp, do you have any insight for this difference? I think tfp also uses the iterative algorithm, runs in XLA, and most computation time (probably I am wrong about this assumption) should lie at leapfrog step... so the performance on this example should be similar.

@neerajprad I don't want to catch up with the "speed" thing in GPU unless it provides many benefits. It is good to keep the benchmarks as-is because we only want to benchmark against the recursive algorithms. However, we should create an issue to track down this problem (I suspect lax.cond plays a role here) in the future.

neerajprad:
@fehiepsi - For large datasets, I suppose that most of the compute will be dominated by tensor ops on the backward pass. I don't think this is worth investigating unless we have a few more examples. In any case, I think the unrolled NUTS implementation using XLA should be pretty much the same in terms of run time even for CPU. It will be surprising if that's not the case.

junpenglao:
Thanks a lot for the early feedback!
I wouldn't expect a large differences as well, one of the main differences (please correct me if I am wrong on the numpyro end) is that TFP version batched at each leapfrog step, and numpyro version batched at the end using vectorized_map. Since I need to implement in a ways to make sure batching works, there are quite a lot of shape handling that might under-perform/out-perform autobatching in terms of memory (my largest concern) and using TF operations that are slow.

Otherwise, once we compile to XLA, everything should give similar speed as TF and JAX are basically different interface to XLA. However, we might see some differences when we do compare large batch size (e.g., num_chain=1, 10, 100, 500)

@fehiepsi the slowness in CPU is pretty strange, maybe it is something to do with compiling to XLA - I usually run once and then run again to do the timing.

fehiepsi:
@junpenglao Yes, numpyro `chain_method='vectorized'` batches at the end, so it wastes computational time if some chains finish their trajectories earlier than the others :( (but to draw a lot of chains in GPU, this might not be a big problem). For cpu benchmark, I follow your suggestion - you can take a look at this [gist](https://gist.github.com/fehiepsi/f35f9c44e3c1814f7a7ed01dcab69462) (edward2 gives 60ms/leapfrog but tfp+XLA gives 80ms/leapfrog - but as I said, this probably comes from the experimental stage of `tf.xla.experimental.compile`; your colab with tf_nightly and `tf.function(..., autograph=False, experimental_compile=True)` might also do better). Anyway, thanks for sharing your suggestions! Using tfp seems to be not as complicated as I have thought previously. :D

junpenglao:
TF got a lot simpler with TF2😉
Also, the edward nuts implemented its own leapfrog with a for loop that built a larger graph if I remember correctly, that might be one reason as well.

benchmarks/covtype.py
```diff
@@ -0,0 +1,399 @@
...
+
+# numpyro
+import jax
+from jax import numpy as np
+import numpyro
+import numpyro.distributions as ndist
+from numpyro.examples.datasets import COVTYPE, load_dataset
+
+# pyro
+import torch
```
fehiepsi: Just a note in case we face this issue in the future: with jaxlib 0.1.37, `import torch` might cause GPU memory error when running numpyro mcmc on this example (I don't have a reason for this and also can't come with an isolated code to raise upstream).

neerajprad:
Closing this since all tasks are resolved, and all changes will remain in the benchmarks branch. We will periodically update this branch, and add git tags to pin status at specific time points - the last one is pinned at `benchmarks-20191222`. 

@junpenglao - Thanks for sharing your thoughts on the possible differences in the tfp implementation which could result in a different perf profile, it makes sense. Please feel free to use models from this branch for profiling if you find it useful. Please share any useful findings with us!

> I wouldn't expect a large differences as well, one of the main differences (please correct me if I am wrong on the numpyro end) is that TFP version batched at each leapfrog step, and numpyro version batched at the end using vectorized_map.

In NumPyro, parallel chains work by either using device parallelism (`pmap`) or vectorization. In the latter case, as @fehiepsi mentioned, we need to wait for all the K chains to meet the terminating condition before we can collect the K samples and then repeat. Is this the same with tfp, or is Alexey's auto-batching solution incorporated in the tfp implementation? 🙂

junpenglao:
Thanks @neerajprad!

Both the current TFP NUTS and Alexey's auto-batched NUTS is doing the same thing as numpyro then: wait for all chains to meet termination (u-turn or divergence), finalized one sample, start the next sample (resample momentum etc).
Potentially we can add a flag to have chains not wait for anyone - terminate when _any_ chains terminated. This will increase the speed but reduce the effective sample size as the tree building would terminated too early for most chain. Not sure how it would effect num_effective_samples per second but certainly an interesting idea to explore.
We dont have pmap version (which IIUC would be the chains wont wait for each other) but maybe it is possible to use the TF pmap to do that.

neerajprad:
Thanks for explaining, @junpenglao. 

> Potentially we can add a flag to have chains not wait for anyone - terminate when any chains terminated. This will increase the speed but reduce the effective sample size as the tree building would terminated too early for most chain. Not sure how it would effect num_effective_samples per second but certainly an interesting idea to explore.

That's something that we can also easily explore with our current setup - since we are wasting less computation, it is possible that the higher number of drawn samples results in a higher effective sample size, despite early termination. It is worth investigating!

Issue #445, opened 2019-11-12
Only use `figure_format = svg` in website


fehiepsi:
Using the `svg` format will create a lot of changes in tutorial PRs, so it is hard to review changes or typos (if any). We should use the default matplotlib backend and only use `svg` when building tutorials, probably through an environment flag.

PR #472, opened 2019-11-28
Only use `figure_format = svg` in website

fehiepsi:
Addresses #445.

@neerajprad It seems that logistic regression example does not work with the caching pattern. In particular, kernel.init is called with new kwargs, so we need to reconstruct `kernel._init_fn` and `kernel._sample_fn`.


neerajprad:
> @neerajprad It seems that logistic regression example does not work with the caching pattern. In particular, kernel.init is called with new kwargs, so we need to reconstruct kernel._init_fn and kernel._sample_fn.

Does it throw an exception? Could you share some code snippet and elaborate on what it is doing counter to expectations?

fehiepsi:
@neerajprad For now, you can see the error with exception [here](https://nbviewer.jupyter.org/github/pyro-ppl/numpyro/blob/6537a0f81625a37e046d8bb8af934e2d4b77556d/notebooks/source/bayesian_regression.ipynb#Model-2:-Predictor---Median-Age-of-Marriage). I'll provide a simple replicate example later. :)

neerajprad:
Oh..I was looking at the logistic regression example, not that one. Don't worry about replicating it; I'll run it and push a fix for that.

fehiepsi:
Oppss, sorry. I meant bayesian regression. @@

neerajprad:
@fehiepsi - is this good to merge?

fehiepsi:
Oh no, I haven't run this again. >"<

fehiepsi:
@neerajprad I think it is good to merge when tests pass. :) I only add the check for the flag "NUMPYRO_SPHINXBUILD" in the environment to disable/enable "svg" format and rerun those notebooks.

PR #473, opened 2019-11-28
Enable x64 tests in travis

neerajprad:
Some minor changes to help with #470:
 - [x] Enable x64 tests; they were not being tested due to a typo.
 - [x] Add an `enable_x64` option to numpyro for convenience.
 - [x] Fix issue with #469 where we were running warmup adaptation during sampling - we shouldn't reset `HMCState.i` to 0 since that will be used by the `hmc` function when deciding whether to run the adaptation.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,7 @@ jobs:
           python: 3.6
           script:
               - pytest -vs -k "not test_example" --deselect test/contrib/test_distributions_contrib.py --durations=100
-              - JAX_ENABLE_x64=1 pytest -vs test/test_mcmc.py -k x64
+              - JAX_ENABLE_X64=1 pytest -vs test/test_mcmc.py -k x64
               - XLA_FLAGS="--xla_force_host_platform_device_count=2" pytest -vs test/test_mcmc.py -k \
                   "chain or pmap or vmap"
         - name: "examples"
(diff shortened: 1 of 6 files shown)


numpyro/infer/mcmc.py
```diff
@@ -752,10 +752,7 @@ def _single_chain_mcmc(self, rng_key, init_params, args, kwargs,
                                     diagnostics_fn=diagnostics)
         states, warmup_state = collect_vals
         # Get first argument of type `HMCState`
-        warmup_state = warmup_state[0]
-        # Note that setting i = 0 may result in recompilation due to python integers having
-        # weak type
-        self._warmup_state = warmup_state._replace(i=np.zeros_like(warmup_state.i))
+        self._warmup_state = warmup_state[0]
```
neerajprad: @fehiepsi - I think this should be enough to get the correct behavior, but let me know if I missed something.

.travis.yml
```diff
@@ -39,7 +39,7 @@ jobs:
           python: 3.6
           script:
               - pytest -vs -k "not test_example" --deselect test/contrib/test_distributions_contrib.py --durations=100
-              - JAX_ENABLE_x64=1 pytest -vs test/test_mcmc.py -k x64
+              - JAX_ENABLE_X64=1 pytest -vs test/test_mcmc.py -k x64
```
fehiepsi: ................................................

fehiepsi:
> Fix issue with #469 where we were running warmup adaptation during sampling

Oh, this might explain why I saw bad n_eff in hmm example. Thanks for the fix!

fehiepsi:
@neerajprad Please merge this as your convenience.

PR #474, opened 2019-11-29
Fix get_potential_fn for dynamic args

neerajprad:
Fixes #472.


test/test_mcmc.py
```diff
@@ -491,3 +491,34 @@ def model(y_obs):
...
+    z = onp.random.randn(10)
+
+    # Run MCMC on zero observations.
+    kernel = NUTS(model)
+    mcmc = MCMC(kernel, 20, jit_model_args=jit_args)
+    mcmc.run(random.PRNGKey(0), a, b=None, z=z)
+    mcmc.num_samples = 10
+    mcmc.run(random.PRNGKey(1), a, b=None, z=z, reuse_warmup_state=True)
+
+    # Re-run on new data - should be much faster.
```
fehiepsi: however, I don't understand why this is faster...

neerajprad: It shoudn't be, sorry. That comment was just copied over, will fix.

PR #476, opened 2019-11-29
Fix bug in fori_collect for init val collection

### LINKED ISSUE #475: Regression in fori_collect when progress bar disabled
I am finding that with progress bar disabled, `fori_collect` is not returning the same results for the sparse regression example:
 - [x] It fetches the wrong `init_state` and runs adaptation twice.
 - [ ] Results returned when progress bar is enabled/disabled are different.
 - [x] When progress bar is disabled, the time taken is almost twice. This seems to be a regression from earlier when disabling the progress bar was much faster.

neerajprad:
Partly fixes #475 - this is only a fix for the first part, namely collecting the wrong `init_state`. I'm still investigating the perf regression.

### DIFF
--- a/numpyro/util.py
+++ b/numpyro/util.py
@@ -181,11 +181,11 @@ def fori_collect(lower, upper, body_fun, init_val, transform=identity,
 
         def _body_fn(i, vals):
             val, collection, start_state = vals
-            val = body_fun(val)
-            i = np.where(i >= lower, i - lower, 0)
-            start_state = lax.cond(i == lower-1,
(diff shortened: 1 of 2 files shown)


neerajprad:
> To trigger the cache from the first run, probably we need to fix collection to a constant shape.

Hmm..we'll not really know the future shapes, so this isn't something we can reliably do. I was hoping that the thing that would take most of the compilation time would be jiting the `body_fn` which shouldn't happen twice (exactly as for the case when progress bar is enabled).

PR #477, opened 2019-11-30
Avoid double compiling issue

### LINKED ISSUE #463: Model compilation happens twice with the latest JAX version
With the latest jax version, many models are compiled twice - once at the beginning of warmup and then again when sampling starts. The reason for this is the introduction of weak types in JAX (all python scalars have a weak type which might be promoted later depending on the dtype of other arrays). 

As such, the arguments to `sample_fn` at the beginning of both the stages may differ in this aspect, i.e. arguments passed by the user may have weak types but those that are used to begin sampling don't, which results in a cache miss. The full impact of this still needs to be assessed. It will probably not matter for small models.

fehiepsi:
Probably this will resolve #463.

Tested on the `hmm` benchmark at #470, where during the warmup phase, body_fn is compiled for the first step, then compiled again for the second step (because the init value contains python scalar).

### DIFF
--- a/numpyro/infer/mcmc.py
+++ b/numpyro/infer/mcmc.py
@@ -6,7 +6,7 @@ from operator import attrgetter
 import os
 import warnings
 
-from jax import jit, lax, partial, pmap, random, vmap, device_get
+from jax import jit, lax, partial, pmap, random, vmap, device_get, device_put
 from jax.core import Tracer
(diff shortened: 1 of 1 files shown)


neerajprad:
I completely forgot about device put (and was manually changing all scalars to arrays!), this is great!

PR #479, opened 2019-11-30
Cache compilation when progbar=False

### LINKED ISSUE #467: Improvements to help benchmark numpyro code

neerajprad:
Fixes #467.

### DIFF
--- a/numpyro/infer/mcmc.py
+++ b/numpyro/infer/mcmc.py
@@ -6,7 +6,7 @@ from operator import attrgetter
 import os
 import warnings
 
-from jax import jit, lax, partial, pmap, random, vmap, device_get
+from jax import jit, lax, partial, pmap, random, vmap, device_get, device_put
 from jax.core import Tracer
 from jax.dtypes import canonicalize_dtype
 from jax.flatten_util import ravel_pytree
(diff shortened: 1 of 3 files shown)


numpyro/infer/mcmc.py
```diff
@@ -732,20 +740,19 @@ def _single_chain_mcmc(self, rng_key, init_params, args, kwargs,
             if self._warmup_state is None:
                 raise ValueError('No `init_state` found; warmup results cannot be reused.')
             num_warmup = 0
-            init_state = self._warmup_state
+            init_state = self._warmup_state._replace(rng_key=rng_key)
```
neerajprad: This was a bug! We were basically returning the same result as the previous run.

neerajprad:
@fehiepsi - I think this should fix the compilation issue. Feel free to push to this branch.

fehiepsi:
This is great! This is what I wanted but doesn't know how to achieve. :D

fehiepsi:
@neerajprad Whoa, this works well with `hmm` benchmark (when I fixed `collection` to a fixed size). Bravo!

I'll revise MCMC in another PR.

numpyro/infer/mcmc.py
fehiepsi: I think for reproducible purpose, it would be nice to allow `rng_key` have a batch_size (in that case, we won't split). Previously, with functional interface, we can verify the consistency with
```
mcmc_1chain.run(random.PRNGKey(0), ...)
mcmc_4chains.run(vmap(random.PRNGKey)(range(4)), ...)
```
Similarly, here, we can verify the consistency of `reuse_warmup_state` by setting
```
mcmc.run(mcmc._warmup_state.rng_key, ...)
```

PR #480, opened 2019-12-01
Separating compiling, warmup phases

fehiepsi:
This PR
+ adds `_compile` method to separate out compiling time (only useful for benchmark purpose so I made it hidden) 
+ adds `warmup` method to separate out the warmup phase and run phase.
+ groups all logics of `collect_warmup`, `reuse_warmup_state` to set_fori_collect_infos method. This makes the code easier to follow.
+ allow `rng_key` have a batch size for reproducible purpose
+ add `compiling_size` to `fori_collect` for the caching purpose (we need `collection` to have the same size across runs)
+ change return_warmup_state to `return_last_val`, which is more suitable for the standalone `fori_collect`.
+ use the same `body_fn` for progbar=True/False in `fori_collect`. This saves a bit of time when we use the progbar, especially for small models.

### Fixes
+ warmup_state is updated after each run. This is unexpected behavior. Having a separated warmup method solves this.
+ warmup_state is the same across multi-chains. This is solved by moving init_state logic to `run` method.

### Side effect

I only discover that we can do this while writing this comment. :D Users can keep drawing `num_samples` samples from the last state by using the pattern
```python
mcmc._warmup_state = mcmc._last_state  # hack: set warmup_state to be last_state
mcmc.run(..., reuse_warmup_state=True)
```

### TODO:
Add tests to verify that
- [x] caching works for single/multi-chain.
- [x] the result is reproducible for single/multi-chain.

### DIFF
--- a/numpyro/infer/mcmc.py
+++ b/numpyro/infer/mcmc.py
@@ -796,7 +796,7 @@ class MCMC(object):
         self._collection_params["collection_size"] = collection_size
 
     def _compile(self, rng_key, *args, extra_fields=(), init_params=None, **kwargs):
-        self._set_fori_collect_infos(0, 0, self.num_samples)
+        self._set_collection_params(0, 0, self.num_samples)
         self.run(rng_key, *args, extra_fields=extra_fields, init_params=init_params, **kwargs)
         rng_key = (_hashable(rng_key),)
         args = tree_map(lambda x: _hashable(x), args)
@@ -827,11 +827,11 @@ class MCMC(object):
         :param kwargs: Keyword arguments to be provided to the :meth:`numpyro.infer.mcmc.MCMCKernel.init`
             method. These are typically the keyword arguments needed by the `model`.
         """
-        self._warmup_state = False
+        self._warmup_state = None
         if collect_warmup:
-            self._set_fori_collect_infos(0, self.num_warmup, self.num_warmup)
+            self._set_collection_params(0, self.num_warmup, self.num_warmup)
         else:
-            self._set_fori_collect_infos(self.num_warmup, self.num_warmup, self.num_samples)
+            self._set_collection_params(self.num_warmup, self.num_warmup, self.num_samples)
         self.run(rng_key, *args, extra_fields=extra_fields, init_params=init_params, **kwargs)
         self._warmup_state = self._last_state
 
@@ -859,7 +859,7 @@ class MCMC(object):
             rng_key = random.split(rng_key, self.num_chains)
 
         if self._warmup_state is not None:
-            self._set_fori_collect_infos(0, self.num_samples, self.num_samples)
+            self._set_collection_params(0, self.num_samples, self.num_samples)
             init_state = self._warmup_state._replace(rng_key=rng_key)
 
         chain_method = self.chain_method
@@ -914,7 +914,7 @@ class MCMC(object):
         self._last_state = last_state
         self._states = states
         self._states_flat = states_flat
-        self._set_fori_collect_infos()
+        self._set_collection_params()
 
     def get_samples(self, group_by_chain=False):
         """
--- a/numpyro/infer/mcmc.py
+++ b/numpyro/infer/mcmc.py
@@ -611,15 +611,16 @@ class NUTS(HMC):
         self._algo = 'NUTS'
 
 
+def _get_value_from_index(xs, i):
+    return tree_map(lambda x: x[i], xs)
+
+
 def _laxmap(f, xs):
     n = tree_flatten(xs)[0][0].shape[0]
 
-    def get_value_from_index(i):
-        return tree_map(lambda x: x[i], xs)
-
     ys = []
     for i in range(n):
-        x = jit(get_value_from_index)(i)
+        x = jit(_get_value_from_index)(xs, i)
         ys.append(f(x))
 
     return tree_multimap(lambda *args: np.stack(args), *ys)
@@ -641,6 +642,18 @@ def _collect_fn(collect_fields):
         return attrgetter(*collect_fields)(x[0])
 
     return collect
+
+
+# XXX: Is there a better hash key that we can use?
+def _hashable(x):
+    # When the arguments are JITed, ShapedArray is hashable.
+    if isinstance(x, Tracer):
+        return x
+    elif isinstance(x, DeviceArray):
+        return x.copy().tobytes()
+    elif isinstance(x, np.ndarray):
+        return x.tobytes()
+    return x
 
 
 class MCMC(object):
@@ -700,6 +713,8 @@ class MCMC(object):
         self._last_state = None
         # HMCState returned by last warmup
         self._warmup_state = None
+        # HMCState returned by hmc.init_kernel
+        self._init_state_cache = {}
         self._cache = {}
         self._fori_collect_infos = {}
         self._set_fori_collect_infos()
@@ -708,16 +723,6 @@ class MCMC(object):
         if self._jit_model_args:
             args, kwargs = (None,), (None,)
         else:
-            # XXX: Is there a better hash key that we can use?
-            def _hashable(x):
-                # When the arguments are JITed, ShapedArray is hashable.
-                if isinstance(x, Tracer):
-                    return x
-                elif isinstance(x, DeviceArray):
-                    return x.copy().tobytes()
-                elif isinstance(x, np.ndarray):
-                    return x.tobytes()
-                return x
             args = tree_map(lambda x: _hashable(x), self._args)
             kwargs = tree_map(lambda x: _hashable(x), tuple(sorted(self._kwargs.items())))
         key = args + kwargs
@@ -736,6 +741,17 @@ class MCMC(object):
             if key is not None:
                 self._cache[key] = fn
         return fn
+
+    def _get_cached_init_state(self, rng_key, args, kwargs):
+        rng_key = (_hashable(rng_key),)
+        args = tree_map(lambda x: _hashable(x), args)
+        kwargs = tree_map(lambda x: _hashable(x), tuple(sorted(kwargs.items())))
+        key = rng_key + args + kwargs
+        try:
+            return self._init_state_cache.get(key, None)
+        # If unhashable arguments are provided, return None
+        except TypeError:
+            pass
 
     def _single_chain_mcmc(self, rng_key, init_state, init_params, args, kwargs, collect_fields=('z',)):
         num_warmup = self.num_warmup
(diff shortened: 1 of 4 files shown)


fehiepsi:
@neerajprad I think this is good to review. This requires an explicit `warmup` call to make `reuse_warmup_state` work. I have tested that caching works with the pattern
```
mcmc._compile(rng_key)
mcmc.warmup(rng_key)
mcmc.run(rng_key1, reuse_warmup_states)
```
for
+ single chain with/without progbar
+ vectorized with/without progbar
+ sequential with progbar

For parallel method, I think it just partly works because I observed some overhead in hmm example.

For sequential without progbar, I think that caching does not work because we apply `lax.map` for the whole process (so caching for fori_collect is not enough). I guess we can make it work but probably not worth the effort.

numpyro/infer/mcmc.py
```diff
@@ -696,24 +709,20 @@ def __init__(self,
         self._jit_model_args = jit_model_args
         self._states = None
         self._states_flat = None
+        # HMCState returned by last run
+        self._last_state = None
         # HMCState returned by last warmup
         self._warmup_state = None
+        # HMCState returned by hmc.init_kernel
+        self._init_state_cache = {}
```
fehiepsi: This is only useful when running `_compile` method for benchmarking.

numpyro/infer/mcmc.py
```diff
@@ -733,75 +742,127 @@ def _hashable(x):
...
+
+    def _set_fori_collect_infos(self, lower=None, upper=None, collect_size=None):
+        self._fori_collect_infos["lower"] = self.num_warmup if lower is None else lower
+        self._fori_collect_infos["upper"] = self.num_warmup + self.num_samples if upper is None else upper
+        self._fori_collect_infos["collect_size"] = collect_size
+
+    def _compile(self, rng_key, *args, extra_fields=(), init_params=None, **kwargs):
+        self._set_fori_collect_infos(0, 0, self.num_samples)
+        self.run(rng_key, *args, extra_fields=extra_fields, init_params=init_params, **kwargs)
+        rng_key = (_hashable(rng_key),)
```
neerajprad: same here, we shouldn't need `_hashable` for rng key.

numpyro/util.py
```diff
@@ -174,9 +174,9 @@ def fori_collect(lower, upper, body_fun, init_val, transform=identity,
         be any Python collection type containing `np.ndarray` objects.
     :param transform: a callable to post-process the values returned by `body_fn`.
     :param progbar: whether to post progress bar updates.
-    :param bool return_init_state: If `True`, the state at iteration `lower-1`,
-        where the collection begins, is also returned. This has the same type
-        as `init_val`.
+    :param bool return_last_val: If `True`, the last value is also returned.
+        This has the same type as `init_val`.
+    :param int collect_size: Size of the returned collection.
```
neerajprad: Could you add that the collection size will be `upper-lower` if this is not specified, and also that if this is larger than `upper-lower` only the top `upper-lower` entries will be non-zero. I think a more general utility should allow for a collection argument (instead of size) that will be used to fill up the entries, but since we are only using this specifically for mcmc, we can keep it as is for now.

numpyro/infer/mcmc.py
```diff
@@ -733,75 +742,127 @@ def _hashable(x):
...
-    def _single_chain_mcmc(self, rng_key, init_params, args, kwargs,
-                           collect_fields=('z',), collect_warmup=False, reuse_warmup_state=False):
+    def _get_cached_init_state(self, rng_key, args, kwargs):
+        rng_key = (_hashable(rng_key),)
+        args = tree_map(lambda x: _hashable(x), args)
+        kwargs = tree_map(lambda x: _hashable(x), tuple(sorted(kwargs.items())))
+        key = rng_key + args + kwargs
+        try:
+            return self._init_state_cache.get(key, None)
+        # If unhashable arguments are provided, return None
```
neerajprad: nit: we should just `return None` below to match the comment above.

numpyro/infer/mcmc.py
```diff
@@ -733,75 +742,127 @@ def _hashable(x):
...
-        return states
+        return states, last_state
+
+    def _single_chain_jit_args(self, init, collect_fields=('z',)):
+        return self._single_chain_mcmc(*init, collect_fields=collect_fields)
+
+    def _single_chain_nojit_args(self, init, model_args, model_kwargs, collect_fields=('z',)):
+        return self._single_chain_mcmc(*init, model_args, model_kwargs, collect_fields=collect_fields)
+
+    def _set_fori_collect_infos(self, lower=None, upper=None, collect_size=None):
```
neerajprad: nit: can we rename this to `_set_collection_arguments(..)` or `_set_collection_params` or something? This can also be a namedtuple `_collection_params = namedtuple(.., ['upper', 'lower', 'collection_size'])`, though its not that important.

numpyro/infer/mcmc.py
```diff
@@ -733,75 +742,127 @@ def _hashable(x):
...
         self._args = args
         self._kwargs = kwargs
+        init_state = self._get_cached_init_state(rng_key, args, kwargs)
+        if self.num_chains > 1 and rng_key.ndim == 1:
+            rng_key = random.split(rng_key, self.num_chains)
+
+        if reuse_warmup_state:
+            self._set_fori_collect_infos(0, self.num_samples, self.num_samples)
+            if self._warmup_state is None:
+                raise ValueError('No `init_state` found; warmup results cannot be reused.')
```
neerajprad: nit: change this to `_warmup_state` to reflect the actual attribute name. Also move up before `self._set_fori_..`.

test/test_mcmc.py
```diff
@@ -100,9 +100,12 @@ def model(data):
     data = true_coef + random.normal(random.PRNGKey(0), (1000,))
     kernel = NUTS(model=model)
     mcmc = MCMC(kernel, num_warmup=num_warmup, num_samples=num_samples)
-    mcmc.run(random.PRNGKey(2), data, collect_warmup=True)
+    mcmc.warmup(random.PRNGKey(2), data, collect_warmup=True)
```
neerajprad: Hmm..this is a breaking change, and will deviate from the pyro-api interface. Can we instead do the following:
 - set `reuse_warmup_state` to `True`.
 - call `self.warmup()` from `run` only if `reuse_warmup_state` is `False` or `reuse_warmup_state = True` and `_warmup_state` is None.

We may need to move the existing logic from `.run` to another method like `.collect` or something so that `.warmup` can call that instead of `.run`.

fehiepsi: The problem is we don't want to have this separation for multi-chains, i.e. a chain doesn't need to wait for other chains to complete the warmup phase.

How about:
+ Making warmup a hidden method, which is only useful for benchmarking if we make the following change:
+ Return low_idx val and last val for each fori_collect run. And warmup state will be stored if `reuse_warmup_state` is False.

WDYT?

neerajprad: > The problem is we don't want to have this separation for multi-chains, i.e. a chain doesn't need to wait for other chains to complete the warmup phase.

Could you explain how this is handled in the current code? 

> Return low_idx val and last val for each fori_collect run. And warmup state will be stored if reuse_warmup_state is False.

That sounds reasonable. I'm not sure if we need to return `last_val`for benchmarking, unless it is to support things like being able to restart from the last state.

More generally, I think amongst the space of solutions, we can probably rate them along the following lines:
 - Whether most of the state setting required for handling multiple chains is confined to a single method. This used to be `.run` but increasingly, things have become complex enough that the logic has started leaking out. Still, I think it will help to contain as much of the complexity to a single method so that we are able to freely refactor other parts without keeping the multi chain constraints in mind all the time.
 - Whether it is compatible with pyro / pyro-api's current interface.
 - Minimizing model re-compilation whenever possible.
 - Any benchmarking specific changes shouldn't impact the interface unless they also support a concrete use case.

fehiepsi: Regarding the user interface, there are two changes for this release:
+ Removing `collect_warmup` arg, which is used to diagnose the warmup phase. Most users won't use it and it can be achieved by using `_warmup` method.
+ Having an additional `reuse_warmup_state` arg, which is useful for #438 (though I prefer to have a hidden method doing this job at least for the next release, I leave it up to you to make the decision).

New hidden methods `_compiling` and `_warmup` are mainly used to separate out compiling/warmup/sampling time for benchmarking. They only affect the `fori_collect` interface.

> Could you explain how this is handled in the current code?

Currently, for each chain, we run both warmup + sampling, rather than for each chain, we run warmup and after finishing, we run sampling. The changes in this PR do not affect that. However, if users run `_warmup` first and run sampling with `reuse_warmup_phase`, then they have to wait for `_warmup` to finish.

> amongst the space of solutions

I guess the changes in this PR only affect `fori_collect` interface. So it is good to go.

> we need to return last_val for benchmarking

So far, it is not useful for:
+ compiling (because val_lower is returned and val_lower=val_last)
+ warmup with collect=False (because val_last=val_lower)
+ run with/without reuse_warmup_state

It is only useful for:
+ warmup with collect_warmup=True (where val_last=warmup_state)
+ restart run from last state (useful when users want to draw a lot of samples but do not have enough RAM, the memory restricts `collection` to have a limited size)

On the other hand, val_lower is useful for:
+ run with reuse_warmup_state

Though they affect the `fori_collect` interface, I think it is good to have both of them. Otherwise, we can make fori_collect with args to get those values a hidden function and only expose the one without `collect_size`, `return_last_val`, `return_init_val`,... args. WDYT? That way, I believe no one will blame us for having those additional args. ;) (anyway, this might not be an important point).

numpyro/infer/mcmc.py
```diff
@@ -733,75 +742,127 @@ def _hashable(x):
...
             method. These are typically the keyword arguments needed by the `model`.
         """
         self._args = args
         self._kwargs = kwargs
+        init_state = self._get_cached_init_state(rng_key, args, kwargs)
+        if self.num_chains > 1 and rng_key.ndim == 1:
+            rng_key = random.split(rng_key, self.num_chains)
+
+        if reuse_warmup_state:
+            self._set_fori_collect_infos(0, self.num_samples, self.num_samples)
```
neerajprad: I think this is the part that I do not  understand - why do we not have to set the collection parameters if `reuse_warmup_state = False`?

test/test_mcmc.py
neerajprad: So it seems like this is already doing what I mentioned, i.e. `run` will run both warmup and sampling. The only difference is that if we are looking to use the last `warmup` state, then we need to call `.warmup` explicitly. I think that's a fine constraint to have. So please disregard my comments above. I had just one question that I asked above.

neerajprad: Regarding removing `reuse_warmup_state`, I think we can probably do that now? The only reason why we needed it earlier was to be able to reuse/rerun warmup, but we can do it now by just calling `.warmup` directly.

fehiepsi: > So it seems like this is already doing what I mentioned, i.e. run will run both warmup and sampling. The only difference is that if we are looking to use the last warmup state, then we need to call .warmup explicitly.

Yep!

> Regarding removing reuse_warmup_state, I think we can probably do that now?

I think it is fine to remove. Though if users want to do `warmup + run` again, then they have to call `.warmup` again or setting `._warmup_state = None`. WDYT?

neerajprad: I think that's fine, i.e. users can continue to use `.run` which will have the same behavior as earlier. Otherwise, if they explicitly call `.warmup` first, we will skip warmup when they use `.run`.

numpyro/infer/mcmc.py
fehiepsi: If `reuse_warmup_state = False`, we have default values: lower=warmup, upper=warmup+num_samples, collect_size=None=num_samples. We reset to this default value after each `.run`, so if any method changes default value, after `.run`, these default values will be reset.

numpyro/infer/mcmc.py
```diff
@@ -733,75 +742,127 @@ def _hashable(x):
                 self._cache[key] = fn
         return fn
 
-    def _single_chain_mcmc(self, rng_key, init_params, args, kwargs,
-                           collect_fields=('z',), collect_warmup=False, reuse_warmup_state=False):
+    def _get_cached_init_state(self, rng_key, args, kwargs):
+        rng_key = (_hashable(rng_key),)
```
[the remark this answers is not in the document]
fehiepsi: unfortunately, `{jax.random.PRNGKey(0): 10}` triggers TypeError...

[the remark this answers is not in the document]
neerajprad: Oh..`rng_key` is just a jax array, makes sense. For some reason, I thought that PRNGKey is hashable.

PR #486, opened 2019-12-03
Fix progbar description when warmup / sample are run separately

### LINKED ISSUE #483: mcmc progbar does not display correctly running phase
The last change does not update the logic of progressbar. So it displays `warmup`, `sample` incorrectly when we run `.warmup` separately from `.run`.

neerajprad:
Fixes #483. Note that this will show `sample` when `collect_warmup=True`, but I think that's fine, since we are indeed collecting samples in addition to doing adaptation.


fehiepsi:
Thanks, @neerajprad ! The behavior of `collect_warmup=True` is expected to me. +1:

PR #487, opened 2019-12-04
Minor changes to infer.util for 0.2.2

neerajprad:
Addresses #455.

### DIFF
--- a/numpyro/infer/util.py
+++ b/numpyro/infer/util.py
@@ -32,11 +32,12 @@ __all__ = [
 
 def log_density(model, model_args, model_kwargs, params, skip_dist_transforms=False):
     """
-    Computes log of joint density for the model given latent values ``params``.
+    (EXPERIMENTAL INTERFACE) Computes log of joint density for the model given
+    latent values ``params``.
 
     :param model: Python callable containing NumPyro primitives.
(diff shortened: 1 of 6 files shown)


numpyro/infer/util.py
```diff
@@ -416,8 +419,11 @@ def constrain_fun(*args, **kwargs):
     return potential_fn, constrain_fun
 
 
-def initialize_model(rng_key, model, *model_args, init_strategy=init_to_uniform(),
-                     dynamic_args=False, **model_kwargs):
+def initialize_model(rng_key, model,
+                     init_strategy=init_to_uniform(),
+                     dynamic_args=False,
+                     model_args=(),
+                     model_kwargs=None):
```
neerajprad: this is the only non-doc related change.

fehiepsi: Could you also mark `find_valid_initial_params` as experimental too? We might need to cache `find_valid_initial_params` when we rerun MCMC with new data. This is usually pretty fast comparing to compile `body_fn` but in some cases, it took a few seconds to finish.

neerajprad: I think that's marked as experimental but I'm changing the varargs to match the other utils.

PR #488, opened 2019-12-04
Bump up version to 0.2.2

neerajprad:


setup.py
```diff
@@ -31,8 +31,8 @@
     author_email='npradhan@uber.com',
     install_requires=[
         # TODO: pin to a specific version for the release (unless JAX's API becomes stable)
-        'jax>=0.1.52',
-        'jaxlib>=0.1.32',
+        'jax==0.1.53',
```
neerajprad: @fehiepsi - I'm using the latest jax version, is that okay with you?

fehiepsi: Sure. I used it recently. 👍

PR #489, opened 2019-12-04
Release patch for python 3.7 incompatibility

neerajprad:
I noticed that MCMC is failing on python-3.7 because of `TypeError: get_progbar_desc_str() got multiple values for argument 'num_warmup'`. This doesn't work in python 3.7:

```python
def f(a, b):
  print(a+b)

functools.partial(f, a='x')('y')
```


numpyro/infer/mcmc.py
```diff
@@ -774,7 +774,7 @@ def _single_chain_mcmc(self, rng_key, init_state, init_params, args, kwargs, col
                                     return_last_val=True,
                                     collection_size=self._collection_params["collection_size"],
                                     progbar_desc=functools.partial(get_progbar_desc_str,
-                                                                   num_warmup=lower_idx),
+                                                                   lower_idx),
```
neerajprad: this is the bug fix.

neerajprad:
Closing in favor of #490.

Issue #491, opened 2019-12-08
Numerical issues in distributions.util.poisson


nkaimcaudle:
Hi, seems the function distributions.util.poisson encounters numerical issues when rate parameter is too high, in the code I see exp(-rate) which will go to zero as rate parameter increases.

Here is my code showing the issue

```
import jax.numpy as np
import jax.random as random
import numpy as onp
import numpyro
from numpyro.distributions.util import poisson

mu = rate = 1000
N = 2**12

A = onp.random.poisson(lam=mu, size=N)
print( A.mean() )  ###prints around 1000.

key = random.PRNGKey(64)
B = poisson(key, rate=rate, shape=(N,))
print( B.mean() ) #prints arounds 87.
```

fehiepsi:
Thanks for the report!

You are right, the algorithm in numpyro is only good for a small rate. For a large rate, I think we need to implement another algorithm e.g. [the one in wikipedia](https://en.wikipedia.org/wiki/Poisson_distribution#Generating_Poisson-distributed_random_variables) or [the transformed rejection method in numpy](https://github.com/numpy/numpy/blob/master/numpy/random/src/distributions/distributions.c#L537).

We'll address this issue for the next release. :)

PR #495, opened 2019-12-17
resolve overflow issue of poisson sampler

fehiepsi:
Addresses #491. The implementation is based on [numpy](https://github.com/numpy/numpy/blob/master/numpy/random/src/distributions/distributions.c#L574), which is also used in PyTorch. The version in [wikipedia](https://en.wikipedia.org/wiki/Poisson_distribution#Generating_Poisson-distributed_random_variables) works but is slow.

I have tested the performance of using `lax.map` in CPU and `lax.vmap` in GPU. The current setting gives the best performance overall.

### DIFF
--- a/numpyro/distributions/util.py
+++ b/numpyro/distributions/util.py
@@ -46,6 +46,8 @@ def categorical(key, p, shape=()):
     return _categorical(key, p, shape)
 
 
+# Ref https://github.com/numpy/numpy/blob/8a0858f3903e488495a56b4a6d19bbefabc97dca/
+# numpy/random/src/distributions/distributions.c#L574
 def _poisson_large(val):
     rng_key, lam = val
     slam = np.sqrt(lam)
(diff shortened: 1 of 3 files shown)


numpyro/distributions/util.py
```diff
@@ -44,27 +46,74 @@ def categorical(key, p, shape=()):
...
-        u = random.uniform(rng_key_u, shape)
-        p = p * u
-        return k, p, rng_key
-
-    k, _, _ = lax.while_loop(lambda val: np.any(val[1] > L), body_fn, (k, p, key))
-    return k - 1
+    rate = lax.convert_element_type(rate, canonicalize_dtype(np.float64))
+    rate = np.broadcast_to(rate, shape)
+    rng_keys = random.split(key, np.size(rate))
+    if xla_bridge.get_backend().platform == 'cpu':
```
neerajprad: I am curious as to why this should differ with the backend.

fehiepsi: For rejective samplers, it is better to use `lax.map` because in the vectorized map, early-finished samples have to wait for other proposals accepted. However, in GPU vectorized is preferred (i.e. faster).

PR #496, opened 2019-12-17
Update for jax >= 0.1.55

fehiepsi:
The new version of jax renames `_promote_args_like` to `_promote_args_inexact`, so we need to rename and pump the requirement for master branch.

### DIFF
--- a/test/test_distributions_util.py
+++ b/test/test_distributions_util.py
@@ -3,12 +3,10 @@ from numbers import Number
 import numpy as onp
 from numpy.testing import assert_allclose
 import pytest
-import scipy.special as osp_special
 
-from jax import grad, jacobian, jit, lax, random, vmap
+from jax import jacobian, lax, random, vmap
 import jax.numpy as np
-from jax.scipy.special import expit
-from jax.util import partial
+from jax.scipy.special import expit, xlog1py, xlogy
 
 from numpyro.distributions.util import (
     binary_cross_entropy_with_logits,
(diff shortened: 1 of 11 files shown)


fehiepsi:
Blocked by https://github.com/google/jax/issues/1875

test/test_distributions_util.py
```diff
@@ -17,70 +15,8 @@
...
-
-@pytest.mark.parametrize('x, y, grad1, grad2', [
-    (np.array([1., 1., 1.]), np.array([1., 2., 3.]),
-     np.log(np.array([1, 2, 3])), np.array([1., 0.5, 1./3])),
-    (np.array([1.]), np.array([1., 2., 3.]),
-     np.sum(np.log(np.array([1, 2, 3]))), np.array([1., 0.5, 1./3])),
-    (np.array([1., 2., 3.]), np.array([2.]),
-     np.log(np.array([2., 2., 2.])), np.array([3.])),
-    (np.array([0.]), np.array([0., 0.]),
-     np.array([-float('inf')]), np.array([0., 0.])),
```
fehiepsi: Using upstream `xlogy` only makes the test for `grad2` here fail (the upstream version returns `nan`, which is caused by `0 / 0`). I think it should be fine to relax that semantic.

fehiepsi:
@neerajprad I think this is unblocked now. Let's see if tests pass.

neerajprad:
There's just one assertion failure for delta distribution where the resulting tensor isn't a `DeviceArray`.

fehiepsi:
Yes, in new version of JAX, broadcast_to a float will be a float.

PR #500, opened 2019-12-22
Deprecate Predictive.get_samples method

fehiepsi:
to be consistent with Pyro.


neerajprad:
Are we using `Predictive.get_samples` anywhere in the examples (that may need updating)?

fehiepsi:
Thanks, @neerajprad !

PR #502, opened 2019-12-25
Downgrade numpy version in setup and docs build

neerajprad:
See - https://github.com/google/jax/issues/1912

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,3 +1,3 @@
-jax>=0.1.37
-jaxlib>=0.1.18
+jax>=0.1.53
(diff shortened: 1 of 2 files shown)


neerajprad:
I'm still noticing issues with the build; will update once those are fixed.

fehiepsi:
I think you would need to ping jax/jaxlib versions. Currently, #496 is blocked.

neerajprad:
Thanks @fehiepsi, that was the issue. I think this should now pass.

neerajprad:
Ready to merge when build is green.

Issue #420, opened 2019-10-29
Support AutoLowRankMultivariateNormal autoguide


fehiepsi:
With LowRankMultivariateNormal distribution implemented at #403, we are ready to add the corresponding autoguide.

PR #503, opened 2019-12-27
Add auto lowrank guide

fehiepsi:
Addresses #420.

### DIFF
--- a/numpyro/contrib/autoguide/__init__.py
+++ b/numpyro/contrib/autoguide/__init__.py
@@ -347,6 +347,63 @@ class AutoMultivariateNormal(AutoContinuous):
         return self._unpack_and_constrain(latent, params)
 
 
+class AutoLowRankMultivariateNormal(AutoContinuous):
+    """
+    This implementation of :class:`AutoContinuous` uses a LowRankMultivariateNormal
+    distribution to construct a guide over the entire latent space.
+    The guide does not depend on the model's ``*args, **kwargs``.
+
+    Usage::
+
+        guide = AutoLowRankMultivariateNormal(model, rank=2, ...)
+        svi = SVI(model, guide, ...)
+    """
+    def __init__(self, model, prefix="auto", init_strategy=init_to_uniform(), rank=None):
+        self.rank = rank
+        super(AutoLowRankMultivariateNormal, self).__init__(
+            model, prefix=prefix, init_strategy=init_strategy)
+
+    def _sample_latent(self, base_dist, *args, **kwargs):
+        sample_shape = kwargs.pop('sample_shape', ())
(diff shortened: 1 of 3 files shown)


numpyro/contrib/autoguide/__init__.py
```diff
@@ -347,6 +347,63 @@ def quantiles(self, params, quantiles):
...
+        self.rank = rank
+        super(AutoLowRankMultivariateNormal, self).__init__(
+            model, prefix=prefix, init_strategy=init_strategy)
+
+    def _sample_latent(self, base_dist, *args, **kwargs):
+        sample_shape = kwargs.pop('sample_shape', ())
+        rank = int(round(self.latent_size ** 0.5)) if self.rank is None else self.rank
+        loc = numpyro.param('{}_loc'.format(self.prefix), self._init_latent)
+        cov_factor = numpyro.param('{}_cov_factor'.format(self.prefix), np.zeros((self.latent_size, rank)))
+        scale = numpyro.param('{}_scale'.format(self.prefix), np.ones(self.latent_size))
```
neerajprad: One thing that we found really helped stabilize learning was starting with a low value of `scale`; it was in fact critical for learning in many models. I would suggest having an `init_scale` parameter like in Pyro. That is however a more general comment since it applies to other autoguides too, so we can address this in a separate PR.

numpyro/contrib/autoguide/__init__.py
```diff
@@ -347,6 +347,63 @@ def quantiles(self, params, quantiles):
...
+        posterior = dist.LowRankMultivariateNormal(loc, cov_factor, cov_diag)
+        return numpyro.sample("_{}_latent".format(self.prefix), posterior, sample_shape=sample_shape)
+
+    def _get_transform(self, params):
+        loc = params['{}_loc'.format(self.prefix)]
+        cov_factor = params['{}_cov_factor'.format(self.prefix)]
+        scale = params['{}_scale'.format(self.prefix)]
+        cov_diag = scale * scale
+        cov_factor = cov_factor * scale[..., None]
+        scale_tril = dist.LowRankMultivariateNormal(loc, cov_factor, cov_diag).scale_tril
```
neerajprad: Do you think we should abstract out the `scale_tril` computation so that we can use that instead of initializing a distribution instance and doing unneeded checks only to compute this attribute?

fehiepsi: I think we just do this one time so it should be fine. :)

numpyro/contrib/autoguide/__init__.py
fehiepsi: Yes, let's address this in another PR.

Issue #419, opened 2019-10-28
Tutorial on predator-prey model


fehiepsi:
One of the coolest features of JAX is the [ode](https://github.com/google/jax/blob/master/jax/experimental/ode.py) submodule. Utilizing this module, I believe that we can run MCMC on models involving solving ode systems (e.g. the model in this [Stan case studies](https://mc-stan.org/users/documentation/case-studies/lotka-volterra-predator-prey.html#data-lynx-and-hare-pelts-in-canada). Probably we can replicate that excellent Stan case study and turn it into a tutorial.

PR #504, opened 2019-12-28
Add ode example

fehiepsi:
Addresses #419. The dataset is longer than the one in reference [1] but it seems that priors in [1] still work. However, I observed that there are 2 modes when running MCMC, depending on random seeds. :(

[ode_plot.pdf](https://github.com/pyro-ppl/numpyro/files/4006717/ode_plot.pdf)

[ode_plot1.pdf](https://github.com/pyro-ppl/numpyro/files/4006718/ode_plot1.pdf)

**TODO:**

- [x] Adjust priors to eliminate the first mode.

### DIFF
--- a/examples/ode.py
+++ b/examples/ode.py
@@ -84,14 +84,16 @@ def main(args):
     # predict populations
     y_pred = Predictive(model, mcmc.get_samples())(PRNGKey(2), data.shape[0])["y"]
     pop_pred = np.exp(y_pred)
-    plt.fill_between(year, data[:, 0], color="y", alpha=0.2, label="true hare")
-    plt.plot(year, data[:, 1], "kx", label="true lynx")
-    plt.plot(year, np.mean(pop_pred[..., 0], 0), "b-", label="pred hare")
-    plt.plot(year, np.mean(pop_pred[..., 1], 0), "g--", label="pred lynx")
+    mu, pi = np.mean(pop_pred, 0), np.percentile(pop_pred, (10, 90), 0)
+    plt.plot(year, data[:, 0], "ko", mfc="none", ms=4, label="true hare", alpha=0.67)
+    plt.plot(year, data[:, 1], "bx", label="true lynx")
+    plt.plot(year, mu[:, 0], "k-.", label="pred hare", lw=1, alpha=0.67)
+    plt.plot(year, mu[:, 1], "b--", label="pred lynx")
+    plt.fill_between(year, pi[0, :, 0], pi[1, :, 0], color="k", alpha=0.2)
+    plt.fill_between(year, pi[0, :, 1], pi[1, :, 1], color="b", alpha=0.3)
     plt.gca().set(ylim=(0, 160), xlabel="year", ylabel="population (in thousands)")
-    handles, labels = plt.gca().get_legend_handles_labels()
-    order = [3, 0, 1, 2]
-    plt.legend([handles[idx] for idx in order], [labels[idx] for idx in order])
+    plt.title("Posterior Predictive with 80% CI.")
+    plt.legend()
 
     plt.savefig("ode_plot.pdf")
     plt.tight_layout()
--- a/examples/ode.py
+++ b/examples/ode.py
@@ -45,7 +45,7 @@ def dz_dt(z, t, alpha, beta, gamma, delta):
     return np.stack([du_dt, dv_dt])
 
 
-ode_integrate = build_odeint(dz_dt, rtol=1e-5, atol=1e-3, mxstep=500)
+predator_prey_int = build_odeint(dz_dt, rtol=1e-5, atol=1e-3, mxstep=500)
 
 
 def model(N, y=None):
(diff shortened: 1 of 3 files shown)


examples/ode.py
```diff
@@ -0,0 +1,112 @@
...
+    z = predator_prey_int(z_init, ts, *theta)
+    # measurement errors, we expect that measured hare has larger error than measured lynx
+    sigma = numpyro.sample("sigma", dist.Exponential(np.array([1, 2])))
+    # measured populations (in log scale)
+    numpyro.sample("y", dist.Normal(np.log(z), sigma), obs=y)
+
+
+def main(args):
+    URL = "http://people.whitman.edu/~hundledr/courses/M250F03/LynxHare.txt"
+    data = pd.read_csv(URL, sep=" (?!$)", names=["year", "hare", "lynx"], index_col=0, engine="python")
```
fehiepsi: @neerajprad could you help me upload the following csv file (or .txt file) to the cloud?
```
import pandas as pd
URL = "http://people.whitman.edu/~hundledr/courses/M250F03/LynxHare.txt"
data = pd.read_csv(URL, sep=" (?!$)", names=["year", "hare", "lynx"], index_col=0, engine="python")
data.to_csv("LynxHare.csv")
```
Thanks! I'll write a reader for the file to remove pandas requirement in this example (currently, this makes test fail).

numpyro/examples/datasets.py
```diff
@@ -53,6 +53,11 @@
 ])
 
 
+LYNXHARE = dset('lynxhare', [
+    'http://people.whitman.edu/~hundledr/courses/M250F03/LynxHare.txt',
```
fehiepsi: @neerajprad Could you help me upload this dataset to cloud?

neerajprad: link - https://d2hg8soec8ck9v.cloudfront.net/datasets/LynxHare.txt

examples/ode.py
```diff
@@ -0,0 +1,112 @@
...
+    plt.fill_between(year, data[:, 0], color="y", alpha=0.2, label="true hare")
+    plt.plot(year, data[:, 1], "kx", label="true lynx")
+    plt.plot(year, np.mean(pop_pred[..., 0], 0), "b-", label="pred hare")
+    plt.plot(year, np.mean(pop_pred[..., 1], 0), "g--", label="pred lynx")
+    plt.gca().set(ylim=(0, 160), xlabel="year", ylabel="population (in thousands)")
+    handles, labels = plt.gca().get_legend_handles_labels()
+    order = [3, 0, 1, 2]
+    plt.legend([handles[idx] for idx in order], [labels[idx] for idx in order])
+
+    plt.savefig("ode_plot.pdf")
```
neerajprad: Also add this example and the figure to the gallery?

examples/ode.py
```diff
@@ -0,0 +1,112 @@
...
+    mcmc = MCMC(NUTS(model, dense_mass=True),
+                args.num_warmup, args.num_samples, num_chains=args.num_chains,
+                progress_bar=False if "NUMPYRO_SPHINXBUILD" in os.environ else True)
+    mcmc.run(PRNGKey(1), N=data.shape[0], y=np.log(data))
+    mcmc.print_summary()
+
+    # predict populations
+    y_pred = Predictive(model, mcmc.get_samples())(PRNGKey(2), data.shape[0])["y"]
+    pop_pred = np.exp(y_pred)
+    plt.fill_between(year, data[:, 0], color="y", alpha=0.2, label="true hare")
```
neerajprad: Could you explain why are we filling in here (I thought these values are fixed measurements)? Also, could you add the predictive CI to the plots?

numpyro/examples/datasets.py
fehiepsi: Thanks, @neerajprad !

examples/ode.py
fehiepsi: It will be added automatically. We have filtered only a few examples: https://github.com/pyro-ppl/numpyro/blob/master/notebooks/source/conf.py#L180

examples/ode.py
fehiepsi: I just make sure that the plot is easy to observe: https://github.com/pyro-ppl/numpyro/files/4006718/ode_plot1.pdf. Let me see if adding CI is helpful for the plot.

neerajprad: I would suggest using a different line style. Fill makes it seem like it could be anywhere between 0 to actual value. We should use a fill for the CI though (I think that will be nice to have as a posterior predictive check).

examples/ode.py
neerajprad: Oh that's great, I didn't realize that.

examples/ode.py
fehiepsi: Actually, I mimic this figure :D https://en.wikipedia.org/wiki/Lotka%E2%80%93Volterra_equations#/media/File:Milliers_fourrures_vendues_en_environ_90_ans_odum_1953_en.jpg

fehiepsi: Screenshot from 2020-01-14 17-15-12
This is the plot with 60% CI. It is hard to appreciate hare's PI. Should I remove it? I want to have hare and lynx in the same plot to illustrate predator-prey pattern.

neerajprad: I think it might look nicer if we replace the true hare population with points or a line and use a light shade for the CI (that's kind of the point of using numpyro instead of just getting point estimates, so will be good to have in the figure). Feel free to commit your changes, and I can try to play around with the plot settings too.

fehiepsi: yeah, let me see if I can come up with something cleaner with CI. :)

fehiepsi: Screenshot from 2020-01-15 11-16-15

It looks much better now. Thanks for your suggestions!

examples/ode.py
fehiepsi: The downside is when building the website, if the content of a script is changed, it will rerun the script to capture the output. I guess it is fine given than we only need to update the examples when we pump up the version (actually, it will be a good time to verify that we still get correct outputs). :)

examples/ode.py
neerajprad: Looks great, thanks!

Issue #497, opened 2019-12-18
[FR] sample adaptive MCMC


fehiepsi:
Requested by @martinjankowiak ! This is a nice sampler to have. For me, this is a good chance to validate our MCMCKernel interface.

Reference: https://papers.nips.cc/paper/9107-sample-adaptive-mcmc.pdf

If the performance is good, we need to push this to Pyro too.

fehiepsi:
@martinjankowiak If you want to implement this, please go ahead! Otherwise, I can work on this this weekend.

martinjankowiak:
i don't have time to do this right now but happy to look at a PR!

PR #508, opened 2020-01-06
implement sample adaptive mcmc

fehiepsi:
Addresses #497.

The interface would be very similar to HMC. Here, I made two personal choices:
+ if adapt_state_size N is None, N = 2d, where d is latent size,
+ current sample will be drawn uniformly from the proposal set of N points.

Tasks

- [x] Implement the sampler
- [x] Add some tests to verify the implementation

Future PRs:

- add an example for benchmarking

Blocked by #496


numpyro/infer/mcmc.py
```diff
@@ -247,11 +253,11 @@ def init_kernel(init_params,
                                             target_accept_prob=target_accept_prob,
                                             find_reasonable_step_size=find_reasonable_ss)
 
-        rng_key_hmc, rng_key_wa = random.split(rng_key)
+        rng_key_hmc, rng_key_wa, rng_key_momentum = random.split(rng_key, 3)
```
fehiepsi: This split gives some failing tests. So I changed their thresholds a bit.

martinjankowiak: can you please explain further?

numpyro/infer/mcmc.py
```diff
@@ -615,6 +621,257 @@ def __init__(self,
...
+            else np.sqrt(inverse_mass_matrix)
+        if adapt_state_size is None:
+            # XXX: heuristic choice
+            adapt_state_size = 2 * z_flat.shape[-1]
+        else:
+            assert adapt_state_size > 1, 'adapt_state_size should be greater than 1.'
+        # NB: mean is init_params
+        zs = z_flat + _sample_proposal(inv_mass_matrix_sqrt, rng_key_zs, (adapt_state_size,))
+        # compute potential energies
+        pes = lax.map(lambda z: pe_fn(unravel_fn(z)), zs)
```
martinjankowiak: can you comment on which quantities are/are not cached from step to step?

numpyro/infer/mcmc.py
```diff
@@ -615,6 +621,257 @@ def __init__(self,
         self._algo = 'NUTS'
 
 
+def _get_proposal_loc_and_scale(samples, loc, scale, new_sample):
```
martinjankowiak: can you please add more comments to this method? will be easier to review

numpyro/infer/mcmc.py
```diff
@@ -615,6 +621,257 @@ def __init__(self,
...
+        # XXX: numpy.delete is not available in JAX
+        zs = np.where(np.arange(zs.shape[0])[:, None] < j, zs_[:-1], zs_[1:])
+        pes = np.where(np.arange(pes.shape[0]) < j, pes_[:-1], pes_[1:])
+        loc = locs[j]
+        inv_mass_matrix_sqrt = scales[j]
+        adapt_state = SAAdaptState(zs, pes, loc, inv_mass_matrix_sqrt)
+
+        # XXX: we make a modification of SA sampler in [1]
+        # in [1], each MCMC state contains N points
+        # here we do resampling to pick randomly a point from N points
```
martinjankowiak: can you please comment further?

numpyro/infer/mcmc.py
```diff
@@ -615,6 +621,257 @@ def __init__(self,
...
+        If model is provided, `potential_fn` will be inferred using the model.
+    :param potential_fn: Python callable that computes the potential energy
+        given input parameters. The input parameters to `potential_fn` can be
+        any python collection type, provided that `init_params` argument to
+        `init_kernel` has the same type.
+    :param int adapt_state_size: The number of points to generate proposal
+        distribution. Defaults to 2 times latent size.
+    :param bool dense_mass:  A flag to decide if mass matrix is dense or
+        diagonal (default to ``dense_mass=True``)
+    :param callable init_strategy: a per-site initialization function.
```
martinjankowiak: can you describe how you're setting the N initial samples?

numpyro/infer/mcmc.py
fehiepsi: This issue happens at `hmc` function (not relevant to this PR). We missed a split to generate a key for the initial `r`. Because this line will affect the following random generator numbers, so some strict tests will fail.

numpyro/infer/mcmc.py
fehiepsi: All information of the set S (the MCMC state of SA sampler) will be cached in `adapt_state`:
+ zs: values of the samples in S
+ pes: potential energies of them (we don't need to recompute those)
+ loc: mean of S
+ scale: scale of S

numpyro/infer/mcmc.py
fehiepsi: In Algorithm 1, line 10 of the paper, the author returns the collections of num_samples x N samples (denoted by z[k, n]). The posterior will contain all those points, and for each quantity `f`, `E(f(z)) = \int f(z)p(z)dz` will be computed as
`1/num_samples sum_k (1/N sum_n f(z[k,n]))`. I think that we can just simply compute that quantity by `1 / num_samples sum_k f(z[k])` where `z[k]` is randomly picked from N points `z[k,n]` (I think it is fine because they have uniform weights).

numpyro/infer/mcmc.py
fehiepsi: done; this function is also tested in `test_get_proposal_loc_and_scale`. :)

numpyro/infer/mcmc.py
martinjankowiak: why would you want to do this? you may lose information (samples)

numpyro/infer/mcmc.py
```diff
@@ -615,6 +621,260 @@ def __init__(self,
...
+        zs_ = np.concatenate([z[None, :], zs])
+        pes_ = np.concatenate([pe[None], pes])
+        if inv_mass_matrix_sqrt.ndim == 2:  # dense_mass
+            log_weights = dist.MultivariateNormal(locs, scale_tril=scales).log_prob(zs_) + pes_
+        else:
+            log_weights = dist.Normal(locs, scales).log_prob(zs_).sum(-1) + pes_
+        log_weights = np.where(np.isnan(log_weights), -np.inf, log_weights)
+        j = categorical_logits(rng_key_reject, log_weights)
+        # XXX: numpy.delete is not available in JAX
+        zs = np.where(np.arange(zs.shape[0])[:, None] < j, zs_[:-1], zs_[1:])
```
martinjankowiak: should this complex logic be a helper fn?

test/test_mcmc.py
```diff
@@ -16,24 +16,27 @@
...
 from numpyro.util import fori_collect
 
 
-@pytest.mark.parametrize('kernel_cls', [HMC, NUTS])
+@pytest.mark.parametrize('kernel_cls', [HMC, NUTS, SA])
 @pytest.mark.parametrize('dense_mass', [False, True])
 def test_unnormalized_normal_x64(kernel_cls, dense_mass):
     true_mean, true_std = 1., 0.5
-    warmup_steps, num_samples = 1000, 8000
+    warmup_steps, num_samples = (100000, 100000) if kernel_cls is SA else (1000, 8000)
```
martinjankowiak: how long do these new tests take?

fehiepsi: They are super fast (a few seconds for all of them as long as we disable progbar). A few observations are:
+ n_eff/num_samples is small (which indicates that the samples are correlated).
+ n_eff/s seems to be better than NUTS.

So with the same num_samples, this will not be as good as NUTS. But given the same amount of time, this can be better than NUTS.

numpyro/infer/mcmc.py
fehiepsi: agreed! I'll separate it out.

numpyro/infer/mcmc.py
fehiepsi: I mainly do this because of several reasons:
+ We want to keep the same MCMC API (otherwise, many logics will need to be revised so that the returned samples will have shape num_chains x num_samples x adapt_state_size x sample_shape). This is the main reason.
+ We don't need so many samples. Typically, SA will give us 100,000,000 - 1000,000,000  samples (from the experimental section of the reference) with num_samples=1000,000 and adapt_state_size=100-1000. I think 1000,000 samples are already large enough for prediction tasks. :D
+ I think if we want more samples, using multi-chains will be better. This SA is pretty suitable for the `vectorized` chain method.

martinjankowiak: ok that makes sense. can we put a comment to that effect in the main docstring? something like "Note that unlike in reference [1], we return a randomly selected (i.e. thinned) subset of approximate posterior samples of size num_chains x num_samples instead of num_chains x num_samples x adapt_state_size."

fehiepsi: Thanks, I'll add that.

numpyro/infer/mcmc.py
```diff
@@ -615,6 +621,302 @@ def __init__(self,
         self._algo = 'NUTS'
 
 
+def _get_proposal_loc_and_scale(samples, loc, scale, new_sample):
+    # get loc/scale of q_{-n} (Algorithm 1, line 5 of ref [1]) for n from 1 -> N
+    # these loc/scale will be stacked to the first dim; so
+    #   proposal_loc.shape[0] = proposal_loc.shape[0] = N
+    # Here, we use the numerical stability procedure in Appendix 6 of [1].
+    weight = 1 / samples.shape[0]
```
neerajprad: I think this file is getting a bit big, and it may be worth having a separate mcmc module with different files for HMC/NUTS and SA.

numpyro/infer/mcmc.py
neerajprad: Makes sense. We can include this in both HMC as well as this class, though I'm not sure how useful it will be to be able to customize this.

fehiepsi: Me too. Let's wait for some user requests before adding it, or if we find setting initial inverse_mass_matrix is useful in our experiments.

PR #509, opened 2020-01-06
Implement cholesky 1-rank update/downdate

fehiepsi:
This PR adds a utility that is necessary for Sample Adaptive MCMC #508. (In Sample Adaptive, we need to update Cholesky of covariance matrix when a sample is added to or removed from a collection of samples.)

### DIFF
--- a/numpyro/distributions/util.py
+++ b/numpyro/distributions/util.py
@@ -206,6 +206,9 @@ def xlog1py(x, y):
 defjvp(xlog1py, _xlog1py_jvp_lhs, _xlog1py_jvp_rhs)
 
 
+# TODO: rename this function, in PyTorch cholesky_inverse computes the inverse using a cholesky
+# input; here we want to take cholesky of the inverse... (the name here is better but
+# we need to change to avoid confusion)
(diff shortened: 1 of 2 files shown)


numpyro/distributions/util.py
```diff
@@ -282,6 +285,42 @@ def vec_to_tril_matrix(t, diagonal=0):
...
+
+        1. A more efficient rank-one covariance matrix update for evolution strategies,
+           Oswin Krause and Christian Igel
+    """
+    batch_shape = lax.broadcast_shapes(L.shape[:-2], x.shape[:-1])
+    L = np.broadcast_to(L, batch_shape + L.shape[-2:])
+    x = np.broadcast_to(x, batch_shape + x.shape[-1:])
+    diag = np.diagonal(L, axis1=-2, axis2=-1)
+    # convert to unit diagonal triangular matrix: L @ D @ T.t
+    L = L / diag[..., None, :]
```
fehiepsi: The implementation is a simplified version of Algorithm 3.1 of Ref [1] (here entries of L are divided by diagonal part by default).

Issue #506, opened 2019-12-30
numpyro equivalent to Pyro's poutine.mask


FlorianWilhelm:
For me, it seems that numpyro's `primitives` module is missing right now the functionality of Pyro's [poutine.mask](http://docs.pyro.ai/en/stable/poutine.html#pyro.poutine.handlers.mask) in order to mask out some of the sample statements elementwise. This is very convenient when dealing with missing values or if some padding is needed to align for instance time series.

neerajprad:
Thanks for creating this feature request. We usually wait for internal use cases and user requests to prioritize new features. We also have a use case for `poutine.mask`, and we'll plan to have it in the next release.

VincentBt:
Also very interested!
To do things like
```
z_t = pyro.sample("z_%d" % t,
                  dist.Normal(z_loc, z_scale)
                      .mask(mini_batch_mask[:, t - 1:t])
                      .to_event(1))
```

fehiepsi:
This is addressed in #515. For masked distribution, please refer to #520.

PR #515, opened 2020-01-10
Add support for mask handler

neerajprad:
Addresses #506.

### DIFF
--- a/numpyro/infer/util.py
+++ b/numpyro/infer/util.py
@@ -63,6 +63,9 @@ def log_density(model, model_args, model_kwargs, params, skip_dist_transforms=Fa
             intermediates = site['intermediates']
             mask = site['mask']
             scale = site['scale']
+            # Early exit when all elements are masked
+            if not_jax_tracer(mask) and mask is not None and not np.any(mask):
+                return jax.device_put(0.), model_trace
             if intermediates:
                 if skip_dist_transforms:
                     log_prob = site['fn'].base_dist.log_prob(intermediates[0][0])
@@ -72,7 +75,8 @@ def log_density(model, model_args, model_kwargs, params, skip_dist_transforms=Fa
                 log_prob = site['fn'].log_prob(value)
 
             # Minor optimizations
-            # XXX: note that this will not work for dynamic masks.
+            # XXX: note that this may not work correctly for dynamic masks, provide
+            # explicit jax.DeviceArray for masking.
             if mask is not None:
                 if scale is not None:
                     log_prob = np.where(mask, scale * log_prob, 0.)
--- a/numpyro/infer/util.py
+++ b/numpyro/infer/util.py
@@ -2,7 +2,7 @@ from functools import partial
 import warnings
 
 import jax
-from jax import device_get, lax, random, value_and_grad, vmap
+from jax import device_get, device_put, lax, random, value_and_grad, vmap
 from jax.flatten_util import ravel_pytree
 import jax.numpy as np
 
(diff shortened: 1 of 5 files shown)


numpyro/infer/util.py
```diff
@@ -61,16 +61,20 @@ def log_density(model, model_args, model_kwargs, params, skip_dist_transforms=Fa
         if site['type'] == 'sample':
             value = site['value']
             intermediates = site['intermediates']
+            mask = site['mask']
+            scale = site['scale']
+            # exit early if mask all
+            if not np.any(mask):
```
fehiepsi: I think you would need to check `not_jax_tracer` here.

numpyro/infer/util.py
```diff
@@ -61,16 +61,20 @@ def log_density(model, model_args, model_kwargs, params, skip_dist_transforms=Fa
...
             if intermediates:
                 if skip_dist_transforms:
                     log_prob = site['fn'].base_dist.log_prob(intermediates[0][0])
                 else:
                     log_prob = site['fn'].log_prob(value, intermediates)
             else:
                 log_prob = site['fn'].log_prob(value)
+            log_prob = np.where(mask, scale * log_prob, 0.)
             log_prob = np.sum(log_prob)
-            if 'scale' in site:
```
fehiepsi: WDYT about this pattern? I guess it would be nice to have this check to save some computations, but I don't recall why we have `scale` (and `mask`) in `sample` primitive by default. :(

neerajprad: I think it is okay to have some default value for `scale` or `mask`, and we can always add other optimizations like in Pyro where we check if the scale is identically zero or whether all values are masked out and treat these cases specially. I found no performance difference without these minor optimizations on the few examples I ran in NumPyro. e.g. in this case we only end up multiplying by a scalar 1 which probably makes no difference under JIT (in fact, we'll need to use `lax.cond` to apply these optimizations correctly under JIT).

numpyro/infer/util.py
neerajprad: Thanks for noting this; I think these optimizations need to be under `lax.cond` if we want to apply them under JIT. I'll also add a test for this.

fehiepsi: I think lax.cond does not have grad rule yet. I think you can just multiply for now.

neerajprad: I see, okay let me run some investigations at my end and update.

neerajprad: So I added some of these optimizations back by using `None` as the default value for mask and scale. Also added a test for jit. I have removed this particular early exit option, since we currently don't need it, i.e. we don't have use cases where we put `mask=False` (we can use `block` if needed). However, as I mentioned, this may not work correctly for dynamic masks under JIT, i.e. if for the same site, the mask changes from `None` to some bool array, this will give a wrong answer under JIT, but I think the only way to avoid that is to not apply any of these optimizations.

numpyro/infer/util.py
```diff
@@ -61,16 +61,27 @@ def log_density(model, model_args, model_kwargs, params, skip_dist_transforms=Fa
...
             if intermediates:
                 if skip_dist_transforms:
                     log_prob = site['fn'].base_dist.log_prob(intermediates[0][0])
                 else:
                     log_prob = site['fn'].log_prob(value, intermediates)
             else:
                 log_prob = site['fn'].log_prob(value)
+
+            # Minor optimizations
+            # XXX: note that this will not work for dynamic masks.
```
fehiepsi: I think this is fine. Users can provide ndarray for dynamic masks and this is JAX restriction, not NumPyro. I believe we can also have a check if mask is False to return early but it might not be important and can cause confusion (the benefit is to skip log prob computation for e.g. deterministic sites). WDYT?

neerajprad: I think for that use case we can probably just use `handlers.block`? Otherwise, I'm not sure how that will work under JIT without resorting to `lax.cond`.

neerajprad: Okay done.

fehiepsi: Oh, I didn't mean that but your solution works much better. Your solution is great!

fehiepsi:
Thanks, @neerajprad! This is a nice handler to have.

PR #516, opened 2020-01-10
Update license to Apache 2

neerajprad:
Addresses https://github.com/pyro-ppl/pyro/issues/2249.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -1,3 +1,6 @@
+# Copyright (c) 2017-2020 Contributors to the Pyro project.
+# SPDX-License-Identifier: Apache-2.0
+
 import os
 import sys
 
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -1,4 +1,4 @@
-# Copyright Contributors to the Pyro project.
+# Copyright (c) 2017-2019 Uber Technologies, Inc.
 # SPDX-License-Identifier: Apache-2.0
 
 import os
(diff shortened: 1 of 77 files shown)


docs/source/conf.py
```diff
@@ -1,3 +1,6 @@
+# Copyright (c) 2017-2020 Contributors to the Pyro project.
```
fritzo: Let's just use "Copyright Contributors to the Pyro project" as recommended by Steve Winslow

neerajprad: Okay, done.

numpyro/compat/util.py
```diff
@@ -1,3 +1,6 @@
+# Copyright Contributors to the Pyro project.
+# SPDX-License-Identifier: Apache-2.0
+
```
fritzo: need an extra newline here for flake8

numpyro/patch.py
```diff
@@ -1,3 +1,6 @@
+# Copyright Contributors to the Pyro project.
+# SPDX-License-Identifier: Apache-2.0
+
```
fritzo: ditto: add an extra newline

neerajprad: Oh I missed this, thanks for catching this. Will update.

setup.py
```diff
@@ -1,3 +1,6 @@
+# Copyright Contributors to the Pyro project.
+# SPDX-License-Identifier: Apache-2.0
+
```
fritzo: Can you also update setup.py as
```diff
- 'License :: OSI Approved :: MIT License',
+ 'License :: OSI Approved :: Apache License 2.0',
```

PR #517, opened 2020-01-12
Fix handler mask trigger GPU platform by default

fehiepsi:
We should not set the default arguments to jax.numpy device array. Otherwise, we can't change the default platform (GPU).


neerajprad:
Could you expand on what you meant by "we should not set the default arguments to jax.numpy device array"?

fehiepsi:
If we have some `jax.numpy` code globally, GPU will be set and we can't change to use CPU by default. I faced this issue while running test code for SA sampler (pytest runs in GPU instead of CPU).

Issue #510, opened 2020-01-08
change init scale of auto_guide for stability


fehiepsi:
Let's add an arg `init_scale=0.1` to auto guides like Pyro.

PR #521, opened 2020-01-14
add init scale to some autoguide

fehiepsi:
Addresses #510. We follow Pyro to set init scale to 0.1.


neerajprad:
There are a couple of test failures, I think we may need to adjust those tests.

PR #522, opened 2020-01-15
Add deterministic primitive

### LINKED ISSUE #462: numpyro equivalent to pymc3's Deterministic
First of all, thank you for this excellent work!

I'm trying to implement the Finnish horseshoe prior (Piironen, J., and Vehtari, A., 2017).
Edit: I should credit Michael Betancourt for his implementation of the Finnish HS [here](https://betanalpha.github.io/assets/case_studies/bayes_sparse_regression.html).
I'm able to calculate the horseshoe as follows:

```python
def finnish_horseshoe(M, m0, N, var, half_slab_df, slab_scale2, tau_tilde, c2_tilde, lambd, beta_tilde):
    
    tau0 = (m0/(M-m0) * (np.sqrt(var)/np.sqrt(1.0*N)))
    tau = tau0 * tau_tilde
    c2 = slab_scale2 * c2_tilde
    lambd_tilde = np.sqrt(c2 * lambd**2 / (c2 + tau**2 * lambd**2))
    beta = tau * lambd_tilde * beta_tilde
    
    return(beta)

def model(y_vals,
          design_mat,
          N, # number of y_vals
          slab_df=1,
          slab_scale=1,
          expected_large_covar_num=10):

    a_prior = dist.Normal(10., 10.)
    a = numpyro.sample("alpha", a_prior)

    # update from pymc3
    half_slab_df = slab_df/2
    variance = y_vals.var()
    slab_scale2 = slab_scale**2
    hs_shape = (design_mat.shape[1],)
    
    beta_tilde = numpyro.sample('beta_tilde', dist.Normal(0., 1.), sample_shape=hs_shape)
    lambd = numpyro.sample('lambd', dist.HalfCauchy(1.), sample_shape=hs_shape)
    tau_tilde = numpyro.sample('tau_tilde', dist.HalfCauchy(1.))
    c2_tilde = numpyro.sample('c2_tilde', dist.InverseGamma(half_slab_df, half_slab_df))
    
    bC = finnish_horseshoe(M = hs_shape[0],
                                    m0 = expected_large_covar_num,
                                    N = N,
                                    var = variance,
                                    half_slab_df = half_slab_df, 
                                    slab_scale2 = slab_scale2,
                                    tau_tilde = tau_tilde,
                                    c2_tilde = c2_tilde,
                                    lambd = lambd, 
                                    beta_tilde = beta_tilde)

    C = np.dot(design_mat, bC)
    mu = a + C

    sig_prior = dist.Exponential(1.)
    sigma = numpyro.sample('sigma', sig_prior)
    return numpyro.sample('obs', dist.Normal(mu, sigma), obs=y_vals)
```

The problem I have is that the sampled values of `bC` are not tracked during sampling. In pymc3 I can track samples of `bC` using `pm.Deterministic`. Is there a numpyro equivalent, or would I have to implement a new distribution class?

Thanks for your help!

neerajprad:
Fixes #462. 

This adds a new primitive site type called `deterministic` which has the benefit that it will be neglected by most effect handlers except `trace`, so deterministic sites won't broadcast under `plate` and it will be cheaper to include these in the model.

Note that this makes a design decision so that the values for deterministic sites are only accessible via `Predictive`. An alternative (and what I had in an earlier commit) would store deterministic nodes as HMC is run and was directly available via `mcmc.get_samples`. However, for some reason there were some serious perf issues with NUTS and SA (I think compilation takes a really long time even though these nodes are shouldn't have any operation recorded by the tracer, but I didn't investigate in detail). The current version has no performance issue, but has a downside that users need to call `Predictive` to access deterministic nodes in the model and these sites are not printed in model summary. If these seem like big cons, we can modify `get_samples` to use `Predictive` if a model has deterministic nodes, but I would like to go with this simpler solution first.

### DIFF
--- a/numpyro/handlers.py
+++ b/numpyro/handlers.py
@@ -184,7 +184,7 @@ class replay(Messenger):
         super(replay, self).__init__(fn)
 
     def process_message(self, msg):
-        if msg['name'] in self.guide_trace:
+        if msg['name'] in self.guide_trace and msg['type'] in ('sample', 'plate'):
             msg['value'] = self.guide_trace[msg['name']]['value']
 
 
@@ -423,6 +423,8 @@ class substitute(Messenger):
         super(substitute, self).__init__(fn)
 
     def process_message(self, msg):
+        if msg['type'] not in ('sample', 'param'):
+            return
         if self.param_map is not None:
             if msg['name'] in self.param_map:
                 msg['value'] = self.param_map[msg['name']]
--- a/numpyro/primitives.py
+++ b/numpyro/primitives.py
@@ -75,8 +75,8 @@ def sample(name, fn, obs=None, rng_key=None, sample_shape=()):
         state to `fn`. In those situations, `rng_key` keyword will take no
         effect.
 
-    :param str name: name of the sample site
-    :param fn: Python callable
+    :param str name: name of the sample site.
+    :param fn: a stochastic function that returns a sample.
     :param numpy.ndarray obs: observed value
     :param jax.random.PRNGKey rng_key: an optional random key for `fn`.
     :param sample_shape: Shape of samples to be drawn.
@@ -140,6 +140,31 @@ def param(name, init_value=None, **kwargs):
         'mask': None,
         'scale': None,
         'cond_indep_stack': [],
+    }
+
+    # ...and use apply_stack to send it to the Messengers
+    msg = apply_stack(initial_msg)
+    return msg['value']
+
+
+def deterministic(name, value):
+    """
+    Used to designate deterministic sites in the model. Note that most effect
+    handlers will not operate on deterministic sites (except
+    :function:`~numpyro.handlers.trace`), so deterministic sites should be
+    side-effect free. The use case for deterministic nodes is to record any
+    values in the model execution trace.
+
+    :param str name: name of the deterministic site.
+    :param numpy.ndarray value: deterministic value to record in the trace.
+    """
+    if not _PYRO_STACK:
+        return value
+
+    initial_msg = {
+        'type': 'deterministic',
+        'name': name,
+        'value': value,
     }
 
     # ...and use apply_stack to send it to the Messengers
@@ -233,6 +258,8 @@ class plate(Messenger):
         return tuple(batch_shape)
 
     def process_message(self, msg):
+        if msg['type'] not in ('sample', 'plate'):
+            return
         cond_indep_stack = msg['cond_indep_stack']
         frame = CondIndepStackFrame(self.name, self.dim, self.subsample_size)
         cond_indep_stack.append(frame)
--- a/numpyro/__init__.py
+++ b/numpyro/__init__.py
@@ -4,7 +4,7 @@
 from numpyro import compat, diagnostics, distributions, handlers, infer, optim
 from numpyro.distributions.distribution import enable_validation, validation_enabled
 import numpyro.patch  # noqa: F401
-from numpyro.primitives import factor, module, param, plate, sample
+from numpyro.primitives import deterministic, factor, module, param, plate, sample
 from numpyro.util import enable_x64, set_host_device_count, set_platform
 from numpyro.version import __version__
 
@@ -14,6 +14,7 @@ set_platform('cpu')
 __all__ = [
     '__version__',
     'compat',
+    'deterministic',
     'diagnostics',
     'distributions',
     'enable_x64',
--- a/numpyro/infer/elbo.py
+++ b/numpyro/infer/elbo.py
@@ -57,10 +57,8 @@ class ELBO(object):
             seeded_model = seed(model, model_seed)
             seeded_guide = seed(guide, guide_seed)
             guide_log_density, guide_trace = log_density(seeded_guide, args, kwargs, param_map)
-            # NB: we only want to substitute params not available in guide_trace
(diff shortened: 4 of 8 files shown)


numpyro/handlers.py
```diff
@@ -184,7 +184,7 @@ def __init__(self, fn, guide_trace):
         super(replay, self).__init__(fn)
 
     def process_message(self, msg):
-        if msg['name'] in self.guide_trace:
+        if msg['name'] in self.guide_trace and msg['type'] in ('sample', 'plate'):
```
fehiepsi: Will this work if both `model` and `guide` have the same `params` site? I think it will work if we change the logic at [these lines](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/infer/elbo.py#L60-L63) (basically removing `model_param_map` there). I think the performance would be the same but you certainly have a better idea than me regarding this. WDYT?

numpyro/primitives.py
```diff
@@ -147,6 +147,31 @@ def param(name, init_value=None, **kwargs):
...
+def deterministic(name, value):
+    """
+    Used to designate deterministic sites in the model. Note that most effect
+    handlers will not operate on deterministic sites (except
+    :function:`~numpyro.handlers.trace`), so deterministic sites should be
+    side-effect free. The use case for deterministic nodes is to record any
+    values in the model execution trace.
+
+    :param str name: name of the deterministic site.
+    :param numpy.ndarray value: deterministic value to record in the trace.
```
fehiepsi: It would be nice if this can support a `callable` to lazily get those values (similar to `pyro.param(..., lambda: ...)`).

fehiepsi:
Both solutions look great to me! For the first one, I think that we should defer `deterministic` to `constrain_fn`, which only runs after we collect all unconstrained samples. I think it will work if you add
```
if v['type'] == 'deterministics':
   has_transformed_dist = True
```
after [this line](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/infer/util.py#L381).

numpyro/primitives.py
neerajprad: Could you elaborate on this? I am not sure about why this needs to be lazy.

neerajprad:
I think your first solution is quite clever, and I quite like it! I am however wary that subtly changing `constrain_fn` to handle this use case and returning deterministic sites too could lead to other bugs and add to unnecessary checks in classes like SVI which make use of `constrain_fn`. I realize that right now we are repeating some computation in getting those deterministic sites, but it seems quite fast and the code is separate enough that we can later change it, if needed. We can also write a separate `postprocess_fn` which calls `constrain_fn` and also returns deterministic sites and is only used by `mcmc`, which should address your concern. WDYT?

numpyro/handlers.py
neerajprad: Hmm..the only reason why I added this is because many of our effect handlers should implicitly operate on a subset of site types but we haven't specified that. Now that we have a few different types, I thought it will be good to specify that. `replay` shouldn't apply to param sites, but `substitute` is more general and allows for that. Let me know if I'm missing some use case you had in mind.

numpyro/primitives.py
fehiepsi: There are two usage cases of `deterministic`:
+ record a value
+ record a transformed value

This PR supports the first usage case. But it will incur unnecessary overhead while running inference for the second usage case (unless users use a flag to disable the computation for inference). Nevermind though, now I think it is better to use a flag for those lazy values. It is always better to be explicit than introducing a library-knowledge pattern (especially when users ask: what is the output of those deterministic sites if its value is lazily evaluated). :)

fehiepsi:
That is a good plan to me!

Btw, the proposed change only affects MCMC because the logic only affects [get_potential_fn](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/infer/util.py#L399).

numpyro/handlers.py
fehiepsi: At the logic in my previous comment, we filter out `params` available in `guide` and substitute to `model`. If we only replay `sample` sites here, for those params available in `guide`, we'll neither substitute or replay. I think if we add those checks, then we would need to `substitute` all params (including those params available in `guide`).

fehiepsi:
Opss, there is still an issue regarding `replay` that we need to discuss.

numpyro/handlers.py
neerajprad: I see. I just removed `model_param_map`, though please correct me if that's an oversimplification and I overlooked something. 😄

neerajprad:
> Btw, the proposed change only affects MCMC because the logic only affects get_potential_fn.

Done; this seems pretty reasonable to me. I think we can rename `constrain_fn` to `postprocess_fn` in `mcmc` for clarity. I have commented out the `print_summary` in the test since that adds considerable additional time specially for SA.

numpyro/handlers.py
fehiepsi: I think it is correct. :)

fehiepsi:
> I think we can rename constrain_fn to postprocess_fn in mcmc for clarity.

Good idea! Let's do this.

Sorry about `constrain_fn` stuff, I thought we don't need to change there but I was wrong. About SA, its slowness makes sense because the array is big: 100,000 x 3000. We shouldn't even print out the stats of those `3000` sites I believe. How about having a flag in `print_summary` to ignore these sites by default?

PR #524, opened 2020-01-16
Minor follow up issues with numpyro.deterministic

neerajprad:
- exclude `deterministic` sites from `print_summary`.
 - change `constrain_fn` to `postprocess_fn` in `mcmc` since the function is effectively postprocessing unconstrained samples and returning deterministic sites.
 - change `infer.util.constrain_fn` to not return deterministic sites by default so that we can avoid checks / bugs in other parts of the codebase that uses this.

### DIFF
--- a/numpyro/infer/mcmc.py
+++ b/numpyro/infer/mcmc.py
@@ -359,10 +359,11 @@ class MCMCKernel(ABC):
     Defines the interface for the Markov transition kernel that is
     used for :class:`~numpyro.infer.MCMC` inference.
     """
-    def constrain_fn(self, model_args, model_kwargs):
+    def postprocess_fn(self, model_args, model_kwargs):
         """
         Function that transforms unconstrained values at sample sites to values
-        constrained to the site's support.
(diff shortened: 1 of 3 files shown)


numpyro/infer/mcmc.py
```diff
@@ -1250,8 +1252,12 @@ def get_extra_fields(self, group_by_chain=False):
         states = self._states if group_by_chain else self._states_flat
         return {k: v for k, v in states.items() if k != 'z'}
 
-    def print_summary(self, prob=0.9):
-        print_summary(self._states['z'], prob=prob)
+    def print_summary(self, prob=0.9, exclude_deterministic=True):
+        # Exclude deterministic sites by default
+        sites = self._states['z']
+        if exclude_deterministic:
```
fehiepsi: nit: I think you would need to add a check `if isinstance(sites, dict) and exclude_deterministic:` here in case `potential_fn` is used instead of `model`.

neerajprad: Good catch, will fix.

PR #526, opened 2020-01-18
Clean up the gallery

fehiepsi:
This PR addresses some small issues while generating the gallery. After this, the gallery will look like this

Screenshot from 2020-01-18 02-08-29

~Blocked by #340.~


neerajprad:
Could you merge with the latest master?

examples/vae.py
```diff
@@ -133,7 +133,7 @@ def reconstruct_img(epoch, rng_key):
 if __name__ == '__main__':
     assert numpyro.__version__.startswith('0.2.3')
     parser = argparse.ArgumentParser(description="parse args")
-    parser.add_argument('-n', '--num-epochs', default=20, type=int, help='number of training epochs')
+    parser.add_argument('-n', '--num-epochs', default=15, type=int, help='number of training epochs')
```
fehiepsi: with 20 epochs, loss returns nan in my system.

examples/ode.py
```diff
@@ -78,7 +78,7 @@ def main(args):
     mcmc = MCMC(NUTS(model, dense_mass=True),
                 args.num_warmup, args.num_samples, num_chains=args.num_chains,
                 progress_bar=False if "NUMPYRO_SPHINXBUILD" in os.environ else True)
-    mcmc.run(PRNGKey(1), N=data.shape[0], y=np.log(data))
+    mcmc.run(PRNGKey(0), N=data.shape[0], y=np.log(data))
```
fehiepsi: When `progress_bar=False`, with random key = 1, MCMC returns constant values (this indicates pe or its grad is nan/inf for all new proposals). I guess the issue due to  JAX `ode` implementation but has no further evidence.

examples/vae.py
neerajprad: Hmm..this is probably worth investigating separately.

PR #536, opened 2020-02-08
expose inverse mass matrix

### LINKED ISSUE #534: Sequential Sampling Strategy
Hi all,

I have a sequential problem where I need to build an initial model and then update this as new data arrives (eg. every month for monthly data) I know that sequential MCMC is a whole separate field but for a few reasons (mostly its generality) I want to find an efficient way of using NUTS in this context. Instead of the brute force method of re running the whole sampling from scratch on an expanding window of data (too slow), I’m considering:

1. Re using parameter estimates and mass matrix from initial sampler run to start the sampler on the expanded window of data and running fewer sampler iterations. 

What’s the best way to get the (dense) mass matrix and initialisation values from run t in to the sampler for run t+1?

Or:

2. Use posterior estimates from first run of the sampler as priors in the second run, and only the new data. Pymc3 has an ‘Interpolated’ Distribution type which samples from an fitted empirical histogram of data to do this. (Example here https://docs.pymc.io/notebooks/updating_priors.html) I guess this could be implemented in numpyro? Alternatively I guess I can use my prior distributions re parameterised based on samples from previous fit...

Again I want to use mass matrix and parameter estimates from previous fits, so need to set up the sampler with these somehow?

fehiepsi:
Resolves #534.

I think it is good to expose the inverse mass matrix. This allows users to specify the mass matrix which is obtained using other dataset or inference methods (e.g. SVI).

### DIFF
--- a/numpyro/infer/mcmc.py
+++ b/numpyro/infer/mcmc.py
@@ -424,6 +424,9 @@ class HMC(MCMCKernel):
     :param float step_size: Determines the size of a single step taken by the
         verlet integrator while computing the trajectory using Hamiltonian
         dynamics. If not specified, it will be set to 1.
+    :param numpy.ndarray inverse_mass_matrix: Initial value for inverse mass matrix.
+        This may be adapted during warmup if adapt_mass_matrix = True.
+        If no value is specified, then it is initialized to the identity matrix.
     :param bool adapt_step_size: A flag to decide if we want to adapt step_size
         during warm-up phase using Dual Averaging scheme.
     :param bool adapt_mass_matrix: A flag to decide if we want to adapt mass
@@ -443,6 +446,7 @@ class HMC(MCMCKernel):
                  potential_fn=None,
                  kinetic_fn=None,
                  step_size=1.0,
+                 inverse_mass_matrix=None,
                  adapt_step_size=True,
                  adapt_mass_matrix=True,
                  dense_mass=False,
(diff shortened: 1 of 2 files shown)


neerajprad:
I'm not sure how useful it will be to directly expose the mass matrix in the format that it is used internally -  with flattened unconstrained parameters concatenated in a certain order. The user will have to do this externally, so we should consider providing helper utilities that make it less cumbersome or allow for the format to be specified in some other way, e.g. a dict? It depends on the use case though - is this being estimated via SVI or samples drawn from past runs?

fritzo:
Note pyro.infer.autoguide exposes `.median()` and `.quantile()` methods that convert an internal concatenated `(loc,scale_tril)` to a dictionary of user-facing `(loc,scale)` tensors. I guess that would work for diagonal mass matrices, but you'd need a pairwise dictionary for non-diagonal.

fehiepsi:
> with flattened unconstrained parameters concatenated in a certain order

Good point!! Let me think about it a bit more to see if there is a workaround... We can follow @fritzo's suggestion or revise some hidden machinery to "not exposed" support this:
```
mcmc._compile(key, args, kwargs)
init_state = replace(mcmc._get_cached_init_state(key, args, kwarngs), "adapt_state.inverse_mass_matrix", user_defined_inverse_mass_matrix)
mcmc._cache_init_state(init_state, key, args, kwargs)
mcmc.run(key, args, kwargs)
```

fehiepsi:
Mass matrix is sensitive to the order of parameters (even if we force an order, it is still non-trivial to users) and works on an unconstrained space. So it is better to not expose it. #534 can still be resolved by modifying some `init_state` or `warmup` states, which will be exposed in a new PR.

So I would like to close this PR, thanks for the discussions!

MrinankSharma:
Hey all. I'm trying to debug my model using the mass matrix. I find adaptation is very slow because elements of the mass matrix have very different values. However, I'm struggling to associate each element of the mass matrix with a parameter in my model, because I don't know the ordering used. 

Does anybody know how to get the ordering? Thanks.

Issue #540, opened 2020-02-19
Construct a contrib.forecasting module


fehiepsi:
Follow up https://github.com/pyro-ppl/pyro/issues/2311, it would be great to have a `contrib` module in NumPyro for forecasting. I expected some differences

- Use NUTS instead of SVI and target a single time series, instead of a hierarchical time series like in Pyro. This is not a good approach for problems with a bunch of "hierarchical" time series. Even in that case, with the inference results of some "important" time series, we can get some insight into the data and modeling, hence improving Pyro models.
- Play around Holt-Winters models instead of Kalman filter. I think it would be great if we can match the functionality of [Rlgt](https://cran.r-project.org/web/packages/Rlgt/vignettes/GT_models.html). This is convenient because we already have constructed some models in that package here.

### Tasks

- [x] Sketch an interface, especially focusing on forecasting. It is best to mimic (but simplify) the Pyro interface.
- [x] Add hyperparameters to control modeling details (e.g. multiplicative vs additive, seasonal, local trend, global trend,...) and parameter priors.
- [x] Consider to incorporate `covariates` (maybe through a regression component?).
- [x] Vectorize over seasonality window (as @neerajprad did in [this workshop tutorial](https://github.com/pyro-ppl/sandbox/blob/master/2019-08-time-series/part_ii_inference.ipynb))

I would favor the approach "get it done then refactoring later" like how we developed the functional mcmc then refactoring into MCMC. :smile_cat:

fritzo:
@slaweku what model class would you recommend? Are there any areas where you think NumPyro's fast HMC or neural nets could outperform competing univariate forecasting tools?

slaweks17:
Sure. A typical univariate tool will fit a model per series, thus losing opportunities from cross-learning. Also. when there are meaningful regressors, available in past and for future (like holidays, calendar features) or limited future covered by prediction horizon (like user attributes, promotions) NNs and generally ML algorithms should do better.

PR #544, opened 2020-02-25
Add forecast module

fehiepsi:
Address #540.

The `GlobalTrendModel` is different from the current forecasting tutorial at the following points:
+ merge train and forecast into a single model
+ have a regression component
+ support all models from Rlgt library
+ the interface is similar to Pyro

TODO:
- [x] add lgt model
- [x] add sgt model
- [x] add sgt2 model
- [x] support forecasting with those models
- [x] ~vectorize the models~ not necessary because we will mainly use cpu
- [x] finalize the api
- [ ] add tests on toy datas
- [x] add docs
- [x] evaluate and backtesting ?


fehiepsi:
@fritzo The current api can be seen at [this notebook](https://github.com/pyro-ppl/numpyro/blob/36457e8e15cd898ded44a2a8d5fd038d429606a1/numpyro/contrib/forecast/lynx.ipynb). WDYT? Do you have any suggestions for additional features that need to be implemented? (e.g. do we need `backtest` for MCMC?)

fritzo:
@fehiepsi The interface looks good, thanks for staying close to the `pyro.contrib.forecast` interface. I think the `backtest()` function is not necessary in the first PR, but I do think it is important for performing experiments and for model exploration. Maybe you could implement it in a follow-up PR?

numpyro/contrib/forecast/gt.py
```diff
@@ -0,0 +1,285 @@
...
+    "pow_season_alpha": 1,
+    "pow_season_beta": 1,
+    "min_sigma": 1e-10,
+    "min_val": 1e-30,
+    "max_val": 1e38,
+}
+
+
+class GlobalTrendModel:
+    def __init__(self,
```
fritzo: nit: now that we're python 3 only, I've have started using the `*,` separator to force kwargs. Does this work with NumPyro? If so you could define this as
```py
class GlobalTrendModel:
    def __init__(self, *,
                 seasonality=1,
                 ...
                 hyperparameters={}):
```
This allows us to freely reorder parameters in the future, while maintaining backwards compatibility.

numpyro/contrib/forecast/gt.py
```diff
@@ -0,0 +1,285 @@
...
+            powx = numpyro.sample("powx", dist.Uniform())
+            omega = sigma * exp_val ** powx + offset_sigma
+
+        if covariates.shape[0] == N:  # training
+            numpyro.sample("y", dist.StudentT(nu, exp_val, omega), obs=data[1:])
+        else:  # forecasting
+            numpyro.sample("y", dist.StudentT(nu, exp_val, omega))
+
+
+# src: https://github.com/cbergmeir/Rlgt/blob/master/Rlgt/src/stan_files/LGT.stan
```
fritzo: The source file is GPL licensed. Can you find a non-GPL source for a similar model?

numpyro/contrib/forecast/gt.py
fehiepsi: Thanks, I'll do it and add a backtest function too.

numpyro/contrib/forecast/gt.py
fehiepsi: Hmm, this is an important point... I'll learn more about the licence to make changes accordingly.

fehiepsi: Looking like there is no alternative unless we add a GPL license. I'll close this PR and think about other models which I can implement myself...

Issue #548, opened 2020-02-26
[FR] add Gumbel distribution


fehiepsi:
As requested @tbsexton at #545, it would be nice to have Gumbel distribution in NumPyro.

References: [wikipedia](https://en.wikipedia.org/wiki/Gumbel_distribution), [pytorch dist](https://pytorch.org/docs/stable/distributions.html#gumbel), [jax sampler](https://jax.readthedocs.io/en/latest/jax.random.html#jax.random.gumbel).

tbsexton:
Also commenting to add info on `gumbel-softmax`, which kicked off this discussion (continuous approximation of categorical distribution). 

[Paper](https://arxiv.org/pdf/1611.01144.pdf), [(deprecated?) Pytorch version with straight-through forward-mode support](https://github.com/pytorch/pytorch/blob/master/torch/nn/functional.py#L1447)

tbsexton:
Wow, thanks to @daydreamt! That was fast. I'd be happy to close this, or re-name it as "Gumbel-Softmax" if that functionality is also wanted. 

Based on the pytorch link above, it seems like they are making use of some pytorch-specific detachments, etc. to implement the straight-through version (argmax in the fwd/sampling direction, softmax in the backward/gradient direction, if I'm understanding correctly). Not sure how such a thing could be done with the JAX backend.

fehiepsi:
@tbsexton A similar version can be implemented using [jax.lax.stop_gradient](https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.stop_gradient.html#jax.lax.stop_gradient), which plays the role of PyTorch `detach`. However, I am not sure how useful it is to solve your problem. Could you open a new issue which sketches your ideas and your feature request? FYI, we have [RelaxedOneHotCategoricalStraightThrough](http://docs.pyro.ai/en/1.2.1/_modules/pyro/distributions/relaxed_straight_through.html#RelaxedOneHotCategoricalStraightThrough) distribution in Pyro, which can be ported to NumPyro without much difficulty.

I would like to close this issue for now and discuss on the new one. :)

PR #550, opened 2020-02-29
Add Gumbel distribution.

daydreamt:
Hi,

thank you for this wonderful library.

This merge request is my attempt in response to https://github.com/pyro-ppl/numpyro/issues/548 

I mostly used the linked Wikipedia's article notation while implementing e.g. the mean, variance and log_likelihood properties, but I named the parameters loc and scale instead of mu and beta, since these style seems preferred in the rest of the codebase. I also reused some of the code from the Gaussian for broadcasting shapes and so on.

The fact that you can sample from Gumbel(0,1), and then multiply by the scale and add the location to sample Gumbel(location, scale) I based it on the third line [here]( https://stats.stackexchange.com/a/374932), but would appreciate a second pair of eyes verifying. (what Jax's random.gumbel returns is [-ln(-ln(U))](https://github.com/google/jax/blob/master/jax/random.py#L988))
Ideally I would like to add a test for this, but I'm wondering what would be the recommended way.
As far as I understand, there is no existing test already to compare the distributions of generated samples, right?


numpyro/distributions/continuous.py
```diff
@@ -326,6 +329,39 @@ def variance(self):
...
+                                     validate_args=validate_args)
+
+    def sample(self, key, sample_shape=()):
+        standard_gumbel_sample = random.gumbel(key, shape=sample_shape + self.batch_shape + self.event_shape)
+        return self.loc + self.scale * standard_gumbel_sample
+
+    @validate_sample
+    def log_prob(self, value):
+        z = (value - self.loc) / self.scale
+        return (np.exp(-z) - z) - np.log(self.scale)
```
fehiepsi: I think you miss `-` sign before exp term.

fehiepsi:
Awesome!!!

> As far as I understand, there is no existing test already to compare the distributions of generated samples, right?

For upstream (JAX) samplers, we don't need to test. If we implement samplers by ourselves, we would need to add some more statistical tests. :) Could you also add a doc entry [at this file](https://github.com/pyro-ppl/numpyro/blob/master/docs/source/distributions.rst#continuous-distributions)

numpyro/distributions/continuous.py
daydreamt: Thank you, I did indeed.

daydreamt:
> > As far as I understand, there is no existing test already to compare the distributions of generated samples, right?
> 
> For upstream (JAX) samplers, we don't need to test. If we implement samplers by ourselves, we would need to add some more statistical tests. :) Could you also add a doc entry [at this file](https://github.com/pyro-ppl/numpyro/blob/master/docs/source/distributions.rst#continuous-distributions)

Hi @fehiepsi,
Since we do not directly get Gumbel(loc, scale), but only Gumbel(0,1) samples from JAX,
I now also added a test for equality of distribution between our generated Gumbel(loc, scale) samples and those from SciPy, and the test is currently only active for the Gumbel distribution.

I'm using the Kolmogorov-Smirnov test instead of the Anderson-Darling because 1) the scipy implementation seems to support less distributions (and we [can't just pass it a sp_dist object](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.anderson.html#scipy.stats.anderson)) and 2) for the many thousands of samples we are generating, the power of the Kolmogorov-Smirnov test seems to be fine (this comes from skimming [this paper](https://www.researchgate.net/profile/Bee_Yap/publication/267205556_Power_Comparisons_of_Shapiro-Wilk_Kolmogorov-Smirnov_Lilliefors_and_Anderson-Darling_Tests/links/5477245b0cf29afed61446e1/Power-Comparisons-of-Shapiro-Wilk-Kolmogorov-Smirnov-Lilliefors-and-Anderson-Darling-Tests.pdf)). I'm definitely open to better alternatives. 

I set the number of samples to generate before testing to 100_000, because with only 10_000 samples while testing all existing continuous distributions, the test rejected the equality of distribution of the samples generated by one of the Chi Squared distributions. Despite the SciPy documentation, it also seems we do not have to set a numpy seed value for the test ( I tried setting many different ones, but the result didn't change)

Issue #543, opened 2020-02-24
Very slow compile on simple model


jhmenke:
Below is an example code of using MCMC to infer some variables *vm* (size: 3) and *va* (size: 2). However, on my machine the runtime of the code is about 3 minutes, mostly due to an abnormally slow compilation process. The results are exactly what i would expect though.

Unfortunately, debugging that is not really my expertise, therefore i have created a minimal example and also attached the compile logs as per the suggestion. I think there is a good chance this is some beginners mistake and not actually a bug in your code, but i'm not quite sure to be honest.

I'm running an Ubuntu derivate 64 bit with Python 3.7 and updated jax/jaxlib/numpyro releases. Thanks to someone who can take a quick peek at this!

> ********************************
> Very slow compile?  If you want to file a bug, run with envvar XLA_FLAGS=--xla_dump_to=/tmp/foo and attach the results.
> ********************************

```
import argparse

import jax
import jax.numpy as np
import jax.random as random
import numpy as onp
import numpyro
import numpyro.distributions as dist
import pandas as pd
from numpyro.distributions.constraints import interval, greater_than
from numpyro.infer import MCMC, NUTS

jax.config.update('jax_platform_name', 'cpu')

# create the data to use
Z = onp.asarray([[onp.nan, 1e-2 + 3e-2j, 2e-2 + 5e-2j],
                 [1e-2 + 3e-2j, onp.nan, 3e-2 + 8e-2j],
                 [2e-2 + 5e-2j, 3e-2 + 8e-2j, onp.nan]], dtype=onp.complex)
Y = -onp.reciprocal(Z)
Y[onp.where(onp.isnan(Y))] = 0.
g, b = np.array(-Y.real), np.array(-Y.imag)
Y[0, 0] = -Y[0, :].sum()
Y[1, 1] = -Y[1, :].sum()
Y[2, 2] = -Y[2, :].sum()
G, B = np.array(Y.real), np.array(Y.imag)
measurements_df = pd.DataFrame(columns=["type", "value", "std"],
                               index=pd.RangeIndex(8),
                               data=[["pl_1_2", 0.888, 0.008],
                                     ["pl_1_3", 1.173, 0.008],
                                     ["p_2", -0.501, 0.01],
                                     ["ql_1_2", 0.568, 0.008],
                                     ["ql_1_3", 0.663, 0.008],
                                     ["q_2", -0.286, 0.01],
                                     ["v_1", 1.006, 0.004],
[...]

PR #553, opened 2020-03-07
Remove find_reasonable_step_size

fehiepsi:
Addresses #543. This PR drops the usage of `find_reasonable_step_size` in `hmc`.

### Reason

In HMC/NUTS, we only compile potential_fn and its grad 1 time: at leap_frog step. However, in find_reasonable_step_size, we compile potential_fn 2 times (to get initial potential energy and in the while-loop to find reasonable step_size) and its grad 1 time (in the while-loop to find reasonable step_size). Those together require JAX to compile a computational graph which is 2.5x larger than if we drop `find_reasonable_step_size`. It is known that JAX compiling time is superlinear w.r.t. its computational graph, so if the model is slow to compile, our MCMC will take a very long time for its initial step. Some evidence:
+ As @jhmenke observed in #543, removing this functionality reduces compiling time from 3m to 11s.
+ In our early development, we observed the slowness and introduced `heuristic_step_size` flag at hmc_init (see https://github.com/pyro-ppl/numpyro/pull/127). But we removed it at 0.2 release to clean up the interface, and the compiling time is acceptable for models we tested.
+ `hmm` example: 39s -> 28s elapsed time
+ `ode` example: 20s compiling -> 10s compiling

With this change, the compiling time of `sample_kernel` will be aligned with the compiling time of `value_and_grad(potential_fn)`.

### Effect

I used the following code to test the effect of removing this feature. Surprisingly, it seems that dual_averaging can do a pretty good job to remedy it
```python
from numpyro.infer.hmc_util import dual_averaging
import jax.numpy as np

ss_init, ss_update = dual_averaging()

ss_state = ss_init()
for i in range(10):
    accept_prob = 0.4  # 0.4, 0.6, 0.9
    ss_state = ss_update(0.8 - accept_prob, ss_state)
    print("current ss", np.exp(ss_state[0]), "    \taverage ss", np.exp(ss_state[1]))
```
With `accept_prob = 0.4`, we get
```
current ss 0.48322508     	average ss 0.48322508
current ss 0.15173525     	average ss 0.24267578
current ss 0.04085883     	average ss 0.111069076
current ss 0.010343172     	average ss 0.047984514
current ss 0.0025725756     	average ss 0.020001529
current ss 0.0006435767     	average ss 0.008161279
current ss 0.00016403742     	average ss 0.0032920958
current ss 4.290037e-05     	average ss 0.0013218971
current ss 1.1554674e-05     	average ss 0.0005309463
current ss 3.2104135e-06     	average ss 0.0002140629
```
we can see that if things are bad, step_size will drop pretty fast. On the other hand `accept_prob = 1`, we get
```
current ss 1.438551     	average ss 1.438551
current ss 2.5671825     	average ss 2.0299568
current ss 4.947172     	average ss 3.0005677
current ss 9.832708     	average ss 4.565091
current ss 19.71587     	average ss 7.0707974
current ss 39.418476     	average ss 11.069319
current ss 78.07797     	average ss 17.42865
current ss 152.67555     	average ss 27.504337
current ss 294.18542     	average ss 43.3985
current ss 558.1096     	average ss 68.34855
```
which shows that step_size (and its average) will be increased quickly if things look good. The length of the last window in warmup_adapter is 50, so `step_size` has the whole room to adjust in this interval. So I believe that inference results won't change if we remove this functionality.

As an example, running `hmm`, I don't observe differences in the inference result.

### DIFF
--- a/numpyro/infer/mcmc.py
+++ b/numpyro/infer/mcmc.py
@@ -27,7 +27,6 @@ from numpyro.infer.hmc_util import (
     IntegratorState,
     build_tree,
     euclidean_kinetic_energy,
-    find_reasonable_step_size,
     velocity_verlet,
     warmup_adapter
 )
@@ -243,15 +242,11 @@ def hmc(potential_fn=None, potential_fn_gen=None, kinetic_fn=None, algo='NUTS'):
                 kwargs = {} if model_kwargs is None else model_kwargs
                 pe_fn = potential_fn_gen(*model_args, **kwargs)
 
-        find_reasonable_ss = partial(find_reasonable_step_size,
-                                     pe_fn, kinetic_fn, momentum_generator)
-
         wa_init, wa_update = warmup_adapter(num_warmup,
                                             adapt_step_size=adapt_step_size,
                                             adapt_mass_matrix=adapt_mass_matrix,
                                             dense_mass=dense_mass,
-                                            target_accept_prob=target_accept_prob,
-                                            find_reasonable_step_size=find_reasonable_ss)
+                                            target_accept_prob=target_accept_prob)
 
         rng_key_hmc, rng_key_wa, rng_key_momentum = random.split(rng_key, 3)
         wa_state = wa_init(z, rng_key_wa, step_size,
--- a/numpyro/infer/mcmc.py
+++ b/numpyro/infer/mcmc.py
@@ -27,6 +27,7 @@ from numpyro.infer.hmc_util import (
     IntegratorState,
     build_tree,
     euclidean_kinetic_energy,
+    find_reasonable_step_size,
     velocity_verlet,
     warmup_adapter
 )
@@ -191,6 +192,7 @@ def hmc(potential_fn=None, potential_fn_gen=None, kinetic_fn=None, algo='NUTS'):
                     target_accept_prob=0.8,
                     trajectory_length=2*math.pi,
                     max_tree_depth=10,
+                    find_heuristic_step_size=False,
                     model_args=(),
                     model_kwargs=None,
                     rng_key=PRNGKey(0)):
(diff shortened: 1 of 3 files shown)


neerajprad:
Always good to see code being simplified! My concern is that there might be cases where we are wasting adaptation steps in getting to a reasonable value of step size, and we may now start giving bad results on models that were previously working (though given your observations, that seems unlikely). Longer term, I think this is probably the right solution anyways since running longer adaptation will be faster than paying for repeated compilation time. In the meantime, a safer alternative would be to gate this behind a `find_heuristic_step_size=False` argument, which by default will not use the `find_reasonable_step_size` heuristic. If we don't notice any issues with the new default, we can later remove it completely  and simplify the interface. WDYT?

fehiepsi:
> n the meantime, a safer alternative would be to gate 

yup, this is safer. Your concern is quite valid! I hope that in that case, we can replace it by other heuristic functions, which do not depend on potential_fn (maybe just using the norm of the last inverse_mm and the current inverse_mm) so that we won't face compiling issue. But it is quite likely that we won't get trouble with this change. :)

fehiepsi:
@neerajprad sorry for the delay. I just update the pattern as you suggested.

numpyro/infer/mcmc.py
```diff
@@ -437,6 +444,8 @@ class HMC(MCMCKernel):
         value is :math:`2\\pi`.
     :param callable init_strategy: a per-site initialization function.
         See :ref:`init_strategy` section for available functions.
+    :param bool find_heuristic_step_size: whether to use an extra heuristic
```
neerajprad, line 447: Maybe we can add to the description - `"Uses a heuristic function to adjust the step size at the end of each adaptation window."`?

Also, just to confirm - this will still use this to initialize the step size but omit doing so at the end of each adaptation window?

fehiepsi: @neerajprad We also skip the initialization of step size. Actually, this function adjusts step size at the "begin" of each adaptation window.

neerajprad: I see, I forgot that this is a more fundamental change. I think the only time this should matter is when we are using very few warmup steps. It seems like we do this at the start and then at the end of each adaptation [window](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/infer/hmc_util.py#L399)?

fehiepsi: > the only time this should matter is when we are using very few warmup steps

Exactly!

> It seems like we do this at the start and then at the end of each adaptation window?

Actually, we do this only at the start of each adaptation window (including the first window).

Issue #568, opened 2020-04-09
mask handler does not behave as expected


dsheldon:
I am trying to use the mask handler to mask invalid observed values (nans). Per [this discussion](https://forum.pyro.ai/t/behavior-of-mask-handler-with-invalid-observation-possible-bug/1719), I replace invalid observations with values that are valid for the particular distribution to avoid nans in the gradient. But now it seems that the default values actually affect the posterior distribution.

Please see this example, where x ~ N(0, 1) and y_i ~ N(x, 0.1) for i = 1 to n:

~~~python
import jax
import jax.numpy as np
from jax.random import PRNGKey

import numpyro
import numpyro.distributions as dist
from numpyro.infer import MCMC, NUTS

def model(n, y=None):
    
    x = numpyro.sample("x", dist.Normal(0, 1))
    
    mask = True 
    if y is not None:
        # Mask nan values, then substitute a default value to
        # avoid nans in gradients
        mask = ~np.isnan(y)
        y = np.where(mask, y, -100) 
        
    with numpyro.handlers.mask(mask):
        y = numpyro.sample("y", dist.Normal(x, 0.1), sample_shape=(n,), obs=y)
        
    return x, y


n = 100
with numpyro.handlers.seed(rng_seed=1):
    x, _ = model(n)
    print("x is", x)
    
y = np.repeat(np.nan, n)

kernel = NUTS(model)
mcmc = MCMC(kernel, num_warmup=1000, num_samples=1000, num_chains=1)
mcmc.run(PRNGKey(1), n, y=y)

mcmc.print_summary()
~~~

Here I provide an observed vector `y` that is all `nan`, so I expect the posterior over x to be equal to the prior. However, the posterior I get is highly concentrated about the default value (-100) used to replace nans:

~~~text
                mean       std    median      5.0%     95.0%     n_eff     r_hat
         x    -99.99      0.01    -99.99   -100.01    -99.97    410.46      1.01
~~~

I'm filing this as an issue because I believe that masked observations are not supposed to affect the log-density or its gradients, but let me know if I've misinterpreted something.

fehiepsi:
Your argument sounds very reasonable to me. Let me look at this in more details.

fehiepsi:
@dsheldon It seems that you miss the keyword `mask_array`:
```
with numpyro.handlers.mask(mask_array=mask):
```

dsheldon:
Oh, wow. I guess I misunderstood how to call this.

[...]

PR #569, opened 2020-04-09
fix bug for models with mask=False

fehiepsi:
Addresses #568.

### DIFF
--- a/numpyro/infer/util.py
+++ b/numpyro/infer/util.py
@@ -59,7 +59,7 @@ def log_density(model, model_args, model_kwargs, params, skip_dist_transforms=Fa
     else:
         model = substitute(model, param_map=params)
     model_trace = trace(model).get_trace(*model_args, **model_kwargs)
-    log_joint = 0.
+    log_joint = jax.device_put(0.)
     for site in model_trace.values():
         if site['type'] == 'sample':
             value = site['value']
@@ -68,7 +68,7 @@ def log_density(model, model_args, model_kwargs, params, skip_dist_transforms=Fa
             scale = site['scale']
             # Early exit when all elements are masked
             if not_jax_tracer(mask) and mask is not None and not np.any(mask):
-                return jax.device_put(0.), model_trace
+                continue
             if intermediates:
                 if skip_dist_transforms:
                     log_prob = site['fn'].base_dist.log_prob(intermediates[0][0])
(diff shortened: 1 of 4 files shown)


numpyro/infer/util.py
```diff
@@ -59,7 +59,7 @@ def log_density(model, model_args, model_kwargs, params, skip_dist_transforms=Fa
     else:
         model = substitute(model, param_map=params)
     model_trace = trace(model).get_trace(*model_args, **model_kwargs)
-    log_joint = 0.
+    log_joint = jax.device_put(0.)
```
neerajprad, line 62: @fehiepsi - maybe we should also add a check to `mask` so that it raises a ValueError if fn is not a callable to ensure that users don't trip up by not specifying this as a keyword:

```python
if fn is not None and not callable(fn):
    raise ValueError("Expected `fn` to be callable; instead found type(fn) = {}.".format(type(fn)))
```

numpyro/infer/util.py
```diff
@@ -68,7 +68,7 @@ def log_density(model, model_args, model_kwargs, params, skip_dist_transforms=Fa
             scale = site['scale']
             # Early exit when all elements are masked
             if not_jax_tracer(mask) and mask is not None and not np.any(mask):
-                return jax.device_put(0.), model_trace
+                continue
```
neerajprad, line 71: Thanks for fixing this. So if everything is masked this will return `0` as the log density?

numpyro/infer/util.py
fehiepsi: good point! I'll add it.

numpyro/infer/util.py
fehiepsi: Yes. We start looping over the sites with `log_joint = 0.`

numpyro/primitives.py
```diff
@@ -42,6 +42,9 @@ def apply_stack(msg):
 
 class Messenger(object):
     def __init__(self, fn=None):
+        if fn is not None and not callable(fn):
+            raise ValueError("Expected `fn` to be callable; instead found type(fn) = {}."
```
neerajprad, line 46: nit: can we change `to be callable` --> `to be a Python callable object` for clarity?

fehiepsi: thanks

Issue #566, opened 2020-04-07
Predictive distribution fails on model with lax.scan


neerajprad:
Refer to the [forum discussion](https://forum.pyro.ai/t/state-space-model-is-lax-scan-compatible-with-numpyro-sample/1758/8)

The following snippet:

```python
def target(T=10, q=1., r=1., phi=0.5, beta=0.5):

    def transition(state, xs):
        i, key = xs
        key1, key2 = jax.random.split(key)
        x0, mu0 = state
        x1 = numpyro.sample(f'x_{i}', dist.Normal(phi * x0, q), rng_key=key1)
        mu1 = beta * mu0 + x1
        y1 = numpyro.sample(f'y_{i}', dist.Normal(mu1, r), rng_key=key2)
        return (x1, mu1), (x1, y1)

    mu0 = x0 = numpyro.sample('x_0', dist.Normal(0, q))
    y0 = numpyro.sample('y_0', dist.Normal(mu0, r))

    key = numpyro.sample('key', dist.PRNGIdentity())
    _, xy = jax.lax.scan(transition, (x0, mu0), (np.arange(1, T), jax.random.split(key, T-1)))
    x, y = xy
    
    return np.append(x0, x), np.append(y0, y)


prior = Predictive(target, posterior_samples = {}, num_samples = 10)
prior_samples = prior(PRNGKey(2), T=10)
```

fails with:

```
UnexpectedTracerError: Encountered an unexpected tracer. Perhaps this tracer escaped through global state from a previously traced function.
The functions being transformed should not save traced values to global state.
Details: Can't lift level Traced<ShapedArray(float32[]):JaxprTrace(level=1/0)> to JaxprTrace(level=0/0).
```

fehiepsi:
@neerajprad I think we can't trace this function
```
numpyro.handlers.trace(numpyro.handlers.seed(target, 0)).get_trace()
```

neerajprad:
Hmm..yes, that makes sense. We can't really collect `x_i`, `y_i` samples unfortunately, so this doesn't work. I would like to keep this issue open to explore better workarounds for such use cases where we may need to use a jax control flow primitive inside a model. I think the only workaround is to make sure that the body of the control flow doesn't contain any pyro primitives, but I wonder if there's a better solution.

fehiepsi:
[...]

PR #572, opened 2020-04-20
Port funsor-based enumeration machinery for profiling

eb8680:
This PR contains NumPyro-compatible versions of some of the Funsor-based enumeration machinery from https://github.com/pyro-ppl/pyro/pull/2307.  It is primarily intended to support testing and performance experiments for integration with HMC, and will be replaced by a more carefully designed version when we are confident in the performance numbers.

Tasks
- [x] Add a couple of example models, maybe Pyro HMM examples
- [x] Get enumeration working for `model_0` and `model_1`
- [x] Get HMC working end-to-end for HMM `model_1`
- [x] Get HMC/NUTS working end-to-end for all 5 HMMs
- ~[ ] Troubleshoot high compile times~
- ~[ ] Understand/fix high memory use~
- ~[ ] Profile compared to Pyro versions~
- [x] Add some MCMC tests
- [x] fix incorrect inference results

Triaged
- ~~Port some tests from https://github.com/pyro-ppl/pyro/pull/2307 - this is a bit difficult at the moment because the tests in Pyro rely heavily on Pyro's old enumeration machinery~~


numpyro/infer/enum_messenger.py
```diff
@@ -0,0 +1,557 @@
...
+
+class trace(OrigTraceMessenger):
+    """
+    This version of TraceMessenger records information necessary to do packing after execution.
+    Each sample site is annotated with a "dim_to_name" dictionary,
+    which can be passed directly to funsor.to_funsor.
+    """
+    def postprocess_message(self, msg):
+        if msg["type"] == "sample":
+            msg["infer"]["dim_to_name"] = NamedMessenger._get_dim_to_name(msg["fn"].batch_shape)
```
eb8680, line 513: This metadata should be sufficient to pack the `log_prob` tensors for sum-product contraction with `funsor.sum_product` in a modified `potential_energy` implementation.

numpyro/infer/util.py
```diff
@@ -164,6 +168,50 @@ def potential_energy(model, inv_transforms, model_args, model_kwargs, params):
     return - log_joint
 
 
+def enum_potential_energy(model, inv_transforms, model_args, model_kwargs, params):
```
eb8680, line 171: @fehiepsi @neerajprad here's the modified `potential_energy` we discussed

numpyro/infer/enum_messenger.py
```diff
@@ -0,0 +1,577 @@
...
+                msg["infer"].get("enumerate") != "parallel":
+            return
+
+        if msg["infer"].get("num_samples", None) is not None:
+            raise NotImplementedError("TODO implement multiple sampling")
+
+        if msg["infer"].get("expand", False):
+            raise NotImplementedError("expand=True not implemented")
+
+        raw_value = msg["fn"].enumerate_support(expand=False).squeeze()
```
eb8680, line 494: @fehiepsi does this functionality exist yet in NumPyro?  If not, can you suggest an alternative implementation for this line?

neerajprad: Hmm..we didn't bother to add this since we didn't have enumeration, but it should be straightforward to add this to our discrete distributions. WDYT, @fehiepsi?

eb8680: I can just add a workaround here for now

numpyro/infer/util.py
neerajprad: Nice..I think this should plug into the existing hmc implementation without issues. Let us know if you face any issues.

numpyro/infer/enum_messenger.py
fehiepsi: I'll add this soon.

eb8680:
Enumeration of values now seems to work and produce correct shapes in a couple of HMMs ported from the Pyro HMM examples.  I haven't yet run HMC successfully, but hopefully I'll be able to do that tonight.

I've been having some issues with the mismatch between Pyro's shape semantics and NumPyro's shape semantics.  For example, in Pyro, every batch dimension must be either a plate dimension or an enumeration/sample dimension, whereas in NumPyro some distributions have batch dimensions that are not accounted for by plates that we would `to_event` away in Pyro.  Also, plates don't expand distribution shapes automatically.  These issues aren't blockers but we should keep them in mind and consider sticking more closely to Pyro as we continue to develop enumeration in NumPyro.

numpyro/infer/util.py
eb8680: I've been trying to run HMC with `enum_potential_energy` but I'm currently a bit stuck trying to work around the assumption baked into multiple parts of the HMC code that every latent variable is differentiable and constrained.  Is there an easy way to hack around this for testing?

neerajprad: I think we'll need to change both `get_potential_fn` and `find_valid_initial_params` in `numpyro.infer.util` to exclude sample sites containing discrete valued distributions.

neerajprad:
> For example, in Pyro, every batch dimension must be either a plate dimension or an enumeration/sample dimension, whereas in NumPyro some distributions have batch dimensions that are not accounted for by plates that we would to_event away in Pyro. Also, plates don't expand distribution shapes automatically. These issues aren't blockers but we should keep them in mind and consider sticking more closely to Pyro as we continue to develop enumeration in NumPyro.

Shape checks in NumPyro are quite lax, and `plate` changes the sample shape without modifying the underlying batch shape (but we can add something like `ExpandedDistribution` which effectively does the same thing while reflecting the correct `batch_shape`). We can also annotate event dims with `.to_event` but we have pretty minimal shape checks at the moment. I think we had to enforce stricter checks in Pyro when we added enumeration, and we may need to do it now in NumPyro too. Please feel free to create any such issues that might be blocking this effort.

eb8680:
Update: NUTS + enumeration now seems to work for 3/5 models, but compile times are extremely long and memory use is through the roof.

neerajprad:
> Update: NUTS + enumeration now seems to work for 3/5 models, but compile times are extremely long and memory use is through the roof.

Does #580 help with the slow compilation times? Taking a quick glance at the code, nothing jumped at me, but if you have any long loops, they will get unrolled by jit, so we may need to use jax's control flow primitives like `lax.fori_loop`.

eb8680:
> Does #580 help with the slow compilation times?

No, that's just to make writing the models easier and avoid indexing bugs.  There are a couple of different Python for-loops in the models (one within the model looping through timesteps, and one in the variable elimination computation in Funsor [here](https://github.com/pyro-ppl/funsor/blob/master/funsor/optimizer.py#L101)).

A possible next step is to write highly simplified versions of these two loops specific to a single model and try to rule out any pathologies in the details of the general implementation.  I'd also like to compare runtimes and compile times to Pyro's HMC/NUTS.

neerajprad:
> There are a couple of different Python for-loops in the models (one within the model looping through timesteps, and one in the variable elimination computation in Funsor here).

I think we will need to rewrite one or both of the loops using JAX's control flow to get reasonable performance. I'll need to understand the Funsor code better to judge how straightforward that would be. I think we tried to run Pyro's HMC/NUTS on these models earlier, but it was unbearably slow.

> A possible next step is to write highly simplified versions of these two loops specific to a single model and try to rule out any pathologies in the details of the general implementation.

That's a good idea, and will give us an upper limit on how fast NUTS sampling will be on these models.

eb8680:
Another update: all 5 models now work, but compilation times and memory usage remain insurmountable barriers.  I've written some simplified versions of the relevant variable elimination computations in `test_funsor_tve_smoke.py` and, while they are still rather slow to compile (taking up to 2 minutes on my laptop in some cases), there is a large unexplained compiler performance increase over the actual HMM examples.  

It's hard to say whether there's a big mistake somewhere in the combined example/inference code that's behind the slowdown, or whether it's just the combination of a super-linear slowdown somewhere in XLA with the much larger traces generated by HMC/NUTS.

fehiepsi:
@eb8680  I'll also troubleshoot those issues. I think that the `for` loops are the main problems. In the `examples/hmm.py` example, using `lax.scan` took me 20s to compile. If I replace `lax.scan` by the naive for loop, compiling took "forever". It is interesting to me to see how to rewrite those `markov` handlers using jax control flows (it might be doable for some specific patterns).

eb8680:
>  It is interesting to me to see how to rewrite those markov handlers using jax control flows

I haven't thought about it too carefully but I believe it should be possible to make effect handlers aware of the JAX control flow ops and generically support the use of `fori_loop` and `scan` with enumeration that way.  We may also need to implement some basic control-flow terms in Funsor to avoid having to write JAX-specific versions of core Funsor transformations like the optimizer.

neerajprad:
> I haven't thought about it too carefully but I believe it should be possible to make effect handlers aware of the JAX control flow ops and generically support the use of fori_loop and scan with enumeration that way. 

This has been a recurring issue (e.g. https://github.com/pyro-ppl/numpyro/issues/566), and our workaround has involved reparametrizing the models (see [forum post](https://forum.pyro.ai/t/hmm-like-model-with-sequences-of-different-lengths/1507/12)) or passing a rng seed explicitly and gathering the end results. This isn't satisfactory though because we can't use effect handlers to record the values at these sample sites or intervene in other ways. You are right that we really need JAX control flow to work with effect handlers, which will probably require expressing these as some kind of program transformations. Will be curious to hear your thoughts.

fehiepsi:
Just a side note: With #553, compiling time is significantly reduced for sequential models because we remove model-compiling in the warmup adapter. However, there are still two places left which affect compiling time:
+ We use [dynamic_args=True](https://github.com/pyro-ppl/numpyro/blob/25d055ad4b5a5e85a02159af87e8db1451243217/numpyro/infer/mcmc.py#L487) in MCMC, which [runs the model two times](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/infer/util.py#L431): 1 to get inv_transforms, 1 to compute potential_energy
+ We compile [build_basetree](https://github.com/pyro-ppl/numpyro/blob/25d055ad4b5a5e85a02159af87e8db1451243217/numpyro/infer/hmc_util.py#L646) (which includes potential_energy) two times: 1 to get the initial leaf and the other in the while loop. Let's see there is a workaround for this issue.

I believe that with those fixes, compiling time will be reduced from 2m to less than 30s, or even much better because the model will be traveled exactly 1 time. I'll try to address those tomorrow. They should be high priority issues.

eb8680:
Another update: compilation speed and memory use have improved with the latest changes in `master`, and the post-compilation performance gains over Pyro are substantial (5-100x) for the HMM examples in this PR run with sequences of length 50 (`--truncate 50`).  Unfortunately compilation is still impractical for truncated data (taking 5-10 minutes on my laptop) and infeasible (I stopped after 20 minutes) for non-truncated versions.  Since most of the easy fixes have landed in `master` already, I suspect the problem is now down to the for-loops in the model and the inference algorithms, which will require extra engineering effort to fix (although I continue to believe this is possible and does not require fundamentally redesigning numpyro).

In another bit of bad news, https://github.com/pyro-ppl/pyro/pull/2307 may not land in Pyro for some time, if ever, because Funsor is not and has never been compatible with Python 3.5 and we are not yet willing to drop this requirement until PyTorch does in PyTorch 1.6.  As a result, this larger Funsor-based refactoring initiative is in limbo and it's unclear whether it's worth finishing, so I'm going to declare victory and close this PR until a final decision is made about how to proceed.

fehiepsi:
> I continue to believe this is possible and does not require fundamentally redesigning numpyro

Thanks, @eb8680 ! I agree. If I can't come up with a solution for for-loop in a week, I'll buy you a beer when we meet.

neerajprad:
> I suspect the problem is now down to the for-loops in the model and the inference algorithms, which will require extra engineering effort to fix (although I continue to believe this is possible and does not require fundamentally redesigning numpyro).

I can't wait to see us make headway on this, given how important this has turned out to be. I'll likely not be able to go through the issues involved for another week, but your optimism is reassuring! :) 

> In another bit of bad news, pyro-ppl/pyro#2307 may not land in Pyro for some time, if ever, because Funsor is not and has never been compatible with Python 3.5 

I think if the compilation issues are resolved, we should be in a good position to reinstate this. I don't think it will be that much longer until pytorch 1.6 is released, and I'm hopeful that we can have all the other issues in numpyro resolved before then (not that pytorch's release blocks this effort at least). Until then, this is victory on the first milestone!

fehiepsi:
@eb8680 Can I update this branch to make it compatible with master branch and add some small examples such as change point detection? I'll try to keep the change minimal and mostly only touch infer/util.py file. I also want to take this chance to see if we can rewrite markov in scan language (which seems to be possible because the latent shapes are periodically repeated - except for one or two first latent states).

eb8680:
@fehiepsi sure, feel free to push directly and make any necessary changes, I'm excited to see if you can get things working with `scan`

numpyro/contrib/funsor/infer_util.py
```diff
@@ -0,0 +1,64 @@
...
+import numpyro
+from numpyro.contrib.funsor.enum_messenger import infer_config, plate as enum_plate, trace as packed_trace
+from numpyro.distributions.util import is_identically_one
+from numpyro.handlers import substitute
+
+funsor.set_backend("jax")
+
+
+@contextmanager
+def plate_to_enum_plate():
```
fehiepsi, line 17: Inside this context, `numpyro.plate` is replaced by `funsor.plate`.

numpyro/infer/util.py
```diff
@@ -360,15 +386,23 @@ def initialize_model(rng_key, model,
...
+    inv_transforms, replay_model, has_enumerate_support, model_trace = _get_model_transforms(
         substituted_model, model_args, model_kwargs)
     constrained_values = {k: v['value'] for k, v in model_trace.items()
                           if v['type'] == 'sample' and not v['is_observed']
                           and not v['fn'].is_discrete}
 
+    if has_enumerate_support:
+        from numpyro.contrib.funsor import config_enumerate, enum
+
+        if not isinstance(model, enum):
```
fehiepsi, line 398: If the model is already wrapped by `enum`, we might skip this step... (this check might be unnecessary).

eb8680:
Great job getting this over the finish line, @fehiepsi !  I'll take a second look at the code tomorrow but I'm especially pleased to see the improvements to broadcasting syntax in the HMM examples after all the other PRs you've finished.

> Please let me know what you think

The funsor integration after your cleanup work here and in other PRs looks really slick!  I think what's in this PR is a suitable MVP and could be released as a `contrib` module.  The main improvements I'd like to see before release are better documentation for new user-facing handlers and primitives, especially `enum`, `infer_config`/`config_enumerate`, and `to_funsor`/`to_data`, and a couple of simpler, more flattering examples.  You might also want to consider temporarily removing `markov` entirely or at least removing its `__iter__` method given the performance problems with sequential models like HMMs.

Going forward, after this PR is merged and a new NumPyro version is released my biggest concern about this functionality other than performance is figuring out how to share some of the code across Pyro and NumPyro - the dimension tracking logic in `DimStack` and `markov` is still quite intricate, has some rough edges I'd like to address, and should be tested more systematically (more on that below), so I'd like to avoid us maintaining more than one copy if possible.

My current thinking is that Funsor is really the right place for `DimStack` and the `named`/`localnamed`/`globalnamed` messengers to live since they're all closely tied to Funsor already, while everything higher level like `enum` and `plate` should stay in Pyro/NumPyro.  However, it's not clear how to accomplish this without moving the base `Messenger` definition/API to Funsor as well (although that may actually be a good idea).  This issue is why I originally indicated that this PR was a draft and would eventually need to be reorganized/redesigned.

> do I need to add more tests?

The enumeration logic in this PR is well-tested in the Pyro version, but it's tested by comparison to Pyro's current enumeration logic rather than directly.  There are a few tests in the Pyro version that are more direct, which should be fairly easy to port and should be enough for a `contrib` module release.  

You're welcome to write more unit tests beyond that; adding more direct low-level unit tests is also a big todo for me in the Pyro version.  Better documentation of the low-level components would also be nice, and is something I started on in the Pyro version before putting the project temporarily on hold.

However, since the code here will be of somewhat limited usefulness until it works with JAX control flow, you might want to focus on a fix for that instead before worrying too much about the current implementation.

> Before this is landed, I think we should move the data file to cloudfront

That dataset is a slice of the polyphonic music data used in the Pyro HMM and DMM examples.  Unfortunately the preprocessing logic there depends on PyTorch.  

Given the outstanding performance issues with sequential models like the HMMs here, I would suggest removing them from this PR entirely (or converting them to tests) and adding some non-sequential examples like a Gaussian mixture model instead, even if you'd prefer not to remove `markov`.  If you convert them to tests you can just generate some fake data in the test code.

fehiepsi:
Thanks for your comments, @eb8680! I'll work through the docs and add more unit tests soon.

> moving the base Messenger definition/API to Funsor

Maybe we just make a duplicated version in funsor and add some wrapper here. We don't have any check for `isinstance(..., Messenger)` so I think this is not a problem.

eb8680:
> Maybe we just make a duplicated version in funsor and add some wrapper here

Another place it could live is in `pyroapi`, but I don't want to put any global state like the handler stack there.  I'll have to think more about possible solutions to this.

eb8680:
@fehiepsi what else is blocking this? Do I just need to do a final review?

fehiepsi:
@eb8680 I marked it blocked because funsor 0.1.3 has not been released yet. I think we can go ahead and use master branch of funsor here in the meantime. Could you help me do the final review?

@neerajprad There are some interface changes in this PR:
+ `infer` keyword is added to `sample`
+ `enum` is added to various utilities in `infer/util.py`
+ I also added `funsor` as a dependency.
Please let me know if you have any suggestions. I'm not sure if we should make those and other changes or if there are any simplifications for them.

I intended to implement mixture models such as [Dawid-Skene models](https://www.aclweb.org/anthology/Q18-1040.pdf) but have faced [label switching](https://mc-stan.org/docs/2_23/stan-users-guide/label-switching-problematic-section.html) issues. I am not sure how to deal with it so if you guys have any experience before, please let me know. I'll make a separate PR for that example.

eb8680:
Ok, I just wanted to make sure you weren't waiting for something from me - no rush.

> Could you help me do the final review?

Sure, will do

fritzo:
Feel free to release funsor 0.1.3

fehiepsi:
@fritzo Sure, I'll ping you when we are ready for 0.3 release. :)

test/contrib/test_funsor.py
```diff
@@ -0,0 +1,92 @@
...
+from jax import random
+import jax.numpy as jnp
+
+import numpyro
+from numpyro.contrib.funsor import markov
+import numpyro.distributions as dist
+from numpyro.infer import MCMC, NUTS
+
+
+def test_gaussian_mixture_model():
```
eb8680, line 16: Do you consider these tests sufficient for merging this PR?  It's up to you - as I mentioned before, there are a few extra tests in the Pyro version that don't depend on Pyro's old enumeration machinery you could port if you wanted to.

numpyro/contrib/funsor/enum_messenger.py
```diff
@@ -0,0 +1,634 @@
...
+        return super().__exit__(*args, **kwargs)
+
+
+##########################################
+# User-facing handler implementations
+##########################################
+
+class plate(GlobalNamedMessenger):
+    """
+    An alternative implementation of :class:`numpyro.primitives.plate` primitive.
```
eb8680, line 420: Clarify here that only this version of `plate` is compatible with enumeration?  Might also want to add a pointer to the `plate_to_enum_plate` context manager.

test/contrib/test_funsor.py
fehiepsi: Oh, I missed that point. I'll try to port some of them here. Thanks!

test/contrib/test_funsor_tve_smoke.py
```diff
@@ -0,0 +1,78 @@
...
+import jax
+import jax.numpy as jnp
+
+import funsor
+from funsor.einsum import einsum
+from funsor.testing import make_einsum_example, make_hmm_einsum, make_plated_hmm_einsum
+from numpyro.contrib.indexing import Vindex
+
+funsor.set_backend("jax")
+pytestmark = pytest.mark.skipif('CI' in os.environ, reason="those tests are slow in CI")
```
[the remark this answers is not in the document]
fehiepsi: Yeah, I'll delete it.

test/contrib/test_funsor.py
eb8680: You can probably just copy this file: https://github.com/pyro-ppl/pyro/pull/2475/files#diff-90ba92d31fd3a057903e4e0663d72b99

fehiepsi: Thanks, I just port it here.

setup.py
```diff
@@ -43,6 +45,8 @@
         'doc': ['sphinx', 'sphinx_rtd_theme', 'sphinx-gallery'],
         'test': [
             'flake8',
+            # TODO: make sure that we release funsor before releasing numpyro
+            'funsor',
```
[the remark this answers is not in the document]
fehiepsi: Thanks, I just remove it because it is duplicated.

eb8680:
As the original PR author I can't approve or merge this directly myself, but feel free to merge when you're ready.

test/contrib/test_funsor.py
```diff
@@ -0,0 +1,190 @@
...
+        n_count_data = count_data.shape[0]
+        alpha = 1 / jnp.mean(count_data)
+        lambda_1 = numpyro.sample('lambda_1', dist.Exponential(alpha))
+        lambda_2 = numpyro.sample('lambda_2', dist.Exponential(alpha))
+        # this is the same as DiscreteUniform(0, 69)
+        tau = numpyro.sample('tau', dist.Categorical(logits=jnp.zeros(70)))
+        idx = jnp.arange(n_count_data)
+        lambda_ = jnp.where(tau > idx, lambda_1, lambda_2)
+        with numpyro.plate("data", n_count_data):
+            numpyro.sample('obs', dist.Poisson(lambda_), obs=count_data)
```
fehiepsi, line 75: @eb8680 Could you help me debug this test? I think currently, enum only works when discrete sites are inside a `plate` statement. :(

fehiepsi:
Thanks for reviewing @eb8680! I'll discuss with Neeraj about the api changes and having `funsor` as a dependency.

test/contrib/test_funsor.py
eb8680: Just pushed a fix - `enum` wasn't properly passing calls to `to_data` to its parent handler.

fehiepsi: Awesome! Thanks a lot, Eli!

numpyro/primitives.py
```diff
@@ -84,6 +84,10 @@ def sample(name, fn, obs=None, rng_key=None, sample_shape=()):
     :param numpy.ndarray obs: observed value
     :param jax.random.PRNGKey rng_key: an optional random key for `fn`.
     :param sample_shape: Shape of samples to be drawn.
+    :param dict infer: an optional dictionary containing additional information
+        for inference algorithms. For example, if `fn` is a discrete distribution,
+        setting `infer={'enumerate': 'parallel'}` will tell MCMC marginalize
```
neerajprad, line 89: nit: `to tell`.

neerajprad: We can fix this subsequently. I'm merging this so that your other PR is unblocked.

PR #573, opened 2020-04-21
add enumerate_support method for discrete distributions

fehiepsi:
This is required to support enumeration in #572.

There is an edge case in Binomial under tracing: when `total_count` is not constant, the error will not be raised. Actually, when `total_count` is not constant, we can still consider `arange(0, max(total_count) + 1)` as the support (in the ExtendedBinomial sense - with `validate_args` is True). It is not important though.

### DIFF
--- a/numpyro/distributions/discrete.py
+++ b/numpyro/distributions/discrete.py
@@ -95,8 +95,11 @@ class BernoulliProbs(Distribution):
     def variance(self):
         return self.probs * (1 - self.probs)
 
-    def enumerate_support(self):
-        return np.arange(2)
+    def enumerate_support(self, expand=False):
+        values = np.arange(2).reshape((-1,) + (1,) * len(self.batch_shape))
+        if expand:
+            values = np.broadcast_to(values, values.shape[:1] + self.batch_shape)
+        return values
 
 
 @copy_docs_from(Distribution)
@@ -129,8 +132,11 @@ class BernoulliLogits(Distribution):
     def variance(self):
         return self.probs * (1 - self.probs)
 
-    def enumerate_support(self):
-        return np.arange(2)
+    def enumerate_support(self, expand=False):
+        values = np.arange(2).reshape((-1,) + (1,) * len(self.batch_shape))
+        if expand:
+            values = np.broadcast_to(values, values.shape[:1] + self.batch_shape)
+        return values
 
 
 def Bernoulli(probs=None, logits=None, validate_args=None):
@@ -177,14 +183,17 @@ class BinomialProbs(Distribution):
     def support(self):
         return constraints.integer_interval(0, self.total_count)
 
-    def enumerate_support(self):
+    def enumerate_support(self, expand=False):
         total_count = np.amax(self.total_count)
         if not_jax_tracer(total_count):
             # NB: the error can't be raised if inhomogeneous issue happens when tracing
             if np.amin(self.total_count) != total_count:
                 raise NotImplementedError("Inhomogeneous total count not supported"
                                           " by `enumerate_support`.")
-        return np.arange(total_count + 1)
+        values = np.arange(total_count + 1).reshape((-1,) + (1,) * len(self.batch_shape))
+        if expand:
+            values = np.broadcast_to(values, values.shape[:1] + self.batch_shape)
+        return values
 
 
 @copy_docs_from(Distribution)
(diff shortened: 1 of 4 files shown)


numpyro/distributions/discrete.py
```diff
@@ -165,6 +171,15 @@ def variance(self):
     def support(self):
         return constraints.integer_interval(0, self.total_count)
 
+    def enumerate_support(self):
```
neerajprad, line 174: @eb8680 - Do we also need the `expand` keyword for `enumerate_support`?

fehiepsi: I did `grep` in Pyro and found no such usage case. :D

neerajprad: I think you'll probably find it in the code for `enum_messenger`.

eb8680: > Do we also need the expand keyword for enumerate_support?

Not really, it should be fine to only implement `expand=False`.

fehiepsi:
@eb8680 do you need `has_enumerate_support` to filter out enumerate sites?

@neerajprad it seems that [in Pyro](https://github.com/pyro-ppl/pyro/blob/dev/pyro/infer/mcmc/util.py#L393), we still let HMC run even for discrete latent variables that do not have `enumerate_support` (e.g. Poisson). I guess we should add some checks for `support` of those sites to raise the issue for users? In NumPyro, we can easily fix the issue by adding an attribute `is_discrete` (or make those discrete distributions subclass DiscreteDistribution). WDYT? I can address the issue in the follow-up PR.

eb8680:
> do you need has_enumerate_support to filter out enumerate sites?

Yeah, that would be useful.

neerajprad:
Yes, I think that will be useful since we need to exclude it in other utilities too. Adding a static `is_discrete` seems reasonable.

test/test_distributions.py
```diff
@@ -962,3 +962,16 @@ def test_generated_sample_distribution(jax_dist, sp_dist, params,
...
+
+
+@pytest.mark.parametrize('jax_dist, params, support', [
+    (dist.BernoulliLogits, (5.,), np.arange(2)),
+    (dist.BernoulliProbs, (0.5,), np.arange(2)),
+    (dist.BinomialLogits, (4.5, 10), np.arange(11)),
+    (dist.BinomialProbs, (0.5, 11), np.arange(12)),
+    (dist.BetaBinomial, (2., 0.5, 12), np.arange(13)),
+    (dist.CategoricalLogits, (np.array([3., 4., 5.]),), np.arange(3)),
+    (dist.CategoricalProbs, (np.array([[0.1, 0.5, 0.4]]),), np.arange(3)),
```
neerajprad, line 974: The shape here should be `(3, 1)` right for the additional batch dim? Maybe we can add some shape asserts too in the test.

fehiepsi: I am not sure if we need to add additional dimensions for `enumerate_support`. Based on [Pyro implementation](https://github.com/pyro-ppl/pyro/pull/2307/files#diff-70ab0ab836b4c309f534671a06651eeaR97), it seems to be not required.

neerajprad: Hmm..the distribution's `enumerate_support` takes into account batch dims.
e.g.
```python
In [4]: dist.Categorical(torch.tensor([[0.1, 0.9]])).enumerate_support()
Out[4]:
tensor([[0],
        [1]])

In [5]: dist.Categorical(torch.tensor([0.1, 0.9])).enumerate_support()
Out[5]: tensor([0, 1])
```

fehiepsi: Yes, but with `expand=False`, it just appends [singleton batch dimensions](https://github.com/pytorch/pytorch/blob/master/torch/distributions/categorical.py#L124). With @eb8680 's implementation, those singleton dimensions are squeezed.

neerajprad: I see. I think there is benefit in keeping the distribution's interface separate and consistent with pytorch's shape semantics. We can munge these shapes for the purpose of inference if needed. An alternative is to not expose this method in the distribution's interface and just provide a private utility function to aid @eb8680's refactor (making use of the integer interval support for the distribution's instance).

eb8680: I agree with @neerajprad, best to stick closely to PyTorch distribution semantics for `enumerate_support`.

fehiepsi: IMO if inhomogeneous enumerate support is not supported, there is no reason to match batch_shape. It is not important so I'll add `expand` argument for consistency.

eb8680: > IMO if inhomogeneous enumerate support is not supported, there is no reason to match batch_shape.

The bigger-picture rationale for matching `batch_shape` in `enumerate_support` is that other algorithms (e.g. Pyro's `TraceTMC_ELBO`) produce sample tensors that are not homogeneous, and we want to be able to reuse the same filtering and variable elimination code across as many such algorithms as possible.

fehiepsi: Thanks for your inputs, @neerajprad @eb8680 !

Issue #554, opened 2020-03-07
'Normal' object has no attribute 'expand'


liseeocean:
Hi,
I find distributions in numpyro don't have .expand but pyro have. How can I get this one?
It works in pyro:
import pyro.distributions as dist
N=10

wi=dist.Normal(0., 1.).expand([N, N]).to_event(2)

Not work in numpyro:

import numpyro.distributions as dist
N=10

wi=dist.Normal(0., 1.).expand([N, N]).to_event(2)

fehiepsi:
@liseeocean This seems like a duplicated issue of https://github.com/pyro-ppl/numpyro/issues/547#issuecomment-591486854. Could you check if it solves your problem?

liseeocean:
Thank you very much for your quick reply! Do you know how to use "from pyro.nn import PyroSample" in numpyro? 
In Pyro:
self.linear = PyroModule[nn.Linear](hidden_size, output_size)
self.fc1.weight = PyroSample(dist.Normal(loc=0., scale=1.))

fehiepsi:
The `.expand()` method is available now. Thanks for making this feature request!

PR #574, opened 2020-04-22
Implement ExpandedDistribution

fehiepsi:
Addresses #554. The implementation is almost identical to Pyro version. I also made `enumerate_support(expand=True)` by default to match PyTorch behavior.

### DIFF
--- a/test/test_distributions.py
+++ b/test/test_distributions.py
@@ -982,3 +982,29 @@ def test_enumerate_support_smoke(jax_dist, params, support, batch_shape, expand)
     if expand:
         expected = np.broadcast_to(expected, support.shape + batch_shape)
     assert_allclose(actual, expected)
+
+
+@pytest.mark.parametrize('jax_dist, sp_dist, params', CONTINUOUS + DISCRETE)
+@pytest.mark.parametrize('prepend_shape', [
+    (),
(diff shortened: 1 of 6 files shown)


test/test_distributions.py
```diff
@@ -982,3 +982,29 @@ def test_enumerate_support_smoke(jax_dist, params, support, batch_shape, expand)
...
+
+@pytest.mark.parametrize('jax_dist, sp_dist, params', CONTINUOUS + DISCRETE)
+@pytest.mark.parametrize('prepend_shape', [
+    (),
+    (2, 3),
+])
+@pytest.mark.parametrize('sample_shape', [
+    (),
+    (4,),
+])
```
neerajprad, line 995: I think we should also add tests to check the interaction of `.to_event` and `.expand` and `log_prob` (you could take a look at `check_sample_shapes` in `test_distributions` in Pyro which is used to test many of these corner cases), but feel free to do that in a separate PR since you are currently blocked by this. Also, scipy distributions don't consistently follow the tensor shaping semantics that we have in pytorch, so it might be best to not take that as a reference.

fehiepsi: Thanks, @neerajprad ! I'll add it to Masked PR.

Issue #520, opened 2020-01-13
Create support for MaskedDistribution


neerajprad:
Refer to the discussion in https://forum.pyro.ai/t/hmm-like-model-with-sequences-of-different-lengths/1507.

TuanNguyen27:
I'm interested in helping out with this !

neerajprad:
@TuanNguyen27 - That's great. You could take a look at Pyro's [MaskedDistribution](https://github.com/pyro-ppl/pyro/blob/dev/pyro/distributions/torch_distribution.py#L227) and [distribution.mask](https://github.com/pyro-ppl/pyro/blob/dev/pyro/distributions/torch_distribution.py#L147) method. Let us know if you need any help.

fehiepsi:
@TuanNguyen27 Did you have a chance to work on this feature? If you want, I can make a PR for this. I need to use this distribution to test out some codes.

TuanNguyen27:
ah, please go head, i have't had time to work on this :)

PR #575, opened 2020-04-23
Implement MaskedDistribution

fehiepsi:
Addresses #520.

Blocked by #574 because we need `expand` method for broadcasting `base_dist` if `mask.ndim > len(batch_shape)`.

 #### TODO

- [x] add more tests for expand

### DIFF
--- a/numpyro/distributions/discrete.py
+++ b/numpyro/distributions/discrete.py
@@ -48,7 +48,7 @@ from numpyro.distributions.util import (
     sum_rightmost,
     validate_sample,
 )
-from numpyro.util import copy_docs_from
+from numpyro.util import copy_docs_from, not_jax_tracer
 
 
 def _to_probs_bernoulli(logits):
@@ -73,6 +73,8 @@ def _to_logits_multinom(probs):
 class BernoulliProbs(Distribution):
     arg_constraints = {'probs': constraints.unit_interval}
     support = constraints.boolean
+    has_enumerate_support = True
+    is_discrete = True
 
     def __init__(self, probs, validate_args=None):
         self.probs = probs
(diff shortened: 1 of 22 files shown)


fehiepsi:
@neerajprad @eb8680 I have also added a couple of bug fixes which are exercised in funsor. This should unblock https://github.com/pyro-ppl/funsor/pull/326. There is only one issue remaining related to [`has_rsample` attribute](https://github.com/pyro-ppl/funsor/blob/master/funsor/distributions.py#L126). Should we add `has_rsample` flag to NumPyro distributions?

numpyro/distributions/discrete.py
```diff
@@ -294,7 +294,7 @@ def variance(self):
 
     @property
     def support(self):
-        return constraints.integer_interval(0, np.shape(self.probs)[-1])
+        return constraints.integer_interval(0, np.shape(self.probs)[-1] - 1)
```
fehiepsi, line 297: this is a bug

numpyro/distributions/discrete.py
```diff
@@ -342,10 +342,10 @@ def variance(self):
 
     @property
     def support(self):
-        return constraints.integer_interval(0, np.shape(self.logits)[-1])
+        return constraints.integer_interval(0, np.shape(self.logits)[-1] - 1)
 
     def enumerate_support(self, expand=True):
-        values = np.arange(self.probs.shape[-1]).reshape((-1,) + (1,) * len(self.batch_shape))
+        values = np.arange(self.logits.shape[-1]).reshape((-1,) + (1,) * len(self.batch_shape))
```
fehiepsi, line 348: we don't need to use `self.probs` here for CategoricalLogits

eb8680:
> Should we add has_rsample flag to NumPyro distributions?

Might as well, we'll need it if/when we add fancier VI methods to NumPyro via Funsor.

PR #580, opened 2020-04-27
Port pyro.ops.indexing to a new numpyro.contrib.indexing

eb8680:
This PR ports Pyro's `pyro.ops.indexing` module, an implementation of [NEP 21](https://numpy.org/neps/nep-0021-advanced-indexing.html), to NumPyro as a new `numpyro.contrib.indexing` module.  The `Vindex` utility is extremely useful for working with enumerated models like the ones in #572.

The code and tests from Pyro have been ported almost verbatim.

### DIFF
--- a/numpyro/contrib/indexing.py
+++ b/numpyro/contrib/indexing.py
@@ -5,7 +5,7 @@ import jax.numpy as np
 
 
 def _is_batched(arg):
-    return isinstance(arg, np.ndarray) and len(arg.shape) > 0
+    return np.ndim(arg) > 0
 
 
 def vindex(tensor, args):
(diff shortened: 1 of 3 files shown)


numpyro/contrib/indexing.py
```diff
@@ -0,0 +1,145 @@
+# Copyright Contributors to the Pyro project.
+# SPDX-License-Identifier: Apache-2.0
+
+import jax.numpy as np
+
+
+def _is_batched(arg):
+    return isinstance(arg, np.ndarray) and len(arg.shape) > 0
```
fehiepsi, line 8: nit: it is better to return `np.ndim(x) > 0` here, which also works for DeviceArray and Python scalars.

numpyro/contrib/indexing.py
```diff
@@ -0,0 +1,145 @@
...
+    [1] https://www.numpy.org/neps/nep-0021-advanced-indexing.html
+        introduces ``vindex`` as a helper for vectorized indexing.
+        This implementation is similar to the proposed notation
+        ``x.vindex[]`` except for slightly different handling of ``Ellipsis``.
+
+    :param np.ndarray tensor: A tensor to be indexed.
+    :param tuple args: An index, as args to ``__getitem__``.
+    :returns: A nonstandard interpetation of ``tensor[args]``.
+    :rtype: np.ndarray
+    """
```
fehiepsi, line 79: nit: could you add an assertation here that `np.ndim(tensor) > 0`.

Issue #189, opened 2019-06-04
Call for user contributed examples and tutorials


neerajprad:
A good way to get acquainted with NumPyro would be to write your model of interest and contribute it to [examples](https://github.com/pyro-ppl/numpyro/tree/master/examples) or as a [notebook](https://github.com/pyro-ppl/numpyro/tree/master/notebooks). 

If you are interested, please open up an issue describing your model so that one of the core contributors can help assess feasibility, put into place any features that we are currently lacking, and work with you on the PR. Also, please feel free to suggest topics / examples that you would like to see covered.

PR #581, opened 2020-04-28
Adding example for testing proportions

freddyaboulton:
In reference to issue #189, I added an example for testing if two proportions are equal using logistic regression. This example uses a self-contained, simulated dataset.

### DIFF
--- a/examples/proportion_test.py
+++ b/examples/proportion_test.py
@@ -87,7 +87,7 @@ def print_results(coef: np.ndarray, interval_size: float = 0.95) -> None:
           f"{(100 * impact_on_probability[0]):.2} to {(100 * impact_on_probability[1]):.2} percentage points."
           )
 
-    print("There is a 95% probability the effect of gender on the log odds of conversion "
+    print(f"There is a {interval_size * 100}% probability the effect of gender on the log odds of conversion "
           f"lies in the interval ({effect_of_gender[0]:.2}, {effect_of_gender[1]:.2f})."
           " Since this interval contains 0, we can conclude gender does not impact the conversion rate.")
 
--- a/examples/proportion_test.py
+++ b/examples/proportion_test.py
@@ -78,9 +78,9 @@ def print_results(coef: np.ndarray, interval_size: float = 0.95) -> None:
     baseline_response = expit(coef[:, 0])
     response_with_calls = expit(coef[:, 0] + coef[:, 1])
 
-    impact_on_probability = hpdi(response_with_calls - baseline_response, prob=0.95)
+    impact_on_probability = hpdi(response_with_calls - baseline_response, prob=interval_size)
 
-    effect_of_gender = hpdi(coef[:, 2], prob=0.95)
+    effect_of_gender = hpdi(coef[:, 2], prob=interval_size)
 
     print(f"There is a {interval_size * 100}% probability that calling customers "
           "increases the chance they'll make a purchase by "
--- a/examples/proportion_test.py
+++ b/examples/proportion_test.py
@@ -1,0 +1,134 @@
+"""
+Proportion Test
+===============================
+You are managing a business and want to test if calling your customers will
+increase their chance of making a purchase. You get 100,000 customer records and call
+roughly half of them and record if they make a purchase in the next three months. 
+You do the same for the half that did not get called. After three months, the data is in - 
+did calling help?
+
+This example answers this question by estimating a logistic regression model where the
+covariates are whether the customer got called and their gender. We place a multivariate
+normal prior on the regression coefficients. We report the 95% highest posterior
+density interval for the effect of making a call.
+"""
+
+
+import argparse
+import os
+from typing import Any, Callable, Tuple
+
+import jax.numpy as np
+import numpyro
+import numpyro.distributions as dist
+import pandas as pd
+from jax import random
+from jax.scipy.special import expit
+from numpyro.diagnostics import hpdi
+from numpyro.infer import MCMC, NUTS
+from scipy import stats
+
+
+def make_dataset() -> Tuple[np.ndarray, np.ndarray]:
+    """
+    Make simulated dataset where potential customers who get a 
+    sales calls have ~2% higher chance of making another purchase.
+    """
+    made_purchase_got_called = stats.bernoulli.rvs(0.084, size=51342)
+    made_purchase_no_calls = stats.bernoulli.rvs(0.061, size=48658)
+    
+    made_purchase = np.concatenate([made_purchase_got_called, made_purchase_no_calls])
+    
+    is_female = stats.bernoulli.rvs(0.5, size=100000)
+    got_called = np.concatenate([np.ones(51342), np.zeros(48658)])
+    design_matrix = np.hstack([np.ones((100000, 1)),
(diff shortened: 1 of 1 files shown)


fehiepsi:
Thanks for contributing this nice example, @freddyaboulton ! I'll review this in detail tomorrow.

examples/proportion_test.py
```diff
@@ -0,0 +1,134 @@
...
+from numpyro.infer import MCMC, NUTS
+from scipy import stats
+
+
+def make_dataset() -> Tuple[np.ndarray, np.ndarray]:
+    """
+    Make simulated dataset where potential customers who get a 
+    sales calls have ~2% higher chance of making another purchase.
+    """
+    made_purchase_got_called = stats.bernoulli.rvs(0.084, size=51342)
```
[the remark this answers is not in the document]
neerajprad: Can we also spell out `num_calls = 51342` and `num_no_calls = 48658` separately.

examples/proportion_test.py
```diff
@@ -0,0 +1,134 @@
...
+    """
+    Print the confidence interval for the effect size with interval_size
+    probability mass.
+    """
+
+    baseline = hpdi(coef[:, 0], prob=0.95)
+    effect_size = hpdi(coef[:, 1], prob=0.95)
+    
+    baseline_response = expit(baseline)
+    response_with_calls = expit(baseline + effect_size)
```
neerajprad, line 80: Should we instead do this on the raw tensors and take `hpdi` later? Also, should we say something about the effect / non-effect of gender?

examples/proportion_test.py
```diff
@@ -0,0 +1,134 @@
...
+    
+    impact_on_probability = response_with_calls - baseline_response
+    
+    print(f"There is a {interval_size * 100}% probability that calling customers "
+          "increases the chance they'll make a purchase by "
```
neerajprad, line 85: This looks good to me since we constructed the dataset, but users extrapolating it to more complex examples with multiple predictors might be problematic. @fehiepsi, @martinjankowiak: WDYT?

examples/proportion_test.py
```diff
@@ -0,0 +1,134 @@
...
+                               got_called.reshape(-1, 1),
+                               is_female.reshape(-1, 1)])
+    
+    return design_matrix, made_purchase
+
+
+def model(design_matrix: np.ndarray, outcome: np.ndarray = None) -> None:
+    """
+    Model definition: Log odds of making a purchase is a linear combination
+    of covariates. Specify aNormal prior over regression coefficients.
```
martinjankowiak, line 54: typo: aNormal

examples/proportion_test.py
```diff
@@ -0,0 +1,134 @@
...
+    :param design_matrix: Covariates. All categorical variables have been one-hot
+        encoded.
+    :param outcome: Binary response variable. In this case, whether or not the
+        customer made a purchase.
+    """
+    
+    # Use multivariate normal prior in case we want to add more covariates later.
```
martinjankowiak, line 61: does this comment still make sense?

examples/proportion_test.py
```diff
@@ -0,0 +1,134 @@
...
+    beta = numpyro.sample('coefficients', dist.MultivariateNormal(loc=0.,
+                                                                  covariance_matrix=np.eye(design_matrix.shape[1])))
+    logits = design_matrix.dot(beta)
+    
+    with numpyro.plate('data', design_matrix.shape[0]):
+            numpyro.sample('obs', dist.Bernoulli(logits=logits), obs=outcome)
```
martinjankowiak, line 67: nit: spacing?

examples/proportion_test.py
martinjankowiak: i think this is fine for this self-contained example 👍

examples/proportion_test.py
```diff
@@ -16,76 +16,81 @@
...
+
     print(f"There is a {interval_size * 100}% probability that calling customers "
           "increases the chance they'll make a purchase by "
           f"{(100 * impact_on_probability[0]):.2} to {(100 * impact_on_probability[1]):.2} percentage points."
-         )
-    
+          )
+
+    print("There is a 95% probability the effect of gender on the log odds of conversion "
+          f"lies in the interval ({effect_of_gender[0]:.2}, {effect_of_gender[1]:.2f})."
```
fehiepsi, line 91: Oh, I just learn that we can format f-string like this. Nice!

examples/proportion_test.py
```diff
@@ -16,76 +16,81 @@
...
     :param design_matrix: Covariates. All categorical variables have been one-hot
         encoded.
     :param outcome: Binary response variable. In this case, whether or not the
         customer made a purchase.
     """
-    
-    # Use multivariate normal prior in case we want to add more covariates later.
+
     beta = numpyro.sample('coefficients', dist.MultivariateNormal(loc=0.,
                                                                   covariance_matrix=np.eye(design_matrix.shape[1])))
```
fehiepsi, line 65: nit: this can be faster if you replace `covariance_matrix` by `scale_tril`. It is still fine if you want to leave it as-is.

examples/proportion_test.py
```diff
@@ -16,76 +16,81 @@
...
-    
-    baseline_response = expit(baseline)
-    response_with_calls = expit(baseline + effect_size)
-    
-    impact_on_probability = response_with_calls - baseline_response
-    
+    baseline_response = expit(coef[:, 0])
+    response_with_calls = expit(coef[:, 0] + coef[:, 1])
+
+    impact_on_probability = hpdi(response_with_calls - baseline_response, prob=0.95)
```
fehiepsi, line 81: I think you need to replace `0.95` by `interval_size`.

examples/proportion_test.py
```diff
@@ -16,76 +16,81 @@
...
+
+    effect_of_gender = hpdi(coef[:, 2], prob=0.95)
```
fehiepsi, line 83: ditto:  `0.95` -> `interval_size`

examples/proportion_test.py
```diff
@@ -16,76 +16,81 @@
...
+
     print(f"There is a {interval_size * 100}% probability that calling customers "
           "increases the chance they'll make a purchase by "
           f"{(100 * impact_on_probability[0]):.2} to {(100 * impact_on_probability[1]):.2} percentage points."
-         )
-    
+          )
+
+    print("There is a 95% probability the effect of gender on the log odds of conversion "
```
fehiepsi, line 90: ditto:  `0.95` -> `interval_size`

examples/proportion_test.py
fritzo: Just be careful to avoid them in Pyro since Pyro still supports Python 3.5 which doesn't allow f-strings.

freddyaboulton:
@fehiepsi @martinjankowiak @neerajprad 

I think I've addressed your comments! Please let me know if you need me to change anything else.

examples/proportion_test.py
fehiepsi: This is still not addressed yet.

freddyaboulton:
@fehiepsi Thank you for the help! Should be all good now.

PR #584, opened 2020-04-30
Reduce compiling time in iterative_build_tree

fehiepsi:
Currently, in `iterative_build_tree`, we compile potential_energy 2 times: one for the first leaf and the other one for the loop. With this PR, we will only compile 1 time.

### Benchmark:

+ time to compile `hmm` example is reduced from 28s to 14s.

### TODO

- [x] There is something wrong for `cond` with `jaxlib >= 0.1.40`. The HMM result is weird.

### DIFF
--- a/numpyro/infer/hmc_util.py
+++ b/numpyro/infer/hmc_util.py
@@ -630,7 +630,7 @@ def _iterative_build_subtree(prototype_tree, vv_update, kinetic_fn,
                         lambda x: _combine_tree(*x, False))
 
         leaf_idx = current_tree.num_proposals
-        # NB: in the special case leaf_idx=0, ckpt_idx_min=1 and ckpt_idx_max=1,
+        # NB: in the special case leaf_idx=0, ckpt_idx_min=1 and ckpt_idx_max=0,
         # the following logic is still valid for that case
         ckpt_idx_min, ckpt_idx_max = _leaf_idx_to_ckpt_idxs(leaf_idx)
         r, _ = ravel_pytree(new_leaf.r_right)
--- a/numpyro/infer/hmc_util.py
+++ b/numpyro/infer/hmc_util.py
@@ -561,11 +561,9 @@ def _get_leaf(tree, going_right):
 def _double_tree(current_tree, vv_update, kinetic_fn, inverse_mass_matrix, step_size,
                  going_right, rng_key, energy_current, max_delta_energy, r_ckpts, r_sum_ckpts):
     key, transition_key = random.split(rng_key)
-    # If we are going to the right, start from the right leaf of the current tree.
-    z, r, z_grad = _get_leaf(current_tree, going_right)
 
-    new_tree = _iterative_build_subtree(current_tree.depth, vv_update, kinetic_fn,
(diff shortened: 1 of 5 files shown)


numpyro/infer/hmc_util.py
```diff
@@ -621,13 +619,19 @@ def _cond_fn(state):
...
         z, r, z_grad = _get_leaf(current_tree, going_right)
         new_leaf = _build_basetree(vv_update, kinetic_fn, z, r, z_grad, inverse_mass_matrix, step_size,
                                    going_right, energy_current, max_delta_energy)
-        new_tree = _combine_tree(current_tree, new_leaf, inverse_mass_matrix, going_right,
-                                 transition_rng_key, False)
+        new_tree = cond(current_tree.num_proposals == 0,
+                        new_leaf,
+                        lambda x: x,
+                        (current_tree, new_leaf, inverse_mass_matrix, going_right, transition_rng_key),
+                        lambda x: _combine_tree(*x, False))
```
fehiepsi, line 630: Now, this `logsumexp` issue happens with `hmm` example. The behavior is pretty weird because if we use the upstream logsumexp, keep the current code in master branch as is, change this line to
```python
        new_tree = cond(current_tree.num_proposals == 0,
                        new_leaf,
                        lambda x: x,
                        (current_tree, new_leaf, inverse_mass_matrix, going_right, transition_rng_key),
                        lambda x: _combine_tree(*x, False))
```
, the issue (all samples are constant) will happen. Note that if we only change this line, the `true` branch will not happen, so the execution should be the same as before (i.e. without using `cond`). Even if I go ahead and set the condition to `current_tree.num_proposals < -1000`, the issue still happens. But with the condition `current_tree.num_proposals != current_tree.num_proposals`, the issue no longer happens.

This points out there are some serious issues (I think so because this might not only happen with `logsumexp`, but also with some other stuffs) which I can't answer (also can't make a simple reproducible code):
+ Why adding `cond` here affects the behavior of `logsumexp`?
+ Why we have different behaviors when the conditions of `cond` are False (using different statements: `x != x`, `x < 0`).
+ Even without using `cond` here, `logsumexp` issue still happens in `neutra` example.
+ `logsumexp` issue does not happen when we disable progress_bar (i.e. running MCMC in a `lax.fori_loop` loop).

I tried various versions of jax/jaxlib and observed the following behavior:
+ `logsumexp` issue happens in `neutra` example for all jaxlib versions
+ `logsumexp` issue happens in `hmm` example with `jaxlib=0.1.40` (and its newer versions) and does not happen with `jaxlib==0.1.39` (using the same jax version `0.1.59`)

I only test in CPU because GPU is pretty slow for hmm.

@mattjj Do you have something in mind about this issue or can you give us some hints to track it down? Currently, this does not block us because we can use a custom `logsumexp` with `lax.stop_gradient`. However, I am a bit worried that the issue might happen somewhere else too. Also, it is a bit inconvenient for users to use a custom `logsumexp`. I wanted to make a simple reproducible code but it is hard to create one (I'm not sure when the issue will happen - as you see, the issue can happen with unrelated changes at `cond` above). Thanks in advance!

Issue #579, opened 2020-04-26
Add logistic distribution


stuhlmueller:
Submissions to [Metaculus](https://www.metaculus.com/) are represented as mixtures of logistic distributions. Having access to logistic distributions would make it easier to use Numpyro in [Ergo](https://github.com/oughtinc/ergo/).

PR #586, opened 2020-05-01
Add logistic distribution

freddyaboulton:
In reference to #579, this PR implements a logistic distribution with parameters `loc` and `scale`.

### DIFF
--- a/numpyro/distributions/continuous.py
+++ b/numpyro/distributions/continuous.py
@@ -1082,3 +1082,34 @@ class Uniform(TransformedDistribution):
     @property
     def variance(self):
         return (self.high - self.low) ** 2 / 12.
+
+
+@copy_docs_from(Distribution)
+class Logistic(Distribution):
+    arg_constraints = {'loc': constraints.real, 'scale': constraints.positive}
+    support = constraints.real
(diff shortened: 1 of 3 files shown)


numpyro/distributions/continuous.py
```diff
@@ -1082,3 +1082,34 @@ def mean(self):
...
+        super(Logistic, self).__init__(batch_shape, validate_args=validate_args)
+
+    def sample(self, key, sample_shape=()):
+        z = random.logistic(key, shape=sample_shape + self.batch_shape + self.event_shape)
+        return self.loc + z * self.scale
+
+    @validate_sample
+    def log_prob(self, value):
+        log_exponent = (self.loc - value) / self.scale
+        log_denominator = np.log(self.scale) + 2 * np.log(1 + np.exp(log_exponent))
```
fehiepsi, line 1105: nit: it is better to use `jax.nn.softplus(log_exponent)` instead of `np.log(1 + np.exp(log_exponent))`.

freddyaboulton: Thanks for the tip @fehiepsi ! Just pushed this change.

PR #588, opened 2020-05-04
add polya-gamma distribution

martinjankowiak:
this is the PG(b, 0) distribution. i do not implement the sampler because it's a bit involved and because i only intend to use this for hmc.

for the pdf see the equation at the bottom of page 5 [here](https://arxiv.org/abs/1205.0310)

computing this is a bit tricky because it's an infinite alternating sum. i simply truncate the sum. this is a bit problematic because as the value x grows larger we need more and more terms. on the other hand, if i'm doing hmc (and haven't initialized things in a crazy way) i should never see large x so i don't really care.

@fehiepsi any suggestions for doing this in a more comprehensive and/or numerically stable way? my current strategy is to treat even terms and odd terms separately. i think it should be reasonably stable but it'll definitely give wrong log_probs for very large x. maybe there's some asymptotic expansion for large x but i'm not sure if it's worth worrying about....

what do i need to add for distribution tests?

[this](https://github.com/pyro-ppl/numpyro/files/4576276/hist.pdf) histogram shows 50k samples collected by HMC using this density (suggesting that it's reasonably numerically stable)


numpyro/distributions/continuous.py
```diff
@@ -1114,3 +1114,38 @@ def mean(self):
...
+            1.5 * np.log(value[..., None]) - 0.125 * np.square(2.0 * all_indices + b) / value[..., None]
+        even_terms = np.take(log_terms, all_indices[::2], axis=-1)
+        odd_terms = np.take(log_terms, all_indices[1::2], axis=-1)
+        logsumexp_even = logsumexp(even_terms, axis=-1)
+        logsumexp_odd = logsumexp(odd_terms, axis=-1)
+        even_odd_ratio = np.exp(logsumexp_odd - logsumexp_even)
+        even_odd_ratio = np.clip(even_odd_ratio, a_max=1.0 - 1.0e-6)
+        log_sum = logsumexp_even + np.log1p(-even_odd_ratio)
+        large_value_result = log_sum
+        small_value_result = np.take(even_terms, 0, axis=-1)
```
fehiepsi, line 1149: Is this necessary? I guess if the later terms (index > 0) are small, then `logsumexp_even ~ first_term`.

martinjankowiak:
@fehiepsi updated the pr:
- added an approximate sampler (useful for initialization i guess)
- explicitly truncated the distribution so that HMC never goes to bad regions of the log_prob
- made the distribution parameter free (concentration =  1)
- consequently simplified some of the numerics

numpyro/distributions/continuous.py
```diff
@@ -1114,3 +1114,37 @@ def mean(self):
...
+class TruncatedPolyaGamma(Distribution):
+    truncation_point = 4.0
+    num_log_prob_terms = 11
+    num_gamma_variates = 20
+    assert num_log_prob_terms % 2 == 1
+
+    arg_constraints = {}
+    support = constraints.interval(0.0, truncation_point)
+
+    def __init__(self, batch_shape=(), validate_args=None):
```
martinjankowiak, line 1129: is it ok to have batch_shape as an arg like this?

test/test_distributions.py
```diff
@@ -1012,3 +1012,22 @@ def test_expand(jax_dist, sp_dist, params, prepend_shape, sample_shape):
...
+
+@pytest.mark.parametrize('batch_shape', [
+    (),
+    (4,),
+])
+def test_polya_gamma(batch_shape, num_points=20000):
+    d = dist.TruncatedPolyaGamma(batch_shape=batch_shape)
+    rng_key = random.PRNGKey(0)
+
+    # test density approximately normalized
```
martinjankowiak, line 1025: note that this also tests for numerical stability

numpyro/distributions/continuous.py
fehiepsi: yup, it is okie.

martinjankowiak:
yet's it's only an approximation but the log_prob is good to one part in a million or so, and the sampler will have some edge effects near the truncation point, but for HMC this is adequate (i've tested elsewhere)

martinjankowiak:
@fehiepsi any idea why this isn't passing travis? does it just need to be restarted?

fehiepsi:
I think the new version of jaxlib breaks Travis. Let me look closer and make a fix.

martinjankowiak:
@fehiepsi i simplified a bit further. this should be good to go once tests pass 👍

numpyro/distributions/continuous.py
```diff
@@ -1143,8 +1143,6 @@ def log_prob(self, value):
...
         even_terms = np.take(log_terms, all_indices[::2], axis=-1)
         odd_terms = np.take(log_terms, all_indices[1::2], axis=-1)
-        logsumexp_even = logsumexp(even_terms, axis=-1)
-        logsumexp_odd = logsumexp(odd_terms, axis=-1)
-        even_odd_ratio = np.exp(logsumexp_odd - logsumexp_even)
-        even_odd_ratio = np.clip(even_odd_ratio, a_max=1.0 - 1.0e-6)
-        return logsumexp_even + np.log1p(-even_odd_ratio) - 0.5 * np.log(2.0 * np.pi)
+        sum_even = np.exp(logsumexp(even_terms, axis=-1))
+        sum_odd = np.exp(logsumexp(odd_terms, axis=-1))
+        return np.log(sum_even - sum_odd) - 0.5 * np.log(2.0 * np.pi)
```
fehiepsi, line 1148: I think this will fail if `sum_even < sum_odd`. Maybe we can clamp the diff by `np.finfo(value).tiny`?

martinjankowiak: this will never happen in the restricted range of the domain. this is one reason for restricting the range

test/test_distributions.py
martinjankowiak: note that computes the log prob throughout the entire domain

zoj613:
Hi @martinjankowiak . I'm not sure if this is relevent but seeing as though your implementation only involves the Gamma convolution method, maybe the on-going work on implementating this distribution in a Numpy/Scipy compatible way might be of interest to you with regards to expanding the code? See: https://github.com/zoj613/polya-gamma

martinjankowiak:
thanks @zoj613 i currently don't need a PG sampler but i'll keep this in mind if i do!

PR #589, opened 2020-05-07
Update jax version

fehiepsi:
This is also to diagnose the Travis issue at #588.

Currently, min jaxlib version is [0.1.45](https://github.com/google/jax/blob/master/jax/lib/__init__.py#L20).

### DIFF
--- a/test/test_flows.py
+++ b/test/test_flows.py
@@ -17,7 +17,7 @@ from numpyro.distributions.util import matrix_to_tril_vec
 
 def _make_iaf_args(input_dim, hidden_dims):
     _, rng_perm = random.split(random.PRNGKey(0))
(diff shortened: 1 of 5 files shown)


test/test_flows.py
```diff
@@ -17,7 +17,7 @@
 
 def _make_iaf_args(input_dim, hidden_dims):
     _, rng_perm = random.split(random.PRNGKey(0))
-    perm = random.shuffle(rng_perm, onp.arange(input_dim))
+    perm = random.permutation(rng_perm, onp.arange(input_dim))
```
neerajprad, line 20: The tests are still failing, I think there are a few more instances that we need to fix.

fehiepsi: Thanks Neeraj, it passed now.

PR #592, opened 2020-05-08
Replace instances of jax.partial with functools.partial.

hawkinsp:
`jax.partial` doesn't add anything useful over `functools.partial`, so there's no value to JAX exporting it from its API. It's better to use the standard Python library function instead.

### DIFF
--- a/numpyro/infer/svi.py
+++ b/numpyro/infer/svi.py
@@ -1,7 +1,7 @@
 # Copyright Contributors to the Pyro project.
 # SPDX-License-Identifier: Apache-2.0
 
-from functools import namedtuple
+from functools import namedtuple, partial
 
 import jax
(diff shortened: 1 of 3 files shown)


hawkinsp:
BTW it looks like your CI build is failing because you have pinned `jax` to an older version, but not also pinned `jaxlib`. You should probably pin both or neither.

neerajprad:
> BTW it looks like your CI build is failing because you have pinned jax to an older version, but not also pinned jaxlib. You should probably pin both or neither.

Thanks! This should be fixed by #589. I have restarted CI tests.

hawkinsp:
Our belief is that `jax.partial` is moot in Python 3. If you have a counterexample I'd very much like to hear it!

PR #601, opened 2020-05-20
Remove base_param_map

### LINKED ISSUE #587: substitute_fn argument in the substitute effect handler.
Is the way to use substitute_fn expected? 

I was expecting that values returned by substitute_fn will be unchanged. However, from the code it appears that these values will be base_value and will be transformed. This is not mentioned in the documentation.

https://github.com/pyro-ppl/numpyro/blob/25d055ad4b5a5e85a02159af87e8db1451243217/numpyro/handlers.py#L432

### LINKED ISSUE #594: AutoContinuous not working correctly with TransformedDistribution
The implementation of `AutoContinuous` will not work with `TransformedDistribution` objects, because the transformations specified for the distribution seem to be never applied. We encountered this issue with a `LogNormal` distribution and here is a minimal example:
```python
import jax.numpy as np
import jax

from numpyro.primitives import sample
from numpyro.infer import SVI, ELBO
from numpyro.contrib.autoguide import AutoDiagonalNormal
from numpyro.optim import SGD

import numpy as onp

data = .3 * onp.random.randn(100) + 1.

def model(x):
    mu = sample('mus', dist.Normal(0., 1.))
    sig = sample('sig', dist.LogNormal(0., 1.))
    sample('x', dist.Normal(mu, sig), obs=x)

optim = SGD(1e-4)
svi = SVI(model, guide, optim, ELBO())

rng_key = jax.random.PRNGKey(0) # works with 1, does not work with 0
svi_state = svi.init(rng_key, data)
init_svi_state = svi_state

for _ in range(10):
    svi_state, loss = svi.update(svi_state, data)
    print(loss)

```

If the key is seeded with 0, this will produce `nan` values (but it will work fine
for a seed of 1, or even if the order of mu and sig in the model is changed for
seed 0). The reason is that for a seed of 0 the guide samples negative initial values 
for sigmas (to which `LogNormal.log_prob` then applies a log).

We believe the issue lies in `AutoContinuous._setup_prototype` (particularly [line 154ff](https://github.com/pyro-ppl/numpyro/blob/5755a2ea845eb7e1e7878334cb711f70317b2512/numpyro/contrib/autoguide/__init__.py#L152))
where, for sample sites with intermediates, the guide just extracts a transform
from unconstrained optimization space to the support of a base distribution and
stores it in its `_inv_transforms` lookup, ignoring all other transformation the
site's distribution applies. The `__call__` implementation therefore applies
only the transform from unconstrained space to the constraint support of
the base distribution, not the target distribution.

The `LogNormal` is implemented as `TransformedDistribution(Normal, ExpTransform())`,
so it's base distribution has real support and therefore the guide extracts an
`IdentityTransform`. The `ExpTransform` seems to be never applied (at least we
could not pinpoint a location where that would be the case), thus the value
returned by the guide can be negative.

Note: `Pareto` is an interesting case, because it doesn't fall into the same
trap. It is a `TransformedDistribution`, applying an `ExpTransform` and an `AffineTransform`,
which the guide ignores, but it's base distribution is `Exponential`, which
has positive support and thus the guide will extract an `ExpTransform` anyways,
ensuring that the values it samples are positive (albeit probably not scaled properly).

~~A pull request with a suggested bugfix will follow soon.~~
Edit: We're not sure what would be the best fix for this.

From how the `__call__` function is written, it seems that a straightforward fix would be to change [line 153](https://github.com/pyro-ppl/numpyro/blob/5755a2ea845eb7e1e7878334cb711f70317b2512/numpyro/contrib/autoguide/__init__.py#L153) from `transform = biject_to(site['fn'].base_dist.support)`
to
`transform = ComposeTransform([biject_to(site['fn'].base_dist.support)] + site['fn'].transforms`, however, this comes with several caveats:
- it only works for `TransformedDistribution` classes that define a `transforms` attribute and will crash for all other distributions that sample intermediates
- there seems to be some way of applying the correct transformations to the latent sample from optimization space, because methods returning parameter values from an `SVIState` do that already. `_unpack_and_constrain` seems to be doing that maybe, but we're not quite sure and wouldn't know how to integrate it into `__call__`, which is currently not using `_unpack_and_constrain` but doing it's own stuff.

Edit 2: We have not much experience with Pyro, but from a quick look it seems that the implementation there is almost identical (minus the obvious structural differences and some case distinctions), so it would be interesting to know whether the same problems persists there, but we currently lack the time to do so..

fehiepsi:
With #599, we no longer need to do "automatic reparameterization" behind the scene for dynamic support. This PR removes stuff related to base_param_map.

Fixes #587 and fixes #594.

### TODO
- [x] remove stuffs in `infer.util`
- [x] make sure all tests pass, probably with the usage of "reparam"
- [x] remove dynamic support in autoguide

### DIFF
--- a/numpyro/contrib/autoguide/__init__.py
+++ b/numpyro/contrib/autoguide/__init__.py
@@ -10,11 +10,10 @@
 from abc import ABC, abstractmethod
 import warnings
 
-from jax import hessian, random, vmap
+from jax import hessian, random
 from jax.experimental import stax
 from jax.flatten_util import ravel_pytree
 import jax.numpy as np
-from jax.tree_util import tree_map
 
 import numpyro
 from numpyro import handlers
@@ -32,14 +31,12 @@ from numpyro.distributions.transforms import (
     biject_to
 )
 from numpyro.distributions.util import cholesky_of_inverse, sum_rightmost
-from numpyro.handlers import seed, substitute
 from numpyro.infer.elbo import ELBO
-from numpyro.infer.util import constrain_fn, find_valid_initial_params, init_to_uniform, log_density, transform_fn
+from numpyro.infer.util import find_valid_initial_params, init_to_uniform, transform_fn
 from numpyro.util import not_jax_tracer
 
 __all__ = [
     'AutoContinuous',
-    'AutoContinuousELBO',
     'AutoGuide',
     'AutoDiagonalNormal',
     'AutoMultivariateNormal',
@@ -113,14 +110,6 @@ class AutoContinuous(AutoGuide):
     Assumes model structure and latent dimension are fixed, and all latent
     variables are continuous.
 
-    .. note::
-        We recommend using :class:`AutoContinuousELBO` as the
-        objective function `loss` in :class:`~numpyro.infer.svi.SVI`.
-        In addition, we recommend using :meth:`sample_posterior` method
-        for drawing posterior samples from the autoguide as it will
-        automatically do any unpacking and transformations required
-        to constrain the samples to the support of the latent sites.
-
     **Reference:**
 
     1. *Automatic Differentiation Variational Inference*,
@@ -145,19 +134,12 @@ class AutoContinuous(AutoGuide):
                                                                    model_args=args,
                                                                    model_kwargs=kwargs)
         self._inv_transforms = {}
-        self._has_transformed_dist = False
         unconstrained_sites = {}
         for name, site in self.prototype_trace.items():
             if site['type'] == 'sample' and not site['is_observed']:
-                if site['intermediates']:
-                    transform = biject_to(site['fn'].base_dist.support)
-                    self._inv_transforms[name] = transform
-                    unconstrained_sites[name] = transform.inv(site['intermediates'][0][0])
-                    self._has_transformed_dist = True
-                else:
-                    transform = biject_to(site['fn'].support)
-                    self._inv_transforms[name] = transform
-                    unconstrained_sites[name] = transform.inv(site['value'])
+                transform = biject_to(site['fn'].support)
+                self._inv_transforms[name] = transform
+                unconstrained_sites[name] = transform.inv(site['value'])
 
         self._init_latent, unpack_latent = ravel_pytree(init_params)
         # this is to match the behavior of Pyro, where we can apply
@@ -165,7 +147,7 @@ class AutoContinuous(AutoGuide):
         self._unpack_latent = UnpackTransform(unpack_latent)
         self.latent_size = np.size(self._init_latent)
         if self.base_dist is None:
-            self.base_dist = dist.Independent(dist.Normal(np.zeros(self.latent_size), 1.), 1)
+            self.base_dist = dist.Normal(np.zeros(self.latent_size), 1.).to_event(1)
         if self.latent_size == 0:
             raise RuntimeError('{} found no latent variables; Use an empty guide instead'
                                .format(type(self).__name__))
@@ -203,10 +185,7 @@ class AutoContinuous(AutoGuide):
             site = self.prototype_trace[name]
             value = transform(unconstrained_value)
             log_density = - transform.log_abs_det_jacobian(unconstrained_value, value)
-            if site['intermediates']:
-                event_ndim = len(site['fn'].base_dist.event_shape)
-            else:
-                event_ndim = len(site['fn'].event_shape)
+            event_ndim = len(site['fn'].event_shape)
             log_density = sum_rightmost(log_density,
                                         np.ndim(log_density) - np.ndim(value) + event_ndim)
             delta_dist = dist.Delta(value, log_density=log_density, event_ndim=event_ndim)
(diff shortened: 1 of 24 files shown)


numpyro/contrib/autoguide/__init__.py
```diff
@@ -203,44 +185,16 @@ def __call__(self, *args, **kwargs):
...
-        if sample_shape:
-            latent_sample = np.reshape(latent_sample, (-1, np.shape(latent_sample)[-1]))
-            unpacked_samples = vmap(unpack_single_latent)(latent_sample)
-            unpacked_samples = tree_map(lambda x: np.reshape(x, sample_shape + np.shape(x)[1:]),
-                                        unpacked_samples)
-        else:
-            unpacked_samples = unpack_single_latent(latent_sample)
-        return unpacked_samples
+    def _unpack_and_constrain(self, latent_sample):
+        return transform_fn(self._inv_transforms, self._unpack_latent(latent_sample))
```
fehiepsi, line 197: We save a bunch of code here. :D

numpyro/contrib/autoguide/__init__.py
```diff
@@ -610,42 +551,3 @@ def _get_transform(self):
             arnn = numpyro.module('{}_arn__{}'.format(self.prefix, i), arn, (self.latent_size,))
             flows.append(BlockNeuralAutoregressiveTransform(arnn))
         return ComposeTransform(flows)
-
-
-class AutoContinuousELBO(ELBO):
```
fehiepsi, line 615: We no longer need this class. cc @lumip

numpyro/contrib/reparam.py
```diff
@@ -55,6 +55,14 @@ def process_message(self, msg):
         new_fn, value = reparam(msg["name"], msg["fn"], msg["value"])
 
         if value is not None:
+            if new_fn is None:
+                msg['type'] = 'deterministic'
+                msg['value'] = value
+                for key in list(msg.keys()):
+                    if key not in ('type', 'name', 'value'):
+                        del msg[key]
+                return
```
fehiepsi, line 64: Returns `deterministic` site here instead of `sample` site. This way, MCMC can report results for deterministic nodes.

numpyro/distributions/transforms.py
```diff
@@ -498,7 +498,7 @@ def __call__(self, x):
         batch_shape = x.shape[:-1]
         if batch_shape:
             unpacked = vmap(self.unpack_fn)(x.reshape((-1,) + x.shape[-1:]))
-            return tree_map(lambda x: np.reshape(x, batch_shape + x.shape[-1:]), unpacked)
+            return tree_map(lambda z: np.reshape(z, batch_shape + z.shape[1:]), unpacked)
```
fehiepsi, line 501: Fix a `reshape` bug (`unpacked`'s event shape starts at 1, rather than -1).

numpyro/handlers.py
```diff
@@ -259,25 +257,28 @@ class condition(Messenger):
        >>> assert exec_trace['a']['value'] == -1
        >>> assert exec_trace['a']['is_observed']
     """
-    def __init__(self, fn=None, param_map=None, substitute_fn=None):
-        self.substitute_fn = substitute_fn
+    def __init__(self, fn=None, param_map=None, condition_fn=None):
```
fehiepsi, line 260: I think this is typo. In the docs, this arg is `condition_fn`.

numpyro/infer/hmc_util.py
```diff
@@ -209,8 +209,8 @@ def update_fn(step_size, inverse_mass_matrix, state):
     return init_fn, update_fn
 
 
-def find_reasonable_step_size(potential_fn, kinetic_fn, momentum_generator, inverse_mass_matrix,
-                              position, rng_key, init_step_size):
+def find_reasonable_step_size(potential_fn, kinetic_fn, momentum_generator,
+                              init_step_size, inverse_mass_matrix, position, rng_key):
```
fehiepsi, line 213: I just reorder the position of `init_step_size` to clean up the code (using the utility `identity`).

numpyro/infer/initialization.py
```diff
@@ -0,0 +1,123 @@
+# Copyright Contributors to the Pyro project.
+# SPDX-License-Identifier: Apache-2.0
+
+from functools import partial
```
fehiepsi, line 4: Just copy 1-1 from `infer/util.py` file for clarity.

numpyro/contrib/reparam.py
```diff
@@ -55,6 +55,14 @@ def process_message(self, msg):
         new_fn, value = reparam(msg["name"], msg["fn"], msg["value"])
 
         if value is not None:
+            if new_fn is None:
+                msg['type'] = 'deterministic'
+                msg['value'] = value
+                for key in list(msg.keys()):
+                    if key not in ('type', 'name', 'value'):
```
neerajprad, line 62: Could you explain why we need to delete here?

numpyro/contrib/reparam.py
neerajprad: How do we handle this in Pyro?

numpyro/contrib/reparam.py
fehiepsi: We don't need to delete, but `deterministic` default msg doesn't hold those keys, so I delete for consistency.

numpyro/contrib/reparam.py
fehiepsi: In Pyro, we use Delta observation `sample` site because Pyro doesn't have `deterministic` site yet.

neerajprad: I see. So with this change, we'll need to wrap this in `Predictive` to get the values at deterministic nodes still, right?

fehiepsi: For both Delta obs or deterministic, we need to use Predictive. But with deterministic, we can retrace the model at postprocess_fn, as we currently do, so no need to use Predictive. I think we need to do the same for autoguide but I'll address it separately.

numpyro/contrib/autoguide/__init__.py
fehiepsi: @neerajprad I'll add a commit to record `deterministic` sites for autoguide. I think we did it previously here.

fehiepsi:
@neerajprad I have made a change at `AutoGuide._unpack_and_constrain` method to return deterministic sites for `sample_posteriors`, `median`, `quantiles` methods. This makes autoguide consistent with mcmc. WDYT?

neerajprad:
> @neerajprad I have made a change at AutoGuide._unpack_and_constrain method to return deterministic sites for sample_posteriors, median, quantiles methods. This makes autoguide consistent with mcmc. WDYT?

In principle, this is a fine change but I would like to discuss what we get by doing this and alternatives, if any. Some of the tests are now failing, btw. How about we make this change and discuss in a separate PR instead (just the autoguide related change; MCMC looks fine)?

fehiepsi:
Yes, let's discuss it when we move autoguide to infer or at #609 (the autoguide part).

neerajprad:
Sounds good. LGTM once the autoguide changes are reverted.

Issue #602, opened 2020-05-20
_NumpyroOptim causes double jit-compilation


lumip:
We've noted that our SVI-trained models are always jit-compiled two times (in the first and second epoch, not more afterwards). This is caused by the initialization of `_NumpyroOptim` in conjunction with [jax issue 3163]( https://github.com/google/jax/issues/3163).

Briefly, `_NumpyroOptim.init` returns a tuple, the first entry of which tracks the iterations the optimizer has seen, which is initialized to `0` as a native python integer. After the first iteration, this value, is instead, a jax type (a `jax.DeviceArray` of type `np.int32`) - this change in type causes jax to trace and compile the model (and guide) again.

Minimal example:
```python
import jax
import jax.numpy as np

from numpyro.optim import Adam

adam = Adam(1e-3)
state = adam.init(np.zeros(10))

my_fn_calls = 0

@jax.jit
def my_fn(state, g):
    global my_fn_calls
    my_fn_calls += 1

    state = adam.update(g, state)
    return state

rng = jax.random.PRNGKey(0)

for i in range(10):
    rng_i = jax.random.fold_in(rng, i)
    g = jax.random.normal(rng_i, (10,))
    state = my_fn(state, g)

assert(my_fn_calls == 1) # fails as my_fn_calls is 2
```

PR #603, opened 2020-05-20
_NumpyroOptim.init emits jax type for iter count

lumip:
fixes the double compilation issue described in #602


fehiepsi:
Awesome!! Thanks for adding the tests and for the tip of using global variable to detect re-compilation. ;)

Issue #598, opened 2020-05-19
0.3 release checklist


fehiepsi:
Here we list issues that we want to address for the next release:

- [x] Port `pyro.infer.reparam` #599
- [x] Remove all `base_param_map`/`transform_with_intermediates` stuffs to simplify our codebase. Remove `AutoContinuousELBO`. #601
- [x] Update `neutra`, `neal_funnel`, `eight_schools` examples to reflect the usage of `reparam`
- [x] Update the pattern to define improper prior #605
- [x] Move `autoguide` to the `infer` module. 
- [x] Refactor `initialization` strategies, following Pyro versions. Revise the pattern of subsitute_fn because it is used incorrectly, as mention at [this comment](https://github.com/pyro-ppl/numpyro/pull/595/files#diff-fee397191e0cc75a2e33eea8f7ecb6d5R335) #606
- [x] Unifies the pattern of returning statistics for deterministic sites
- [x] #585 Addresses compiling issues
- [x] Add some `control_flow` primitives #595
- ~[ ] #567 Enhance `kinetic_energy` API, based on recent refactoring in Pyro. This should allow users to set initial mass matrix #536~ **edit** It is not clear which API we should change... so I would like to defer this to future PRs when structure mass matrix is required.
- [ ] Initial support for models with discrete latent variable. We can defer advanced stuff such as `markov` for later releases.
- [x] Update readme eight school example using reparam.
- [x] Update notebooks
- [x] Update jax/jaxlib to the latest version:
    + Use the convention `import jax.numpy as jnp` #590
    + Use the upstream logsumexp again
    + Use `lax.switch` for multiple branches
    + Use the upstream poisson sampler
- [x] some tweaks
    + remove device put in poisson
    + match Pyro api for `condition`, `substitute`, `scale`, Pareto, MultivariateAffineTransform...
    + verify all the docs are generated as expected
    + PRNGKey log_prob issue
    + remove contrib/distributions

fehiepsi:
@neerajprad I want to remove the support for improper prior through `pyro.param`. The reasons are:
+ We can achieve the same purpose by using `sample(name, dist.Foo().mask(False))`. Using `sample` is better because we can draw samples from the (improper) prior to get initial values; rather than a fixed value from `param` statement.
+ The distinguish between `param` and `sample` is clearer.
+ Keeps the behavior like in Pyro, without losing functionality.

There are some other technical benefits:
[...]

PR #605, opened 2020-05-23
Switch to use `mask(False)` for improper prior

fehiepsi:
Per discussions at #598, it is better to use this pattern for improper prior.


test/test_mcmc.py
```diff
@@ -270,8 +275,8 @@ def test_mcmc_progbar():
     num_warmup, num_samples = 10, 10
 
     def model(data):
-        mean = numpyro.param('mean', 0.)
-        std = numpyro.param('std', 1., constraint=constraints.positive)
+        mean = numpyro.sample('mean', dist.Normal(0, 1).mask(False))
+        std = numpyro.sample('std', dist.LogNormal(0, 1).mask(False))
```
fehiepsi, line 279: @neerajprad Here is an equivalent improper prior, but with this, `init_to_prior` will work (rather than being constant at `init_value` of `numpyro.param`)

test/test_mcmc.py
```diff
@@ -127,7 +127,10 @@ def test_improper_normal():
 
     def model(data):
         alpha = numpyro.sample('alpha', dist.Uniform(0, 1))
-        loc = numpyro.param('loc', 0., constraint=constraints.interval(0., alpha))
+        with reparam(config={'loc': TransformReparam()}):
```
neerajprad, line 130: We need the `reparam` for the dynamic upper bound `alpha`, right?

Purely from an interface point of view, this seems like a regression. :) This is probably not a pattern that we would like to advertise. I still appreciate that we can clear off a lot of our tech debt, and if anyone wants to have optimizable params with dynamic support, there is a way to do this. This seems like a fringe enough use case that it is better to opt for code simplicity.

fehiepsi: You are right, this is a regression. However, the change simplifies much of code, it makes things less confusing (there are several issues come from the confusing base_param_map, substitute_fn only applies for base value,...). We also gain some performance with the change.

About this improper prior, there are two ways of thinking:
+ flat prior over (0, 1)
+ flat prior over (0, alpha)
Before this change, the behavior is flat prior over (0, 1). Is it the behavior that we want? I am not sure. With this change, it is clear that users want a flat prior over (0, 1). To get a flat prior over (0, alpha), I think we just need to add `handlers.mask(False)` context out of this reparam context. So things are more explicit to me. :D

neerajprad: I read this as flat prior over (0, alpha). I would imagine that something like the following should work. Is that not the case?

```python
with reparam(config={'loc': TransformReparam()}):
    loc = numpyro.sample('loc', dist.Uniform(0, alpha).mask(False))
```

fehiepsi: The issue is TransformReparam does not work with `MaskedDistribution` yet. So you would need to use `mask` handler here. I'll make an issue for it. After that, your code will work as you expected.

Issue #585, opened 2020-05-01
Optimize compiling time for MCMC


fehiepsi:
Now we start looking at large models. We need to remove unnecessary traveling the model to reduce compiling time.

#### Single-chain init

We travel 4 times: 1 to find `inv_transforms`, 1 to find init sampling state (which involves computing potential energy), 2 to find valid initial params (which involves computing both `inv_transforms` and potential energy).

- [x] We should only need to travel 2 times here: we can use the output of `find valid initial params` to define `inv_transform` and set the init sampling state (because `find_valid_initial_params` also computes pe and its grad)

#### Single-chain sample

- [x] After #584 and a small change to `get_potential_fn`, we travel 1 time for each sampling step to compute potential energy. This is expected!

#### Multi chain

- [ ] Currently, we do `pmap(init + sample)`. But it is better to do `pmap/map/vmap(init) + pmap(sample)` because `init + sample` produces much larger jaxpr than each single one and JAX compiling time is nonlinear w.r.t. jaxpr length.

#### heuristic step size

- [x] (pretty low priority) probably we can also find heuristic step size with only 1 time for compiling (if users want to use this functionality)

fehiepsi:
Currently, using `pmap(init + sample)` really simplifies a lot of code for us. So probably the last task is unnecessary unless we face some slow compiling issues again.

mcguenther:
> `pmap`

o pmap

PR #607, opened 2020-05-24
Add some profiling tests

fehiepsi:
Those tests are used to observe the enhancements while working on #585.

@lumip: thanks for giving a great idea to test the number of compiling times

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -34,8 +34,9 @@ setup(
     author_email='npradhan@uber.com',
     install_requires=[
         # TODO: pin to a specific version for the release (until JAX's API becomes stable)
-        'jax>=0.1.65',
-        'jaxlib>=0.1.45',
+        'jax==0.1.67',
+        # check min version here: https://github.com/google/jax/blob/master/jax/lib/__init__.py#L20
+        'jaxlib>=0.1.47',
         'tqdm',
     ],
     extras_require={
--- a/test/test_compile.py
+++ b/test/test_compile.py
@@ -1,0 +1,66 @@
+# Copyright Contributors to the Pyro project.
+# SPDX-License-Identifier: Apache-2.0
+
+import pytest
+
+from jax import lax, random
+from jax.lib import xla_bridge
+
+import numpyro
+from numpyro.contrib.autoguide import AutoDiagonalNormal
+import numpyro.distributions as dist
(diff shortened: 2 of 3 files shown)


test/test_compile.py
```diff
@@ -0,0 +1,66 @@
...
+def test_mcmc_one_chain(deterministic):
+    GLOBAL["count"] = 0
+    mcmc = MCMC(NUTS(model), 100, 100)
+    mcmc.run(random.PRNGKey(0), deterministic=deterministic)
+    mcmc.get_samples()
+
+    if deterministic:
+        assert GLOBAL["count"] == 8
+    else:
+        assert GLOBAL["count"] == 7
```
fehiepsi, line 36: @eb8680 @neerajprad this is pretty bad; with recent refactoring, in principle, we should just need to travel through the model 3 times:
+ 1 to inspect the model to get information such as prototype params, inverse transforms, has deterministic sites or not, using enumerate or not,...
+ 1 to find initial params
+ 1 to run mcmc loops

setup.py
```diff
@@ -34,8 +34,9 @@
     author_email='npradhan@uber.com',
     install_requires=[
         # TODO: pin to a specific version for the release (until JAX's API becomes stable)
-        'jax>=0.1.65',
-        'jaxlib>=0.1.45',
+        'jax==0.1.67',
```
fehiepsi, line 37: Tests are failing with `jax==0.1.68` due to a bug at `np.median` method.

test/test_compile.py
neerajprad: We are traversing through the model 7 times, but are we also calling into `jit` 7 times? The latter is what we are interested in optimizing.

neerajprad: FYI..not for testing, but for local debugging you can use the jax flag `jax.flags.FLAGS['log_jax_compiles'] = True` to log every time the model is being compiled.

fehiepsi: For single chain, the body_fn of find_valid_initial_params travels through the model 2 times. The sample_fn of hmc travels through the model 2 times, so the situation might be not that bad. But for multi-chains, I think we run pmap over about 6 or 7 traces, which is pretty bad.

neerajprad: Thanks for explaining, that makes sense.

PR #608, opened 2020-05-24
Update the forecasting with the usage of Predictive

fehiepsi:
Originally, we wrote the tutorial to introduce some handlers and the usage of `vmap` to forecast values. But for many cases, we don't need to learn about handlers to use NumPyro. So this PR adds a small section to illustrate that we can use Predictive utility to get the same result with just 2 lines of code.


neerajprad:
This looks good. If we have introduced effect handlers in other tutorials, I am fine with even removing the whole forecasting part and using `Predictive` by default if that keeps things simple.

fehiepsi:
Yeah, you are right. We already covered handlers in regression tutorial. Let me remove that part.

PR #609, opened 2020-05-25
Reduce compiling time in MCMC

fehiepsi:
We are seeing more and more sophisticated applications of NumPyro, so it is time to address some performance issues. Recent refactoring work makes it easier to address #585. The target is to reduce the number of times we travel through the model, which can be verified using tests in #607.

Note: we need to make sure that `jit_model_args` still works.

### Tasks
- [x] Refactoring utilities in `infer/util.py`
- [x] Refactoring HMC/MCMC logics.
- [x] Optimize AutoGuide
- ~[ ] Optimize multi-chain~ this is defered to a later PR
- [x] Make sure jit_model_args still work
- [x] Make sure all tests pass.


test/test_compile.py
```diff
@@ -0,0 +1,66 @@
...
+def test_mcmc_one_chain(deterministic):
+    GLOBAL["count"] = 0
+    mcmc = MCMC(NUTS(model), 100, 100)
+    mcmc.run(random.PRNGKey(0), deterministic=deterministic)
+    mcmc.get_samples()
+
+    if deterministic:
+        assert GLOBAL["count"] == 4
+    else:
+        assert GLOBAL["count"] == 3
```
fehiepsi, line 36: By some small changes, we are able to reduce the number of time we trace the model to 3 :dagger:

fehiepsi:
@neerajprad To reduce the number of compiling time, I made several changes:
+ `find_initial_valid_params` now returns initial params in addition to its potential energy and their gradients. This way, we save time at `verlocity_verlet.init` (no need to run `value_and_grad(potential_fn)(z)` there).
+ With the default `init_to_uniform` strategy, we don't need to trace the model to generate initial values (this applies for `init_to_feasible` and `init_to_value` too).
+ We don't need to trace `model` to get `inv_transforms` in `get_potential_fn`.
+ `get_model_transforms` now returns `inv_transforms` and `prototype_params`, we can use this prototype_params to setup initial values of while_loop in `find_initial_valid_params`.

This way, whenever we want to collect more information from `model` (e.g. enumerate,...), we just need to add a new key. This reduces the number of times we need to change the interface to add new functionality.

neerajprad:
Do we have a target model that is large enough that we can use to benchmark the changes in this PR?

fehiepsi:
I think we can use `hmm` with an option to switch `lax.scan` to a python scan. I'll do it. :)

fehiepsi:
@neerajprad I run tests for  `python examples/hmm.py --unroll-loop --num-unsupervised 100`. It turns out that this just saves a few seconds: 54s vs 58s. Haizz... I have thought that this makes a huge change. :(

neerajprad:
> @neerajprad I run tests for python examples/hmm.py --unroll-loop --num-unsupervised 100. It turns out that this just saves a few seconds: 54s vs 58s. Haizz... I have thought that reducing the number of times we trace the model makes a huge change. :(

No worries. Code simplification is always good, but if we are making a lot of minor optimizations that adds complexity, we can use this as a benchmark and see what makes sense to add. I haven't looked at the details btw, but wanted to have an example to test how do each of the changes affect the compilation / execution time.

neerajprad:
Btw..how about some of the examples that @eb8680 had in his PR? Do you think the `hmm` example is similar?

fehiepsi:
I think it is similar, but I'll check if we gain some benefits there.

numpyro/contrib/autoguide/__init__.py
```diff
@@ -185,8 +174,8 @@ def __call__(self, *args, **kwargs):
         result = {}
 
         for name, unconstrained_value in self._unpack_latent(latent).items():
-            transform = self._inv_transforms[name]
             site = self.prototype_trace[name]
+            transform = biject_to(site['fn'].support)
```
fehiepsi, line 178: We will use `postprocess_fn` to constrain latent samples, so we don't need to store `inv_transforms`. If we don't want to replay model when there are deterministic sites, then we can simply remove that behavior in then `dynamic_args=False` branch of `get_potential_fn` utility. But I think we should record `deterministic` sites.

numpyro/contrib/autoguide/__init__.py
```diff
@@ -434,12 +430,10 @@ def _get_transform(self, params):
         def loss_fn(z):
             params1 = params.copy()
             params1['{}_loc'.format(self.prefix)] = z
-            # we are doing maximum likelihood, so only require `num_particles=1` and an arbitrary rng_key.
-            return ELBO().loss(random.PRNGKey(0), params1, self.model, self,
-                               *self._args, **self._kwargs)
```
fehiepsi, line 439: We don't store `self._args, self._kwargs` so I move this `loss` function to `setup_prototype` method.

numpyro/infer/hmc_util.py
```diff
@@ -181,13 +181,16 @@ def velocity_verlet(potential_fn, kinetic_fn):
...
         """
         :param z: Position of the particle.
         :param r: Momentum of the particle.
+        :param potential_energy: Potential energy at `z`.
+        :param z_grad: gradient of potential energy at `z`.
         :return: initial state for the integrator.
         """
-        potential_energy, z_grad = value_and_grad(potential_fn)(z)
+        if potential_energy is None or z_grad is None:
+            potential_energy, z_grad = value_and_grad(potential_fn)(z)
```
fehiepsi, line 193: We can avoid evaluating potential_fn if `potential_energy` and `z_grad` is available.

numpyro/infer/mcmc.py
```diff
@@ -72,7 +72,8 @@ def _get_num_steps(step_size, trajectory_length):
     return num_steps.astype(canonicalize_dtype(np.int64))
 
 
-def _sample_momentum(unpack_fn, mass_matrix_sqrt, rng_key):
+def momentum_generator(prototype_r, mass_matrix_sqrt, rng_key):
```
fehiepsi, line 75: cleanup: I want to move the logic to pack/unpack into a single function here. This way, we don't need to store a global attribute `momentum_generator` in `hmc`.

neerajprad:
@fehiepsi - let me know whenever you'd like me to take a look.

numpyro/infer/mcmc.py
```diff
@@ -507,23 +517,10 @@ def init(self, rng_key, num_warmup, init_params=None, model_args=(), model_kwarg
...
-        # Find valid initial params
-        if self._model and not init_params:
-            init_params, is_valid = find_valid_initial_params(rng_key, self._model,
-                                                              init_strategy=self._init_strategy,
-                                                              model_args=model_args,
-                                                              model_kwargs=model_kwargs)
-            if not_jax_tracer(is_valid):
-                if device_get(~np.all(is_valid)):
-                    raise RuntimeError("Cannot find valid initial parameters. "
-                                       "Please check your model again.")
```
fehiepsi, line 526: I have moved those logics to `init_state` method. The only behavior change here is: `init_params` arg takes no effect when `self._model is not None`. If we want to preserve the old behavior (i.e. using `init_params` for `model` and skip `find_valid_inititial_params`, we can simply add an args `init_params` to `initialize_model` class).

numpyro/infer/mcmc.py
```diff
@@ -1084,7 +1077,9 @@ def _single_chain_mcmc(self, rng_key, init_state, init_params, args, kwargs, col
             init_state = self.sampler.init(rng_key, self.num_warmup, init_params,
                                            model_args=args, model_kwargs=kwargs)
         if self.postprocess_fn is None:
-            self.postprocess_fn = self.sampler.postprocess_fn(args, kwargs)
+            postprocess_fn = self.sampler.postprocess_fn(args, kwargs)
+        else:
+            postprocess_fn = self.postprocess_fn
```
fehiepsi, line 1082: I believe this is the intended behavior: if users provide postprocess_fn, we use it; otherwise, use `sampler.postprocess_fn` with "current" args, kwargs. I am not sure why this does not raise any issue previously.

numpyro/infer/util.py
```diff
@@ -108,14 +109,17 @@ def constrain_fn(model, transforms, model_args, model_kwargs, params, return_det
         sites from the model. Defaults to `False`.
     :return: `dict` of transformed params.
     """
-    params_constrained = transform_fn(transforms, params)
-    substituted_model = substitute(model, param_map=params_constrained)
+    def substitute_fn(site):
+        if site['name'] in params:
+            return biject_to(site['fn'].support)(params[site['name']])
```
fehiepsi, line 114: @neerajprad This is unconstrained reparam that I mentioned. This way, we don't need to store `inv_transforms`.

numpyro/infer/util.py
```diff
@@ -124,29 +128,43 @@ def potential_energy(model, inv_transforms, model_args, model_kwargs, params):
...
+        name = site['name']
+        if name in params:
+            p = params[name]
+            t = biject_to(site['fn'].support)
+            p_constrained = t(p)
+            log_det = np.sum(t.log_abs_det_jacobian(p, p_constrained))
+            if site['scale'] is not None:
+                log_det = site['scale'] * log_det
+            numpyro.factor('_{}_log_det'.format(name), log_det)
+            return p_constrained
```
fehiepsi, line 146: This is unconstrained reparam with jacobian adjustment.

numpyro/infer/util.py
```diff
@@ -168,84 +201,84 @@ def body_fn(state):
...
+            constrained_values, inv_transforms = {}, {}
+            for k, v in model_trace.items():
+                # TODO: filter out discrete sites here
+                if v['type'] == 'sample' and not v['is_observed']:
+                    constrained_values[k] = v['value']
+                    inv_transforms[k] = biject_to(v['fn'].support)
+            params = transform_fn(inv_transforms,
+                                  {k: v for k, v in constrained_values.items()},
+                                  invert=True)
+        else:  # this branch doesn't require tracing the model
```
fehiepsi, line 217: For `init_to_uniform`, `init_to_value`, we don't need to trace the model to get initial values. We handle that logic here.

numpyro/infer/util.py
```diff
@@ -168,84 +201,84 @@ def body_fn(state):
...
 
     # Handle possible vectorization
     if rng_key.ndim == 1:
-        init_params, is_valid = _find_valid_params(rng_key)
+        (init_params, pe, z_grad), is_valid = _find_valid_params(rng_key, exit_early=True)
     else:
-        init_params, is_valid = lax.map(_find_valid_params, rng_key)
-    return init_params, is_valid
+        (init_params, pe, z_grad), is_valid = lax.map(_find_valid_params, rng_key)
+    return (init_params, pe, z_grad), is_valid
```
fehiepsi, line 252: Returns also `pe` and `grad` to avoid recomputing those terms in `hmc`.

numpyro/infer/util.py
```diff
@@ -303,30 +333,43 @@ def initialize_model(rng_key, model,
...
-    init_params, is_valid = find_valid_initial_params(rng_key, model,
-                                                      init_strategy=init_strategy,
-                                                      model_args=model_args,
-                                                      model_kwargs=model_kwargs)
+    init_strategy = init_strategy if isinstance(init_strategy, partial) else init_strategy()
+    if init_strategy.func is init_to_value:
+        init_values = init_strategy.keywords.get("values")
+        unconstrained_values = transform_fn(inv_transforms, init_values, invert=True)
+        init_strategy = _init_to_unconstrained_value(values=unconstrained_values)
+    prototype_params = transform_fn(inv_transforms, constrained_values, invert=True)
```
fehiepsi, line 362: If the init strategy is `init_to_value`, we transform it into the unconstrained domain.

fehiepsi:
@neerajprad Thanks! This is ready to review now. Please let me know if you have any concern. :)

neerajprad:
Thanks for adding your comments and guiding me, very helpful. I'll take a look at this today. Could you provide a guide for review? e.g.
 - what changes are simply code refactoring / simplification.
 - what changes result in reduction in compilation time.

fehiepsi:
@neerajprad I have updated the comments at the places we save tracing time. Basically, we save 4 traces using the following tricks:
+ using unconstrain reparam in `constrain_fn` (no need to find inv_transforms for `postprocess_fn`)
+ using unconstrain reparam in `potential_energy` (no need to find inv_transforms for `potential_fn`)
+ using prototype_params in `find_valid_initial_params` for `init_to_uniform/init_to_feasible/init_to_value` strategies (this does not take effect for other strategies)
+ keep `potential_energy`, `grad` terms in `find_intial_valid_params` and provide it to `hmc_init`. This way, we don't need to recompute those terms to get an initial velocity verlet state.

All other changes are to reorganize and clean up.

numpyro/infer/mcmc.py
fehiepsi: The issue is when using args, kwargs at later runs, when args, kwargs change. Before this pr, I think we use the same args, kwargs for different runs here.

numpyro/infer/util.py
```diff
@@ -39,7 +42,7 @@ def log_density(model, model_args, model_kwargs, params):
     """
     model = substitute(model, param_map=params)
     model_trace = trace(model).get_trace(*model_args, **model_kwargs)
-    log_joint = jax.device_put(0.)
+    log_joint = np.array(0.)
```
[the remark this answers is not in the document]
fehiepsi: Both gives us the same device array. This is not important though.

examples/stochastic_volatility.py
```diff
@@ -85,7 +85,7 @@ def main(args):
     _, fetch = load_dataset(SP500, shuffle=False)
     dates, returns = fetch()
     init_rng_key, sample_rng_key = random.split(random.PRNGKey(args.rng_seed))
-    init_params, potential_fn, constrain_fn = initialize_model(init_rng_key, model, model_args=(returns,))
+    init_params, potential_fn, constrain_fn, _ = initialize_model(init_rng_key, model, model_args=(returns,))
```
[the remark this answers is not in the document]
fehiepsi: Haha, I'll revert this back. I think the best option is to return a dict here so we can get more information from the model if needed. But you are right that this is unnecessary change.

examples/hmm.py
```diff
@@ -193,6 +198,7 @@ def main(args):
     parser.add_argument('-n', '--num-samples', nargs='?', default=1000, type=int)
     parser.add_argument('--num-warmup', nargs='?', default=500, type=int)
     parser.add_argument("--num-chains", nargs='?', default=1, type=int)
+    parser.add_argument("--unroll-loop", action='store_true')
```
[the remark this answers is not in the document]
fehiepsi: That makes me sad :( I thought there should be much improvement with the change. But at least, we are in much better position than the last release (thanks to some other changes: disable finding heuristic params or avoild compiling two times the base tree when iteratively create a tree).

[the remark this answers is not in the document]
neerajprad: That's good news though. :) It shows that for the most part, we shouldn't have to do a lot of micro-optimizations like we had to do in Pyro. But 5s is still quite significant 45s --> 40s.

examples/stochastic_volatility.py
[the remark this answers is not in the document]
fehiepsi: Hmm, this is used to initialize autoguide too (to save one time we trace the model). Historically, we expose some utilities in `infer` such as `transform_fn`, `constrain_fn`, `find_valid_params`, because they are used at various places and some are used in neutra example. For now, I believe only `initialize_model` is useful and can be used for many inference types: autoguide, kernel (hmc, nuts, sa)... I think it is a good time to decide which is better here in the long term. Do you have a suggestion? (I just want to find a better solution - keeping the old interface is fine to me)

fehiepsi:
Working with JAX, I think the main issue is: under jit context, it is better to make the function as small as possible; otherwise, compiling time will negatively affect the total inference time. I do agree that among those optimizations only the following (which belongs to `sample_fn`) are useful for single chain:
+ avoid finding `inv_transforms` in `potential_fn`
+ disable `find_reasonable_step_size` by default
+ setup a prototype tree to iteratively build a tree
because we only compile `sample_fn`. That's why we've seen a significant improvement using the last two optimizations and some improvement in this PR with the first optimization.

However, for multi-chains, we do everything under `pmap`, which will compile everything from the beginning to the end. This is the situation that makes other optimizations worth. E.g. here is what I observed with HMM example

| HMM (`--num-chains 2`)  | Time |
| ------------- | ------------- |
| Master branch  | 37s  |
| This PR | 29s  |
| This PR (without ParamInfo) | 33s |


| HMM (`--num-chains 2 --unroll-loop --num-unsupervised 50`)  | Time |
| ------------- | ------------- |
| Master branch  | 76s  |
| This PR | 46s  |
| This PR (without ParamInfo) | 59s |

examples/stochastic_volatility.py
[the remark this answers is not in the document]
neerajprad: I'm fine with changing the return type, but if we are doing that, maybe let us think about what is the minimal, sufficient information that we can return from `initialize_model`, maybe anticipating future requirements. I think a model trace seems like good idea, anything else that we should think about? Do you think we should package this in a `namedtuple` called `models_info` so that it is easier to tack on additional fields in the future?

neerajprad:
> under jit context, it is better to make the function as small as possible; otherwise, compiling time will negatively affect the total inference time.

As you pointed out earlier, I wonder if we'll be forced to revisit this choice of using `pmap` for everything vs. only using it to run chains post initialization, since this will be hard to keep track of as the code evolves (we'll specifically need to do perf regression testing for the multi chain case). Do you still plan to introduce additional optimizations for pmap, or are all planned changes in this PR?

examples/stochastic_volatility.py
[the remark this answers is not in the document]
fehiepsi: I think that returning a namedtuple is a good solution and can be extended in the future without having to worry about backward compatibility. I don't have an idea right now what fields else that we need. I'll add a TODO note here so that we can revisit this issue when we truly need to extend the return results.

fehiepsi:
To further optimize multi chains, we need to change the logic of MCMC a bit. I don't see that we need to change the other utilities as we did in this PR. It is a bit low priority for me (actually, the job might be more involved than my expectation because we need to consider other chain methods such as `vectorized` or `sequential` too). For parallel method (pmap), what we currently do in the master branch
+ pmap(intialize_model + hmc_init + hmc_sample)

What we have in this PR is
+ pmap(initialize_model + hmc_sample)

If we split things out, we get:
+ pmap(initialize_model) + pmap(hmc_init) + pmap(hmc_sample)

or if we use ParamInfo:
+ pmap(initialize_model) + pmap(hmc_sample)

So eliminating pmap(hmc_init) will be a plus point for both splitting and non-splitting. About splitting (which I intended to do in a later PR, but low priority and probably not for the next release) vs non-splitting, do you think that it is worth addressing (given that it will make no harm to single-chain)? It is more like an issue of JAX compiling mechanism, rather than our implementation.

examples/stochastic_volatility.py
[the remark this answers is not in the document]
neerajprad: If we are using a namedtuple, it doesn't matter as much. It should be easy to add other fields (though I think mostly anything else can be derived from the trace).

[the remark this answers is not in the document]
fehiepsi: Just to make sure we are on the same page: I intended to revert the changes and add a TODO here (to return a namedtuple when we need new fields) to preserve backward compatibility for this unnecessary change (I can understand your concern). But I'm happy to make a change to use namedtuple now if you agree (which seems to be so based on your last comment).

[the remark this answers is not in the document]
neerajprad: I think we have sufficient evidence that there is other auxiliary information that will be useful, so I would be in favor of using a namedtuple here (which is a more stable interface anyways). WDYT?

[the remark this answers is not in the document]
fehiepsi: Absolutely agreed! Thanks for your suggestions, I'll make the change.

neerajprad:
Very happy to see how this shaped up! I'm wondering if it makes sense to expose all our utility functions like `get_potential_fn` since some of these utilities are more coupled now due to the optimizations in this PR. I still think those utilities have wider utility, but if you think some of these should be reserved for internal use only, we can make changes in subsequent PRs.

fehiepsi:
Yes, they might be useful for users so let's keep those utilities exposed as-is. Thanks for reviewing, @neerajprad !

PR #611, opened 2020-05-30
note on profiling mcmc runs

vanAmsterdam:


numpyro/infer/mcmc.py
```diff
@@ -1180,6 +1180,11 @@ def run(self, rng_key, *args, extra_fields=(), init_params=None, **kwargs):
             with the input type to `potential_fn`.
         :param kwargs: Keyword arguments to be provided to the :meth:`numpyro.infer.mcmc.MCMCKernel.init`
             method. These are typically the keyword arguments needed by the `model`.
+
+        .. note:: jax allows python code to continue even when the compiled code has not finished yet.
+        This can cause troubles when trying to profile the code for speed.
```
fehiepsi, line 1185: Thanks, @vanAmsterdam! Could you verify that `make docs` will generate an expected result? I think you would need to indent this line and the following lines.

PR #612, opened 2020-06-02
Add Improper Distribution

fehiepsi:
This mimics https://github.com/pyro-ppl/pyro/pull/2516 but adds an invalid `sample` method to generate random samples in the `support`. I also add `batch_shape=()` by default to make the usage of this distribution less verbose (there are some distributions in NumPyro also have the argument `batch_shape=()` by default).

@fritzo I should ask this in your PR: should this distribution have `validate_args` argument?

cc @vanAmsterdam: hopes that this will unblock your work using `master` branch

### DIFF
--- a/numpyro/distributions/constraints.py
+++ b/numpyro/distributions/constraints.py
@@ -54,8 +54,21 @@ import jax.numpy as np
 
 
 class Constraint(object):
+    """
+    Abstract base class for constraints.
+
+    A constraint object represents a region over which a variable is valid,
+    e.g. within which a variable can be optimized.
+    """
     def __call__(self, x):
         raise NotImplementedError
+
+    def check(self, value):
+        """
+        Returns a byte tensor of `sample_shape + batch_shape` indicating
+        whether each event in value satisfies this constraint.
+        """
+        return self(value)
 
 
 class _Boolean(Constraint):
--- a/numpyro/distributions/distribution.py
+++ b/numpyro/distributions/distribution.py
@@ -30,10 +30,10 @@ from contextlib import contextmanager
 import warnings
 
 import jax.numpy as np
-from jax import lax, random
+from jax import lax
 
-from numpyro.distributions.constraints import is_dependent, real
-from numpyro.distributions.transforms import Transform, biject_to
+from numpyro.distributions.constraints import Constraint, is_dependent, real  # noqa: F401
+from numpyro.distributions.transforms import Transform
 from numpyro.distributions.util import lazy_property, sum_rightmost, validate_sample
 from numpyro.util import not_jax_tracer
 
@@ -352,28 +352,49 @@ class ExpandedDistribution(Distribution):
 
 class ImproperUniform(Distribution):
     """
-    A helper distribution with zero :meth:`log_prob` and improper
-    :meth:`sample`.
+    A helper distribution with zero :meth:`log_prob` over the `support` domain.
 
-    .. note:: `sample` method does not generate samples distributed uniformly over the support.
-        Instead, it only returns valid values belong to the support.
+    .. note:: `sample` method is not implemented for this distribution.
+
+    **Usage**::
+
+        >>> from numpyro.distributions import constraints
+        >>>
+        >>> def model():
+        ...     # ordered vector with length 10
+        ...     x = sample('x', ImproperUniform(constraints.ordered_vector, (), event_shape=(10,))
+        ...
+        ...     # real matrix with shape (3, 4)
+        ...     y = sample('y', ImproperUniform(constraints.real, (), event_shape=(3, 4))
+        ...
+        ...     # a shape-(6, 8) batch of length-5 vectors greater than 3
+        ...     z = sample('z', ImproperUniform(constraints.greater_than(3), (6, 8), event_shape=(5,))
+
+    If you want to set improper prior over all values greater than `a`, where `a` is
+    another random variable, you might use
+
+        >>> x = sample('x', ImproperUniform(constraints.greater_than(a), (), event_shape=()))
+
+    or if you want to reparameterize it
+
+        >>> from numpyro.distributions import constraints, transforms
+        >>> from numpyro.contrib.reparam import reparam, TransformReparam
+        >>>
+        >>> with reparam(config={'x': TransformReparam()}):
+        ...     x = sample('x',
+        ...                TransformedDistribution(ImproperUniform(constraints.positive, (), ()),
+        ...                                        transforms.AffineTransform(a, 1)))
+
+    :param Constraint support: the support of this distribution.
+    :param tuple batch_shape: batch shape of this distribution. It is usually safe to
+        set `batch_shape=()`.
+    :param tuple event_shape: event shape of this distribution.
     """
     arg_constraints = {}
-    unconstrained_radius = 2
 
-    def __init__(self, support, event_shape, batch_shape=(), validate_args=None):
+    def __init__(self, support, batch_shape, event_shape, validate_args=None):
         self.support = support
         super().__init__(batch_shape, event_shape, validate_args=validate_args)
-
-    def sample(self, key, sample_shape=()):
-        transform = biject_to(self.support)
-        prototype_value = np.zeros(self.event_shape)
-        unconstrained_event_shape = np.shape(transform.inv(prototype_value))
-        shape = sample_shape + self.batch_shape + unconstrained_event_shape
-        unconstrained_samples = random.uniform(key, shape,
-                                               minval=-self.unconstrained_radius,
-                                               maxval=self.unconstrained_radius)
-        return transform(unconstrained_samples)
 
     @validate_sample
     def log_prob(self, value):
--- a/numpyro/distributions/distribution.py
+++ b/numpyro/distributions/distribution.py
@@ -350,7 +350,7 @@ class ExpandedDistribution(Distribution):
         return np.broadcast_to(self.base_dist.variance, self.batch_shape + self.event_shape)
 
 
-class Improper(Distribution):
+class ImproperUniform(Distribution):
     """
     A helper distribution with zero :meth:`log_prob` and improper
     :meth:`sample`.
@@ -361,9 +361,9 @@ class Improper(Distribution):
     arg_constraints = {}
     unconstrained_radius = 2
 
-    def __init__(self, support, event_shape, batch_shape=()):
+    def __init__(self, support, event_shape, batch_shape=(), validate_args=None):
         self.support = support
-        super().__init__(batch_shape, event_shape)
+        super().__init__(batch_shape, event_shape, validate_args=validate_args)
 
     def sample(self, key, sample_shape=()):
         transform = biject_to(self.support)
@@ -375,10 +375,18 @@ class Improper(Distribution):
                                                maxval=self.unconstrained_radius)
         return transform(unconstrained_samples)
 
+    @validate_sample
     def log_prob(self, value):
-        batch_shape = value.shape[:len(np.shape(value)) - len(self.event_shape)]
+        batch_shape = np.shape(value)[:np.ndim(value) - len(self.event_shape)]
         batch_shape = lax.broadcast_shapes(batch_shape, self.batch_shape)
         return np.zeros(batch_shape)
+
+    def _validate_sample(self, value):
+        mask = super(ImproperUniform, self)._validate_sample(value)
+        batch_dim = np.ndim(value) - len(self.event_shape)
+        if batch_dim < np.ndim(mask):
+            mask = np.all(np.reshape(mask, np.shape(mask)[:batch_dim] + (-1,)), -1)
+        return mask
 
 
 class Independent(Distribution):
--- a/numpyro/distributions/distribution.py
+++ b/numpyro/distributions/distribution.py
@@ -354,36 +354,45 @@ class ImproperUniform(Distribution):
     """
     A helper distribution with zero :meth:`log_prob` over the `support` domain.
 
-    .. note:: `sample` method is not implemented for this distribution.
+    .. note:: `sample` method is not implemented for this distribution. In autoguide and mcmc,
+        initial parameters for improper sites are derived `init_to_uniform` or `init_to_value`
+        strategies.
 
-    **Usage**::
+    **Usage:**
 
-        >>> from numpyro.distributions import constraints
-        >>>
-        >>> def model():
-        ...     # ordered vector with length 10
-        ...     x = sample('x', ImproperUniform(constraints.ordered_vector, (), event_shape=(10,))
-        ...
-        ...     # real matrix with shape (3, 4)
-        ...     y = sample('y', ImproperUniform(constraints.real, (), event_shape=(3, 4))
-        ...
-        ...     # a shape-(6, 8) batch of length-5 vectors greater than 3
-        ...     z = sample('z', ImproperUniform(constraints.greater_than(3), (6, 8), event_shape=(5,))
+    .. doctest::
+
+       >>> from numpyro import sample
+       >>> from numpyro.distributions import ImproperUniform, Normal, constraints
+       >>>
+       >>> def model():
+       ...     # ordered vector with length 10
+       ...     x = sample('x', ImproperUniform(constraints.ordered_vector, (), event_shape=(10,)))
+       ...
+       ...     # real matrix with shape (3, 4)
+       ...     y = sample('y', ImproperUniform(constraints.real, (), event_shape=(3, 4)))
+       ...
+       ...     # a shape-(6, 8) batch of length-5 vectors greater than 3
+       ...     z = sample('z', ImproperUniform(constraints.greater_than(3), (6, 8), event_shape=(5,)))
 
     If you want to set improper prior over all values greater than `a`, where `a` is
     another random variable, you might use
 
-        >>> x = sample('x', ImproperUniform(constraints.greater_than(a), (), event_shape=()))
+       >>> def model():
+       ...     a = sample('a', Normal(0, 1))
+       ...     x = sample('x', ImproperUniform(constraints.greater_than(a), (), event_shape=()))
 
     or if you want to reparameterize it
 
-        >>> from numpyro.distributions import constraints, transforms
-        >>> from numpyro.contrib.reparam import reparam, TransformReparam
-        >>>
-        >>> with reparam(config={'x': TransformReparam()}):
-        ...     x = sample('x',
-        ...                TransformedDistribution(ImproperUniform(constraints.positive, (), ()),
-        ...                                        transforms.AffineTransform(a, 1)))
+       >>> from numpyro.distributions import TransformedDistribution, transforms
+       >>> from numpyro.contrib.reparam import reparam, TransformReparam
+       >>>
+       >>> def model():
+       ...     a = sample('a', Normal(0, 1))
+       ...     with reparam(config={'x': TransformReparam()}):
+       ...         x = sample('x',
+       ...                    TransformedDistribution(ImproperUniform(constraints.positive, (), ()),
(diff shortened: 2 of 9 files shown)


numpyro/distributions/distribution.py
```diff
@@ -350,6 +350,45 @@ def variance(self):
...
+    A helper distribution with zero :meth:`log_prob` and improper
+    :meth:`sample`.
+
+    .. note:: `sample` method does not generate samples distributed uniformly over the support.
+        Instead, it only returns valid values belong to the support.
+    """
+    arg_constraints = {}
+    unconstrained_radius = 2
+
+    def __init__(self, support, event_shape, batch_shape=(), validate_args=None):
```
fritzo, line 364: Can we make this agree with Pyro? Either `batch_shape, event_shape` or `batch_shape=(), event_shape=()`? If you really feel strongly that `event_shape, batch_shape=()` is optimal, we should change the Pyro version.

:+1: I agree a `validate_args` is a good idea.

numpyro/distributions/distribution.py
```diff
@@ -350,6 +350,45 @@ def variance(self):
         return np.broadcast_to(self.base_dist.variance, self.batch_shape + self.event_shape)
 
 
+class ImproperUniform(Distribution):
+    """
+    A helper distribution with zero :meth:`log_prob` and improper
```
fritzo, line 355: nit: "improper sample" -> "arbitrarily distributed sample" or something. The term 'improper' specifically means "non normalized and possibly non-nonrmalizable" as in "improper uniform prior over the real line".

numpyro/distributions/distribution.py
fehiepsi: How about using `shape, event_dim=None` here where `event_dim=len(shape)` by default? I like it better than the current resolution. Either having `event_shape=()` or `event_dim=0` will lead to wrong results, while having `batch_shape=()` or `event_dim=len(shape)` by default will work for most usage cases (except for models with strong batch shape restriction, e.g. models with enumerate). If we go for this, I think we need to resolve the difference of `event_dim` in Pyro and `event_ndim` in NumPyro for Delta distribution too.

Using `batch_shape, event_shape` is good to me. But I worry that we need to explain those arguments properly. In addition, the previous usage of improper distributions do not require this distinction of shape:
```
param('x', np.ones(3), constraints.positive)
```
or
```
sample('x', dist.LogNormal(zeros(3), 1).mask(False), constraints.positive)
```

I also think @neerajprad will have more inputs about the interface choice. :)

fritzo: Yeah, I like `shape, event_dim=None`. My main use case is to replicate an existing distribution, so your suggestion would allow
```py
dist.ImproperUniform(fn.support, fn.shape(), fn.event_dim)
```
Does that also work in NumPyro?

fehiepsi: We don't have `.shape()` method and `event_dim` property. The main reason is at the early stage, we ported torch Distribution to NumPyro, rather than Pyro Distribution, so there are some missing pieces of stuff here in NumPyro. We'll add them. Thanks for pointing it out!

neerajprad: Either looks fine to me, but I don't fully understand why the default of treating all dims as event dims is a better choice than `batch_shape=(), event_shape=()`. Could you elaborate? The latter has the advantage that we won't have to specify either batch or event shape for a sample site under `plate`.

fehiepsi: @neerajprad The main reason is for many constraints such as `low_cholesky` or `simplex`, event_shape will not be `()`. If we go with `batch_shape=(), event_shape=()`, we should add some warnings for users (e.g. by checking `biject_to(support).event_dim < len(event_shape)`). Do you have a solution for this issue?

vanAmsterdam:
> This mimics [pyro-ppl/pyro#2516](https://github.com/pyro-ppl/pyro/pull/2516) but adds an invalid `sample` method to generate random samples in the `support`. I also add `batch_shape=()` by default to make the usage of this distribution less verbose (there are some distributions in NumPyro also have the argument `batch_shape=()` by default).
> 
> @fritzo I should ask this in your PR: should this distribution have `validate_args` argument?
> 
> cc @vanAmsterdam: hopes that this will unblock your work using `master` branch

@fehiepsi yes, this (combined) with a `sample(..., obs=...)` statement gives back the same flexibility as before; I made a tiny educational example for myself: https://gist.github.com/vanAmsterdam/c76164de2f39cc515dbf81b27ffa4b75

numpyro/distributions/distribution.py
neerajprad: I see. I suppose the most common use case I thought was either the positive or interval support, for which I think we'll need to pass in special arguments under `plate` which can be confusing. Correct me if I'm mistaken? I agree that with other constraints like simplex the default will not work, and in that case we can raise a warning if the `event_shape` is different from `constraint.event_dim`, but it seems to me that we should cater the defaults to the more common use case. WDYT?

fehiepsi: I don't worry about batch_shape, which can be set to `()` by default, and using `plate` will work. But I think the default `event_shape` should depend on the constraint. So to me, users should specify it or we add an early warning. I don't see why `positive` or `interval` is more popular (maybe they are useful for bayesian imputation?). To my knowledge, using improper distribution is popular for special constraints, e.g. in ordered regression (because there is no handy distribution over `ordered_vector` support - though users can use something like `dist.TransformedDistribution(dist.Normal(0, 1), OrderedTransform()).mask(False)` - FYI, the reason I supported improper distribution using `param` statement is to deal with `ordered_vector` constraint).

numpyro/distributions/distribution.py
fehiepsi: Thanks for explaining! I'll make the change.

Re `sample` method: I will raise not ImplementationError here as you did. It is better.

numpyro/distributions/distribution.py
neerajprad: Interesting use case, I had the simpler improper uniform in mind. 

> But I think the default event_shape should depend on the constraint. So to me, users should specify it or we add an early warning.

That seems fine - can we raise a warning if the user hasn't specified an event shape and the constraint has a non-trivial event dim? I still think treating all the dims as event dims by default could be problematic under a plate handler (will need to check), but it might be working right now since we aren't using `ExpandedDistribution` in `plate`.

fehiepsi: > can we raise a warning if the user hasn't specified an event shape and the constraint has a non-trivial event dim

@fritzo WDYT? We can keep the same interface but having a warning in NumPyro. You can drop this check in Pyro if you don't like this solution.

fritzo: I agree with Neeraj that, if we prioritize broadcasting support, then a default of `event_dim=()` is more useful. Indeed I would prefer a different interface for `pyro.deterministic()` because (e.g. again just minutes ago) it causes problems with batching to assume `event_dim=value.dim()`. Maybe you don't care so much about broadcasting in NumPyro, but in Pyro broadcasting is crucial for speed, hence `event_dim=()` seems like the right default there.

> raise a warning if the user hasn't specified an event shape and the constraint has a non-trivial event dim

Sounds reasonable if `is_validation_enabled()`, otherwise there is not yet a cheap way to query a constraint's `.event_dim`.

fehiepsi: So let me summarize the situations. Please correct me if I misunderstand something:

+ Let's first forget about the speed gain when we avoid using the `.expand` method under the `plate` context.

If users need to specify something related to `shape`, then it should be `event_shape` (because under plate, `.expand` is automatically retrieved).

Another point is to make `event_shape` or `shape` optional, set them to () by default, and raise a warning or let users debug things by themselves. For this point, I choose to raise a warning because the common usage of improper distribution is vector or matrix constraints, not scalar constraints.

+ Now, let's concern the speed gain when users specify `batch_shape` and want to avoid calling the `.expand` method under the hood.

Because users already specify `batch_shape`, we want to make `event_shape` optional to have shorter code? So if we specify something like `shape`, a better option is to make `even_dim=0` by default. But this comes with the trade-off of warning/debugging.

My opinions about `speed`, `short`, and `usage` are:
+ *speed*: most users don't know that specifying `batch_shape` will gain speed, so I wonder if that is a good point to deliver to users. If they know about it, then they can specify both `batch_shape`, `event_shape` correctly without difficulty.
+ *short*: this comes with the trade-off of warning/debugging.
+ *usage*: speed and short work best for scalar constraints, which is not common for improper constraint in my opinion.

My best option is still requiring `event_shape` or `shape` with `event_ndim=len(shape)` by default. The cost of using `expand` operator is acceptable to me. If we let `event_shape=()` by default, then we should add a warning.

fritzo: > the common usage of improper distribution is vector or matrix constraints

It sounds like your use case is different from `pyro.contrib.epidemiology` where all four uses cases are scalar. My use cases all concern automatically converting a model from causal dag form to factor graph form, which is especially useful in autovectorization.

I am happy with the current Pyro interface of
```py
ImproperUniform(support, batch_shape, event_shape)
```
One issue relying on `.expand()`
```
ImproperUniform(support, event_shape).expand(batch_shape)
```
is that the support itself could be batched, so the temporary object would be inconsistent.

Would it be a reasonable compromise to simply avoid default values, as we have done in Pyro?

neerajprad: My only concern has been that if we provide this, users may use this to implement improper priors like they do in Stan by declaring a parameter like `real<lower=0> sigma`. I suppose another way to do it would be using `LogNormal(0., 1.).mask(False)`, but this seems like a natural way to support the same and something that I would expect a new user to reach out for just going by the distribution's name, instead of using masking logic.

```
with numpyro.plate('N', 10):
    sigma = numpyro.sample('sigma', ImproperUniform(constraints.positive))  # instead of the earlier param
    ...
```

I don't think the code above will work as expected unless we specify that `event_shape=()`, which seems a bit odd. On the other hand, I think it is okay to expect users to provide `event_shape` when using `ImproperUniform` with a `LowerCholesky` constraint, for instance. This is just based on my rough estimate of how many times I have seen Stan code with a simple interval or positive/real constraint vs more complex constraints when specifying an improper prior. That said, this is a minor issue and I could be missing some other considerations as I haven't followed all the discussions closely (I don't think speed matters here).

neerajprad: @fehiepsi - I think your earlier suggestion of using `shape=(), event_dims=None` should work fine for the simple example above since the event shape will be determined at `__init__`. Sorry that was my mistake in failing to recognize that earlier. So either that or using Pyro's default behavior looks good to me. 😄

fehiepsi: I think Stan code 'real' means scalar here, i.e. event_shape=(). If it is a vector constraint, I believe Stan code will need something like vector(n), which means event_shape=(n,). So I dont think we introduce anything new comparing Stan code. But if we require batch shape, as in my earlier comment, we need to explain it properly, need to point to Pyro tensor shape tutorial, and mention that users can set it to (), which will work for most usage cases. I think it is fine given that this is a special distribution. But let's hold this for a day to see if there is a better solution.

> the support itself could be batched,

I think if it is positive with batch 10, we just need to specify event_shape=10. But I think that I miss your point here.

fritzo: Here is an example where `ImproperUniform(...).expand()` would creates an inconsistent intermediate distribution:
```py
population = torch.tensor([1000., 2000., 3000.])
with pyro.plate("region", 3):
    incidence = pyro.sample("incidence",
                            ImproperUniform(support=interval(0, population),
                                            batch_shape=(3,),
                                            event_shape=())) 
```

fehiepsi: Interesting... I thought we need to specify event_shape=population.shape here. Let me look at it more closely, probably your notion of event_shape is different from mine.

fehiepsi: After going through compartmental models and a long thought for this, I still do not understand why it's better to use args `batch_shape, event_shape` and to introduce that interface to users. Obviously that it does not make the internal code or user code shorter. Actually, we started with a discussion about the interface and we agreed that the solutions are in principle equivalent there. Then we raised some issues with using default values: `event_shape=()` or `event_shape=None`. The former requires a warning, while the latter might introduce some bugs or performance issues. Because I don't understand those latter issues, I'll go with Fritz's solution and try to deliver some messages to users through docs (current docs lack those pieces of information).

It isn't a compromise between solutions. It is just because I don't understand the reasoning behind the issues of the better (to me) solution and because I think we have put a lot of thoughts about this issue (which might not worth our time).

Thanks for discussions, @fritzo @neerajprad! I'll go with the current Pyro solution. I tried to simplify pieces of stuff for users, but I might be wrong that it will be actually helpful. :)

neerajprad: These discussions were very helpful at least for me to understand the issues involved, and I still don't think I fully do. 

@fehiepsi - My suggestion would be to have tests for the cases discussed here (specially cases with `plate` and batched constraints, and constraints with event shape), and verify that the sample and log_prob shapes agree with what we would expect (using whatever parametrization makes the most sense to you). In @fritzo's example above, for instance, I think even with `event_shape=(3,)`, we'll get incorrectly shaped samples with the existing implementation, but you probably have thought about it much more and have a solution. That will also make it easier to clearly see and discuss different ways to parametrize this, and how the resulting model code looks like.

fehiepsi: @neerajprad I don't have a solution. I just don't see the issue with it. I'll add more tests to understand better.

numpyro/distributions/distribution.py
```diff
@@ -350,6 +350,45 @@ def variance(self):
...
+
+    def sample(self, key, sample_shape=()):
+        transform = biject_to(self.support)
+        prototype_value = np.zeros(self.event_shape)
+        unconstrained_event_shape = np.shape(transform.inv(prototype_value))
+        shape = sample_shape + self.batch_shape + unconstrained_event_shape
+        unconstrained_samples = random.uniform(key, shape,
+                                               minval=-self.unconstrained_radius,
+                                               maxval=self.unconstrained_radius)
+        return transform(unconstrained_samples)
```
fritzo, line 376: Do you plan to replace this with `NotImplementedError` and require uses to use `InitMessenger` or similar (which is more flexible)?

fehiepsi: I plan to do so. I need to change something in how to interpret shape of initial unconstrained parameters (which works for improper distribution but might not work for general priors).

numpyro/distributions/constraints.py
```diff
@@ -54,9 +54,22 @@
...
+    """
+    Abstract base class for constraints.
+
+    A constraint object represents a region over which a variable is valid,
+    e.g. within which a variable can be optimized.
+    """
     def __call__(self, x):
         raise NotImplementedError
 
+    def check(self, value):
```
fehiepsi, line 66: We expose this doc so I want to match PyTorch behavior here.

fehiepsi:
@fritzo @neerajprad I have removed `sample` method and matched the constructor with the interface of Pyro. Because there is no sample method, I changed the logic of `init_to_uniform`, `init_to_median` a bit to bypass the NotImplementedError. Also, please let me know if there is any confusing point in the docs.

I also added the test `test_improper_expand` but still couldn't catch the expand issue. I still think that things should work as expected. >"< Could you elaborate it a bit more?

fehiepsi:
Hi @vanAmsterdam, the ordinal-regression gist looks great! Do you want to make a PR for it? I believe `ordinal_regression` is a very good example for the next NumPyro release.

vanAmsterdam:
> Hi @vanAmsterdam, the ordinal-regression gist looks great! Do you want to make a PR for it? I believe `ordinal_regression` is a very good example for the next NumPyro release.

sure! here it is: https://github.com/pyro-ppl/numpyro/pull/619#issue-428460022

numpyro/distributions/distribution.py
```diff
@@ -350,6 +350,75 @@ def variance(self):
         return np.broadcast_to(self.base_dist.variance, self.batch_shape + self.event_shape)
 
 
+class ImproperUniform(Distribution):
+    """
+    A helper distribution with zero :meth:`log_prob` over the `support` domain.
+
+    .. note:: `sample` method is not implemented for this distribution. In autoguide and mcmc,
+        initial parameters for improper sites are derived `init_to_uniform` or `init_to_value`
```
neerajprad, line 358: nit: `derived from`

test/test_infer_util.py
```diff
@@ -253,3 +253,17 @@ def model(data):
...
+    def model():
+        population = np.array([1000., 2000., 3000.])
+        with numpyro.plate("region", 3):
+            numpyro.sample("incidence",
+                           dist.ImproperUniform(support=constraints.interval(0, population),
+                                                batch_shape=(3,),
+                                                event_shape=(3,)))
+
+    model_info = initialize_model(random.PRNGKey(0), model)
+    assert model_info.param_info.z['incidence'].shape == (3, 3)
```
neerajprad, line 269: What I wanted to verify for this example is that with `event_shape=()`, this should be  (3,). Is that correct?

fehiepsi: Let me test it. I don't know what happens when users provide an invalid event_shape.

fehiepsi: Thanks, @neerajprad! I just added that test case and fixed the issue at `init_to_uniform`. Does the fix sound correct to you?

numpyro/infer/initialization.py
```diff
@@ -22,13 +23,17 @@ def init_to_median(site=None, num_samples=15):
     if site['type'] == 'sample' and not site['is_observed'] and not site['fn'].is_discrete:
         rng_key = site['kwargs'].get('rng_key')
         sample_shape = site['kwargs'].get('sample_shape')
-        samples = site['fn'].sample(rng_key, sample_shape=(num_samples,) + sample_shape)
-        return np.median(samples, axis=0)
+        try:
+            samples = site['fn'].sample(rng_key, sample_shape=(num_samples,) + sample_shape)
+            return np.median(samples, axis=0)
+        except NotImplementedError:
+            return init_to_uniform(site)
```
neerajprad, line 30: Can we just return `prototype_value = np.full(site['fn'].shape(), np.nan)` instead of calling `init_to_uniform` so that we don't need to additional work like calling `random.split`?

test/test_infer_util.py
```diff
@@ -253,3 +253,18 @@ def model(data):
...
+
+
+@pytest.mark.parametrize('event_shape', [(3,), ()])
+def test_improper_expand(event_shape):
+
+    def model():
+        population = np.array([1000., 2000., 3000.])
+        with numpyro.plate("region", 3):
+            numpyro.sample("incidence",
+                           dist.ImproperUniform(support=constraints.interval(0, population),
```
neerajprad, line 265: Can we also add a `log_prob` shape check here?


```python
d = dist.ImproperUniform(support=constraints.interval(0, population)
incidence = numpyro.sample("incidence",
                                                d,
                                                batch_shape=(3,),
                                                event_shape=event_shape))
assert d.log_prob(incidence).shape == (3,)
```

test/test_infer_util.py
neerajprad: Great, that makes sense.

numpyro/infer/initialization.py
fehiepsi: For substitute_fn, I think it only works if we substitue(seed(model), ...) because seed will give rng_key for the site we want to apply substitute_fn. Inside substitute_fn, we dont use sample primitive, so there is no random.split here IIUC.

fehiepsi:
Thank @neerajprad and @fritzo for reviewing!

PR #613, opened 2020-06-02
Filter out discrete sites

fehiepsi:
This filters out discrete sites in some utilities. In addition, I correct the shape of `log_det` term in potential_energy computation, which is useful for enum potential energy.

### DIFF
--- a/numpyro/infer/util.py
+++ b/numpyro/infer/util.py
@@ -11,6 +11,7 @@ import jax.numpy as np
 
 import numpyro
 from numpyro.distributions.transforms import biject_to
+from numpyro.distributions.util import sum_rightmost
 from numpyro.handlers import seed, substitute, trace
 from numpyro.infer.initialization import init_to_uniform, init_to_value
 from numpyro.util import not_jax_tracer, while_loop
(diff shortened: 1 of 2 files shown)


numpyro/infer/util.py
```diff
@@ -120,6 +121,21 @@ def substitute_fn(site):
             (return_deterministic and (v['type'] == 'deterministic'))}
 
 
+def _unconstrain_reparam(params, site):
```
fehiepsi, line 124: I split this out to use it for `enum_potential_energy`.

numpyro/infer/util.py
```diff
@@ -120,6 +121,21 @@ def substitute_fn(site):
...
+    name = site['name']
+    if name in params:
+        p = params[name]
+        t = biject_to(site['fn'].support)
+        value = t(p)
+
+        log_det = t.log_abs_det_jacobian(p, value)
+        log_det = sum_rightmost(log_det, np.ndim(log_det) - np.ndim(value) + len(site['fn'].event_shape))
```
fehiepsi, line 132: @eb8680 will this resolve your logic at [this line](https://github.com/pyro-ppl/numpyro/pull/572/files#diff-fee397191e0cc75a2e33eea8f7ecb6d5R197)? Here, `log_det` shape and `site.log_prob` shape will have the same shape (or broadcastable).

eb8680: I think that logic was to account for dimensions that weren't marked as plate or enumeration dimensions, which includes both event dimensions and batch dimensions that should have been plates but weren't.  My code wasn't tested directly so if your code ensures that all such dimensions have been removed then it's probably an improvement.

fehiepsi: This only guarantees that this new `log_det` site will have the same `dim_to_name`, which is specified [here](https://github.com/pyro-ppl/numpyro/pull/572/files#diff-bd74ce0e030e5b37b3cb5308e7963720R527), as the corresponding sample site. To be honest, I think this is correct but not sure if it is equivalent to the way you resolved in `enum_potential_energy`. Here this log_det site will be an observed site with event_shape (0,) (it is [unit distribution](https://github.com/pyro-ppl/pyro/blob/dev/pyro/distributions/unit.py)) and batch_shape = log_prob.shape.

fehiepsi:
Thanks for reviewing!

PR #614, opened 2020-06-02
Cleanup autoguide before transfering 1-1 to the main module

fehiepsi:
This PR matches the interface of autoguide to be similar to Pyro. There are two differences:
+ here we use `latent_size` while Pyro uses `latent_dim`, should we make a switch? I remember that we already discussed this difference, but can't recall the resolution.
+ post-inference methods require `params` argument. I think this is necessary because there is no global param store for SVI. I also feel this way is more explicit.

I have optimized the implementation a bit by using `_get_posterior` rather than `_get_transform`. The first method can return `Normal(loc, scale)` while the later one has to go through an extra `transform` layer. This optimization also matches the behavior in Pyro.

I also added `get_posterior` method and replace the property `base_dist` by the method `get_base_dist` to match Pyro.

Hopefully, all tests will pass without any change.

### DIFF
--- a/numpyro/contrib/autoguide/__init__.py
+++ b/numpyro/contrib/autoguide/__init__.py
@@ -121,7 +121,6 @@ class AutoContinuous(AutoGuide):
     """
     def __init__(self, model, prefix="auto", init_strategy=init_to_uniform):
         self.init_strategy = init_strategy
-        self._base_dist = None
         super(AutoContinuous, self).__init__(model, prefix=prefix)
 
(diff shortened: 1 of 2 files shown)


neerajprad:
Do we need to make any updates to the notebooks?

fehiepsi:
We don't have any tutorials yet for autoguide. Or you mean some docs?

neerajprad:
> We don't have any tutorials yet for autoguide. Or you mean some docs?

I was remembering the neutra example, but that's using the neutra reparam, so please disregard that.

Issue #610, opened 2020-05-27
Support MaskedDistribution, ExpandedDistribution for TransformReparam


fehiepsi:
cc @neerajprad

fehiepsi:
@neerajprad Looking like your issue requires to handle the composition of those distributions? Or we just need to support some simple composition?
```
dist.TransformedDistribution(...).mask(...).expand(...)
```

neerajprad:
@fehiepsi - it will be nice if we could support arbitrary compositions, but let me think about it a bit to see if there's a natural way to do this.

fehiepsi:
Closed because ExpandedDistribution has been supported. For MaskedDistribution, we can use `ImproperUniform` instead (as illustrated in the docs of `ImproperUniform`).

PR #616, opened 2020-06-03
Use ExpandedDistribution in numpyro.plate

### LINKED ISSUE #615: Use ExpandedDistribution in numpyro.plate
Since #574, I think we should be able to switch our existing logic to use `ExpandedDistribution` instead of our current logic. This will align the underlying implementation with that of Pyro and possibly aid with funsor integration.

neerajprad:
Fixes #615. I noticed this while reviewing #612. 

 - [x] Check that there is no perf regression.

### DIFF
--- a/numpyro/distributions/distribution.py
+++ b/numpyro/distributions/distribution.py
@@ -267,6 +267,9 @@ class ExpandedDistribution(Distribution):
 
     def __init__(self, base_dist, batch_shape=()):
         self.base_dist = base_dist
+        if isinstance(base_dist, ExpandedDistribution):
+            batch_shape = self._broadcast_shape(base_dist.batch_shape, batch_shape)
+            base_dist = base_dist.base_dist
         super().__init__(base_dist.batch_shape, base_dist.event_shape)
         # adjust batch shape
         self.expand(batch_shape)
--- a/numpyro/distributions/distribution.py
+++ b/numpyro/distributions/distribution.py
@@ -311,11 +311,9 @@ class ExpandedDistribution(Distribution):
 
     def __init__(self, base_dist, batch_shape=()):
         if isinstance(base_dist, ExpandedDistribution):
-            base_dist = base_dist.base_dist
-        self.base_dist = base_dist
-        if isinstance(base_dist, ExpandedDistribution):
             batch_shape = self._broadcast_shape(base_dist.batch_shape, batch_shape)
             base_dist = base_dist.base_dist
+        self.base_dist = base_dist
         super().__init__(base_dist.batch_shape, base_dist.event_shape)
         # adjust batch shape
         self.expand(batch_shape)
--- a/numpyro/primitives.py
+++ b/numpyro/primitives.py
@@ -265,19 +265,16 @@ class plate(Messenger):
         frame = CondIndepStackFrame(self.name, self.dim, self.subsample_size)
         cond_indep_stack.append(frame)
         expected_shape = self._get_batch_shape(cond_indep_stack)
-        dist_batch_shape = msg['fn'].batch_shape if msg['type'] == 'sample' else ()
-        overlap_idx = max(len(expected_shape) - len(dist_batch_shape), 0)
-        trailing_shape = expected_shape[overlap_idx:]
-        # e.g. distribution with batch shape (1, 5) cannot be broadcast to (5, 5)
-        broadcast_shape = lax.broadcast_shapes(trailing_shape, dist_batch_shape)
-        if broadcast_shape != dist_batch_shape:
-            raise ValueError('Distribution batch shape = {} cannot be broadcast up to {}. '
-                             'Consider using unbatched distributions.'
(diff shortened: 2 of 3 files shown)


test/contrib/test_reparam.py
```diff
@@ -49,7 +49,6 @@ def model():
 
     with handlers.trace() as tr:
         value = handlers.seed(model, 0)()
-    assert isinstance(tr["x"]["fn"], dist.TransformedDistribution)
```
neerajprad, line 52: @fehiepsi - The `TransformedDistribution` will be the base distribution now, so I am not sure if this PR can go through until we resolve #610.

fehiepsi: Yeah, I will address it.

numpyro/primitives.py
```diff
@@ -265,19 +265,16 @@ def process_message(self, msg):
...
+        if msg['type'] == 'sample':
+            dist_batch_shape = msg['fn'].batch_shape
+            if 'sample_shape' in msg['kwargs']:
+                dist_batch_shape = msg['kwargs']['sample_shape'] + dist_batch_shape
+                msg['kwargs']['sample_shape'] = ()
+            overlap_idx = max(len(expected_shape) - len(dist_batch_shape), 0)
+            trailing_shape = expected_shape[overlap_idx:]
+            broadcast_shape = lax.broadcast_shapes(trailing_shape, dist_batch_shape)
+            batch_shape = expected_shape[:overlap_idx] + broadcast_shape
+            msg['fn'] = msg['fn'].expand(batch_shape)
```
fehiepsi, line 277: Should we also expand when sample_shape is available at process or postprocess trace handler for consistency?

fehiepsi: Also, maybe at the constructor of ExpandedDist, we can check if base dist is expanded dist. If so, we just need to store `base_dist.base_dist`. This is convenient for inspecting the sites under multiple plate statements.

neerajprad: > Should we also expand when sample_shape is available at process or postprocess trace handler for consistency?

`sample_shape` should already incorporated. Right now, the behavior is that the distribution's intrinsic shape is `sample_shape + dist.batch_shape` which needs to broadcast with all the plates. This was what we had earlier, so the behavior is kept consistent. We can instead do `sample_shape + broadcasted shape from plates and batch shape`, but that can be surprising. Ideally, we shouldn't use `sample_shape` with `plate` to avoid this kind of confusion.

> Also, maybe at the constructor of ExpandedDist, we can check if base dist is expanded dist. If so, we just need to store base_dist.base_dist. 

Let me see how we are doing this in Pyro, I agree multiple layers of `ExpandedDist` is not great for debugging.

fehiepsi: I see. It is a bit confusing. I like your change because it solves the shape broadcasting issue that we have before (the ValueError in the previous code).

neerajprad: Regarding..
> Also, maybe at the constructor of ExpandedDist, we can check if base dist is expanded dist. If so, we just need to store base_dist.base_dist.

Just by using nested plates, we will never hit this condition where the `base_dist` is already an `ExpandedDistribution`. This can only happen if the user uses `.expand` and `plate` together. Does that make sense? We can still introduce this optimization, but it shouldn't affect the vast majority of models.

fehiepsi: I see. I already add this optimization in #617 (just to make sure that we can use `TransformReparam` with multiple `.expand` op).

fehiepsi:
You'll need to merge the master branch to pass CI.

neerajprad:
> Which performance regressions that you think about? I think those shape manipulations here and in expand will negligible under jit.

I just wanted to see if we noticed any slowness in our test suite, but there is negligible effect, as you noted. This should be good to go once tests pass.

numpyro/distributions/distribution.py
```diff
@@ -313,6 +313,9 @@ def __init__(self, base_dist, batch_shape=()):
         if isinstance(base_dist, ExpandedDistribution):
             base_dist = base_dist.base_dist
         self.base_dist = base_dist
+        if isinstance(base_dist, ExpandedDistribution):
+            batch_shape = self._broadcast_shape(base_dist.batch_shape, batch_shape)
+            base_dist = base_dist.base_dist
```
fehiepsi, line 318: This optimization is addressed above. Do you think we need extra logic here?

neerajprad: I missed that during merge, I think we should still do an explicit broadcast so that. `dist.expand(5, 4).expand(6, 4)` raises an error. Will fix.

fehiepsi: Hmm, `.expand(5, 4).expand(6, 4)` seems to be invalid, hence an error is expected. Did you mean something like: `.expand(5, 1).expand(1, 4)`?

fehiepsi: You are right!!

PR #617, opened 2020-06-04
support Expand Reparam

fehiepsi:
Addresses #610.

I also add some missing methods from pyro TorchDistribution and rename `event_ndim` to `event_dim` in Delta distribution for consistency.

@neerajprad I can't find a proper way to support MaskedDistribution. Basically, mask(t(dist)) is different from `t(mask(dist))`. The difference between two log_probs is the log determinant of the transform. Adding jacobian adjustment will complicate the code. In addition, the composition of `Masked(Expand(Masked))` is complicated to deal with. :( It seems to be a regression, but I think users can use `Uniform`, which works with TransformReparam, for `Improper(interval)` (~we miss support for the dynamic improper `greater_than`~).

### DIFF
--- a/numpyro/contrib/reparam.py
+++ b/numpyro/contrib/reparam.py
@@ -97,10 +97,16 @@ class TransformReparam(Reparam):
     """
     def __call__(self, name, fn, obs):
         assert obs is None, "TransformReparam does not support observe statements"
+        batch_shape = fn.batch_shape
+        if isinstance(fn, dist.ExpandedDistribution):
+            fn = fn.base_dist
         assert isinstance(fn, dist.TransformedDistribution)
 
         # Draw noise from the base distribution.
-        x = numpyro.sample("{}_base".format(name), fn.base_dist)
+        # We need to make sure that we have the same batch_shape
+        reinterpreted_batch_ndims = fn.event_dim - fn.base_dist.event_dim
+        x = numpyro.sample("{}_base".format(name),
+                           fn.base_dist.to_event(reinterpreted_batch_ndims).expand(batch_shape))
 
         # Differentiably transform.
         for t in fn.transforms:
--- a/numpyro/distributions/distribution.py
+++ b/numpyro/distributions/distribution.py
@@ -150,6 +150,29 @@ class Distribution(object):
         """
         return self._event_shape
 
+    @property
+    def event_dim(self):
+        """
+        :return: Number of dimensions of individual events.
+        :rtype: int
+        """
+        return len(self.event_shape)
+
+    def shape(self, sample_shape=()):
+        """
+        The tensor shape of samples from this distribution.
+
+        Samples are of shape::
+
+            d.shape(sample_shape) == sample_shape + d.batch_shape + d.event_shape
+
+        :param tuple sample_shape: the size of the iid batch to be drawn from the
+            distribution.
+        :return: shape of samples.
+        :rtype: tuple
+        """
+        return sample_shape + self.batch_shape + self.event_shape
+
     def sample(self, key, sample_shape=()):
         """
         Returns a sample from the distribution having shape given by
@@ -228,6 +251,8 @@ class Distribution(object):
         """
         if reinterpreted_batch_ndims is None:
             reinterpreted_batch_ndims = len(self.batch_shape)
+        elif reinterpreted_batch_ndims == 0:
+            return self
         return Independent(self, reinterpreted_batch_ndims)
 
     def enumerate_support(self, expand=True):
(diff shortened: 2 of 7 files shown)


numpyro/distributions/distribution.py
```diff
@@ -246,8 +271,25 @@ def expand(self, batch_shape):
         :return: an instance of `ExpandedDistribution`.
         :rtype: :class:`ExpandedDistribution`
         """
+        batch_shape = tuple(batch_shape)
```
fehiepsi, line 274: In Pyro, it is [common](http://pyro.ai/examples/tensor_shapes.html#It-is-always-safe-to-assume-dependence) to use `.expand([n])` so it is better to support it here.

numpyro/distributions/distribution.py
```diff
@@ -246,8 +271,25 @@ def expand(self, batch_shape):
...
+        if batch_shape == self.batch_shape:
+            return self
```
fehiepsi, line 276: For those trivial cases, we don't use the derived distribution.

neerajprad:
> @neerajprad I can't find a proper way to support MaskedDistribution. Basically, mask(t(dist)) is different from t(mask(dist)). The difference between two log_probs is the log determinant of the transform. 

That seems fine to me, +1 to favoring code simplicity over supporting specialized cases. IIUC, instead of using `param` statements in HMC for improper priors, we now have two alternatives - using sample with `.mask(False)`, and the upcoming `ImproperUniform` distribution. Could you walk through a few cases when to use either of these, e.g. what to use with dynamic constraints? For the earlier [example](https://github.com/pyro-ppl/numpyro/pull/605#discussion_r431347344), can we use the ImproperUniform? It will be a good idea to document all the cases involved.

fehiepsi:
For proper distributions, there are 3 scenarios with dynamic constraints:
+ using transform reparam: I think this is the recommended way to resolve dynamic constraint issues. IIRC, thanks to this reparam, we removed dynamic constraint support. Using transform reparam is also helpful for mixing but it is a different story.
+ the model has deterministic sites: without transform reparam, the inference and post-process still work correctly thanks to unconstrained reparam (because we don't have to carry `inv_transforms` around and the model is replayed for post-processing)
+ the model has no deterministic sites: inference works but post-process does not work correctly because we didn't replay the model. I think this is a regression. We need to detect if there are dynamic constraints in the model. Previously, we checked if a site is `TransformedDistribution` but that does not cover all cases (e.g. Expanded/Independent of transformed distribution). I'm not sure what is the best way to deal with this issue. Could you give me some suggestions? Here a solution I have in mind: just simply check the support, we set `replay=True` if it is `interval` or `greater_than` where their args `low`, `high` are not Python scalars (which is a rare situation - where priors are Uniform(0, 1) or Uniform(1, 2) or TruncatedCauchy(0.5, 3, 4)...). WDYT?

About `ImproperUniform` vs `.mask(False)`, I think it is better to use `ImproperUniform` from now on (`.mask(False)` works but it has limitations: there are some constraints which have no corresponding distributions).

About improper distribution with dynamic constraint, as we discussed above, currently we don't have a mechanism to reparameterize it (hence mixing might be bad). This is the third scenario that I mentioned above: the inference works but postprocessing failed.

neerajprad:
I see, the combination of `ExpandedDistribution` + `TransformedDistribution` in case 3 will become more important with #615, though overall I'm thinking that handling dynamic constraints should be low priority. For now, let us just put a `TODO` and have a more sophisticated logic for deciding when to replay the model, if this issue comes up. 

You'll need to merge changes from master.

fehiepsi:
Yes, I added a TODO for it. I'll add a small scalar check there to filter out most incorrect usages.

numpyro/contrib/reparam.py
```diff
@@ -97,10 +97,16 @@ class TransformReparam(Reparam):
     """
     def __call__(self, name, fn, obs):
         assert obs is None, "TransformReparam does not support observe statements"
+        batch_shape = fn.batch_shape
+        if isinstance(fn, dist.ExpandedDistribution):
+            fn = fn.base_dist
```
neerajprad, line 102: We have `_wrap/_unwrap` logic in pyro's `Reparam` for [Independent](https://github.com/pyro-ppl/pyro/blob/73af4ff/pyro/infer/reparam/reparam.py). Do you think having something like that for expanded and masked distributions might be helpful here?

numpyro/primitives.py
```diff
@@ -137,8 +137,6 @@ def param(name, init_value=None, **kwargs):
         'args': (init_value,),
         'kwargs': kwargs,
         'value': None,
-        'mask': None,
-        'scale': None,
```
neerajprad, line 141: We are not yet using `MaskedDistribution` in MaskMessenger in Pyro. Are there any issues that this might cause when integrating with funsor? See - https://github.com/pyro-ppl/pyro/pull/2284. cc @fritzo.

numpyro/distributions/distribution.py
```diff
@@ -150,6 +150,29 @@ def event_shape(self):
         """
         return self._event_shape
 
+    @property
+    def event_dim(self):
```
neerajprad, line 154: I would suggest creating separate PRs for non-brainer changes like these, so that we can merge it asap. :) Don't worry about it for this one though.

numpyro/contrib/reparam.py
fehiepsi: I think something like `_unexpand` and `_expand` would do the job here. Let's make those helpers when we have more reparameterization classes and the logic here is repeated. I am not sure about Masked because I would like to avoid dealing with jacobian in TransformReparam.

numpyro/primitives.py
fehiepsi: I am not sure. In [enum PR](https://github.com/pyro-ppl/numpyro/pull/572/files#diff-fee397191e0cc75a2e33eea8f7ecb6d5R186), `mask` is used to mask out the log_prob term, as in the usual `log_density` implementation. I am not sure about the following too: "automatic broadcasting to the mask shape when the mask is enumerated". AFAIK we just use `mask` for log_prob computation and that job is handled in MaskedDistribution.log_prob method.

PR #618, opened 2020-06-05
Filter out most models with static supports in MCMC

fehiepsi:
This PR adds a simple logic to decide if we need to replay the model. With this, we can replay model in some special cases and don't need to replay models for most cases.

### DIFF
--- a/docs/source/primitives.rst
+++ b/docs/source/primitives.rst
@@ -17,7 +17,7 @@ plate
 
 plate_stack
 -----------
-.. autoclass:: numpyro.primitives.plate_stack
+.. autofunction:: numpyro.primitives.plate_stack
 
 deterministic
 -------------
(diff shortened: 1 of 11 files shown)


docs/source/primitives.rst
```diff
@@ -17,7 +17,7 @@ plate
 
 plate_stack
 -----------
-.. autoclass:: numpyro.primitives.plate_stack
+.. autofunction:: numpyro.primitives.plate_stack
```
fehiepsi, line 20: use `autoclass` raises a warning with doctest

numpyro/infer/util.py
```diff
@@ -265,8 +266,17 @@ def get_model_transforms(model, model_args=(), model_kwargs=None):
...
+            support = v['fn'].support
+            inv_transforms[k] = biject_to(support)
+            # XXX: the following code filters out most situations with dynamic supports
+            args = ()
+            if isinstance(support, _GreaterThan):
+                args = ('lower_bound',)
+            elif isinstance(support, _Interval):
+                args = ('lower_bound', 'upper_bound')
+            for arg in args:
+                if not isinstance(getattr(support, arg), (int, float)):
```
neerajprad, line 278: 👍

PR #619, opened 2020-06-05
Example of ordinal regression with ImproperDistribution

vanAmsterdam:
as requested here https://github.com/pyro-ppl/numpyro/pull/612#issuecomment-639229617

can someone help out with updating the docs? Haven't done that before

### DIFF
--- a/notebooks/ordinalregression.ipynb
+++ b/notebooks/ordinalregression.ipynb
@@ -1,0 +1,219 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Ordinal regression with ImproperUniform distribution"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from jax import numpy as np, random\n",
+    "import numpyro\n",
+    "from numpyro import sample\n",
+    "from numpyro.distributions import constraints, Normal, ImproperUniform, Categorical, OrderedLogistic\n",
+    "from numpyro.infer.mcmc import NUTS, MCMC\n",
+    "import pandas as pd\n",
+    "import seaborn as sns\n",
+    "\n",
+    "num_chains = 4\n",
+    "numpyro.set_host_device_count(num_chains)\n",
+    "numpyro.set_platform('cpu')"
+   ]
+  },
+  {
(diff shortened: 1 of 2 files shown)


fehiepsi:
Hi @vanAmsterdam , the content looks great. Thanks for contributing your example! Please don't worry about the doc format, we'll check it again.

Might I add a section on the usage of
```
c_y = sample("cutpoints",
                      dist.TransformedDistribution(dist.Normal(0, 1).expand([nclasses]),
                                                   OrderedTransform()))
```
to put a weak prior on the intercepts, instead of the cutpoints `c_y`?

vanAmsterdam:
Thanks; yes I'll write something;

Do you mean adding the `TransformedDistribution` method as `model3`, another way of getting the same result?
I'm not sure I follow what you mean with intercepts as opposed to cutpoints. What intercepts do you mean?

fehiepsi:
Sorry, I used the language in Statistical Rethinking book. It seems that that terminology is not common.

What I meant is the models in the tutorial use improper priors for cutpoints. In my last comment, we use a "proper" prior for those cutpoints. The motivation is:
+ assume that predictor is 0, cutpoints `c` orderly lie in the log-odds space (y-axis of [this figure](https://en.wikipedia.org/wiki/Logit#/media/File:Logit.svg)) and each c_i corresponds to a cumulative probability p_i of ordered categories (i.e. if the outputs are scores of students, then p_i is the probability that the score <= i)
+ We can put positive-support priors on the differences (c_{i + 1} - c_{i}) of those cutpoints. In my last comment, I used LogNormal(0, 1) priors. That also means that `log(c_{i+1} - c_{i})` (which belongs to the real domain of OrderedTransform - they are called intercepts in the above book IIRC) has Normal(0,1) prior.

Does that make sense to you? Do you want to write some words about it in the tutorial?

fehiepsi:
@vanAmsterdam Feel free to ignore my comment on the usage of `TransformedDistribution`. We can improve the content in a follow-up PR. :)

vanAmsterdam:
I’m a bit slow to respond due to a deadline, are you in a hurry to close the issue or can it linger for a few weeks?

fehiepsi:
Absolutely no! I just worried that you found the usage of TransformedDistribution a bit counterintuitively to address. Please take your time! I hope you enjoy doing open-source contribution so I don't want something like 'hurry' take away such pleasure. ;)

PR #625, opened 2020-06-10
move autoguide the the main module

fehiepsi:
We already have many autoguides and the interface is pretty similar to Pyro. So as a part of 0.3 release, I want to move it to the main module. The files are copied 1-1 with only change at `self.latent_size` -> `self.latent_dim` to match Pyro.

### TODO
- [x] add a note to docs mention how to get `params`.


numpyro/infer/autoguide.py
```diff
@@ -1,12 +1,7 @@
 # Copyright Contributors to the Pyro project.
 # SPDX-License-Identifier: Apache-2.0
```
neerajprad, line 3: I would suggest that we do this in a way where we raise a warning first so that existing code doesn't immediately break, like we did in Pyro:

```python

from numpyro.infer.autoguide import *  # noqa F403

warnings.warn("numpyro.contrib.autoguide has moved to numpyro.infer.autoguide. "
              "The contrib alias will stop working in future versions.",
              FutureWarning)
```

neerajprad:
@fehiepsi - LGTM. If the interface is the same as Pyro, we can also expose this through `pyro-api` so that we can continue to maintain a uniform interface across both.

numpyro/infer/autoguide.py
fehiepsi: Oh, good idea!!!! I have to make another PR blocked because of this issue. With this, I won't have to do that. Thanks!

fehiepsi:

> If the interface is the same as Pyro

Unfortunately, the post-inference interface is different at the `params` requirement. :(

PR #626, opened 2020-06-11
Update jax version to test Travis with latest jax release

fehiepsi:
Also, remove `poisson` and `categorical_logits` sampler because we have an upstream version now. `logsumexp` issue and warning issue, which has blocked us from testing the latest jax release in Travis) is also fixed upstream.

### DIFF
--- a/numpyro/distributions/discrete.py
+++ b/numpyro/distributions/discrete.py
@@ -38,12 +38,10 @@ from numpyro.distributions.util import (
     binary_cross_entropy_with_logits,
     binomial,
     categorical,
-    categorical_logits,
     clamp_probs,
     get_dtype,
     lazy_property,
     multinomial,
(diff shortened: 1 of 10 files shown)


numpyro/distributions/discrete.py
```diff
@@ -543,7 +541,7 @@ def __init__(self, rate, validate_args=None):
         super(Poisson, self).__init__(np.shape(rate), validate_args=validate_args)
 
     def sample(self, key, sample_shape=()):
-        return poisson(key, self.rate, shape=sample_shape + self.batch_shape)
+        return random.poisson(key, device_put(self.rate), shape=sample_shape + self.batch_shape)
```
fehiepsi, line 544: Upstream implementation requires device_array `rate`.

PR #629, opened 2020-06-12
update README file to use reparam handler

fehiepsi:
This PR updates README description to reflect some changes in automatic reparameterization mechanism. I also mentioned that we support many flexible autoguides for ADVI.

@neerajprad should we add some import statements so that users can reproduce code in the readme without having to look at examples to make correct import statements?


neerajprad:
> @neerajprad should we add some import statements so that users can reproduce code in the readme without having to look at examples to make correct import statements?

That sounds good - maybe we can use the `<details>` tag if the imports are too many so as to not detract from the actual example code. 

<details>
 <summary>imports</summary>

```python
import numpyro.handlers as handlers
```

</details>

neerajprad:
I think there's something wrong with travis - there is a duplicate `travis CI` stage.

README.md
```diff
@@ -50,71 +67,82 @@ Let us infer the values of the unknown parameters in our model by running MCMC u
...
 >>> mcmc.run(rng_key, J, sigma, y=y, extra_fields=('potential_energy',))
+
 ```
 
 We can print the summary of the MCMC run, and examine if we observed any divergences during inference. Additionally, since we collected the potential energy for each of the samples, we can easily compute the expected log joint density.
 
 ```python
 >>> mcmc.print_summary()
-
+<BLANKLINE>
```
neerajprad, line 77: Why `<BLANKLINE>` - is this meant to be rendered differently?

fehiepsi:
@neerajprad I have added the import section and reformatted the outputs a bit to make sure that doctest pass locally `python -m doctest -v README.md`.

README.md
```diff
@@ -16,20 +16,36 @@ NumPyro is a small probabilistic programming library that provides a NumPy backe
...
+>>> import numpyro
+>>> import numpyro.distributions as dist
+>>> from numpyro.infer import MCMC, NUTS, Predictive
+>>> from numpyro.infer.reparam import TransformReparam
+
+```
+
+</details>
 
 The data is given by:
```
neerajprad, line 42: Now I'm wondering if we should put imports separately for each of the code blocks? For instance, for this one, all we need is `import numpy as np` for which we don't need to use the `details` block. I think that way we may not even need to use the `details` block if there are only a few imports for each of the code blocks.

README.md
neerajprad: Oh..so `<BLANKLINE>` is for doctest?

fehiepsi: The output of [doctest](https://docs.python.org/3/library/doctest.html#how-are-docstring-examples-recognized) expects <BLANKLINE> here:  "If expected output does contain a blank line, put <BLANKLINE> in your doctest example each place a blank line is expected." Do you know a trick to remove it?

README.md
fehiepsi: You are right! Each block contains at most 2 import lines. Let me make the change.

README.md
neerajprad: One option is to add `# doctest: +SKIP` to the specific lines. I think these are likely to be flaky so its a good idea to skip asserting on the actual output in these cases.

README.md
```diff
@@ -16,122 +16,146 @@ NumPyro is a small probabilistic programming library that provides a NumPy backe
...
-  theta[4]      3.62      4.66      3.75     -3.87     11.17   1091.53      1.00
-  theta[5]      3.92      4.43      4.06     -2.41     11.09   1179.74      1.00
-  theta[6]      5.88      4.84      5.34     -1.45     13.11    881.38      1.00
-  theta[7]      4.63      4.86      4.64     -3.57     11.80   1065.27      1.00
-
+>>> mcmc.print_summary(exclude_deterministic=False)
+<BLANKLINE>
+                   mean       std    median      5.0%     95.0%     n_eff     r_hat
+           mu      4.08      3.51      4.14     -1.69      9.71    720.43      1.00
+          tau      3.96      3.31      3.09      0.01      8.34    488.63      1.00
```
neerajprad, line 116: `n_eff` seems to have gone down for both `mu` and `tau`. Is this just noise?

README.md
fehiepsi: Yes, I think we can do so and add this doctest to travis.

README.md
fehiepsi: I think so.

neerajprad:
Could you merge the changes from master? I think that should get us past the duplicate build issues.

fehiepsi:

PR #631, opened 2020-06-12
Remove `mask` field in the trace

fehiepsi:
and defer the logic to `MaskedDistribution`.

Without this, we are incorrectly masking out `log_det` sites (which is constructed in substitute_fn) in unconstrained reparam.


neerajprad:
You'll need to merge changes from the latest master.

numpyro/distributions/util.py
```diff
@@ -346,6 +346,17 @@ def clamp_probs(probs):
     return jnp.clip(probs, a_min=finfo.tiny, a_max=1. - finfo.eps)
 
 
+def is_identically_one(x):
+    """
+    Check if argument is exactly the number one. True for the number one;
+    false for other numbers; false for ndarrays.
+    """
+    if isinstance(x, (int, float)):
```
neerajprad, line 354: Do we need to check for numpy scalars separately?

fehiepsi: I am not so sure how to check that properly because some numpy scalars are already subclasses of `int, float`, some are not. This simplifies stuff under `plate`, where `scale = 1.` for most of the situations.

PR #632, opened 2020-06-14
Added the Laplace distribution.

nh2liu:
I've been using the Laplace distribution as a prior and thought it would be a good addition to numpyro.

This PR implements the Laplace distribution parameterized by `loc` and `scale`.


fehiepsi:
Thanks, @nh2liu !

Issue #620, opened 2020-06-07
Support `do` handler


fehiepsi:
While answering a [question](https://forum.pyro.ai/t/beginners-toy-examples-tutorials-for-exact-inference-on-discrete-variables/1911) in the forum, I found that `do` handler will be useful if users want to intervene with a value for a site to compute conditional distribution, rather than joint distribution.

Here is how [poutine.do](https://github.com/pyro-ppl/pyro/blob/dev/pyro/poutine/do_messenger.py) messenger implemented in Pyro. Basically, the idea is to [condition](http://num.pyro.ai/en/stable/handlers.html#condition) and add a field `stop=True` (equivalent to using [block](http://num.pyro.ai/en/stable/handlers.html#block) handler) to the message.

TuanNguyen27:
I'd like to work on this over the weekend if no one claims it by then !

fehiepsi:
Thank @TuanNguyen27 for addressing this issue!

PR #638, opened 2020-06-16
Add `do` handler

TuanNguyen27:
Solves #620


TuanNguyen27:
Thanks for the review, @fehiepsi ! I'm going to address your comments shortly :)

test/test_counterfactual.py
```diff
@@ -0,0 +1,76 @@
...
+            if observations[name] is not None:
+                assert tr.nodes[name]['is_observed']
+                assert_allclose(observations[name], tr.nodes[name]['value'])
+            if interventions[name] is not None:
+                assert_allclose(interventions[name], actual_values[name])
+            if interventions[name] != observations[name]:
+                assert_not_equal(interventions[name], tr.nodes[name]['value'])
+
+
+def assert_not_equal(x, y, prec=1e-5, msg=''):
```
fehiepsi, line 72: nit: I guess you can [assert_raises](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_raises.html#numpy.testing.assert_raises) here:
`assert_raises(AssertionError, assert_allclose, x, y, atol=prec)`. Can we also keep `atol` as-is, instead of using `prec`? You can also remove this utility and use `assert_raises(AssertionError, assert_allclose, x, y)` to match the pattern of other tests in `test_handlers` (this is not necessary if you found `assert_not_equal` clearer).

test/test_counterfactual.py
```diff
@@ -0,0 +1,76 @@
...
+from numpy.testing import assert_allclose
+
+
+@pytest.mark.parametrize('intervene,observe,flip', [
+    (True, False, False),
+    (False, True, False),
+    (True, True, False),
+    (True, True, True),
+])
+def test_counterfactual_query(intervene, observe, flip):
```
fehiepsi, line 18: Thanks for adding this test!

nit: I think we can move this to `test_handlers` file.

numpyro/handlers.py
```diff
@@ -437,3 +438,76 @@ def process_message(self, msg):
...
+            msg['_intervener_id'] = self._intervener_id
+
+            # split node, avoid reapplying self recursively to new node
+            new_msg = msg.copy()
+            apply_stack(new_msg)
+
+            intervention = self.data.get(msg['name'])
+            msg['name'] = msg['name'] + "__CF"  # mangle old name
+            msg['value'] = intervention
+            if msg['is_observed']:
```
fehiepsi, line 510: Just my curiousity: why do we need this? I think `do(condition(...))` is allowable.

numpyro/handlers.py
```diff
@@ -437,3 +438,76 @@ def process_message(self, msg):
...
+      >>> import jax.numpy as jnp
+      >>> import numpyro
+      >>> from numpyro.handlers import do, trace, seed
+      >>> import numpyro.distributions as dist
+      >>> def model(x):
+      ...     s = numpyro.sample("s", dist.Normal(), constraint=constraints.positive)
+      ...     z = numpyro.sample("z", dist.Normal(x, s))
+      ...     return z ** 2
+      >>> with trace() as exec_trace:
+      ...     z_square = seed(intervened_model, 0)(x)
```
fehiepsi, line 480: typo: missing the definition of `intervened_model`.

numpyro/handlers.py
```diff
@@ -437,3 +438,76 @@ def process_message(self, msg):
...
+    set the return values of those sites equal to the values
+    as if they were hard-coded to those values
+    and introduce fresh sample sites with the same names
+    whose values do not propagate.
+    Composes freely with :func:`~numpyro.handlers.condition`
+    to represent counterfactual distributions over potential outcomes.
+    See Single World Intervention Graphs [1] for additional details and theory.
+
+    This is equivalent to replacing `z = numpyro.sample("z", ...)` with
+    `z = jnp.array(1.)`
```
fehiepsi, line 456: nit: you can use `z = 1.` for simplicity

numpyro/handlers.py
```diff
@@ -437,3 +438,76 @@ def process_message(self, msg):
...
+    **Example:**
+
+    .. doctest::
+
+      >>> import jax.numpy as jnp
+      >>> import numpyro
+      >>> from numpyro.handlers import do, trace, seed
+      >>> import numpyro.distributions as dist
+      >>> def model(x):
+      ...     s = numpyro.sample("s", dist.Normal(), constraint=constraints.positive)
```
fehiepsi, line 476: If you use `sample`, you should use something like `dist.LogNormal()` prior for positive constraint. (`sample` does not have `constraint` arg)

test/test_handlers.py
```diff
@@ -116,6 +117,24 @@ def model():
         handlers.condition(model, {'y': 3.})()
 
 
+def test_do():
```
fehiepsi, line 120: this is tested in docs some we  might skip it here.

test/test_counterfactual.py
TuanNguyen27: I agree ! I was lazy so i copy pasted stuff from pyro's version of this test...:D

numpyro/handlers.py
TuanNguyen27: ah, I think I also copied this from pyro's implementation, let me remove it.

numpyro/handlers.py
```diff
@@ -437,3 +438,74 @@ def process_message(self, msg):
...
+      ...     return z ** 2
+      >>> intervened_model = handlers.do(model, data={"z": 1.})
+      >>> with trace() as exec_trace:
+      ...     z_square = seed(intervened_model, 0)(x)
+      >>> assert exec_trace['z']['value'] != 1.
+      >>> assert not exec_trace['z']['is_observed']
+      >>> assert not exec_trace['z'].get('stop', None)
+      >>> assert z_square == 1
+    """
+    def __init__(self, fn=None, data=None):
```
fehiepsi, line 486: I think this is almost done. We just need to change `data` to `param_map` here and throughout docs and tests. In NumPyro, we used `param_map` for `condition`, `substitute`,... instead of `data`.

@neerajprad Do you think it is better to change all the names to `data`? The benefit is it will match Pyro api, but we have to raise Deprecation error if users use `param_map`. I don't have a strong opinion on this.

neerajprad: IIRC `param_map` was specifically in the context of `substitute` since we can't have a global param store. I'm not sure if that distinction is relevant now, and otherwise I think it is a good idea to keep parity with Pyro wherever possible. We can change the argument for condition (and substitute) in a separate PR.

fehiepsi: Agreed! I'll try to collect some differences that I know in a separate PR.

numpyro/handlers.py
```diff
@@ -437,3 +438,74 @@ def process_message(self, msg):
 
         if value is not None:
             msg['value'] = value
+
+
+class do(Messenger):
+    """
+    Given a stochastic function with some sample statements
```
neerajprad, line 445: nit: can we somewhat justify the docstring text?

numpyro/handlers.py
```diff
@@ -84,9 +84,10 @@
...
 import jax.numpy as jnp
 
-from numpyro.primitives import Messenger
+from numpyro.primitives import Messenger, apply_stack
 from numpyro.util import not_jax_tracer
 
+
 __all__ = [
     'block',
     'condition',
```
fehiepsi, line 93: Could you also add `'do'` here?

numpyro/handlers.py
TuanNguyen27: @fehiepsi do you still want me to change `data` to `param_map` ?

fehiepsi: Ah, sorry, I think you can keep it because we are going to change `param_map` to `data` for other handlers.

test/test_handlers.py
```diff
@@ -240,3 +240,63 @@ def guide():
...
+            model = handlers.do(model, data=interventions)
+        if observe:
+            model = handlers.condition(model, param_map=observations)
+    elif flip and intervene and observe:
+        model = handlers.do(
+            handlers.condition(model, param_map=observations),
+            data=interventions)
+
+    tr = handlers.trace(model).get_trace()
+    actual_values = tr["_RETURN"]["value"]
```
fehiepsi, line 277: In NumPyro, there is not `_RETURN` site. You can use
```
with handlers.trace() as tr:
    actual_values = model()
```
to get both trace `tr` and the returned value `actual_values`.

fehiepsi:
@TuanNguyen27 I think your PR is blocked because of recent changes in JAX. Please ignore unrelated issues and ping me if this is ready to review.

test/test_handlers.py
```diff
@@ -243,6 +243,66 @@ def guide():
...
+            y = numpyro.sample("y", dist.Normal(x, 1))
+            z = numpyro.sample("z", dist.Normal(y, 1))
+            w = numpyro.sample("w", dist.Normal(z, 1))
+            return dict(x=x, y=y, z=z, w=w)
+
+    if not flip:
+        if intervene:
+            model = handlers.do(model, data=interventions)
+        if observe:
+            model = handlers.condition(model, param_map=observations)
```
fehiepsi, line 271: Could you switch `param_map` to `data`? We just change it in a recent PR. Sorry about that!

test/test_handlers.py
```diff
@@ -243,6 +243,66 @@ def guide():
...
+    elif flip and intervene and observe:
+        model = handlers.do(
+            handlers.condition(model, param_map=observations),
```
fehiepsi, line 274: ditto: `param_map` to `data`

test/test_handlers.py
```diff
@@ -243,6 +243,66 @@ def guide():
...
+        actual_values = model()
+    for name in sites:
+        # case 1: purely observational query like handlers.condition
+        if not intervene and observe:
+            if observations[name] is not None:
+                assert tr[name]['is_observed']
+                assert_allclose(observations[name], actual_values[name])
+                assert_allclose(observations[name], tr[name]['value'])
+            if interventions[name] != observations[name]:
+                assert_raises(AssertionError, assert_allclose, interventions[name], actual_values[name])
```
fehiepsi, line 287: For `assert_raises` statements, I think you need to add:
```
if interventions[name] is not None:
    assert_raises(AssertionError, assert_allclose, interventions[name], actual_values[name])
```

test/test_handlers.py
```diff
@@ -243,6 +243,66 @@ def guide():
...
+        # case 2: purely interventional query like old handlers.do
+        elif intervene and not observe:
+            assert not tr[name]['is_observed']
+            if interventions[name] is not None:
+                assert_allclose(interventions[name], actual_values[name])
+            assert_raises(AssertionError, assert_allclose, observations[name], tr[name]['value'])
```
fehiepsi, line 293: ditto: `if observations[name] is not None`

test/test_handlers.py
```diff
@@ -243,6 +243,66 @@ def guide():
...
+            assert_raises(AssertionError, assert_allclose, interventions[name], tr[name]['value'])
```
fehiepsi, line 294: ditto: `if interventions[name] is not None`

test/test_handlers.py
```diff
@@ -243,6 +243,66 @@ def guide():
...
+        # case 3: counterfactual query mixing intervention and observation
+        elif intervene and observe:
+            if observations[name] is not None:
+                assert tr[name]['is_observed']
+                assert_allclose(observations[name], tr[name]['value'])
+            if interventions[name] is not None:
+                assert_allclose(interventions[name], actual_values[name])
+            if interventions[name] != observations[name]:
+                assert_raises(AssertionError, assert_allclose, interventions[name], tr[name]['value'])
```
fehiepsi, line 303: ditto: `if interventions[name] is not None`

TuanNguyen27: so in this case, is it `if (interventions[name] != observations[name]) and interventions[name] is not None:` or just `if interventions[name] is not None` ?

fehiepsi: I think the former is right. We need those `is not None` check because `assert_allclose` can't compare None values. You can test locally with `pytest -vs test/test_handlers.py -k counterfactual`.

TuanNguyen27:
@fehiepsi sorry for the delay...will try to fix the tests shortly :)

fehiepsi:
@TuanNguyen27 Please take your time. Just in case you want to include this in the next release (I think your PR is nice to have), we are going to make a release in a week or so. IIRC, adding those `if ... is not None:` fixes the issue but if not, please let me know :D. After fixing them, could you also add `do` handler to [this docs file](https://github.com/pyro-ppl/numpyro/blob/master/docs/source/handlers.rst)?

fehiepsi:
@TuanNguyen27 Do you want me to add those `if interventions[name] is not None` (because those `assert_raises` do not work with `None` values) as in my previous comment to your tests and merge this?

TuanNguyen27:
> @TuanNguyen27 Do you want me to add those `if interventions[name] is not None` (because those `assert_raises` do not work with `None` values) as in my previous comment to your tests and merge this?

Yes please ! Sorry that this has been delayed again and again ...

fehiepsi:
No worries! Thanks for this great PR! I personally learn a lot here. :)

PR #641, opened 2020-06-19
Support `hide` argument in block handler

### LINKED ISSUE #640: Support `hide` argument in `block`
If we want to hide some sites, using `hide_fn=lambda site: site['name'] == 'x'` is a bit inconvenient. We can mimic Pyro by allowing an extra `hide` argument, so users can just simply use `hide=['x']`.

fehiepsi:
Fixes #640. The `hide` argument is useful when we want to compute the conditional distribution p(y|x=x0), which is `block(condition_or_substitute_or_do(fn, {'x': x0}), hide=['x']))`

### DIFF
--- a/numpyro/contrib/distributions/distribution.py
+++ b/numpyro/contrib/distributions/distribution.py
@@ -9,10 +9,25 @@
 # Copyright (c) 2003-2019 SciPy Developers.
 # All rights reserved.
 from contextlib import contextmanager
+try:
+    from scipy._lib._util import getargspec_no_self
+except ImportError:
+    from collections import namedtuple
+    from scipy._lib._util import getfullargspec_no_self
 
-from scipy._lib._util import getargspec_no_self
+    ArgSpec = namedtuple('ArgSpec', ['args', 'varargs', 'keywords', 'defaults'])
+
+    def getargspec_no_self(obj):
+        return ArgSpec(*getfullargspec_no_self(obj)[:4])
 import scipy.stats as osp_stats
-from scipy.stats._distn_infrastructure import instancemethod, rv_frozen, rv_generic
+from scipy.stats._distn_infrastructure import rv_frozen, rv_generic
+try:
+    from scipy.stats._distn_infrastructure import instancemethod
+except ImportError:
+    import types
+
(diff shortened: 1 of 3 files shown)


neerajprad:
Could you merge the latest master?

fehiepsi:

neerajprad:
There's some unrelated scipy import error it seems.

fehiepsi:
@neerajprad I have added a fix to be compatible with scipy 1.5. I guess at some point, we should remove the `contrib.distributions` module. WDYT?

neerajprad:
> @neerajprad I have added a fix to be compatible with scipy 1.5. I guess at some point, we should remove the contrib.distributions module. WDYT?

Agreed, I don't think there should be any blockers to removing that.

fehiepsi:
Yeah, I'll add it to 0.3 task.

numpyro/contrib/distributions/distribution.py
```diff
@@ -9,10 +9,25 @@
...
+        return ArgSpec(*getfullargspec_no_self(obj)[:4])
 import scipy.stats as osp_stats
-from scipy.stats._distn_infrastructure import instancemethod, rv_frozen, rv_generic
+from scipy.stats._distn_infrastructure import rv_frozen, rv_generic
+try:
+    from scipy.stats._distn_infrastructure import instancemethod
+except ImportError:
+    import types
+
+    def instancemethod(func, obj, cls):
```
neerajprad, line 29: yikes, I think it is time we remove this module. :)

numpyro/contrib/distributions/distribution.py
```diff
@@ -9,10 +9,25 @@
 # Copyright (c) 2003-2019 SciPy Developers.
 # All rights reserved.
 from contextlib import contextmanager
+try:
+    from scipy._lib._util import getargspec_no_self
+except ImportError:
+    from collections import namedtuple
+    from scipy._lib._util import getfullargspec_no_self
```
neerajprad, line 16: I think if we want this module, we should ideally reimplement these utility functions in contrib, otherwise there is no guarantee that they will continue to exist in future scipy versions.

PR #642, opened 2020-06-19
Add scope handler

### LINKED ISSUE #593: add namespace feature?

fehiepsi:
Fixes #593. This is also useful when users use `scan` for prediction: they can use the same transition function but with different scope `train` and `forecast`.

I don't have a strong opinion on `namespace` vs `scope` or `::` vs `/` so I am happy to change them if reviewers feel that the current names are not conventional. The current name matches `pyro.contril.namespace.scope` behavior.

### DIFF
--- a/numpyro/handlers.py
+++ b/numpyro/handlers.py
@@ -92,6 +92,7 @@ __all__ = [
     'condition',
     'replay',
     'scale',
+    'scope',
     'seed',
     'substitute',
     'trace',
@@ -177,7 +178,8 @@ class replay(Messenger):
        -0.20584235
        >>> assert replayed_trace['a']['value'] == exec_trace['a']['value']
     """
-    def __init__(self, fn, guide_trace):
+    def __init__(self, fn=None, guide_trace=None):
+        assert guide_trace is not None
         self.guide_trace = guide_trace
         super(replay, self).__init__(fn)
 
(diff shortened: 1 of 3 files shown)


numpyro/handlers.py
```diff
@@ -380,6 +382,37 @@ def process_message(self, msg):
         msg["scale"] = self.scale if msg.get('scale') is None else self.scale * msg['scale']
 
 
+class scope(Messenger):
+    """
+    This handler prepend a prefix followed by a ``/`` to the name of sample sites.
```
neerajprad, line 387: nit: `prepends`.

neerajprad:
Looks great! I am merging this PR, we can fix the docstring in a subsequent one.

PR #648, opened 2020-06-27
Fix a bug of init_to_value strategy

fehiepsi:
In `find_valid_init_params`, we check if a partial function is `init_to_unconstrained_value` using `if init_strategy.func is _init_to_unconstrained_value`, but in `initialize_model`, `_init_to_unconstrained_value(values=unconstrained_values)` is not a partial function. This PR fixes this issue and adds a test for it.

**Edit:** Actually, this solves several issues, as explained in the comment.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -93,7 +93,7 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install jaxlib
-        pip install jax>=0.1.70
+        pip install jax
         pip install .[examples,test]
         pip freeze
     - name: Test with pytest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -93,7 +93,7 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install jaxlib
-        pip install jax
+        pip install jax==0.1.70
         pip install .[examples,test]
         pip freeze
     - name: Test with pytest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -93,7 +93,7 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install jaxlib
-        pip install jax==0.1.70
+        pip install jax>=0.1.70
         pip install .[examples,test]
         pip freeze
     - name: Test with pytest
--- a/numpyro/compat/ops.py
+++ b/numpyro/compat/ops.py
@@ -21,7 +21,7 @@ def ones(*sizes, **kwargs):
     elif len(sizes) == 1:
         if isinstance(sizes[0], (tuple, list)):
             sizes = sizes[0]
-    if not all([isinstance(s, int) for s in sizes]):
+    if not np.all([isinstance(s, int) for s in sizes]):
         raise ValueError('Invalid data type for `size` provided.')
     return jnp.ones(sizes, **kwargs)
 
@@ -32,6 +32,6 @@ def zeros(*sizes, **kwargs):
     elif len(sizes) == 1:
         if isinstance(sizes[0], (tuple, list)):
             sizes = sizes[0]
-    if not all([isinstance(s, int) for s in sizes]):
+    if not np.all([isinstance(s, int) for s in sizes]):
         raise ValueError('Invalid data type for `size` provided.')
     return jnp.ones(sizes, **kwargs)
--- a/numpyro/distributions/discrete.py
+++ b/numpyro/distributions/discrete.py
@@ -25,7 +25,7 @@
 # ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 # POSSIBILITY OF SUCH DAMAGE.
 
-import math
+import numpy as np
 
 from jax import device_put, lax
 from jax.dtypes import canonicalize_dtype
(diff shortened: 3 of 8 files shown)


numpyro/infer/util.py
```diff
@@ -365,10 +365,10 @@ def initialize_model(rng_key, model,
                                                     model_kwargs=model_kwargs)
 
     init_strategy = init_strategy if isinstance(init_strategy, partial) else init_strategy()
-    if init_strategy.func is init_to_value:
+    if (init_strategy.func is init_to_value) and not replay_model:
```
fehiepsi, line 368: I added a check `not replay_model` in case there are dynamic supports in the model. If there is dynamic support, we will use the default `init_to_value` strategy (because given a constrained value, with dynamic support, the corresponding unconstrained value might change depending on the random seed). This is just a minor point.

numpyro/distributions/discrete.py
```diff
@@ -441,7 +443,7 @@ def __init__(self):
         super(PRNGIdentity, self).__init__(event_shape=(2,))
 
     def sample(self, key, sample_shape=()):
-        return jnp.reshape(random.split(key, jnp.product(sample_shape).astype(jnp.int32)),
+        return jnp.reshape(random.split(key, math.prod(sample_shape)),
```
fehiepsi, line 446: the new version of JAX requires the argument of `jnp.product` to be an ndarray.

.github/workflows/ci.yml
```diff
@@ -93,7 +93,7 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install jaxlib
-        pip install jax
+        pip install jax==0.1.70
```
fehiepsi, line 96: I temporarily restrict JAX version because the example `ode` is failing with the latest version, probably due to [this PR](https://github.com/google/jax/pull/3562). I'll try to make a reproducible code to report it.

test/test_mcmc.py
```diff
@@ -41,8 +41,8 @@ def potential_fn(z):
     mcmc.run(random.PRNGKey(0), init_params=init_params)
     mcmc.print_summary()
     hmc_states = mcmc.get_samples()
-    assert_allclose(jnp.mean(hmc_states), true_mean, rtol=0.05)
-    assert_allclose(jnp.std(hmc_states), true_std, rtol=0.05)
+    assert_allclose(jnp.mean(hmc_states), true_mean, rtol=0.07)
+    assert_allclose(jnp.std(hmc_states), true_std, rtol=0.07)
```
fehiepsi, line 45: A precision issue in x64 mode of the latest jax version.

numpyro/compat/ops.py
```diff
@@ -21,7 +21,7 @@ def ones(*sizes, **kwargs):
     elif len(sizes) == 1:
         if isinstance(sizes[0], (tuple, list)):
             sizes = sizes[0]
-    if not all([isinstance(s, int) for s in sizes]):
+    if not np.all([isinstance(s, int) for s in sizes]):
```
fehiepsi, line 24: Here `all` is `jax.numpy.all`, which does not accept `list` input anymore.

numpyro/distributions/discrete.py
```diff
@@ -441,7 +443,7 @@ def __init__(self):
         super(PRNGIdentity, self).__init__(event_shape=(2,))
 
     def sample(self, key, sample_shape=()):
-        return jnp.reshape(random.split(key, jnp.product(sample_shape).astype(jnp.int32)),
+        return jnp.reshape(random.split(key, np.prod(sample_shape).astype(np.int32)),
```
fehiepsi, line 446: same here, `jnp.prod` does not accept `list/tuple` input anymore.

numpyro/infer/autoguide.py
```diff
@@ -183,7 +183,7 @@ def unpack_single_latent(latent):
             unpacked_samples = self._unpack_latent(latent)
             # add param sites in model
             unpacked_samples.update({k: v for k, v in params.items() if k in self.prototype_trace
-                                     and v['type'] == 'param'})
+                                     and self.prototype_trace[k]['type'] == 'param'})
```
fehiepsi, line 186: this is a bug.

numpyro/infer/util.py
```diff
@@ -365,10 +365,10 @@ def initialize_model(rng_key, model,
...
         init_values = init_strategy.keywords.get("values")
         unconstrained_values = transform_fn(inv_transforms, init_values, invert=True)
-        init_strategy = _init_to_unconstrained_value(values=unconstrained_values)
+        init_strategy = partial(_init_to_unconstrained_value, values=unconstrained_values)
```
neerajprad, line 371: Sorry, now that I'm taking another look, I don't think I understand this change. Wouldn't `_init_to_unconstrained_value(values=unconstrained_values)` return a `partial` function?

neerajprad: Okay, I see the issue. Can we just pass in `init_to_value(values=..)`?

fehiepsi: I was lazy to make init_to_unconstrained_value compatible with other init strategies, hence introducing the bug. :(

With init to uniform and init to unconstrained values, we don't have to trace the model in find_valid_initial_params to generate unconstrained values. With init to value, we need to trace the model to get the transforms, then using transforms to transform the values to unconstrained values.

neerajprad: I see. If we called `find_valid_initial_params` with `init_to_value(values=constrained_values)`, are we going to be using the constrained values when generating the initial params?

fehiepsi: Yes, we still use constrained values. But we need to trace the model in that case. This extra logic only happens when we use initialize_model with init to values.

PR #649, opened 2020-06-29
[wip] General Stein Variational Inference implementation + high-level interface for training/inference

ahmadsalim:
Hi Pyro Devs!

I have implemented a general version of Stein VI (that works with any guide/model) supporting different kernels on top of NumPyro. You can read some of the papers integrated here:
[EinStein VI](https://github.com/pyro-ppl/numpyro/files/4845182/EinsteinVI.pdf) (although the MCMC stuff will be factored out for this PR, since there are some stuff to be ironed out).

I also added a high-level interface `VI` with `train` and `predict` functions, to allow a better deep prob. prog. training and prediction flow. It will support things like callbacks and validation automatically, making it more like a simplified version of Keras.

I will work on refactoring things, tests, etc. but I hope you find the things I did useful.

Thanks!
AS


fehiepsi:
Impressive!!! It is really nice that `Stein` API is pretty similar to the current interface. <3

FYI, we are going to release version 0.3 so we want to focus on making NumPyro robust at the moment. I see that there are many new ideas in your PR. Personally, I like the `train/callback` interface. Together with other changes such as `rng_key`, `init_strategy`, `create_plates`,... I think we need to discuss them in more detail. While refactoring, could you split your PR into smaller PRs to make it easier to review? You can put new ideas in `contrib` module if it breaks some of the current interfaces.

ahmadsalim:
Glad you like the direction! I will see how I can refactor things more so that reviewing becomes easier.

Many of the changes are because I have been trying to reuse the initialization interface to reinitialize things dynamically for Stein particles. I can split them into a separate PR, but I am not sure how to otherwise handle initialization for Stein.

Suggested PR split:
* Autoguides/re-initializable guides
* initialization procedure updates
* rng_key primitive
* Stein VI with new VI interface.
* I can factor out examples separately in a PR

ahmadsalim:
I forgot to mention, that this is joint work with @OlaRonning :)!

fehiepsi:
> Suggested PR split:

Sounds good to me! I particularly have not yet understood the reason behind `reinit_param`. Hopefully, after some discussions, we can come up with a good solution for it, without having to change the interface much.

fehiepsi:
Hi @OlaRonning, after #800, I think SteinVI can be integrated as a standalone inference algorithm (many ideas here have been refactored out in separate PRs - except for high level interface for training) without changing much of the current code base. And I think there is no need for the `reinit` mechanism (by using some handlers such as `lift` or we can create a new one if needed). But I might miss something fundamentally so let's discuss more details (I would like to hear more about your plan on the integration too) in the seminar. :)

OlaRonning:
@fehiepsi Great to hear you are joining for the seminar! I'll get an overview of exactly what this integration entails, and we can discuss specific plans then.

fehiepsi:
Awesome!! Thanks, Ola!

PR #653, opened 2020-06-29
rename param_map to data in handlers

fehiepsi:
@neerajprad There is still `param_map` left in [ELBO.loss](http://num.pyro.ai/en/stable/svi.html#numpyro.infer.elbo.ELBO.loss). Should we rename it to `params` to make it consistent with `AutoGuide` or utilities in `infer/util.py`? I just notice when trying to grep `param_map` and have less opinion about it. WDYT?

### DIFF
--- a/numpyro/distributions/discrete.py
+++ b/numpyro/distributions/discrete.py
@@ -548,7 +548,7 @@ class Poisson(Distribution):
         super(Poisson, self).__init__(jnp.shape(rate), validate_args=validate_args)
 
     def sample(self, key, sample_shape=()):
-        return random.poisson(key, device_put(self.rate), shape=sample_shape + self.batch_shape)
+        return random.poisson(key, self.rate, shape=sample_shape + self.batch_shape)
 
     @validate_sample
     def log_prob(self, value):
--- a/numpyro/handlers.py
+++ b/numpyro/handlers.py
@@ -269,7 +269,12 @@ class condition(Messenger):
        >>> assert exec_trace['a']['value'] == -1
        >>> assert exec_trace['a']['is_observed']
     """
-    def __init__(self, fn=None, data=None, condition_fn=None):
+    def __init__(self, fn=None, data=None, condition_fn=None, param_map=None):
+        if param_map is not None:
+            data = param_map
+            warnings.warn(FutureWarning,
+                          "'param_map' argument is renamed to 'data'. We will remove"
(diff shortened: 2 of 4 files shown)


numpyro/infer/util.py
```diff
@@ -151,7 +151,7 @@ def potential_energy(model, model_args, model_kwargs, params):
 
 def _init_to_unconstrained_value(site=None, values={}):
     if site is None:
-        return partial(init_to_value, values=values)
+        return partial(_init_to_unconstrained_value, values=values)
```
fehiepsi, line 154: @neerajprad I think this is my original intention. My last fix is a bit misleading (though it still solves the issue)... :(

numpyro/distributions/discrete.py
```diff
@@ -550,7 +550,7 @@ def __init__(self, rate, validate_args=None):
         super(Poisson, self).__init__(jnp.shape(rate), validate_args=validate_args)
 
     def sample(self, key, sample_shape=()):
-        return random.poisson(key, device_put(self.rate), shape=sample_shape + self.batch_shape)
+        return random.poisson(key, self.rate, shape=sample_shape + self.batch_shape)
```
fehiepsi, line 553: we no longer need this in the last JAX release

numpyro/infer/util.py
neerajprad: Thanks, now I understand what's happening. :)

numpyro/handlers.py
```diff
@@ -241,12 +241,12 @@ def process_message(self, msg):
...
 class condition(Messenger):
     """
-    Conditions unobserved sample sites to values from `param_map` or `condition_fn`.
+    Conditions unobserved sample sites to values from `data` or `condition_fn`.
     Similar to :class:`~numpyro.handlers.substitute` except that it only affects
     `sample` sites and changes the `is_observed` property to `True`.
 
     :param fn: Python callable with NumPyro primitives.
-    :param dict param_map: dictionary of `numpy.ndarray` values keyed by
+    :param dict data: dictionary of `numpy.ndarray` values keyed by
```
neerajprad, line 249: What do you think about raising a `FutureWarning` with `param_map`? We can remove all of these deprecated arguments in the next release.

fehiepsi: Sure, I'll add a FutureWarning for this.

PR #654, opened 2020-06-30
Match Pyro interface

fehiepsi:
These are all points I know that diverge from Pyro:

+ the order of parameters of Pareto: due to #153 
+ init_to_prior -> init_to_sample: I believe that we started with `init_to_prior` in Pyro MCMC but Pyro autoguide uses `init_to_sample`, so we keep that name in Pyro for MCMC too. I think we need to be consistent here because some users asked for this strategy in the forum and I keep misspelling those names.
+ scale_factor -> scale: I believe the original reason why we use `scale` in NumPyro is because of the statement `super(scale, self).__init__(fn)`, which has the class name `scale` conflict with `scale` arg.
+ `MultivariateAffineTransform` is `LowerCholeskyAffine` in Pyro. I hesitate in changing this, mainly because all NumPyro transforms have the trailing `Transform`. So I really need more opinions on this change.

**Edit:** I also run `isort` because the order of some variables is changed due to the names changed. However, it also solves the orders of other places, which populate the diffs at the import sections a bit. Please ignore those changes because I didn't do anything special there. I will comment on the interface changes in the comments below.


numpyro/distributions/continuous.py
```diff
@@ -953,9 +952,9 @@ def variance(self):
 
 @copy_docs_from(Distribution)
 class Pareto(TransformedDistribution):
-    arg_constraints = {'alpha': constraints.positive, 'scale': constraints.positive}
+    arg_constraints = {'scale': constraints.positive, 'alpha': constraints.positive}
 
-    def __init__(self, alpha, scale=1., validate_args=None):
+    def __init__(self, scale, alpha, validate_args=None):
```
fehiepsi, line 957: Pareto args are reordered.

numpyro/handlers.py
```diff
@@ -381,14 +383,14 @@ class scale(Messenger):
     This is typically used for data subsampling or for stratified sampling of data
     (e.g. in fraud detection where negatives vastly outnumber positives).
 
-    :param float scale_factor: a positive scaling factor
+    :param float scale: a positive scaling factor
```
fehiepsi, line 386: `scale_factor` -> `scale`

numpyro/distributions/transforms.py
```diff
@@ -335,7 +335,7 @@ def log_abs_det_jacobian(self, x, y, intermediates=None):
         return x[..., -n:].sum(-1)
 
 
-class MultivariateAffineTransform(Transform):
+class LowerCholeskyAffine(Transform):
```
fehiepsi, line 338: `MultivariateAffineTransform` -> `LowerCholeskyAffine`

numpyro/infer/initialization.py
```diff
@@ -30,7 +30,7 @@ def init_to_median(site=None, num_samples=15):
             return init_to_uniform(site)
 
 
-def init_to_prior(site=None):
+def init_to_sample(site=None):
```
fehiepsi, line 33: init_to_prior -> init_to_sample

fritzo: This seems reasonable since `init_to_median` is an alternative way to use the prior. I guess we could be even more verbose with `init_to_prior_sample`, `init_to_prior_median`, but I'm not sure other options. Maybe `init_to_generated`?

numpyro/distributions/continuous.py
fritzo: One improvement I've been starting to make in Pyro is to clearly separate args from kwargs via `, *,`. This makes it safer to make future changes to kwargs, since kwargs after `, *,` can be specified only by name, not by position. We can't do this pervasively in Pyro because Pyro 1.0 supported Python 2. However since NumPyro is not yet 1.0 you might consider pervasively separating args from kwargs. E.g. you could require `validate_args` to be specified by name
```py
def __init__(self, scale, alpha, *, validate_args=None):
    ...
```
Just a thought 🙂

fritzo:
Looks great, but I don't know enough to approve!

neerajprad:
Another thing that I was reminded of was renaming `posterior_samples` to `samples` in `Predictive` since it is used more generally, but we should make that change to both Pyro and NumPyro and not diverge.

numpyro/distributions/continuous.py
fehiepsi: I think this is a good idea. Let me create a separate issue to track down the progress for it.

fehiepsi: Thanks for your input!

fehiepsi:
> renaming posterior_samples to samples

Sounds reasonable to me! Users also use Predictive to get samples from priors. I think it is better to make a change in Pyro first to grasp other opinions, then we can make a change in NumPyro.

PR #657, opened 2020-07-01
Update forecasting tutorial using scan primitive

fehiepsi:
Suggested by @neerajprad!


neerajprad:
Thanks, this will be really useful as a reference example for everyone!

Just some minor nits:

> , we have use condition handler here. 

use --> used

> The reason is we also want to use this model for forecasting. In forecasting, future values of y are non-observable, so obs=y[t] will become invalid. 

"will become invalid" - would suggest making it more specific, i.e. does this raise an index out of bounds error?

> Using condition, when the length of scan is larger than the length of the conditioned/observed site, non-observed values will be sampled from the distribution of that site.

unobserved values?

neerajprad:
On another note, I think we need an additional indent after `doctest...` for the [scan documentation](http://num.pyro.ai/en/latest/primitives.html?highlight=scan#scan). I'm not sure if the doctest is running, but it is showing up in the documentation.

fehiepsi:
Thanks Neeraj, I'll address them soon.

fehiepsi:
@neerajprad I just incorporate your suggestions.

Issue #645, opened 2020-06-22
Allow for calculation of log_likelihood by chain?


vanAmsterdam:
When checking inference, diagnosing log_likelihoods by chain could provide some insight, e.g. if the posterior has different modes. Currently passing samples that are `group_by_chain`-ed to `log_likelihood` will lead to broadcasting errors. Maybe we can add something like this:

```python
import jax, numpyro
from jax import numpy as np, random, vmap
from numpyro import distributions as dist, sample
from numpyro.infer.mcmc import MCMC, NUTS
from numpyro.infer.util import log_likelihood
from functools import partial
numpyro.set_host_device_count(2)

rkeys = random.split(random.PRNGKey(0), 2)
X     = random.normal(rkeys[0], shape=(100,))
Y     = 0.5 * X + random.normal(rkeys[1], shape=(100,))
def m(X, Y = None):
    a = sample('a', dist.Normal())
    b = sample('b', dist.Normal())
    with numpyro.plate('obs', X.shape[0]):
        mu_hat = a + b * X
        sample('obs_Y', dist.Normal(mu_hat, 1), obs=Y)

mcmckey = random.PRNGKey(1)

mcmc = MCMC(NUTS(m), 100, 250, num_chains=2)
mcmc.run(mcmckeys[0], X, Y)

smps  = mcmc.get_samples(group_by_chain=True)
# lls  = log_likelihood(m, smps, X, Y) # -> throws an error

def log_likelihood2(model, posterior_samples, grouped_by_chain=False, *args, **kwargs):
    if grouped_by_chain:
        llfun = lambda samples, model, *args, **kwargs: log_likelihood(model, samples, *args, **kwargs)
        return vmap(partial(llfun, model=model, *args, **kwargs))(posterior_samples)
    else:
        return log_likelihood(model, posterior_sampels, *args, **kwargs)

lls = log_likelihood2(m, smps, grouped_by_chain=True, X=X, Y=Y)

print(lls['obs_Y'].shape) # -> (2, 250, 100)
```

the only thing I have been struggling with still is to keep the possibility to pass *args to the model call...

fehiepsi:
[...]

Issue #639, opened 2020-06-17
Allow arbitrary sample_sample in Predictive


fehiepsi:
Currently, we assume the sample_shape in Predictive is `(num_samples,)`, but it might be more convenient to allow arbitrary shapes here. Some motivations are:
+ We can apply Predictive on samples with `group_by_chain=True`. For models with discrete latent variables, we don't replay the model to collect the value of discrete sites so we need to use Predictive to collect those samples. Of course, we can do `{k: v.reshape(...)}` but it is a bit verbose.
+ I found that using Predictive is more convenient than using handlers to collect trace and get sites' values. I can recall that this question is already raised in the forum. Actually, this is the main motivation that I raise this issue. In my gmm example, I just want to draw the prediction for "one" cluster means/scales sample.

Cons:
+ We might change `num_samples` to `sample_shape` or adding a new keyword `sample_shape` for this purpose.

What do you think, @neerajprad?

neerajprad:
I am not opposed to it, but would like to understand the use case better. For case 1, we should be able to flatten the chain dim and use `Predictive`, right? For the second case, could you point me to the code so that I understand what the issues are.

fehiepsi:
Yes, you are right that this is not an issue. I just found it more convenient to avoid `reshape` stuff. Here is my code for gmm (there is nothing special with it), predictions will have the leading dimension with size 1. If we allow something like `sample_shape=()`, the output won't have that leading dimension. At the time I write this, I found that `batch_ndim` might be a better argument (by default, it will be 1; with chains, it will be 2; with single sample, it will be 0).

```python
K = 3
D = 4

def gmm(data=None, N=None):
    if N is None:
        assert data is not None
        N = data.shape[0]

    with numpyro.plate("num_components", K):
        means = numpyro.sample("means", dist.Normal().expand([D]).to_event(1))
        scale_trils = numpyro.sample("scale_trils", dist.LKJCholesky(D))

    mixture_weights = numpyro.sample("phi", dist.Dirichlet(jnp.ones(K)))
    with numpyro.plate("num_observations", N):
        z = numpyro.sample("assignments", dist.Categorical(mixture_weights))
        numpyro.sample("obs", dist.MultivariateNormal(means[z], scale_tril=scale_trils[z]), obs=data)

nuts_kernel = NUTS(gmm)
mcmc = MCMC(nuts_kernel, num_warmup=1000, num_samples=1000)
mcmc.run(random.PRNGKey(0), data)
[...]

PR #660, opened 2020-07-02
Add batch_ndims argument in Predictive and log_likelihood

fehiepsi:
Addresses #645 and #639.

This PR also adds a utility `soft_vmap` to avoid duplicated code in Predictive and log_likelihood. It might also be useful for other purposes. @martinjankowiak you might find `soft_vmap`, which is adapted from your code, useful for your experiment. :D

### DIFF
--- a/numpyro/distributions/transforms.py
+++ b/numpyro/distributions/transforms.py
@@ -372,7 +372,7 @@ class LowerCholeskyTransform(Transform):
 class MultivariateAffineTransform(LowerCholeskyAffine):
     def __init__(self, loc, scale_tril):
         warnings.warn("MultivariateAffineTransform is renamed to LowerCholeskyAffine.",
-                      DeprecationWarning)
+                      FutureWarning)
         super().__init__(loc, scale_tril)
 
 
--- a/numpyro/distributions/transforms.py
+++ b/numpyro/distributions/transforms.py
@@ -371,8 +371,8 @@ class LowerCholeskyTransform(Transform):
 
 class MultivariateAffineTransform(LowerCholeskyAffine):
     def __init__(self, loc, scale_tril):
-        warnings.warn(DeprecationWarning,
-                      "MultivariateAffineTransform is renamed to LowerCholeskyAffine.")
+        warnings.warn("MultivariateAffineTransform is renamed to LowerCholeskyAffine.",
+                      DeprecationWarning)
         super().__init__(loc, scale_tril)
 
 
--- a/numpyro/infer/initialization.py
+++ b/numpyro/infer/initialization.py
@@ -33,7 +33,7 @@ def init_to_median(site=None, num_samples=15):
 
 def init_to_prior(site=None):
     warnings.warn("`init_to_prior` strategy is renamed to `init_to_sample`.",
-                  DeprecationWarning)
+                  FutureWarning)
     return init_to_sample(site=site)
 
 
--- a/numpyro/infer/initialization.py
+++ b/numpyro/infer/initialization.py
@@ -32,8 +32,8 @@ def init_to_median(site=None, num_samples=15):
 
 
 def init_to_prior(site=None):
-    warnings.warn(DeprecationWarning,
-                  "`init_to_prior` strategy is renamed to `init_to_sample`.")
+    warnings.warn("`init_to_prior` strategy is renamed to `init_to_sample`.",
+                  DeprecationWarning)
     return init_to_sample(site=site)
 
 
--- a/numpyro/infer/util.py
+++ b/numpyro/infer/util.py
@@ -453,8 +453,8 @@ class Predictive(object):
         prototype_site = batch_shape = batch_size = None
         for name, sample in posterior_samples.items():
             if batch_shape is not None and sample.shape[:batch_ndims] != batch_shape:
-                raise ValueError(f"Batch shapes at site {name} and {prototype_site}"
-                                 f"should be the same, but got"
+                raise ValueError(f"Batch shapes at site {name} and {prototype_site} "
+                                 f"should be the same, but got "
                                  f"{sample.shape[:batch_ndims]} and {batch_shape}")
             else:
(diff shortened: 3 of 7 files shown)


numpyro/util.py
```diff
@@ -298,3 +298,29 @@ def unravel_pytree(arr):
         return tree_unflatten(treedef, unravel_list(arr))
 
     return flat, unravel_pytree
+
+
+def soft_vmap(fn, xs, batch_shape, chunk_size=None):
```
martinjankowiak, line 303: @fehiepsi could you please add a docstring for this? i'd like to better understand what it's doing...

fehiepsi: Sure, I'll add. This basically computes `fn(xs)` by vectorizing `fn(x)` over `batch_shape` (i.e. `xs.shape = batch_shape + 'x.shape'`). The optional `chunk_size` allows the job taken in chunk: by spliting `xs` into chunks `xs_1`,..., `xs_n`, computing `fn(xs_1)`, ... `fn(xs_n)` using `vmap`, and finally merging the results. `chunk_size=None` means that there is only 1 chunk, which is the same as `vmap(fn)(x)`. `chunk_size=1` means that we don't use `vmap` at all, which is the same as `map(fn)(x)`.

The difference w.r.t. your script is `batch_shape` argument, which has length 1 in your script but has arbitrary length here (to allow calculating `num_chains x num_samples x ...`, which has length 2, forms).

numpyro/util.py
```diff
@@ -298,3 +298,42 @@ def unravel_pytree(arr):
         return tree_unflatten(treedef, unravel_list(arr))
 
     return flat, unravel_pytree
+
+
+def soft_vmap(fn, xs, batch_ndims=1, chunk_size=None):
+    """
+    Vectorizing map a function `fn` over `batch_ndims` leading axes of `xs`.
+    The job is splitted into chunks to reduce memory usages.
```
neerajprad, line 306: nits: 
Vectorizing map that maps a function `fn` ..
splitted --> split

Maybe we can replace the last line with: This uses `jax.vmap` over smaller chunks of the batch dimension to keep memory usage constant.

test/test_infer_util.py
```diff
@@ -130,14 +131,21 @@ def test_prior_predictive():
     assert predictive_samples["obs"].shape == (100,) + data.shape
 
 
-def test_log_likelihood():
+@pytest.mark.parametrize('batch_shape', [(), (100,), (2, 50)])
+def test_log_likelihood(batch_shape):
     model, data, _ = beta_bernoulli()
-    samples = Predictive(model, return_sites=["beta"], num_samples=100)(random.PRNGKey(1))
-    loglik = log_likelihood(model, samples, data)
+    samples = Predictive(model, return_sites=["beta"], num_samples=200)(random.PRNGKey(1))
```
neerajprad, line 137: What will happen if we provide `batch_ndims=2` here?

numpyro/infer/util.py
```diff
@@ -488,21 +514,48 @@ def get_samples(self, rng_key, *args, **kwargs):
...
-        model_trace = trace(substitute(model, samples)).get_trace(*args, **kwargs)
+        substituted_model = substitute(model, samples) if isinstance(samples, dict) else model
+        model_trace = trace(substituted_model).get_trace(*args, **kwargs)
         return {name: site['fn'].log_prob(site['value']) for name, site in model_trace.items()
                 if site['type'] == 'sample' and site['is_observed']}
 
-    return vmap(single_loglik)(posterior_samples)
+    prototype_site = batch_shape = None
+    for name, sample in posterior_samples.items():
+        if batch_shape is not None and sample.shape[:batch_ndims] != batch_shape:
```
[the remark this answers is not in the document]
neerajprad: Sorry, outdated comment. This is being inferred by the first element in the dict, so should be fine.

numpyro/util.py
fehiepsi: I just add the docstring.

test/test_infer_util.py
fehiepsi: I believe the output will have shape `(1, 200, ...)` but let me test it.

fehiepsi: That's true, I just add a test for that in `test_prior_predictive`.

numpyro/distributions/transforms.py
```diff
@@ -371,8 +371,8 @@ def log_abs_det_jacobian(self, x, y, intermediates=None):
 
 class MultivariateAffineTransform(LowerCholeskyAffine):
     def __init__(self, loc, scale_tril):
-        warnings.warn(DeprecationWarning,
-                      "MultivariateAffineTransform is renamed to LowerCholeskyAffine.")
+        warnings.warn("MultivariateAffineTransform is renamed to LowerCholeskyAffine.",
+                      DeprecationWarning)
```
neerajprad, line 375: Sorry, I might have missed this earlier.`DeprecationgWarning` will be swallowed by application code. We should use `FutureWarning` instead. See https://docs.python.org/3/library/warnings.html#warning-categories.

numpyro/infer/initialization.py
```diff
@@ -32,8 +32,8 @@ def init_to_median(site=None, num_samples=15):
 
 
 def init_to_prior(site=None):
-    warnings.warn(DeprecationWarning,
-                  "`init_to_prior` strategy is renamed to `init_to_sample`.")
+    warnings.warn("`init_to_prior` strategy is renamed to `init_to_sample`.",
```
neerajprad, line 35: same here about DeprecationWarning.

numpyro/util.py
fehiepsi: Thanks for suggesting!

numpyro/distributions/transforms.py
fehiepsi: Thanks! This explains why I didn't see the warnings when I used `predictive.get_samples(...)`. :D

Issue #557, opened 2020-03-09
Add example usage of numpyro.infer.SA


martinjankowiak:
perhaps to baseball.py?

PR #664, opened 2020-07-03
Add Sample Adaptive MCMC algo to baseball example

fehiepsi:
Addresses #557.

I have changed the behavior of SA a bit to recompute loc/scale at each step because I think this is the better default behavior.  The reason is with SA, we need a lot of samples (100,000 - 1,000,000) so after a while, `scale` term will diverge from the true value. It is nice to have a config setting to control how often we recompute, but for now, it seems not necessary.

I also resolve the `nan` issue when the initial scale is NaN.

### DIFF
--- a/numpyro/infer/mcmc.py
+++ b/numpyro/infer/mcmc.py
@@ -770,7 +770,7 @@ def _sa(potential_fn=None, potential_fn_gen=None):
             if cov.shape == ():  # JAX returns scalar for 1D input
                 cov = cov.reshape((1, 1))
             cholesky = jnp.linalg.cholesky(cov)
-            scale = jnp.where(jnp.isnan(cholesky), scale, cholesky)
+            scale = jnp.where(jnp.any(jnp.isnan(cholesky)), scale, cholesky)
         else:
             scale = jnp.std(zs, 0)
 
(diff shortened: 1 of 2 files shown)


numpyro/infer/mcmc.py
```diff
@@ -763,6 +761,19 @@ def sample_kernel(sa_state, model_args=(), model_kwargs=None):
         if potential_fn_gen:
             pe_fn = potential_fn_gen(*model_args, **model_kwargs)
         zs, pes, loc, scale = sa_state.adapt_state
+        # we recompute loc/scale after each iteration to avoid precision loss
```
martinjankowiak, line 764: what's the resulting change in the computational complexity?

numpyro/infer/mcmc.py
```diff
@@ -763,6 +761,19 @@ def sample_kernel(sa_state, model_args=(), model_kwargs=None):
...
+        # XXX: consider to expose a setting to do this job pediodically
```
martinjankowiak, line 765: typo: pediodically

numpyro/infer/mcmc.py
fehiepsi: It is still O(Nxd^2) where `N` is `adapt_state_size` (size of each adaptive state) and `d` is the latent dimension. The computation here is O(Nd^2) because computing covariance matrix took O(Nxd^2) while computing cholesky term is O(d^3). In a later step, we compute `log_prob` of mvn distribution with `batch_shape = (N + 1,)`, which has complexity `O(Nd^2)`. I haven't looked into details but I think `_get_proposal_loc_and_scale` also took O(Nxd^2) to compute the scale of each combination of proposals. Anyway, I believe recomputing here does not add more complexity to the overall computation.

Issue #665, opened 2020-07-03
Behavior of `condition` handler is different from `pyro.condition`


fehiepsi:
In Pyro, we allow `condition` on observed sites, but this is not allowed in NumPyro. Should we relax this constraint? cc @neerajprad 

(I think this issue will block #638: we can not compose `do(condition(...))`)

fritzo:
@eb8680 ?

eb8680:
I don't see why this would block #638, can you explain?

fehiepsi:
With `condition(model, {'z': 1})`, `z` will be come an observed site. Then in `do.process_msg(z_site)`, we do `apply_stack(z_site.copy())`, which will do `condition` again on the observed `z_site.copy()`.

fehiepsi:
Actually, this constraint is [relaxed in Pyro](https://github.com/pyro-ppl/pyro/pull/2157/files#diff-8b4f7d602fb71525c816c41901267f78) to support `do` handler. :D

PR #666, opened 2020-07-06
relax condition handler

fehiepsi:
Addresses #665 and unblocks #638.


eb8680:
Looks like there's a test you'll need to update as well

fehiepsi:
> there's a test you'll need to update

Yeah, I just fix it. Thanks!

Issue #663, opened 2020-07-03
TransformedDistribution broken in MCMC?


vanAmsterdam:
A NegativeHalfNormal distribution, defined with an AffineTransform gives negative samples when sampling directly, but when used in mcmc context, only positive samples. This works properly in the pip-version of numpyro 0.2.4 but not in the development version

```python

from jax import numpy as jnp, random
import numpyro
from numpyro.distributions import constraints, Normal, HalfNormal, TransformedDistribution
from numpyro.distributions.transforms import AffineTransform
from numpyro.infer.mcmc import NUTS, MCMC

class NegativeHalfNormal(TransformedDistribution):
    arg_constraints = {'scale': constraints.positive}

    def __init__(self, scale=1., validate_args=None):
        base_dist = HalfNormal(scale)
        self.scale = base_dist.scale
        super(NegativeHalfNormal, self).__init__(base_dist, AffineTransform(0, -1), validate_args=validate_args)

    @property
    def mean(self):
        return -self.base_dist.mean

    @property
    def variance(self):
        return self.base_dist.variance

NegativeHalfNormal().variance

simkey = random.PRNGKey(0)
X      = Normal().sample(simkey, (100,)) - 0.25

def model(X):
    mu = numpyro.sample('mu', NegativeHalfNormal(2.5))
    with numpyro.plate('obs', X.shape[0]):
        numpyro.sample('X', Normal(mu, 1), obs=X)

smps = NegativeHalfNormal().sample(random.PRNGKey(0), (100,))
print(smps.mean()) # -0.748

mcmc_key = random.PRNGKey(12345)
kernel = NUTS(model)
mcmc   = MCMC(kernel, num_warmup=250, num_samples=750)
mcmc.run(mcmc_key, X)
mcmc.print_summary() # mu = 0.04

```

fehiepsi:
@vanAmsterdam Currently, AffineTransform only works for a positive `scale`. What NumPyro 0.24 does can be translated to the code
```python
def model(X):
    from numpyro.infer.reparam import TransformReparam
    with numpyro.handlers.reparam(config={'mu': TransformReparam()}):
        mu = numpyro.sample('mu', NegativeHalfNormal(2.5))

mcmc.print_summary(exclude_deterministic=False)
```
in the `master` branch (see the eight school example in README). Does it serve your purpose?

[...]

PR #668, opened 2020-07-07
Improve compiling time for the option `find_heuristic_step_size=True`

fehiepsi:
Addresses #585.

This optimization has been low-priority to me because I thought that `find_heuristic_step_size=True` is not important to use. But recently, @vanAmsterdam mentioned in #663 that our MCMC might have some regressions w.r.t. the last release. Though I suspected that random seed plays a role there, while waiting for the feedback in #663, it is still good to optimize compiling if users want to enable that option.

Tested on https://github.com/pyro-ppl/numpyro/issues/543, with `find_heuristic_step_size=True`, compiling time is reduced from almost 1 minute (master branch) to 12s (this PR). Note that only took about 1-2s with `find_heuristic_step_size=False`.

### DIFF
--- a/numpyro/infer/hmc_util.py
+++ b/numpyro/infer/hmc_util.py
@@ -213,7 +213,7 @@ def velocity_verlet(potential_fn, kinetic_fn):
 
 
 def find_reasonable_step_size(potential_fn, kinetic_fn, momentum_generator,
-                              init_step_size, inverse_mass_matrix, position, rng_key):
+                              init_step_size, inverse_mass_matrix, z_info, rng_key):
     """
     Finds a reasonable step size by tuning `init_step_size`. This function is used
     to avoid working with a too large or too small step size in HMC.
@@ -228,7 +228,8 @@ def find_reasonable_step_size(potential_fn, kinetic_fn, momentum_generator,
     :param momentum_generator: A generator to get a random momentum variable.
     :param float init_step_size: Initial step size to be tuned.
     :param inverse_mass_matrix: Inverse of mass matrix.
-    :param position: Current position of the particle.
+    :param tuple z_info: A tuple containing current position of the particle,
+        its potential energy, and the corresponding gradients.
     :param jax.random.PRNGKey rng_key: Random key to be used as the source of randomness.
     :return: a reasonable value for step size.
     :rtype: float
@@ -239,8 +240,9 @@ def find_reasonable_step_size(potential_fn, kinetic_fn, momentum_generator,
     target_accept_prob = jnp.log(0.8)
 
     _, vv_update = velocity_verlet(potential_fn, kinetic_fn)
-    z = position
-    potential_energy, z_grad = value_and_grad(potential_fn)(z)
+    z, potential_energy, z_grad = z_info
+    if potential_energy is None or z_grad is None:
+        potential_energy, z_grad = value_and_grad(potential_fn)(z)
     finfo = jnp.finfo(get_dtype(init_step_size))
 
     def _body_fn(state):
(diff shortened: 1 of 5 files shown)


numpyro/infer/hmc_util.py
```diff
@@ -355,9 +357,10 @@ def warmup_adapter(num_adapt_steps, find_reasonable_step_size=None,
     adaptation_schedule = jnp.array(build_adaptation_schedule(num_adapt_steps))
     num_windows = len(adaptation_schedule)
 
-    def init_fn(z, rng_key, step_size=1.0, inverse_mass_matrix=None, mass_matrix_size=None):
+    def init_fn(z_info, rng_key, step_size=1.0, inverse_mass_matrix=None, mass_matrix_size=None):
         """
-        :param z: Initial position of the integrator.
+        :param tuple z_info: A tuple containing the initial position of the integrator, its
```
neerajprad, line 362: nit: can we just reuse `IntegratorState` namedtuple for `z_info` and define `r` as None? Also, not important but we can also define a defaults argument in `NamedTuple` and set it to `None`. (See [workaround](https://stackoverflow.com/questions/11351032/namedtuple-and-default-values-for-optional-keyword-arguments) for python 3.6).

numpyro/infer/hmc_util.py
```diff
@@ -390,25 +393,26 @@ def init_fn(z, rng_key, step_size=1.0, inverse_mass_matrix=None, mass_matrix_siz
...
         return HMCAdaptState(step_size, inverse_mass_matrix, mass_matrix_sqrt,
                              ss_state, mm_state, window_idx, rng_key)
 
-    def update_fn(t, accept_prob, z, state):
+    def update_fn(t, accept_prob, z_info, state):
         """
         :param int t: The current time step.
         :param float accept_prob: Acceptance probability of the current trajectory.
-        :param z: New position drawn at the end of the current trajectory.
+        :param z_info: A tuple contains new position drawn at the end of the current trajectory,
```
neerajprad, line 414: nit: can we reuse this z_info description in all the remaining docstrings too?

numpyro/infer/hmc_util.py
fehiepsi: Thanks, that makes sense to me. I've made the change.

numpyro/infer/hmc_util.py
fehiepsi: I have change `z_info` tuple to IntegratorState, so we can simply say "the new integrator state" here

test/test_hmc_util.py
```diff
@@ -226,7 +226,8 @@ def potential_fn(q):
     fn = (jit(find_reasonable_step_size, static_argnums=(0, 1, 2))
           if jitted else find_reasonable_step_size)
     rng_key = random.PRNGKey(0)
-    step_size = fn(potential_fn, kinetic_fn, p_generator, init_step_size, m_inv, q, rng_key)
+    step_size = fn(potential_fn, kinetic_fn, p_generator, init_step_size, m_inv,
+                   (q, None, None), rng_key)
```
neerajprad, line 230: These tests also need to be modified, I think.

vanAmsterdam:
> Addresses #585.
> 
> 
> 
> This optimization has been low-priority to me because I thought that `find_heuristic_step_size=True` is not important to use. But recently, @vanAmsterdam mentioned in #663 that our MCMC might have some regressions w.r.t. the last release. Though I suspected that random seed plays a role there, while waiting for the feedback in #663, it is still good to optimize compiling if users want to enable that option.
> 
> 
> 
> Tested on https://github.com/pyro-ppl/numpyro/issues/543, with `find_heuristic_step_size=True`, compiling time is reduced from almost 1 minute (master branch) to 12s (this PR). Note that only took about 1-2s with `find_heuristic_step_size=False`.

the regressions are probably due to a bug in my own code

test/test_hmc_util.py
fehiepsi: Thanks Neeraj! I just fix them and will merge if tests pass due to a typo (missing a comma) at setup file, which removes the dependency of numpyro on `jax`.

PR #669, opened 2020-07-07
tweak sparse regression example

martinjankowiak:
this brings the model into exact alignment with the reference (wasn't quite right before)

### DIFF
--- a/examples/sparse_regression.py
+++ b/examples/sparse_regression.py
@@ -185,7 +185,6 @@ def sample_theta_space(X, Y, active_dims, msq, lam, eta1, xisq, c, sigma):
 
     k_xx = kernel(kX, kX, eta1, eta2, c) + sigma ** 2 * jnp.eye(N)
     L = cho_factor(k_xx, lower=True)[0]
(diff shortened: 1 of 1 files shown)


examples/sparse_regression.py
```diff
@@ -43,7 +44,7 @@ def dot(X, Z):
 
 
 # The kernel that corresponds to our quadratic regressor.
-def kernel(X, Z, eta1, eta2, c, jitter=1.0e-6):
+def kernel(X, Z, eta1, eta2, c, jitter=1.0e-4):
```
martinjankowiak, line 47: need more jitter because the prior now has more mass at sigma=0

fehiepsi: Did you face Cholesky issues or `nan` during sampling? I think we addressed it by setting `energy=inf` but if it happens with this example, I need to take a look again.

martinjankowiak: no sampling was fine; it's more that some downstream predictive stuff was less stable

PR #670, opened 2020-07-08
Add less_than constraint

fehiepsi:
Addresses #663. I was wrong there that Pyro didn't have this constraint.

This PR adds `less_than` constraint and its corresponding bijective. For that purpose, I also support the case `scale < 0` in AffineTransform in case `scale` is not a tracer. If `scale` is a tracer or not `< 0`, we still assume that `scale > 0` because, e.g, there is no defined codomain for `ComposedTransform([ExpTransform(), AffineTransform(0, scale=jnp.array([-1, 1]), domain=constraints.positive)])`

cc @vanAmsterdam After this, you don't need to change your code to use `reparam` handler.

### DIFF
--- a/numpyro/distributions/transforms.py
+++ b/numpyro/distributions/transforms.py
@@ -20,6 +20,7 @@ from numpyro.distributions.util import (
     sum_rightmost,
     vec_to_tril_matrix
 )
+from numpyro.util import not_jax_tracer
 
 __all__ = [
     'biject_to',
@@ -79,7 +80,10 @@ class AbsTransform(Transform):
 
 
 class AffineTransform(Transform):
-    # TODO: currently, just support scale > 0
+    """
+    .. note:: When `scale` is a JAX tracer, we always assume that `scale > 0`
+        when calculating `codomain`.
+    """
     def __init__(self, loc, scale, domain=constraints.real):
         self.loc = loc
         self.scale = scale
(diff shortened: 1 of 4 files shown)


numpyro/distributions/transforms.py
```diff
@@ -92,10 +96,24 @@ def codomain(self):
...
+                return constraints.less_than(self(self.domain.lower_bound))
+            # we suppose scale > 0 for any tracer
+            else:
+                return constraints.greater_than(self(self.domain.lower_bound))
+        elif isinstance(self.domain, constraints.less_than):
+            if not_jax_tracer(self.scale) and jnp.all(self.scale < 0):
+                return constraints.greater_than(self(self.domain.upper_bound))
+            # we suppose scale > 0 for any tracer
+            else:
+                return constraints.less_than(self(self.domain.upper_bound))
```
fritzo, line 109: I like that in NumPyro you added a customizable `domain` arg. That might also be helpful in PyTorch, especially for `TransformedDistribution`. When we discussed it in the past we decided that the constraint transform logic would be overly complex.

Indeed the logic here is getting quite complex, e.g. you need to make a `> 0` assumption. Do you think it would be simpler to add a customizable `codomain` arg as well? Maybe similarly `TransformedDistribution` should take a `.support` arg in case it cannot be computed directly. I know I've created subclasses of `TransformedDistribution` merely to add a proper `.support`.

fehiepsi: > created subclasses of TransformedDistribution merely to add a proper .support

I also faced this issue in the past. I agree that having arg `support` would be simpler in situations such as `TransformedDist(LogNormal(), ExpTransform())` because `codomain` of `ExpTransform` is fixed to `positive`. But I am also happy with the logic
```
    @property
    def support(self):
        domain = self.base_dist.support
        for t in self.transforms:
            t.domain = domain
            domain = t.codomain
        return domain
```
in `TransformedDistribution`, which should cover most usage cases.

Similarly, having a customizable `codomain` is convenient but we still need this logic in case users didn't provide `codomain`. IIRC, we have `domain` arg here so that users can provide a correct sequence of `domain/codomain` in TransformedDistribution in the constructor (for this, having `codomain` arg, as you said, would be much simpler), but later, we used the above logic to compute `support` because it is nontrivial for users to handle that `domain/codomain` sequence.

fehiepsi:
> Is a JAX tracer any non-static tensor?

You can see its docs [here](https://github.com/google/jax/blob/master/docs/faq.rst#different-kinds-of-jax-values). In my interpretation, it is a temporary object that does not hold an actual value (the actual value is computed in XLA, llvm, or something).

numpyro/distributions/transforms.py
fritzo: Thanks for explaining!

PR #672, opened 2020-07-08
Examples for mixture models

fehiepsi:
In this PR, I want to create an example for the new enum implementation. I found that [this reference](https://www.aclweb.org/anthology/Q18-1040.pdf) provides many models with different structures, which will be useful for users who are new to `enum` and `plate`.

Tasks
- [x] Add Multinomial model
- [x] Add Dawid-Skene model
- [x] Add Multi-Annotator Competence Estimation (MACE) model
- [x] Add Hierarchical Dawid-Skene model
- [x] Add Item Difficulty model
- [x] Add Logistic Random Effects model
- [x] Add a script for doing inference using the above models
- [x] Mention that label switching is not addressed and use LocScaleReparam

What will be omitted
- Post-processing samples to deal with "label switching" issue (mentioned in Section 6.2 of the above reference). I will create a separate issue to get help from readers who know more about how to do this task.


examples/annotation.py
```diff
@@ -0,0 +1,339 @@
...
+    pi = numpyro.sample("pi", dist.Dirichlet(jnp.ones(num_classes)))
+
+    with numpyro.plate("item", num_items, dim=-2):
+        c = numpyro.sample("c", dist.Categorical(pi))
+
+        # here we use Vindex to allow broadcasting for the second index `c`
+        # ref: http://num.pyro.ai/en/latest/utilities.html#numpyro.contrib.indexing.vindex
+        with numpyro.plate("position", num_positions):
+            numpyro.sample(
+                "y", dist.Categorical(Vindex(beta)[positions, c, :]), obs=annotations
```
fehiepsi, line 122: @fritzo `Vindex` truly helps here.

eb8680:
These are great, we should port them to Pyro as well.

examples/annotation.py
```diff
@@ -0,0 +1,339 @@
...
+         1, 1, 1, 1, 2, 2, 1, 2, 2, 1, 1, 2, 1, 1, 1,
+         1, 3, 1, 2, 2, 3, 2, 3, 2, 1, 1, 1, 2, 1, 2],
+    ]).T
+    # we minus 1 because in Python, the first index is 0
+    return positions - 1, annotations - 1
+
+
+def multinomial(annotations):
+    """
+    This model corresponses to the plate diagram in Figure 1 of reference [1].
```
martinjankowiak, line 84: here and below; typo: corresponses -> corresponds

numpyro/contrib/funsor/infer_util.py
```diff
@@ -0,0 +1,108 @@
...
+        return {}
+
+    return infer_config(fn, config_fn)
+
+
+def log_density(model, model_args, model_kwargs, params):
+    """
+    Similar to :func:`numpyro.infer.util.log_density` but works for models
+    with discrete latent variables. Internally, this uses :mod:`funsor`
+    to evalutate the joint log probability.
```
martinjankowiak, line 63: typo: evalutate

fehiepsi: oops... good catch!

Issue #675, opened 2020-07-09
Support LocScaleReparam


fehiepsi:
I found that writing TransformedReparam is too verbose for several examples, e.g. the eight school example in README or [this example](https://github.com/pyro-ppl/numpyro/blob/b06be2b46c8b026e365c9504d2b6e4956ea268e8/examples/annotation.py#L218) in #672. So it would be nice to port LocScaleReparam in Pyro to here to simplify those programs.

PR #677, opened 2020-07-10
Port LocScaleReparam

fehiepsi:
Addresses #675.

 This port Pyro [LocScaleRepram](https://github.com/pyro-ppl/pyro/blob/dev/pyro/infer/reparam/loc_scale.py) to NumPyro with the following differences:
+ array input for `center` parameter is not supported.
+ an extra logic for `expand` is needed because in NumPyro, we expand distributions under `plate` using the wrapper `ExpandedDistribution`. To simplify the implementation, I made `Independent(...).expand(batch_shape)` returns an `Independent(ExpandedDistribution(...))` distribution instead of an `ExpandedDistribution`.

### DIFF
--- a/numpyro/infer/reparam.py
+++ b/numpyro/infer/reparam.py
@@ -7,8 +7,8 @@ import jax.numpy as jnp
 
 import numpyro
 import numpyro.distributions as dist
-from numpyro.distributions import biject_to
-from numpyro.distributions.util import sum_rightmost
+from numpyro.distributions import biject_to, constraints
+from numpyro.distributions.util import is_identically_one, sum_rightmost
 from numpyro.infer.autoguide import AutoContinuous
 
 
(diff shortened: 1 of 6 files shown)


numpyro/infer/reparam.py
```diff
@@ -26,6 +26,93 @@ def __call__(self, name, fn, obs):
...
+    :param float centered: optional centered parameter. If None (default) learn
+        a per-site per-element centering parameter in ``[0,1]``. If 0, fully
+        decenter the distribution; if 1, preserve the centered distribution
+        unchanged.
+    :param shape_params: list of additional parameter names to copy unchanged from
+        the centered to decentered distribution.
+    :type shape_params: tuple or list
+    """
+    def __init__(self, centered=None, shape_params=()):
+        assert centered is None or isinstance(centered, (int, float))
```
fehiepsi, line 78: @fritzo do you use Tensor `centered`? In NumPyro, under `jit` context, we can't check if `centered` belongs to the interval [0, 1] so I skip the support for it. If you think it is useful, I'll remove this assertation.

fritzo: Yes, the main idea of the [original paper](https://arxiv.org/abs/1906.03028) is to learn per-element `centered` parameters. If you simplify anything, I would simplify to make `centered` required and to always be a tensor and assume it is learned.

fritzo: Oh I see, you call `param` statements internally. I think this is a fine assumption.

numpyro/infer/reparam.py
```diff
@@ -26,6 +26,93 @@ def __call__(self, name, fn, obs):
...
+        self.shape_params = shape_params
+
+    def __call__(self, name, fn, obs):
+        assert obs is None, "LocScaleReparam does not support observe statements"
+        centered = self.centered
+        if is_identically_one(centered):
+            return name, fn, obs
+        event_shape = fn.event_shape
+        fn, event_dim = self._unwrap(fn)
+        fn, batch_shape = self._unexpand(fn)
```
fritzo, line 93: I am not convinced that the shape of the centered param is compatible with `_unwrap`ping here. I think we may have different understandings of the typical use cases of batch_shape.

Does it make sense to you to initialize `centered` to `jnp.full(batch_shape + event_shape, 0.5)` below?

fehiepsi: Assume that `fn` is `dist.Normal(0, 1).expand([2]).to_event(1)`. In Pyro, it is the same as `dist.Normal(jnp.zeros(2), jnp.ones(2)).to_event(1)`. After `unwrap`, we get `dist.Normal(jnp.zeros(2), jnp.ones(2))` and `event_dim=1`. Later, we define `centered = jnp.full(event_shape, 0.5)`.

In NumPyro, after `unwrap`, we get `dist.ExpandedDistribution(dist.Normal(0, 1), (2,))` and `event_dim=1`. After `unexpand`, we get `fn=dist.Normal(0, 1)`, `event_dim=1`, `batch_shape=(2,)`. Because `centered` in Pyro is defined independently of `batch_shape`, I think two implementations give the same behavior. WDYT?

fehiepsi: hmm, I think we need to change `centered` to `full(shape, 0.5)` with `shape = (original fn).shape()` in both Pyro and NumPyro.

fritzo: hmm, it has been a while since I looked at the `LocScaleReparam` interface, but I believe I was thinking:

> Users may want to share statistical strength across various batch dimensions, but probably not across event dimensions. Probably the most common usage pattern for learnable `centered` is to create one learnable parameter per event element, i.e. `pyro.param(..., full(event_shape, 0.5))`. In all less common cases, users can create a `centered = pyro.param(...)` statement outside of the `pyro.reparam()` context and pass the `centered` parameter to the `LocScaleReparam`.

Given that perspective, I am happy with the Pyro interface. Is it possible in NumPyro to add support for tensors, but maybe omit some of the short-circuit optimizations in the Pyro version?

fehiepsi: Got it! I believe all I need to do is to change the first assertation in the constructor to `isinstance(centered, (ndarray, ndgeneric, DeviceArray, Tracer,...)` and only do validation for `int, float` cases. The `_unexpand` logic here seems not relevant.

fritzo:
@fehiepsi what is your merge policy on NumPyro? Do authors hit the merge button, or do reviewers?

fehiepsi:
Any is good, as long as reviewers approve. It is less strict than Pyro, but we usually let reviewers merge.

nkaimcaudle:
Is it possible to mention LocScaleReparam in the text above the example code in README? It hasn't been updated and is still talking about AffineTransform.

fehiepsi:
Thanks, @nkaimcaudle! I'll update it.

Issue #674, opened 2020-07-08
Metropolis Hasting MCMC for simulating spin model


PHYSICSIDEA:
Dear numpyro team,

So I am dealing with what is so-called spin models in physics, whereby the state space is discrete and of very large dimension. The standard procedure to approximate the true distribution ( which is theoretically a Boltzmann distribution) of such systems consists of generating a Markov chain of states, one after another, by specifying a transition probability from the current to the next state. For the Metropolis-Hasting method, this transition probability is of the form min (1, exp(-dH)), where dH is equal to the change of a function of the state H (Hamiltonian in physics literature) corresponding to this transition. The reason for taking this procedure, instead of defining a prior distribution is that the state space is exceedingly large and one hence prefers to sample only those regions of considerable weights. The mentioned MH procedure allows one to do so. As I am a total newbie,  could you point me out to a tutorial or start reading for implementing MH procedure in numpyro? Thanks.

fehiepsi:
Hi @PHYSICSIDEA, you can create an MH kernel by subclassing [MCMCKernel](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/infer/mcmc.py#L44) and implementing `init` and `sample` method. Could you take a look there and see if that interface is enough for your purpose? I am happy to discuss with you any difficulty that you got.

fehiepsi:
@PHYSICSIDEA We just expose [MCMCKernel in docs](http://num.pyro.ai/en/latest/mcmc.html#numpyro.infer.mcmc.MCMCKernel). I hope that subclassing it and implementing `init`, `sample` would be enough for your MH kernel compatible with `MCMC` class. Please let me know if you need further clarification either here or in [forum](https://forum.pyro.ai/).

PR #680, opened 2020-07-12
Add MetropolisHastings algorithm to example of MCMCKernel

fehiepsi:
Addresses #674. This also fixes some issues if users' MCMC state does not have `z`, `diverging`, or `mean_accept_prob` fields, or even not a namedtuple.

cc @PHYSICSIDEA

### DIFF
--- a/numpyro/infer/mcmc.py
+++ b/numpyro/infer/mcmc.py
@@ -37,8 +37,10 @@ def get_diagnostics_str(mcmc_state):
         return '{} steps of size {:.2e}. acc. prob={:.2f}'.format(mcmc_state.num_steps,
                                                                   mcmc_state.adapt_state.step_size,
                                                                   mcmc_state.mean_accept_prob)
-    else:
+    elif hasattr(mcmc_state, "mean_accept_prob"):
         return 'acc. prob={:.2f}'.format(mcmc_state.mean_accept_prob)
+    else:
+        return ''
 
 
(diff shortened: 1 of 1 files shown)


neerajprad:
The example looks great. The changes to support arbitrary fields works for now, but I think that instead of hardcoding `('z', 'diverging')` as the set of default fields to collect in `MCMC`, we should instead have a property `MCMCKernel.default_collect_fields` so that the onus of naming their state parameters is on the users, and we can use that within the `MCMC` class. The current interface would still work as users will be able to optionally ask for additional fields via `extra_fields`. WDYT?

fehiepsi:
I think it is a good idea. I like it. Maybe it is more useful to have two fields: `default_sample_field`, `default_extra_fields`? A property for `get_diagnostics_str` function is also useful for the general kernel I think. Do you want to address it in a separate PR?

neerajprad:
+1 on `get_diagnostics_str`. I'm not yet sure if making a distinction between the sample field and other fields will be important, but it is worth keeping in mind.

>  Do you want to address it in a separate PR?

Sure, I'll add this in a separate PR.

fehiepsi:
> making a distinction between the sample field and other fields

I just thought that if `get_samples` returns only `z` field, then we need a way to separate it (I assumed that if users define `x` as a sample field, they will want `mcmc.get_samples` returns `x` field). I don't have a strong opinion on this though. Requiring `z` as the sample field is also good IMO.

Issue #678, opened 2020-07-10
Improve compiling time for numpyro.markov


fehiepsi:
This issue tracks progress toward improving the compiling time of `numpyro.contrib.funsor.markov`.

This can be separated into 3 tasks:
- [x] Collecting all log_prob terms together with their dimension information. We might rewrite `markov` using a functional interface `transition_fn`, then using `scan` for this purpose.
- [x] Using something like [sequential_sum_product](https://github.com/pyro-ppl/funsor/blob/master/funsor/sum_product.py#L135) to compute sum product of those terms in `O(log(time))`.
- ~Generalize the implementation to `history > 1`.~ This is addressed per user-request.

fehiepsi:
For the first task, I think we can do something like
```
state_init = trace(transition_fn)(init_input)

def scan_body(prev_state):
     state_odd = trace(transition_fn)(prev_state)
     state_even = trace(transition_fn)(state_odd)
     return state_even, (state_odd, state_even)

last_state, (odd_states, even_states) = scan(body_fn, state_init, markov_length)
```

fehiepsi:
So far, I come up with the following sketch

```python
# This uses `enum-messengers` branch and replace `handlers.trace` in
# `numpyro.contrib.control_flow.scan` by `numpyro.contrib.funsor.trace`.
from contextlib import contextmanager
import jax.numpy as jnp
import funsor

import numpyro
from numpyro.contrib.control_flow import scan
from numpyro.contrib.funsor import markov, to_funsor, trace, enum, log_density
import numpyro.distributions as dist


def fun_markov(f, init, xs):
    assert xs.shape[0] % 2 == 1

    with numpyro.handlers.scope(prefix="_init"):
        carry, y0 = f(init, xs[0])
    xs_reshape = jnp.reshape(xs[1:], (-1, 2) + xs.shape[1:])

    def f_f(carry, x):
        assert x.shape[0] == 2
        print(carry.shape)
        with numpyro.handlers.scope(prefix="_odd"):
            carry, y_odd = f(carry, x[0])
        print(carry.shape)
        with numpyro.handlers.scope(prefix="_even"):
            carry, y_even = f(carry, x[1])
        print(carry.shape)
        y = jnp.stack([y_odd, y_even])
        return carry, y

    last_carry, y = scan(enum(f_f), carry, xs_reshape)
    assert y.shape[1] == 2
[...]

PR #682, opened 2020-07-13
Implement markov(history=1) by scanning discrete latent variables

fehiepsi:
Changes to support experiments in https://github.com/pyro-ppl/numpyro/issues/678.

### Tasks
- [x] Support enum in scan
- [x] Add some tests
- [x] Add some examples

### DIFF
--- a/examples/hmm_enum.py
+++ b/examples/hmm_enum.py
@@ -29,6 +29,10 @@ density in parallel. In addition, the stacked form allows us
 to use the parallel-scan algorithm in [2], which reduces parallel
 complexity from O(length) to O(log(length)).
 
+Data are taken from [3]. However, the original source of the data
+seems to be the Institut fuer Algorithmen
+und Kognitive Systeme at Universitaet Karlsruhe.
+
 **References:**
 
     1. *Pyro's Hidden Markov Model example*,
@@ -36,11 +40,14 @@ complexity from O(length) to O(log(length)).
     2. *Temporal Parallelization of Bayesian Smoothers*,
        Simo Sarkka, Angel F. Garcia-Fernandez
        (https://arxiv.org/abs/1905.13002)
+    3. *Modeling Temporal Dependencies in High-Dimensional Sequences:
+       Application to Polyphonic Music Generation and Transcription*,
+       Boulanger-Lewandowski, N., Bengio, Y. and Vincent, P.
 """
 
 import argparse
 import logging
-import pickle
+import os
 import time
 
 from jax import random
@@ -50,8 +57,12 @@ import numpyro
 from numpyro.contrib.control_flow import scan
 from numpyro.contrib.indexing import Vindex
 import numpyro.distributions as dist
+from numpyro.examples.datasets import JSB_CHORALES, load_dataset
 from numpyro.handlers import mask
 from numpyro.infer import HMC, MCMC, NUTS
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
 
 
 # Let's start with a simple Hidden Markov Model.
@@ -228,34 +239,34 @@ def main(args):
 
     model = models[args.model]
 
-    with open('./hmm_enum_data.pkl', 'rb') as f:
-        data = pickle.load(f)
-    data['sequences'] = data['sequences'][0:args.num_sequences]
-    data['sequence_lengths'] = data['sequence_lengths'][0:args.num_sequences]
+    _, fetch = load_dataset(JSB_CHORALES, split='train', shuffle=False)
+    lengths, sequences = fetch()
+    sequences = sequences[0:args.num_sequences].astype("int32")
+    lengths = lengths[0:args.num_sequences]
 
-    logging.info('-' * 40)
-    logging.info('Training {} on {} sequences'.format(
-        model.__name__, len(data['sequences'])))
-    sequences = jnp.array(data['sequences'], dtype=jnp.int32)
-    lengths = jnp.array(data['sequence_lengths'])
+    logger.info('-' * 40)
+    logger.info('Training {} on {} sequences'.format(
+        model.__name__, len(sequences)))
 
     # find all the notes that are present at least once in the training set
     present_notes = ((sequences == 1).sum(0).sum(0) > 0)
-    # remove notes that are never played (we remove 37/88 notes)
+    # remove notes that are never played (we remove 44/88 notes with default args)
     sequences = sequences[..., present_notes]
 
     if args.truncate:
         lengths = lengths.clip(0, args.truncate)
         sequences = sequences[:, :args.truncate]
 
-    logging.info('Starting inference...')
+    logger.info('Each sequence has shape {}'.format(sequences[0].shape))
+    logger.info('Starting inference...')
     rng_key = random.PRNGKey(2)
     start = time.time()
     kernel = {'nuts': NUTS, 'hmc': HMC}[args.kernel](model)
-    mcmc = MCMC(kernel, args.num_warmup, args.num_samples, progress_bar=True)
+    mcmc = MCMC(kernel, args.num_warmup, args.num_samples, args.num_chains,
+                progress_bar=False if "NUMPYRO_SPHINXBUILD" in os.environ else True)
     mcmc.run(rng_key, sequences, lengths, args=args)
     mcmc.print_summary()
-    logging.info('\nMCMC elapsed time: {}'.format(time.time() - start))
+    logger.info('\nMCMC elapsed time: {}'.format(time.time() - start))
 
 
 if __name__ == '__main__':
--- a/examples/hmm_enum.py
+++ b/examples/hmm_enum.py
@@ -39,7 +39,8 @@ def model_1(sequences, lengths, args, include_prior=True):
                     numpyro.sample("y", dist.Bernoulli(probs_y[x.squeeze(-1)]), obs=y)
         return (x, t + 1), None
 
-    scan(transition_fn, (0, 0), jnp.swapaxes(sequences, 0, 1))
+    scan(transition_fn, (jnp.zeros((num_sequences, 1), dtype=jnp.int32), 0),
+         jnp.swapaxes(sequences, 0, 1))
 
 
 # Next let's add a dependency of y[t] on y[t-1].
--- a/examples/hmm_enum.py
+++ b/examples/hmm_enum.py
@@ -77,7 +77,7 @@ logger.setLevel(logging.INFO)
 # prior. The latent state is x, and the observed state is y.
 def model_1(sequences, lengths, args, include_prior=True):
     num_sequences, max_length, data_dim = sequences.shape
-    with mask(mask_array=include_prior):
+    with mask(mask=include_prior):
         probs_x = numpyro.sample("probs_x",
                                  dist.Dirichlet(0.9 * jnp.eye(args.hidden_dim) + 0.1)
                                      .to_event(1))
@@ -89,7 +89,7 @@ def model_1(sequences, lengths, args, include_prior=True):
     def transition_fn(carry, y):
         x_prev, t = carry
         with numpyro.plate("sequences", num_sequences, dim=-2):
-            with mask(mask_array=(t < lengths)[..., None]):
+            with mask(mask=(t < lengths)[..., None]):
                 x = numpyro.sample("x", dist.Categorical(probs_x[x_prev]))
                 with numpyro.plate("tones", data_dim, dim=-1):
                     numpyro.sample("y", dist.Bernoulli(probs_y[x.squeeze(-1)]), obs=y)
@@ -108,7 +108,7 @@ def model_1(sequences, lengths, args, include_prior=True):
 #     y[t-1] --> y[t] --> y[t+1]
 def model_2(sequences, lengths, args, include_prior=True):
     num_sequences, max_length, data_dim = sequences.shape
-    with mask(mask_array=include_prior):
+    with mask(mask=include_prior):
         probs_x = numpyro.sample("probs_x",
                                  dist.Dirichlet(0.9 * jnp.eye(args.hidden_dim) + 0.1)
                                      .to_event(1))
@@ -121,7 +121,7 @@ def model_2(sequences, lengths, args, include_prior=True):
     def transition_fn(carry, y):
         x_prev, y_prev, t = carry
         with numpyro.plate("sequences", num_sequences, dim=-2):
-            with mask(mask_array=(t < lengths)[..., None]):
+            with mask(mask=(t < lengths)[..., None]):
                 x = numpyro.sample("x", dist.Categorical(probs_x[x_prev]))
                 # Note the broadcasting tricks here: to index probs_y on tensors x and y,
                 # we also need a final tensor for the tones dimension. This is conveniently
@@ -153,7 +153,7 @@ def model_2(sequences, lengths, args, include_prior=True):
 def model_3(sequences, lengths, args, include_prior=True):
     num_sequences, max_length, data_dim = sequences.shape
     hidden_dim = int(args.hidden_dim ** 0.5)  # split between w and x
-    with mask(mask_array=include_prior):
+    with mask(mask=include_prior):
         probs_w = numpyro.sample("probs_w",
                                  dist.Dirichlet(0.9 * jnp.eye(hidden_dim) + 0.1)
                                      .to_event(1))
@@ -168,7 +168,7 @@ def model_3(sequences, lengths, args, include_prior=True):
     def transition_fn(carry, y):
         w_prev, x_prev, t = carry
         with numpyro.plate("sequences", num_sequences, dim=-2):
-            with mask(mask_array=(t < lengths)[..., None]):
+            with mask(mask=(t < lengths)[..., None]):
                 w = numpyro.sample("w", dist.Categorical(probs_w[w_prev]))
                 x = numpyro.sample("x", dist.Categorical(probs_x[x_prev]))
                 # Note the broadcasting tricks here: to index probs_y on tensors x and y,
@@ -200,7 +200,7 @@ def model_3(sequences, lengths, args, include_prior=True):
 def model_4(sequences, lengths, args, include_prior=True):
     num_sequences, max_length, data_dim = sequences.shape
     hidden_dim = int(args.hidden_dim ** 0.5)  # split between w and x
-    with mask(mask_array=include_prior):
+    with mask(mask=include_prior):
         probs_w = numpyro.sample("probs_w",
                                  dist.Dirichlet(0.9 * jnp.eye(hidden_dim) + 0.1)
                                      .to_event(1))
@@ -216,7 +216,7 @@ def model_4(sequences, lengths, args, include_prior=True):
     def transition_fn(carry, y):
         w_prev, x_prev, t = carry
         with numpyro.plate("sequences", num_sequences, dim=-2):
-            with mask(mask_array=(t < lengths)[..., None]):
+            with mask(mask=(t < lengths)[..., None]):
                 w = numpyro.sample("w", dist.Categorical(probs_w[w_prev]))
                 x = numpyro.sample("x", dist.Categorical(Vindex(probs_x)[w, x_prev]))
                 with numpyro.plate("tones", data_dim, dim=-1) as tones:
--- a/examples/hmm_enum.py
+++ b/examples/hmm_enum.py
@@ -1,0 +1,232 @@
+# Copyright Contributors to the Pyro project.
+# SPDX-License-Identifier: Apache-2.0
+
+import argparse
+import logging
+import pickle
+import time
+
+from jax import random
+import jax.numpy as jnp
+
+import funsor
+
+import numpyro
+from numpyro.contrib.control_flow import scan
+import numpyro.distributions as dist
+from numpyro.handlers import mask, seed
+from numpyro.infer import HMC, MCMC, NUTS
+from numpyro.contrib.indexing import Vindex
+
+
+def model_1(sequences, lengths, args, include_prior=True):
+    num_sequences, max_length, data_dim = sequences.shape
+    with mask(mask_array=include_prior):
+        probs_x = numpyro.sample("probs_x",
+                                 dist.Dirichlet(0.9 * jnp.eye(args.hidden_dim) + 0.1)
+                                     .to_event(1))
+        probs_y = numpyro.sample("probs_y",
+                                 dist.Beta(0.1, 0.9)
+                                     .expand([args.hidden_dim, data_dim])
+                                     .to_event(2))
+
+    def transition_fn(carry, y):
+        x, t = carry
+        with numpyro.plate("sequences", num_sequences, dim=-2):
+            with mask(mask_array=(t < lengths)[..., None]):
+                x = numpyro.sample("x", dist.Categorical(probs_x[x]))
+                with numpyro.plate("tones", data_dim, dim=-1):
+                    numpyro.sample("y", dist.Bernoulli(probs_y[x.squeeze(-1)]), obs=y)
+        return (x, t + 1), None
+
+    scan(transition_fn, (0, 0), jnp.swapaxes(sequences, 0, 1))
+
+
+# Next let's add a dependency of y[t] on y[t-1].
+#
+#     x[t-1] --> x[t] --> x[t+1]
+#        |        |         |
+#        V        V         V
+#     y[t-1] --> y[t] --> y[t+1]
+"""
+def model_2(sequences, lengths, args, include_prior=True):
+    num_sequences, max_length, data_dim = sequences.shape
+    with numpyro_mask(mask_array=include_prior):
+        probs_x = pyro_sample("probs_x",
+                              dist.Dirichlet(0.9 * jnp.eye(args.hidden_dim) + 0.1)
+                                  .to_event(1))
+
+        probs_y = pyro_sample("probs_y",
+                              dist.Beta(0.1, 0.9)
+                                  .expand([args.hidden_dim, 2, data_dim])
+                                  .to_event(3))
+    tones_plate = pyro_plate("tones", data_dim, dim=-1)
+    with pyro_plate("sequences", num_sequences, dim=-2) as batch:
+        lengths = lengths[batch]
+        x, y = 0, 0
+        for t in pyro_markov(range(max_length)):
+            with numpyro_mask(mask_array=(t < lengths)[..., None]):
+                x = pyro_sample("x_{}".format(t), dist.Categorical(probs_x[x]),
+                                infer={"enumerate": "parallel"})
+                logging.info(f"x[{t}]: {x.shape}")
+                # Note the broadcasting tricks here: to index probs_y on tensors x and y,
+                # we also need a final tensor for the tones dimension. This is conveniently
+                # provided by the plate associated with that dimension.
+                with tones_plate as tones:
+                    y = pyro_sample("y_{}".format(t),
+                                    dist.Bernoulli(probs_y[x, y, tones]),
+                                    obs=sequences[batch, t]).astype(jnp.int32)
+
+
+# Next consider a Factorial HMM with two hidden states.
+#
+#    w[t-1] ----> w[t] ---> w[t+1]
+#        \ x[t-1] --\-> x[t] --\-> x[t+1]
+#         \  /       \  /       \  /
+#          \/         \/         \/
+#        y[t-1]      y[t]      y[t+1]
+#
+# Note that since the joint distribution of each y[t] depends on two variables,
+# those two variables become dependent. Therefore during enumeration, the
+# entire joint space of these variables w[t],x[t] needs to be enumerated.
+# For that reason, we set the dimension of each to the square root of the
+# target hidden dimension.
+def model_3(sequences, lengths, args, include_prior=True):
+    num_sequences, max_length, data_dim = sequences.shape
+    hidden_dim = int(args.hidden_dim ** 0.5)  # split between w and x
+    with numpyro_mask(mask_array=include_prior):
+        probs_w = pyro_sample("probs_w",
+                              dist.Dirichlet(0.9 * jnp.eye(hidden_dim) + 0.1)
+                                  .to_event(1))
+        probs_x = pyro_sample("probs_x",
+                              dist.Dirichlet(0.9 * jnp.eye(hidden_dim) + 0.1)
+                                  .to_event(1))
+        probs_y = pyro_sample("probs_y",
+                              dist.Beta(0.1, 0.9)
+                                  .expand([args.hidden_dim, 2, data_dim])
+                                  .to_event(3))
+
+    tones_plate = pyro_plate("tones", data_dim, dim=-1)
+    with pyro_plate("sequences", num_sequences, dim=-2) as batch:
+        lengths = lengths[batch]
+        w, x = 0, 0
+        for t in pyro_markov(range(max_length)):
+            with numpyro_mask(mask_array=(t < lengths)[..., None]):
+                w = pyro_sample("w_{}".format(t), dist.Categorical(probs_w[w]),
+                                infer={"enumerate": "parallel"})
+                logging.info(f"w[{t}]: {w.shape}")
+                x = pyro_sample("x_{}".format(t), dist.Categorical(probs_x[x]),
+                                infer={"enumerate": "parallel"})
+                logging.info(f"x[{t}]: {x.shape}")
+                with tones_plate as tones:
+                    pyro_sample("y_{}".format(t), dist.Bernoulli(probs_y[w, x, tones]),
+                                obs=sequences[batch, t])
+
+
+# By adding a dependency of x on w, we generalize to a
+# Dynamic Bayesian Network.
+#
+#     w[t-1] ----> w[t] ---> w[t+1]
+#        |  \       |  \       |   \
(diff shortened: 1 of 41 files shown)


numpyro/contrib/control_flow/scan.py
```diff
@@ -194,7 +270,14 @@ def g(*args, **kwargs):
...
-        apply_stack(msg)
+    if not msg["kwargs"].get("enum", False):
+        for msg in pytree_trace.trace.values():
+            apply_stack(msg)
+    else:
+        for msg in pytree_trace.trace.values():
+            with LocalNamedMessenger():
+                dim_to_name = msg["infer"].get("dim_to_name")
+                to_funsor(msg["value"], dim_to_name=OrderedDict([(k, dim_to_name[k]) for k in sorted(dim_to_name)]))
+                apply_stack(msg)
```
eb8680, line 280: This logic is fine, the `LocalNamedMessenger` here ensures that these dimensions have no effect on the rest of model execution.

numpyro/contrib/funsor/enum_messenger.py
```diff
@@ -253,9 +253,9 @@ def _get_dim_to_name(batch_shape, dim_to_name=None, dim_type=DimType.LOCAL):
 
         # interpret all names/dims as requests since we only run this function once
         for dim in range(-batch_dim, 0):
-            if batch_shape[dim] == 1:
-                continue
             name = dim_to_name.get(dim, None)
+            if batch_shape[dim] == 1 and name is None:
+                continue
```
eb8680, line 258: Per discussion on Slack, this change was made because the time dimension on the left sometimes necessitates empty dimensions appearing before they have been assigned a name.

numpyro/contrib/funsor/enum_messenger.py
```diff
@@ -501,6 +501,8 @@ def process_message(self, msg):
         if msg["type"] != "sample" or \
                 msg.get("done", False) or msg["is_observed"] or msg["infer"].get("expand", False) or \
                 msg["infer"].get("enumerate") != "parallel" or (not msg["fn"].has_enumerate_support):
+            if msg["type"] == "control_flow":
+                msg["kwargs"]["enum"] = True
```
eb8680, line 505: nit: this should probably live in `infer_config`

numpyro/contrib/funsor/infer_util.py
```diff
@@ -70,6 +71,28 @@ def config_fn(site):
...
+        prev_vars = markov_vars[time_var]
+        prev_to_curr = {}
+        curr_vars = []
+        for k in prev_vars:
+            curr_var = "/".join(k.split("/")[1:])
+            curr_vars.append(curr_var)
+            prev_to_curr[k] = curr_var
+        prev_and_curr = prev_vars | frozenset(curr_vars)
+        with funsor.interpreter.interpretation(funsor.terms.lazy):
+            eliminate = (sum_vars | prod_vars) - prev_and_curr
```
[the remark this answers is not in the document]
fehiepsi: Time variables do not appear in `sum_vars` and `prod_vars` so I think this is fine.

numpyro/contrib/funsor/infer_util.py
```diff
@@ -70,6 +71,28 @@ def config_fn(site):
     return infer_config(fn, config_fn)
 
 
+def compute_markov_factors(time_to_factors, markov_vars, sum_vars, prod_vars):
```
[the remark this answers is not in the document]
fehiepsi: Assume `x` is a discrete latent, `markov_vars[time]` is [`_init/x`]. If there are two discrete latents, `markov_vars[time]` is [`_init/x`, `_init/w`]. `sum_vars` are all variables `_init/x`, `_init_w`, `x`, `w`, `y`. `prod_vars` contains plate dimensions (inside and outside scan).

numpyro/contrib/funsor/infer_util.py
```diff
@@ -109,11 +134,26 @@ def model(*args, **kwargs):
...
+                        s for s in dim_to_name.values() if s.startswith("_init/"))
+                    break
+            if time_dim is None:
+                log_factors.append(log_prob)
             sum_vars |= frozenset({site['name']})
             prod_vars |= frozenset(f.name for f in site['cond_indep_stack'] if f.dim is not None)
 
+    if len(time_to_factors) > 0:
+        markov_factors = compute_markov_factors(time_to_factors, markov_vars, sum_vars, prod_vars)
+        log_factors = log_factors + markov_factors
```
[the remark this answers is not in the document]
fehiepsi: Good question! `plate` inside and outside `scan` should be good, but using `scan` inside `plate` does not work (we can easily move `plate` inside `scan` though). However, I think the scan `transition` might not work if it depends on an outside discrete latent variable (I am not sure, I need to add a test if it works).

> no factors outside of the scan can depend on fresh variables from inside the scan

I am not sure. Factors outside of the scan can depend on the last carry value. I'll need to add a test for it.

numpyro/contrib/funsor/enum_messenger.py
fehiepsi: yeah, I'll add this comment. thanks! :D

numpyro/contrib/funsor/enum_messenger.py
fehiepsi: Hmm, I need `scan_wrapper` knows that `enum=True` but `scan_wrapper` only accesses to `args`, `kwargs`, not `infer` field.

numpyro/contrib/funsor/infer_util.py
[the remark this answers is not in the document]
fehiepsi: For factors depend on the last x, I guess we need to reshape the last carry value... I am not sure if dimensions are aligned correctly. Let me test this first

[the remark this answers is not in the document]
fehiepsi: To me: make sure to add test for plate.

test/contrib/test_funsor.py
```diff
@@ -188,3 +193,94 @@ def testing():
...
+            probs = init_probs if x is None else transition_probs[x]
+            x = numpyro.sample("x", dist.Categorical(probs))
+            numpyro.sample("y", dist.Normal(locs[x], 1), obs=y)
+            return x, None
+
+        scan(transition_fn, None, data)
+
+    actual_log_joint = log_density(enum(config_enumerate(fun_model)), (data,), {}, {})[0]
+    expected_log_joint = log_density(enum(config_enumerate(fun_model)), (data,), {}, {})[0]
+    assert_allclose(actual_log_joint, expected_log_joint)
```
fehiepsi, line 223: Ops, I tested fun_model to itself.........

numpyro/contrib/funsor/infer_util.py
```diff
@@ -109,11 +134,26 @@ def model(*args, **kwargs):
...
                 log_prob = scale * log_prob
 
-            log_prob = funsor.to_funsor(log_prob, output=funsor.reals(), dim_to_name=site['infer']['dim_to_name'])
-            log_factors.append(log_prob)
+            dim_to_name = site["infer"]["dim_to_name"]
+            log_prob = funsor.to_funsor(log_prob, output=funsor.reals(), dim_to_name=dim_to_name)
+
+            time_dim = None
+            for dim, name in dim_to_name.items():
+                if name.startswith("_time"):
```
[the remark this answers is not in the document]
fehiepsi: I explained the name in the docs of `scan`.

numpyro/contrib/funsor/infer_util.py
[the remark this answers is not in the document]
fehiepsi: @eb8680 It seems that `markov` does not allow outside-dependency
```python
import jax.numpy as jnp
import funsor
import numpyro
from numpyro.contrib.control_flow import scan
from numpyro.contrib.funsor import config_enumerate, markov, to_funsor, trace, enum, log_density

data = jnp.repeat(jnp.array([0, 1]), (7, 4))
transition_probs = jnp.array([[0.8, 0.2], [0.1, 0.9]])
locs = jnp.array([-1.0, 1.0])

def model(data):
    x = None
    for i, y in markov(enumerate(data)):
        probs = jnp.array([0.5, 0.5]) if x is None else transition_probs[x]
        x = numpyro.sample(f"x_{i}", dist.Categorical(probs))
        numpyro.sample(f"y_{i}", dist.Normal(locs[x], 1), obs=y)
    numpyro.sample("w", dist.Normal(jnp.zeros(2)[x]))

with trace() as tr, enum():
    numpyro.handlers.seed(config_enumerate(model), rng_seed=1)(data)
print(tr["w"]["infer"])
log_density(enum(config_enumerate(model)), (data,), {}, {'w': jnp.ones(2)})[0]
```
returns
```
{'dim_to_name': OrderedDict([(-1, '_pyro_dim_1')])}
DeviceArray([-15.72769, -15.72769], dtype=float32)
```
Is this a bug or unsupported?

numpyro/contrib/funsor/infer_util.py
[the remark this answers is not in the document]
eb8680: If you have a time-global variable that couples a bunch of time-local variables (like a global parameter in a state-space model's transition matrix), then eliminating that global variable couples all of the local variables and breaks the conditional independence that allows the use of `sequential_sum_product`.  I think you'll need to run `sequential_sum_product` before this step, rather than after.

numpyro/contrib/funsor/infer_util.py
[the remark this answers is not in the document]
eb8680: > Is this a bug or unsupported?

This is unsupported under `markov` as it is currently defined, but it's a common enough scenario that we should probably change `markov` to support it.

numpyro/contrib/funsor/infer_util.py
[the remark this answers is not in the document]
fehiepsi: Hmm, I think I understand this wrong. Using `eliminate=frozenset()` passes `plate` and global discrete tests.

numpyro/contrib/funsor/infer_util.py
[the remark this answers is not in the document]
fehiepsi: I just remove `sum_vars` and rename `markov_vars` to `init_vars` for clarity.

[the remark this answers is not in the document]
fehiepsi: > the expected .inputs and .output of the return value and the arguments in time_to_factors

`time_to_factors[time]` contains all log_probs with time `time`. I think all of them have `reals()` outputs. `inputs` are `dim_to_name` at those sites having `time` dimension.

numpyro/contrib/funsor/infer_util.py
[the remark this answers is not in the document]
eb8680: Yes, when `eliminate == frozenset()` this works fine, but when that's not the case it may be incorrect to sum out variables in this order.

numpyro/contrib/funsor/infer_util.py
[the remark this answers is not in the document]
fehiepsi: I just made a change that guarantees shape of the last carry is consistent w.r.t. sequential strategy. For now, do you think that we can add a note that there should be no variable depends on the last carry? (or there is a simple fix here?)

numpyro/contrib/funsor/infer_util.py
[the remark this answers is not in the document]
fehiepsi: It seems that the test for multiple plates at the same dimension does not pass due to this issue. I think I need to follow your strategy: using `sequential_sum_product` before this step. Did you suggested that we need to eliminate time dimensions in `x[t]|x[t-1]` and `y[t]|y[t-1]` separately?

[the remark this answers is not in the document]
fehiepsi: Yes, so I think we need to set eliminate=frozenset(). The tricky thing is when we have two plate statements at dim=-1, z[t]|x[t] and y[t]|z[t], we need to concatenate z and y, eliminate dimensions, then split the result. The current code will allocate two separate dimensions for two plates, which leads to wrong result. Do you have a solution for this?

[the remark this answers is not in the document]
fehiepsi: Because we don't eliminate any variable, I think we can use `sum(log_factors)` here for simplicity. However, we still need to have a mechanism to merge variables that allocate the same dimension. It is the job of `sequential_sum_product` I guess?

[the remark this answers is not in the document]
eb8680: > However, we still need to have a mechanism to merge variables that allocate the same dimension

Sorry, I'm not sure I understand the question.  What do you mean here? What is the difference between `z[t, i]` and `z[t]`, which I assume are supposed to be `.inputs` of a factor funsor?  In this function, all of the log-factors are funsors, so I don't see why you can't just sum the log-factors.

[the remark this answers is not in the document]
fehiepsi: Yes, we can take the sum, but assume z has addition plate D1 at dim=-1 and y has addition plate D2 at dim=-1, taking the sum will return a funsor with inputs time x prev x curr x D1 x D2. But I think the correct behavior is time x prev x curr x (D1 + D2).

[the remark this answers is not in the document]
eb8680: > but assume z has addition plate D1 at dim=-1 and y has addition plate D2 at dim=-1

Sorry, I still don't understand what you mean when you say that a funsor input `z` has an additional plate at dim=-1. Inputs are just inputs; all plates should be included in the log-factor funsors as additional inputs.  Can you give a concrete example set of log-factor funsors that exhibit this behavior?

test/contrib/test_funsor.py
```diff
@@ -188,3 +193,198 @@ def testing():
...
+            return x, None
+
+        scan(transition_fn, None, data)
+
+    actual_log_joint = log_density(enum(config_enumerate(fun_model), -2), (data,), {}, {})[0]
+    expected_log_joint = log_density(enum(config_enumerate(model), -2), (data,), {}, {})[0]
+    assert_allclose(actual_log_joint, expected_log_joint)
+
+
+def test_scan_enum_plates_same_dim():
```
fehiepsi, line 264: @eb8680 I meant this test.

numpyro/contrib/funsor/infer_util.py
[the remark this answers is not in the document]
eb8680: I see, I understand the situation you have in mind in `test_scan_enum_plates_same_dim`. I don't see why the fact that the plates share a dimension (dim=-1) should matter from the perspective of funsor, since everything is named and unordered.  However, I do understand now what you mean about stacking the plate dimensions (`D1 + D2`) rather than broadcasting to the cartesian product (`D1 x D2`).  

You can run tensor variable elimination (`funsor.sum_product.sum_product`) to eliminate those plates before applying `sequential_sum_product`, as you were doing before.  This would have the same effect as stacking the plates, but there are subtle restrictions on what can happen inside of them in the model for variable elimination to remain tractable - conservatively, any enumeration inside should be restricted to a single timestep (so the value of an enumerated sample site within one of the plates may not leave it or be used at the next timestep).  The situation in `test_scan_enum_plates_same_dim` is a special case of this where no new variables are enumerated within either plate.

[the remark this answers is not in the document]
fehiepsi: Thanks a lot, @eb8680!!! I just eliminate all dimensions that do not appear in markov sites (discrete sites that depend on the last step) and ... all test cases that I can think about passed. Could you take a look at the tests and suggest to me other scenarios that we need to tests?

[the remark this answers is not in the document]
fehiepsi: Looking like I am ready to port `hmm_enum` example now. :D thanks for your help!

fehiepsi:
@eb8680 @neerajprad Looking like all tests and examples pass. Compiling time is much more reasonable now (without truncation, model 1 took less than 1m in my slow Windows subsystem for Linux). :D 

@neerajprad Could you help me store the data file to cloudfront?

examples/hmm_enum.py
```diff
@@ -0,0 +1,279 @@
...
+    def transition_fn(carry, y):
+        x_prev, t = carry
+        with numpyro.plate("sequences", num_sequences, dim=-2):
+            with mask(mask_array=(t < lengths)[..., None]):
+                x = numpyro.sample("x", dist.Categorical(probs_x[x_prev]))
+                with numpyro.plate("tones", data_dim, dim=-1):
+                    numpyro.sample("y", dist.Bernoulli(probs_y[x.squeeze(-1)]), obs=y)
+        return (x, t + 1), None
+
+    x_init = jnp.zeros((num_sequences, 1), dtype=jnp.int32)
```
fehiepsi, line 87: Instead of `x_init = 0` in Pyro, we need to make `x_init` have the same shape as the output (in non-enumerate version).

examples/hmm_enum.py
```diff
@@ -0,0 +1,279 @@
...
+
+Note that difference from [1], which uses Python loop, here we use
+:func:`~numpryo.contrib.control_flow.scan` to reduce compilation
+times (only one step needs to be compiled) of the model. Under the
+hood, `scan` stacks all the priors' parameters and values into
+an additional time dimension. This allows us computing the joint
+density in parallel. In addition, the stacked form allows us
+to use the parallel-scan algorithm in [2], which reduces parallel
+complexity from O(length) to O(log(length)).
+
```
fehiepsi, line 31: @eb8680 Could you suggest me a paragraph mentioning about how to obtain the data and how to preprocess it?

eb8680: I think the data you're using is a single batch from the music dataset used in the Pyro HMM and DMM examples. The relevant Pyro file, with a bit of commentary, is [here](https://github.com/pyro-ppl/pyro/blob/dev/examples/dmm/polyphonic_data_loader.py).  Unfortunately the preprocessing logic there depends heavily on PyTorch, so the easiest thing to do is probably to do the preprocessing once in Pyro, save the resulting transformed data, and store it in Cloudfront.

numpyro/contrib/funsor/enum_messenger.py
eb8680: Sorry, I meant that this logic setting `msg["kwargs"]["enum"] = True` should live in `infer_config`, not in `enum`.

numpyro/contrib/funsor/infer_util.py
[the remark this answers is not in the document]
eb8680: You can just add a note about the restriction for now.  I'm not sure there's a quick fix, although I do think it should be changed at some point.

examples/hmm_enum.py
```diff
@@ -0,0 +1,279 @@
...
+                                 dist.Dirichlet(0.9 * jnp.eye(args.hidden_dim) + 0.1)
+                                     .to_event(1))
+        probs_y = numpyro.sample("probs_y",
+                                 dist.Beta(0.1, 0.9)
+                                     .expand([args.hidden_dim, data_dim])
+                                     .to_event(2))
+
+    def transition_fn(carry, y):
+        x_prev, t = carry
+        with numpyro.plate("sequences", num_sequences, dim=-2):
```
eb8680, line 80: This is a nit for these models, but these `sequences` plates should really live outside of `transition_fn` to avoid tempting users to potentially break enumeration by enumerating variables inside plates across timesteps.  Were there issues supporting that? You may need to do the same trick with `plate` that you did with `trace` and `substitute`, where you push the `plate` handler inside of the `scan` wrapper, but that doesn't seem so bad.

numpyro/contrib/control_flow/scan.py
```diff
@@ -78,7 +83,99 @@ def _subs_wrapper(subs_map, i, length, site):
...
+                msg["value"] = jnp.reshape(value, prepend_shapes + jnp.shape(value))
+            elif fn_batch_ndim < value_batch_ndims:
+                msg["fn"] = tree_map(lambda x: jnp.reshape(x, prepend_shapes + jnp.shape(x)), fn)
+
+
+def scan_enum(f, init, xs, length, reverse, rng_key=None, substitute_stack=None):
+    # XXX: This implementation only works for history size=1 but can be
+    # extended to history size > 1 by running `f` `history_size` times
+    # for initialization. However, `sequential_sum_product` does not
+    # support history size > 1, so we skip supporting it here.
```
eb8680, line 104: Note `funsor.sum_product.sarkka_bilmes_product` does support `history > 1`.

numpyro/contrib/control_flow/scan.py
```diff
@@ -78,7 +83,99 @@ def _subs_wrapper(subs_map, i, length, site):
...
+                trace = {}
+        else:
+            with handlers.block(), packed_trace() as trace, promote_shapes(), enum(), markov():
+                new_carry, y = config_enumerate(seeded_fn)(carry, x)
+
+            # store shape of new_carry at a global variable
+            nonlocal carry_shape_at_t1
+            carry_shape_at_t1 = [jnp.shape(x) for x in tree_flatten(new_carry)[0]]
+            # make new_carry have the same shape as carry
+            # FIXME: is this rigorous?
```
eb8680, line 139: You may want to add some unit tests for this

examples/hmm_enum.py
fehiepsi: I see. I think we just need to mention what is the structure of the data. E.g. in `pyro`, the data has 88 tones but it seems to only have 44 tones here.

numpyro/contrib/funsor/infer_util.py
```diff
@@ -70,6 +71,33 @@ def config_fn(site):
...
+    :param frozenset prod_vars: all plate dimensions in the trace.
+    :returns: a list of factors after eliminate time dimensions
+    """
+    markov_factors = []
+    for time_var, log_factors in time_to_factors.items():
+        prev_vars = time_to_init_vars[time_var]
+        # remove `_init/` prefix to convert prev to curr
+        prev_to_curr = {k: "/".join(k.split("/")[1:]) for k in prev_vars}
+        # we eliminate all plate dimensions not available at discrete sites.
+        eliminate_vars = (sum_vars | prod_vars) - time_to_markov_dims[time_var]
```
eb8680, line 90: Does this `eliminate_vars` include variables only enumerated *within* a single timestep in `scan`? For example, suppose that for whatever reason you had a discrete HMM whose emission distributions were Gaussian mixtures rather than Gaussian - you can eliminate the latent mixture component in the emission distribution in the initial `funsor.sum_product.sum_product` call below, but not the latent state of the markov chain.

numpyro/contrib/control_flow/scan.py
fehiepsi: Oh, I didn't know about that. Nice, thanks for pointing out! I temp into trying it this weekend. :D

numpyro/contrib/control_flow/scan.py
```diff
@@ -78,7 +83,99 @@ def _subs_wrapper(subs_map, i, length, site):
...
+            elif subs_type == 'substitute':
+                seeded_fn = handlers.substitute(seeded_fn, substitute_fn=subs_fn)
+
+        if init:
+            with handlers.scope(prefix="_init"):
+                new_carry, y = seeded_fn(carry, x)
+                trace = {}
+        else:
+            with handlers.block(), packed_trace() as trace, promote_shapes(), enum(), markov():
+                new_carry, y = config_enumerate(seeded_fn)(carry, x)
```
eb8680, line 133: Add a comment explaining what's happening here for maintainability.

numpyro/contrib/control_flow/scan.py
fehiepsi: The thing is it passes all the test and I can't figure out when it possibly fails. :(

examples/hmm_enum.py
eb8680: I think there are only 44 tones here because the batch I grabbed only happened to contain 44 notes.  I would suggest using the full dataset instead now that the compilation issues have been resolved.

numpyro/contrib/control_flow/scan.py
eb8680: Things may get more complicated if you decide to support `history > 1` and there are enumerated variables with different periods.

numpyro/contrib/funsor/infer_util.py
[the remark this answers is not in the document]
fehiepsi: Sure, I'll add to the docs

numpyro/contrib/control_flow/scan.py
```diff
@@ -78,7 +83,99 @@ def _subs_wrapper(subs_map, i, length, site):
                                f" but got {value_ndim}. Please report the issue to us!")
 
 
-def scan_wrapper(f, init, xs, length, reverse, rng_key=None, substitute_stack=[]):
+class promote_shapes(Messenger):
+    # a helper messenger to promote shapes of `fn` and `value`
```
eb8680, line 87: nit: add a comment or docstring with example old and new shapes for `fn` and `value` for maintainability.

numpyro/contrib/funsor/infer_util.py
fehiepsi: Does this [test](https://github.com/pyro-ppl/numpyro/pull/682/files#diff-fed0396c70de3697c3e0eb518df45b00R303) cover that usage case? Here we not eliminate only dimensions in the latent state of the markov chain.

eb8680: Yes - if you run that test and set a debugger breakpoint here, what are the contents of `eliminate_vars`?  Does it contain `w`?

examples/hmm_enum.py
fehiepsi: Do you mean to support something like
```
with plate(...):
    scan(...)
```
? If so, I think that it is tricky to put time dimensions at the right position (it seems to be doable but might require non-enum users write correct plate codes). Could you give an example on how things could go wrong so I can add to the docs instead?

eb8680: All of the [restrictions from Pyro's enumeration tutorial](http://pyro.ai/examples/enumeration.html#Dependencies-among-plates) still apply.  A simple failure case would be having an enumerated variable outside of a plate depend on a variable enumerated inside that plate at the previous timestep (`x_t ~ p( . | x_t-1, y_i_t-1)`, `y_t_i ~ p( . | x_t)`).

numpyro/contrib/funsor/enum_messenger.py
fehiepsi: Do you suggest that when we use `config_enumerate(fn)`, we will add `enum` kwargs to `control_flow` msg? Currently, `infer_config` does
```
    if msg["type"] in ("sample", "param"):
            msg["infer"].update(self.config_fn(msg))
```
so we need to change it to
```
        if msg["type"] in ("sample", "param"):
            msg["infer"].update(self.config_fn(msg))
        if msg["type"] == "control_flow":
         msg["kwargs"].update(self.config_fn(msg))
```
and in `config_enumerate`, we need to check if `msg` is `control_flow` to return `{"enum": True}`. It is a bit verbose to me and violate the purpose of `infer_config`. In addition, we don't need to use `config_enumerate` if users already provide `infer={"enumerate": "parallel"}` in `sample` statements. In that case, the only handler that affects `control_flow` is `enum`.

eb8680: > In that case, the only handler that affects control_flow is enum.

Ah, yes, good point, I guess what you have here is fine.

fehiepsi:
Thanks for reviewing Eli, I will address your comments soon! I am not sure how large the "batch" of the original dataset is, probably too large to run MCMC? About nested scan, I haven't tested if it works out of box because I didn't design `scan` to be used in nested pattern. Let me test and if it doesn't work, I'll make an issue for it. For `plates` outside `scan`, I think that we can create a separate issue. The most complicated issue is to move "time" dimension to the correct position for data of observed sites in `transition_fn`, of condition/substitute handlers, and all the collected `sample` msg of latent sites. It might be nice to have but I am not sure if there is a simple way to do it.

examples/hmm_enum.py
fehiepsi: Thanks, I didn't know about those restrictions. I'll add a link to it and learn more about them.

eb8680:
> I am not sure how large the "batch" of the original dataset is

I wouldn't be surprised if NumPyro's HMC + `sequential_sum_product` could handle the smallest of the four original music datasets directly, or could at least do so with the help of the parallel scan primitive from #686 and a GPU or TPU.  If it does end up being too large, I suggest switching to another dataset or loading a full music dataset and taking a slice here for reproducibility, since I don't remember which one I took these from...

> For plates outside scan, I think that we can create a separate issue.

It's up to you to decide what to include in the upcoming release, but since support for this is a notable difference from enumeration in Pyro, I suggest adding some logic in this PR to raise a `NotImplementedError` in `enum.process_msg` if a plate handler is on the handler stack, with a helpful error message that would explain how the user could fix their model.  Ditto for nested `scan` - have `enum` raise a `NotImplementedError` with an informative error message unless/until you support that case.  You may also want to add xfailing tests for both situations.

I'll have to think a bit more about whether there's an easier implementation strategy for supporting plates outside `scan`.

numpyro/contrib/funsor/infer_util.py
fehiepsi: Yes, it is `frozenset({'y', '_init/w', '_init/y', 'D', 'w'}`. Is it expected?

fehiepsi:
Thanks Eli! Currently, we raised `RuntimeError` for `scan` under `plate`. I have switched it to `NotImplementedError` and told users to see `scan` docs for example on how to move `plate` into `scan`. About nested scan, it does not work even for continuous case due to substitute logic, which does not . I have added a sentence there tell users that nested scan is not supported in the docs. I'll study a bit about the data to see if we can use the original version here.

examples/hmm_enum.py
fehiepsi: I can confirm that there are 44 notes with `num_sequences=17`. :D

fehiepsi:
@eb8680 Thanks for throughout reviews! I think I have addressed most of your comments. The remaining one is at `eliminate_vars` which I am not sure if the current behavior is at your expectation. If you notice that I missed something else, please let me know.

fehiepsi:
I also want to comment that although compiling time is not an issue anymore, MCMC is still slow for this dataset, especially at early steps where NUTS takes long trajectory to get u-turn/divergence. Probably it is faster in GPU because we have vectorized computations but I don't have GPU to test (my desktop is unusable after delivered to my new home :( ).

numpyro/contrib/funsor/infer_util.py
```diff
@@ -109,11 +140,35 @@ def model(*args, **kwargs):
...
+                if name.startswith("_time"):
+                    time_dim = funsor.Variable(name, funsor.domains.bint(site["value"].shape[dim]))
+                    time_to_factors[time_dim].append(log_prob)
+                    time_to_init_vars[time_dim] |= frozenset(
+                        s for s in dim_to_name.values() if s.startswith("_init"))
+                    break
+            if time_dim is None:
+                log_factors.append(log_prob)
+
             sum_vars |= frozenset({site['name']})
```
eb8680, line 157: Don't add `site['name']` to `sum_vars` here if it was observed

numpyro/contrib/funsor/infer_util.py
eb8680: Yeah, `w`, `_init/w` and `D` are expected.  `y` shouldn't be in `sum_vars` in the first place, since it's not an input of any of the log-factor funsors. I left a comment indicating where that should be fixed.

numpyro/contrib/funsor/infer_util.py
fehiepsi: Thanks, I just made the commit.

eb8680:
> Probably it is faster in GPU because we have vectorized computations but I don't have GPU to test

@fehiepsi I ran `python hmm_enum.py -m 2 --device gpu --num-sequences 110` on a cloud GPU (Tesla K80, ~12GB RAM) and it finished in a couple hours. I suspect that GPU could just about handle the entire dataset (`--num-sequences 229`), but there seems to be some initial processing (which I haven't looked into) that uses quite a bit more GPU memory than the actual sampling steps and causes an out-of-memory allocation error when I try to increase `num_sequences` further.

fehiepsi:
@eb8680 Before sampling, we try to find valid initial parameters, which involves evaluating the joint density. With hidden_dim=16, tones=51, num_sequences=229, length=129, the log_prob at `x` has shape `128 x 16 x 16 x 229 x 1` and at `y` is `128 x 16 x 1 x 229 x 51`, which are a bit large for involved operators I guess.

eb8680:
> Before sampling, we try to find valid initial parameters, which involves evaluating the joint density

Hmm, would it be possible to add initialization strategies that don't enumerate discrete variables?  That seems like a useful optimization, although probably not worth looking at before the release.

fehiepsi:
We find initial parameters sequentially, so memory requirement should be similar to sampling. I think you can verify it by separating out `compile` and `sample` steps as in [this benchmark](https://github.com/pyro-ppl/numpyro/blob/benchmarks/benchmarks/hmm.py#L91). I think your observation of high memory usage before the actual sampling might not be relevant to it.

fehiepsi:
@eb8680 I tried to inspect the execution in different ways and observed the same thing as yours: memory used for initialization is much higher than sampling...

fehiepsi:
I think the issue is `numpyro.plate` returns `None` instead of the list of indices. Could you verify if it fixes the issue in your system? It fixes the issue in my system.

eb8680:
Nice! Now the HMM examples run on the full dataset on that same GPU.

PR #684, opened 2020-07-15
Promote shape when unflatten expand

fehiepsi:
Previously, we can't scan transition functions including sites such as `dist.Normal(0, 1).expand([10])`. After this PR, it is possible.

This is also useful to reshape distributions when scanning `markov` transition.

### DIFF
--- a/numpyro/distributions/distribution.py
+++ b/numpyro/distributions/distribution.py
@@ -34,7 +34,7 @@ import jax.numpy as jnp
 
 from numpyro.distributions.constraints import is_dependent, real
 from numpyro.distributions.transforms import Transform
-from numpyro.distributions.util import lazy_property, sum_rightmost, validate_sample
+from numpyro.distributions.util import lazy_property, promote_shapes, sum_rightmost, validate_sample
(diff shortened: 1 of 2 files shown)


numpyro/distributions/distribution.py
```diff
@@ -413,33 +413,11 @@ def variance(self):
...
-            # issue). So we don't have to worry about it.
-            raise ValueError("base_dist's batch_shape and expand shape have different lengths."
-                             " This will lead to ambiguous results when unflattening a"
-                             " scanned/vmapped version of this distribution."
-                             " To avoid this issue, make sure that your base_dist's"
-                             " parameters have the same batch_shape as this expand distribution.")
+        prepend_ndim = len(self.batch_shape) - len(self.base_dist.batch_shape)
+        base_dist = tree_util.tree_map(
+            lambda x: promote_shapes(x, shape=(1,) * prepend_ndim + jnp.shape(x))[0],
+            self.base_dist)
```
fehiepsi, line 419: Just until today, I realize that this hack works. Previously, I thought it is tricky to reshape distributions. :D

PR #690, opened 2020-07-21
Allow default doc inheritance in sphinx

neerajprad:
Our `copy_from_docs` utility function is not working currently, i.e. there is not docstring produces for the annotated methods. This removes the function and relies on sphinx's default docstring inheritance mechanism. Also adds the missing `funsor.rst` file to the docs source.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,3 +1,4 @@
 jax>=0.1.65
 jaxlib>=0.1.45
+funsor>=0.1.2
 tqdm
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -66,11 +66,9 @@ extensions = [
     'sphinx.ext.viewcode',
 ]
 
-# Disable documentation inheritance so as to avoid inheriting
-# docstrings in a different format, e.g. when the parent class
-# is a PyTorch class.
+# Enable documentation inheritance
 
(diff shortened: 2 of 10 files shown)


fritzo:
FYI the only reason we created a `copy_docs_from` hack in Pyro was because PyTorch uses NumPy-style dosctrings whereas Pyro uses rst-style (i think) docstrings, and inheriting broke sphinx. I suspect JAX also uses NumPy-style, but maybe you don't inherit from anything in JAX. Let me know if you find a cleaner workaround!

neerajprad:
> FYI the only reason we created a copy_docs_from hack in Pyro was because PyTorch uses NumPy-style dosctrings whereas Pyro uses rst-style (i think) docstrings, and inheriting broke sphinx when the two styles were combined. I suspect JAX also uses NumPy-style, but maybe you don't inherit from anything in JAX. Let me know if you find a cleaner workaround!

IIRC, when we had this helper decorator, sphinx did not have doc inheritance but even after the feature was released sometime in 2018, it was causing issues in Pyro because of our distributions wrapping over PyTorch (it was probably just format mismatch as you mention) and we had to turn it off. We don't inherit anything from JAX, so it makes sense to rely on the default doc inheritance from sphinx, but we had forgotten about the feature and were using our helper class instead.

PR #696, opened 2020-07-25
fix reshape issue of chain samples

### LINKED ISSUE #671: Cannot reshape the output of chained MCMC
This issue is created to make sure that https://github.com/google/jax/issues/3597 is resolved before we release the next version.

fehiepsi:
Resolves #671.

### DIFF
--- a/numpyro/infer/mcmc.py
+++ b/numpyro/infer/mcmc.py
@@ -7,7 +7,7 @@ from operator import attrgetter
 import os
 import warnings
 
-from jax import jit, lax, pmap, random
+from jax import device_put, jit, lax, pmap, random
 from jax.core import Tracer
 from jax.interpreters.xla import DeviceArray
 from jax.lib import xla_bridge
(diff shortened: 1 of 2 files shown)


numpyro/infer/mcmc.py
```diff
@@ -453,6 +453,8 @@ def run(self, rng_key, *args, extra_fields=(), init_params=None, **kwargs):
                     states, last_state = lax.map(partial_map_fn, map_args)
             elif chain_method == 'parallel':
                 states, last_state = pmap(partial_map_fn)(map_args)
+                # TODO: remove when https://github.com/google/jax/issues/3597 is resolved
+                states = device_put(states)
```
fehiepsi, line 457: @mattjj It seems that using `device_put` fixes the issue https://github.com/google/jax/issues/3597 for us. Do you know if there are side effects if we do this?

mattjj: It'll collect the results into one "device" (the default one), producing a DeviceArray rather than ShardedDeviceArray. That means inter-"device" communication, though on CPU where the devices are just different cores that is at worst some copying, which is probably necessary to form a contiguous array with the result. 

If this gets passed again into a pmapped computation then something inefficient might happen, in that without the device_put there may be less copying/communicating. That's why we try to leave things as ShardedDeviceArrays and only collect them into DeviceArrays when forced to by some downstream operation. But it sounds like in this case that's not a concern!

So this looks good to me. (At some point upstream JAX will be fixed so that this is no longer necessary, and we'll let you know when that happens on the JAX issue.)

fehiepsi: Thanks for your clear explanation, Matt!

PR #697, opened 2020-07-25
Bump to version 0.3.0

fehiepsi:
### Tasks
- [x] draft release notes
- [x] ping jax/jaxlib to the latest version
- [x] make funsor an optional package
- [x] resolve the circular dependency of numpyro and funsor. ~Any idea for this?~
- [x] merge #696 
- [x] check if the tests in https://github.com/fehiepsi/rethinking-numpyro/pull/4 pass
- [x] (optional) merge #460 
- [x] (optional) check if tutorial pages are rendered correctly and include all examples

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,18 +182,13 @@ Pyro users will note that the API for model specification and inference is large
 
 ## Installation
 
-> **Limited Windows Support:** Note that NumPyro is untested on Windows, and might require building jaxlib from source. See this [JAX issue](https://github.com/google/jax/issues/438) for more details. Alternatively, you can install [Windows Subsystem for Linux](https://docs.microsoft.com/en-us/windows/wsl/) as use NumPyro on it as on a Linux system. See also [CUDA on Windows Subsystem for Linux](https://developer.nvidia.com/cuda/wsl) if you want to use GPUs on Windows.
+> **Limited Windows Support:** Note that NumPyro is untested on Windows, and might require building jaxlib from source. See this [JAX issue](https://github.com/google/jax/issues/438) for more details. Alternatively, you can install [Windows Subsystem for Linux](https://docs.microsoft.com/en-us/windows/wsl/) and use NumPyro on it as on a Linux system. See also [CUDA on Windows Subsystem for Linux](https://developer.nvidia.com/cuda/wsl) if you want to use GPUs on Windows.
 
 To install NumPyro with a CPU version of JAX, you can use pip:
 
 ```
 pip install numpyro
 ```
-or
-```
-pip install numpyro[funsor]
-```
-if you want to do inference for models with discrete latent variables.
 
 To use NumPyro on the GPU, you will need to first [install](https://github.com/google/jax#installation) `jax` and `jaxlib` with CUDA support.
 
@@ -204,8 +199,7 @@ You can also install NumPyro from source:
 ```
 git clone https://github.com/pyro-ppl/numpyro.git
 # install jax/jaxlib first for CUDA support
-pip install -e .
-# or `pip install -e .[funsor]
+pip install -e .[dev]  # contains additional dependencies for NumPyro development
 ```
 
 ## Frequently Asked Questions
--- a/README.md
+++ b/README.md
@@ -182,13 +182,18 @@ Pyro users will note that the API for model specification and inference is large
 
 ## Installation
 
-> **Limited Windows Support:** Note that NumPyro is untested on Windows, and will require building jaxlib from source. See this [JAX issue](https://github.com/google/jax/issues/438) for more details.
+> **Limited Windows Support:** Note that NumPyro is untested on Windows, and might require building jaxlib from source. See this [JAX issue](https://github.com/google/jax/issues/438) for more details. Alternatively, you can install [Windows Subsystem for Linux](https://docs.microsoft.com/en-us/windows/wsl/) as use NumPyro on it as on a Linux system. See also [CUDA on Windows Subsystem for Linux](https://developer.nvidia.com/cuda/wsl) if you want to use GPUs on Windows.
 
 To install NumPyro with a CPU version of JAX, you can use pip:
 
 ```
 pip install numpyro
 ```
+or
+```
+pip install numpyro[funsor]
+```
+if you want to do inference for models with discrete latent variables.
 
 To use NumPyro on the GPU, you will need to first [install](https://github.com/google/jax#installation) `jax` and `jaxlib` with CUDA support.
 
@@ -199,7 +204,8 @@ You can also install NumPyro from source:
 ```
 git clone https://github.com/pyro-ppl/numpyro.git
 # install jax/jaxlib first for CUDA support
-pip install -e .[dev]
+pip install -e .
+# or `pip install -e .[funsor]
 ```
 
 ## Frequently Asked Questions
--- a/setup.py
+++ b/setup.py
@@ -36,13 +36,10 @@ setup(
         # TODO: pin to a specific version for the release (until JAX's API becomes stable)
         'jax==0.1.73',
         # check min version here: https://github.com/google/jax/blob/master/jax/lib/__init__.py#L20
-        'jaxlib==0.1.52',
+        'jaxlib==0.1.51',
         'tqdm',
     ],
     extras_require={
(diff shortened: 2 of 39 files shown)


neerajprad:
> resolve the circular dependency of numpyro and funsor. Any idea for this?

@fehiepsi - I don't think funsor should directly depend on `numpyro` or `pyro`. Where do you see this issue?

fehiepsi:
I meant `funsor[jax]` [depends on numpyro](https://github.com/pyro-ppl/funsor/blob/master/setup.py#L50). I guess we can add an if/else [here](https://github.com/pyro-ppl/funsor/blob/master/funsor/jax/distributions.py#L6) to check if `numpyro` is available and remove `numpyro` dependency from `funsor[jax]`?

neerajprad:
> I meant funsor[jax] depends on numpyro. I guess we can add an if/else here to check if numpyro is available and remove numpyro dependency from funsor[jax]?

What happens if we install `funsor` as is and simply let it use whatever version of jax/numpyro is available? Does that cause any issues?

fehiepsi:
If users install funsor alone, then it is fine. But if they install funsor[jax], pip will also install some specific pre-released version of NumPyro. When we release numpyro, it will cause conflict with the pre-released version that has been installed when funsor is installed. I am not sure if this can be resolved easily. The only consistent solution that I know is to release NumPyro 0.3.0rc1, then release funsor[jax] with numpyro>=0.3.0rc1. Then release NumPyro 0.3.

I would prefer the inconsistent solution: release funsor[jax] with numpyro>=0.3.0 and numpyro 0.3.0 at the same time.

Do you have another solution?

neerajprad:
> When we release numpyro, it will cause conflict with the pre-released version that has been installed when funsor is installed. 

What if you release funsor with `numpyro>=0.2.4`? Then when we release numpyro, users should still be able to use the same funsor version with an upgraded `numpyro==0.3.0`. Maybe I don't fully understand the issue.

fehiepsi:
Sure, let me see how much work it needs to make funsor works with numpyro 0.2.4

neerajprad:
> Sure, let me see how much work it needs to make funsor works with numpyro 0.2.4

Oh I see the problem. So funsor requires some features from numpyro's master? I thought that it only uses numpyro's distributions whose interface is relatively static. In that case, I think your earlier suggestion  of pinning to 0.3 and then doing a release sounds better.

README.md
```diff
@@ -199,7 +204,8 @@ You can also install NumPyro from source:
 ```
 git clone https://github.com/pyro-ppl/numpyro.git
 # install jax/jaxlib first for CUDA support
-pip install -e .[dev]
+pip install -e .
+# or `pip install -e .[funsor]
```
fehiepsi, line 208: typo: `

README.md
```diff
@@ -182,13 +182,18 @@ Pyro users will note that the API for model specification and inference is large
...
 To install NumPyro with a CPU version of JAX, you can use pip:
 
 ```
 pip install numpyro
 ```
+or
+```
+pip install numpyro[funsor]
+```
+if you want to do inference for models with discrete latent variables.
```
fehiepsi, line 196: I have changed README a bit to mention about Windows limitedness and funsor installation.

README.md
```diff
@@ -182,13 +182,18 @@ Pyro users will note that the API for model specification and inference is large
 
 ## Installation
 
-> **Limited Windows Support:** Note that NumPyro is untested on Windows, and will require building jaxlib from source. See this [JAX issue](https://github.com/google/jax/issues/438) for more details.
+> **Limited Windows Support:** Note that NumPyro is untested on Windows, and might require building jaxlib from source. See this [JAX issue](https://github.com/google/jax/issues/438) for more details. Alternatively, you can install [Windows Subsystem for Linux](https://docs.microsoft.com/en-us/windows/wsl/) as use NumPyro on it as on a Linux system. See also [CUDA on Windows Subsystem for Linux](https://developer.nvidia.com/cuda/wsl) if you want to use GPUs on Windows.
```
neerajprad, line 185: as -> and

README.md
neerajprad: I think this installation method (`pip install numpyro[funsor]`) should be in `installing from source` below?

README.md
neerajprad: I think it will be nice to include test dependencies into `dev` and mention both `funsor` (for end users experimenting with more advanced machinery) as well as `dev` for users who are looking to do numpyro development either as contributors or for tinkering things on their own system.

setup.py
```diff
@@ -33,22 +33,27 @@
...
-        'jax>=0.1.70',
+        'jax==0.1.73',
         # check min version here: https://github.com/google/jax/blob/master/jax/lib/__init__.py#L20
-        'jaxlib>=0.1.47',
+        'jaxlib==0.1.52',
         'tqdm',
     ],
     extras_require={
+        'funsor': [
+            'funsor @ git+https://github.com/pyro-ppl/funsor.git@b4db46acc5ab615abd2e1297f65ff5e70e961876#egg=funsor',
```
neerajprad, line 44: I suppose this is only a placeholder atm?

fehiepsi: Yes, I will change it to funsor when funsor is released (before this PR is merged).

README.md
fehiepsi: So we will add `funsor` to `dev` and keep `pip install -e .[dev]` here? That looks good to me.

README.md
fehiepsi: yes, will change

README.md
neerajprad: Either is fine. Keeping it separate is okay, we should just mention the use cases.

```
pip install -e .[funsor]  # for experimental `numpyro.contrib.funsor` module for inference on models with discrete variables
pip install -e .[dev]  # contains additional dependencies for NumPyro development
```

setup.py
```diff
@@ -33,22 +33,27 @@
...
     author_email='npradhan@uber.com',
     install_requires=[
-        # TODO: use the release version of funsor for the release
-        'funsor @ git+https://github.com/pyro-ppl/funsor.git@b4db46acc5ab615abd2e1297f65ff5e70e961876#egg=funsor',
         # TODO: pin to a specific version for the release (until JAX's API becomes stable)
-        'jax>=0.1.70',
+        'jax==0.1.73',
         # check min version here: https://github.com/google/jax/blob/master/jax/lib/__init__.py#L20
-        'jaxlib>=0.1.47',
+        'jaxlib==0.1.52',
```
fehiepsi, line 39: use `0.1.51` because 0.1.52 seems to fail in mac os.

setup.py
```diff
@@ -34,9 +34,9 @@
     author_email='npradhan@uber.com',
     install_requires=[
         # TODO: pin to a specific version for the release (until JAX's API becomes stable)
-        'jax>=0.1.70',
+        'jax==0.1.73',
         # check min version here: https://github.com/google/jax/blob/master/jax/lib/__init__.py#L20
-        'jaxlib>=0.1.47',
+        'jaxlib==0.1.51',
```
fehiepsi, line 39: Pin to 0.1.51 due to https://github.com/google/jax/issues/3867

setup.py
```diff
@@ -47,7 +47,7 @@
             'pyro-api>=0.1.1'
         ],
         'dev': [
-            'funsor @ git+https://github.com/pyro-ppl/funsor.git@b4db46acc5ab615abd2e1297f65ff5e70e961876#egg=funsor',
+            'funsor',
```
neerajprad, line 50: Should we specify the version number (>= current version)?

neerajprad: We can do this later, it doesn't block us from merging this and releasing.

neerajprad:
Good to merge?

fehiepsi:
Yeah!!!!

fehiepsi:
After this, could you help me render the tutorial site to update new examples?

neerajprad:
> After this, could you help me render the tutorial site to update new examples?

Sure, let me try to build it first.

neerajprad:
Okay, tutorials should be up. Could you check and see if everything looks good?

fehiepsi:
They look good to me. Thanks for the help!

Issue #703, opened 2020-07-28
Implement some popular distributions


fehiepsi:
This issue tracks some popular distributions that are not available in NumPyro yet.

- [x] [Geometric distribution](https://pytorch.org/docs/master/distributions.html#geometric)
- [x] [ZeroInflatedNegativeBinomial](http://docs.pyro.ai/en/stable/distributions.html#zeroinflatednegativebinomial)
- [x] [Weibull distribution](https://pytorch.org/docs/master/distributions.html#weibull) (supported in `contrib.tfp`)
- [x] [DirichletMultinomial](http://docs.pyro.ai/en/stable/distributions.html#dirichletmultinomial) (supported in `contrib.tfp`)
- [ ] [FoldedDistribution](http://docs.pyro.ai/en/stable/distributions.html#foldeddistribution)

TuanNguyen27:
TFP also has [Geometric ](https://www.tensorflow.org/probability/api_docs/python/tfp/distributions/Geometric) & [Weibull](https://www.tensorflow.org/probability/api_docs/python/tfp/distributions/Weibull) distribution support, in case #710 is successful :)

fehiepsi:
Yeah, if those distributions in `numpyro.contrib.tfp.distributions` work well, we will probably move them to the main module. There are some interesting distributions in tfp such as [JointDistributionSequential](https://www.tensorflow.org/probability/api_docs/python/tfp/distributions/JointDistributionSequential).

nkaimcaudle:
Is it possible to also add [SinhArcsinh transformation](https://www.tensorflow.org/probability/api_docs/python/tfp/bijectors/SinhArcsinh) ?

fehiepsi:
@nkaimcaudle Sure, I just created a FR for it. :)

dirmeier:
Ciao all, 

I'm working on two models with ZINB and Weibull likelihoods right now, if you are interested to have implementations for both of these, I could open up a PR. 

Maybe it could also make sense to have the alternative parameterization of the NB (https://mc-stan.org/docs/2_26/functions-reference/nbalt.html), since that is the one mainly used for regression models.

Best,
Simon

fehiepsi:
Hi @dirmeier, your proposals sound great to me. Regarding NB, you could name the other parameterization `NegativeBinomial2`. :)

PR #709, opened 2020-08-05
Implement geometric distribution

xidulu:
https://github.com/pyro-ppl/numpyro/issues/703

As title,

Implementation largely follows torch.distributions.Geometric.

p.s. I know something about torch.distributions but I am relatively new to Jax (still learning the best practice of using Jax.)


numpyro/distributions/util.py
```diff
@@ -215,6 +215,20 @@ def multinomial(key, p, n, shape=()):
     return _multinomial(key, p, n, n_max, shape)
 
 
+@partial(jit, static_argnums=(2,))
+def _geometric(key, p, shape):
+    if shape is None:
+        shape = jnp.shape(p)
+    dtype = get_dtype(p)
+    tiny = jnp.finfo(dtype).eps
+    u = random.uniform(key, shape, dtype, tiny)
```
fehiepsi, line 224: nit: because 1 is a valid value for `u`, I guess we can take `u = random.uniform(key, shape, dtype)` and compute `jnp.log1p(-u)` instead of `jnp.log(u)` without having to clamp the value.

numpyro/distributions/util.py
```diff
@@ -215,6 +215,20 @@ def multinomial(key, p, n, shape=()):
...
+    return jnp.floor(jnp.log(u) / jnp.log1p(-p))
```
fehiepsi, line 225: nit: I think it is better to move this implementation to `sample` method.

numpyro/distributions/discrete.py
```diff
@@ -592,3 +593,76 @@ def mean(self):
...
+    def sample(self, key, sample_shape=()):
+        return geometric(key, self.probs, sample_shape + self.batch_shape)
+
+    @validate_sample
+    def log_prob(self, value):
+        if self._validate_args:
+            self._validate_sample(value)
+        value, probs = promote_shapes(value, self.probs)
+        probs = jnp.where((probs == 1) & (value == 0), 0, probs)
+        return value * jnp.log1p(-probs) + jnp.log(probs)
```
fehiepsi, line 651: You can use [this implementation](https://github.com/pyro-ppl/pyro/blob/dev/pyro/distributions/torch.py#L161) for GeometricLogits. I found it more stable.

numpyro/distributions/discrete.py
```diff
@@ -592,3 +593,76 @@ def mean(self):
...
+    def probs(self):
+        return _to_probs_bernoulli(self.logits)
+
+    def sample(self, key, sample_shape=()):
+        return geometric(key, self.probs, sample_shape + self.batch_shape)
+
+    @validate_sample
+    def log_prob(self, value):
+        if self._validate_args:
+            self._validate_sample(value)
```
fehiepsi, line 648: In NumPyro, you can remove this check (because it is handled in `@validate_sample` decorator).

numpyro/distributions/discrete.py
```diff
@@ -592,3 +593,76 @@ def mean(self):
...
+                                             validate_args=validate_args)
+
+    def sample(self, key, sample_shape=()):
+        return geometric(key, self.probs, sample_shape + self.batch_shape)
+
+    @validate_sample
+    def log_prob(self, value):
+        if self._validate_args:
+            self._validate_sample(value)
+        value, probs = promote_shapes(value, self.probs)
```
fehiepsi, line 615: nit: no need to promote shapes

numpyro/distributions/discrete.py
```diff
@@ -592,3 +593,76 @@ def mean(self):
...
     def variance(self):
         return (1 - self.gate) * self.rate * (1 + self.rate * self.gate)
+
+
+class GeometricProbs(Distribution):
+    arg_constraints = {'probs': constraints.unit_interval}
+    support = constraints.nonnegative_integer
+    is_discrete = True
+
+    def __init__(self, probs=None, validate_args=None):
```
fehiepsi, line 603: nit: `probs=None` -> `probs`

numpyro/distributions/discrete.py
```diff
@@ -592,3 +593,76 @@ def mean(self):
...
+    def variance(self):
+        return (1. / self.probs - 1.) / self.probs
+
+
+class GeometricLogits(Distribution):
+    arg_constraints = {'logits': constraints.real}
+    support = constraints.nonnegative_integer
+    is_discrete = True
+
+    def __init__(self, logits=None, validate_args=None):
```
fehiepsi, line 633: nit: `logits=None` -> `logits`

numpyro/distributions/discrete.py
xidulu: That looks great, didn't find that in the torch repo.

numpyro/distributions/util.py
xidulu: But does that mean JIT will not be used? I actually follow the practice of other distributions such as Categorical to put the sampling method in util.py.

numpyro/distributions/util.py
xidulu: I follow the implementation in torch https://github.com/pytorch/pytorch/blob/master/torch/distributions/geometric.py#L78
Also, could `log1p(-u)` be little unstable when `u -> 1`?

fehiepsi: In most frameworks, `uniform` samples values in [0, 1), so it should be fine. We can use the trick `1 - u` to get a uniform sample in (0, 1]. If you prefer to sample from (eps, 1), feel free to ignore my comment.

numpyro/distributions/util.py
fehiepsi: Actually, if you sample from `Geometric`, `random.uniform` will be compiled for you. We often put complicated samplers to `util` to simplify codes in the main implementation. We will move categorical sampler to `CategoricalProbs.sample` when its performance is improved (currently, it is slow when the number of categories are large).

fehiepsi: Just a side note: when you move this sampler to GeometricLogits, you can replace `jnp.log1p(-p)` by `-soft_plus(logits)`

numpyro/distributions/util.py
xidulu: Oh, I got that.

Issue #704, opened 2020-07-28
Integrate with Tensorflow Probability distribution


fehiepsi:
TFP has some [subtraces](https://www.tensorflow.org/probability/api_docs/python/tfp/substrates/jax/distributions), including JAX. It is helpful to build a layer so that users can use TFP distributions in NumPyro. This helps avoid duplication of effort and settles some collaborations between two communities in the future.

It is nice to make TFP Normal distribution works in NumPyro first. Probably we can start with `numpyro.contrib.tfp.distributions.Normal`...

TuanNguyen27:
While working on #705 I found [Oryx](https://github.com/tensorflow/probability/tree/master/spinoffs/oryx) ; [documentation](https://www.tensorflow.org/probability/oryx/notebooks/a_tour_of_oryx) which is another good candidate for integration with numpyro :D

fehiepsi:
@TuanNguyen27 Do you have some suggestions on how to integrate it in NumPyro? The ability to transform a sampling function to a log probability function looks pretty cool! It would be great if we can use it to simplify some implementations or enable a new way to do inference.

TuanNguyen27:
I don't know numpyro internals well enough to make suggestion, but if you have an example implementation in numpyro that could be replaced by the submodules implemented in Oryx I can prototype something out :D I am interested in both refactoring numpyro & new inference !

fehiepsi:
@TuanNguyen27 For now, I don't have any idea in mind. :(

TuanNguyen27:
PPS: We have another candidate, https://github.com/google/objax :D

PR #710, opened 2020-08-11
Support tensorflow probability

fehiepsi:
Initial support for #704. MCMC just works out of the box. :D

I don't use tensorflow probability so I don't add tests for it throughout. But I am happy to fix any issue that users get while using `numpyro.contrib.tfp.distributions`.

NB: some changes at `infer/initialization.py` are necessary to make sure it work with tfp distributions `sample` method.

One issue with tfp is those distributions do not have `support` or `enumerate_support` attributes. I resolve it by setting `support=real` by default and copying NumPyro distribution's support to the corresponding TFP distribution.

After this merged, I will add a note in forum to call for tests from users.


test/contrib/test_tfp.py
```diff
@@ -0,0 +1,59 @@
...
+from numpy.testing import assert_allclose
+
+import jax.numpy as jnp
+from jax import random
+
+import numpyro
+from numpyro.infer import MCMC, NUTS
+
+
+@pytest.mark.filterwarnings("ignore")
```
fehiepsi, line 14: `import tfp` raises some warnings so we filter it here

numpyro/contrib/tfp/distributions.py
```diff
@@ -0,0 +1,56 @@
...
+            # resolve FooProbs/FooLogits namespaces
+            if type(numpyro_dist_class).__name__ == "function":
+                if not hasattr(numpyro_dist, _name + "Logits"):
+                    continue
+                numpyro_dist_class = getattr(numpyro_dist, _name + "Logits")
+            _PyroDist.support = numpyro_dist_class.support
+            _PyroDist.enumerate_support = numpyro_dist_class.enumerate_support
+            _PyroDist.arg_constraints = numpyro_dist_class.arg_constraints
+            _PyroDist.has_enumerate_support = numpyro_dist_class.has_enumerate_support
+            _PyroDist.is_discrete = numpyro_dist_class.is_discrete
```
eb8680, line 45: How much of this can be automated using methods from the TFP distributions? That seems like a key step in replacing current NumPyro distributions. 

It looks like all TFP distributions implement the methods [`param_shapes`](https://www.tensorflow.org/probability/api_docs/python/tfp/distributions/Distribution#param_shapes) and `_default_event_space_bijector`, which I believe can be used to infer (part of?) `arg_constraints` and `support` respectively.

.github/workflows/ci.yml
```diff
@@ -64,6 +64,7 @@ jobs:
         pip install https://github.com/pyro-ppl/funsor/archive/master.zip
         pip install jaxlib
         pip install jax
+        pip install tensorflow-cpu tfp-nightly
```
eb8680, line 67: Does TFP depend on tensorflow by default, or is that optional?

brianwa84: The TF dependency is optional. If all you want is jax, `from tensorflow_probability.substrates import jax as tfp`

fehiepsi: Thanks @brianwa84! It is so helpful.

neerajprad: @fehiepsi - can we just `pip install tensorflow_probability tfp-nightly` here?

numpyro/contrib/tfp/distributions.py
```diff
@@ -0,0 +1,56 @@
...
+for _name, _Dist in tfd.__dict__.items():
+    if not isinstance(_Dist, type):
+        continue
+    if not issubclass(_Dist, tfd.Distribution):
+        continue
+    if _Dist is tfd.Distribution:
+        continue
+
+    try:
+        _PyroDist = locals()[_name]
```
neerajprad, line 29: I don't think we should need this since this module doesn't have any distribution in the `locals()` scope.

numpyro/contrib/tfp/distributions.py
neerajprad: As a temporary measure, we can also have very shallow wrappers for specific distributions like we have in `pyro.distributions.torch`, which will allow us to remove our own distributions code and slowly thin out the wrappers over time.

fehiepsi: Yeah, I think we can leverage `_default_event_space_bijector` to define default support in the Mixin class and have shallow wrappers for extra properties.

.github/workflows/ci.yml
fehiepsi: Yes, I think we can remove `tensorflow-cpu`. Let me remove it in the next commit to see if there is any issue happens.

numpyro/contrib/tfp/distributions.py
fehiepsi: That's right! But we will add some wrappers soon. :D

fehiepsi:
@neerajprad I need to incorporate you and Eli's suggestions but didn't find time to look into more details yet. :( I will try to address this soon.

fehiepsi:
Closed in favor of #718.

PR #713, opened 2020-08-18
add example for ordinal regression

vanAmsterdam:
this is an update to https://github.com/pyro-ppl/numpyro/pull/619

I got some merge conflicts locally so I deleted the branch and created this fresh PR


notebooks/source/ordinalregression.ipynb
```diff
@@ -0,0 +1,239 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Ordinal regression with the ImproperUniform distribution"
```
fehiepsi, line 7: I think it is better to use "Ordinal Regression" for further enhancements in the future. For example, it is not necessary to restrict ourselves to "improper" world, where we can add any extra log prob terms to the joint log probability (as in the `model2`) to create an "unnormalized" distribution. I can add a "proper" version in a follow-up PR to clarify that.

notebooks/source/ordinalregression.ipynb
```diff
@@ -0,0 +1,239 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "metadata": {},
+   "outputs": [
+    {
+     "data": {
+      "text/plain": [
+       "<AxesSubplot:xlabel='Y', ylabel='X'>"
```
fehiepsi, line 86: I think you can add `;` at the end of the code `sns.violinplot(x='Y', y='X', data=df)` to remove this text.

notebooks/source/ordinalregression.ipynb
```diff
@@ -0,0 +1,239 @@
...
+    "from numpyro.infer.hmc import NUTS\n",
+    "import pandas as pd\n",
+    "import seaborn as sns"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "First, generate data some with ordinal structure"
```
fehiepsi, line 39: typo: `data some`

notebooks/source/ordinalregression.ipynb
```diff
@@ -0,0 +1,239 @@
...
+    "mcmc   = MCMC(kernel, num_warmup=250, num_samples=750)\n",
+    "mcmc.run(mcmc_key, X,Y, nclasses)\n",
+    "mcmc.print_summary()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "If we have additional information on the parameters that we want to encode in a prior, we can add a `sample` statement that uses a 'regular' prior with an `obs` argument, conditioning the parameter drawn from the `ImproperUniform` on the chosen prior"
```
fehiepsi, line 164: It would be great if you explain this approach a little bit more. I don't have much idea but something like mentioning that this is equivalent to using an "unnormalized" prior (created by restricting the support of a Normal distribution to the "ordered" domain). :)

notebooks/source/ordinalregression.ipynb
```diff
@@ -0,0 +1,239 @@
...
+   "outputs": [],
+   "source": [
+    "from jax import numpy as np, random\n",
+    "import numpyro\n",
+    "from numpyro import sample\n",
+    "from numpyro.distributions import constraints, Normal, ImproperUniform, Categorical, OrderedLogistic\n",
+    "from numpyro.infer.mcmc import MCMC\n",
+    "from numpyro.infer.hmc import NUTS\n",
+    "import pandas as pd\n",
+    "import seaborn as sns"
```
fehiepsi, line 32: Could you add the assertation `assert numpyro.__version__.startswith('0.3.0')` at the end of the code cell?

notebooks/source/ordinalregression.ipynb
vanAmsterdam: I've tried to elaborate on this a bit. Is is better phrased right now? Do you have additional suggestions?

fehiepsi: Thanks, @vanAmsterdam! It look great to me.

PR #717, opened 2020-08-27
Integrate haiku.Module and flax.nn.Module with numpyro

### LINKED ISSUE #705: Integrate with flax/haiku
Currently, we support [stax](https://jax.readthedocs.io/en/latest/jax.experimental.stax.html) but it lacks some popular layers such as RNN. Some other candidates include [flax](https://github.com/google/flax), [haiku](https://github.com/deepmind/dm-haiku). It might not take much effort to use those libraries in NumPyro.

We might start with some utilities/handlers in `numpyro.contrib.flax` or `numpyro.contrib.haiku`, add some examples (which illustrates how to set priors for some parameters of neural nets), and gradually enhance the API.

Some references:
+ [lift](http://num.pyro.ai/en/stable/handlers.html#lift) handler: converts a `param` site to a `sample` site
+ [module](http://num.pyro.ai/en/stable/primitives.html#module) primitive: move nn's parameters to `param` statement
+ [VAE example](http://pyro.ai/numpyro/examples/vae.html)

TuanNguyen27:
Closes #705


TuanNguyen27:
Thanks for the review! Will update my PR with tests shortly :) 

> * those primitives to the [docs](https://github.com/pyro-ppl/numpyro/blob/master/docs/source/primitives.rst) (after `module` entry)

Do i need to rename the original `numpyro.module` to `numpyro.stax_module` ?

fehiepsi:
I think eventually, we will move those implementations to `numpyro.module` (probably with a keyword to identify which package we will use. So it is better to keep `numpyro.module` as is.

TuanNguyen27:
@fehiepsi I am running into some lint problem: In `init.py` for both `haiku_module` and `flax_module` I am testing the import of the library, but i don't actually use it for anything and flake8 is not happy. Is there a way to suppress this ? 

```
try:
    import flax
except ImportError:
    raise ImportError("Looking like you want to use flax and/or haiku to declare "
                      "nn modules. This is an experimental feature. "
                      "You need to install `flax` to be able to use this feature. "
                      "It can be installed with `pip install flax`.")
```

fehiepsi:
@TuanNguyen27 You can add `# noqa: F401` at the end of the `import` line (assume F401 is the lint error)

numpyro/contrib/haiku_module/nn.py
```diff
@@ -0,0 +1,32 @@
...
+    module_key = name + '$params'
+    nn_params = numpyro.param(module_key)
+    if nn_params is None:
+        if input_shape is None:
+            raise ValueError('Valid value for `input_shape` needed to initialize.')
+        # feed in dummy data to init params
+        rng_key = numpyro.sample(name + '$rng_key', PRNGIdentity())
+        nn_params = nn.init(rng_key, jnp.ones(input_shape))
+        numpyro.param(module_key, nn_params)
+    return lambda x: nn.apply(nn_params, rng_key, x)
```
fehiepsi, line 32: I am curious why `rng_key` is necessary for `haiku`? Is it used for `Dropout` or something? If it is important, we need to split the keys, otherwise, it would be nice if we can set it to `None`.

numpyro/contrib/haiku_module/__init__.py
```diff
@@ -0,0 +1,7 @@
+try:
```
fehiepsi, line 1: If you decide to split out `flax` and `haiku`, it is better to rename those modules to `flax` and `haiku` so we can do `from numpyro.contrib.flax import flax_module` instead of `from numpyro.contrib.flax_module import flax_module`.

numpyro/contrib/haiku_module/nn.py
TuanNguyen27: indeed it could be set to None, rng key is optional for `.apply` :)

TuanNguyen27:
@fehiepsi using partial works, and the `haiku_module` is behaving the same as `flax_module` ! Do you want me to implement `random_module` in this PR as well, or should i make a separate PR for that ?

numpyro/contrib/nn.py
```diff
@@ -0,0 +1,55 @@
+# Copyright Contributors to the Pyro project.
+# SPDX-License-Identifier: Apache-2.0
+
+from flax import nn as flax_nn
```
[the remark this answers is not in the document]
fehiepsi: @TuanNguyen27 I guess I made you confused. Here I just meant to move those imports into `flax_module` function, kind of
```python
def flax_module(...):
    try:
        import flax
    except:
        ...
```

With this, all `flax_module`, `haiku_module`, `random_module` primitives are gathered into a single namespace `numpyro.contrib.nn`. By the way, is it better to rename this namespace to `numpyro.contrib.module` (because `numpyro.nn` is used for nn layers, rather than primitives)?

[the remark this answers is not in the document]
TuanNguyen27: ahhh i see what you mean now! Does this module still need `__init__.py` ? We can leave that empty too right ?

fehiepsi:
I think it is better to defer it to another PR. There, you can add some nice examples based on your experience with `pyvarinf` (by turning a nn to a bayesian nn with e.g. `random_module(..., prior=dist.Normal(0, 0.1))` and using AutoDiagonalNormal to do inference, we achieve the same functionality). ;)

numpyro/contrib/nn.py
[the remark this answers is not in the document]
fehiepsi: > Does this module still need __init__.py?

If you make `module` a folder then `__init__` is necessary but you need to expose `flax_module`, `haiku_module`, `random_module` functions to this init file (so users can do `from numpyro.contrib.module import ...` - without doing so, users have to do `from numpyro.contrib.module.flax_module import flax_module`). Because the implementation is short, I prefer to put `flax_module`, `haiku_module`, `random_module` functions in `contrib/module.py` file for simplicity. But it depends on you.

TuanNguyen27:
@fehiepsi should i also add the notebook where i have flax / haiku working with VAE example ?

test/contrib/test_module.py
```diff
@@ -0,0 +1,43 @@
...
+    linear_module = flax.nn.Dense.partial(features=1)
+    nn = flax_module("nn", linear_module, 1)
+    mean = nn(x)
+    numpyro.sample("y", numpyro.distributions.Normal(mean, 0.1), obs=y)
+
+
+def test_module():
+    with handlers.trace() as flax_tr, handlers.seed(rng_seed=1):
+        flax_model(X, Y)
+    flax_params = flax_tr["nn$params"]
```
fehiepsi, line 35: nit: To get the actual value of `param` site, it is clearer to use `flax_tr["nn$params"]["value"]`. In some cases, it is different from `flax_tr["nn$params"]["args"][0]` (e.g. when we substitute some values to those param sites.

test/contrib/test_module.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+def flax_model(x, y):
+    linear_module = flax.nn.Dense.partial(features=1)
+    nn = flax_module("nn", linear_module, 1)
+    mean = nn(x)
+    numpyro.sample("y", numpyro.distributions.Normal(mean, 0.1), obs=y)
+
+
+def test_module():
```
fehiepsi, line 32: Could you separating this into 2 different tests `test_flax_module` and `test_haiku_module`?

test/contrib/test_module.py
```diff
@@ -0,0 +1,43 @@
+# Copyright Contributors to the Pyro project.
+# SPDX-License-Identifier: Apache-2.0
+
+import haiku as hk
+import flax
+from numpyro.contrib.module import haiku_module, flax_module
+import numpy as np
+from numpyro import handlers
+import numpyro
```
fehiepsi, line 9: tip: you can run `isort` to sort those import statements.

test/contrib/test_module.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+X = np.array(list(range(100)))
+Y = np.array([x*2 + 2.0 for x in list(range(100))])
+param_keys = ['type', 'name', 'fn', 'args', 'kwargs', 'value', 'scale', 'cond_indep_stack']
+
+
+def haiku_model(x, y):
+    linear_module = hk.transform(lambda x: hk.Linear(1)(x))
+    nn = haiku_module("nn", linear_module, 1)
```
fehiepsi, line 19: nit: `1` -> `(1,)` to make it clear that this in a shape)

test/contrib/test_module.py
```diff
@@ -0,0 +1,43 @@
...
+    assert nn.args[0]
+    mean = nn(x)
+    numpyro.sample("y", numpyro.distributions.Normal(mean, 0.1), obs=y)
+
+
+def flax_model(x, y):
+    linear_module = flax.nn.Dense.partial(features=1)
+    nn = flax_module("nn", linear_module, 1)
```
fehiepsi, line 27: nit: `1` -> `(1,)`

test/contrib/test_module.py
```diff
@@ -0,0 +1,43 @@
...
+
+import haiku as hk
+import flax
+from numpyro.contrib.module import haiku_module, flax_module
+import numpy as np
+from numpyro import handlers
+import numpyro
+
+
+X = np.array(list(range(100)))
```
fehiepsi, line 12: nit: np.arange(100)

test/contrib/test_module.py
```diff
@@ -0,0 +1,43 @@
...
+Y = np.array([x*2 + 2.0 for x in list(range(100))])
```
fehiepsi, line 13: nit: `X = np.arange(100)`, `Y = 2 * x + 2`, no need for `param_keys`. I think you can move those global variables to each of `test_flax_module`, `test_haiku_module`.

test/contrib/test_module.py
TuanNguyen27: separate functions within `test_module.py`, or two separate files ?

TuanNguyen27:
@fehiepsi it is tricky to install haiku on github actions...it doesn't have the official pip install haiku, and instead i needed to do `pip install git+https://github.com/deepmind/dm-haiku` but it seems to be breaking on github action.

fehiepsi:
Oh, so `haiku` does not have pypi release yet. To pin to a specific github version, we usually add
```
dm-haiku @ https://github.com/deepmind/dm-haiku/archive/v0.0.2.zip
```
or if the released version 0.0.2 is not compatible with `numpyro`,
```
dm-haiku @ git+https://github.com/deepmind/dm-haiku/@[commit_hash]#egg=dm-haiku
```
here `[commit_hash]` is the latest commit in master branch (e.g. at the time I wrote this comment: `db32b3974eed081779a5edd0d9ff56b9a76c8260`)

test/contrib/test_module.py
fehiepsi: I meant into 2 functions. We rarely combine tests unless doing so will improve testing time.

PR #719, opened 2020-08-30
Support BFGS minimizer

### LINKED ISSUE #716: Support jax.scipy minimizers in NumPyro
After jax 0.1.74, `bfgs` minimizer is available in JAX. It would be nice to have a BFGS wrapper in `numpyro.optim`.

In jax/scipy, 'BFGS' is provided through `method` keyword in [minimize](https://docs.scipy.org/doc/scipy/reference/optimize.minimize-bfgs.html#optimize-minimize-bfgs). I am not sure if we should make a wrapper `numpyro.optim.Minimize` for `minimize` or we should split those methods into separated classes such as `numpyro.optim.BFGS`. With the former, new minimizers in JAX will be supported in NumPyro without modifications. With the latter, it is more explicit and friendlier to PyTorch users. Either way looks good to me.

cc @Joshuaalbert who implemented BFGS minimizer in JAX

fehiepsi:
Resolves #716 

This PR adds Minimize optimizer. To support this optimizer, I added a new method `.step` in `NumPyroOptim` because the current `update` method only updates states based on the current gradients, while in BFGS, we need to reevaluate the function multiple time to achieve the optimal value.

Test should be covered in doctest.


fritzo:
Hi @fehiepsi , I see you are adding a lot of new interfaces in NumPyro. Can we discuss in the next Pyro dev meeting whether we should port any of the new interfaces back to Pyro, or possibly make alternative interfaces in Pyro so that it is easier to write backend-agnostic models?

fehiepsi:
Sure, @fritzo! There are a bunch of new stuffs to discuss. :D

numpyro/optim.py
```diff
@@ -55,6 +60,23 @@ def update(self, g: _Params, state: _IterOptState) -> _IterOptState:
         opt_state = self.update_fn(i, g, opt_state)
         return i + 1, opt_state
 
+    def step(self, fn: Callable, state: _IterOptState) -> _IterOptState:
```
neerajprad, line 63: This seems clean, but the `step` / `update` names seem a bit arbitrary. Another option could be to add `fn=None` keyword argument to `update`.

fehiepsi: Good idea!! I think we should go for that option.

fehiepsi: @neerajprad One issue with move the implementation to `update` is the output will be inconsistent. Here `step` returns both `update` output and loss value. How about renaming `step` to something else?

neerajprad (approved):

numpyro/optim.py
neerajprad: Agreed, the new name seems more suitable.

PR #721, opened 2020-08-31
Add a thin wrapper to use TFP mcmc kernels for NumPyro models

### LINKED ISSUE #715: Support TFP mcmc kernels
TFP [kernels](https://www.tensorflow.org/probability/api_docs/python/tfp/substrates/jax/mcmc/RandomWalkMetropolis#one_step) takes potential function in its construction and has the method [one_step](https://www.tensorflow.org/probability/api_docs/python/tfp/substrates/jax/mcmc/RandomWalkMetropolis#one_step) which update `(current_state, previous_kernel_results, seed)`. This is equivalent to how [MCMCKernel](http://num.pyro.ai/en/stable/mcmc.html#numpyro.infer.mcmc.MCMCKernel) works in NumPyro.

It should be straightforward to add wrappers in `contrib.tfp` for TFP MCMC kernels because both api are equivalent. However, in NumPyro, we support both `model` and `potential_fn` so I think the wrappers in `contrib.tfp.mcmc` should take the same name but `target_log_prob_fn` will be replaced by `model=None, potential_fn=None` (all other args, kwargs are kept the same).

While looking into this, I also think that we should make `HMCState`, `SAState` a default attribute of their corresponding classes. This centralizes all information related to kernel to 1 class, rather than separated into a pair of (state_definition, kernel_class) as in the current implementation.

WDYT about the above 2 issues @neerajprad?

After this, we can use tfp mcmc kernels for NumPyro models. :D

fehiepsi:
Resolves #715.

~This also fixes a bug in the current MCMC kernels, where init_strategy is not respected.~

### Tasks

- [x] Implement the mixin class
- [x] Loop over all TFP kernels and subclass the mixin
- [x] Add tests


numpyro/contrib/tfp/mcmc.py
```diff
@@ -0,0 +1,194 @@
...
+
+    def init(self, rng_key, num_warmup, init_params=None, model_args=(), model_kwargs={}):
+        # non-vectorized
+        if rng_key.ndim == 1:
+            rng_key, rng_key_init_model = random.split(rng_key)
+        # vectorized
+        else:
+            rng_key, rng_key_init_model = jnp.swapaxes(vmap(random.split)(rng_key), 0, 1)
+        init_params = self._init_state(rng_key_init_model, model_args, model_kwargs, init_params)
+        if self._potential_fn and init_params is None:
```
neerajprad, line 127: I see that you are using `vmap` for vectorization - are tfp mcmc kernels not auto vectorized?

test/contrib/test_tfp.py
```diff
@@ -102,3 +103,62 @@ def model(data):
     mcmc.print_summary()
     samples = mcmc.get_samples()
     assert_allclose(jnp.mean(samples['p_latent'], 0), true_probs, atol=0.05)
+
+
+@pytest.mark.parametrize('kernel, kwargs', [
+    ('RandomWalkMetropolis', dict()),
+    ('SliceSampler', dict(step_size=1.0, max_doublings=5))
```
neerajprad, line 110: How many other samplers are there, and we tested if the wrappers work as expected?

numpyro/contrib/tfp/mcmc.py
fehiepsi: I just took a look at tfp mcmc. It seems there is another utility [sample_chain](https://www.tensorflow.org/probability/api_docs/python/tfp/substrates/jax/mcmc/sample_chain) for it. It seems to me that it just does the same job here.

test/contrib/test_tfp.py
fehiepsi: There are several more. Each of them has different `kwargs`. I'll loop over them and add corresponding kwargs in the test.

fehiepsi: @neerajprad I have added tests for remaining kernels. Only [ReplicaExchangeMC](https://www.tensorflow.org/probability/api_docs/python/tfp/substrates/jax/mcmc/ReplicaExchangeMC) is not working due to some of its reshape logics. I think we need to do something like `vmap(potential_fn)` so the potential_fn can be evaluated on multiple replicas of params. I will try to support it in a follow-up PR if possible...

fehiepsi: Just want to update, ReplicaExchangeMC should work now. I think we have supported all TFP kernels. And any user-defined tfp kernel can be converted to a NumPyro compatible kernel with `TFPKernel[custom_kernel]`. Cheers, :D

numpyro/contrib/tfp/mcmc.py
```diff
@@ -0,0 +1,233 @@
...
+for _name, _Kernel in tfp.mcmc.__dict__.items():
+    if not isinstance(_Kernel, type):
+        continue
+    if not issubclass(_Kernel, tfp.mcmc.TransitionKernel):
+        continue
+    if 'target_log_prob_fn' not in inspect.getfullargspec(_Kernel).args:
+        continue
+
+    try:
+        _PyroKernel = locals()[_name]
```
neerajprad, line 209: Since we aren't importing any numpyro mcmc kernels here, is this for a future use case?

numpyro/contrib/tfp/mcmc.py
```diff
@@ -0,0 +1,233 @@
...
+import jax.numpy as jnp
+import tensorflow_probability.substrates.jax as tfp
+
+from numpyro.infer import init_to_uniform
+from numpyro.infer.mcmc import MCMCKernel
+from numpyro.infer.util import initialize_model
+from numpyro.util import identity
+
+
+TFPKernelState = namedtuple('TFPKernelState', ['z', 'kernel_results', 'rng_key'])
```
neerajprad, line 19: Any idea how these tfp kernels works with CUDA? Can we set the jax device and would things automatically run on the GPU?

neerajprad:
You'll need to merge in changes from master. Looks great to me, it is actually surprising that we are able to reuse so much of the excellent work done by the TFP folks. I just had some comments for my own understanding.

numpyro/contrib/tfp/mcmc.py
fehiepsi: Yes, I just test that we can switch device using `numpyro.set_platform` as usual

numpyro/contrib/tfp/mcmc.py
fehiepsi: Thanks, it is unnecessary codes. I just remove it.

fehiepsi:
Thanks for reviewing, @neerajprad!!

neerajprad (approved):

PR #725, opened 2020-09-07
Relax summing up simplex constraint to align with Pyro.

elchorro:
Due to machine precision draws from Dirichlet are sometimes rejected as they fail the simplex constraint. Current Numpyro's implementation does not relax the summing up constraint on both sides.


fehiepsi (approved):
Thanks for the fix, @elchorro!

PR #732, opened 2020-09-10
Mention about the default platform in README

### LINKED ISSUE #727: NumPyro disables GPU on import
When numpyro is imported, the platform is set to 'cpu' by default in the init file. 
This was very confusing to me since suddenly jax stopped recognising my GPU--only after a bit of digging did I discover it was due to the numpyro import statement. 

I think it would make sense to either display a warning or reconsider changing the platform, since it's a bit unexpected for a module to change some global state on import.

fehiepsi:
Resolves #727. This PR adds a note to README to mention that we use CPU by default, which is different from JAX default behavior. Is there a better idea?

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,7 +194,7 @@ To use NumPyro on the GPU, you will need to first [install](https://github.com/g
 
 To run NumPyro on Cloud TPUs, you can use pip to install NumPyro as above and setup the TPU backend as detailed [here](https://github.com/google/jax/tree/master/cloud_tpu_colabs).
 
-> **Default Platform:** Different from JAX, which uses GPU as the default platform, we use CPU as the default platform. You can use [set_platform](http://num.pyro.ai/en/stable/utilities.html#set-platform) utility to switch to other platforms such as GPU or TPU at the begin of your program.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -194,6 +194,8 @@ To use NumPyro on the GPU, you will need to first [install](https://github.com/g
 
 To run NumPyro on Cloud TPUs, you can use pip to install NumPyro as above and setup the TPU backend as detailed [here](https://github.com/google/jax/tree/master/cloud_tpu_colabs).
 
+> **Default Platform:** Different from JAX, which uses GPU as the default platform, we use CPU as the default platform. You can use [set_platform](http://num.pyro.ai/en/stable/utilities.html#set-platform) utility to switch to other platforms such as GPU or TPU at the begin of your program.
```
martinjankowiak, line 197: In contrast to JAX, which uses GPU as the default platform, we use CPU as the default platform. You can use the [set_platform](http://num.pyro.ai/en/stable/utilities.html#set-platform) utility to switch to other platforms such as GPU or TPU at the beginning of your program

fehiepsi: Thanks Martin! I made the change. :D

martinjankowiak (approved):

PR #733, opened 2020-09-10
Validate the model when cannot find valid initial params.

### LINKED ISSUE #731: Add a better error message for how to debug "Cannot find valid initial parameters" issue
Recently, several users got this issue but it is not easy to diagnose the issue just by looking at the `model`. We should add a better error message, e.g. to mention about `numpyro.validation_enabled()` utility, which tells us at which latent site, we get invalid parameters/values.

fehiepsi:
Resolves #731. As explained there, it might be tricky to find bugs when the inference cannot find valid initial parameters. With this PR, when that happens, we can recognize at which site, things go wrong.

@rim30 could you run your code with this branch to see where causes the problem?

TODO

- [x] incorporate init_strategy because some distributions such Improper does not have `sample` method.

### DIFF
--- a/numpyro/infer/util.py
+++ b/numpyro/infer/util.py
@@ -427,9 +427,22 @@ def initialize_model(rng_key, model,
 
     if not_jax_tracer(is_valid):
         if device_get(~jnp.all(is_valid)):
-            with numpyro.validation_enabled(), \
+            with numpyro.validation_enabled(), trace() as tr, \
                     seed(rng_seed=rng_key if jnp.ndim(rng_key) == 1 else rng_key[0]):
+                # validate parameters
                 model(*model_args, **model_kwargs)
+                # validate values
+                for site in tr.values():
+                    if site['type'] == 'sample':
+                        with warnings.catch_warnings(record=True) as ws:
+                            site['fn']._validate_sample(site['value'])
+                        if len(ws) > 0:
+                            for w in ws:
+                                # at site information to the warning message
+                                w.message.args = ("Site {}: {}".format(site["name"], w.message.args[0]),) \
+                                    + w.message.args[1:]
+                                warnings.showwarning(w.message, w.category, w.filename, w.lineno,
+                                                     file=w.file, line=w.line)
             raise RuntimeError("Cannot find valid initial parameters. Please check your model again.")
     return ModelInfo(ParamInfo(init_params, pe, grad), potential_fn, postprocess_fn, model_trace)
 
--- a/numpyro/infer/util.py
+++ b/numpyro/infer/util.py
@@ -427,6 +427,9 @@ def initialize_model(rng_key, model,
 
     if not_jax_tracer(is_valid):
         if device_get(~jnp.all(is_valid)):
+            with numpyro.validation_enabled(), \
+                    seed(rng_seed=rng_key if jnp.ndim(rng_key) == 1 else rng_key[0]):
+                model(*model_args, **model_kwargs)
             raise RuntimeError("Cannot find valid initial parameters. Please check your model again.")
     return ModelInfo(ParamInfo(init_params, pe, grad), potential_fn, postprocess_fn, model_trace)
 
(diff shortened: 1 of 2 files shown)


rim30:
Hi @fehiepsi ,

Sorry for the massive delay, I was a bit busy until today. So I installed numpyro from your branch (https://github.com/fehiepsi/numpyro.git@validate) and got this error:

numpyro_error



The time series that i am using is: [1146.,  488.,  753.,  583.,  553.,  832.,  807.,  875.,
              945.,  795.,  862., 1322.,  890.,  911.,  990.,  791.,
              910.,  957.,  838.,  956.,  920.,  945., 1192., 1921.,
              987.,  907.,  762.,  859.,  843.,  804.,  785.,  942.,
              822.,  727.] which you can clearly see it is not negative.

fehiepsi:
@rim30 How about replacing `exp_val = jnp.clip(exp_val, a_min=0)` by
```python
exp_val = jnp.clip(exp_val, a_min=1e-30, a_max=1e38)
```
?

numpyro/infer/util.py
```diff
@@ -427,6 +427,21 @@ def initialize_model(rng_key, model,
 
     if not_jax_tracer(is_valid):
         if device_get(~jnp.all(is_valid)):
+            with numpyro.validation_enabled(), trace() as tr:
```
neerajprad, line 430: A more informative warning / error message is definitely needed. I am thinking that can we simply run `initialize_model` with `validation_enabled` (I wouldn't expect that to add any material overhead)? Is the resulting warning message not informative enough?

fehiepsi: Yes, it is simpler (I also don't worry about the overhead) but there are two issues with that:
+ validation is only useful for the first try (under jax loop, we can't prompt the warning/error for the later tries)
+ displaying the warning message for the first try might not be useful for users when we can find a valid one in a later try

What do you think?

neerajprad: Thanks for explaining, @fehiepsi, both of your points make a lot of sense.

numpyro/infer/util.py
```diff
@@ -427,6 +427,21 @@ def initialize_model(rng_key, model,
...
+                # validate values
+                for site in tr.values():
+                    if site['type'] == 'sample':
+                        with warnings.catch_warnings(record=True) as ws:
+                            site['fn']._validate_sample(site['value'])
+                        if len(ws) > 0:
+                            for w in ws:
+                                # at site information to the warning message
+                                w.message.args = ("Site {}: {}".format(site["name"], w.message.args[0]),) \
+                                    + w.message.args[1:]
```
neerajprad, line 442: What does a sample warning message look like?

neerajprad:
LGTM. Do you think this handles most of the forum questions you have been getting, or were they due to other numerical issues outside of distributions?

fehiepsi:
This helps detect some issues in the forum, one for data not belong to the support and one for wrong parameter.

numpyro/infer/util.py
fehiepsi: Thanks for reviewing, @neerajprad! I'm a bit busy today but will run this again to display the warning message tomorrow. Here I just want to add site information to the warning message, because not many users know how to use `warnings` to turn a warning to an error to debug.

neerajprad (approved):

numpyro/infer/util.py
fehiepsi: @neerajprad the following model
```python
import numpyro
import numpy as np
def model():
    x = numpyro.sample("x", numpyro.distributions.Normal())
    numpyro.sample("obs", numpyro.distributions.Normal(x), obs=float('nan'))
mcmc = numpyro.infer.MCMC(numpyro.infer.NUTS(model), 10, 10)
mcmc.run(np.array([0, 0], dtype='uint32'))
```
gives the warning
```
UserWarning: Site obs: Out-of-support values provided to log prob method. The value argument should be within the support.
```

Issue #724, opened 2020-09-07
HMCECS in numpyro


OlaRonning:
Hi,

@LysSanzMoreta and I are working on a project where we use [HMCECS](https://arxiv.org/abs/1708.00955), and so made a model-specific implemented in Numpyro. 
This implementation assumes the model to have the form:
```
def model(obs, feat, *args, **kwargs):
	pass
```
and then implement HMC-in-Gibbs as a loop, following the algorithm:
1. update subsample;
2. update parameters (by HMC) evaluated on the subsample.

We want to add HMCECS to Numpyro; however, we need feedback on a good design for handling the subsampling. The current version has a strong assumption on the parameters of the model (i.e. first and second params are obs and feats), which we would like to avoid; Pyro has plates that support subsamples, with user-defined subsample schemes. Are there plans for something similar in Numpyro, and if so what would it take to implement?

fehiepsi:
Hi @OlaRonning, I think the current `plate` statement supports subsampling (we have a [small test](https://github.com/pyro-ppl/numpyro/blob/master/test/test_handlers.py#L204) for it, also see [this thread](https://github.com/pyro-ppl/numpyro/issues/546) for a discussion). Could you be a little more explicit on "user-defined subsample schemes"? I think that we can implement some specific handlers for your purpose.

> The current version has a strong assumption on the parameters of the model (i.e. first and second params are obs and feats), which we would like to avoid

Yeah, I agree. I think we can avoid it, by either using a handler or modifying your HMCECS implementation a bit. I am looking forward to seeing your PR. :)

LysSanzMoreta:
I hope I can help, by "user-defined subsample schemes" we would like to be able to imitate NUTS on the automatic set up of the parameter "subsample size" (same as they do with step size for example). Therefore the subsample size will change /will be corrected according to the likelihood.

fehiepsi:
Unfortunately, I think that dynamic subsample size can only work with some masking mechanism. Kind of taking a subsample of max_subsample_size and masking out some log probabilities if the current subsample size is less than that max value. Could you provide some pseudo code so we can see if any additional mechanism is needed? I still believe that if something works with `obs, feats` then it will also work without them, as long as users provide enough information in the model or in the inference method. Btw, what does `feats` stand for?

OlaRonning:
Hi @fehiepsi,
Thanks for linking the subsampling discussion!

[...]

Issue #556, opened 2020-03-09
Support numpyro.subsample


fehiepsi:
As [requested](https://github.com/pyro-ppl/numpyro/issues/546#issuecomment-592002328) by @justindomke, it would be nice to have native support for subsampling, without requiring users to write subsample code.

This also comes from the internal discussion we have today.

fritzo:
Note there is a new [pyro.subsample()](http://docs.pyro.ai/en/stable/primitives.html#pyro.primitives.subsample) primitive that might be a good place to start.

PR #734, opened 2020-09-11
Support data subsampling

fehiepsi:
Addresses #556. This will also be helpful for #724.

@fritzo The implementation here is a bit simpler than Pyro. But I am not sure if I miss any point. Could you help me review this PR?


numpyro/primitives.py
```diff
@@ -224,38 +239,44 @@ def __init__(self, name, size, subsample_size=None, dim=None):
...
             raise ValueError('dim arg must be negative.')
-        self.dim = dim
-        self._validate_and_set_dim()
+        self.dim, self._indices = self._subsample(self.name, self.size, self.subsample_size, dim)
         super(plate, self).__init__()
 
-    def _validate_and_set_dim(self):
+    # XXX: different from Pyro, this method returns dim and indices
+    @staticmethod
+    def _subsample(name, size, subsample_size, dim):
```
fehiepsi, line 247: I made this a staticmethod to be able to reuse it in `contrib.funsor`. The api is also more similar to Pyro.

numpyro/contrib/funsor/enum_messenger.py
```diff
@@ -442,10 +442,11 @@ def __init__(self, name, size, subsample_size=None, dim=None):
...
             raise ValueError('dim arg must be negative.')
         self.dim = dim
+        _, indices = OrigPlateMessenger._subsample(self.name, self.size, self.subsample_size, dim)
         self._indices = funsor.Tensor(
-            funsor.ops.new_arange(funsor.tensor.get_default_prototype(), self.size),
-            OrderedDict([(self.name, funsor.bint(self.size))]),
-            self.size
+            indices,
+            OrderedDict([(self.name, funsor.bint(self.subsample_size))]),
+            self.subsample_size
```
fehiepsi, line 449: @eb8680 I made those changes to support subsampling. Could you take a look to see if this is a reasonable change?

numpyro/handlers.py
```diff
@@ -596,7 +599,7 @@ def __init__(self, fn=None, data=None, substitute_fn=None):
         super(substitute, self).__init__(fn)
 
     def process_message(self, msg):
-        if (msg['type'] not in ('sample', 'param')) or msg.get('_control_flow_done', False):
+        if (msg['type'] not in ('sample', 'param', 'plate')) or msg.get('_control_flow_done', False):
```
fehiepsi, line 602: I added `plate` here so we can substitute if needed. This is useful for #724 I guess. The behavior should be similar to the way we use `plate(..., subsample=)` in Pyro.

numpyro/handlers.py
```diff
@@ -547,7 +547,10 @@ def __init__(self, fn=None, rng_seed=None):
 
     def process_message(self, msg):
         if (msg['type'] == 'sample' and not msg['is_observed'] and
-                msg['kwargs']['rng_key'] is None) or msg['type'] == 'control_flow':
+                msg['kwargs']['rng_key'] is None) or msg['type'] in ['plate', 'control_flow']:
+            # no need to create a new key when value is available
+            if msg['value'] is not None:
+                return
```
fehiepsi, line 553: This saves some unnecessary splitting random keys.

numpyro/contrib/funsor/enum_messenger.py
eb8680: Looks right to me. The corresponding code in `pyro.contrib.funsor` is here: https://github.com/pyro-ppl/pyro/blob/dev/pyro/contrib/funsor/handlers/plate_messenger.py#L63

eb8680:
Looks mostly correct to me, but you might want to add tests that exercise replay and nested/reused plates, like the model in [this Pyro test](https://github.com/pyro-ppl/pyro/blob/dev/tests/infer/test_valid_models.py#L681).

> The implementation here is a bit simpler than Pyro. But I am not sure if I miss any point.

Part of the extra complexity in Pyro comes from supporting sequential plates, which I guess isn't a goal here, and part of it comes from treating subsample sites as `sample` sites, which was just a poor early design choice in Pyro. Also, are you planning to add a [`numpyro.subsample` primitive](http://docs.pyro.ai/en/stable/primitives.html#pyro.primitives.subsample)?

fehiepsi:
> add tests that exercise replay and nested/reused plates

Thanks @eb8680 ! I didn't even think about this. Thanks for pointing it out. :)

> planning to add a numpyro.subsample primitive?

I didn't have that plan. Let me see if it is easy to address.

eb8680:
> Let me see if it is easy to address.

The implementation in Pyro is here in `SubsampleMessenger._postprocess_message`: https://github.com/pyro-ppl/pyro/blob/dev/pyro/poutine/subsample_messenger.py#L134

Here's a simplified version that does not include parameter subsampling:
```py
    def postprocess_message(self, msg):  # new method for plate
        if msg["type"] == "subsample" and self.dim is not None:
            event_dim = msg["kwargs"].get("event_dim")
            if event_dim is not None:
                assert event_dim >= 0
                dim = self.dim - event_dim
                shape = msg["value"].shape
                if len(shape) >= -dim and shape[dim] != 1:
                    if is_validation_enabled() and shape[dim] != self.size:
                        statement = "pyro.subsample(..., event_dim={})".format(event_dim)
                        raise ValueError(
                            "Inside pyro.plate({}, {}, dim={}) invalid shape of {}: {}"
                            .format(self.name, self.size, self.dim, statement, shape))
                    if self.subsample_size < self.size:
                        value = msg["value"]
                        new_value = value.index_select(dim, self._indices)
                        msg["value"] = new_value
```

fehiepsi:
Thanks for the snippet, Eli! My intention is just making a quick PR for this but it is worth to extend this to support subsample and param primitives. Also Pyro allows subsample_size is None in model and not None in guide. I remember @fritzo said that he will use NumPyro if we have this feature implemented. That's a good motivation. :))

numpyro/contrib/funsor/enum_messenger.py
```diff
@@ -438,14 +439,14 @@ class plate(GlobalNamedMessenger):
     def __init__(self, name, size, subsample_size=None, dim=None):
         self.name = name
         self.size = size
-        self.subsample_size = size if subsample_size is None else subsample_size
         if dim is not None and dim >= 0:
             raise ValueError('dim arg must be negative.')
-        self.dim = dim
+        self.dim, indices = OrigPlateMessenger._subsample(self.name, self.size, subsample_size, dim)
```
fehiepsi, line 444: I changed the behavior here: set `self.dim` to be the output of `_subsample`, instead of default value in the constructor.

fehiepsi:
@eb8680 I have added `subsample` primitive and support subsample for it and `param`. Also made a small change to allow `subsample_size=None` in `model` but not None in `guide`.

fritzo (commented):
Note sure if this first subsample PR is the right place for thorough tests of SVI, but I found Pyro's [test_subsample_gradient()](https://github.com/pyro-ppl/pyro/blob/90bbb3e4314a7b7d2cba6f36a440aee521681dda/tests/infer/test_gradient.py#L40) to be invaluable in detecting subsampling issues with very little compute cost. In particular it helped us correctly scale the different parts of the `ScoreParts` tuple.

Does this PR aim to support subsampling in SVI? If so do you already have good tests for `poutine.scale()`?

numpyro/primitives.py
```diff
@@ -221,41 +241,50 @@ class plate(Messenger):
...
-                break
-            dim -= 1
-        if self.dim is None:
-            self.dim = dim
+        if dim is None:
+            new_dim = -1
+            while True:
+                if new_dim not in occupied_dims:
+                    break
+                new_dim -= 1
```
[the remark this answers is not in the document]
fehiepsi: Thanks!

fehiepsi:
@fritzo Thanks for reviewing! I just add `subsample_gradient` test. We have some tests for `handlers.scale` in `test_mask` and `test_scale`, but we haven't supported `ScoreParts` stuff in NumPyro yet.

> Does this PR aim to support subsampling in SVI?

My main motivation is to support https://github.com/pyro-ppl/numpyro/issues/724, where subsample is required (see [this comment](https://github.com/pyro-ppl/numpyro/issues/724#issuecomment-688491882)). But SVI should work I believe.

setup.py
```diff
@@ -51,8 +51,8 @@
             'ipython',
             'isort',
             'flax',
-            'dm-haiku @ https://github.com/deepmind/dm-haiku/archive/v0.0.2.zip',
-            'tfp-nightly',  # TODO: change this to stable release or a specific nightly release
+            'dm-haiku',
+            'tfp-nightly==0.12.0.dev20200911',  # TODO: change this to stable release or a specific nightly release
```
fehiepsi, line 55: I merged #739 to here because without it, all PRs will fail.

numpyro/handlers.py
fritzo: So do I understand correctly that in numpyro, in contrast to pyro, `plate` does not consume randomness? Note in Pyro it is `plate` that creates a `_Subsample` site that consumes randomness.

test/test_handlers.py
```diff
@@ -235,6 +258,102 @@ def test_plate(model):
...
+    num_particles = 50000
+    optimizer = optim.Adam(0.1)
+    elbo = ELBO(num_particles=num_particles)
+    svi = SVI(model, guide, optimizer, loss=elbo)
+    svi_state = svi.init(random.PRNGKey(0), None)
+    params = svi.optim.get_params(svi_state.optim_state)
+    normalizer = 2 if subsample else 1
+    if subsample_size == 1:
+        subsample = jnp.array([0])
+        _, grads1 = value_and_grad(lambda x: svi.loss.loss(
```
fritzo, line 340: could we also check that the values agree, since it's not much extra work?

fritzo (commented):
Looks great, thanks for implementing this!

numpyro/handlers.py
fehiepsi: @fritzo IIUC your question, then in NumPyro, `_subsample_fn` plays a similar role. If `value != None`, we will use the random key here to generate subsample indices.

fritzo: thanks for explaining!

numpyro/primitives.py
```diff
@@ -202,13 +211,24 @@ def module(name, nn, input_shape=None):
     return functools.partial(nn_apply, nn_params)
 
 
+def _subsample_fn(size, subsample_size, rng_key=None):
+    assert rng_key is not None, "Missing random key to generate subsample indices."
+    return random.permutation(rng_key, size)[:subsample_size]
```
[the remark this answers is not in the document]
fehiepsi: I think you did some [bechmarks](https://github.com/pyro-ppl/pyro/pull/2383) for this. In JAX, both are [equivalent](https://github.com/google/jax/blob/master/jax/random.py#L592) so using `random.choice` is also good to me.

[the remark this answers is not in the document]
fritzo: Ha ha thanks for finding the evidence. No need for change then.

test/test_handlers.py
fehiepsi: Hi @fritzo, I just added a test for value agreement. But I am not sure if that is what you mentioned.

fritzo (approved):

fritzo:
LGTM, @eb8680 did you have any other comments?

eb8680 (dismissed):
Looks good, thanks for adding the extra tests!

fehiepsi:
Thanks for reviewing, @eb8680 and @fritzo! I just observed something wrong in `Predictive` for models with multiple `subsample` statements. I'll try to fix it and add the failing test in this PR.

test/test_handlers.py
```diff
@@ -235,6 +258,110 @@ def test_plate(model):
...
+    data = jnp.arange(100.)
+    subsample_size = 7
+
+    with handlers.trace() as guide_trace, handlers.seed(rng_seed=0):
+        with numpyro.plate("a", len(data), subsample_size=subsample_size):
+            pass
+
+    with handlers.seed(rng_seed=1), handlers.replay(guide_trace=guide_trace):
+        with numpyro.plate("a", len(data)):
+            subsample_data = numpyro.subsample(data, event_dim=0)
```
fehiepsi, line 303: This test exercises `subsample` works under `replay` handler.

numpyro/handlers.py
```diff
@@ -192,7 +192,7 @@ def __init__(self, fn=None, guide_trace=None):
         super(replay, self).__init__(fn)
 
     def process_message(self, msg):
-        if msg['name'] in self.guide_trace and msg['type'] in ('sample', 'plate'):
+        if msg['type'] in ('sample', 'plate') and msg['name'] in self.guide_trace:
```
fehiepsi, line 195: I switch the order of two sides of `and` because some primitives such as `subsample` do not have `name` field.

fritzo (approved):

PR #738, opened 2020-09-12
Update ordinal regression tutorial to include proper prior

fehiepsi:
cc @vanAmsterdam


vanAmsterdam (dismissed):
good addition; maybe add a reason why / when a proper prior is desirable?

fehiepsi:
Improper prior can be convenient when normalization constant is not important (e.g. in HMC). But there are other kinds of inferences which require proper distributions (e.g. SVI). I'll mention about that. (actually, I think it is better to reason the other way around: we should use proper distributions, improper distributions should only be used with caution in mind)

neerajprad:
@fehiepsi - is this PR awaiting review still, or should we close this if it isn't valid (if so, could you add a comment as to why)?

fehiepsi:
Hi @neerajprad, I just wanted to fix the comment for why proper distribution is required. A better reason would be: it is tricky to sample from improper distribution, so in SVI, we prefer a proper one...

fehiepsi:
@neerajprad I just add another argument to use proper prior in MCMC and skip mentioning about SVI (though the same reason can be applied here - we can't use Improper distribution for the guide because we can't sample from it to calculate ELBO).

neerajprad (approved):

neerajprad:
LGTM. We can mention about SVI too, your comments were valid. From what I have seen, improper distributions are used often to place uninformative priors in MCMC, though I'm not sure whether it makes much difference in results for this example.

PR #739, opened 2020-09-13
Pin tfp nightly version

fehiepsi:
because the latest version is causing tests to fail (e.g. #738 does not pass Actions). There is a new flattening logic in the constructor of tfp distributions that prevents us from creating a thin wrapper. I think it will be fixed in the future.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -51,7 +51,7 @@ setup(
             'ipython',
             'isort',
             'flax',
-            'dm-haiku @ https://github.com/deepmind/dm-haiku/archive/v0.0.2.zip',
+            'dm-haiku',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -51,8 +51,8 @@
             'ipython',
             'isort',
             'flax',
-            'dm-haiku @ https://github.com/deepmind/dm-haiku/archive/v0.0.2.zip',
-            'tfp-nightly',  # TODO: change this to stable release or a specific nightly release
+            'dm-haiku',
```
fehiepsi, line 54: We can install `dm-haiku` from pypi.

fehiepsi:
This is already merged in master.

PR #740, opened 2020-09-13
Avoid `nan` acceptance probability in SA

fehiepsi:
Currently, when `log_weights_[-1] = inf`,
```
log_weights_[-1] - logsumexp(log_weights_)
```
will lead to `nan` acceptance probability. This PR clips `log_weights` by `finfo.max` value to prevent that issue.

cc  @awakhloo

### DIFF
--- a/numpyro/infer/sa.py
+++ b/numpyro/infer/sa.py
@@ -173,6 +173,9 @@ def _sa(potential_fn=None, potential_fn_gen=None):
         else:
             log_weights_ = dist.Normal(locs_, scales_).log_prob(zs_).sum(-1) + pes_
         log_weights_ = jnp.where(jnp.isnan(log_weights_), -jnp.inf, log_weights_)
+        log_weights_ = jnp.where(jnp.isposinf(log_weights_),
+                                 jnp.finfo(log_weights_.dtype).max,
+                                 log_weights_)
         # get rejecting index
         j = random.categorical(rng_key_reject, log_weights_)
         zs = _numpy_delete(zs_, j)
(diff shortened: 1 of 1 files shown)


fehiepsi:
@martinjankowiak Is it better to clip the nonsense value `inf` by `finfo.max` or `-inf`? The former means that we put the main weight at that `inf` sample, while the latter means that we put `0` weight to that value. In the last commit, I used `finfo.max` but using `-inf` makes more sense to me.

fehiepsi:
Anyway, after thinking more about this, I think it is safer to use `-inf` log weights for invalid values: nan, +inf.

numpyro/infer/sa.py
```diff
@@ -172,7 +172,8 @@ def sample_kernel(sa_state, model_args=(), model_kwargs=None):
             log_weights_ = dist.MultivariateNormal(locs_, scale_tril=scales_).log_prob(zs_) + pes_
         else:
             log_weights_ = dist.Normal(locs_, scales_).log_prob(zs_).sum(-1) + pes_
-        log_weights_ = jnp.where(jnp.isnan(log_weights_), -jnp.inf, log_weights_)
+        # mask invalid values (nan, +inf) by -inf
+        log_weights_ = jnp.where(jnp.isfinite(log_weights_), log_weights_, -jnp.inf)
```
neerajprad, line 176: LGTM, but when would this ever be `+inf`?

fehiepsi: It comes from a diffusion drift model in our forum (where the model has lax.scan). I guess something like 1/0 is involved or precision loss during the loop... (the model is pretty complicated and involves sequential matrix multiplication....).

neerajprad (approved):

PR #743, opened 2020-09-14
Support Flax random module

fehiepsi:
Towards nn -> bnn in numpyro. :)

Pair-coded with @TuanNguyen27 

Tested with both SVI and MCMC api.

### Tasks
- [x] add docs and examples
- [x] add batching and prediction to test


numpyro/contrib/module.py
```diff
@@ -71,6 +82,150 @@ def haiku_module(name, nn, input_shape=None):
...
+        >>> for i in tqdm.trange(n_iterations):
+        ...     svi_state, loss = update_fn(svi_state, x_train, y_train, batch_size)
+        >>>
+        >>> params = svi.get_params(svi_state)
+        >>> n_test_data = 100
+        >>> x_test, y_test = generate_data(n_test_data)
+        >>> predictive = Predictive(model, guide=guide, params=params, num_samples=1000)
+        >>> y_pred = predictive(random.PRNGKey(1), x_test[:100])["obs"].copy()
+        >>> assert loss < 3000
+        >>> assert np.sqrt(np.mean(np.square(y_test - y_pred))) < 1
```
fehiepsi, line 196: @TuanNguyen27 Finally, this example works well. :)

docs/requirements.txt
```diff
@@ -3,3 +3,5 @@ jaxlib>=0.1.45
 git+git://github.com/pyro-ppl/funsor@master#egg=funsor
 tqdm
 tfp_nightly  # TODO: change this to stable release
+flax
```
martinjankowiak, line 6: should these be optional?

numpyro/infer/hmc.py
```diff
@@ -412,11 +411,6 @@ def _init_state(self, rng_key, model_args, model_kwargs, init_params):
                 init_strategy=self._init_strategy,
                 model_args=model_args,
                 model_kwargs=model_kwargs)
```
martinjankowiak, line 413: why is this being removed?

numpyro/contrib/module.py
```diff
@@ -71,6 +82,150 @@ def haiku_module(name, nn, input_shape=None):
...
+def random_flax_module(name, nn_module, prior, *, input_shape=None):
+    """
+    A primitive to place a prior over the parameters of the Flax module `nn_module`.
+
+    :param str name: name of NumPyro module
+    :param flax.nn.Module: the module to be registered with NumPyro
+    :param prior: a NumPyro distribution or a Python dict with parameter names as keys and
+        respective distributions as values. For example::
+
+            net = random_flax_module("net",
```
neerajprad, line 130: How are we aligning the names in the `"prior"` dict and the NN parameter names? Are they easily available for a flax model, and what happens to the parameter names when multiple modules are combined?

numpyro/contrib/module.py
```diff
@@ -71,6 +82,150 @@ def haiku_module(name, nn, input_shape=None):
...
+                                     flax.nn.Dense.partial(features=1),
+                                     prior={"bias": dist.Cauchy(), "kernel": dist.Normal()},
```
neerajprad, line 132: For an MLP, if we just want to make the last layer bayesian, would this API allow for it and how would we do that (would a delta distribution work here or can we just skip some parameters entirely)?

docs/requirements.txt
fehiepsi: We need this to generate docs. Those are still optional in the main setup.

numpyro/infer/hmc.py
fehiepsi: This is a deprecation message. Previously, we treat `param` sites are improper latent variables but that support has been removed from 0.2 release IIRC. Now, for parameters of BNN, we still hold their shape information in `param` sites. Displaying this message is unnecessary for that usage case.

numpyro/contrib/module.py
fehiepsi: Parameters of `flax` and `haiku` are stored in nested dicts. For example, given the below model:
```
{'Dense_0': {'kernel': DeviceArray([[-0.5536316,  1.1597457, -0.6334757]], dtype=float32),
  'bias': DeviceArray([0., 0., 0.], dtype=float32)},
 'Dense_1': {'kernel': DeviceArray([[-0.09242683, -0.05430904, -0.3762646 ],
               [-0.1693897 ,  0.14864841, -0.57235295],
               [ 0.19652358,  0.6845585 , -0.71639156]], dtype=float32),
  'bias': DeviceArray([0., 0., 0.], dtype=float32)},
 'Dense_2': {'kernel': DeviceArray([[ 0.38516834],
               [-0.25442997],
               [-0.8081337 ]], dtype=float32),
  'bias': DeviceArray([0.], dtype=float32)},
 'Dense_3': {'kernel': DeviceArray([[-0.46138018],
               [ 0.3994321 ],
               [-0.8642299 ]], dtype=float32),
  'bias': DeviceArray([0.], dtype=float32)}}
```
An example for prior is: `{"Dense_1.kernel": dist.Normal(0, 1), "Dense_3.bias": dist.Normal(0, 1)}`. I used **flat** name convention as in PyTorch for this, as in [test_update_params](https://github.com/pyro-ppl/numpyro/pull/743/files#diff-b710b48cf627f2b6524be99a2d4194dbR60). Alternatively, we can use nested dict for `prior` here but I found it less flexible. What do you think?

numpyro/contrib/module.py
fehiepsi: Yes, `update_params` only lifts those in `prior` to `sample` statements (see its test, where the parameters `a.b.e` is kept as-is). At `net$params` site:
+ the nested structure is kept as-is
+ parameters not in `prior` are kept as-is
+ parameters in `prior` becomes `ParamShape` (which only holds `shape` information). When the optimizer flattens those values, they become empty. So we don't optimize anything here.

The output of this random module will be an NN with the following parameters (stored in `new_params` in the code):
+ the nested structure is kept as-is
+ parameters not in `prior` are kept as-is
+ parameters in `prior` will take the values drawn from their corresponding prior.

So the make the last layer bayesian, we can set `prior={"Dense_3.kernel": dist.Normal(0, 1), "Dense_3.bias": dist.Normal(0, 1)}`. Because I used flatten-with-dot name convention in `prior`, I think we can make `_update_params` more flexibile by using regex, i.e. `prior={"Dense_3.*": dist.Normal(0, 1)}`. I can do it in the follow-up PR (I believe it just took two or three lines of code to replace the check `flatten_name in prior` by a pattern matching for the regex expressions in `prior`'s keys). What do you think?

(using regex can be helpful when we want to set different priors for `kernel` and `bias`, across all layers)

numpyro/contrib/module.py
fehiepsi: If we decide to go with the "flatten name" convention, then I will need to add a note for it.

neerajprad: I think a flattened name convention (over nested dictionary) is fine. It would be more user-friendly if users could specify the prior at the level of the module itself so that they didn't need to understand the naming convention for chained Flax modules, but that will require a `NumpyroModule` that inherits from a Flax module and can be chained together similarly. I'll think about it. Lets proceed with this design for the time being.

numpyro/contrib/module.py
```diff
@@ -1,22 +1,32 @@
...
+def flax_module(name, nn_module, *, input_shape=None):
     """
     Declare a :mod:`~flax` style neural network inside a
     model so that its parameters are registered for optimization via
     :func:`~numpyro.primitives.param` statements.
 
     :param str name: name of the module to be registered.
-    :param flax.nn.Module nn: a `flax` Module which has .init and .apply methods
+    :param flax.nn.Module nn_module: a `flax` Module which has .init and .apply methods
     :param tuple input_shape: shape of the input taken by the
```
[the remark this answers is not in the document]
fehiepsi: I doubt that this is not supported in stax, flax, and haiku. Stax requires input_shape while flax and haiku requires a prototype input to be able to create init params. In those frameworks, some modules such as Dense/Linear only requires output info specified in the construction. I think if they design those nns such that they have both input and output infos in the construction, then we can infer param_shape from those infos. Not sure what is the reason behind that design.

numpyro/contrib/module.py
fehiepsi: You are right. I also find it tricky to specify correct path for parameters. This might be not a problem of flax/haiku users but for new users who want to construct a bnn, it is inconvenient. Looking like https://github.com/google/objax is the solution for us. Its interface is pretty similar to PyTorch.

neerajprad (approved):

numpyro/contrib/module.py
```diff
@@ -71,6 +81,166 @@ def haiku_module(name, nn, input_shape=None):
...
+        ...     with numpyro.plate("batch", x.shape[0], subsample_size=batch_size):
+        ...         batch_x = numpyro.subsample(x, event_dim=0)
+        ...         batch_y = numpyro.subsample(y, event_dim=0) if y is not None else None
+        ...         mean, rho = net(batch_x)
+        ...         sigma = nn.softplus(rho)
+        ...         numpyro.sample("obs", dist.Normal(mean, sigma), obs=batch_y)
+        >>>
+        >>> n_train_data = 5000
+        >>> x_train, y_train = generate_data(n_train_data)
+        >>> guide = autoguide.AutoDiagonalNormal(model, init_strategy=init_to_feasible)
```
neerajprad, line 195: You probably want to change this to `init_loc_fn`?

neerajprad (dismissed):

numpyro/contrib/module.py
fehiepsi: Thanks, @neerajprad! I just made the fix. I also fixed the doc of `plate` to not excute usage code, which makes `make doctest` fail in my system (not sure why it passed Github Action).

neerajprad (approved):

PR #747, opened 2020-09-20
Add AutoNormal guide

### LINKED ISSUE #746: Port autoguide AutoNormal from Pyro
I am trying to fit a bnn in #743 with ELBO loss but couldn't optimize it with stochastic KL. It might work if I use AutoNormal with TraceMeanField_ELBO, so I would like to add this feature to try that option.

fehiepsi:
Resolves #746.

With this guide, we can leverage MeanFieldELBO to make SVI converge faster.

There are several differences from Pyro:
- `cond_indep_stack` is a class in Pyro but it is a namedtuple in NumPyro. This makes it not possible to set the attribute `full_size` to `CondIndepStackFrame`. I made a workaround by storing an additional attribute `_prototype_frame_full_sizes` along with `_prototype_frames`.
- `locs` and `scales` are PyroModule in Pyro, but they do not have similar alternatives in NumPyro. So I stored `_init_locs` values and used `numpyro.param` for them instead.
- Because there are no `locs` and `scales`, I used the following names for those parameters: `f"{site_name}_{self.prefix}_loc"` and `f"{site_name}_{self.prefix}_scale"`. There are other alternatives in naming, but I have no preference among them. Suggestions are very welcome!
- When `support` of latent sites are real/real_vector, I skip the transform logic and use a single `Normal` guide (instead of Normal + Delta). With this, I can leverage MeanFieldELBO to improve SVI results.

#### TODO

- [x] Add test

### DIFF
--- a/numpyro/infer/autoguide.py
+++ b/numpyro/infer/autoguide.py
@@ -24,7 +24,7 @@ from numpyro.distributions.transforms import (
     UnpackTransform,
     biject_to
 )
-from numpyro.distributions.util import cholesky_of_inverse, sum_rightmost
+from numpyro.distributions.util import cholesky_of_inverse, periodic_repeat, sum_rightmost
 from numpyro.infer.elbo import ELBO
 from numpyro.infer.util import init_to_uniform, initialize_model
 from numpyro.nn.auto_reg_nn import AutoregressiveNN
@@ -52,6 +52,12 @@ class AutoGuide(ABC):
 
     :param callable model: a pyro model
     :param str prefix: a prefix that will be prefixed to all param internal sites
+    :param callable init_strategy: A per-site initialization function.
+        See :ref:`init_strategy` section for available functions.
+    :param callable create_plates: An optional function inputing the same
+        ``*args,**kwargs`` as ``model()`` and returning a :class:`numpyro.plate`
+        or iterable of plates. Plates not returned will be created
+        automatically as usual. This is useful for data subsampling.
     """
 
     def __init__(self, model, *, prefix="auto", init_strategy=init_to_uniform, create_plates=None):
@@ -61,6 +67,7 @@ class AutoGuide(ABC):
         self.create_plates = create_plates
         self.prototype_trace = None
         self._prototype_frames = {}
+        self._prototype_frame_full_sizes = {}
 
     def _create_plates(self, *args, **kwargs):
         if self.create_plates is None:
@@ -74,8 +81,9 @@ class AutoGuide(ABC):
             self.plates = {p.name: p for p in plates}
         for name, frame in sorted(self._prototype_frames.items()):
             if name not in self.plates:
-                size = self._prototype_plate_sizes[name]
-                self.plates[name] = numpyro.plate(name, size, dim=frame.dim, subsample_size=frame.size)
+                full_size = self._prototype_frame_full_sizes[name]
+                self.plates[name] = numpyro.plate(name, full_size, dim=frame.dim,
+                                                  subsample_size=frame.size)
         return self.plates
 
     @abstractmethod
@@ -110,7 +118,7 @@ class AutoGuide(ABC):
                 dynamic_args=False,
                 model_args=args,
                 model_kwargs=kwargs)
-        self._init_values = init_params[0]
+        self._init_locs = init_params[0]
 
         self._prototype_frames = {}
         self._prototype_plate_sizes = {}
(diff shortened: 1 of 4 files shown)


numpyro/infer/autoguide.py
```diff
@@ -50,13 +52,39 @@ class AutoGuide(ABC):
...
+        else:
+            plates = self.create_plates(*args, **kwargs)
+            if isinstance(plates, numpyro.plate):
+                plates = [plates]
+            assert all(isinstance(p, numpyro.plate) for p in plates), \
+                "create_plates() returned a non-plate"
+            self.plates = {p.name: p for p in plates}
+        for name, frame in sorted(self._prototype_frames.items()):
+            if name not in self.plates:
+                full_size = self._prototype_frame_full_sizes[name]
```
fehiepsi, line 84: In Pyro, we set `full_size` to be `frame.size` here. Is it a mistake? Without this change, I can't run SVI for the following model in #743
```python
        >>> def model(x, y=None, batch_size=None):
        ...     module = Net.partial(n_units=32)
        ...     net = random_flax_module("nn", module, dist.Normal(0, 0.1), input_shape=())
        ...     with numpyro.plate("batch", x.shape[0], subsample_size=batch_size):
        ...         batch_x = numpyro.subsample(x, event_dim=0)
        ...         batch_y = numpyro.subsample(y, event_dim=0) if y is not None else None
        ...         mean, rho = net(batch_x)
        ...         sigma = nn.softplus(rho)
        ...         numpyro.sample("obs", dist.Normal(mean, sigma), obs=batch_y)
```
unless I added `create_plates` to the guide.

fritzo: It looks like in Pyro we set `full_size = getattr(frame, "full_size", frame.size)` [here](https://github.com/pyro-ppl/pyro/blob/feeee7bcae6aefee6cca089f0d452e191aaf7535/pyro/infer/autoguide/guides.py#L347), and that the `full_size` attribute is set by `SubsampleMessenger` [here](https://github.com/pyro-ppl/pyro/blob/feeee7bcae6aefee6cca089f0d452e191aaf7535/pyro/poutine/subsample_messenger.py#L126). I think you'll also need the `frame.full_size` for param subsampling.

numpyro/infer/autoguide.py
```diff
@@ -50,13 +52,39 @@ class AutoGuide(ABC):
...
-    def __init__(self, model, prefix='auto'):
-        assert isinstance(prefix, str)
+    def __init__(self, model, *, prefix="auto", init_strategy=init_to_uniform, create_plates=None):
         self.model = model
         self.prefix = prefix
+        self.init_strategy = init_strategy
+        self.create_plates = create_plates
         self.prototype_trace = None
+        self._prototype_frames = {}
+        self._prototype_frame_full_sizes = {}
```
fritzo, line 70: As mentioned below I believe this is redundant given `self._prototype_frames`

numpyro/infer/autoguide.py
```diff
@@ -81,19 +109,155 @@ def sample_posterior(self, rng_key, params, *args, **kwargs):
...
+        result = {}
+        for name, site in self.prototype_trace.items():
+            if site["type"] != "sample" or isinstance(site["fn"], dist.PRNGIdentity) or site["is_observed"]:
+                continue
+
+            event_dim = self._event_dims[name]
+            init_loc = self._init_locs[name]
+            with ExitStack() as stack:
+                for frame in site["cond_indep_stack"]:
+                    stack.enter_context(plates[frame.name])
```
fritzo, line 203: (Note @eb8680 this is exactly the plate normal form I mentioned in https://github.com/pyro-ppl/pyro/pull/2635)

numpyro/infer/autoguide.py
fehiepsi: In Pyro, each frame is a class (subclass of a namedtuple), so we can set the attribute `full_size` for it. But in NumPyro, it is a namedtuple, which does not allow setting attributes. So I made this workaround.

numpyro/infer/autoguide.py
fehiepsi: Yeah, that Pyro code works for subsampled params. In the above model, we have subsampled data, so I think we need to use
```python
pyro.plate(name, full_size, dim=frame.dim, subsample_size=frame.size)
```
here instead of the current one in Pyro
```python
pyro.plate(name, frame.size, dim=frame.dim)
```
so in case `create_plates` is None, the `plate` in guide will be the same as the `plate` in model. (without this, I think `guide` will always generate `subsample=torch.arange(subsample_size)` and that constant index tensor will be used for subsampling in `model`)

numpyro/infer/autoguide.py
fehiepsi: @fritzo You can find descriptions for the differences between NumPyro and Pyro in the first post. :)

numpyro/infer/autoguide.py
fritzo: Thanks for explaining, I think I'm starting to understand. I believe historically Pyro only allowed guide subsampling via user-defined `create_plates()` functions, and ignored model subsampling (or mis-handled it as you suggest). Do I understand correctly that in your fixed version, the model's `subsample_size` will now determine the autoguide's strategy for automatic subsampling (rather than manual subsampling via `create_plates()`)?

I'll try to implement your fix in Pyro to help me understand. PR forthcoming 😄

fritzo: Here's your fix in Pyro 😄 https://github.com/pyro-ppl/pyro/pull/2638

fritzo (approved):
Looks great! I look forward to adding autoguide tests to proapi

numpyro/infer/autoguide.py
fehiepsi: Yes, I believe you got it right. :D

fehiepsi:
Thanks for reviewing!

PR #748, opened 2020-09-20
Add mean field ELBO

fehiepsi:
ported from Pyro's `Trace_MeanField_ELBO`.

This adds minimal support for MeanFieldELBO to be used in bayesian nn. The implementation follows closely Pyro's one.

### TODO
- [x] add test


numpyro/infer/elbo.py
```diff
@@ -73,6 +80,104 @@ def single_particle_elbo(rng_key):
...
+    assert set(model_sites) == set(guide_sites)
+    if model_sites != guide_sites:
+        warnings.warn("Failed to verify mean field restriction on the guide. "
+                      "To eliminate this warning, ensure model and guide sites "
+                      "occur in the same order.\n" +
+                      "Model sites:\n  " + "\n  ".join(model_sites) +
+                      "Guide sites:\n  " + "\n  ".join(guide_sites))
+
+
+class MeanFieldELBO(ELBO):
```
fehiepsi, line 110: @fritzo @neerajprad I need a bit of discussion for this. In NumPyro, `Trace_ELBO` -> `ELBO`, then how about `TraceMeanField_ELBO`? Do we have a plan to rename Pyro objectives for Pyro 2.0?

fritzo: In the one hand, I generally feel that users can tolerate a very limited amount of breakage before wanting to switch to another library, and I'd prefer to spend our limited breakage budget on interesting structural changes (like `numpyro.scan` and `PyroModule`) rather than on boring renamings. OTOH if people really think it's worth renaming, we could do so in Pyro 2 (but still provide backwards-compatible aliases).

fehiepsi: If we don't have plan to rename in Pyro, then it is better to use TraceMeanField_ELBO here in my opinion. WDYT @neerajprad ?

neerajprad: I think the reason why it was `Elbo` in numpyro is because the initial interface was based off of minipyro (rather than Pyro), which had dropped the `Trace` prefix. If we add the `Trace` prefix, we should do so uniformly. It will be more confusing to see `Elbo` and `TraceMeanFieldELBO`. In any case, the generic API should still use the names that we are using in Pyro, and resolve other interface differences which already exist.

p.s. - I thought we started with `Elbo` as the class name and not `ELBO`. The former would be the preferred way of naming classes that are acronyms as per pep-8.

neerajprad:
You'll need to merge changes from master.

numpyro/infer/elbo.py
fehiepsi: Thanks for the input, @neerajprad! It seems that I renamed `Elbo` to `ELBO` at some point in the past...

> If we add the Trace prefix, we should do so uniformly.

Which one do you prefer? My preference order is (top means "like most"):
+ Trace_ELBO and TraceMeanField_ELBO as in Pyro, deprecate ELBO. This is good because we have matched much of the code in early development to Pyro.
+ ELBO and TraceMeanField_ELBO as in this PR. This is also good to me and I don't see the confusion here for users.
+ ELBO and MeanFieldELBO as in the early stage of this PR
+ ELBO and MeanField_ELBO ...
+ Other options?

neerajprad: I think (1) sounds most reasonable, simply because the names match exactly with Pyro which dwarfs any other minor considerations. If we do not want to add deprecation warnings in this release, we can go with (2) and change `ELBO` to `Trace_ELBO` in a later release. However, since you have already made other changes to match Pyro API, we might as well bundle these changes with this release (maybe this should be a `0.4` release instead of a patch release?).

fehiepsi: Good idea! Thanks, I'll go with that. I feel having 0.4 release is more aligned with JAX 0.2 release. :D we also have new stuffs rather than just patches.

numpyro/infer/__init__.py
```diff
@@ -24,6 +24,7 @@
     'log_likelihood',
     'ELBO',
```
neerajprad, line 25: Do you want to do the renaming in this PR?

fehiepsi: @neerajprad I'll do it in a separate PR. I just fixed an issue at `contrib/module.py` where params given to `optim.init` method has a different order w.r.t. params got from `optim.get_params` method. It is due to in JAX, `unflatten_tree(flatten_tree(x))` is the same dict as `x` but having keys sorted. If keys are not consistent, latent sites will be inconsistent after we lifted them to `sample` sites, which violate mean field assumption: latent sites in guide should be as the same order as latent sites in model.

fehiepsi:
@fritzo @neerajprad Could you help me review this PR? After this, I'll deprecate ELBO in favor of Trace_ELBO and release NumPyro 0.4.0. Thanks!

numpyro/distributions/kl.py
```diff
@@ -0,0 +1,180 @@
+# Copyright Contributors to the Pyro project.
+# SPDX-License-Identifier: Apache-2.0
+
+# The implementation follows the design in PyTorch: torch.distributions.kl.py
```
fritzo, line 4: @fehiepsi it might be simpler to use Matt Rocklin's [multipledispatch](https://github.com/mrocklin/multipledispatch) library, which we use in Funsor. The original motivation for adding this pairwise KL stuff in PyTorch is that PyTorch didn't want to add multipledispatch as a dependency. But since NumPyro already depends on Funsor, you won't be adding a dependency.
```py
from multipledispatch import dispatch

@dispatch(Distribution, ExpandedDistribution)
def kl_divergence(p, q):
    kl = kl_divergence(p, q.base_dist)
    shape = lax.broadcast_shapes(p.batch_shape, q.batch_shape)
    return jnp.broadcast_to(kl, shape)

...
```

fehiepsi: Yeah, it is a good idea. But `funsor` is currently an optional dependency of numpyro, so I would like to defer it to a future PR. I'll make an issue to track this.

fritzo (approved):
Generally looks good. I haven't checked the math in detail, so I assume you've faithfully translated the Pyro version.

numpyro/infer/elbo.py
```diff
@@ -73,6 +80,104 @@ def single_particle_elbo(rng_key):
...
+                      "Guide sites:\n  " + "\n  ".join(guide_sites))
+
+
+class TraceMeanField_ELBO(ELBO):
+    """
+    A trace implementation of ELBO-based SVI. This is currently the only
+    ELBO estimator in NumPyro that uses analytic KL divergences when those
+    are available.
+
+    .. warning:: This estimator may give incorrect results if the mean-field
```
martinjankowiak, line 116: is it worth including the full requirements?

```

    The mean field condition is a sufficient but not necessary condition for
    this estimator to be correct. The precise condition is that for every
    latent variable `z` in the guide, its parents in the model must not include
    any latent variables that are descendants of `z` in the guide. Here
    'parents in the model' and 'descendants in the guide' is with respect
    to the corresponding (statistical) dependency structure. For example, this
    condition is always satisfied if the model and guide have identical
    dependency structures.
```

numpyro/infer/elbo.py
```diff
@@ -73,6 +80,104 @@ def single_particle_elbo(rng_key):
...
+                        log_prob = jnp.sum(model_site["fn"].log_prob(model_site["value"]))
+                        log_prob = scale_and_mask(log_prob, scale=model_site["scale"])
+                        elbo_particle = elbo_particle + log_prob
+                    else:
+                        guide_site = guide_trace[name]
+                        try:
+                            kl_qp = kl_divergence(guide_site["fn"], model_site["fn"])
+                            kl_qp = scale_and_mask(kl_qp, scale=guide_site["scale"])
+                            if isinstance(kl_qp, (int, float)):
+                                kl_qp_sum = kl_qp * np.prod(guide_site["fn"].batch_shape)
```
martinjankowiak, line 157: what's going on here? doesn't `kl_qp` include`guide_site["fn"].batch_shape` already?

numpyro/infer/elbo.py
```diff
@@ -73,6 +80,104 @@ def single_particle_elbo(rng_key):
...
+            subs_model = substitute(replay(seeded_model, guide_trace), data=param_map)
+            model_trace = trace(subs_model).get_trace(*args, **kwargs)
+            _check_mean_field_requirement(model_trace, guide_trace)
+
+            elbo_particle = 0
+            for name, model_site in model_trace.items():
+                if model_site["type"] == "sample" and not isinstance(model_site["fn"], dist.PRNGIdentity):
+                    if model_site["is_observed"]:
+                        log_prob = jnp.sum(model_site["fn"].log_prob(model_site["value"]))
+                        log_prob = scale_and_mask(log_prob, scale=model_site["scale"])
```
fehiepsi, line 149: nit: use `_get_log_prob_sum` here for simplicity

numpyro/infer/elbo.py
fehiepsi: Thanks, I'll add it.

numpyro/infer/elbo.py
fehiepsi: I thought so too. This follows [this line](https://github.com/pyro-ppl/pyro/blob/dev/pyro/infer/trace_mean_field_elbo.py#L108) in Pyro. I guess we need it for [KL[masked, masked]](https://github.com/pyro-ppl/numpyro/pull/748/files#diff-a54d4ede5b490415bf09bbe4c4dd8103R169) where it returns `0.` when `mask=False`. There might be other places which I don't have knowledge about. Nevertheless, I believe we can remove this logic by simply using `kl_qp_sum = jnp.sum(kl_qp)` here without if/else. WDYT?

fehiepsi: Anyway, I just remove this logic. It is better. Thanks!

martinjankowiak: hmm. @fritzo i feel like we may have pair-coded that line... do you perhaps remember why...?

fritzo: There must have been a shortcut where we returned a broadcastable scalar value rather than a batched tensor, and the `np.prod` is required to get the power right when summing over many identical tensor elements. However I cannot think of a case other than `{-math.inf, 0, math.inf}`, and in those cases the power has no effect. You might simply convert this to an assertion like
```py
assert (not isinstance(kl_qp, (int, float)) or (kl_qp in (-math.inf, 0, math.inf))
```
(Aside: I wish Python had an "implies" operator so we didn't have to write "a implies b" as "(not a) or b")

neerajprad (approved):
LGTM, pending any further review comments from @fritzo, @martinjankowiak.

fritzo (approved):

fehiepsi:
Thanks everyone for reviewing!
________________________________

Issue #749, opened 2020-09-24
NumPyro is breaking with JAX 0.2


fehiepsi:
In JAX 0.2 release, many of our tests are failing... I think it is related to the new [omnistaging](https://jax.readthedocs.io/en/latest/CHANGELOG.html#jax-0-2-0-september-23-2020) mechanism.

This issue discusses several strategies that we would follow for our next release:
- Release 0.3.1 and pin jax version to `0.1.77`
- Release 0.3.1, pin jax version to `0.2.0`, disable omnistaging by default
- Try to fix all issues if possible, then release 0.3.1 with 0.2.0.

The first option is the safest because we don't understand much how stable the omnistaging is. Hopefully, it is stable enough, the issue lies at some other JAX commits, and all failing tests can be easily fixed.

neerajprad:
How easy is it to fix the tests/other issues with `0.2.0`. I agree that it is best to pin to `0.1` for the time being.

fehiepsi:
I am trying to fix those issues. Hopefully, they will be easy to fix. :)

PR #750, opened 2020-09-24
Pin jax version to 0.1.77

fehiepsi:
A temporary solution for #749 to unblock our PRs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,7 @@ jobs:
         python -m pip install --upgrade pip
         pip install https://github.com/pyro-ppl/funsor/archive/master.zip
         pip install jaxlib
-        pip install jax
+        pip install jax<0.2.0
         pip install .[doc,test]
         pip freeze
     - name: Lint with flake8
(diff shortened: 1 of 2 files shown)


neerajprad (approved):

.github/workflows/ci.yml
```diff
@@ -27,7 +27,7 @@ jobs:
         python -m pip install --upgrade pip
         pip install https://github.com/pyro-ppl/funsor/archive/master.zip
         pip install jaxlib
-        pip install jax
+        pip install "jax<0.2"
```
neerajprad, line 30: It will be best to pin this to `0.1.77` because I am pretty sure our current dev branch will be incompatible with older versions of JAX.

neerajprad (changes requested):
Lets pin to `0.1.77` explicitly or use a range like `pip install "jax>=0.1.xxx,<0.2"` that we know works.

.github/workflows/ci.yml
fehiepsi: Yeah, I just made the change.

neerajprad (approved):

PR #753, opened 2020-09-25
Some changes to match Pyro api

### LINKED ISSUE #751: Rename `init_strategy` in SVI to `init_loc_fn`
We used `init_strategy` to be consistent with MCMC at our early development. In recent releases, we are gradually making two APIs consistent but we missed this point. It would be nice to make two API consistent.

fehiepsi:
Fixes #751.

This PR renames `init_strategy` to `init_loc_fn` in autoguide and `value` to `v` in Delta distribution. We missed those points in the last release.

In addition, I addressed a small point in autoguide, where we might return both `sample` and `param` sites when computing `median`, `quantiles`,... This PR makes sure that we do not include `param` sites in the result.


numpyro/infer/autoguide.py
```diff
@@ -52,18 +52,18 @@ class AutoGuide(ABC):
...
+    :param callable init_loc_fn: A per-site initialization function.
         See :ref:`init_strategy` section for available functions.
     :param callable create_plates: An optional function inputing the same
         ``*args,**kwargs`` as ``model()`` and returning a :class:`numpyro.plate`
         or iterable of plates. Plates not returned will be created
         automatically as usual. This is useful for data subsampling.
     """
 
-    def __init__(self, model, *, prefix="auto", init_strategy=init_to_uniform, create_plates=None):
+    def __init__(self, model, *, prefix="auto", init_loc_fn=init_to_uniform, create_plates=None):
```
neerajprad, line 63: Since this is a patch release and we have already moved this out of contrib, this will break existing code. We should add a deprecation warning here.

fehiepsi: Oops, I forgot... :( thanks for reminding, I will add depreciation warning here and Delta too .

neerajprad: Actually, could you add `FutureWarning` instead which won't be swallowed by python?

neerajprad (approved):

PR #754, opened 2020-09-25
Pump jax and tfp-nightly version

### LINKED ISSUE #749: NumPyro is breaking with JAX 0.2

fehiepsi:
Fixes #749 

My target is to release NumPyro 0.3.1 with `jax=0.2` and `tfp-nightly=0.12.0.dev20200923` or a released version for `tfp` if possible.

All the fixes are:
- `jnp` -> `np` in several places, which many happens in examples, notebooks, and some nn flows
- skip register pytree for tensorflow wrappers because tensorflow distributions are pytrees now (in a recent commit)
- `trace(jit(model)).get_trace()` stores ShapedArray, so in `test_handlers.py::test_plate`, we can't compare values of `trace(jit(model))` and `trace(model)` anymore. I think that is an expected behavior but not sure why the test passed previously. To fix this, I only compare shapes of those values.
- change precisions at some tests
- ReplicaExchangeMC is not fully compatible with omnistaging yet. I made a [PR](https://github.com/tensorflow/probability/pull/1087) to fix this issue upstream.

### Tasks

- [x] fix all failing errors


fehiepsi:
Hi @brianwa84, do you know if TFP 0.12 will be released soon?

examples/annotation.py
```diff
@@ -86,7 +88,7 @@ def multinomial(annotations):
     """
     This model corresponds to the plate diagram in Figure 1 of reference [1].
     """
-    num_classes = int(jnp.max(annotations)) + 1
+    num_classes = int(np.max(annotations)) + 1
```
fehiepsi, line 91: This is the main error: we can't cast `int` to a DeviceArray under jit anymore. Something likes those will cause errors now:
```
def f(x):
    return np.ones(10)[x]

jax.jit(f)(jnp.array([1, 2]))
```
or
```
def f(x):
    a = int(jnp.ones(()))
    return x + a

jax.jit(f)(jnp.array([1, 2]))
```

neerajprad:
> My target is to release NumPyro 0.3.1 with jax=0.2 and tfp-nightly=0.12.0.dev20200923 or a released version for tfp if possible.

Maybe we can ask when the next release of tfp is happening, but since this is an optional dependency, also feel free to proceed with the release using the nightly version.

examples/annotation.py
neerajprad: So under jit, function arguments cannot be later used as concrete arrays?

numpyro/nn/block_neural_arn.py
```diff
@@ -21,8 +26,8 @@ def BlockMaskedDense(num_blocks, in_factor, out_factor, bias=True, W_init=glorot
     input_dim, out_dim = num_blocks * in_factor, num_blocks * out_factor
     # construct mask_d, mask_o for formula (8) of Ref [1]
     # Diagonal block mask
-    mask_d = jnp.identity(num_blocks)[..., None]
-    mask_d = jnp.tile(mask_d, (1, in_factor, out_factor)).reshape(input_dim, out_dim)
+    mask_d = np.identity(num_blocks)[..., None]
+    mask_d = np.tile(mask_d, (1, in_factor, out_factor)).reshape(input_dim, out_dim)
```
neerajprad, line 30: I am not entirely clear why we need `np.tile` here. Could you elaborate?

examples/annotation.py
fehiepsi: I think they are still concrete arrays. The new change is we can't apply numpy/python operators on DeviceArray under jit any more (previously, it is doable as long as that DeviceArray is constant - i.e. not derived from traced inputs). In addition, under jit, we can't use DeviceArray indexing for non-DeviceArray array. We have to do something like
```
def f(x):
    return jnp.ones(10)[x]
```
or
```
def f(x):
    return device_put(np.ones(10))[x]
```

numpyro/nn/block_neural_arn.py
fehiepsi: If we use `jnp.tile`, this will become a DeviceArray and the error happens at `dense_logdet[mask_d.astype(bool)]`: `IndexError: Array boolean indices must be concrete.` In jax 0.2, it is required that `mask_d.astype(bool)` is a concrete boolean index so I fixed it by creating a concrete array instead of DeviceArray.

neerajprad: I see, thanks for explaining.

neerajprad (approved):

brianwa84:
> Hi @brianwa84, do you know if TFP 0.12 will be released soon?

We've historically followed the tensorflow release cadence, but I think we could be open to doing more frequent releases if we could better automate things, I've raised it to @csuter to discuss.

fehiepsi:
Thanks, @brianwa84! It totally makes sense to follow the tensorflow release. I also like the fact that we can install the nightly version directly using pip. It is pretty convenient. Let us see how it goes with recent integrations of NumPyro with TFP. Hopefully, we can set up more closely collaborations in the future.

PR #760, opened 2020-09-29
Deprecated ELBO in favor of Trace_ELBO

fehiepsi:
This change is per discussion at #748.

### DIFF
--- a/numpyro/infer/elbo.py
+++ b/numpyro/infer/elbo.py
@@ -39,6 +39,8 @@ class Trace_ELBO:
     :param num_particles: The number of particles/samples used to form the ELBO
         (gradient) estimators.
     """
+    def __init__(self, num_particles=1):
+        self.num_particles = num_particles
 
     def loss(self, rng_key, param_map, model, guide, *args, **kwargs):
         """
@@ -77,10 +79,10 @@ class Trace_ELBO:
 
 
 class ELBO(Trace_ELBO):
-    def __init_subclass__(cls, **kwargs):
+    def __init__(self, num_particles=1):
         warnings.warn("Using ELBO directly in SVI is deprecated. Please use Trace_ELBO class instead.",
                       FutureWarning)
-        super().__init_subclass__(**kwargs)
+        super().__init__(num_particles=num_particles)
 
 
 def _get_log_prob_sum(site):
--- a/numpyro/infer/elbo.py
+++ b/numpyro/infer/elbo.py
@@ -1,6 +1,9 @@
 # Copyright Contributors to the Pyro project.
 # SPDX-License-Identifier: Apache-2.0
 
+from abc import ABCMeta, abstractmethod
+import warnings
+
 from jax import random, vmap
 from jax.lax import stop_gradient
 import jax.numpy as jnp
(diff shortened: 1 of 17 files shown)


numpyro/infer/elbo.py
```diff
@@ -10,7 +13,59 @@
...
+            name.
+        :param model: Python callable with NumPyro primitives for the model.
+        :param guide: Python callable with NumPyro primitives for the guide.
+        :param args: arguments to the model / guide (these can possibly vary during
+            the course of fitting).
+        :param kwargs: keyword arguments to the model / guide (these can possibly vary
+            during the course of fitting).
+        :return: negative of the Evidence Lower Bound (ELBO) to be minimized.
+        """
+        warnings.warn("Using ELBO directly in SVI is deprecated. Please using Trace_ELBO class instead.",
```
neerajprad, line 43: nit: "Please use..."

numpyro/infer/elbo.py
```diff
@@ -10,7 +13,59 @@
...
+        # Return (-elbo) since by convention we do gradient descent on a loss and
+        # the ELBO is a lower bound that needs to be maximized.
+        if self.num_particles == 1:
+            return - single_particle_elbo(rng_key)
+        else:
+            rng_keys = random.split(rng_key, self.num_particles)
+            return - jnp.mean(vmap(single_particle_elbo)(rng_keys))
+
+
+class Trace_ELBO(ELBO):
```
neerajprad, line 68: Can't we simply do the following (until we deprecate `ELBO` in the next release)?

```python
class Trace_ELBO(ELBO):
    pass
```

neerajprad: Better still, I think we can rename the base class to `Trace_ELBO` and use the following. This will raise a FutureWarning for subclasses too and moreover we can just simply delete these lines in the next release.

```python
class ELBO(Trace_ELBO):
    def __init_subclass__(self):
        warnings.warn(...
```

fritzo: i'm confused, it looks like `ELBO` is now an abstract class

fehiepsi: Yes, `ELBO` now is an abstract class. Custom objectives should have the same signature for `loss` method. I think @neerajprad's idea is to raise the warning as early as possible. In my commit, I raised the warning in `loss` method, which is not a good idea. I think we can go with Neeraj solution and add to the warning message: "ELBO will become the base class for other objectives in a future release." and make other classes subclass `Trace_ELBO` instead (or we can make each class separated from each other for now).

What do both of you think? Do I get it correctly?

neerajprad: @fehiepsi - Thanks for explaining. We had a simple `ELBO` class earlier, but now that we have two subclasses `Trace_ELBO` and `TraceMeanField_ELBO`, I suppose we are following Pyro's design of having an abstract base class called `ELBO` instead. In that case, I have the same confusion as @fritzo - can we actually instantiate `ELBO` and call the `loss` method (even though its annotated as abstract)? If not, consider not making it an abstract class at least for now.

fehiepsi: Oh, I didn't know that we can't instantiate an abstract class with abstract methods. Thanks!

I just made the change. Is it reasonable now? It seems that `__init_subclass__` is unnecessary so I used `__init__` instead.

neerajprad (approved):

PR #764, opened 2020-09-30
Use `raise .. from ..` where appropriate

### LINKED ISSUE #761: Fix exception chaining all over the codebase
I recently went over [PyTorch](https://github.com/pytorch/pytorch/pull/43836) and [Detectron2](https://github.com/facebookresearch/detectron2/pull/1945), suggesting a fix in the way that Python 3's exception chaining is used.

As described in detail in [the article](https://blog.ram.rachum.com/post/621791438475296768/improving-python-exception-chaining-with), exception chaining ([PEP 3134](https://www.python.org/dev/peps/pep-3134/)) can be used to make exceptions more user-friendly, and in that case, the syntax `raise .. from ..` needs to be used.

When `raise .. from ..` is used, there will be a line saying `The above exception was the direct cause of the following exception` between tracebacks. However, when we implicitly chaining exceptions (meaning when `raise .. from ..` syntax is not used), the message will be `During handling of the above exception, another exception occurred` which can confuse users.

Specifically, the following should be used in order to chain exceptions:
```python
try:
    something_which_raises_OldError()
except OldError as e:
    raise NewError("A more user-friendly exception message.") from e 
```
instead of:
```python
try:
    something_which_raises_OldError()
except OldError:
    raise NewError("A more user-friendly exception message.")
```

Here is one example which needs to be fixed. The below case needs a fix because it means, "while handling `KeyError`, `NotImplementedError` **unexpectedly** occurs."
https://github.com/pyro-ppl/numpyro/blob/298f3c9907fab8eb1dc08b1acf636edc2b6ba4cc/numpyro/distributions/transforms.py#L556-L559

If this suggestion sounds good and reasonable, I'd be happy to submit a PR! Let me know your thoughts on this!

akihironitta:
## Motivation
Fixes and closes #761.

## Description of the changes
I changed lines where `raise`ing an exception in `except` to either:
1. `raise .. from e` to explicitly chain exceptions where the exception `e` has valuable information and seems useful for users.
2. `raise .. from None` to unchain exceptions where the previous exception has no valuable information and seems noisy and redundant.

### DIFF
--- a/numpyro/compat/infer.py
+++ b/numpyro/compat/infer.py
@@ -132,7 +132,7 @@ class SVI(svi.SVI):
         except TypeError as e:
             if 'not a valid JAX type' in str(e):
                 raise TypeError('NumPyro backend requires args, kwargs to be arrays or tuples, '
-                                'dicts of arrays.')
+                                'dicts of arrays.') from e
             else:
                 raise e
         params = jit(super(SVI, self).get_params)(self.svi_state)
--- a/numpyro/contrib/funsor/__init__.py
+++ b/numpyro/contrib/funsor/__init__.py
@@ -3,11 +3,11 @@
 
 try:
     import funsor
-except ImportError:
+except ImportError as e:
     raise ImportError("Looking like you want to do inference for models with "
                       "discrete latent variables. This is an experimental feature. "
                       "You need to install `funsor` to be able to use this feature. "
-                      "It can be installed with `pip install funsor`.")
+                      "It can be installed with `pip install funsor`.") from e
 
 from numpyro.contrib.funsor.enum_messenger import enum, infer_config, markov, plate, to_data, to_funsor, trace
 from numpyro.contrib.funsor.infer_util import config_enumerate, log_density, plate_to_enum_plate
--- a/numpyro/contrib/module.py
+++ b/numpyro/contrib/module.py
@@ -35,11 +35,11 @@ def flax_module(name, nn_module, *, input_shape=None):
     """
     try:
         import flax  # noqa: F401
-    except ImportError:
+    except ImportError as e:
         raise ImportError("Looking like you want to use flax to declare "
                           "nn modules. This is an experimental feature. "
                           "You need to install `flax` to be able to use this feature. "
-                          "It can be installed with `pip install flax`.")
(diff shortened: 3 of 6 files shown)


akihironitta:
## List of lines
I created [a simple script](https://gist.github.com/akihironitta/4223c1b32404b36c1b349d70c4c93b4d) to list all the lines containing `raise` in `except` clause.
- [ ] https://github.com/pyro-ppl/numpyro/blob/8ce9fd36c96d5ec3786aecc670a565065dd70852/numpyro/contrib/funsor/__init__.py#L7
- [ ] https://github.com/pyro-ppl/numpyro/blob/8ce9fd36c96d5ec3786aecc670a565065dd70852/numpyro/contrib/module.py#L39
- [ ] https://github.com/pyro-ppl/numpyro/blob/8ce9fd36c96d5ec3786aecc670a565065dd70852/numpyro/contrib/module.py#L75
- [ ] https://github.com/pyro-ppl/numpyro/blob/8ce9fd36c96d5ec3786aecc670a565065dd70852/numpyro/contrib/tfp/__init__.py#L7
- [ ] https://github.com/pyro-ppl/numpyro/blob/8ce9fd36c96d5ec3786aecc670a565065dd70852/numpyro/distributions/transforms.py#L559
- [ ] https://github.com/pyro-ppl/numpyro/blob/8ce9fd36c96d5ec3786aecc670a565065dd70852/numpyro/compat/infer.py#L134
- [ ] https://github.com/pyro-ppl/numpyro/blob/8ce9fd36c96d5ec3786aecc670a565065dd70852/numpyro/compat/infer.py#L137
- [ ] https://github.com/pyro-ppl/numpyro/blob/8ce9fd36c96d5ec3786aecc670a565065dd70852/numpyro/infer/reparam.py#L186
- [ ] https://github.com/pyro-ppl/numpyro/blob/8ce9fd36c96d5ec3786aecc670a565065dd70852/test/test_distributions.py#L438

akihironitta:
I actually didn't use `raise .. from None` because all the exceptions seem to have valuable information and it looks inappropriate to get rid of the information by using `raise .. from None`.

fehiepsi (approved):
LGTM, this is very helpful. Thanks, @akihironitta!

PR #773, opened 2020-10-03
Add numpyro.collapse

### LINKED ISSUE #756: Port poutine.collapse to NumPyro
Pyro introduced a `poutine.collapse()` effect in https://github.com/pyro-ppl/pyro/pull/2620, allowing exact conjugate inference over blocks of code, similar to @neerajprad's [pyro.contrib.conjugate](https://github.com/pyro-ppl/pyro/blob/dev/pyro/contrib/conjugate/infer.py).

@fehiepsi I think this would be an edifying intro to using Funsor for general pattern matching. Once `poutine.collapse()` is implemented in NumPyro, we can add [new conjugacy patterns](https://github.com/pyro-ppl/funsor/issues/367) to Funsor. Currently I believe only Beta-Binomial and Gaussian-Gaussian are implemented. I think some good starter Funsor tasks would be implementing patterns for say Beta-Bernoulli and then patterns for plates, as in Beta-plated(Binomial).

cc @eb8680

fehiepsi:
Resolves #756.

### TODO:
- [x] debug why the pattern is not recognized
- [x] add beta bernoulli xfail test

Blocked by https://github.com/pyro-ppl/funsor/pull/377

### DIFF
--- a/numpyro/contrib/tfp/distributions.py
+++ b/numpyro/contrib/tfp/distributions.py
@@ -90,7 +90,23 @@ def _transform_to_bijector_constraint(constraint):
     return BijectorTransform(constraint.bijector)
 
 
-class TFPDistributionMixin(NumPyroDistribution):
+_TFPDistributionMeta = type(tfd.Distribution)
+
+
+# XXX: we create this mixin class to avoid metaclass conflict between TFP and NumPyro Ditribution
+class _TFPMixinMeta(_TFPDistributionMeta, type(NumPyroDistribution)):
+    def __init__(cls, name, bases, dct):
+        # XXX: _TFPDistributionMeta.__init__ registers cls as a PyTree
+        # for some reasons, when defining metaclass of TFPDistributionMixin to be _TFPMixinMeta,
+        # TFPDistributionMixin will be registered as a PyTree 2 times, which is not allowed
+        # in JAX, so we skip registering TFPDistributionMixin as a PyTree.
+        if name == "TFPDistributionMixin":
+            super(_TFPDistributionMeta, cls).__init__(name, bases, dct)
+        else:
+            super(_TFPMixinMeta, cls).__init__(name, bases, dct)
+
+
+class TFPDistributionMixin(NumPyroDistribution, metaclass=_TFPMixinMeta):
     """
     A mixin layer to make TensorFlow Probability (TFP) distribution compatible
     with NumPyro internal.
@@ -118,11 +134,11 @@ class TFPDistributionMixin(NumPyroDistribution):
         return self.support is None
 
 
-class InverseGamma(tfd.InverseGamma):
+class InverseGamma(tfd.InverseGamma, TFPDistributionMixin):
     arg_constraints = {"concentration": constraints.positive, "scale": constraints.positive}
 
 
-class OneHotCategorical(tfd.OneHotCategorical):
+class OneHotCategorical(tfd.OneHotCategorical, TFPDistributionMixin):
     arg_constraints = {"logits": constraints.real_vector}
     has_enumerate_support = True
     support = constraints.simplex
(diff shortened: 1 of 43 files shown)


numpyro/contrib/tfp/distributions.py
```diff
@@ -90,7 +90,23 @@ def _transform_to_bijector_constraint(constraint):
...
+_TFPDistributionMeta = type(tfd.Distribution)
+
+
+# XXX: we create this mixin class to avoid metaclass conflict between TFP and NumPyro Ditribution
+class _TFPMixinMeta(_TFPDistributionMeta, type(NumPyroDistribution)):
+    def __init__(cls, name, bases, dct):
+        # XXX: _TFPDistributionMeta.__init__ registers cls as a PyTree
+        # for some reasons, when defining metaclass of TFPDistributionMixin to be _TFPMixinMeta,
+        # TFPDistributionMixin will be registered as a PyTree 2 times, which is not allowed
+        # in JAX, so we skip registering TFPDistributionMixin as a PyTree.
```
fritzo, line 102: Wow, that must have been tricky to debug!

numpyro/distributions/distribution.py
```diff
@@ -71,7 +71,23 @@ def validation_enabled(is_validate=True):
         enable_validation(distribution_validation_status)
 
 
-class Distribution(object):
+# TODO Remove import guard once funsor is a required dependency.
```
fritzo, line 74: Could you refactor to use the updated `COERCIONS` pattern as in https://github.com/pyro-ppl/pyro/pull/2641 ? I'm hoping we can refactor `contrib.epidemiology` to use the same `COERCIONS` mechanism, thus keeping Numpyro aligned with Pyro will make it easier to eventually port `contrib.epidemiology` to Numpyro.

numpyro/handlers.py
```diff
@@ -242,6 +243,66 @@ def process_message(self, msg):
...
 
 
+class collapse(trace):
+    """
+    EXPERIMENTAL Collapses all sites in the context by lazily sampling and
+    attempting to use conjugacy relations. If no conjugacy is known this will
+    fail. Code using the results of sample sites must be written to accept
+    Funsors rather than Tensors. This requires ``funsor`` to be installed.
+    """
+    def __init__(self, *args, **kwargs):
```
fritzo, line 253: Ditto: update `__init__`, `__enter__` and `__exit__` as in https://github.com/pyro-ppl/pyro/pull/2641

numpyro/handlers.py
```diff
@@ -242,6 +243,66 @@ def process_message(self, msg):
...
+
+        funsor.set_backend("jax")
+        super().__init__(*args, **kwargs)
+
+    def process_message(self, msg):
+        from funsor.terms import Funsor
+
+        if msg["type"] == "sample":
+            if msg["value"] is None:
+                msg["value"] = msg["name"]
```
fritzo, line 264: Is there no need for `msg["done"] = True` as in Pyro?

test/test_handlers.py
```diff
@@ -509,3 +509,84 @@ def model():
...
+def test_collapse_beta_binomial_plate():
+    data = np.array([0., 1., 5., 5.])
+
+    def model():
+        c = numpyro.sample("c", dist.Gamma(1, 1))
+        with handlers.collapse():
+            probs = numpyro.sample("probs", dist.Beta(c, 2))
+            with numpyro.plate("plate", len(data)):
+                numpyro.sample("obs", dist.Binomial(10, probs),
+                               obs=data)
```
fritzo, line 583: cc @neerajprad I believe this is basically the mobb model 😄

fritzo (changes requested):

numpyro/contrib/tfp/distributions.py
fehiepsi: Yeah, it is tricky. Especially to understand `__call__` `__new__` `__init__` logics of metaclass. >___<

numpyro/distributions/distribution.py
fehiepsi: Sure, I didn't notice that you put a new pattern in Pyro.

numpyro/handlers.py
fehiepsi: We don't have `msg["done"]` in NumPyro. What does it do?

fritzo: @eb8680 can you explain `msg["done"]` vs `msg["stop"]`?

fritzo (approved):

numpyro/handlers.py
eb8680: `msg["done"]` indicates that `msg["value"]` is final and should not be changed by subsequent handlers.  In this case it would equivalent to `msg["value"] is not None`, although it can occasionally be useful to distinguish between these situations.

eb8680 (approved):
LGTM

PR #775, opened 2020-10-06
Make omnistaging work with validate_args=True

### LINKED ISSUE #771: Omnistaging does not work when enabling distributions' validation
Hi again,

I have updated to 0.4 and my probit model (that runs in the previous version) gives the following error:

ConcretizationTypeError: Abstract tracer value encountered where concrete value is expected.

the model I fairly simple, (see below), I thought it might be cdf (jax.scipy.stats.norm.cdf) but could not make sense of it.

Have you experienced such changes by updating ?

MODEL:

def model(nwifeinc,
                  educ,
                  exper,
                  exper2,
                  age,
                  kidslt6,
                  kidsge6,
                  y=None):
    
    alpha = numpyro.sample("alpha", dist.Normal(0, 1))
    b1 = numpyro.sample("b1", dist.Normal(0, 1))
    b2 = numpyro.sample("b2", dist.Normal(0, 1))
    b3 = numpyro.sample("b3", dist.Normal(0, 1))
    b4 = numpyro.sample("b4", dist.Normal(0, 1))
    b5 = numpyro.sample("b5", dist.Normal(0, 1))
    b6 = numpyro.sample("b6", dist.Normal(0, 1))
    b7 = numpyro.sample("b7", dist.Normal(0, 1))
    xb =  alpha + b1*nwifeinc  + b2*educ +b3*exper + b4*exper2 + b5*age  + b6*kidslt6 + b7*kidsge6
    phi =  cdf(xb)  
    numpyro.sample("y", dist.BernoulliProbs(phi), obs=y)

fehiepsi:
Fixes #771. The issue is with omnistaging, under jit, operators such as `jnp.all(concrete_array)` will return a TracedArray, which cannot be converted to a Python bool object. I think "omnistaging" is a bit annoying but given that it is the default behavior in JAX 0.2, we have to live with it. :(

### DIFF
--- a/numpyro/distributions/discrete.py
+++ b/numpyro/distributions/discrete.py
@@ -183,12 +183,14 @@ class BinomialProbs(Distribution):
         return constraints.integer_interval(0, self.total_count)
 
     def enumerate_support(self, expand=True):
-        total_count = jnp.amax(self.total_count)
-        if not_jax_tracer(total_count):
+        if not_jax_tracer(self.total_count):
+            total_count = np.amax(self.total_count)
             # NB: the error can't be raised if inhomogeneous issue happens when tracing
-            if jnp.amin(self.total_count) != total_count:
+            if np.amin(self.total_count) != total_count:
                 raise NotImplementedError("Inhomogeneous total count not supported"
                                           " by `enumerate_support`.")
+        else:
+            total_count = jnp.amax(self.total_count)
         values = jnp.arange(total_count + 1).reshape((-1,) + (1,) * len(self.batch_shape))
         if expand:
             values = jnp.broadcast_to(values, values.shape[:1] + self.batch_shape)
@@ -200,6 +202,7 @@ class BinomialLogits(Distribution):
                        'total_count': constraints.nonnegative_integer}
     has_enumerate_support = True
     is_discrete = True
+    enumerate_support = BinomialProbs.enumerate_support
 
     def __init__(self, logits, total_count=1, validate_args=None):
         self.logits, self.total_count = promote_shapes(logits, total_count)
@@ -234,18 +237,6 @@ class BinomialLogits(Distribution):
     @property
     def support(self):
         return constraints.integer_interval(0, self.total_count)
-
-    def enumerate_support(self, expand=True):
-        total_count = jnp.amax(self.total_count)
-        if not_jax_tracer(total_count):
-            # NB: the error can't be raised if inhomogeneous issue happens when tracing
-            if jnp.amin(self.total_count) != total_count:
-                raise NotImplementedError("Inhomogeneous total count not supported"
-                                          " by `enumerate_support`.")
-        values = jnp.arange(total_count + 1).reshape((-1,) + (1,) * len(self.batch_shape))
-        if expand:
-            values = jnp.broadcast_to(values, values.shape[:1] + self.batch_shape)
-        return values
 
 
 def Binomial(total_count=1, probs=None, logits=None, validate_args=None):
--- a/numpyro/distributions/discrete.py
+++ b/numpyro/distributions/discrete.py
@@ -412,7 +412,10 @@ class OrderedLogistic(CategoricalProbs):
                        'cutpoints': constraints.ordered_vector}
 
     def __init__(self, predictor, cutpoints, validate_args=None):
-        predictor, = promote_shapes(predictor, shape=jnp.shape(predictor) + (1,))
+        if jnp.ndim(predictor) == 0:
+            predictor, = promote_shapes(predictor, shape=(1,))
+        else:
+            predictor = predictor[..., None]
         predictor, self.cutpoints = promote_shapes(predictor, cutpoints)
         self.predictor = predictor[..., 0]
         cumulative_probs = expit(cutpoints - predictor)
--- a/numpyro/distributions/discrete.py
+++ b/numpyro/distributions/discrete.py
@@ -412,7 +412,8 @@ class OrderedLogistic(CategoricalProbs):
                        'cutpoints': constraints.ordered_vector}
 
     def __init__(self, predictor, cutpoints, validate_args=None):
-        predictor, self.cutpoints = promote_shapes(jnp.expand_dims(predictor, -1), cutpoints)
+        predictor, = promote_shapes(predictor, shape=jnp.shape(predictor) + (1,))
+        predictor, self.cutpoints = promote_shapes(predictor, cutpoints)
         self.predictor = predictor[..., 0]
         cumulative_probs = expit(cutpoints - predictor)
         # add two boundary points 0 and 1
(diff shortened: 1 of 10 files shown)


test/test_distributions.py
```diff
@@ -748,6 +748,22 @@ def test_distribution_constraints(jax_dist, sp_dist, params, prepend_shape):
         d.log_prob(oob_samples)
 
 
+def test_omnistaging_invalid_param():
```
neerajprad, line 751: We have tests that check whether arg constraints are satisfied, but I suppose we do not have tests that check validation for out of support values in distributions? If so, it will be good to add tests for all relevant distributions and do a patch release. WDYT?

neerajprad: This test is fine, but my concern is if we have missed something like this for other distributions.

fehiepsi: Yeah, I think we already have some tests that check for both constraints and support. Let me also add `jit` check for them.

fehiepsi: Actually, you raised a very good point. It is just lucky that my test passed for `positive` constraint. For many other constraints, no warning is raised! :( The issue is if the constraint uses `jax.numpy` to check, the output will be a traced array (under jit), regardless the input is a DeviceArray or not.

fehiepsi:
@neerajprad Omnistaging is a bit annoying for validation code. Currently, I found no simply way to have some "check" under `jit` without adding some dispatching mechanism to some of our constraint implementations. For arg validation, there are a few places that we use `jnp.broadcast_to`, which will create Tracer output. To fix it, I tried to use `promote_shapes` for all of them. :(

fritzo:
@fehiepsi can you explain more about the difficulties and need for dispatch mechanism? I'm just curious because I have been thinking about what additional metadata might be useful to add to `Constraint` objects, e.g. maybe an `.event_dim` attribute or richer shape attributes like (`.batch_shape`, `.event_shape`).

fehiepsi:
Yes, I meant so. With the new behavior of JAX, outside `jit`, we have numpy.ndarray and DeviceArray while under `jit`, they become numpy.ndarray and ShapedArray (i.e. tracer).  For validation, currently, we use the code
```
mask = support.check(value)
if not_tracer(mask):
    if not jax.numpy.all(mask):
        raise ...
```
Assume `mask` is a numpy ndarray (or python True/False), `not_tracer(mask)` will be True. Then we need to check if all values in `mask` is True. If we use `jax.numpy.all`, `jax.numpy.all(mask)` will be a tracer under jit. So `if not jax.numpy.all(mask)` will raise a JAX error because we cannot get the actual value of `jax.numpy.all(mask)` here.

In this PR, I used `numpy.all(mask)` instead. In addition, to be able to raise the validation warning/error, it is required that `mask` is not a tracer (i.e. an abstract shaped array). If we use `jax.numpy` operators in `support.check` implementation, the output will always be a tracer so I added the above dispatching mechanism. With that, if `value` is a numpy ndarray, we will get a numpy ndarray `mask`. Of course, the validation code still does not work if `value` is a device array. However, I think it is still useful to detect issues such as: using `0` value for scale parameter or data contains `nan`,...

If the code is run outside of `jit` context, then the validation will still work because `numpy.all(...)` can apply for a DeviceArray.

Re what additional metadata might be useful: Currently, we have `event_dim` but I think having something like TFP [forward_event_shape](https://www.tensorflow.org/probability/api_docs/python/tfp/bijectors/Affine#forward_event_shape) method would be useful to:
+ determine changes in event_dim of `domain` and `codomain` (e.g. LKJ)
+ determine changes in shapes of `domain` and `codomain` (e.g. Simplex)
Currently, we can't give a correct `event_shape` of a TransformedDistribution due to the missing of such method.
In addition, I think having a method `forward_domain` will also be useful, in case we want to use `AffineTransform` with `positive` domain. Currently, users need to define a correct support in e.g. [Weibull](https://github.com/pytorch/pytorch/blob/master/torch/distributions/weibull.py#L10) because we don't have such logic for the composed of `PowerTransform` and `AffineTransform`.

numpyro/distributions/distribution.py
```diff
@@ -150,9 +152,9 @@ def __init__(self, batch_shape=(), event_shape=(), validate_args=None):
                     continue
                 if is_dependent(constraint):
                     continue  # skip constraints that cannot be checked
-                is_valid = jnp.all(constraint(getattr(self, param)))
+                is_valid = constraint(getattr(self, param))
                 if not_jax_tracer(is_valid):
-                    if not is_valid:
+                    if not np.all(is_valid):
```
neerajprad, line 157: For my understanding, if we were to always cast a python bool to a numpy or jax bool type (i.e. the default mask would be a `jnp.array(True)`), we would not have needed any change here. Is that correct?

numpyro/distributions/discrete.py
```diff
@@ -183,12 +183,14 @@ def support(self):
         return constraints.integer_interval(0, self.total_count)
 
     def enumerate_support(self, expand=True):
-        total_count = jnp.amax(self.total_count)
-        if not_jax_tracer(total_count):
+        if not_jax_tracer(self.total_count):
+            total_count = np.amax(self.total_count)
```
neerajprad, line 187: Same as my comment below - is this `np` needed because `self.total_count` could be a python scalar. In that case does it make sense to use `jnp.array` or `device_put` everywhere? That way, we can freely use `jnp` functions instead of a mix of `np` and `jnp` which becomes hard to reason about.

neerajprad:
>  For arg validation, there are a few places that we use jnp.broadcast_to, which will create Tracer output. To fix it, I tried to use promote_shapes for all of them. :(

If we can avoid this by using `device_put` (basically not having any python scalar types in distributions and masking etc.), then that would seem preferable and will help avoid future regressions. That said, my understanding of the actual issues could be quite off, so correct me if I am mistaken.

numpyro/distributions/distribution.py
fehiepsi: We we need this change because under `jit`. `jnp.all(True)`, `device_put(True)` will create a tracer. So validation code won't activate.
```python
from jax import device_put, jit
import jax.numpy as jnp

def f():
    assert not not_jax_tracer(jnp.all(True))
    assert not not_jax_tracer(device_put(True))

jit(f)()  # pass
```

You are right that if we cast `mask` in `_validate_sample` method to a device array, then the error will be fixed. In that case, no validation for parameters/sample are activated under `jit`. It is fine for single chain mcmc because we trace the model outside of `jit` context, so if there are some wrong specification, the users will get the warning/error. But for multi chain mcmc, no validation will be activated because we run everything under `pmap`.

neerajprad: Thanks for explaining about `pmap`, I missed that. In that case, is it correct to say that for distributions that have all parameters as jax devices (no python scalars), there won't be any validation performed under `pmap` because `not_jax_tracer(param)` will be true?

fehiepsi: Yes, that's true.

neerajprad (approved):

neerajprad:
Thanks for explaining your changes, @fehiepsi.

Issue #774, opened 2020-10-05
Bayesian Neural Network doesn't work for my dataset


Nikolin4:
Hi there!

I am writting to you because  bayesian neural model does not produce the desired output.
I have one hot encoding data, also X.shape = (N,20), also the data has 20 features, each for an amino acid.
The targets are  Y.shape = (N,2), also one feature stands for phi angles, and the other feature stands for psi angles.
N stands for the numbers of the rows.

My objectives are to predict the phi, psi, kappa, and mu where mu and kappa are the parameters for von Misses distribution,
phi and psi are the torsion angles for proteins.  (kappa.shape, mu.shape = (N, 2)## phi.shape,  psi.shape = (N,1))

Can anyone help develop this model using latent variables?
I cannot share the code from here, but let me know how we can work on it

This model is taken from http://pyro.ai/numpyro/examples/bnn.html .
The model in the following is written directly here without having tested. 
(I haven't shared the code, but this is what I use)
Please contribute to build this model, maybe by adding latent variables, or anything might be useful.
I can't get my head around this problem
Hope to hear from you

```python
def model(X, Y, D_H):
	D_X, D_Y = X.shape[1], 2
  w1 = numpyro.sample("w1", dist.Normal( 0,1 ), sample_shape (D_X, D_H))
  b1 = numpyro.sample("b1", dist.Normal(0,1), sample_shape(D_H,))
  z1 = softmax( jnp.matmul(X,w1) +b1)

  w2 = numpyro.sample("w2", dist.Normal( 0,1 ), sample_shape (D_H D_H))
  b2 = numpyro.sample("b2", dist.Normal(0,1), sample_shape(D_H,))
  z2 = softmax( jnp.matmul(z2,w2) +b2)


  w3 = numpyro.sample("w3", dist.Normal(0,1   ), sample_shape=(D_H, 2*D_Y)  )
  b3 = numpyro.sample("b3", dist.Normal(0,1), sample_shape = (2*D_Y, ) )
  z3 = jnp.matmul(z2, w3) + b3
  kappa = numpyro.deterministic("kappa", sigmoid(z3[ : ,  :2 ] ) ) #Activation on kappas needed
  mu = numpyro.deterministic("mu",  z3[ : , 2: ] ) #No activation needed
  numpyro.sample('phi', dist.VonMises(mus[: ,0], kappas[: , 0], obs= Y[: ,0]) 
[...]

PR #776, opened 2020-10-07
Fix von mises log prob and sample

fehiepsi:
Hopefully, this will be helpful for #774.

This fixes VonMises.log_prob and `von_mises_centered` sampler. They are tested against scipy. In the master branch, we missed taking `log` after applying Bessel function on concentration parameter. In addition, because `bessel_i0e` is exponentially scaled Bessel function, we need to plus `concentration` after taking the log. In addition, I made some enhancements:
+ using `jax.scipy.special.i1e` instead of `lax.bessel_i1e` because the latter can't be applied on int scalar/arrays.
+ scale `value - loc` to [0, 2pi] before taking `cos`. This is helpful when `value - loc` is very large, which causes discontinuity as in the picture below
image

Blocked by #777

cc @Nikolin4 @OlaRonning

### DIFF
--- a/numpyro/distributions/util.py
+++ b/numpyro/distributions/util.py
@@ -420,7 +420,7 @@ def _von_mises_centered(key, concentration, shape, dtype):
         w = jnp.where(done, w, (1. + s * z) / (s + z))  # Update where not done
 
         y = concentration * (s - w)
-        v = random.uniform(key=uni_vkey, shape=shape, dtype=concentration.dtype, minval=-1., maxval=1.)
+        v = random.uniform(key=uni_vkey, shape=shape, dtype=concentration.dtype)
 
         accept = (y * (2. - y) >= v) | (jnp.log(y / v) + 1. >= y)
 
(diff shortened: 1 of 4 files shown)


Nikolin4 (approved):

Nikolin4 (approved):

fehiepsi:
@OlaRonning Could you help me review this PR? I think those fixes would be helpful for your experiments with VonMises.

OlaRonning:
Hi @fehiepsi, I'll have a look this afternoon.

numpyro/distributions/util.py
```diff
@@ -420,7 +420,7 @@ def body_fn(*args):
         w = jnp.where(done, w, (1. + s * z) / (s + z))  # Update where not done
 
         y = concentration * (s - w)
-        v = random.uniform(key=uni_vkey, shape=shape, dtype=concentration.dtype, minval=-1., maxval=1.)
+        v = random.uniform(key=uni_vkey, shape=shape, dtype=concentration.dtype)
```
OlaRonning, line 423: Is the reasoning that we can remove the bounds that [-1,1] ~ Reals?

OlaRonning (approved):
See question below, otherwise LGTM, well-done @fehiepsi!

neerajprad (approved):

PR #777, opened 2020-10-08
Pass tests for jax 0.2.1

fehiepsi:
There is an [issue](https://github.com/google/jax/issues/4490) at `np.mean` in jax 0.2.1 that makes our CI failed. Because the issue just raises some UserWarning and it only affects a few tests, I haven't pinned jax to 0.2.0. Instead, I made a few changes in our tests to make CI pass.


fritzo (approved):

fehiepsi:
Thanks, Fritz!

PR #780, opened 2020-10-14
Use original plate implementation in enum

fehiepsi:
This PR defers `postprocess_msg` and `postprocess_msg` implementation in enum plate to the original implementation. This is helpful to avoid inconsistent logics of enum vs without enum. In particular, this helps to fix the user bug in [forum](https://forum.pyro.ai/t/data-shape-mismatch-with-nested-plates/2270).


eb8680:
Can you explain what's actually different here, functionally? How does this address the question on the forum - is it the fact that you're not using `self.dim` anymore?  Does the HMM example still work?

test/contrib/test_funsor.py
```diff
@@ -196,6 +197,14 @@ def testing():
         testing()
 
 
+def test_nested_plate():
+    with enum(first_available_dim=-3):
+        with enum_plate("a", 5):
+            with enum_plate("b", 2):
+                x = numpyro.sample("x", dist.Normal(0, 1), rng_key=random.PRNGKey(0))
+                assert x.shape == (2, 5)
```
eb8680, line 205: Was this assertion false before this PR?

fehiepsi:
Previously, we used the logic: `if msg['type'] not in ('sample',): return super().process_message(msg)`, so the below code will only run if `msg['type'] == 'sample'`. On the other hand, the logic in `numpyro.plate` runs if `if msg['type'] not in ('param', 'sample', 'plate'):`. This PR just makes the logic consistent.

test/contrib/test_funsor.py
fehiepsi: Yes, actually, there is an error that happens even before this assertation. Currently, plate "a" will have `dim=-1`, plate "b" will also have `dim=-1`, this makes it impossible to expand `Normal(0, 1)` to both (5,) and (2,) shapes.

fehiepsi:
> Does the HMM example still work?

Unfortunately, HMM example `model4` breaks at some point (before this PR). How do you know? I think it is unrelated so will try to fix it in this PR, but if you think it is related, then I will try to fix it first and let this PR block.

test/contrib/test_funsor.py
eb8680: I guess I'm puzzled at how the change in this PR fixes that - as far as I can tell, setting aside `subsample`/`param` statements, the only behavioral difference is that the logic in `plate.process_message` now runs for `"plate"` sites and `super().process_message(msg)` now does not run for `"plate"` sites, since the bulk of the code in `plate.process_message` was copied verbatim from `numpyro.primitives.plate`.

Seems like I really should update `numpyro.contrib.funsor` ASAP.

eb8680:
> Unfortunately, HMM example model4 breaks at some point (before this PR). How do you know?

Just a guess, since it's the most complicated enumeration example in NumPyro at the moment and makes use of nested plates.

fehiepsi:
@eb8680 I think I see the issues now. There are two places that make things wrong
+ In the PR #734, I added `postprocess_msg` to enum plate (because I also did so in the original plate). However, enum plate is inherited by `GlobalNamedMessenger`, which has its own `postprocess_msg` method implemented. I didn't notice that so after that PR, the `GlobalNamedMessenger.postprocess_msg` does not run anymore. That's why we got some incorrect named dimensions at observed variables.
+ When I said hmm model 3/4 is not working anymore, I used `hidden_dim=3`. I tried to roll back to the first PR that supports enum scan but the issues still happen. It turns out that in enum scan implementation, I used the field `dim_to_name` to determine the largest dimension. Here are dimensions at each sites for hidden_dim=3 and hidden_dim=4 (corresponding to the actual hidden_dim=1 and hidden_dim=2 because in that model `hidden_dim = int(sqrt(args.hidden_dim))`): (each line is name -> batch_shape -> dim_to_name)
```
_probs_w_log_det () OrderedDict()
probs_w () OrderedDict()
_probs_x_log_det () OrderedDict()
probs_x () OrderedDict()
_probs_y_log_det () OrderedDict()
probs_y () OrderedDict()
_init/w (1, 229, 1) OrderedDict([(-2, 'sequences')])
_init/x (1, 1, 229, 1) OrderedDict([(-2, 'sequences')])
_init/y (1, 1, 229, 51) OrderedDict([(-2, 'sequences'), (-1, 'tones')])
w (2, 1, 1, 1, 229, 1) OrderedDict([(-6, '_pyro_dim_6'), (-2, 'sequences')])
x (2, 1, 1, 1, 1, 229, 1) OrderedDict([(-7, '_pyro_dim_7'), (-2, 'sequences')])
y (2, 1, 1, 1, 1, 229, 51) OrderedDict([(-7, '_pyro_dim_7'), (-2, 'sequences'), (-1, 'tones')])
```

```
_probs_w_log_det () OrderedDict()
probs_w () OrderedDict()
_probs_x_log_det () OrderedDict()
probs_x () OrderedDict()
_probs_y_log_det () OrderedDict()
probs_y () OrderedDict()
_init/w (1, 229, 1) OrderedDict([(-2, 'sequences')])
_init/x (1, 1, 229, 1) OrderedDict([(-2, 'sequences')])
_init/y (1, 1, 229, 51) OrderedDict([(-2, 'sequences'), (-1, 'tones')])
w (2, 1, 1, 1, 229, 1) OrderedDict([(-6, '_time_w'), (-2, 'sequences')])
x (2, 1, 1, 1, 1, 229, 1) OrderedDict([(-7, '_time_w'), (-2, 'sequences')])
y (2, 1, 1, 1, 1, 229, 51) OrderedDict([(-7, '_time_w'), (-2, 'sequences'), (-1, 'tones')])
```
You can see that time dimensions are not available in the first case. Here are time dimensions for both cases in the enum scan implementation:
```
# name -> time dimension -> batch_shape
time dimension at : w -2 (2, 1, 1, 1, 229, 1)
time dimension at:  x -2 (2, 1, 1, 1, 1, 229, 1)
time dimension at:  y -2 (2, 1, 1, 1, 1, 229, 51)
```

```
time dimension at:  w -5 (2, 1, 1, 2, 229, 1)
time dimension at:  x -6 (2, 1, 1, 2, 1, 229, 1)
time dimension at:  y -6 (2, 2, 2, 1, 1, 229, 51)
```

I have addressed the first issue by running `GlobalNamedMessenger.postprocess_msg` if the msg is either `to_data` or `to_funsor` and tried to clean up the implementation to make things consistent. If you don't like that clean up, I can copy verbatim again the implementation of original plate `process_msg` and `postprocess_msg` to enum plate.

For the second issue, I derived `time_dim` by using `fn['site']` and `fn['value']` instead of using `fn['infer']['dim_to_name']`.

After those fixes, HMM enum works again and provides correct named dimension. I have added enum tests to `test_examples`. However, there is another issue when we summary the result if the model has Dirichlet latent variable with event_shape = (1,). I made a small fix in `MCMC` to account for that case and added corresponding test `test_trivial_dirichlet`.

numpyro/infer/mcmc.py
```diff
@@ -338,7 +338,10 @@ def _single_chain_mcmc(self, init, args, kwargs, collect_fields):
         states = dict(zip(collect_fields, states))
         # Apply constraints if number of samples is non-zero
         site_values = tree_flatten(states[self._sample_field])[0]
-        if len(site_values) > 0 and site_values[0].size > 0:
+        # XXX: lax.map still works if some arrays have 0 size
+        # so we only need to filter out the case site_value.shape[0] == 0
+        # (which happens when lower_idx==upper_idx)
+        if len(site_values) > 0 and jnp.shape(site_values[0])[0] > 0:
```
fehiepsi, line 344: @neerajprad Could you help me review this change? If we use `site_values[0].size` here, those values with shape `(num_samples, 0)` will not be post-processed, hence some errors will happen when we use the samples for other tasks (or even when we print the summary). For simplex constraint (the support of Dirichlet), if the constrained value has shape (1,), then its unconstrained value will have shape (0,). That will give use constant constrained samples. I believe this is not useful, but it is better to let users decide if they want to do so, rather than just throwing a misleading error e.g. `ValueError: cannot reshape array of size 0 into shape (1,0)` when I summarize a `hmm` model with this prior.

eb8680 (approved):
@fehiepsi thanks for the detailed explanation, the latest changes make sense to me and I can now understand how they fix the underlying issues.

fehiepsi:
Thanks for reviewing, @eb8680!

PR #783, opened 2020-10-19
Bump to 0.4.1 (again)

fehiepsi:
@neerajprad I missed this place. Without this, upload to PyPI raise some errors. :(

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -47,7 +47,7 @@ setup(
             'pyro-api>=0.1.1'
         ],
         'dev': [
-            'funsor @ git+https://github.com/pyro-ppl/funsor.git@6575ac2c3f7ac25a6a1e5f1107b1b7c072edd992',
+            'funsor',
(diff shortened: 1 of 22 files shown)


setup.py
```diff
@@ -47,7 +47,7 @@
             'pyro-api>=0.1.1'
         ],
         'dev': [
-            'funsor @ git+https://github.com/pyro-ppl/funsor.git@6575ac2c3f7ac25a6a1e5f1107b1b7c072edd992',
+            'funsor',
```
neerajprad, line 50: So the latest release of funsor has this commit?

fehiepsi: oh, good question! I need to verify this...

fehiepsi: yes, it has.

neerajprad (approved):

fehiepsi:
Thanks, @neerajprad!

PR #784, opened 2020-10-21
Adding conda install instructions

### LINKED ISSUE #658: installation / deployment to conda-forge
Is anyone working on making this library available via conda-forge?

freddyaboulton:
Fix #658

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,6 +204,11 @@ git clone https://github.com/pyro-ppl/numpyro.git
 pip install -e .[dev]  # contains additional dependencies for NumPyro development
 ```
 
+You can also install NumPyro with conda:
+```
+conda install -c conda-forge numpyro
+```
+
 ## Frequently Asked Questions
 
 1. Unlike in Pyro, `numpyro.sample('x', dist.Normal(0, 1))` does not work. Why?


fehiepsi:
Thanks, @freddyaboulton! In my opinion, it is better to not officially support `conda` installation for now due to several reasons:
+ Conda installation is not officially supported by jax.
+ Conda-forge jaxlib does not support gpu installation. Installing `numpyro` with conda will reinstall those installed-with-pip jax/jaxlib packages IIUC.
+ In numpyro release, we pin `jax`/`jaxlib` to specific versions because jax devs update their packages more often and sometimes, those new versions break numpyro user code. By pinning the versions, we are more flexible to develop new features while not having to keep updated with jax releases.

How about just considering `conda` installation as unofficially supported for now? We can reiterate the issue in the future when jax officially supports conda (I think the above issues will be resolved by that time).

freddyaboulton:
@fehiepsi Sounds good to me! Thanks for explaining.

fehiepsi (approved):
Following the recent move in Pyro and it seems that the conda-forge version is working well, we can start officially supporting this. Thanks for contributing @freddyaboulton!

PR #791, opened 2020-10-30
Update documentation page

### LINKED ISSUE #788: Merge tutorial page to readthedocs
Currently, tutorials are not exposed to the main [numpyro site](http://num.pyro.ai/en/stable/). It would be nice to follow https://github.com/pyro-ppl/pyro/issues/2670 and integrate the tutorial site to readthedocs. Not only it will be more accessible for users, but also:
+ resolves https://github.com/pyro-ppl/numpyro/issues/695 . Everytime we update `master` and  notebooks/examples, readthedocs will automatically it for us.
+ reduces the overhead to render the site http://pyro.ai/numpyro/.

Also, it would be nice to:
- ~[ ] automatically render the changelogs~ this requires internet access, which potentially causes rendering issues in readthedocs
- [x] add open in colab link for tutorials (see https://github.com/pyro-ppl/pyro/issues/2581)
- [x] links to other tutorial sites (e.g. Statistical rethinking) **edit**: this is already mentioned in a section of README

fehiepsi:
Resolves #788

This PR does:
+ Merge tutorial site to docs site in the form of galleries
+ Add colab link in the front of each tutorial
+ Add README to docs site
+ Add link to changelog

**Note** any thumbnail image can be put into `docs/source/_static/img/examples` or `docs/source/_static/img/tutorials` folders . It should have the same filename with the corresponding tutorial/example.

TODO
- [x] Add custom image for examples
- ~[ ] Add developer guide~ defer to another pr

Current results
rtd


fritzo:
Looks good to me. It's a little surprising to see api docs pushed down to the bottom, but I guess I usually access reference docs via search rather than the TOC, so I guess this is fine. I think the only times I use API docs are to look up one of many distributions, transforms, or handlers, and in those cases I do navigate to the Distributions, Transforms, or Handlers section via the TOC.

fehiepsi:
Thanks, Fritz! The only reason that I pushed it down is to make the galleries more visible in the index page. But it seems that I can simply set max_depth to a smaller number for API section to achieve that purpose. I will revert it back to the top position. :)

eb8680:
This is great! It's so much easier for people to orient themselves this way, and it also just feels like there's a lot more documentation now despite not adding any new content.  The Colab links look great too.

Does this affect the URLs of the examples?  It would be nice to preserve those if possible, since there are lots of forum posts that link to tutorials.

fritzo:
:+1: to ensuring links work so as to preserve old forum urls

fehiepsi:
Currently, tutorials stay at `pyro.ai/numpyro/` and examples stay at `pyro.ai/numpyro/examples/`. With this PR, tutorials (with the same filenames) are stored at `num.pyro.ai/en/[version]/tutorials` and `num.pyro.ai/en/[version]/examples` (where `[version]` is `stable`, `0.4.0`,... - a nice thing with this way is: we can keep the old "released" versions of tutorials/examples, which works with the old version of pyro/numpyro). I think we can make `pyro.ai/numpyro` and `pyro.ai/numpyro/examples` redirect to `num.pyro.ai/en/0.4.1/tutorials` and `num.pyro.ai/en/0.4.1/examples`.

@neerajprad Does this make sense at all? I just guess that if we write a script to redirect all html links in [this folder](https://github.com/pyro-ppl/website/tree/master/numpyro) following the pattern of Pyro [time series](https://github.com/pyro-ppl/website/blob/master/time/index.html) then old links will be preserved.
```
<script type="text/javascript">
window.location = "http://pyro.ai/examples/#time-series"
</script>
```
But I don't have real experience with websites so I am worried that I miss something.

fehiepsi:
@fritzo Does this look better?

numpyro1
numpyro2

fritzo:
@fehiepsi it looks ok either way, up to you. I do like that, if the api reference is in the top, that it is collapsed to a single line by default

fritzo:
GPyTorch also has nice looking documentation. They keep api reference at the bottom, but collapse tutorials by default so the TOC fits on one page. I like how that looks https://docs.gpytorch.ai/en/v1.1.1/

neerajprad:
>  I think we can make pyro.ai/numpyro and pyro.ai/numpyro/examples redirect to num.pyro.ai/en/0.4.1/tutorials and num.pyro.ai/en/0.4.1/examples

I think bumping up the version number in the URL each time isn't best for SEO - it might be better to have a static URL (stable/latest) that points to the latest version number (just like our usual docs page). Then we can just redirect `pyro.ai/numpyro/examples` to `num.pyro.ai/en/latest/examples` etc. WDYT?

fehiepsi:
> the api reference is in the top, that it is collapsed to a single line by default

Make sense to me. :)

> They keep api reference at the bottom, but collapse tutorials by default so the TOC fits on one page.

Actually, many Python packages put API reference at the bottom. :D  But I don't have a preference for this. I think making some tutorial thumnails visible on the index page is more accessible (however, this points to further questions: what is a good single figure to, together with the title, reflect the content of a tutorial - which is interesting to think about, at least for me :) ). Otherwise, readers will go to the index page and use the side bar to walk through the content, instead of using the nice (?!) gallery.

> redirect pyro.ai/numpyro/examples to num.pyro.ai/en/latest/examples

Actually, this is what I did at https://github.com/pyro-ppl/website/pull/9 because those tutorials are only available in the `latest` rtd branch. Thanks for your input, @neerajprad!

neerajprad:
Would this rebuild all tutorials when we do `make docs`? It will be nice to keep docs build fast for quick testing since some of the MCMC examples take a long time to run.

docs/source/conf.py
```diff
@@ -109,6 +127,102 @@
...
+with open('getting_started.rst', 'wt') as f:
+    f.write(nbsphinx.markdown2rst(text))
+
+
+# -- Copy notebook files
+
+if not os.path.exists('tutorials'):
+    os.makedirs('tutorials')
+
+# remove files that are updated or not available in notebooks/source
```
neerajprad, line 178: I did not exactly follow this part - why do we need to copy these files from `notebooks/source` directory? Also will these files persist after we have built the documentation?

fehiepsi:
> Would this rebuild all tutorials when we do make docs?

Yes, this will render the current tutorials/examples, not execute them because it will be slow. Rendering the gallery seems to be much faster than building the documentation.

docs/source/conf.py
fehiepsi: Yes, these files will persist so I have to add some entries to [.gitignore](https://github.com/pyro-ppl/numpyro/pull/791/files#diff-bc37d034bad564583790a46f19d807abfe519c5671395fd494d8cce506c42947R6) to skip committing those files. This works but not a good solution. I don't think `sphinx` can provide an easy way to specify files outside of `source` directory but let me check it again.

neerajprad: I see. If that doesn't work, we can temporarily create these directories but clean it up before exiting.

fehiepsi: I just took a look at gpytorch and they did the same thing: copy [examples folder](https://github.com/cornellius-gp/gpytorch/blob/8f9b44fc57dbb0a13b568946f07a37e9332f92c4/docs/source/conf.py#L107) and add an entry to `.gitignore`. :D

> clean it up before exiting

If we clean those up then no need to add those md5hash codes, so I'll remove those codes too. I think we can extend the `make docs` command for this. Let me see.

fehiepsi: @neerajprad I added `git clean -dfx docs/source/` to `Makefile`, which does
```
Removing docs/source/examples/
Removing docs/source/getting_started.rst
Removing docs/source/tutorials/
```
at the end of `make docs`.

jpchen:
Looks great, it definitely is much cleaner to have everything serviced and version controlled through rtd. WRT endpoints, the better way would be to redirect at the DNS level not the [application level](https://en.wikipedia.org/wiki/List_of_network_protocols_(OSI_model)), but manual URL redirects is fine for maintaining deprecated links.

docs/source/conf.py
neerajprad: Sounds good. Minor nit, but I think this will also delete any other new files and leave untracked files if the whole thing failed due to other reasons. Can we just wrap the whole operation in a try, finally block and only delete the directories created?

```python
try:
  # create directories for sphinx
finally:
  # delete these directories (check if they exist first)
```

fehiepsi: I just added `-` in front of sphinx-build command so `git clean` will be executed even if the build fails for some reason. I think it is better because try/catch can't deal with `examples` folder (which is created by the `sphinx-gallery` extension).

neerajprad (approved):

fehiepsi:
Thanks for the helpful reviews, everyone! <3

PR #796, opened 2020-11-01
Bernoulli distribution returns an int instead of a boolean

### LINKED ISSUE #789: Bernoulli returns a Boolean instead of an integer
Currently, in Bernoulli distribution, we sample `boolean` arrays but using `arange(2)` for enumerate support. We should use 0, 1 values in `sample` method for consistency.

fehiepsi:
Resolves #789.

This also makes `sample` return the same dtype (int) as `enumerate_support`.

Tests should pass because our `Bollean` constraint actually checks if `x == 1` or `x == 0`.

### DIFF
--- a/numpyro/distributions/discrete.py
+++ b/numpyro/distributions/discrete.py
@@ -83,7 +83,7 @@ class BernoulliProbs(Distribution):
 
     def sample(self, key, sample_shape=()):
         samples = random.bernoulli(key, self.probs, shape=sample_shape + self.batch_shape)
-        return samples.astype(canonicalize_dtype(jnp.int64))
+        return samples.astype(jnp.result_type(samples, int))
 
     @validate_sample
     def log_prob(self, value):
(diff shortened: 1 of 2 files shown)


numpyro/distributions/discrete.py
```diff
@@ -82,7 +82,8 @@ def __init__(self, probs, validate_args=None):
         super(BernoulliProbs, self).__init__(batch_shape=jnp.shape(self.probs), validate_args=validate_args)
 
     def sample(self, key, sample_shape=()):
-        return random.bernoulli(key, self.probs, shape=sample_shape + self.batch_shape)
+        samples = random.bernoulli(key, self.probs, shape=sample_shape + self.batch_shape)
+        return samples.astype(canonicalize_dtype(jnp.int64))
```
neerajprad, line 86: For clarity, this will only coerce to int64 if the corresponding flag is set but by default this should use int32, right? If that's the case, I wonder if there is a more idiomatic way of doing this in JAX. This can be a source of confusion for anyone new to the codebase.

fehiepsi: Looking at [jax.numpy](https://github.com/google/jax/blob/master/jax/_src/numpy/lax_numpy.py) source code, it seems that we can use
```
samples.astype(jnp.result_type(x, int))
```
Does this look better to you?

neerajprad: That seems fine - basically a `jnp.array(True) * 0` will give result in a `int32` type. Can we use this pattern?

fehiepsi: Sure! :) I'll make that change.

neerajprad (approved):
Thanks for making this fix!

PR #799, opened 2020-11-03
Simplify training phase of SVI

### LINKED ISSUE #792: Simplify training phase of SVI
Currently, to train SVI, users need to know either
+ `jax.lax.fori_loop`
+ `jax.lax.scan`
+ or `jax.jit`

Either way, users code will look like
```
svi = SVI(model, guide, ...)
svi_state = svi.init(rng_key)
def body_fn:
    ...
svi_state = fori_loop/scan/python loop
params = svi.get_params(svi_state)
```

To me, this is unfortunate and in most of the cases (given the fact that we already support subsampling via `numpyro.subsample`), the model input will be constant. I think it is better to support a high-level api for the above job. For example, we can add a method `svi.run` (which mimics `mcmc.run`) which do all the job, kind of
```
svi = SVI(model, guide, ...)
svi.run(rng_key, num_steps, *args, progress_bar=True, **kwargs)
params = svi.get_params()
losses = svi.get_losses()
```

With this, we
+ need not to know what svi_state is
+ need not to know how to use those jax transforms
+ switch between with progress_bar and no progress_bar easily
+ can see that the code is more intuitively

fehiepsi:
Resolves #792. This PR will either add a utility `svi_run`??? or a method `SVI.run`??? to simplify the training phase of SVI.

For the latter, we should keep in mind that `SVI.run` is deprecated in Pyro. However, I guess we needn't worry much about it because Pyro SVI does not have `init`, `update`, `evaluate` methods too.

But the question is: should `SVI.run` returns the last state and the collected losses?


neerajprad:
> But the question is: should SVI.run returns the last state and the collected losses?

This gets into the same kind of issues like `mcmc.run`. How about we follow the same pattern and `svi.run` not return anything by default. To get parameters we can do `svi.get_params()`. It should also print the loss so that it is easy to see whether it is converging and we can always add other diagnostic methods later. Also, it is probably best to put an experimental warning on this interface.

fehiepsi:
> svi.run not return anything by default. To get parameters we can do svi.get_params()

I first went with that design (it seems to be cleanest) but got stuck at getting the losses. Could you be more explicit about this? I feel `loss` is like `extra_fields` in MCMC (actually, we can add `loss` to `SVIState`). So some things like `svi.run(..., extra_fields=("loss",))`, `svi.get_extra_fields` can make SVI and MCMC closer. But I can't see if getting other information, besides `loss`, from `svi_state` is helpful at all... haizz, I have many thoughts back and forth for this. Let me make a draft version first to make things clearer.

neerajprad:
Please disregard my earlier comment. If `SVI.run` is a convenience (experimental) method, then it does not make sense to add more methods just to support getting params and losses. In that case, it should probably return a result object or named tuple with losses and params directly. It should also probably print losses at some interval so that we can see if the loss is going down.

I am not sure if there are other design alternatives. The current pattern of getting `init_state` and updating it by calls to `svi.update` is the most idiomatic one as far as JAX goes (that's how NNs, optimizers etc work). We can make it closer to Pyro but we'll need to store parameters in the class and wrap it in a poutine every single time `step` is called which seems questionable design wise.

numpyro/infer/svi.py
```diff
@@ -137,6 +135,45 @@ def update(self, svi_state, *args, **kwargs):
...
+            with tqdm.trange(num_steps) as t:
+                for i in t:
+                    svi_state, loss = jit(body_fn)(svi_state, None)
+                    t.set_postfix_str("Loss: {:.4e}".format(loss), refresh=False)
+                    losses.append(loss)
+            losses = jnp.stack(losses)
+        else:
+            svi_state, losses = lax.scan(body_fn, svi_state, None, length=num_steps)
+
+        return self.get_params(svi_state), losses
```
neerajprad, line 175: Do we foresee other things that we might want to return in the future (I can't think of anything at the moment)? If so, it will be better to use a named tuple so that we don't need to introduce a backwards incompatible change in the future.

fehiepsi: Makes sense to me. I'll make the change.

fehiepsi: One thing we might want to return is the last `svi_state`, which is used for evaluation. But I am not sure how useful it is. For evaluation, we can construct `svi_state` back from `params` by using `SVIState(self.optim.init(params), rng_key)`. To me, it is better to have a method `get_state(self, params, rng_key)` if `svi_state` is needed.

The current return values seem good to me.

numpyro/infer/svi.py
```diff
@@ -137,6 +144,46 @@ def update(self, svi_state, *args, **kwargs):
...
+            holds the optimized values at :class:`numpyro.param` sites,
+            and `losses` is the collected loss during the process.
+        :rtype: SVIRunResult
+        """
+        def body_fn(svi_state, carry):
+            svi_state, loss = self.update(svi_state, *args, **kwargs)
+            return svi_state, loss
+
+        svi_state = self.init(rng_key, *args, **kwargs)
+        if progress_bar:
```
neerajprad, line 174: I think that for SVI printing the loss at some regular frequency is quite useful to see the trend - is the loss going down, how fast is it going down and whether it has converged. Using a progress bar might not be as useful in that regard since the trend isn't quite as clear. If we are to use a progress bar, how about adding % decrease from initial loss value ("x% decrease") along with the loss to help visualize the trend.

numpyro/infer/svi.py
neerajprad: I see - yes, we don't need to add anything else right now. I was only asking since if we wanted to have such an option for the future, we could build that flexibility in right now.

numpyro/infer/svi.py
fehiepsi: I agree that more diagnostic infos would be pretty helpful here.

This requires more control values for the report (or we can hard-code them to some reasonable values). Currently, I think that we can return the average loss (over the last ... steps) to account for the fluctuations. About `% decrease`, should we record it with respect to some checkpoint `iter` (initial losses are usually pretty high)?

neerajprad: > This requires more control values for the report (or we can hard-code them to some reasonable values). Currently, I think that we can return the average loss (over the last ... steps) to account for the fluctuations.

I think we can just refresh the bar less frequently or do some exponential smoothing if the fluctuations are too high. 

> About % decrease, should we record it with respect to some checkpoint iter (initial losses are usually pretty high)?

It should be in the log scale so can't be too bad? I have also seen problems, specially with bad initialization, where it doesn't go down much at all.

fehiepsi: > we can just refresh the bar less frequently

I think we should do this. Currently, I see the losses change so fast that I can't keep up with them.

> specially with bad initialization

Haha, I have seen the reverse more: bad initialization will have a very high loss, then it dropped down like 1e6 to 100... XD

Just to confirm your recommendation, "x% decrease" is computed as: (losses[-1] / losses[0]) * 100? What happens when `loss` is negative?

neerajprad: > "x% decrease" is computed as: (losses[-1] / losses[0]) * 100? What happens when loss is negative?

I meant ((losses[0] - losses[-1])/losses[0]) * 100. This should be a positive quantity if the optimization is working as expected (it can on occasion go negative I think but should be rare; in any case, we shouldn't change what we display).

> bad initialization will have a very high loss, then it dropped down like 1e6 to 100.

In cases like this, it will start off with 0% decrease and rapidly go to 99% decrease. That might not be super useful but using some other loss value as baseline will be harder to explain. Another thing we could do is to just mention the initial loss in the progress bar instead of reporting a percent decrease if you think that will be more generally useful, or just not have a progress bar at all and report loss after every 5% progress.
```
[5%  -- Iter: 200] Loss: 100
[10% -- Iter: 400] Loss: ...
```

fehiepsi: > Another thing we could do is to just mention the initial loss

This looks more readable to me. And I think it is good idea to report loss every 5% progress. I also prefer reporting average loss of the last 5% progress. Let me see if I can integrate all those information in progbar.

> just not have a progress bar

I think looking at progbar slowly runs is more fun than waiting for the next print. :D

fehiepsi: @neerajprad I just updated the progbar. Currently, it is like this
```
100%|██████████████████████████████████████████████| 10000/10000 [00:11<00:00, 890.02it/s, init loss: 8.6413, avg. loss [9501-10000]: -1.7689]
```

neerajprad:
Looks pretty good to me!

neerajprad (approved):

PR #801, opened 2020-11-03
Raise errors early if missing plates for models with discrete latent variable

### LINKED ISSUE #790: Warn on missing plate
Requested by @fehiepsi: Briefly, for models with discrete latent variables, sampling outside of plate context yields some fairly confusing stack traces during inference. Numpyro could instead warn during model validation. 

https://forum.pyro.ai/t/global-latent-variables-numpyro-crash/2313

fehiepsi:
Resolves #790 

This is a temporary fix for #790. If plates are missing, we will get the error
```
     56     def __exit__(self, *args, **kwargs):
---> 57         assert _PYRO_STACK[-1] is self
     58         _PYRO_STACK.pop()
```
I know about this issue while playing with some discrete models, but it is hard for users to guess where the error comes from. So this PR adds a validation code for that.

### DIFF
--- a/numpyro/infer/util.py
+++ b/numpyro/infer/util.py
@@ -357,6 +357,21 @@ def _guess_max_plate_nesting(model_trace):
     return max_plate_nesting
 
 
+# TODO: follow pyro.util.check_site_shape logics for more complete validation
+def _validate_model(model_trace):
+    # XXX: this validates plate statements under `enum`
+    sites = [site for site in model_trace.values()
+             if site["type"] in ("sample", "control_flow")]
+
+    for site in sites:
+        batch_dims = len(site["fn"].batch_shape)
+        if site.get('_control_flow_done', False):
+            batch_dims = batch_dims - 1  # remove time dimension under scan
+        plate_dims = -min([0] + [frame.dim for frame in site["cond_indep_stack"]])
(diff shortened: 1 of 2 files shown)


numpyro/infer/util.py
```diff
@@ -357,6 +357,21 @@ def _guess_max_plate_nesting(model_trace):
     return max_plate_nesting
 
 
+# TODO: follow pyro.util.check_site_shape logics for more complete validation
```
neerajprad, line 360: Do distribution dims follow the same convention as in Pyro, i.e. [enum dims][batch dims][event dims], where the number of `batch_dims` equals the maximum number of nested plates in the model? I would like to better understand what class of errors would this validation logic catch and what would it miss.

fehiepsi: Yes, it follows the same convention (except for `scan`, we have `[time_dim][enum dims][batch dims][event dims]`). (this validation does not check for `enum dims` - which is defered to funsor logic)

Pyro validation has some logic to check if some expected shape is met. This only checks the number of batch dimensions agrees with the dimension of the leftmost plate.

eb8680 (approved):
This makes sense to me, hopefully it helps people fix their mixture model code.

PR #804, opened 2020-11-05
Add prng_key primitive

### LINKED ISSUE #798: Simplify the patterns with PRNGIdentity
Currently, `PRNGIdentity` is a special distribution that returns a random key under `seed` handler. One issue with it is we need to have extra logics to filter this site out at various places. For examples at
+ [elbo](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/infer/elbo.py#L159)
+ [log_density](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/infer/util.py#L53)
+ autoguide [here](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/infer/util.py#L53) and [here](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/infer/util.py#L53)
+ [get model's transform](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/infer/util.py#L271)
and I think we will get some issues if we use `contrib.funsor` with this distribution.

We can add "filter-out" logics to those places (there are only a few of them) but I think it is easy to forget. In addition, if this site appears under `plate` statement, then it becomes an `ExpandedDistribution`, so the logics at above places do not apply anymore.

This issue is created to discuss about ways to simplify the pattern of using PRNGIdentity. Currently, generating a random key is only used in AutoGuide and `numpyro.module`, so I think we are flexible to create a new pattern for it. One possible solution is to make a primitive to generate a PRNGKey, instead of relying on `sample` primitive.

fehiepsi:
This PR tries to resolve #798. Based on the changes, it seems that making this primitive helps.

### DIFF
--- a/numpyro/distributions/discrete.py
+++ b/numpyro/distributions/discrete.py
@@ -439,7 +439,7 @@ class PRNGIdentity(Distribution):
     def __init__(self):
         warnings.warn("PRNGIdentity distribution is deprecated. To get a random "
                       "PRNG key, you can use `numpyro.prng_key()` instead.",
-                      DeprecationWarning)
+                      FutureWarning)
         super(PRNGIdentity, self).__init__(event_shape=(2,))
 
     def sample(self, key, sample_shape=()):
--- a/numpyro/distributions/discrete.py
+++ b/numpyro/distributions/discrete.py
@@ -437,6 +437,9 @@ class PRNGIdentity(Distribution):
     is_discrete = True
 
     def __init__(self):
+        warnings.warn("PRNGIdentity distribution is deprecated. To get a random "
+                      "PRNG key, you can use `numpyro.prng_key()` instead.",
+                      DeprecationWarning)
         super(PRNGIdentity, self).__init__(event_shape=(2,))
 
     def sample(self, key, sample_shape=()):
--- a/docs/source/primitives.rst
+++ b/docs/source/primitives.rst
@@ -27,6 +27,10 @@ deterministic
 -------------
 .. autofunction:: numpyro.primitives.deterministic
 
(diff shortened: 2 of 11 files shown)


neerajprad (approved):
LGTM, this was a much needed cleanup to the interface. 
 - Are there any differences worth noting w.r.t. vmap, composability with other effect handlers, etc. My impression is that this should not lead to any unexpected differences.
 - Should we deprecate `PRNGIdentity` now?

fehiepsi:
Thanks for reviewing, @neerajprad! Currently, we have "guards" (e.g. `if msg['type'] not in ('sample',): return`) in each handler so no effect handler beside `seed` can affect this primitive. I am not sure about `vmap` so I added some tests for it. It seems to work as expected to me.

neerajprad (dismissed):

numpyro/distributions/discrete.py
```diff
@@ -437,6 +437,9 @@ class PRNGIdentity(Distribution):
     is_discrete = True
 
     def __init__(self):
+        warnings.warn("PRNGIdentity distribution is deprecated. To get a random "
+                      "PRNG key, you can use `numpyro.prng_key()` instead.",
+                      DeprecationWarning)
```
neerajprad, line 442: Sorry, I didn't notice this. Could you change this to `FutureWarning` instead? See https://docs.python.org/3/library/warnings.html#warning-categories.

fehiepsi: Oh, thanks! You keep reminding me about this but I forgot. >__<

neerajprad (approved):

Issue #781, opened 2020-10-16
Usability issues when warmup samples are collected


neerajprad:
If we have `collect_warmup=True` when running warmup separately we do not discard samples from the warmup phase. However, when we later run `mcmc.run` and call `get_samples`, the warmup samples are discarded. 
 - Would it make sense to return the previously collected warmup samples, in this case?
 - Also when collecting warmup samples, the progress bar reads `Sample`, which can be confusing.

fehiepsi:
Yeah, we need to adjust the message for warmup. For the first point, I think we can call `mcmc.get_samples()` after calling `warmup()`. Do you suggest having a different method to get warmup samples (which is stored by some hidden attributes _warmup_states e.g.)?

neerajprad:
> Do you suggest having a different method to get warmup samples (which is stored by some hidden attributes _warmup_states e.g.)?

Yes, something like that. Basically, if we have collected warmup samples, then it will be nice to have `get_samples` return those too. Stan has a way of getting warmup samples by specifying `inc_warmup`, but since we discard warmup samples by default, it isn't possible to have `include_warmup` argument in `get_samples`. There are a few design choices:

To collect samples from warmup as well:
 - the user needs to `mcmc.warmup(collect_warmup_samples=True, ..)` followed by `mcmc.get_samples`, then do `mcmc.run()` and finally concatenate the warmup samples with the new samples from `mcmc.get_samples`. - This is what we have now.
 - do `mcmc.warmup(collect_warmup_samples=True, ...); mcmc.run(...); mcmc.get_samples(...)` which should return the warmup samples as well since we specified that these should be included.
 - same as above but include a keyword argument `include_warmup=False` to `mcmc.get_samples(...)` which will only include the warmup samples if they were collected and if `include_warmup=True`.

fehiepsi:
Yeah, each of those design choices makes sense to me. The third one is what we have first, but then refactor out to clean up the interface if I remember correctly. Please go with any you prefer. But if we go with the later ones, I feel it might be better to have a new method to get those warmup samples. That way, `get_samples` has a consistent meaning. WDYT?

fehiepsi:
I think we can collect warmup samples using the current API, as in the first design choice above. Let's address the API issue if we find the current usage is a bit inconvenient.

PR #805, opened 2020-11-05
Fix progbar message when running warmup

fehiepsi:
Addresses part of #781. This displays the correct label for warmup phase, when running `warmup`.

I also remove `hmc`, which has been deprecated for a long time.


neerajprad (approved):

neerajprad:
Thanks for the fix!

PR #808, opened 2020-11-06
Try to fix colab link

fehiepsi:
I am not sure why in my machine, colab link is correct, while at http://num.pyro.ai/en/latest/tutorials/bayesian_regression.html, colab link has an extra "v" at tag `0.4.1`.

~My only assumption is on readthedocs, `tag = "v" + pip_version if version[:3] < "0.3" else pip_version` will give `tag = "v0.4.1"` given `pip_version = "0.4.1"`. It is pretty strange.~

**Edit:** the bug is trivial. I should use `pip_version` instead of `version` when computing `tag`.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -133,7 +133,7 @@ with open('../../numpyro/version.py') as f:
             break
     pip_version = line.rstrip().split(" = ")[-1].strip("'").strip('"')
     # resolve the issue: tags for versions before 0.3.0 have the prefix 'v'.
-    tag = ("v" + pip_version) if version[:3] < "0.3" else pip_version
+    tag = "v" + pip_version if pip_version[:3] < "0.3" else pip_version
 
 
 # This is processed by Jinja2 and inserted before each notebook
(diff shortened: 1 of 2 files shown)


docs/source/conf.py
```diff
@@ -48,6 +48,8 @@
     from numpyro import __version__  # noqaE402
     version = __version__
 
+    html_context = {'github_version': 'master'}
```
fehiepsi, line 51: On readthedocs, `github_version` always points to the corresponding tag (e.g. `master`, `0.4.1`, `v0.2.0`). However, when building locally, `github_version` will not be defined. This makes sure that colab link is valid even for local development. This is not important though.

docs/source/conf.py
```diff
@@ -146,13 +139,13 @@
     <div class="admonition note">
       Interactive online version:
       <span style="white-space: nowrap;">
-        <a href="https://colab.research.google.com/github/pyro-ppl/numpyro/blob/{tag}/{{ docname }}">
+        <a href="https://colab.research.google.com/github/pyro-ppl/numpyro/blob/{{ env.config.html_context.github_version }}/{{ docname }}">
```
fehiepsi, line 142: This change will make colab link always points to the correct location. :)

neerajprad (approved):

PR #809, opened 2020-11-07
Add test for beta plate binomial conjugacy

fehiepsi:
This is used to test for funsor PR: https://github.com/pyro-ppl/funsor/pull/392

### DIFF
--- a/numpyro/handlers.py
+++ b/numpyro/handlers.py
@@ -277,18 +277,22 @@ class collapse(trace):
         COERCIONS.append(self._coerce)
         return super().__enter__()
 
-    def __exit__(self, *args, **kwargs):
+    def __exit__(self, exc_type, exc_value, traceback):
+        if exc_type is not None:
+            return super().__exit__(exc_type, exc_value, traceback)
         import funsor
 
         _coerce = COERCIONS.pop()
         assert _coerce is self._coerce
-        super().__exit__(*args, **kwargs)
+        super().__exit__(exc_type, exc_value, traceback)
 
         # Convert delayed statements to pyro.factor()
         reduced_vars = []
         log_prob_terms = []
         plates = frozenset()
         for name, site in self.trace.items():
+            if site["type"] != "sample":
+                continue
             if not site["is_observed"]:
                 reduced_vars.append(name)
             dim_to_name = {f.dim: f.name for f in site["cond_indep_stack"]}
--- a/test/test_handlers.py
+++ b/test/test_handlers.py
@@ -569,7 +569,6 @@ def test_collapse_beta_bernoulli():
     svi.update(svi_state)
 
 
-@pytest.mark.xfail(reason="missing pattern in Funsor")
 def test_collapse_beta_binomial_plate():
     data = np.array([0., 1., 5., 5.])
 
--- a/numpyro/contrib/ns.py
+++ b/numpyro/contrib/ns.py
@@ -1,0 +1,40 @@
+from numpyro.handlers import reparam
+from numpyro.infer.reparam import Reparam
+from numpyro.infer import log_likelihood
+
+from jaxns.nested_sampling import NestedSampler
+from jaxns.prior_transforms import PriorChain
+
+
+class UnitCubeReparam(Reparam):
+    # support distributions with quantiles implementation
+    def __call__(self, name, fn, obs):
+        assert obs is None, "TransformReparam does not support observe statements"
+        fn, batch_shape = self._unexpand(fn)
+        # if fn is transformed distribution, get transform and base_dist, reparam the base_dist
+        # TODO: extract quantiles function
+
+
+class NestedSampling:
+    def __init__(self, model, num_live_points, *, sampler_name='slice', **ns_kwargs):
+        self.model = model
+        self.num_live_points = num_live_points
+        self.sampler_name = sampler_name
+        self.ns_kwargs = ns_kwargs  # max_samples, termination_frac,...
+
+    def run(self, rng_key, *args, **kwargs):
+        # Step 1: reparam the model so that latent sites have Uniform(0, 1) priors
+        reparam_model = reparam(self.model,
+                                config=lambda msg: None if msg.get("is_observed") else UnitCubeReparam())
+
+        # Step 2: compute the likelihood of the model
+        def ll_fn(*params):
+            return list(log_likelihood(reparam_model, params, *args, batch_ndims=0, **kwargs).values())[0]
+
+        # Step 3: use NestedSampler with empty prior chain
+        prior_chain = PriorChain()
+        ns = NestedSampler(ll_fn, prior_chain, collect_samples=True, sampler_name=self.sampler_name)
+        results = ns(rng_key, self.num_live_points)
+        samples = results["samples"]
+        # TODO: transform samples back or rerun reparam_model to get deterministic sites
+        return samples


test/test_handlers.py
```diff
@@ -591,6 +590,91 @@ def guide():
...
+    def model():
+        x = numpyro.sample("x", dist.Exponential(1))
+        with handlers.collapse():
+            with numpyro.plate("d", d):
+                # TODO: verify that to_event works here
+                beta0 = numpyro.sample("beta0", dist.Normal(0, 1).expand([S]).to_event(1))
+                # TODO: address beta0 is a str, which cannot do infer_param_domain
+                beta = numpyro.sample("beta", dist.MultivariateNormal(beta0, jnp.eye(S)))
+            # FIXME: beta is a string here, how to apply numeric operators
+            mean = jnp.ones((T, d)) @ beta
```
fehiepsi, line 665: @fritzo @eb8680 looking like this is a good test to have better coverage for `collapse`. I think that we might return a Variable instead of a string, so we don't need to infer domain for `beta0` and here we can use `funsor.ops.matmul`.

fritzo: :+1: We fixed this to return a `Variable` in Pyro https://github.com/pyro-ppl/pyro/pull/2676

Regarding `funsor.ops.matmul` I would prefer the `@` version because it minimizes mention of magical funsors. I think that would work in `beta @ ones` but not in `ones @ beta`. I wonder if we can overload jnp.ones to return `NotImplemented` if faced with a Funsor so that `jnp.ones(...) @ ` would defer to `Funsor.__rmatmul__()`?

fehiepsi: Oops, thanks @fritzo! I forgot about it...

> wonder if we can overload `jnp.array.__matmul__()`

It seems that it works:
```python
import jax.numpy as jnp

class A():
    def __rmatmul__(self, other):
        return "a"

assert jnp.ones(3) @ A() == "a"
```

fritzo: Great, so `jnp.ones(...) @ funsor.Variable(...)` should also work!

test/test_handlers.py
```diff
@@ -591,6 +590,115 @@ def guide():
...
+def test_collapse_diag_normal_plate_normal():
+    d = 3
+    data = np.ones((5, d))
+
+    def model():
+        x = numpyro.sample("x", dist.Normal(0, 1))
+        with handlers.collapse():
+            with handlers.plate("data", len(data)):
+                y = numpyro.sample("y", dist.Normal(x, 1.).expand([d]).to_event(1))
+                numpyro.sample("z", dist.Normal(y, 1.).to_event(1), obs=data)
```
fehiepsi, line 662: This fails due to 2 reasons:
+ `y.output` is Reals[d], which will raise error while infer_param_domain in funsor: `Output mismatch: Reals[2] vs Real`
+ to_event is not available for Funsor distributions.

test/test_handlers.py
```diff
@@ -591,6 +590,115 @@ def guide():
...
+    def model():
+        x = numpyro.sample("x", dist.Normal(0, 1))
+        with handlers.collapse():
+            with numpyro.plate("d", d, dim=-1):
+                beta0 = numpyro.sample("beta0", dist.Normal(x, 1.).expand([d, S]).to_event(1))
+                beta = numpyro.sample(
+                    "beta", dist.MultivariateNormal(beta0, scale_tril=jnp.eye(S)))
+
+            # this fails because beta shape is (3,) while it should be (2, 3)
+            mean = jnp.ones((T, d)) @ beta
```
fehiepsi, line 688: This fails because `beta.output` is `Reals[S]` while we need it to be `Reals[d, S]`

test/test_handlers.py
```diff
@@ -591,6 +590,115 @@ def guide():
...
+            with numpyro.plate("data", T, dim=-1):
+                numpyro.sample("obs", dist.MultivariateNormal(mean, scale_tril=jnp.eye(S)), obs=data)
```
fehiepsi, line 690: This fails due to a similar reason to the `diag_normal_plate_normal` test above:
+ funsor requires `mean.output` is `Reals[S]` while its output after taking `matmul` is `Reals[T, S]`

numpyro/handlers.py
```diff
@@ -263,31 +278,48 @@ def __init__(self, *args, **kwargs):
...
-        if msg["type"] == "sample":
-            if msg["value"] is None:
-                msg["value"] = msg["name"]
+        import funsor
+
+        # Eagerly convert fn and value to Funsor.
+        dim_to_name = {f.dim: f.name for f in msg["cond_indep_stack"]}
+        dim_to_name.update(self.preserved_plates)
+        if isinstance(msg["fn"], (Independent, ExpandedDistribution)):
+            msg["fn"] = _eager_expand_fn(msg["fn"])
```
fehiepsi, line 290: The implementation is almost the same as Pyro version, only different here.

numpyro/primitives.py
```diff
@@ -316,7 +317,8 @@ def process_message(self, msg):
         cond_indep_stack = msg['cond_indep_stack']
         frame = CondIndepStackFrame(self.name, self.dim, self.subsample_size)
         cond_indep_stack.append(frame)
-        if msg['type'] == 'sample':
+        # only expand if fn is Distribution, not a Funsor
+        if msg['type'] == 'sample' and isinstance(msg['fn'], Distribution):
```
fehiepsi, line 321: Not sure if we have a workaround for this. If `fn` is a Funsor, we can't use the below code to expand it.

test/test_handlers.py
fehiepsi: Please ignore my comments. Pyro doesn't suffer from those issues. It seems that we missed something for jax backend in Funsor.

numpyro/handlers.py
```diff
@@ -245,6 +246,20 @@ def process_message(self, msg):
             msg['stop'] = True
 
 
+def _eager_expand_fn(fn):
```
fritzo, line 249: Do I understand correctly that this function eliminates lazy `ExpandedDistribution`s and replaces them with eagerly `tree_map`-expanded distributions, and that this is needed because Funsor distributions do not support Expanded distributions? If so could you add a commend explaining that?

I guess an alternative would be to support `ExpandedDistribution`s in Funsor, or to do this conversion in the funsor layer. @eb8680 does that seem possible?

fritzo (dismissed):

numpyro/handlers.py
fehiepsi: Yes, it is. I'll add a comment for clarity. I thought from our last discussion, this stuff should be treated in Pyro/NumPyro?

eb8680: > I guess an alternative would be to support ExpandedDistributions in Funsor, or to do this conversion in the funsor layer.

We could certainly implement an equivalent of `.expand`. I think the easiest thing to do to address this case would be to add a `to_funsor` pattern for `ExpandedDistribution`.

fehiepsi: > add a `to_funsor` pattern for `ExpandedDistribution`

Yeah, it seems that you both agree to do this. I can take an effort for it.

eb8680: I think it would basically be the same as what you have here, plus a final step that calls `to_funsor` on the eagerly expanded result.

PR #813, opened 2020-11-12
Use tqdm.auto for notebook friendly tqdm

vishwakftw:
In Colab, the progress bar shown when running MCMC doesn't look good because each iteration / sample is printed on a new line.

Using `tqdm.auto` resolves this.

### DIFF
--- a/numpyro/util.py
+++ b/numpyro/util.py
@@ -8,7 +8,7 @@ import random
 import re
 
 import numpy as np
-import tqdm
+from tqdm.auto import tqdm
 
 import jax
 from jax import device_put, jit, lax, ops, vmap
@@ -221,7 +221,7 @@ def fori_collect(lower, upper, body_fun, init_val, transform=identity,
             # special case, only compiling
             jit(_body_fn)(0, vals)
         else:
-            with tqdm.trange(upper) as t:
+            with tqdm(range(upper)) as t:
                 for i in t:
                     vals = jit(_body_fn)(i, vals)
                     t.set_description(progbar_desc(i), refresh=False)


fehiepsi:
@vishwakftw Could you help me incorporate the fix for [svi](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/infer/svi.py#L176)?

I think a better solution to avoid the print issue is to use `tqdm.trange(..., position=0)`. The widget with `auto` is nice but it is not rendered well. What do you think?

vishwakftw:
I thought that the `tqdm.notebook.tqdm` was specifically designed for notebooks, but personally don't have a preference. I can just change it to have the argument `position=0` if that's what you think would be nice.

fehiepsi:
Yes, it is displayed nicely when executed (assume that users already install widget, jupyterlab nb widgets,...). The problem is it is not rendered nicely after saved. For example, [this gist](https://gist.github.com/fehiepsi/ad9b2eb2d3718a5db1bbd2d2e4a93f5d) displays `HBox` and this [Pyro example](http://pyro.ai/examples/jit.html), which uses `tqdm.notebook`, not display progress bar at all. I am not sure how to resolve those issues. About `position=0`, I just tested it in Colab and it seems to work. I am not sure if there is any side effect with that solution. :(

vishwakftw:
I see, sure. I'll make the changes shortly.

vishwakftw:
Strangely, this has seemed to resolve itself on Colab.

fehiepsi:
> this has seemed to resolve itself on Colab

:smile: surprises make the world a better place. Thanks, @vishwakftw!

PR #816, opened 2020-11-18
AutoDelta implementation.

xidulu:
AutoGuide for MAP estimation,

### DIFF
--- a/numpyro/infer/autoguide.py
+++ b/numpyro/infer/autoguide.py
@@ -339,7 +339,7 @@ class AutoDelta(AutoGuide):
     def sample_posterior(self, rng_key, params, sample_shape=()):
         locs = {k: params["{}_{}_loc".format(k, self.prefix)] for k in self._init_locs}
         latent_samples = {
-            k:jnp.broadcast_to(v, sample_shape + jnp.shape(v)) for k, v in locs.items()
+            k: jnp.broadcast_to(v, sample_shape + jnp.shape(v)) for k, v in locs.items()
         }
         return latent_samples
 
--- a/numpyro/infer/autoguide.py
+++ b/numpyro/infer/autoguide.py
@@ -4,6 +4,7 @@
 # Adapted from pyro.infer.autoguide
 from abc import ABC, abstractmethod
 from contextlib import ExitStack
+from numpyro.infer.initialization import init_to_median
 import warnings
 
 import numpy as np
@@ -43,6 +44,7 @@ __all__ = [
     'AutoMultivariateNormal',
     'AutoBNAFNormal',
     'AutoIAFNormal',
+    'AutoDelta',
 ]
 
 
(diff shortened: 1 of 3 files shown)


numpyro/infer/autoguide.py
```diff
@@ -261,6 +263,95 @@ def quantiles(self, params, quantiles):
...
+        super().__init__(model, prefix=prefix, init_loc_fn=init_loc_fn, create_plates=create_plates)
+
+    def _setup_prototype(self, *args, **kwargs):
+        super()._setup_prototype(*args, **kwargs)
+
+        for name, site in self.prototype_trace.items():
+            if site["type"] != "sample" or site["is_observed"]:
+                continue
+
+            event_dim = site["fn"].event_dim + jnp.ndim(self._init_locs[name]) - jnp.ndim(site["value"])
```
fehiepsi, line 301: nit: this can be simplified to `event_dim = site["fn"].event_dim` because we are working on constrained domain (this needs the line `self._init_locs = ...` in the above comment).

numpyro/infer/autoguide.py
```diff
@@ -261,6 +263,95 @@ def quantiles(self, params, quantiles):
...
+                result[name] = numpyro.sample(name, site_fn)
+
+        return result
+
+    def sample_posterior(self, rng_key, params, sample_shape=()):
+        locs = {k: params["{}_{}_loc".format(k, self.prefix)] for k in self._init_locs}
+        with handlers.seed(rng_seed=rng_key):
+            latent_samples = {}
+            for k in locs:
+                latent_samples[k] = numpyro.sample(k, dist.Delta(locs[k]).expand_by(sample_shape))
```
fehiepsi, line 341: This is nice but I wonder if we can simply expand each value in `locs` to avoid extra `random.split` under `handlers.seed`.

fehiepsi (commented):
Very clean! Thanks a lot, @xidulu! Could you also add an entry to [docs](https://github.com/pyro-ppl/numpyro/blob/master/docs/source/autoguide.rst). I just have some comments below.

numpyro/infer/autoguide.py
```diff
@@ -261,6 +263,95 @@ def quantiles(self, params, quantiles):
...
+        automatically as usual. This is useful for data subsampling.
+    """
+    def __init__(self, model, *, prefix='auto', init_loc_fn=init_to_median,
+                 create_plates=None):
+        self.init_loc_fn = init_loc_fn
+        self._event_dims = {}
+        super().__init__(model, prefix=prefix, init_loc_fn=init_loc_fn, create_plates=create_plates)
+
+    def _setup_prototype(self, *args, **kwargs):
+        super()._setup_prototype(*args, **kwargs)
```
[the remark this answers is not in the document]
xidulu: Oh, yeah, I totally forgot about the constraint.  = =

numpyro/infer/autoguide.py
xidulu: My implementation follows the sample_posterior() from AutoNormal,

Do you mean, directly call apply reshape on locs[k] instead of using expand_by ?

fehiepsi: I think we can simply return: `{k: jnp.broadcast_to(sample_shape + jnp.shape(v)) for k, v in locs.items}`. What do you think?

xidulu: Oh I got it, does it mean ```with handlers.seed(rng_seed=rng_key):```  could just be removed? Since delta distribution is deterministic.

fehiepsi: Yes, we don't need rng_key here.

numpyro/infer/autoguide.py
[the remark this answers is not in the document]
xidulu: @fehiepsi 
How about 
```
self._init_locs = {k: v for k, v in self._postprocess_fn(self._init_locs).items()}
```
Will it have the same effect?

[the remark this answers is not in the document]
fehiepsi: Sorry, I should make it clearer. postprocess might return deterministic sites, which we need to filter them out.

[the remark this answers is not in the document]
xidulu: @fehiepsi 
Now the deterministic sites should be filtered out.
https://github.com/pyro-ppl/numpyro/pull/816/files#diff-84384c2f732a05bdda11794d2be295bd6e174d644562a1c565260d289464e95aR297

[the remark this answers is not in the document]
fehiepsi: Thanks! It is a bug in my above code. ^^

fehiepsi (approved):
LGTM, thanks, @xidulu! Please let me know if this is already to merge. :)

xidulu:
> LGTM, thanks, @xidulu! Please let me know if this is already to merge. :)

Go ahead plz.

PR #819, opened 2020-11-19
Raise a better error message when rng_key is missing

fehiepsi:
Currently, when `rng_key` is missing. Users will get a very long error message
```
...
...
~/jax/jax/interpreters/partial_eval.py in trace_to_subjaxpr_dynamic(fun, main, in_avals)
   1172     trace = DynamicJaxprTrace(main, core.cur_sublevel())
   1173     in_tracers = map(trace.new_arg, in_avals)
-> 1174     ans = fun.call_wrapped(*in_tracers)
   1175     out_tracers = map(trace.full_raise, ans)
   1176   jaxpr, out_avals, consts = frame.to_jaxpr(in_tracers, out_tracers)

~/jax/jax/linear_util.py in call_wrapped(self, *args, **kwargs)
    158 
    159     try:
--> 160       ans = self.f(*args, **dict(self.params, **kwargs))
    161     except:
    162       # Some transformations yield from inside context managers, so we have to

~/jax/jax/random.py in _random_bits(key, bit_width, shape)
    309   """Sample uniform random bits of given width and shape using PRNG key."""
    310   if not _is_prng_key(key):
--> 311     raise TypeError("_random_bits got invalid prng key.")
    312   if bit_width not in (8, 16, 32, 64):
    313     raise TypeError("requires 8-, 16-, 32- or 64-bit field width.")

TypeError: _random_bits got invalid prng key.
```
But it is not meaningful. This PR adds a better error message when it happens.

### DIFF
--- a/numpyro/distributions/continuous.py
+++ b/numpyro/distributions/continuous.py
@@ -38,6 +38,7 @@ from numpyro.distributions.distribution import Distribution, TransformedDistribu
 from numpyro.distributions.transforms import AffineTransform, ExpTransform, InvCholeskyTransform, PowerTransform
 from numpyro.distributions.util import (
     cholesky_of_inverse,
+    is_prng_key,
     lazy_property,
     matrix_to_tril_vec,
     promote_shapes,
@@ -63,7 +64,7 @@ class Beta(Distribution):
         super(Beta, self).__init__(batch_shape=batch_shape, validate_args=validate_args)
 
     def sample(self, key, sample_shape=()):
-        assert key is not None
+        assert is_prng_key(key)
         return self._dirichlet.sample(key, sample_shape)[..., 0]
 
     @validate_sample
@@ -91,7 +92,7 @@ class Cauchy(Distribution):
         super(Cauchy, self).__init__(batch_shape=batch_shape, validate_args=validate_args)
 
     def sample(self, key, sample_shape=()):
-        assert key is not None
+        assert is_prng_key(key)
         eps = random.cauchy(key, shape=sample_shape + self.batch_shape)
         return self.loc + eps * self.scale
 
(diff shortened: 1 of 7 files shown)


numpyro/distributions/distribution.py
```diff
@@ -269,6 +269,13 @@ def _validate_sample(self, value):
 
     def __call__(self, *args, **kwargs):
         key = kwargs.pop('rng_key')
+        # FIXME: This also raises error for Delta site.
+        # Is there a better way to raise this error?
```
fehiepsi, line 273: I get a better way is to add an assertation at `sample` method of all distributions that requires `sample`. It is a bit verbose but will resolve the following situations:
+ Delta site does not need `rng_key`
+ The error should happen at the base_dist of ExpandedDistribution, rather ExpandedDistribution itself. Current, when we missed `rng_key` at `ExpandedDistribution`, we got the following error: `ValueError: Missing random key for ExpandedDistribution distribution.`

fehiepsi: What do you think, @neerajprad?

neerajprad: I agree that it will be better to add it in `sample`. That way, if we use the distributions directly, it will still raise an error if the rng key is absent.

fehiepsi: That is a good point. I'll make a change.

fehiepsi: Done. :)

numpyro/distributions/continuous.py
```diff
@@ -63,6 +63,7 @@ def __init__(self, concentration1, concentration0, validate_args=None):
         super(Beta, self).__init__(batch_shape=batch_shape, validate_args=validate_args)
 
     def sample(self, key, sample_shape=()):
+        assert key is not None
```
neerajprad, line 66: Should we instead `assert isinstance(key, random.PRNGKey)` instead? This will also work for `dist.Normal(0., 1.).sample((10,))`, but if you think the error raised in that case is meaningful enough, this should be fine.

fehiepsi: Good point! There is a hidden method in JAX to check if an array is a PRNGKey (random.PRNGKey alone is a function, not class). I think we can copy it to util and use it here.

fehiepsi: @neerajprad I have changed those assertations to `assert is_prng_key(key)`. This looks better now. Thanks!

neerajprad (approved):

numpyro/distributions/continuous.py
neerajprad: Looks great!

PR #821, opened 2020-11-20
Fix a bug at MCMC.print_summary

fehiepsi:
There, we assume `sample_field` is `z`, which might not be true for other MCMC kernels.

Tested on MH example, where I changed `sample_field` to `u`.

I also revise `scale` handler docstring to mention that `scale` can be a ndarray. cc @xidulu

### DIFF
--- a/numpyro/contrib/tfp/distributions.py
+++ b/numpyro/contrib/tfp/distributions.py
@@ -24,7 +24,7 @@ def _get_codomain(bijector):
         return constraints.positive
     elif bijector.__class__.__name__ == "GeneralizedPareto":
         loc, scale, concentration = bijector.loc, bijector.scale, bijector.concentration
-        if not_jax_tracer(concentration) and np.all(concentration < 0):
+        if not_jax_tracer(concentration) and np.all(np.less(concentration, 0)):
             return constraints.interval(loc, loc + scale / jnp.abs(concentration))
         # XXX: here we suppose concentration > 0
         # which is not true in general, but should cover enough usage cases
--- a/numpyro/distributions/util.py
+++ b/numpyro/distributions/util.py
@@ -329,8 +329,7 @@ def signed_stick_breaking_tril(t):
     # we omit the step of computing s = z * z_cumprod by using the fact:
     #     y = sign(r) * s = sign(r) * sqrt(z * z_cumprod) = r * sqrt(z_cumprod)
     z = r ** 2
-    z1m_cumprod = jnp.cumprod(1 - z, axis=-1)
-    z1m_cumprod_sqrt = jnp.sqrt(z1m_cumprod)
+    z1m_cumprod_sqrt = jnp.cumprod(jnp.sqrt(1 - z), axis=-1)
 
     pad_width = [(0, 0)] * z.ndim
     pad_width[-1] = (1, 0)
@@ -477,6 +476,14 @@ def periodic_repeat(x, size, dim):
     result = jnp.repeat(x, repeats, axis=dim)
     result = result[(Ellipsis, slice(None, size)) + (slice(None),) * (-1 - dim)]
     return result
+
+
+# src: https://github.com/google/jax/blob/5a41779fbe12ba7213cd3aa1169d3b0ffb02a094/jax/_src/random.py#L95
+def is_prng_key(key):
+    try:
+        return key.shape == (2,) and key.dtype == np.uint32
+    except AttributeError:
+        return False
 
 
 # The is sourced from: torch.distributions.util.py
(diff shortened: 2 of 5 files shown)


fehiepsi:
@neerajprad Could you help me review this PR? This also fixes an issue of `scale` like the one in [this forum thread](https://forum.pyro.ai/t/scaling-log-probability-with-different-values/2406).

numpyro/contrib/tfp/distributions.py
```diff
@@ -24,7 +24,7 @@ def _get_codomain(bijector):
         return constraints.positive
     elif bijector.__class__.__name__ == "GeneralizedPareto":
         loc, scale, concentration = bijector.loc, bijector.scale, bijector.concentration
-        if not_jax_tracer(concentration) and np.all(concentration < 0):
+        if not_jax_tracer(concentration) and np.all(np.less(concentration, 0)):
```
fehiepsi, line 27: Currently,
```
import jax
import numpy as np
import jax.numpy as jnp

def f(x):
    if not_jax_tracer(y):
        if np.all(y <= 0):
            print("nooo")
    return x + y

y = jnp.arange(10.)
jax.jit(f)(1)
```
will fail because although `concentration` is not a tracer, `concentration < 0` will be a tracer if concentration is a DeviceArray. For validation checks like this, we need to use all numpy ops.

neerajprad: These nuances are getting very subtle as compared to the earlier behavior, and I am not sure if I fully understand when to use `numpy` vs `jax.numpy`, and now there is this additional distinction between numpy and python ops. :)

numpyro/distributions/util.py
```diff
@@ -329,8 +329,7 @@ def signed_stick_breaking_tril(t):
     # we omit the step of computing s = z * z_cumprod by using the fact:
     #     y = sign(r) * s = sign(r) * sqrt(z * z_cumprod) = r * sqrt(z_cumprod)
     z = r ** 2
-    z1m_cumprod = jnp.cumprod(1 - z, axis=-1)
-    z1m_cumprod_sqrt = jnp.sqrt(z1m_cumprod)
+    z1m_cumprod_sqrt = jnp.cumprod(jnp.sqrt(1 - z), axis=-1)
```
neerajprad, line 332: +1!

numpyro/contrib/tfp/distributions.py
fehiepsi: Yeah, for device array, python op `concentration < 0` will be deferred to something like `jnp.less(concentration, 0)` or so. The rule of thumb now is to always use `numpy` ops (not python ops). It is quite unfortunate. >___<

numpyro/infer/mcmc.py
```diff
@@ -491,11 +492,25 @@ def get_extra_fields(self, group_by_chain=False):
...
         # Exclude deterministic sites by default
         sites = self._states[self._sample_field]
         if isinstance(sites, dict) and exclude_deterministic:
-            sites = {k: v for k, v in self._states[self._sample_field].items()
-                     if k in self._last_state.z}
+            state_sample_field = attrgetter(self._sample_field)(self._last_state)
+            # XXX: there might be the case that state.z is not a dictionary but
+            # its postprocessed value `sites` is a dictionary.
+            # TODO: in general, when both `sites` and `state.z` are dictionaries,
+            # they can have different key names, not necessary due to deterministic
```
neerajprad, line 509: Could you explain this further - when will `sites` and `state.z` have different names?

numpyro/contrib/tfp/distributions.py
neerajprad: Thanks for explaining! Any idea what's the larger motivation for this change? Are we moving towards a world where everything will be JITed by default?

numpyro/infer/mcmc.py
fehiepsi: This depends on `post_processing` code, e.g. `state.z = {'ABC_b': 1}` while `sites = {'b': 1}`. The below code will return an empty dict because it thinks `'b'` is an deterministic site.

numpyro/contrib/tfp/distributions.py
fehiepsi: I am not sure. This behavior has appeared since jax 0.2. I haven't caught the fact that `a < 0` will trigger a jax op if a is a DeviceArray until now.

numpyro/infer/mcmc.py
neerajprad: LGTM, but I am thinking that we should keep track of deterministic and latent nodes separately so that we don't need to bother with all this downstream. We can refactor this later.

neerajprad (approved):

PR #822, opened 2020-11-21
expanded haiku/flax modules to take general args/kwargs for init. 

### LINKED ISSUE #705: Integrate with flax/haiku

quattro:
Pull Req to expand the flax/haiku module functions to allow more general network parameterizations. Can now instantiate based on args, kwargs, or input_shape. fixes #705

### DIFF
--- a/numpyro/contrib/module.py
+++ b/numpyro/contrib/module.py
@@ -18,7 +18,7 @@ __all__ = [
 ]
 
 
-def flax_module(name, nn_module, *args, input_shape=None, **kwargs):
+def flax_module(name, nn_module, *, input_shape=None, **kwargs):
     """
     Declare a :mod:`~flax` style neural network inside a
     model so that its parameters are registered for optimization via
@@ -28,6 +28,8 @@ def flax_module(name, nn_module, *args, input_shape=None, **kwargs):
     :param flax.nn.Module nn_module: a `flax` Module which has .init and .apply methods
     :param tuple input_shape: shape of the input taken by the
         neural network.
+    :param kwargs: optional keyword arguments to initialize flax neural network
+        as an alternative to `input_shape`
     :return: a callable with bound parameters that takes an array
         as an input and returns the neural network transformed output
         array.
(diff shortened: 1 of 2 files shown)


quattro:
There are some bugs/typos that need to be fixed beyond the style issues raised in flake

numpyro/contrib/module.py
```diff
@@ -18,7 +18,7 @@
 ]
 
 
-def flax_module(name, nn_module, *, input_shape=None):
+def flax_module(name, nn_module, *args, input_shape=None, **kwargs):
```
fehiepsi, line 21: The reason we have `*` here is to avoid breaking changes for whatever reason in the future. I guess having `*args` is not important. If a user uses an arg here and we change the way it works in the future, things might break. Also, can you also add docstring for `kwargs`?

numpyro/contrib/module.py
```diff
@@ -42,19 +42,19 @@ def flax_module(name, nn_module, *, input_shape=None):
     module_key = name + '$params'
     nn_params = numpyro.param(module_key)
     if nn_params is None:
-        if input_shape is None:
-            raise ValueError('Valid value for `input_shape` needed to initialize.')
+        if input_shape is not None:
+            args = (jnp.ones(input_shape),)
```
fehiepsi, line 46: nit: I guess `jnp.zeros(...)` is a better default input (at least for linear modules). What do you think? This might not be important though.

quattro: I dont think the values are actually used. I'm not an expert in either flax/haiku, but it appears that its just dummy-data to bootstrap the initialized of nn.Module weights. The network weights are initialized according to their respective init plans.

fehiepsi: Thanks for explaining! That makes to me.

numpyro/contrib/module.py
quattro: 👍 sounds good.

fehiepsi (approved):
Looks awesome, thanks a lot for making those primitives more flexible, @quattro! Also, thanks for adding many test cases, they are very useful!

Issue #820, opened 2020-11-19
Sample Tensorflow distributions with `numpyro.sample()` fail inside `with handlers.seed():` statement


cerbelaut:
Hi,
I am working on a project for which I'd like to use tensorflow distributions (the TruncatedNormal distribution, because the TFP implementation allows to define both a lower and an upper bound). However, the following code snippet fails.

```python
with handlers.seed(rng_seed=random.PRNGKey(0)):
    numpyro.sample("normal", tfd.Normal(0, 1))
```

Error:
```
TypeError: iteration over a 0-d array
```

After investigating this, it seems the error is from the `TFPDistributionMixin` class, in the `__call__` method, where the sampled value is return whether or not the `sample_intermediates` is True, while it appears in my sample code the call function is expected to return the sample value along with the intermediates (which could be set to `[]` I assume in my case).

```python
    def __call__(self, *args, **kwargs):
        key = kwargs.pop('rng_key')
        kwargs.pop('sample_intermediates', False)
        return self.sample(*args, seed=key, **kwargs)
``` 

Potential fix:
```python
    def __call__(self, *args, **kwargs):
        key = kwargs.pop('rng_key')
        sample_intermediates = kwargs.pop('sample_intermediates', False)
        if sample_intermediates:
            return self.sample(*args, seed=key, **kwargs), []
        return self.sample(*args, seed=key, **kwargs)
```

fehiepsi:
Whoa, what a bug! I have used several TFP distributions but only in inference, not sampling, so haven't seen this error yet. Thanks, @cerbelaut! Do you want to submit a PR for this, together with the above regression test? :)

cerbelaut:
Sure @fehiepsi, I could find some time during the weekend or next week ! Thanks for the reply ;)

fehiepsi:
Thanks, @cerbelaut!

PR #824, opened 2020-11-23
debug sampling of tfp distributions using numpyro.sample function

cerbelaut:
debug sampling using `numpyro.sample` with tfp distributions by returning `[]` when `sample_intermediates` is `True`.


cerbelaut:
@fehiepsi here is the PR. It seems I can't link this PR to the issue I raised last week about tfp distributions in inference.

fehiepsi (approved):
Awesome, thanks for the fix @cerbelaut!

fehiepsi:
Re link to an issue: you can pass it URL (which is `https://github.com/pyro-ppl/numpyro/issues/820`)  to the comment, like this https://github.com/pyro-ppl/numpyro/issues/820

PR #826, opened 2020-11-24
Implement mixed hmc

### LINKED ISSUE #785: [FR] Mixed HMC for models with discrete latent variables
Currently, NumPyro uses enumeration to marginalize out discrete latent variables. However, when there are many discrete latent variables, using enumeration will be slow because the dimensional size is blown up exponentially w.r.t. the number of discrete latent variables. It seems to me that @StannisZhou's Mixed HMC [1] is a good inference algorithm to use in that case (e.g. the model in #769, which is pretty slow under enumeration).

**References**:
+ [Mixed Hamiltonian Monte Carlo for Mixed Discrete and Continuous Variables](https://arxiv.org/abs/1909.04852)
+ [Discontinuous Hamiltonian Monte Carlo for discrete parameters and discontinuous likelihoods](https://arxiv.org/abs/1705.08510)
+ JAX implementation for Mixed HMC: https://github.com/StannisZhou/mixed_hmc

fehiepsi:
Resolves #785.

### Tasks
- [ ] Implement the core functionality (mimic [hmcecs implementation](https://gist.github.com/fehiepsi/b4a5a80b245600b99467a0264be05fd5))
- [ ] Add tests
- [ ] Add an example

### DIFF
--- a/numpyro/contrib/mcmc/mixed_hmc.py
+++ b/numpyro/contrib/mcmc/mixed_hmc.py
@@ -1,0 +1,5 @@
+# Sketch the prototype
+# + sketch a structure like hmc_ecs prototype
+# + implement discrete step functionality
+# + use enum for non-event discrete sites and mixed for event discrete sites
+# + test on gmm potential


fritzo:
(moving conversation from slack)
> @fritzo Thanks for explaining the M-HMC [1] paper so clearly! Of their experiments, it looks like the GMM and CTR models both have discrete latent variables inside plates with no downstream dependencies outside the plates, therefore you should be able to independently make all reflect/refract decisions in parallel. Their other experiment (variable selection in regression) does not admit the same parallel decision trick though. I can see how they might consider the trick too "model specific", but I'd argue that a PPL like Pyro can automatically detect when the trick is applicable.
[1] https://arxiv.org/abs/1909.04852 (edited)


> @fehiepsi Make sense to me. This look like the way we propose a new subsample indices (considering it as an subsample-size discrete variable), i.e. one single clock auxilary variable for that plate-size variable.
> 
> Hmm... it seems to be different. It is not clear to me how to make refract/reflex decisions in parallel.
>
> I see your point now. Nice observation!

I think a nice interface to specify this would be to follow the pattern we used for parallel enumeration, where a user could specify an infer config like `infer={"enumerate": "sequential"}` versus `infer={"enumerate": "parallel"}`. Thinking about this in more detail, I think we could independently decide whether to parallelize over each of the enclosing plates. So e.g. if we parallelize by default we could specify something a set of enclosing plates outside of which this particular sample site has downstream dependencies (which prevents parallelism), something like `infer={"dependent_plates": ["plate1", "plate2"]}`.

fehiepsi:
@fritzo I think it is doable without much change in the implementation. Based on those "infer" information, we can maintain 3 behaviors for a discrete D-size variable `x`
+ D clock positions + D momentum variable, each one corresponding to each item in `x`: current implementation
+ 1 clock position + D momentum variables for `x`: this is what you suggested?
+ 1 clock position + 1 momentum variable for `x`: this is useful for a gibbs update or subsample update.

numpyro/infer/mixed_hmc.py
```diff
@@ -0,0 +1,211 @@
...
+from numpyro.infer.hmc_gibbs import DiscreteHMCGibbs, HMCGibbsState
+from numpyro.infer.hmc_util import euclidean_kinetic_energy, warmup_adapter
+from numpyro.util import cond, fori_loop, identity, ravel_pytree
+
+
+class MixedHMC(DiscreteHMCGibbs):
+    """
+    Implementation of Mixed Hamiltonian Monte Carlo (reference [1]).
+
+    .. note:: The number of discrete sites to update each time (`n_D` in reference [1])
```
martinjankowiak, line 19: nit: The number of discrete sites to update at each MCMC iteration (`n_D` in reference [1]) is fixed at 1

numpyro/infer/mixed_hmc.py
```diff
@@ -0,0 +1,211 @@
...
+        is fixed at value 1.
+
+    **References**
+
+    1. *Mixed Hamiltonian Monte Carlo for Mixed Discrete and Continuous Variables*,
+       Guangyao Zhou (2020)
+    2. *Peskun's theorem and a modified discrete-state Gibbs sampler*,
+       Liu, J. S. (1996)
+
+    :param inner_kernel: A :class:`~numpyro.infer.hmc.HMC` kernel.
```
martinjankowiak, line 29: some of these docstrings are missing defaults?

numpyro/infer/mixed_hmc.py
```diff
@@ -0,0 +1,211 @@
...
+        >>> mcmc.print_summary()
+        >>> samples = mcmc.get_samples()["x"]
+        >>> assert abs(jnp.mean(samples) - 1.3) < 0.1
+        >>> assert abs(jnp.var(samples) - 4.36) < 0.5
+    """
+
+    def __init__(self, inner_kernel, num_discrete_updates, *, random_walk=False, modified=True):
+        super().__init__(inner_kernel, random_walk=random_walk, modified=modified)
+        if inner_kernel._algo == "NUTS":
+            raise ValueError("The algorithm is designed to work with HMC and not to work with NUTS.")
```
martinjankowiak, line 67: nit: This algorithm only works with HMC and does not support NUTS.

numpyro/infer/mixed_hmc.py
```diff
@@ -0,0 +1,211 @@
...
+            rng_key, hmc_state, z_discrete, ke_discrete, arrival_times = vals
+            idx = jnp.argmin(arrival_times)
+            # NB: length of each sub-trajectory is scaled from the current min(arrival_times)
+            # (see the note at total_time below)
+            trajectory_length = arrival_times[idx] * time_unit
+            arrival_times = arrival_times - arrival_times[idx]
+            arrival_times = ops.index_update(arrival_times, idx, 1.)
+
+            # this is a trick, so that in a sub-trajectory of HMC, we always accept the new proposal
+            pe = jnp.inf if self.inner_kernel._algo == "HMC" else hmc_state.potential_energy
```
martinjankowiak, line 142: when does `self.inner_kernel._algo != "HMC"` ?

fehiepsi: I leave it to remind me that for NUTS, this trick is not required (and using `inf` in NUTS will give wrong results). I think we can relax the assertation `assert inner_kernel._algo == "HMC"` in the future, when some theory is verified.

numpyro/infer/mixed_hmc.py
```diff
@@ -0,0 +1,232 @@
...
+        num_discretes = self._support_sizes_flat.shape[0]
+        if self._discrete_mass is None:
+            # with total time is 1, we expect to update discrete vars num_discrete_steps times;
+            # there are num_discrete discrete vars, so each of them is expected to update
+            # num_discrete_steps / num_discretes times;
+            self._discrete_mass = jnp.full(num_discretes, num_discretes / self._num_trajectories)
+        # we expect to take 1 hmc step after 1 discrete step
+        hmc_mass = jnp.array(1.) / self._num_trajectories
+        self._clock_mass = jnp.concatenate([self._discrete_mass, hmc_mass[None]])
+        num_warmup = num_warmup * self._num_trajectories
```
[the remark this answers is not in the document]
fehiepsi: It is from an old commit, when adaptation is not ported to this class yet.

fehiepsi:
@martinjankowiak Thanks for reviewing! I have benchmarked the effect of `step_size <-> trajectory_length` and found that hmc is quite sensible to trajectory_length (in terms of n_eff). With the same model, the change can be better or worse depending on the initial trajectory_length. So I feel that the change is better in the sense that it keeps the trajectory_length unchanged (previously, with trajectory_length=7, step_size=2, then the actual trajectory_length is 6 with 3 steps - users can just simply set the initial length to 6 and still keep the dynamics unchanged).

About MixedHMC, currently the tests do not pass for models with only discrete latent variables and the results are biased regardless how high `num_samples` is. After taking a look at the paper again, we think that the final MH correction is not quite correct. Will write up something and send to you.

StannisZhou:
Models with only discrete latent variables are actually [where I started](https://arxiv.org/abs/1909.04852v1). In the Laplace momentum case the behavior is quite simple, similar to a systematic scan sampler but batching multiple site updates and accumulating energy differences to make the MH correction at a later time. But for these cases I didn't find clear advantages of the approach, unlike in the mixed case.

What are the models for which the tests are failing?

fehiepsi:
Hi @StannisZhou, as far as I see. It fails for models `Bernoulli(0.8)` or `Binomial(10, 0.3)`. The last MH correction targets a different distribution. In the particular case, where proposals are Gibbs update (i.e. random_walk=False, modified=False), the acceptance ratio is:
`p(x_new) / p(x) * [p(x_new) / p(x) * Q(x) / Q(x_new)] = p^2(x_new) / p^2(x) * Q(x) / Q(x_new)` where the formula inside `[...]` is 1, the acceptance ratio of a Gibbs update (`Q = p`).
So mixed hmc scheme is targeting p^2 distribution, not p. What do you think?

StannisZhou:
Screenshot from 2021-01-29 14-22-39
It sounds like you are using Gibbs updates? See this paragraph at the end of page 3 in https://arxiv.org/pdf/1909.04852.pdf I would always use (in your terminology) `modified=True`.

fehiepsi:
Hi @StannisZhou, I'm not sure why `modified=True` is important in case we just take 1 discrete step, but let's just consider that case. For binary distribution, the term Q(x_new)/Q(x) is always 1 in both random walk and Gibbs cases, so the final MH correction always accepts the proposal. This makes the scheme correct. But in cases where the size of the support is >=3, e.g. `Binomial(2, 0.8)` or `Binomial(10, 0.3)`, the final MH acceptance ratio Q(x_new)/Q(x) will not always be equal to 1 when we use modified Gibbs proposal (random walk proposal is fine).

Another issue is: in Bernoulli(0.8) case, if we take 2 discrete steps, then when we start at the discrete value 1, we will never reach the discrete value 0 because if we accept the proposal in the first step, then discrete kinetic energy will be increased by an amount that always allows us to accept the proposal in the second step. So the chain will stuck at value `1`.

StannisZhou:
Hi @fehiepsi , thanks for bringing up this very important issue. I still need to look a bit more into this, but a few points I know at the moment:

1. My memory was a bit fuzzy since It has been a while, but if you look at Algorithm 1 on page 3 of [my original the discrete only formulation](https://arxiv.org/pdf/1909.04852v1.pdf), you can see that there was no final MH correction. And it makes sense since In the discrete only case, the integration of the Hamiltonian dynamics is exact, and no correction is needed. This would resolve your failing tests, and all the points you raised (since mixed HMC reduces to a regular MH update in such cases). I revised my earlier comment to reflect this point.
2. The acceptance ratio you gave (`p(x_new) / p(x) * [p(x_new) / p(x) * Q(x) / Q(x_new)] = p^2(x_new) / p^2(x) * Q(x) / Q(x_new)`) does not seem to be consistent with what I have in the paper. In my terminology, the log acceptance ratio is `-(E-E_0)` (line 10 Algorithm 1). In the discrete only case you mentioned, `E=U + k_0 - \Delta E`, and `E_0 = U_0 + k_0`, and `\Delta E = U - U_0 + log Q(x|x_0) - log Q(x_0|x)` (line 19 of Algorithm 1), so the log acceptance ratio is `-[U - U_0 - (U - U_0 + log Q(x|x_0) - log Q(x_0|x))] = log Q(x|x_0) - log Q(x_0|x)`. In your terminology, the acceptance ratio should actually be `p(x_new) / p(x) * [p(x) / p(x_new) * Q(x_new) / Q(x)] = Q(x_new) / Q(x)`. For random walk proposals this ratio is always 1. So if you fix the acceptance ratio your tests shouldn't be failing for random walk proposals.
3. Although for random walk proposals it works, the remaining `log Q(x|x_0) - log Q(x_0 | x)` term still looking problematic, and would break non-random walk proposals. It should be naturally 0 for such cases. I'm still looking into this. I have some hypotheses, and I think there's a chance you actually helped me find the simpler formulation I've been trying (but failing) to get, although there might also be mess-ups in my current formulation. Will keep you posted.

fehiepsi:
>  the acceptance ratio should actually be p(x_new) / p(x) * [p(x) / p(x_new) * Q(x_new) / Q(x)] = Q(x_new) / Q(x). For random walk proposals this ratio is always 1. So if you fix the acceptance ratio your tests shouldn't be failing for random walk proposals

Agreed! The formula `p^2(x_new) ...` is used for not-modified Gibbs case, where Q(x_new)/Q(x) = p(x_new)/p(x). And as I mentioned in my last comment, the scheme is fine for binary distributions or higher support size distributions with modified/not-modified random walk proposals. It will likely cause issues for not-modified Gibbs proposal in general or for modified Gibbs proposal with support size >= 3.

About a simpler formulation, we mimiced the scheme in your paper but separating out the HMC halminton dynamics and discrete halmintonian dynamics. In other words, we consider the continuous variable as a discrete variable and HMC dynamics with MH proposal as a Gibbs proposal (i.e. Delta_E = 0). That way, we don't need the final MH correction and it allows us to the full NUTS update each time we want to update the continuous variable. We are still making the mass adaptation scheme (to control the speed of updates) to be more polished to be able to deliver it to users. Without mass adaptation, the scheme will be similar to random Gibbs scan as you mentioned.

StannisZhou:
Agreed about the periodicity case. I basically made a technical assumption when proving correctness.

Your comments led me to think there might be mess-ups in my cancellation reasonings in the proof which led to the seemingly incorrect final correction. I'm looking into this point.

Could you clarify the point about separating the two parts? If you are separating the two parts and making each Hamilton dynamics self-contained with its own MH correction, how is that different from just alternating between discrete and continuous updates in HMC-within-Gibbs?

fehiepsi:
Yes, that is similar to a sequence of Gibbs updates in MH-within-Gibbs, but updating (not reseting) the kinetic energy each time we update a variable, as in your paper. We just found that the scheme makes it easier to adjust the selection probability for which variable we want to update each time (e.g. some variables we want to update more often than the others). There are other ways to adjust the selection probabilty but the scheme in your paper is more intuitive to me.

In NUTS (and HMC with MH correction too), the accept ratio (after all internal MH corrections in HMC/NUTS) will always be 1, so having an auxiliary kinetic energy is not important. But it is good to use your scheme (with constant kinetic energy) to adjust the speed, hence we can adapt for how often we want to perform NUTS update in an MCMC step.

StannisZhou:
So you are indeed making an MH correction after each continuous HMC update. I don't think that's necessarily the best scheme in the HMC case, but for NUTS it does make more sense.

fehiepsi:
Updated: by incoporating the updated version by @StannisZhou, tests pass now. The new formula has the last acceptance ratio = 1 when the continuous HMC dynamic is exact - so it looks more reasonable than previously.

StannisZhou:
Thanks @fehiepsi ! I'm attaching the updated algorithm as reference while I'm working to fix the MH term in the paper.
Screenshot from 2021-02-07 17-18-41.

StannisZhou:
Update: I have fixed the arxiv version of the paper and the code on github. I additionally included a [script](https://github.com/StannisZhou/mixed_hmc/blob/master/scripts/simple_gmm/test_on_1d_gmm_with_jax.py) to showcase sampling with different proposal distributions.

A brief postmortem:

1. The mistake was in Lemma 4 in supplementary, and was introduced when I adapted the proof from discrete-only case to mixed discrete and continuous case. I have fixed the proof and included (previously omitted) detailed intermediate calculations.
2. The initial incorrect MH correction is the same as the correct one when we use (modified or not) random-walk proposals and for modified proposals on binary variables. In addition, for models where there are at most 2 nonzero entries in conditional distributions (e.g. the 1D GMM example I used, where components are well separated), previous MH correction with modified Gibbs is indistinguishable from the correct MH correction (previous delta_E = log(1 - p_new) - log(1 - p) \approx log(p) - log(p_new) = U_new - U = new delta_U). These happen to cover all my experiments and sanity checks (mostly on GMM and to a less degree on BLR) so I didn't spot the error from the numerical experiments.

martinjankowiak:
great to hear. thanks @StannisZhou for following up!

numpyro/infer/mixed_hmc.py
```diff
@@ -0,0 +1,220 @@
...
+
+    1. *Mixed Hamiltonian Monte Carlo for Mixed Discrete and Continuous Variables*,
+       Guangyao Zhou (2020)
+    2. *Peskun's theorem and a modified discrete-state Gibbs sampler*,
+       Liu, J. S. (1996)
+
+    :param inner_kernel: A :class:`~numpyro.infer.hmc.HMC` kernel.
+    :param int num_discrete_updates: Number of times to update discrete variables.
+        Defaults to the number of discrete latent variables.
+    :param bool random_walk: If False, Gibbs sampling will be used to draw a sample from the
```
martinjankowiak, line 36: nit: ... conditional `p(gibbs_site | remaining sites)`, where `gibbs_site` is one of the discrete sample sites in the model.

numpyro/infer/mixed_hmc.py
```diff
@@ -0,0 +1,220 @@
...
+        conditional `p(gibbs_site | remaining sites)`. Otherwise, a sample will be drawn uniformly
+        from the domain of `gibbs_site`. Defaults to False.
+    :param bool modified: whether to use a modified proposal, as suggested in reference [2], which
+        always proposes a new state for the current Gibbs site. Defaults to True.
```
martinjankowiak, line 40: nit: current Gibbs site (i.e. discrete site)

numpyro/infer/mixed_hmc.py
```diff
@@ -0,0 +1,220 @@
...
+        ...     c = numpyro.sample("c", dist.Categorical(probs))
+        ...     numpyro.sample("x", dist.Normal(locs[c], 0.5))
+        ...
+        >>> probs = jnp.array([0.15, 0.3, 0.3, 0.25])
+        >>> locs = jnp.array([-2, 0, 2, 4])
+        >>> kernel = MixedHMC(HMC(model, trajectory_length=1.2), num_discrete_updates=20)
+        >>> mcmc = MCMC(kernel, 1000, 100000, progress_bar=False)
+        >>> mcmc.run(random.PRNGKey(0), probs, locs)
+        >>> mcmc.print_summary()  # doctest: +SKIP
+        >>> samples = mcmc.get_samples()["x"]
```
martinjankowiak, line 64: make it clear both samples are available?

```
samples_x = mcmc.get_samples()["x"]
samples_c = mcmc.get_samples()["c"]
```

numpyro/infer/mixed_hmc.py
martinjankowiak: do you still want to keep this?

martinjankowiak (approved):

fehiepsi:
Thanks for reviewing and feedback, @martinjankowiak and @StannisZhou!

martinjankowiak (approved):

PR #831, opened 2020-12-02
Expose potential_fn_gen for composed kernels

fehiepsi:
Currently, if we use HMC/NUTS as an inner kernel, there is no way for the outer kernel to access the potential function that is created by the inner kernel. This PR exposes it through a hidden attribute. We might consider exposing it when the composing pattern becomes clearer.

### DIFF
--- a/numpyro/infer/sa.py
+++ b/numpyro/infer/sa.py
@@ -243,6 +243,7 @@ class SA(MCMCKernel):
         self._dense_mass = dense_mass
         self._init_strategy = init_strategy
         self._init_fn = None
+        self._potential_fn_gen = None
         self._postprocess_fn = None
         self._sample_fn = None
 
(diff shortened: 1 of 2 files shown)


numpyro/infer/sa.py
```diff
@@ -258,6 +259,7 @@ def _init_state(self, rng_key, model_args, model_kwargs, init_params):
             init_params = init_params[0]
             # NB: init args is different from HMC
             self._init_fn, sample_fn = _sa(potential_fn_gen=potential_fn)
+            self._potential_fn_gen = potential_fn
```
martinjankowiak, line 262: lgtm but why `_gen`?

martinjankowiak (approved):

numpyro/infer/sa.py
fehiepsi: This creates a potential_fn given inputs args/kwargs, hence the name. We distinguished `potential_fn` and `potential_fn_gen` in `hmc` implementation.

PR #832, opened 2020-12-02
Allow a callable to return an init value in param primitive

fehiepsi:
This is useful if users want to set random initial values for the parameters. This can be helpful for SteinVI implementation, where it requires different initial parameter values for each particle. (we can use `lift` to lift param sites to sample sites, then run the model multiple times to get those param values but it might be a bit verbose)

Actually, many tests/examples in Pyro use callables init_tensor to return random values. So it would be nice to be able to do so in NumPyro. In addition, it will be a bit cheaper/less memory requirement for SVI if users use the pattern: `lambda rng_key: some_value` because we only need to run the callable to get the initial values (in later updates, the values are substituted by params from optimizer)

cc @OlaRonning

### DIFF
--- a/numpyro/primitives.py
+++ b/numpyro/primitives.py
@@ -30,12 +30,6 @@ def apply_stack(msg):
             msg['value'], msg['intermediates'] = msg['fn'](*msg['args'],
                                                            sample_intermediates=True,
                                                            **msg['kwargs'])
-        elif msg['type'] == 'param':
-            value = msg['fn'](*msg['args'], **msg['kwargs'])
-            if callable(value):
-                msg['value'] = value(prng_key())
-            else:
-                msg['value'] = value
         else:
             msg['value'] = msg['fn'](*msg['args'], **msg['kwargs'])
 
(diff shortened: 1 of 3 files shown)


numpyro/primitives.py
```diff
@@ -33,6 +33,10 @@ def apply_stack(msg):
         else:
             msg['value'] = msg['fn'](*msg['args'], **msg['kwargs'])
 
+        if msg['type'] == 'param' and callable(msg['value']):
+            rng_key = prng_key()
```
neerajprad, line 37: What is `prng_key`?

fehiepsi: it is https://github.com/pyro-ppl/numpyro/pull/804. :D

neerajprad: Haha..I completely blanked out on the name and didn't see the definition here.

neerajprad: How about adding this after the `if` clause:
```python
if msg['type'] == 'sample':
  ...
elif msg['type'] == 'param':
  ...
else:
  ...
```

Actually, I'm thinking that if you modify the `fn` in the `init_msg` for `param` to store the initialization function and pass the rng_key to args, we may not even need to change anything here.

fehiepsi: Thanks, I just moved the logic to param primitive. It seems cleaner.

numpyro/primitives.py
```diff
@@ -151,13 +153,21 @@ def param(name, init_value=None, **kwargs):
...
     if not _PYRO_STACK:
+        assert not callable(init_value), \
+            "A callable init_value needs to be put inside a numpyro.handlers.seed handler."
         return init_value
 
+    if callable(init_value):
+        def fn(init_fn, *args, **kwargs):
+            return init_fn(prng_key())
+    else:
+        fn = identity
```
neerajprad, line 164: LGTM. For my own understanding, do we use kwargs to pass in constraints?

neerajprad (approved):

numpyro/primitives.py
fehiepsi: Yes, in svi we will use kwargs from param node in model trace to decide if we need to transform the value.

PR #833, opened 2020-12-11
Stein based inference

OlaRonning:
This PR introduces inference using the Stein method. Hugh shout-out to @ahmadsalim, who has done the majority of both design and implementation! 

This implementation does not use the`lift` handler for initializing the stein particles. I suggest introducing this in a separate PR once this version is merged as the PR is already quite large.

If it would help with reviewing, I can include an example using Stein.
The WIP [design doc](https://docs.google.com/document/d/1HzOyDSKwhzWOPUbfA_t35t3QSJvm1t_dHWAxq78blg8/edit?usp=sharing).

## TODOs
 - [x] Smoke test inference
 - [x] Update 2D toy example
 - [x] Update BNN example to use SVI run signature and comment
 - [x] Update DMM example to use SVI run signature and comment
 - [x] Factor VI out of EinSteinVI
 - [x] Factor out stein_point_mcmc
 - [x] Rewrite [check_model_guide_match](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/util.py#L511-L607) to allow reinit guides
 - [x] Correct already initialized warning in [init_to_uniform](https://github.com/pyro-ppl/numpyro/pull/833/files#diff-54efd1dc6b0e9987f90a4bd4768f517f229694843ddc52f902f3e51fb7be5e0aR77-R82) to allow reinit guides.


fehiepsi:
Hi @OlaRonning, could you add some tests/examples so we can see how to use the current API? Also, do you want me to double-check the logic somewhere or I can trust this implementation based on additional tests that you will add? Just let me know what I can do for your convenience.

One issue is if we release the next NumPyro version, it will take more iterations to change the API because we need to add deprecation messages. So hopefully we can come up with a nice API for users that we all agree on. :D 

I have some side questions: Is this fast / memory hungry? What is the recommended system to run SteinVI?

OlaRonning:
Hi @ fehiepsi,

I'll update with examples and start writing a unit-test sweep for the module. I believe double-checking the logic (as in checking correctness) would be a momentous task, which we should put off till the unit test has found the "simple" issues. If you have time, please check that we don't use the Jax constructs (e.g. [the vmaps](https://github.com/pyro-ppl/numpyro/blob/c2e6ae08c3de6e4ab4cddc44ce4c47a64e76660e/numpyro/infer/einstein/stein.py#L173-L176), etc.) in a suboptimal way. 

The high-level user API is
``` python
def model(*args, **kwargs):
	...

guide = AutoDelta(model)  # Any autoguide would do
stein = Stein(model, WrappedGuide(guide), Adam(0.001), ELBO(), RBFKernel())  # WrappedGuide needed for lifting
stein.init(rng_key)
for i in range(num_updates):
	stein.update(model_args, model_kwargs)
```
We can add a `stein.run` method similar to the one that exists for SVI.

The critical API consideration is whether to have the user explicitly wrap the guide? I feel this is the most flexible approach as it decouples the particles' global initialization strategy from the inference implementation. Techniques such as those introduced in [Stein Points](https://arxiv.org/pdf/1803.10161.pdf) would only consider the guide and not directly integrate with the Stein module.
```
stein = Stein(model, WrappedGuide(guide, init_strategy=init_with_noise(init_to_uniform())), ...) 
```

In terms of recommendations, I've been running it with GPU backend, and this is the preferred configuration, as Jax (in my experience) does not parallelize on CPU.  Anecdotally, the method is computationally intensive. I'll get some exact numbers on this once I've set-up profiling on my machine.

fehiepsi:
Thanks, @OlaRonning! I'll look at those vmap calls. For the API, would it be easier to add `init_strategy` to `Stein` constructor? That way, we have the same interface as SVI
```python
guide = AutoDelta(model)  # Any autoguide would do
stein = Stein(model, guide, Adam(0.001), ELBO(), RBFKernel(), init_strategy=...)
```
where under the hood, `Stein` will work with `WrappedGuide(guide, init_strategy=init_with_noise(init_to_uniform()))`. This way, WrappedGuide is just an internal helper and needs not to be exposed to users.

numpyro/infer/einstein/stein.py
```diff
@@ -0,0 +1,342 @@
...
+        if self.kernel_fn.mode == 'norm' or self.kernel_fn.mode == 'vector':
+            return kernel(x, y) * v
+        else:
+            return kernel(x, y) @ v
+
+    def _kernel_grad(self, kernel, x, y):
+        if self.kernel_fn.mode == 'norm':
+            return jax.grad(lambda x: kernel(x, y))(x)
+        elif self.kernel_fn.mode == 'vector':
+            return jax.vmap(lambda i: jax.grad(lambda x: kernel(x, y)[i])(x)[i])(jnp.arange(x.shape[0]))
```
fehiepsi, line 107: What is the shape of `x`, `y` and `kernel(x, y)` here? It seems to be that we can computing `kernel(x, y)` multiple times, across different `i`, which can be avoided by using `jacfwd`, `jacfwd`, or `vmap` directly over `x` I guess (I still need to understand the shapes first).

numpyro/infer/einstein/stein.py
```diff
@@ -0,0 +1,342 @@
...
+        else:
+            return jax.vmap(lambda l: jnp.sum(jax.vmap(lambda m: jax.grad(lambda x: kernel(x, y)[l, m])(x)[m])
+                                              (jnp.arange(x.shape[0]))))(jnp.arange(x.shape[0]))
```
fehiepsi, line 110: similar to the above: `kernel(x, y)` is computed repeatedly...

numpyro/infer/einstein/reinit_guide.py
```diff
@@ -0,0 +1,55 @@
...
+        def _find_valid_params(rng_key):
+            guide_trace = handlers.trace(handlers.seed(self.fn, rng_key)).get_trace(*args, **kwargs)
+            params = {name: site['value'] for name, site in guide_trace.items()
+                      if site['type'] == 'param' and self._reinit_hide_fn(site)}
+            for site in guide_trace.values():
+                if site['type'] == 'param' and not self._reinit_hide_fn(site):
+                    params[site['name']] = self.init_strategy(site, reinit_param=lambda _: True)
+            return params
+
+        init_params = jax.vmap(_find_valid_params)(rng_keys)
```
fehiepsi, line 44: I usually use `vmap` on `gpu` and use `lax.map` on `cpu` to reduce compiling time. But this is fine I think.

numpyro/infer/einstein/stein.py
ahmadsalim: The kernel_fn to compute kernel for pair of particles.
                 Modes: norm `(d,) (d,)-> ()`,  vector `(d,) (d,) -> (d)`, or matrix `(d,) (d,) -> (d,d)`

ahmadsalim: From SteinKernel 😄

ahmadsalim (commented):
Thanks for the review @fehiepsi .  I have added responses and will take a look at `init_strategy`.

fehiepsi:
@OlaRonning You can upload some data to https://github.com/pyro-ppl/datasets.

OlaRonning:
Hi @fehiepsi,

The PR is becoming ready to be reviewed. I will work on the design document over the next couple of days, so I don't think it makes sense to check the SteinVI part yet. However, I've updated
- Predictive to allow for [particle methods](https://github.com/pyro-ppl/numpyro/pull/833/files#diff-4a25871ef0f2ccc75b72b9048fc97fd9f084f56ec4852eb1db439f2286da8ed8R911-R920),
- [InitiatiInitiation](https://github.com/pyro-ppl/numpyro/blob/4597abf5da6d82e40c1d263a8d25c6f7cca07a98/numpyro/infer/initialization.py) to allow for reinit guides, and
-  batching for [raveling pytrees](https://github.com/pyro-ppl/numpyro/pull/833/files#diff-9ac3e2356dfd580188cf33cb996419c99efec180c6f519eb93e50877410db917R455-R462). 

These should be the changes that could affect the rest of the framework, so if you could look at them, that would be helpful. If you want to try out the examples, I recommend running them on a GPU.

fehiepsi:
Oops, I missed your message. Sorry, I'll take a pass later this week.

fehiepsi:
For my convenience, could you split this PR into smaller PRs? In particular, I don't quite understand the reasoning behind the first three points

> Predictive to allow for particle methods,

Why not using Predictive, then reshape the output? Or allowing `num_samples` taking a tuple of ints.

> InitiatiInitiation to allow for reinit guides

Why not using `numpyro.param(..., init_value=lambda rng: random.normal(rng, ...))`? Could we think of a solution that does not add new stuffs to the current api first?

> batching for raveling pytrees

Why not using `vmap(ravel_pytree)`?

docs/source/index.rst
```diff
@@ -56,6 +56,9 @@ NumPyro documentation
    tutorials/bayesian_imputation
    examples/gp
    examples/bnn
+   examples/stein_bnn
+   examples/stein_mixture_dmm
+   examples/stein_2d_toy
```
fehiepsi, line 61: How about moving those examples to a separate section? I just feel that `Applications` is a bit long for now and needs refactoring. :) In Pyro, we have [Other inference algorithms](https://pyro.ai/examples/baseball.html) section which we can mimic and use for SteinVI, nested sampling, hmcecs, MixedHMC, SA,...

examples/stein_2d_toy.py
```diff
@@ -0,0 +1,235 @@
+# Copyright Contributors to the Pyro project.
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+Example: 2d toy distribution with SteinVI
+===================================================
+In this example we infer a 2d star distribution with the different kernels available for SteinVI.
```
fehiepsi, line 7: I think we need to have a longer docstring to introduce SteinVI and its basic usage. By the way, could you also add a thumbnail for this example?

docs/source/index.rst
```diff
@@ -56,6 +56,9 @@ NumPyro documentation
    tutorials/bayesian_imputation
    examples/gp
    examples/bnn
+   examples/stein_bnn
```
fehiepsi, line 59: Each example is long so it is better to have separate PRs for them.

fehiepsi (commented):
Hi Ola, I took another pass but the PR is still large to review. I suggest separating some examples/utilities into smaller PRs.

examples/stein_2d_toy.py
```diff
@@ -0,0 +1,235 @@
...
+    stein_constr = partial(
+        SteinVI,
+        model=model,
+        guide=guide,
+        optim=Adagrad(step_size=0.1),
+        loss=Trace_ELBO(),
+        num_particles=args.num_particles,
+        init_strategy=init_with_noise(
+            init_to_value(values={"x_auto_loc": np.array([[0.0, 0.0]])}),
+            noise_scale=3.0,
```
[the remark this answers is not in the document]
OlaRonning: Sorry I did not update the examples when merging #1263. With it `init_with_noise` is obsolete if we make `noise_scale` a parameter for `SteinVI` that controls the find_params [scale](https://github.com/pyro-ppl/numpyro/blob/4326d4edd80a6791396a474ca25130dbe6b0a273/numpyro/contrib/einstein/steinvi.py#L144).

examples/stein_2d_toy.py
OlaRonning: That makes sense. I also started a [notebook version](https://github.com/aleatory-science/numpyro/blob/einstein_benchmark/notebooks/source/kernels.ipynb) of this example. Considering the intent of this example is usage, I think the notebook format is better. I'll open a separate PR with the notebook.

examples/stein_2d_toy.py
```diff
@@ -0,0 +1,235 @@
...
+        super(Star, self).__init__(
+            event_shape=(2,), batch_shape=batch_shape, validate_args=validate_args
+        )
+
+    def log_prob(self, value):
+        lps = []
+        for i in range(self.n_comp):
+            lps.append(
+                dist.MultivariateNormal(self.mus[i], self.covs[i]).log_prob(value)
+            )
```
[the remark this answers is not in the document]
OlaRonning: That makes sense. I'll clean this.

numpyro/infer/__init__.py
```diff
@@ -17,6 +17,7 @@
     init_to_sample,
     init_to_uniform,
     init_to_value,
+    init_with_noise,
```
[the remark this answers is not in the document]
OlaRonning: I thought I had already removed `init_with_noise`. I'll do so; it's not necessary.

test/contrib/einstein/test_einstein_kernels.py
```diff
@@ -71,7 +70,7 @@
         lambda d: {},
         lambda x: -0.02 / 12 * x[0] ** 4 - 0.5 / 12 * x[1] ** 4 - x[0] * x[1],
         {
-            "matrix": np.array(
+            "matrix": jnp.array(
```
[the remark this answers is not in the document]
OlaRonning: Reverted to use `np.array` for all cases.

fehiepsi (commented):
LGTM overall, pls let me know if you need me to take extra eyes at some points.

OlaRonning:
Thanks for the review. I'll update the code with your suggestions and change the examples to the latest version. If they run as expected, I'm comfortable merging.

numpyro/contrib/einstein/util.py
```diff
@@ -62,12 +62,17 @@ def get_parameter_transform(site):
...
     flat = vmap(lambda x: ravel_pytree(x)[0])(flat_pytree)
     unravel_fn = ravel_pytree(tree_map(lambda x: x[0], flat_pytree))[1]
-    return flat, lambda _flat: tree_multimap(
-        lambda x, shape: x.reshape(shape), vmap(unravel_fn)(_flat), shapes
+    return (
+        flat,
+        unravel_fn,
+        lambda _flat: tree_multimap(
+            lambda x, shape: x.reshape(shape), vmap(unravel_fn)(_flat), shapes
+        ),
```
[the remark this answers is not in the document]
OlaRonning: I'll add a docstring. I think I can avoid the third return.

test/contrib/einstein/test_stein.py
```diff
@@ -0,0 +1,443 @@
...
+
+
+KERNEL_TEST_CASES = [
+    TKERNEL(
+        RBFKernel,
+        lambda d: {},
+        lambda x: x,
+        {
+            "norm": 0.040711474,
+            "vector": jnp.array([0.056071877, 0.7260586]),
```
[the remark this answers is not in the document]
OlaRonning: Ups, I'll change to numpy arrays.

numpyro/primitives.py
```diff
@@ -424,9 +424,11 @@ class plate(Messenger):
     :param int dim: Optional argument to specify which dimension in the tensor
         is used as the plate dim. If `None` (default), the rightmost available dim
         is allocated.
+    :param float scale: Optional argument to scale subsample.
+        is allocated.
```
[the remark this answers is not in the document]
OlaRonning: I'll take it out and use `scale` in the example instead.

fehiepsi (approved):

fehiepsi:
Woohoo, thanks for working on this! This has been a while! :D

OlaRonning:
Yes, it has taken a while. If I recall, the initial estimate was a month. Thanks for all the help @fehiepsi! Your feedback has significantly improved the project C:

PR #835, opened 2020-12-12
Add capture-recapture models

### LINKED ISSUE #701: Tutorial/example on capture-recapture (CJS) models
This is a good exercise to get familiar with the enumeration mechanism, which is available in the last 0.3 release.

### References

+ [Pyro example](http://pyro.ai/examples/capture_recapture.html) which uses SVI. We can port those models to NumPyro and use MCMC for inference.
+ [Stan example](https://mc-stan.org/docs/2_23/stan-users-guide/mark-recapture-models.html) on CJS models. The reference will walk you go through the actual computations to marginalize discrete latent variables, in case you are interested in. Then you might appreciate how convenient Pyro enumeration is when it can handle all those complicated math for you.
+ Some [markov models](http://pyro.ai/numpyro/examples/hmm_enum.html) and [mixture models](http://pyro.ai/numpyro/examples/annotation.html) using enumeration.

fehiepsi:
Resolves #701. 

The translation is pretty straightforward.

However, there is an issue of `scan` does not compose well with the `unconstrained` handler, so I will move the fix for it to a separate PR.

Blocked by
- [x] https://github.com/pyro-ppl/funsor/pull/406.
- [x] #837


examples/cjs.py
```diff
@@ -0,0 +1,288 @@
...
+    def transition_fn(carry, y):
+        first_capture_mask, z = carry
+        with numpyro.plate("animals", N, dim=-1):
+            with handlers.mask(mask=first_capture_mask):
+                mu_z_t = first_capture_mask * phi * z + (1 - first_capture_mask)
+                # NumPyro exactly sums out the discrete states z_t.
+                z = numpyro.sample("z", dist.Bernoulli(dist.util.clamp_probs(mu_z_t)))
+                mu_y_t = rho * z
+                numpyro.sample("y", dist.Bernoulli(dist.util.clamp_probs(mu_y_t)), obs=y)
+
```
martinjankowiak, line 76: does this treat `t=0` differently? in the pyro code `first_capture_mask` is all zeros for the first time slice?

fehiepsi: It is the same. I just unroll the first step. Could you double-check the logic?

examples/cjs.py
```diff
@@ -0,0 +1,288 @@
...
+    phi_mean = numpyro.sample("phi_mean", dist.Uniform(0.0, 1.0))  # mean survival probability
+    phi_logit_mean = logit(phi_mean)
+    # controls temporal variability of survival probability
+    phi_sigma = numpyro.sample("phi_sigma", dist.Uniform(0.0, 10.0))
+    rho = numpyro.sample("rho", dist.Uniform(0.0, 1.0))  # recapture probability
+
+    def transition_fn(carry, y):
+        first_capture_mask, z = carry
+        with handlers.reparam(config={"phi_logit": LocScaleReparam(0)}):
+            phi_logit_t = numpyro.sample("phi_logit", dist.Normal(phi_logit_mean, phi_sigma))
```
fehiepsi, line 139: @martinjankowiak I think the only change in models is the use of reparam here. Without this, it is pretty slow because each trajectory took 1024 leapfrog steps... Other changes are equivalent to Pyro:
+ Pyro clamps `probs` by default. Without clamping, things like `Bernoulli(0).log_prob(1)` will lead to finite potential energy but `nan` gradients.
+ It is tricky to use `if else` logic in JAX, so I unrolled the first step.
+ for loop -> transition_fn + scan

numpyro/examples/datasets.py
```diff
@@ -36,6 +37,11 @@
 ])
 
 
+DIPPER_VOLE = dset('dipper_vole', [
+    'https://d2hg8soec8ck9v.cloudfront.net/datasets/dipper_vole.zip',
```
fehiepsi, line 41: @neerajprad Could you help me upload this file [dipper_vole.zip](https://github.com/pyro-ppl/numpyro/files/5684306/dipper_vole.zip) to cloudfront?

examples/cjs.py
martinjankowiak: maybe i don't understand how numpyro's `scan` works but to me it looks like the `t=0` iteration is missing.
how do you get an observation statement with `obs=capture_history[:, 0]`?

martinjankowiak: or is it just that that observation statement has zero `log_prob` anyway because of the mask?

fehiepsi: Yes, I moved the case t=0 out of the loop. At t=0, `mu_z_t=1` so `z=Bernoulli(mu_z_t)=1` (which is the same as the initial `z`) and `first_capture_mask = first_capture_mask | capture_history[:, 0] = capture_history[:, 0]`. Because the mask of first step is False, adding an obs statement is not necessary. We can add it but the code will be more populated:
```
first_capture_mask = False
with numpyro.plate("animals", N, dim=-1):
    with handlers.mask(mask=first_capture_mask):
        numpyro.sample("y", dist.Bernoulli(rho), obs=capture_history[:, 0])
first_capture_mask |= capture_history[:, 0].astype(bool)
```
comparing to 1 line of code
```
first_capture_mask = capture_history[:, 0].astype(bool)
```

numpyro/contrib/funsor/infer_util.py
```diff
@@ -141,18 +141,18 @@ def model(*args, **kwargs):
...
             dim_to_name = site["infer"]["dim_to_name"]
-            log_prob = funsor.to_funsor(log_prob, output=funsor.reals(), dim_to_name=dim_to_name)
+            log_prob_factor = funsor.to_funsor(log_prob, output=funsor.reals(), dim_to_name=dim_to_name)
 
             time_dim = None
             for dim, name in dim_to_name.items():
                 if name.startswith("_time"):
-                    time_dim = funsor.Variable(name, funsor.domains.bint(site["value"].shape[dim]))
-                    time_to_factors[time_dim].append(log_prob)
+                    time_dim = funsor.Variable(name, funsor.domains.bint(log_prob.shape[dim]))
```
fehiepsi, line 149: @eb8680 I think this is a bug so I fixed it here. This makes enum fails when we have sites with `event_shape != ()`. I already add tests for the current examples (in those examples, `log_det` terms have `event_shape = (0,)`). What do you think?

martinjankowiak (dismissed):
capture_recapture.py looks good to me! probably want @neerajprad to look at the internal changes?

fehiepsi:
Thanks for reviewing, @martinjankowiak! The internal changes will be clean up after #837. Those models run pretty well (though compiling time is slow due to this [jax issue](https://github.com/google/jax/issues/5162)) and give pretty good n_eff (n_eff ~ 1000 with 1000 samples). Hopefully, mixed gibbs stuff can achieve a similar performance. :D

numpyro/examples/datasets.py
fehiepsi: cc @neerajprad in case you missed to notification :D thanks!

neerajprad: We don't have access to cloudfront anymore, unfortunately. :( We should discuss and figure out where to house these datasets (we only have read access to the remaining ones).

neerajprad (dismissed):

numpyro/examples/datasets.py
neerajprad: @fehiepsi - I don't have an opinion on where we should put the dataset. My only concern is that it should have high availability for CI. We can start with whatever is convenient and if that works fine for some time, we can migrate the remaining datasets. Let me know if you want me to put this on my personal aws S3.

fritzo: Another easy option is raw.githubusercontent.com. We could move all datasets to https://github.com/pyro-ppl/datasets or something.

neerajprad: I created the pyro-ppl/datasets repo. @fehiepsi - I think you were also in favor of this. Do you want to commit your zip file or upload the three files separately? I think they are small enough, so it shouldn't matter either ways.

fehiepsi: Thanks, @neerajprad! I'll upload the zip file there and update the link here.

neerajprad (approved):

PR #836, opened 2020-12-13
Incorporate input_event_dim/output_event_dim to transforms

fehiepsi:
This feature is incorporated in PyTorch distributions recently. I found it useful for #835, where I need to know what is the ndim of a constrained sample, given an unconstrained sample.

I believe this change will cover more ComposedTransform transforms or TransformedDistribution distributions, but still not enough. We need something like [forward_event_shape](https://www.tensorflow.org/probability/api_docs/python/tfp/bijectors/Bijector#forward_event_shape) to provide correct batch_shape, event_shape for TransformedDistributions (currently, only len(batch_shape), len(event_shape) are correct)). But it can be deferred to the future if needed.

**Example** for a not supported case
```
from jax import random
d = dist.TransformedDistribution(dist.Normal(0, 1).expand([6]), dist.transforms.LowerCholeskyTransform())
d.event_shape, d.sample(random.PRNGKey(0)).shape
```
returns `(1, 6)` and `(3, 3)`

### DIFF
--- a/numpyro/distributions/distribution.py
+++ b/numpyro/distributions/distribution.py
@@ -36,7 +36,7 @@ from jax import lax, tree_util
 import jax.numpy as jnp
 
 from numpyro.distributions.constraints import is_dependent, real
-from numpyro.distributions.transforms import Transform
+from numpyro.distributions.transforms import ComposeTransform, Transform
 from numpyro.distributions.util import lazy_property, promote_shapes, sum_rightmost, validate_sample
 from numpyro.util import not_jax_tracer
 
@@ -725,9 +725,23 @@ class TransformedDistribution(Distribution):
         # to pay attention to any inference function that inspects
         # transformed distribution's shape.
         shape = base_distribution.batch_shape + base_distribution.event_shape
-        event_dim = max([len(base_distribution.event_shape)] + [t.event_dim for t in transforms])
-        batch_shape = shape[:len(shape) - event_dim]
-        event_shape = shape[len(shape) - event_dim:]
+        base_ndim = len(shape)
+        transform = ComposeTransform(self.transforms)
+        transform_input_event_dim = transform.input_event_dim
+        if base_ndim < transform_input_event_dim:
+            raise ValueError("Base distribution needs to have shape with size at least {}, but got {}."
+                             .format(transform_input_event_dim, base_ndim))
+        event_dim = transform.output_event_dim + max(self.base_dist.event_dim - transform_input_event_dim, 0)
+        # See the above note. Currently, there is no way to interpret the shape of output after
+        # transforming. To solve this issue, we need something like Bijector.forward_event_shape
+        # as in TFP. For now, we will prepend singleton dimensions to compromise, so that
+        # event_dim, len(batch_shape) are still correct.
+        if event_dim <= base_ndim:
+            batch_shape = shape[:base_ndim - event_dim]
+            event_shape = shape[base_ndim - event_dim:]
+        else:
+            event_shape = (1,) * (event_dim - base_ndim) + shape
+            batch_shape = ()
         super(TransformedDistribution, self).__init__(batch_shape, event_shape, validate_args=validate_args)
 
     @property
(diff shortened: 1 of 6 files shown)


numpyro/distributions/distribution.py
```diff
@@ -725,9 +725,23 @@ def __init__(self, base_distribution, transforms, validate_args=None):
...
+        event_dim = transform.output_event_dim + max(self.base_dist.event_dim - transform_input_event_dim, 0)
+        # See the above note. Currently, there is no way to interpret the shape of output after
+        # transforming. To solve this issue, we need something like Bijector.forward_event_shape
+        # as in TFP. For now, we will prepend singleton dimensions to compromise, so that
+        # event_dim, len(batch_shape) are still correct.
+        if event_dim <= base_ndim:
+            batch_shape = shape[:base_ndim - event_dim]
+            event_shape = shape[base_ndim - event_dim:]
+        else:
+            event_shape = (1,) * (event_dim - base_ndim) + shape
```
neerajprad, line 743: Is there a reason to prepend these singleton dims instead of suffixing these?

numpyro/distributions/transforms.py
```diff
@@ -147,7 +155,21 @@ def codomain(self):
...
     @property
     def event_dim(self):
-        return max(p.event_dim for p in self.parts)
+        raise ValueError("Please use `.input_event_dim` or `.output_event_dim` instead.")
+
+    @property
+    def input_event_dim(self):
+        input_event_dim = self.parts[-1].input_event_dim
+        for part in self.parts[len(self.parts) - 1::-1]:
+            input_event_dim = part.input_event_dim + max(input_event_dim - part.output_event_dim, 0)
```
neerajprad, line 164: Will this ever be anything other than `part[0].input_event_dim` in practice?

numpyro/distributions/distribution.py
fehiepsi: Nope, I think we can also add singleton dims to the last. Either way will not be correct, but we only need to maintain the correct event_dim here.

numpyro/distributions/transforms.py
fehiepsi: Yes, as in the tests, we have `transforms.AffineTransform(0, 2).event_dim = 0` but the composed one will have `input_event_dim = 1`. In general, `Transform.inv` should revert `input_event_dim` and `output_event_dim`, but it is clear that after composed, `output_event_dim != parts[-1].output_event_dim`. This preserves the symmetric.

numpyro/distributions/distribution.py
neerajprad: Hmm..I was thinking that the latter might be more accurate if we compare at each dimension, but it seems like both might be equally wrong. If for example, we have a LowerCholesky constraint on a univariate normal with `batch_shape=(3,)`, then actual event shape = `(2, 2` (assuming we include the diagonal), but this will give us `(1, 3)` which is completely wrong. Is having `shape` useful in other contexts?

numpyro/distributions/transforms.py
neerajprad: I see, that makes sense!

numpyro/distributions/distribution.py
fehiepsi: I found only three places using `.event_shape` or `.shape()`: in [funsor](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/contrib/funsor/enum_messenger.py#L529), [autoguide](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/infer/autoguide.py#L420), and [scan](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/contrib/control_flow/scan.py#L60). All of them only use `event_shape` to calculate `event_dim`. In my opinion, only `batch_shape` and `event_dim` are useful in inference algorithms. :)

neerajprad: I agree with that assessment. What I was trying to get at is should we make it more obvious that we don't know what the event shape is, by say having `(-1,) * event_dim` instead of having something that sort of looks correct but can be buggy? That will broadcast fine to a single dim and otherwise raise a loud error.

fehiepsi: Thanks, your suggestion looks good to me!

neerajprad (approved):

PR #837, opened 2020-12-13
Make scan compatible with unconstrain reparam

fehiepsi:
Currently, at each scan step, we transform and compute logdet of the whole unconstrained values. This both wastes computation resources because we only need to transform a particular value at the current time step and gives wrong results because logdet is counted multiple times. This PR addresses that issue and adds a regression test.

I also move `infer_config` out of `contrib` to avoid depending on `funsor`. In addition, it seems that this handler is also useful outside of funsor context. (note that we also have `poutine.infer_config` in Pyro).

### DIFF
--- a/numpyro/contrib/control_flow/scan.py
+++ b/numpyro/contrib/control_flow/scan.py
@@ -106,7 +106,7 @@ class promote_shapes(Messenger):
 
 
 def scan_enum(f, init, xs, length, reverse, rng_key=None, substitute_stack=None):
-    from numpyro.contrib.funsor import config_enumerate, enum, infer_config, markov
+    from numpyro.contrib.funsor import config_enumerate, enum, markov
     from numpyro.contrib.funsor import trace as packed_trace
 
     # XXX: This implementation only works for history size=1 but can be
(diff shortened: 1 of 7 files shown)


numpyro/contrib/control_flow/scan.py
```diff
@@ -29,14 +29,20 @@ def tree_flatten(self):
...
+        aux_trace, site_names = aux_data
         trace, = children
-        for name, site in trace.items():
-            site.update(aux_data[name])
-        return cls(trace)
+        trace_with_aux = {}
+        for name in site_names:
+            trace[name].update(aux_trace[name])
+            trace_with_aux[name] = trace[name]
+        return cls(trace_with_aux)
```
fehiepsi, line 45: I changed the logic here a bit to return a trace with sites having the same order as the original trace. No performance gain. This only makes it easier to debug (when I printed out the trace, the order of the sites was a bit surprising to me). :)

test/contrib/test_control_flow.py
```diff
@@ -78,3 +79,33 @@ def body_fn(z, val):
     with trace(), seed(rng_seed=0), substitute(data={"z": data}):
         zs = model()
     assert_allclose(zs, data)
+
```
neerajprad, line 82: We might need to run the time series notebook again to see if the results look reasonable (if anything, it should be better)?

neerajprad (approved):
LGTM. This seems like the cleanest way to handle this. The composability issues that this points to will always exist, but good to see that we have workarounds for most practical use cases.

test/contrib/test_control_flow.py
fehiepsi: That tutorial does not have time-dependent latent variables so it should be the same IMO.

fehiepsi:
Thanks for reviewing, @neerajprad! I'll merge this to clean up the capture-recapture PR. :)

neerajprad:
You may want to clean up the `# TODO: leverage t.input_event_dim, t.output_event_dim when they are available` in the other PR then. :)

PR #839, opened 2020-12-14
add HMCGibbs sampler

martinjankowiak:


numpyro/infer/hmc_gibbs.py
```diff
@@ -0,0 +1,108 @@
...
+class HMCGibbs(MCMCKernel):
+    """
+    HMC-within-Gibbs. This inference algorithm allows the user to combine
+    general purpose gradient-based inference (HMC or NUTS) with custom
+    Gibbs samplers.
+
+    :param inner_kernel: One of :class:`~numpyro.infer.HMC` or :class:`~numpyro.infer.NUTS`.
+    :param gibbs_fn: A Python callable that returns a dictionary of Gibbs samples conditioned
+        on the HMC sites. Must include an argument `rng_key` that should be used for all sampling.
+        Must also include arguments for all HMC and Gibbs sites.
```
fehiepsi, line 41: Hmm, I think a rng_key, one dict (or two separate dicts) of current gibbs + unconstrained samples. Assuming that we have many latent samples, requiring users to specify all of them in `gibbs_fn` definition might not be a good idea.

Generally, we would like to ask users to use the same signature here. Probably we can create an abstract Gibbs function class and ask users to inherit from it. Something like
```
class FooGibbsFn:
    def __init__(self, model, ...):
        # inspect support sizes,...

    def __call__(self, rng_key, hmc_sites, gibbs_sites, postprocess_fn=None):
        hmc_sites = postprocess_fn(hmc_sites)  # unconstrain -> constrain
        ...
        return new_gibbs_sites
```

Probably the bug that you mentioned is due to this missing of `uncontrained - > constrained` stuff here.

numpyro/infer/hmc_gibbs.py
```diff
@@ -0,0 +1,108 @@
...
+    :param gibbs_sites: a list of site names for the latent variables that are covered by the Gibbs sampler.
+    """
+
+    sample_field = "z"
+
+    def __init__(self, inner_kernel, gibbs_fn, gibbs_sites):
+        if not (isinstance(inner_kernel, HMC) or isinstance(inner_kernel, NUTS)):
```
fehiepsi, line 48: nit: NUTS is a subclass of HMC, so no need to check for it here.

numpyro/infer/hmc_gibbs.py
```diff
@@ -0,0 +1,108 @@
...
+        model_kwargs = {} if model_kwargs is None else model_kwargs.copy()
+        rng_key, key_u, key_z = random.split(rng_key, 3)
+        prototype_trace = trace(seed(self.model, key_u)).get_trace(*model_args, **model_kwargs)
+
+        gibbs_sites = {name: site["value"] for name, site in prototype_trace.items() if name in self._gibbs_sites}
+        model_kwargs["_gibbs_sites"] = gibbs_sites
+        hmc_state = self.inner_kernel.init(key_z, num_warmup, init_params, model_args, model_kwargs)
+
+        z = {**gibbs_sites, **hmc_state.z}
+        _, self._unravel_fn = ravel_pytree(gibbs_sites)
```
fehiepsi, line 83: nit: this line is not used anywhere

numpyro/infer/hmc_gibbs.py
```diff
@@ -0,0 +1,108 @@
...
+
+from jax import device_put, random, value_and_grad
+
+from numpyro.handlers import condition, seed, trace
+from numpyro.infer.mcmc import MCMCKernel
+from numpyro.infer.hmc import HMC, NUTS
+from numpyro.util import ravel_pytree
+
+
+HMCGibbs_State = namedtuple("HMCGibbs_State", "z, hmc_state, rng_key")
```
fehiepsi, line 16: nit: probably `HMCGibbsState` is more similar to `HMCState` but it depends on your preference

numpyro/infer/hmc_gibbs.py
```diff
@@ -0,0 +1,108 @@
...
+        with condition(data=gibbs_values):
+            model(*args, **kwargs)
+    return fn
+
+
+class HMCGibbs(MCMCKernel):
+    """
+    HMC-within-Gibbs. This inference algorithm allows the user to combine
+    general purpose gradient-based inference (HMC or NUTS) with custom
+    Gibbs samplers.
```
fehiepsi, line 36: nit: I guess this is already flexible enough for discrete and hmc_ecs cases. But can we mark the interface as EXPERIMENTAL? I would like to think how far we can compose this with the sequential proposal scheme in mixed hmc.

numpyro/infer/hmc_gibbs.py
```diff
@@ -0,0 +1,109 @@
...
+    :param gibbs_sites: a list of site names for the latent variables that are covered by the Gibbs sampler.
+    """
+
+    sample_field = "z"
+
+    def __init__(self, inner_kernel, gibbs_fn, gibbs_sites):
+        if not isinstance(inner_kernel, HMC):
+            raise ValueError("inner_kernel must be a HMC or NUTS sampler.")
+        if not callable(gibbs_fn):
+            raise ValueError("gibbs_fn must be a callable")
```
fehiepsi, line 52: We need to assert `inner_kernel.model is not None` here. It is trickier to support `potential_fn` input IMO.

test/test_hmc_gibbs.py
```diff
@@ -0,0 +1,146 @@
...
+import numpyro
+import numpyro.distributions as dist
+from numpyro.infer import HMC, MCMC, NUTS, HMCGibbs
+
+
+def _linear_regression_gibbs_fn(X, XX, XY, Y, rng_key, beta, log_sigma=None, sigma=None):
+    N, P = X.shape
+
+    if sigma is None:
+        sigma = jnp.exp(log_sigma)
```
fehiepsi, line 23: Just my guess to pass the test: you can change those two lines to
```
sigma = jnp.exp(log_sigma if sigma is None else sigma)
```

numpyro/infer/hmc_gibbs.py
```diff
@@ -0,0 +1,143 @@
...
+        Must also include arguments `hmc_sites` and `gibbs_sites`, each of which is a dictionary
+        with keys that are site names and values that are sample values. Note that a given `gibbs_fn`
+        may not need make use of all these sample values.
+    :param gibbs_sites: a list of site names for the latent variables that are covered by the Gibbs sampler.
+
+    **Example**
+
+    .. doctest::
+
+    >>> from jax import random
```
martinjankowiak, line 54: hope this is correct formatting!

docs/source/mcmc.rst
```diff
@@ -22,6 +22,12 @@ MCMC Kernels
     :show-inheritance:
     :member-order: bysource
 
+.. autoclass:: numpyro.infer.hmc.HMCGibbs
```
fehiepsi, line 25: typo: hmc -> hmc_gibbs

numpyro/infer/hmc_gibbs.py
fehiepsi: I think you will need to indent here (not so sure about that). You can test the docstring with `make doctest`.

numpyro/infer/hmc_gibbs.py
```diff
@@ -0,0 +1,143 @@
...
+    ...     x = numpyro.sample("x", dist.Normal(0.0, 2.0))
+    ...     y = numpyro.sample("y", dist.Normal(0.0, 2.0))
+    ...     numpyro.sample("obs", dist.Normal(x + y, 1.0), obs=jnp.array([1.0]))
+
+    >>> def gibbs_fn(rng_key, gibbs_sites, hmc_sites):
+    ...    y = hmc_sites['y']
+    ...    new_x = dist.Normal(0.8 * (1-y), math.sqrt(0.8)).sample(rng_key)
+    ...    return {'x': new_x}
+
+    >>> hmc_kernel = HMC(model)
```
fehiepsi, line 70: nit: it is better to use `NUTS` to mention that this **HMC**Gibbs class can work with NUTS sampler. :D

fehiepsi (approved):
LGTM. Thanks, @martinjankowiak!

fehiepsi (approved):

PR #842, opened 2020-12-15
Simplifies Stein utilities and adds tests

fehiepsi:
As title.

@OlaRonning Could you help me review this? For some reason, I couldn't make the tests pass with the `master` implementation (with `len(batch_shape)>=2`, it throws some vmap error, and with `batch_shape=()`, the `assert_allclose` failed). If I understand something wrong, please let me know.

Currently, the test for PrecondMatrixKernel raises assertation error
```
E       AssertionError: 
E       Not equal to tolerance rtol=1e-07, atol=0
E       
E       Mismatched elements: 2 / 4 (50%)
E       Max absolute difference: 128.7461
E       Max relative difference: 2.29563e+10
E        x: array([[128.7461,   0.    ],
E              [  0.    , 128.7461]], dtype=float32)
E        y: array([[5.608312e-09, 0.000000e+00],
E              [0.000000e+00, 9.347186e-05]], dtype=float32)
```

Probably there is something wrong with this PR. How did you come up with those numerical values, @OlaRonning? FYI, pyro gp kernels are [tested against GPy package](https://gist.github.com/fehiepsi/cf55756ea44dbf87bbb51ab331738a87).

I think we need to think about adding more tests for stein kernels. So we can be more confident that there is a bug on the user side, rather than in the implementation.


numpyro/infer/einstein/kernels.py
```diff
@@ -205,6 +205,7 @@ def mode(self):
         return self._mode
 
     def compute(self, particles, particle_info, loss_fn):
+        # FIXME(fehiepsi): is the usage of `numpy.random` here safe?
```
fehiepsi, line 208: @OlaRonning Does those `np.random` calls happen 1 time and not happen under jit?

OlaRonning: @fehiepsi good question, it should happen once even under jit. If I've understood correctly, the return from `npr.randn` should be cached in jaxpr as constants.

OlaRonning:
@fehiepsi, I'll have a look at the PR tomorrow. To compute the values in the test, I wrote a small script in NumPy; I have it on my workstation.
I agree a larger test sweep would be desirable, in particular, if we can compare the kernel computations with reference implementations in other frameworks.

numpyro/infer/einstein/kernels.py
fehiepsi: I see. Thanks, it seems that this is used to have the same random_weights/biases and the same choice of particles
```
particles = particles[npr.choice(particles.shape[0], self.bandwidth_subset)]
```
across vectorized/parallel maps (`vmap`, `pmap`,...). I am not sure about this design decision, especially for the particle subset above. That determination does not match the word "random" as I expected in dropout. Here if we run `jit(compute)` multiple times, I guess the same subset will be selected while if we run `compute` multiple times, different subsets will be selected each time. Could you suggest a note to add to the docs for this behavior?

OlaRonning:
Hi @fehiepsi,
So had a closer look, first of all, I'll update the third particle in `PARTICLES_2D` as it causes the hessian 
``` python
  lambda x: np.array([[12 * x[0] ** 2, 0.], [0., -3 * x[1]]])
```
to become singular. The Jax NumPy version of `inv` puts in `inf`, which gets caught by the `jnp.maximum(-jnp.inf, 1e-5)` and so the problem is silent. I'll update the tests and make a PR for your branch.

OlaRonning:
Found a problem with the `RBF` kernel, so will make the PR for NumPyro master.

fehiepsi:
Thanks for the help, really appreciate it @OlaRonning !

> the issue with the vmap, can you be more specific as to where this occurs

It happens for the test cases with `len(batch_shape) >= 2` in the previous version but not with the changes here. Reproducible code if you are interested in
```python
import jax
import jax.numpy as jnp
mlambda = jnp.ones((2, 3, 4))
mlambdasqrt = jax.vmap(lambda ml: jnp.diag(jnp.maximum(ml, 1e-5) ** 0.5),
                       in_axes=tuple(range(jnp.ndim(mlambda) - 1)))(mlambda)
```

numpyro/infer/einstein/utils.py
```diff
@@ -1,37 +1,41 @@
...
 def sqrth(m):
     """ Map a matrix to a positive definite matrix and square root it. """
     mlambda, mvec = jnp.linalg.eigh(m)
-    if jnp.ndim(mlambda) >= 2:
-        mlambdasqrt = jax.vmap(lambda ml: jnp.diag(jnp.maximum(ml, 1e-5) ** 0.5),
-                               in_axes=tuple(range(jnp.ndim(mlambda) - 1)))(mlambda)
-        msqrt = mvec @ mlambdasqrt @ jnp.swapaxes(mvec, -2, -1)
-    else:
-        mlambdasqrt = jnp.diag(jnp.maximum(mlambda, 1e-5) ** 0.5)
-        msqrt = mvec * mlambdasqrt * jnp.swapaxes(mvec, -2, -1)
```
fehiepsi, line 25: @OlaRonning `msqrt = mvec * mlambdasqrt * jnp.swapaxes(mvec, -2, -1)` should be `msqrt = (mvec * mlambdasqrt) @ jnp.swapaxes(mvec, -2, -1)`?

OlaRonning: Nice catch. Yes, that is correct!
``` python
>>> arr = np.array([[2,1],[1,5]])
>>> mlambda, mvec = np.linalg.eigh(arr)
>>> mvec * np.diag(mlambda) * mvec.T
array([[1.5547002 , 0.        ],
       [0.        , 4.85747583]])
>>> (mvec * mlambda) @ mvec.T
array([[2., 1.],
       [1., 5.]])
```
Won't update it as this PR fixes with cleaner code and [test](https://github.com/pyro-ppl/numpyro/pull/842/files/896cb36c16a127ff8c97cacebd0e00e43a92e3c7#diff-22c8c05fa7e59344b962bb707c5840651f9c21bd9404591e2e4c38545e58a9e1R97-R103).

fehiepsi:
Thanks for reviewing and helping me fix the test @OlaRonning! Looking like the remaining thing is to discuss how random works in the random feature kernel. Would like to hear more of your thought on that. I am a bit worried that the numpy.random usage makes things a bit diverge from jax random philosophy.

numpyro/infer/einstein/kernels.py
OlaRonning: Sorry I've gotten back regarding this. 

I re-read the [SVGD as Moment matching paper](https://arxiv.org/pdf/1810.11693.pdf) and in the Empirical Experiments section (Appendix D), the specify that the random parameters are selected once (at the beginning) and fixed over iteration. That this is consistent with the random feature representation (eq. 5) they use in their analysis is not entirely apparent to me.

The current naming is foremost faithful to the paper that introduces them. However, I appreciate the disconnect with the "common" semantic of an iterative random process. The kernels are written to use JIT compiling, and so a warning that triggers if the kernel is used without JIT compilation would suffice?

OlaRonning: Regarding the divergence from Jax explicitly parsing the state of the PRNG, I agree that using `numpy.random` is undesirable. With that in mind, I think it should be rewritten using `jax.random`, but it maintains it's current behavior.

fehiepsi: Thanks for explaining! It is clearer to me now. Unlike other kernels, the particle shapes here are fixed and unlike dropout, the particle subset is also fixed. Looking like we should add an assertation
```python
if self._particles_shape is None:
    self._particles_shape = particles.shape
else:
    assert particles.shape == self._particles_shape, \
        "Expect particles.shape={}, but got {}.".format(self._particles_shape, particles.shape)
```
and revise
```python
if self.bandwidth_subset is not None:
    particles = particles[npr.choice(particles.shape[0], self.bandwidth_subset)]
```
into something like
```python
if self.bandwidth_subset is not None:
    if self._bandwidth_subset_indices is not None:
        self._bandwidth_subset_indices = npr.choice(particles.shape[0], self.bandwidth_subset)
    particles = particles[self._bandwidth_subset_indices
```

This still suffers from the issue: parallel map steinvi with different seeds will use the same weight/bias/subset_indices. To match JAX behavior, I think we can add `rng_key` to the constructor and initialize those infos (weights, bias, subset indices) using that random key. This will require users to specify a random seed for this kernel. What do you think?

OlaRonning: That looks great! 

Would it be possible to get the `rng_key` using the `primitive` instead of adding it as a parameter? To keep the signature consistent.

fehiepsi: It seems reasonable but will require running this `compute` method under `seed` handler. Did you run this method under `seed` handler in `svi.init`? We used similar approach in `numpyro.module`:
```python
params = numpyro.param(...)
if params is None:
    rng = numpyro.prng_key()
    params = module.get_initial_params(rng)
```

OlaRonning: Currently, a kernel object is passed as a parameter when instantiating as in
``` python
stein = Stein(model, WrappedGuide(guide), Adam(1e-3), TraceELBO(),RandomFeatureKernel())
```
and not instantiated by `Stein` (as I remembered it). With that in mind, I think adding a rng_key to `__init__` for `RandomFeatureKernel` is a fine solution
``` python
stein = Stein(model, WrappedGuide(guide), Adam(1e-3), TraceELBO(),RandomFeatureKernel(rng_key))
```

fehiepsi: I just took a look at your Stein PR #833. It turns out that the kernel is used in `_svgd_loss_and_grads`, which is not run under the `.init()` method. It seems to be tricky to keep the signature consistent here. A solution is to follow SVI/stax/flax/haiku pattern, which is to have a `kernel.init` method that does the initialization job. In SVI or Stein, the `init` method is passed with `rng_key`, so we can use a random key there to initialize the kernel. Something likes
```python
class EinSteinVI(...):
    def init(self, rng_key, ...):
        rng_key, rng_kernel = random.split(rng_key)
        # no-op for all kernels except for `RandomFeatureKernel`
        # (this can also be a chance to precompute any constant terms in each kernel)
        self.kernel_fn.init(rng_key, particles_shape)
```
That way, we can keep the signature consistent. What do you think?

fehiepsi: I just pushed a commit with the above approach. My only concern is if `particles_shape` is easy to obtain in `Stein.init`. My impression is `Stein.init` and `Stein.update` have a similar signature, so this formation should be available in `init` method. But I might misunderstand something...

test/test_einstein_kernels.py
```diff
@@ -34,7 +38,7 @@
     T(RandomFeatureKernel,
       lambda d: {},
       lambda x: x,
-      {'norm': 12.190277}),
+      {'norm': 15.486456}),
```
fehiepsi, line 41: @OlaRonning Because the random pattern is changed (using jax.random instead of numpy.random), this test failed. So I changed the number here. I think testing against a random seed is not a good way but have no other ideas.

OlaRonning: LGTM

fehiepsi:
@OlaRonning Would you mind if I close this PR and move the current code `infer/einstein` in `master` branch to `contrib/einstein`? That way it will be easier for you to update the api (e.g. to address some issues in this PR) and contribute with less obligation to write careful tests. It will also easier for me (I'll mainly look into the big picture, rather than details) to review your Stein PRs. If we keep it in `infer`, after the release, you will need to add deprecation messages for any change in the API. What do you think?

OlaRonning:
That is fine with me. Must have missed some discussion. I'll have a look tomorrow morning.

fehiepsi:
Please take your time. I'm just trying to clean up tasks for releasing. We can address issues discussed so far after the release. :)

numpyro/infer/einstein/kernels.py
OlaRonning: The stein particles are initialized during `Stein.init`  [here](https://github.com/pyro-ppl/numpyro/blob/c2e6ae08c3de6e4ab4cddc44ce4c47a64e76660e/numpyro/infer/einstein/stein.py#L269-L271), so their shape will be available in `Stein.init`.

fehiepsi: Hi @OlaRonning, could you take over this PR and make a separate PR for this? I think you understand how to make the change consistent much better than me. It is also faster for us to move Stein inference forward. :)

OlaRonning: Yes absolutely. Just for clarity, you suggest making a separate PR for `kernel.init` (also updating the `Stein.init`), factor your current `RandomFeatureKernel.init` out of this PR, and merging the remainder?

fehiepsi: Yup, but you might not need to have several PRs for this: just a single one similar to this but being more completed is enough (e.g. I'm not sure if `def init(self, rng_key, particles_shape):` is a good signature for the init method).

OlaRonning: I see, thanks for clarifying. I'll code and see how it works out :)

OlaRonning:
@fehiepsi I've created a branch on aleatory-science from your fork. I'll do the refactor there, and then we can do the git-foo to turn it into PR for Numpyro.

fehiepsi:
Thank you, @OlaRonning! Let me close this PR. :)

PR #843, opened 2020-12-16
Fixed bandwidth for RBF kernel and test cases

OlaRonning:
This PR fixes the test cases to be consistent with the assumptions for Stein kernels. The Precondition Kernel is now tested for the constant case as it is simpler to verify the computation manually.

@fehiepsi checked with #842 and this now works as expected.

### DIFF
--- a/test/test_einstein_kernels.py
+++ b/test/test_einstein_kernels.py
@@ -15,9 +15,10 @@ from numpyro.infer.einstein.kernels import (
     PrecondMatrixKernel
 )
 
+jnp.set_printoptions(precision=100)
 T = namedtuple('TestSteinKernel', ['kernel', 'particle_info', 'loss_fn', 'kval'])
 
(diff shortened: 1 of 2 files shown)


test/test_einstein_kernels.py
```diff
@@ -79,5 +80,6 @@ def test_kernel_forward(kernel, particles, particle_info, loss_fn, tparticles, m
     d, = tparticles[0].shape
     kernel_fn = kernel(mode=mode).compute(particles, particle_info(d), loss_fn)
     value = kernel_fn(*tparticles)
+    print(value)
```
fehiepsi, line 83: nit: is this print intended?

fehiepsi (dismissed):
Thanks for the fix, @OlaRonning! LGTM.

test/test_einstein_kernels.py
OlaRonning: Ups, leftover from checking with #842.

fehiepsi (approved):

PR #844, opened 2020-12-16
Add has_rsample / rsample attribute to distributions

fehiepsi:
@neerajprad Do we need to test for those attributes?

### DIFF
--- a/docs/source/distributions.rst
+++ b/docs/source/distributions.rst
@@ -49,6 +49,14 @@ TransformedDistribution
     :show-inheritance:
     :member-order: bysource
 
+Delta
+-----
+.. autoclass:: numpyro.distributions.distribution.Delta
+    :members:
+    :undoc-members:
+    :show-inheritance:
+    :member-order: bysource
+
 Unit
 ----
 .. autoclass:: numpyro.distributions.distribution.Unit
@@ -328,14 +336,6 @@ CategoricalLogits
 CategoricalProbs
 ----------------
 .. autoclass:: numpyro.distributions.discrete.CategoricalProbs
-    :members:
-    :undoc-members:
-    :show-inheritance:
-    :member-order: bysource
-
-Delta
------
-.. autoclass:: numpyro.distributions.discrete.Delta
     :members:
     :undoc-members:
     :show-inheritance:
--- a/numpyro/distributions/distribution.py
+++ b/numpyro/distributions/distribution.py
@@ -33,7 +33,6 @@ import warnings
 import numpy as np
 
 from jax import lax, tree_util
-from jax.dtypes import canonicalize_dtype
 import jax.numpy as jnp
 
 from numpyro.distributions.constraints import is_dependent, real
@@ -863,14 +862,14 @@ class Delta(Distribution):
         batch_dim = jnp.ndim(v) - event_dim
         batch_shape = jnp.shape(v)[:batch_dim]
         event_shape = jnp.shape(v)[batch_dim:]
-        self.v = lax.convert_element_type(v, canonicalize_dtype(jnp.float64))
+        self.v = v
         # NB: following Pyro implementation, log_density should be broadcasted to batch_shape
         self.log_density = promote_shapes(log_density, shape=batch_shape)[0]
         super(Delta, self).__init__(batch_shape, event_shape, validate_args=validate_args)
 
     def sample(self, key, sample_shape=()):
         shape = sample_shape + self.batch_shape + self.event_shape
-        return jnp.broadcast_to(device_put(self.v), shape)
+        return jnp.broadcast_to(self.v, shape)
 
     @validate_sample
     def log_prob(self, value):
--- a/numpyro/distributions/distribution.py
+++ b/numpyro/distributions/distribution.py
@@ -418,14 +418,14 @@ class ExpandedDistribution(Distribution):
         return samples.reshape(sample_shape + self.batch_shape + self.event_shape)
 
     def rsample(self, key, sample_shape=()):
-        return self._sample(self.base_dist.rsample, sample_shape)
+        return self._sample(self.base_dist.rsample, key, sample_shape)
 
     @property
     def support(self):
         return self.base_dist.support
 
     def sample(self, key, sample_shape=()):
-        return self._sample(self.base_dist.sample, sample_shape)
+        return self._sample(self.base_dist.sample, key, sample_shape)
 
     def log_prob(self, value):
         shape = lax.broadcast_shapes(self.batch_shape,
--- a/numpyro/distributions/distribution.py
+++ b/numpyro/distributions/distribution.py
@@ -195,7 +195,7 @@ class Distribution(metaclass=DistributionMeta):
 
     @property
     def has_rsample(self):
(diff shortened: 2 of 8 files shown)


fritzo (commented):
In Pyro we define roughly
```py
def __call__(self, sample_shape=()):
    if self.has_rsample:
        return self.rsample(sample_shape)
    else:
        return self.sample(sample_shape)
```
It might make sense to follow this convention so that SVI implementations agree? Note then you'd need to change a couple of the `.sample()` implementations in this PR to defer to `base_dist.sample` rather than `base_dist.__call__()`.

eb8680:
> It might make sense to follow this convention so that SVI implementations agree?

We could also consider removing `Distribution.__call__` in Pyro - it's another awkward API relic from the early days.

numpyro/distributions/distribution.py
```diff
@@ -193,6 +193,16 @@ def event_dim(self):
...
         return len(self.event_shape)
 
+    @property
+    def has_rsample(self):
+        return not self.is_discrete
+
+    def rsample(self, key, sample_shape=()):
+        if self.has_rsample:
+            self.sample(key, sample_shape=sample_shape)
+        else:
```
neerajprad, line 203: nit: no `else` needed here.

numpyro/distributions/distribution.py
```diff
@@ -193,6 +193,16 @@ def event_dim(self):
         """
         return len(self.event_shape)
 
+    @property
+    def has_rsample(self):
+        return not self.is_discrete
```
neerajprad, line 198: We already have a more granular `reparametrized_params` available. I'm not sure if all our continuous distributions are reparametrized? If so, a safer choice might be `return set(self.reparametrized_params) == set(self.arg_constraints.keys())`.

neerajprad:
> @neerajprad Do we need to test for those attributes?

I suppose it will be nice to add some small tests that check - consistency between `reparametrized_params` and `rsample`, and whether we are correctly assigning this for compound distributions like MaskedDistribution/TransformedDistribution, but it is up to your judgment. :) We are doing gradient testing for rsample anyways which is the more important part.

numpyro/distributions/distribution.py
fehiepsi: Makes sense to me. I'll make the change. One thing is reparameterized notion is a bit ambiguous. If I understand correctly, TFP folks implement reparameterized sampler for gamma, then the paper implies dirichlet or studentt using gamma sampler are also reparameterized. But it is a bit different from @fritzo and @martinjankowiak work, where any derived sampler needs its own reparam sampler.

fritzo: fyi: I would say that also in Martin's and my work, we get downstream samplers like Dirichlet and StudentT for free. However those "free" samplers used by tfp involve latent variables (e.g. the gamma in a student t sampler), and therefore introduce variance. What Martin and I did was to derive a custom reparametrization for Dirichlet that is deterministic and therefore has lower variance than the "free" version that tfp uses. But in both cases it is safe to set `.has_rsample = True`.

neerajprad: @fehiepsi - I wasn't sure about distributions like `LKJCholesky`, for instance.

fehiepsi: @fritzo Thanks! I was hesitating to set `has_rsample=True` for Dirichlet/StudentT because I couldn't make them pass the pathwise tests. But it seems that the "free" version is not expected to pass those tests because they have higher variance. I'll update the PR to set `has_rsample=True` for those distributions.

@neerajprad Looking like we can set `rsample` for it because it is a transformed distribution of a Beta distribution. :D

fehiepsi:
@neerajprad Thanks for your suggestion! The additional tests catch a couple of bugs. =.= Here, I also exposed more "reparameterized" params, which are tested in `test_sample_gradient` except for parameters inherit from gamma samplers.

numpyro/distributions/distribution.py
```diff
@@ -193,6 +193,16 @@ def event_dim(self):
         """
         return len(self.event_shape)
 
+    @property
+    def has_rsample(self):
+        return set(self.reparametrized_params) == set(self.arg_constraints)
```
fritzo, line 198: **[low priority]** I like @neerajprad's suggestion here. I suppose we could be even more precise; I don't know the JAX syntax, but in PyTorch we could write:
```py
return all(name in self.reparametrized_params
           for name in self.arg_constraints
           if getattr(self, name).requires_grad)
```

docs/source/distributions.rst
```diff
@@ -49,6 +49,14 @@ TransformedDistribution
     :show-inheritance:
     :member-order: bysource
 
+Delta
```
fehiepsi, line 52: I moved Delta out of Discrete because its behavior is more special.

numpyro/distributions/distribution.py
fehiepsi: It is trickier here because, in jax, the gradients are not stored in the variables: the function `grad(f)` can be applied for any ndarray.

test/test_distributions.py
```diff
@@ -99,7 +99,7 @@ def sample(self, key, sample_shape=()):
 }
 
 CONTINUOUS = [
-    T(dist.Beta, 1., 2.),
+    T(dist.Beta, 0.2, 1.1),
```
fehiepsi, line 102: nothing special, I just changed this for pathwise testing but forgot to reset the values =.=

fehiepsi:
I have no idea why on CI, discrete distributions have `rsample`... I can't replicate it in my system.

test/test_distributions.py
```diff
@@ -338,20 +370,32 @@ def test_unit(batch_shape):
...
+    gamma_derived_params = {
+        "Gamma": ["concentration"],
+        "Beta": ["concentration1", "concentration0"],
+        "Chi2": ["df"],
+        "InverseGamma": ["concentration"],
+        "LKJ": ["concentration"],
+        "LKJCholesky": ["concentration"],
+        "StudentT": ["df"]
+    }.get(jax_dist.__name__, [])
+    reparameterized_params = [p for p in jax_dist.reparametrized_params
```
neerajprad, line 383: Is there a reason to not test these?

fehiepsi: I think this test is not designed for pathwise gradient:
```python
import torch
import pyro
import pyro.distributions as dist

def f(x):
    pyro.set_rng_seed(0)
    return dist.Gamma(x, 1).rsample()

x = torch.tensor(1., requires_grad=True)
f(x).backward()
eps = 1e-3
numerical = (f(x + eps) - f(x - eps)) / (2 * eps)
print(x.grad, numerical.detach())
```
print `tensor(1.9066) tensor(1422.4575)`.

fehiepsi: We have a separate test `test_pathwise_gradient` below, but it only works for `gamma` and `chi2`. Other distributions have high variance.

neerajprad: Thanks for explaining, @fehiepsi! LGTM.

neerajprad (approved):

PR #848, opened 2020-12-19
Support scan markov with history > 1

### LINKED ISSUE #702: Support markov with history size greater than 1
Currently, we supported `markov(history_size=1)` using `scan`. As suggested by @eb8680, we can use [sarkka_bilmes_product](https://github.com/pyro-ppl/funsor/blob/d139ed18ed19ca70fe2c2b2ab2595ffefe317655/funsor/sum_product.py#L288) for `history_size > 1`.

fehiepsi:
Resolves #702.

Tasks:
- [x] Revise log_density implementation to incorporate this feature.
- [x] Add hmm example with history > 1
- [x] Add test to verify the consistency of density, output carry, ys with unrolled scan

### DIFF
--- a/numpyro/contrib/control_flow/scan.py
+++ b/numpyro/contrib/control_flow/scan.py
@@ -87,25 +87,35 @@ def _subs_wrapper(subs_map, i, length, site):
                                " which is currently not supported. Please report the issue to us!")
 
 
-class promote_shapes(Messenger):
-    # a helper messenger to promote shapes of `fn` and `value`
-    #   + msg: fn.batch_shape = (2, 3), value.shape = (3,) + fn.event_shape
-    #     process_message(msg): promote value so that value.shape = (1, 3) + fn.event_shape
+class _promote_fn_shapes(Messenger):
+    # a helper messenger to promote shapes of `fn`
     #   + msg: fn.batch_shape = (3,), value.shape = (2, 3) + fn.event_shape
     #     process_message(msg): promote fn so that fn.batch_shape = (1, 3).
-    def process_message(self, msg):
+    def postprocess_message(self, msg):
         if msg["type"] == "sample" and msg["value"] is not None:
             fn, value = msg["fn"], msg["value"]
             value_batch_ndims = jnp.ndim(value) - fn.event_dim
             fn_batch_ndim = len(fn.batch_shape)
-            prepend_shapes = (1,) * abs(fn_batch_ndim - value_batch_ndims)
-            if fn_batch_ndim > value_batch_ndims:
-                msg["value"] = jnp.reshape(value, prepend_shapes + jnp.shape(value))
-            elif fn_batch_ndim < value_batch_ndims:
+            if fn_batch_ndim < value_batch_ndims:
+                prepend_shapes = (1,) * (value_batch_ndims - fn_batch_ndim)
                 msg["fn"] = tree_map(lambda x: jnp.reshape(x, prepend_shapes + jnp.shape(x)), fn)
 
 
-def scan_enum(f, init, xs, length, reverse, rng_key=None, substitute_stack=None, history=1):
+def _promote_scanned_value_shapes(value, fn):
+    # a helper function to promote shapes of `value`
+    #   + msg: fn.batch_shape = (T, 2, 3), value.shape = (T, 3,) + fn.event_shape
+    #     process_message(msg): promote value so that value.shape = (T, 1, 3) + fn.event_shape
+    value_batch_ndims = jnp.ndim(value) - fn.event_dim
+    fn_batch_ndim = len(fn.batch_shape)
+    if fn_batch_ndim > value_batch_ndims:
+        prepend_shapes = (1,) * (fn_batch_ndim - value_batch_ndims)
+        return jnp.reshape(value, jnp.shape(value)[:1] + prepend_shapes + jnp.shape(value)[1:])
+    else:
+        return value
+
+
+def scan_enum(f, init, xs, length, reverse, rng_key=None, substitute_stack=None, history=1,
+              first_available_dim=None):
     from numpyro.contrib.funsor import config_enumerate, enum, markov
     from numpyro.contrib.funsor import trace as packed_trace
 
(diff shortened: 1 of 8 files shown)


fehiepsi:
@eb8680 Could you help me review this PR? Most of the changes are generalized from history=1 -> history>1. There is a complicated issue to return the correct shape for the last carry: assume history=2, then we run step 0, step 1 outside of `lax.scan`, then use `lax.scan` for steps `2 -> (T - 1)`. Under enumeration, we have 3 different shapes for `carry`, but we can only interpret those shapes using the input carry and output carry at step 2. So one of them is missing.

We can run step 0, step 1, step 2, then using `lax.scan` for steps `3 -> (T - 1)`. This way we can interpret all carry shapes (using infos at steps 2 and 3) but need to unroll `history * 2` steps (in general). So unless it is really needed, I would like to unroll the number of steps as small as possible. :) Anyway, we already added a note previously that: `there should not have any site outside of scan depend on the first output of scan (the last carry value)`

numpyro/contrib/control_flow/scan.py
```diff
@@ -152,24 +149,33 @@ def body_fn(wrapped_carry, x, prefix=None):
...
-    with markov():
+    with markov(history=history):
         wrapped_carry = (0, rng_key, init)
-        wrapped_carry, (_, y0) = body_fn(wrapped_carry, x0)
-        if length == 1:
-            ys = tree_map(lambda x: jnp.expand_dims(x, 0), y0)
-            return wrapped_carry, (PytreeTrace({}), ys)
-        wrapped_carry, (pytree_trace, ys) = lax.scan(body_fn, wrapped_carry, xs_, length - 1, reverse)
+        y0s = []
+        for i in range(history):
```
eb8680, line 167: For clarification, after changing this to `for i in range(2 * history)` what other changes would be necessary to get the correct final `carry_shape`?

numpyro/contrib/control_flow/scan.py
```diff
@@ -187,24 +193,34 @@ def body_fn(wrapped_carry, x, prefix=None):
...
+    ys = tree_multimap(lambda z0, z: jnp.reshape(z, z.shape[:1] + jnp.shape(z0)[1:]), y0s, ys)
+    # then join with y0s
+    ys = tree_multimap(lambda z0, z: jnp.concatenate([z0, z], axis=0), y0s, ys)
     # we also need to reshape `carry` to match sequential behavior
-    if length % 2 == 0:
+    i = (length - 1) % (history + 1)
+    # XXX: unless we unroll more steps, we only know the correct shapes starting from
+    # the `history`-th iteration; that means we only know carry shapes when
+    # i == history - 1 or i == history, which corresponds to input carry or output carry
+    # at the `history`-th iteration.
```
eb8680, line 204: @fehiepsi I think I understand this issue, thanks for explaining.  How much compilation overhead does unrolling `2 * history` steps add in the HMM examples?  Unrolling `history` steps in exchange for incorrect output `carry` shapes is a very subtle optimization that only produces constant-time savings in program length, and may cause maintenance difficulties down the line, so it's important to be sure that it's worth it.

fehiepsi: Thinking about it again, I guess it won't take much compiling time because history is usually small (1 or 2). Let me address the issue in this PR.

eb8680 (approved):
LGTM after latest changes.

fehiepsi:
Thanks for reviewing, @eb8680! Looking like there is a bug when I unrolled more steps. Looking into it...

numpyro/contrib/control_flow/scan.py
```diff
@@ -87,45 +87,53 @@ def _subs_wrapper(subs_map, i, length, site):
...
-class promote_shapes(Messenger):
-    # a helper messenger to promote shapes of `fn` and `value`
-    #   + msg: fn.batch_shape = (2, 3), value.shape = (3,) + fn.event_shape
-    #     process_message(msg): promote value so that value.shape = (1, 3) + fn.event_shape
+class _promote_fn_shapes(Messenger):
+    # a helper messenger to promote shapes of `fn`
     #   + msg: fn.batch_shape = (3,), value.shape = (2, 3) + fn.event_shape
     #     process_message(msg): promote fn so that fn.batch_shape = (1, 3).
-    def process_message(self, msg):
+    def postprocess_message(self, msg):
```
fehiepsi, line 94: we only want to do this right before getting the trace.

numpyro/contrib/control_flow/scan.py
```diff
@@ -141,35 +149,54 @@ def body_fn(wrapped_carry, x, prefix=None):
...
 
         if init:
-            with handlers.scope(prefix="_init"):
-                new_carry, y = seeded_fn(carry, x)
+            # handler the name to match the pattern of sakkar_bilmes product
+            with handlers.scope(prefix='P' * (unroll_steps - i), divider='_'):
+                new_carry, y = config_enumerate(seeded_fn)(carry, x)
                 trace = {}
         else:
-            with handlers.block(), packed_trace() as trace, promote_shapes(), enum(), markov():
```
fehiepsi, line 148: The usage of `enum()`, `markov()` here still gives correct log density but does not recycle the dimensions at this step. To resolve this, I have moved `block` right before the generator `markov(unroll_steps + 1, history)` and remove `markov` handler here.

numpyro/contrib/control_flow/scan.py
```diff
@@ -87,45 +87,53 @@ def _subs_wrapper(subs_map, i, length, site):
...
     from numpyro.contrib.funsor import trace as packed_trace
 
-    # XXX: This implementation only works for history size=1 but can be
-    # extended to history size > 1 by running `f` `history_size` times
-    # for initialization. However, `sequential_sum_product` does not
-    # support history size > 1, so we skip supporting it here.
-    # Note that `funsor.sum_product.sarkka_bilmes_product` does support history > 1.
+    # amount number of steps to unroll
+    history = min(history, length)
+    unroll_steps = min(2 * history - 1, length)
```
fehiepsi, line 124: This is the minimum number of unroll_steps to get a correct last carry shape. But it is safe to use any number greater than this number.

fehiepsi:
@eb8680 Though the previous implementation gives correct log density, the carry shapes are not recycled with
```python
with markov(history=history):
    for i in range(unroll_steps):
        transition_fn()
    lax.scan(with markov(): transition_fn())
```
logic. I have revised it to be
```python
for i in markov(range(unroll_steps + 1)):
    if i < unroll_steps:
        transition_fn()
    else:
        lax.scan(transition_fn())
```
Previously, this issue is not detected due to a typo in the test, where I computed `expected` and `actual` values on the same model:
```python
    expected_x_prev, expected_x_curr = enum(config_enumerate(model))()
    actual_x_prev, actual_x_curr = enum(config_enumerate(model))()  # typo: should be fun_model
```

eb8680 (approved):
LGTM. Now that `sarkka_bilmes_product` is actually being used, I guess we should probably try to move away from the `P` prefix naming scheme upstream in `sarkka_bilmes_product` to a safer interface so no non-Markov names from user code with `P` in them get incorrectly mangled.  Maybe the simplest thing to do would be to make the prefixes longer (e.g. `_PREV_` instead of `P`).

fehiepsi:
> no non-Markov names from user code with P in them get incorrectly mangled

Good point! `_PREV_` should work well here.

PR #849, opened 2020-12-20
Expose logits in FooProbs distributions

fehiepsi:
As title. This is to be closer to Pyro API and makes things easier to debug.


fritzo (approved):

fehiepsi:
Thanks for reviewing!

Issue #769, opened 2020-10-01
Example for spike-and-slab regression


fehiepsi:
Per [a question](https://forum.pyro.ai/t/regression-variable-selection/2236) asked by a user in the forum, it would be nice to have a tutorial/example for this type of regression. I searched for some examples available and found some nice ones below.

+ https://github.com/AsaCooperStickland/Spike_And_Slab with Gibbs sampling
+ http://www.batisengul.co.uk/post/spike-and-slab-bayesian-linear-regression-with-variable-selection/ with Metropolis-Hastings within Gibbs sampling

None of those examples marginalize discrete latent variables so this would be a good example for the enumeration mechanism in NumPyro. I would expect with marginalization, the result will be better than the methods in the above ones.

For ones who are interested in this issue, please ping me with any question you have.

xidulu:
This example looks pretty interesting to me.

I believe the *model* could implemented like this?

```python
def model(X, y):
    N, d = X.shape
    pi = numpyro.param('pi', np.ones((d,)) * 0.5, constraint=unit_interval)
    xi = numpyro.sample('xi', dist.Bernoulli(probs=pi), infer={'enumerate': 'parallel'})
    w = numpyro.sample('w', dist.Normal(0, xi))
    return numpyro.sample('obs', dist.Normal(np.dot(X, w)), obs=y)
```

Haven't figured out how to write the inference part, is `Enum_elbo` available in Numpyro now?

fehiepsi:
Hi @xidulu, I think you can use MCMC for this model. We haven't supported enum for SVI yet. Regarding the model, I think it does not make sense to use `xi` for scale of Normal distribution (e.g. scale can't take 0 values). Your model looks a bit different from [spike and slab regression](https://arxiv.org/pdf/1812.07259.pdf). I guess there is a typo or something?

xidulu:
Hi @fehiepsi 
My knowledge on spike and slab prior mostly comes from : Eq 11 in https://arxiv.org/pdf/1810.04045.pdf

<img width="546" alt="Screen Shot 2020-10-10 at 3 40 37 AM" src="https://user-images.githubusercontent.com/26022201/95625016-5fa06f80-0aaa-11eb-8b69-92a2aadf8141.png">
<img width="428" alt="Screen Shot 2020-10-10 at 3 42 44 AM" src="https://user-images.githubusercontent.com/26022201/95625174-ac844600-0aaa-11eb-985b-00a3f08d84b0.png">

[...]

PR #850, opened 2020-12-21
Add discrete and subsample gibbs_fn

fehiepsi:
This PR adds a discrete Gibbs fn for using HMCGibbs over discrete variables and a "naive" subsample Gibbs fn for doing subsampling HMC.

TODO:
- [x] implement discrete gibbs fn
- [x] implement subsample gibbs fn
- [x] add tests
- (optional) add a spike-and-slab example: #769 (defer to later PR or user contribution)


numpyro/infer/hmc_gibbs.py
```diff
@@ -149,3 +153,178 @@ def potential_fn(z_gibbs, z_hmc):
...
+        >>> import numpyro.distributions as dist
+        >>> from numpyro.infer import MCMC, NUTS, HMCGibbs, discrete_gibbs_fn
+        ...
+        >>> def model(probs, locs):
+        ...     c = numpyro.sample("c", dist.Categorical(probs))
+        ...     numpyro.sample("x", dist.Normal(locs[c], 0.5))
+        ...
+        >>> probs = jnp.array([0.15, 0.3, 0.3, 0.25])
+        >>> locs = jnp.array([-2, 0, 2, 4])
+        >>> gibbs_fn = discrete_gibbs_fn(model, probs, locs)
```
martinjankowiak, line 210: this interface seems a bit confusing since this isn't really a gibbs step. it's a mh step that apply its own mh correction; furthermore there's no complete conditional here.... maybe we need to rename some of these things?

numpyro/infer/hmc_gibbs.py
```diff
@@ -149,3 +153,178 @@ def potential_fn(z_gibbs, z_hmc):
...
+    prototype_trace = trace(seed(model, rng_seed=0)).get_trace(*model_args, **model_kwargs)
+    support_sizes = {
+        name: jnp.broadcast_to(site["fn"].enumerate_support(False).shape[0], jnp.shape(site["value"]))
+        for name, site in prototype_trace.items()
+        if site["type"] == "sample" and site["fn"].has_enumerate_support and not site["is_observed"]
+    }
+
+    def gibbs_fn(rng_key, gibbs_sites, hmc_sites):
+        # convert to unconstrained values
+        z_hmc = {k: biject_to(prototype_trace[k]["fn"].support).inv(v)
```
martinjankowiak, line 230: given our interface i guess this leads to a wasteful flow unconstrained => constrained => unconstrained?

numpyro/infer/hmc_gibbs.py
```diff
@@ -149,3 +153,178 @@ def potential_fn(z_gibbs, z_hmc):
...
+        # get support_sizes of gibbs_sites
+        support_sizes_flat, _ = ravel_pytree({k: support_sizes[k] for k in gibbs_sites})
+        num_discretes = support_sizes_flat.shape[0]
+
+        rng_key, rng_permute = random.split(rng_key)
+        idxs = random.permutation(rng_key, jnp.arange(num_discretes))
+
+        def body_fn(i, val):
+            idx, support_size = idxs[i], support_sizes_flat[i]
+            return _discrete_step(*val, potential_fn=potential_fn, idx=idx, support_size=support_size)
```
martinjankowiak, line 251: this does many MH corrections instead of one. is that preferred?

numpyro/infer/hmc_gibbs.py
```diff
@@ -149,3 +153,178 @@ def potential_fn(z_gibbs, z_hmc):
...
+        ...     numpyro.sample("x", dist.Normal(locs[c], 0.5))
+        ...
+        >>> probs = jnp.array([0.15, 0.3, 0.3, 0.25])
+        >>> locs = jnp.array([-2, 0, 2, 4])
+        >>> gibbs_fn = discrete_gibbs_fn(model, probs, locs)
+        >>> kernel = HMCGibbs(NUTS(model), gibbs_fn, gibbs_sites=["c"])
+        >>> mcmc = MCMC(kernel, 1000, 100000, progress_bar=False)
+        >>> mcmc.run(random.PRNGKey(0), probs, locs)
+        >>> samples = mcmc.get_samples()["x"]
+        >>> assert abs(jnp.mean(samples) - 1.3) < 0.1
```
martinjankowiak, line 215: nit: if you're using this as a test add the variance? i believe it's 
`assert abs(jnp.var(samples) - 4.36) < 0.5`

numpyro/infer/hmc_gibbs.py
```diff
@@ -149,3 +153,178 @@ def potential_fn(z_gibbs, z_hmc):
...
+
+
+def subsample_gibbs_fn(model, *model_args, **model_kwargs):
+    """
+    [EXPERIMENTAL INTERFACE]
+
+    Returns a gibbs_fn to be used in :class:`HMCGibbs`, which works for subsampling
+    statements using :class:`~numpyro.plate` primitive. This implements the Algorithm 1
+    of reference [1] but uses a naive estimation (without control variates) of log likelihood,
+    hence might incur a high bias (reference [2]).
```
martinjankowiak, line 267: bias-> variance?
isn't this asymptotically exact?

martinjankowiak (approved):

numpyro/infer/hmc_gibbs.py
fehiepsi: Yes, we also waste some other computations such as:
+ the initial potential_energy - this information iin HMCGibbs is not accessible through the signatures of `gibbs_fn`.
+ the last potential_energe - we recompute it in HMCGibbs

numpyro/infer/hmc_gibbs.py
fehiepsi: This is more like a sequence of MH updates in a Gibbs scheme (var1 -> var2 -> var3 -> ...). I think I should change the implementation to the Metropolized/Modified Gibbs sampler in reference [1], that is
+ propose a discrete sample proportional to its conditional distribution (this requires evaluating log_density for all possible values in the support)
+ modify accept prob to `(1 - p_old) / (1 - p_new)`
and rename the current implementation to discrete_metropolis_fn. What do you think?

numpyro/infer/hmc_gibbs.py
fehiepsi: I am not sure if this is asymptotically exact. [This reference](https://arxiv.org/abs/1404.4178) has a proof for convergences when control variates are used.

numpyro/infer/hmc_gibbs.py
fehiepsi: Reading literatures through the day, I think it is better to use Gibbs update, instead of metropolized version here. The motivation of metropolized sampler is to try to move away from the current position. But assuming that the discrete variable is Bernoulli with prob=0.5, using metropolized version will always give acceptance probably = 1, hence the sampled sequence will be something like 0, 1, 0, 1, 0, 1,... which is periodic. I think I should use Gibbs sampler from now on, even in mixed hmc implementation. (for mixed hmc, when calculating Delta_E, we should try to remove the involvement of Q...)

numpyro/infer/hmc_gibbs.py
fehiepsi: will switch to gibbs sampling here.

numpyro/infer/hmc_gibbs.py
fehiepsi: Thanks, let me update the implementation and add this test.

numpyro/infer/hmc_gibbs.py
fehiepsi: @martinjankowiak I just updated the implementation of `_discrete_step` that uses a Gibbs update. Could you help me verify the math there?

numpyro/infer/hmc_gibbs.py
```diff
@@ -149,3 +154,188 @@ def potential_fn(z_gibbs, z_hmc):
...
+    rng_key, z, pe, _ = fori_loop(0, support_size - 1, body_fn, init_val)
+    return rng_key, z, pe
+
+
+def discrete_gibbs_fn(model, *model_args, **model_kwargs):
+    """
+    [EXPERIMENTAL INTERFACE]
+
+    Returns a gibbs_fn to be used in :class:`HMCGibbs`, which works for discrete latent sites
+    with enumerate support. The order of updated index is permuted at each step.
```
martinjankowiak, line 199: nit: "The site update order is randomly permuted at each step."

numpyro/infer/hmc_gibbs.py
```diff
@@ -149,3 +154,188 @@ def potential_fn(z_gibbs, z_hmc):
...
+
+    # XXX: we can't vmap potential_fn over all proposals and sample from the conditional
+    # categorical distribution because support_size is a traced value, i.e. its value
+    # might change across different discrete variables;
+    # so here we will loop over all proposals and use an online scheme to sample from
+    # the conditional categorical distribution
+    def body_fn(i, val):
+        rng_key, z, pe, weight_logsumexp = val
+        rng_key, rng_accept = random.split(rng_key)
+        proposal = jnp.where(i >= z_discrete_flat[idx], i + 1, i)
```
martinjankowiak, line 174: i don't understand this where is the randomness coming from?

numpyro/infer/hmc_gibbs.py
```diff
@@ -149,3 +154,188 @@ def potential_fn(z_gibbs, z_hmc):
...
+        >>> assert abs(jnp.mean(samples).copy() - 1.) < 0.1
+
+    """
+    prototype_trace = trace(seed(model, rng_seed=0)).get_trace(*model_args, **model_kwargs)
+    plate_sizes = {
+        name: site["args"]
+        for name, site in prototype_trace.items()
+        if site["type"] == "plate" and site["args"][0] > site["args"][1]  # i.e. size > subsample_size
+    }
+    enum = any(site["type"] == "sample"
```
martinjankowiak, line 318: this is used in a few places. perhaps move to a helper in infer/util?

fehiepsi: I think I will update `discrete_gibbs_fn` to not use this logic. Currently, `support_sizes` already holds all enumerable sites. We don't need to recompute this information in `discrete_gibbs_fn.gibbs_fn`

numpyro/infer/hmc_gibbs.py
fehiepsi: Based on the weights, we will decide if we will accept the proposal or not (`random.bernoulli(rng_accept, transition_prob)` below). For example, assume we only have 1 variable, support_size=4 and the probs are, 0.3, 0.1, 0.2, 0.4; idx=0: Then starting at 0, we will decide if we move to 1 based on the transition_prob 0.1/(0.3 + 0.1) = 0.25. Then we will decide if we move to 2 based on the transition prob `0.2 / (0.2 + 0.25) = 4/9`. Finally, we will decide if we move to 3 based on the transition prob `0.4 / (0.4 + 4.9)`. Is it reasonable?

martinjankowiak: doesn't this mean you have a 75% chance of staying at 0?

fehiepsi: Yes, that is for first step. If it stays at 0, then for the second step, you will have 5/9 chance to stay at 0.

fehiepsi: Opps, chance to stay at 0 for the second step is 2/3. I just edited the numbers.

fehiepsi: So chance to stay at 0 until the end is: 3/4 * 2/3 * 3/5 = 0.3

martinjankowiak: 👍

martinjankowiak (approved):

PR #854, opened 2020-12-30
Add various discrete gibbs functions method

fehiepsi:
### Motivation

+ PyMC3 supports [2 methods](https://docs.pymc.io/api/inference.html#pymc3.step_methods.metropolis.CategoricalGibbsMetropolis): `random_walk=True or False` (corresponding to uniform proposal and proportional proposal there) with `modified=True` (i.e. modified Gibbs)
+ In Mixed-HMC, [various proposals](https://github.com/StannisZhou/mixed_hmc/blob/master/momentum/hmc/mixed_hmc_jax.py#L166) are used: besides the above 2 methods, the original Gibbs is also used (i.e. `random_walk=False, modified=False`).

To benchmark against those frameworks, I added options to switch between them.

### API changes

+ new keyword `random_walk` and `modified` are introduced
+ model_args and model_kwargs are arguments of `discrete_gibbs_fn` and `subsample_gibbs_fn`. This allows us more flexibility to add options for future enhancements (e.g. there are various subsample methods) without having to introduce a new `foo_gibbs_fn`.

### Implementation changes

+ The hidden method `_discrete_foo_proposal` is revised to return only the proposal and its corresponding `log_accept_ratio`, and not perform MH acceptance/rejection. The reason is in a sequential scheme of mixed-hmc, that accept_ratio information is needed to adjust the current kinetic energy and I would like to reuse those functions instead of reimplementing them in mixed hmc.

### DIFF
--- a/numpyro/infer/hmc_gibbs.py
+++ b/numpyro/infer/hmc_gibbs.py
@@ -156,12 +156,13 @@ class HMCGibbs(MCMCKernel):
         return HMCGibbsState(z, hmc_state, rng_key)
 
 
-def _discrete_step(rng_key, z_discrete, pe, potential_fn, idx, support_size):
+def _discrete_gibbs_step(rng_key, z_discrete, pe, potential_fn, idx, support_size):
     # idx: current index of `z_discrete_flat` to update
     # support_size: support size of z_discrete at the index idx
 
     # get z proposal
     z_discrete_flat, unravel_fn = ravel_pytree(z_discrete)
+    pe_0 = pe
 
     # XXX: we can't vmap potential_fn over all proposals and sample from the conditional
     # categorical distribution because support_size is a traced value, i.e. its value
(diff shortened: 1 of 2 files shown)


numpyro/infer/hmc_gibbs.py
```diff
@@ -156,42 +156,106 @@ def potential_fn(z_gibbs, z_hmc):
...
-        rng_key, z, pe, weight_logsumexp = val
-        rng_key, rng_accept = random.split(rng_key)
+        rng_key, z, pe, log_weight_sum = val
+        rng_key, rng_transition = random.split(rng_key)
         proposal = jnp.where(i >= z_discrete_flat[idx], i + 1, i)
         z_new_flat = ops.index_update(z_discrete_flat, idx, proposal)
         z_new = unravel_fn(z_new_flat)
         pe_new = potential_fn(z_new)
-        weight_new = pe - pe_new
+        log_weight_new = pe_0 - pe_new
```
fehiepsi, line 179: This is a bug in the previous implementation. We should use the same constant `pe_0` during those transitions.

numpyro/infer/hmc_gibbs.py
```diff
@@ -156,42 +156,106 @@ def potential_fn(z_gibbs, z_hmc):
...
 
     init_val = (rng_key, z_discrete, pe, jnp.array(0.))
-    rng_key, z, pe, _ = fori_loop(0, support_size - 1, body_fn, init_val)
-    return rng_key, z, pe
+    rng_key, z_new, pe_new, _ = fori_loop(0, support_size - 1, body_fn, init_val)
+    log_accept_ratio = jnp.array(0.)
+    return rng_key, z_new, pe_new, log_accept_ratio
+
 
+def _discrete_modified_gibbs_propose(rng_key, z_discrete, pe, potential_fn, idx, support_size, stay_prob=0.):
```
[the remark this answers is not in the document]
martinjankowiak: does `stay_prob` need a `ValueError` check?

numpyro/infer/hmc_gibbs.py
```diff
@@ -201,6 +265,24 @@ def discrete_gibbs_fn(model, *model_args, **model_kwargs):
...
+    :param callable model: a callable with NumPyro primitives. This should be the same model
+        as the one used in the `inner_kernel` of :class:`HMCGibbs`.
+    :param tuple model_args: Arguments provided to the model.
+    :param dict model_kwargs: Keyword arguments provided to the model.
+    :param bool random_walk: If False, Gibbs sampling will be used to draw a sample from the
+        conditional `p(gibbs_site | remaining sites)`. Otherwise, a sample will be drawn uniformly
+        from the domain of `gibbs_site`.
+    :param bool modified: whether to use a modified proposal, as suggested in reference [1], which
+        always proposes a new state for the current Gibbs site.
+        The modified scheme appears in literature under the name "modified Gibbs sampler" or
```
martinjankowiak, line 277: nit: "in the literature"

martinjankowiak (dismissed):

numpyro/infer/hmc_gibbs.py
```diff
@@ -156,42 +156,106 @@ def potential_fn(z_gibbs, z_hmc):
...
     # get z proposal
     z_discrete_flat, unravel_fn = ravel_pytree(z_discrete)
+    pe_0 = pe
 
     # XXX: we can't vmap potential_fn over all proposals and sample from the conditional
     # categorical distribution because support_size is a traced value, i.e. its value
     # might change across different discrete variables;
     # so here we will loop over all proposals and use an online scheme to sample from
     # the conditional categorical distribution
     def body_fn(i, val):
```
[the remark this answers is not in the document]
fehiepsi: yeah, nice catch! I'll move it out

numpyro/infer/hmc_gibbs.py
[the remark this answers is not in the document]
fehiepsi: Thanks, I just added an assertion for it.

numpyro/infer/hmc_gibbs.py
```diff
@@ -156,42 +156,93 @@ def potential_fn(z_gibbs, z_hmc):
...
-        transition_prob = expit(weight_new - weight_logsumexp)
-        z, pe = cond(random.bernoulli(rng_accept, transition_prob),
-                     (z_new, pe_new), identity,
-                     (z, pe), identity)
-        weight_logsumexp = jnp.logaddexp(weight_new, weight_logsumexp)
-        return rng_key, z, pe, weight_logsumexp
-
+    body_fn = partial(_discrete_gibbs_proposal_body_fn,
+                      z_discrete_flat, unravel_fn, pe, potential_fn, idx)
     init_val = (rng_key, z_discrete, pe, jnp.array(0.))
```
martinjankowiak, line 191: i'm a bit confused about why an initial log weight of `jnp.array(0.)` gives you the behavior you want

martinjankowiak (approved):

numpyro/infer/hmc_gibbs.py
fehiepsi: This is Gibbs sampling, so we want to draw samples proportional to `e^-pe_i` (i.e. `log_weight_i = pe_init - pe_i` - the term `pe_init` is just a constant to scale down `pe_i` to improve precision) of each value `i` in the support. When `i = z_init`, `log_weight_i = pe_init - pe_init = 0.`.

martinjankowiak: ok that makes sense. i guess it would ideally be `jnp.max(log_weights)` but that's unavailable because of the online scheme?

fehiepsi: Yes, you are right.

Issue #539, opened 2020-02-19
GPU Memory


ross-h1:
GPU Memory 

Hi guys, have been having some problems with GPU Memory allocation, especially with fori_collect. I’ve read the JAX note on this here: 

https://jax.readthedocs.io/en/latest/gpu_memory_allocation.html 

- which has helped a bit, though does slow things down. Without using XLA_PYTHON_CLIENT_ALLOCATOR="platform", JAX seems very reluctant to give back memory after each batch of samples. I’m working on sampling in small batches and trying to transfer samples back to CPU memory after use. See code below for a version of fori_collect which batches sampling, in case of use to anyone…

The particular problem I have is with a large number of variables (circa 900) and dense_mass=True. HMCAdaptState which is stored for each sample then becomes very large as the number of samples increases (despite holding the same values post warmup). Is it possible to store HMCAdaptState as an attribute of the sampler, rather than the sample? Alternatively some kind of transfer between GPU to CPU memory and/or file backend post generation of each sample?

Thanks again for an awesome project! Ross

Here is a batched fori_collect, with samples stored as Numpy arrays after generation. This uses the Nvidia-ml-py3 library to access GPU memory usage:

(pip3 install nvidia-ml-py3)

Setup:

Import os

os.environ["XLA_PYTHON_CLIENT_ALLOCATOR"]="platform"

GPU_mem_state=None
try:
    from pynvml import *
    nvmlInit()
    GPUhandle=nvmlDeviceGetHandleByIndex(0)
    numpyro.set_platform('gpu')
    def GPU_mem_state():
        info = nvmlDeviceGetMemoryInfo(GPUhandle)
        return 'Used GPU Memory MB {:,}'.format(onp.round(info.used/1000000,2))
except:
    print ('Cant initialise GPU, Using CPU’)

Code:

def merge_flat(trace,samples):
    if trace is None:
        trace = [onp.atleast_1d(onp.asarray(f)) for f in samples]
    else:
        trace = [onp.concatenate([trace[k], onp.atleast_1d(onp.asarray(samples[k]))]) for k in range(len(trace))]
    return trace

def collect_max(start_collect, num_iter, sample_kernel, last_state, max_n=25):
    """
    As per numpyro fori_collect, but runs in batches of max_n size
    and returns flattened tree back on the cpu via merge_flat.
    """

    n=0
    last_val = last_state
    trace = None
    n_main, n_extra = np.divmod(num_iter, max_n)
[...]

PR #861, opened 2021-01-05
Expose some mcmc states for sequential sampling strategy

### LINKED ISSUE #534: Sequential Sampling Strategy

fehiepsi:
Resolves #534. This is also requested by @rexdouglass in https://github.com/pyro-ppl/numpyro/issues/539 

This is just a solution, mainly for discussion. I'm not sure if this is a good API so if reviewers have other ideas, please let me know.

### DIFF
--- a/numpyro/infer/mcmc.py
+++ b/numpyro/infer/mcmc.py
@@ -386,19 +386,17 @@ class MCMC(object):
     @property
     def init_state(self):
         """
-        The initial state to run MCMC. If this attribute is not None,
-        :meth:`run` will skip calling `self.sampler.init(...)` method to
-        obtain the initial state.
+        The initial state of the MCMC chain. If this attribute is None,
+        :meth:`run` will call `self.sampler.init(...)` method for initialization.
 
-        .. warning:: If you want to set a value for this attribute, then making sure
-            that some fields such as `HMCState.i`, `HMCState.mean_accept_prob` are
-            reseted to 0. In addition, if the initial state has invalid fields such as
-            `HMCState.potential_energy` (because e.g. `data` of the new `run` has been
-            changed), the sampler might get trouble during some initial MCMC steps.
-            To resolve that, we can set `potential_energy` to a high value (says `1e38`),
-            so that it is more likely to accept a new proposal (which will have a correct
-            `potential_energy` field) in the first MCMC step.
-            For example
+        .. warning:: If you want to set a value for this attribute, then you will need to make sure
+            that certain fields such as `HMCState.i` and `HMCState.mean_accept_prob` are
+            reset to 0. In addition, if any of the fields of the initial state have invalid values
+            (e.g. `HMCState.potential_energy` may be invalid if `data` of the new `run` has been
+            changed), the sampler might get stuck or otherwise perform suboptimally.
+            To resolve this issue, you can set `potential_energy` to a large value (say `1e38`),
+            so that the sampler is more likely to accept a new proposal in the first MCMC step
+            (and which will have a correct `potential_energy`). For example
 
             .. code-block:: python
 
@@ -420,8 +418,8 @@ class MCMC(object):
     @property
     def warmup_state(self):
         """
-        The state before sampling phase. If this attribute is not None,
-        :meth:`run` will skip warmup phase and start with the state
+        The state before the sampling phase. If this attribute is not None,
+        :meth:`run` will skip the warmup phase and start with the state
         specified in this attribute.
 
         .. note:: This attribute can be used to sequentially draw MCMC samples. For example,
@@ -444,7 +442,7 @@ class MCMC(object):
     @property
     def last_state(self):
         """
-        The state obtained at the end of sampling phase.
+        The final MCMC state at the end of the sampling phase.
         """
         return self._last_state
 
--- a/numpyro/infer/mcmc.py
+++ b/numpyro/infer/mcmc.py
@@ -266,6 +266,7 @@ class MCMC(object):
         # HMCState returned by last warmup
         self._warmup_state = None
         # HMCState returned by hmc.init_kernel
+        self._init_state = None
         self._init_state_cache = {}
         self._cache = {}
         self._collection_params = {}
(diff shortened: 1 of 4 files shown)


numpyro/infer/mcmc.py
```diff
@@ -382,11 +383,75 @@ def _compile(self, rng_key, *args, extra_fields=(), init_params=None, **kwargs):
...
+                mcmc.init_state = mcmc.last_state._replace(i=0, mean_accept_prob=0.)
+                mcmc.run(random.PRNGKey(1), data1)
+                samples_for_data1 = mcmc.get_samples()
+
+            In addition, if the initial state has invalid fields such as
+            `HMCState.potential_energy` (because e.g. `data` of the new `run` has been changed),
+            the sampler might get trouble during some initial MCMC steps.
+            To resolve that, we can set `potential_energy` to a high value (says `1e38`),
+            so that it is more likely to accept a new proposal in the
+            first MCMC step.
```
fehiepsi, line 411: @martinjankowiak Does this warning make sense to you? Any further notes you want to deliver to users?

numpyro/infer/mcmc.py
```diff
@@ -382,11 +383,76 @@ def _compile(self, rng_key, *args, extra_fields=(), init_params=None, **kwargs):
         except TypeError:
             pass
 
+    @property
+    def init_state(self):
+        """
+        The initial state to run MCMC. If this attribute is not None,
```
martinjankowiak, line 389: nit:
The initial state of the MCMC chain. If this attribute is `None`, 
:meth:`run` will call the `self.sampler.init(...)` method for initialization.

numpyro/infer/mcmc.py
```diff
@@ -382,11 +383,76 @@ def _compile(self, rng_key, *args, extra_fields=(), init_params=None, **kwargs):
...
+        .. warning:: If you want to set a value for this attribute, then making sure
+            that some fields such as `HMCState.i`, `HMCState.mean_accept_prob` are
+            reseted to 0. In addition, if the initial state has invalid fields such as
+            `HMCState.potential_energy` (because e.g. `data` of the new `run` has been
+            changed), the sampler might get trouble during some initial MCMC steps.
+            To resolve that, we can set `potential_energy` to a high value (says `1e38`),
+            so that it is more likely to accept a new proposal (which will have a correct
+            `potential_energy` field) in the first MCMC step.
+            For example
+
```
martinjankowiak, line 402: nit:
If you want to set a value for this attribute, then you will need to make sure
that certain fields such as `HMCState.i` and `HMCState.mean_accept_prob` are
reset to 0. In addition, if any of the fields of the initial state have invalid values 
(e.g. `HMCState.potential_energy` may be invalid if `data` of the new `run` has been
changed), the sampler might get stuck or otherwise perform suboptimally.
To resolve this issue, you can set `potential_energy` to a large value (say `1e38`),
so that the sampler is more likely to accept a new proposal in the first MCMC step 
(and which will have a correct `potential_energy`).

numpyro/infer/mcmc.py
```diff
@@ -382,11 +383,76 @@ def _compile(self, rng_key, *args, extra_fields=(), init_params=None, **kwargs):
...
+        return self._init_state
+
+    @init_state.setter
+    def init_state(self, state):
+        self._init_state = state
+
+    @property
+    def warmup_state(self):
+        """
+        The state before sampling phase. If this attribute is not None,
```
martinjankowiak, line 423: nit:
        The state before the sampling phase. If this attribute is not None,
        :meth:`run` will skip the warmup phase and start with the state
        specified in this attribute.

numpyro/infer/mcmc.py
```diff
@@ -382,11 +383,76 @@ def _compile(self, rng_key, *args, extra_fields=(), init_params=None, **kwargs):
...
+        return self._warmup_state
+
+    @warmup_state.setter
+    def warmup_state(self, state):
+        self._warmup_state = state
+
+    @property
+    def last_state(self):
+        """
+        The state obtained at the end of sampling phase.
```
martinjankowiak, line 447: nit: The final MCMC state at the end of the sampling phase.

martinjankowiak (dismissed):

fehiepsi:
Thanks for your suggestions, Martin! I will need to discuss with Neeraj about the API. Especially, I am not sure if https://github.com/pyro-ppl/numpyro/issues/781 is related and can be resolved here.

numpyro/infer/mcmc.py
```diff
@@ -382,11 +383,74 @@ def _compile(self, rng_key, *args, extra_fields=(), init_params=None, **kwargs):
         except TypeError:
             pass
 
+    @property
+    def init_state(self):
+        """
+        The initial state of the MCMC chain. If this attribute is None,
+        :meth:`run` will call `self.sampler.init(...)` method for initialization.
```
neerajprad, line 390: My concern with exposing this is that it might give the impression that users can use this to specify initial value of parameters or mass matrix (which isn't true). However, there are other auxiliary variables like `z_grad` and those mentioned in the warning that are non-obvious and would result in leakage of internal details. Could you elaborate on the use case for this?

numpyro/infer/mcmc.py
```diff
@@ -382,11 +383,74 @@ def _compile(self, rng_key, *args, extra_fields=(), init_params=None, **kwargs):
...
+
+        """
+        return self._init_state
+
+    @init_state.setter
+    def init_state(self, state):
+        self._init_state = state
+
+    @property
+    def warmup_state(self):
```
neerajprad, line 419: If we are to expose this, what do you think about calling this `post_warmup_state` for clarity?

neerajprad:
> I am not sure if #781 is related and can be resolved here.

To resolve this, I think we'll need to separately collect warmup samples and have a `return_warmup` argument to `get_samples(return_warmup=True)` to return these when requested. I am not sure if it is worth looking into in this PR (or at all since I seem to be the only one with that issue 😄 ).

numpyro/infer/mcmc.py
fehiepsi: I agree that this warning is tricky to deliver to users. The motivation here is to start the chain on new data with the previously adapted mass matrix, step_size, and the last sample, as requested by @martinjankowiak  and in https://github.com/pyro-ppl/numpyro/issues/534.

For those who want such a warm start feature, they can still achieve that by modifying the warmup state by `state._replace(i=0, pe=1e6)` and taking care of other related diagnostics information (wrong `z_grad` is fine because `z_grad` on the old data only drives momentum on the first leapfrog step - but a wrong pe might lead to rejecting all proposals). So I think we should not expose it.

numpyro/infer/mcmc.py
fehiepsi: What do you mean by `post`? Does it mean posterior?

fehiepsi:
> seem to be the only one with that issue

I think I am better at understanding that issue now. It is not about how to collect those warmup samples, but about having a more intuitive API to do that job. We can do it later if it is needed. :) Your suggestions there make sense to me.

numpyro/infer/mcmc.py
neerajprad: I meant post as in `after` as opposed to pre. e.g. `post-warmup draws` from pystan.

numpyro/infer/mcmc.py
neerajprad: > The motivation here is to start the chain on new data with the previously adapted mass matrix, step_size, and the last sample

Do you mean starting from these values and running adaptation from there? IIUC there are three things that we can specify for initialization - initial values (which is possible using `init_to_value`?), step size and mass matrix. This method is to be able to expose the latter two. Is that correct?

fehiepsi: Yes, it will re-learn them but started from the previous values. This might be helpful for tricky posteriors, which require spending much time on initial steps...

numpyro/infer/mcmc.py
fehiepsi: I have no preference so I'll make the change. Thanks for your suggestion!

numpyro/infer/mcmc.py
neerajprad: Hmm..how about things like the adaptation window index, wouldn't that need to be reset? 

Another choice might be to have something like `MCMCKernel.get_init_state(self, *args, **kwargs)` method. For HMC, it could be `get_init_state(self, init_params, mass_matrix=None, step_size=None)` which can internally do all this book-keeping (e.g. setting i to 0). That might be a bit more verbose but easier to explain since we can tell the users to use the `get_init_state` method of the corresponding kernel without other caveats.

```python
kernel = HMC(...)
mcmc = MCMC(kernel, ...)
mcmc.initial_state = kernel.get_init_state(init_params, mass_matrix)
mcmc.run(...)
```

fehiepsi: Makes sense to me. We also need args, kwargs so it is more like `kernel.init` method (except that we won't try to find valid initial params). We have discussed how to expose the inverse mass matrix to the API in https://github.com/pyro-ppl/numpyro/pull/536 but found that it is a bit ambiguous. I think we should skip this feature unless there are more requests for it. What do you think?

> how about things like the adaptation window index

We use [hmc_state.i](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/infer/hmc.py#L314) for the update so it should be fine.

neerajprad: I agree, I think it is fine to discuss a bit more before finalizing this. Power users can already do this if they need to, but we should provide an intuitive interface if we are going to advertise this.

neerajprad (approved):

PR #865, opened 2021-01-06
Make subsample faster in CPU

fehiepsi:
This PR is motivated by @insperatum in a [forum topic](https://forum.pyro.ai/t/extremely-slow-subsample/2499) and Jake's answer [here](https://github.com/google/jax/issues/5328#issuecomment-755445979) on the differences between PyTorch CPU and jax implementation.

### Benchmarks
```python
import numpyro
from jax import random, jit

size = 100000
subsample_size = 1000

@jit
def subsample_fn(rng_key):
    return numpyro.primitives._subsample_fn(size, subsample_size, rng_key)

key0, key1 = random.PRNGKey(0), random.PRNGKey(1)
x = subsample_fn(key0).copy()
%time x = subsample_fn(key1).copy()
```
returns
```
CPU times: user 262 µs, sys: 52 µs, total: 314 µs
Wall time: 253 µs
```
while in PyTorch `%time y = torch.randperm(size)[:subsample_size]`, it took
```
CPU times: user 26.5 ms, sys: 1.37 ms, total: 27.9 ms
Wall time: 2.51 ms
```
and in previous implementation, it took
```
CPU times: user 60.2 ms, sys: 269 µs, total: 60.4 ms
Wall time: 54.1 ms
```

The reason for high performance comparing to PyTorch is in PyTorch, we took a permutation of full size first, then collecting a subset. Here, we only take a permutation of size `subsample_size`. This is observed by @fritzo at [this discussion](https://github.com/pyro-ppl/numpyro/pull/734#discussion_r490357510).
```


fritzo (dismissed):
Nice work! We'll have to port this to Pyro if anyone complains about slow subsampling speed there.

fritzo:
Do you need to ensure double precision so that `random.uniform` will sample sets larger than 24e6?

fehiepsi:
Thanks for reviewing and pointing that issue out, @fritzo! I have switched to `randint` instead of using `uniform`. It is a bit slower `451 µs ± 2.06 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)` but is still pretty fast and does not suffer from precision issue.

fritzo (approved):
Sorry I lost track of this!

PR #868, opened 2021-01-09
Warn instead of raising ValueError when substituting a different size subsample indices

fehiepsi:
The motivation is to allow us evaluating log density of the full data, given a model with [subsample](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/infer/hmc_gibbs.py#L395) statement:
```python
def model(data):
     with numpyro.plate("N", data.shape[0], subsample_size=100):
         batch = numpyro.subsample(data, event_dim=0)
         numpyro.sample("obs", dist.Normal(x, 1), obs=batch)
```
Currently, I couldn't find a simpler way to achieve that. :(

### DIFF
--- a/numpyro/primitives.py
+++ b/numpyro/primitives.py
@@ -4,6 +4,7 @@
 from collections import namedtuple
 from contextlib import ExitStack, contextmanager
 import functools
+import warnings
 
 from jax import lax, random
 import jax.numpy as jnp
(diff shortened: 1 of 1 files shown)


neerajprad:
This looks reasonable to me. However, I don't have context to understand when this is useful. Are there situations where you want to subsample during inference but evaluate on the entire dataset?

neerajprad (approved):

fehiepsi:
Yes, in subsample hmc, we need to compute log probability on the whole dataset in `init` method to evalutate some quantity. I think that we can just simply block that subsample size (using `handlers.block`), but it might potentially cause some hidden errors when users didn't specify `dim` in `plate` sites. For example, with the following pattern
```python
with plate("A", 10):
    with plate("B", 10, subsample_size=2):
        with plate("C", 10):
            ...
```
if we block site `B`, then dim of plate `C` will be `-2`, which is incorrect.

neerajprad:
I see - I had missed the earlier HMCGibbs PRs and was therefore confused. Thanks for explaining!

PR #870, opened 2021-01-12
Block updates for gibbs subsampling

OlaRonning:
Introduces block updates into `subsample_gibbs_fn`.

### DIFF
--- a/numpyro/infer/hmc_gibbs.py
+++ b/numpyro/infer/hmc_gibbs.py
@@ -359,7 +359,7 @@ def discrete_gibbs_fn(model, model_args=(), model_kwargs={}, *, random_walk=Fals
     return gibbs_fn
 
 
-def subsample_gibbs_fn(model, model_args=(), model_kwargs={}, num_blocks={}):
+def subsample_gibbs_fn(model, model_args=(), model_kwargs={}, *, num_blocks=1):
     """
     [EXPERIMENTAL INTERFACE]
 
@@ -368,8 +368,9 @@ def subsample_gibbs_fn(model, model_args=(), model_kwargs={}, num_blocks={}):
     of reference [1] but uses a naive estimation (without control variates) of log likelihood,
     hence might incur a high variance.
 
-    The function can parition named subsample statements and update only one block in the parition
+    The function can partition named subsample statements and update only one block in the parition
     to improve acceptance rate of proposed subsamples as detailed in [3].
+    .. note:: New subsample indices are proposed randomly with replacement at each MCMC step.
 
     **References:**
 
@@ -377,7 +378,7 @@ def subsample_gibbs_fn(model, model_args=(), model_kwargs={}, num_blocks={}):
        Dang, K. D., Quiroz, M., Kohn, R., Minh-Ngoc, T., & Villani, M. (2019)
     2. *Speeding Up MCMC by Efficient Data Subsampling*,
        Quiroz, M., Kohn, R., Villani, M., & Tran, M. N. (2018)
-    3. *The Blcok Pseudo-Margional Sampler*,
+    3. *The Block Pseudo-Margional Sampler*,
         Tran, M.-N., Kohn, R., Quiroz, M. Villani, M. (2017)
 
     :param callable model: A callable with NumPyro primitives. This should be the same model
@@ -385,8 +386,7 @@ def subsample_gibbs_fn(model, model_args=(), model_kwargs={}, num_blocks={}):
 
     :param tuple model_args: Arguments provided to the model.
     :param dict model_kwargs: Keyword arguments provided to the model.
-    :param dict num_blocks: Number of blocks to partition subsample sites into.
-                            Defaults to 1 block if site is not present.
+    :param int num_blocks: Number of blocks to partition subsample into.
     :return: A callable `gibbs_fn` to be used in :class:`HMCGibbs`
 
     **Example**
@@ -420,10 +420,7 @@ def subsample_gibbs_fn(model, model_args=(), model_kwargs={}, num_blocks={}):
         for name, site in prototype_trace.items()
         if site["type"] == "plate" and site["args"][0] > site["args"][1]  # i.e. size > subsample_size
     }
-    valid_blocks = all(plate_sizes[name][1] >= num_block for name, num_block in num_blocks.items())
-    assert valid_blocks, "Blocks updates must use block_size <= subsample_size."
 
-    num_blocks.update({name: 1 for name in plate_sizes if name not in num_blocks})
     enum = any(site["type"] == "sample"
                and not site["is_observed"]
                and site["fn"].has_enumerate_support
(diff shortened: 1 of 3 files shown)


numpyro/infer/hmc_gibbs.py
```diff
@@ -358,7 +359,7 @@ def body_fn(i, val):
     return gibbs_fn
 
 
-def subsample_gibbs_fn(model, model_args=(), model_kwargs={}):
+def subsample_gibbs_fn(model, model_args=(), model_kwargs={}, block_updates={}):
```
fehiepsi, line 362: Could you add `, *,` before this new argument (like `discrete_gibbs_fn`)? I don't have a strong opinion about `block_updates` but would something like `block_sizes` or `num_blocks` deliver a better context?

fehiepsi (commented):
Thanks, @OlaRonning! The test and api look good to me. I just have a question to understand the underlying implementation.

numpyro/infer/hmc_gibbs.py
OlaRonning: I'll make the update. Agree that `num_blocks` is a better name.

numpyro/infer/hmc_gibbs.py
```diff
@@ -374,11 +375,13 @@ def subsample_gibbs_fn(model, model_args=(), model_kwargs={}):
...
        Quiroz, M., Kohn, R., Villani, M., & Tran, M. N. (2018)
 
-    :param callable model: a callable with NumPyro primitives. This should be the same model
+    :param callable model: A callable with NumPyro primitives. This should be the same model
         as the one used in the `inner_kernel` of :class:`HMCGibbs`.
+
     :param tuple model_args: Arguments provided to the model.
     :param dict model_kwargs: Keyword arguments provided to the model.
-    :return: a callable `gibbs_fn` to be used in :class:`HMCGibbs`
+    :param dict block_updates: Number of blocks for named sites.
```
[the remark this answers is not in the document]
OlaRonning: Sorry for the spares comments. I've updated the comment and the example. The idea is introduced [here](https://arxiv.org/pdf/1603.02485.pdf) and is basically to partition subsamples and only change one block in the partition to induce correlation.

numpyro/infer/hmc_gibbs.py
```diff
@@ -367,18 +368,26 @@ def subsample_gibbs_fn(model, model_args=(), model_kwargs={}):
     of reference [1] but uses a naive estimation (without control variates) of log likelihood,
     hence might incur a high variance.
 
+    The function can parition named subsample statements and update only one block in the parition
```
fehiepsi, line 371: typo: partition

numpyro/infer/hmc_gibbs.py
```diff
@@ -422,8 +435,16 @@ def gibbs_fn(rng_key, gibbs_sites, hmc_sites):
...
             size, subsample_size = plate_sizes[name]
-            rng_key, subkey = random.split(rng_key)
-            u_new[name] = random.choice(subkey, size, (subsample_size,), replace=False)
+            num_block = num_blocks[name]
+            rng_key, subkey, block_key = random.split(rng_key, 3)
+            block_size = subsample_size // num_block
+
+            chosen_block = random.randint(block_key, shape=(), minval=0, maxval=num_block)
+            new_idx = random.randint(subkey, minval=0, maxval=size, shape=(subsample_size,))
+            block_mask = (jnp.arange(subsample_size) // block_size == chosen_block).astype(int)
```
fehiepsi, line 444: nit: no need to use `astype` here, then you can use `u_new[name] = jnp.where(block_mask, new_idx, gibbs_sites[name])`

numpyro/infer/hmc_gibbs.py
```diff
@@ -367,18 +368,26 @@ def subsample_gibbs_fn(model, model_args=(), model_kwargs={}):
     of reference [1] but uses a naive estimation (without control variates) of log likelihood,
     hence might incur a high variance.
 
+    The function can parition named subsample statements and update only one block in the parition
+    to improve acceptance rate of proposed subsamples as detailed in [3].
+
```
fehiepsi, line 373: Could you add a note:
```
.. note:: Subsample indices are sampled with replacement at each MCMC step.
```

fehiepsi (approved):
Thanks, @OlaRonning! I understand the PR much better now. The logic looks great to me. I just have a few nits to comment.

numpyro/infer/hmc_gibbs.py
OlaRonning: With integer arithmetics the control flow can be executed using fully data parallel operations. So while it comprises readability it's measurably faster when run on parallel hardware.

fehiepsi: Oh, it is good to know. Make sense to me but I need to double check. In the mean time, you can define `rest_mask = 1 - block_mask` I guess.

numpyro/infer/hmc_gibbs.py
```diff
@@ -359,7 +359,7 @@ def body_fn(i, val):
     return gibbs_fn
 
 
-def subsample_gibbs_fn(model, model_args=(), model_kwargs={}, block_updates={}):
+def subsample_gibbs_fn(model, model_args=(), model_kwargs={}, num_blocks={}):
```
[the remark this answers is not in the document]
OlaRonning: Ups, put it on the todo and got distracted by a bug in the code. I'll update with the ',*,'.

numpyro/infer/hmc_gibbs.py
fehiepsi: Just for reference,
```python
import jax.numpy as jnp
from jax import jit, lax, config
config.update("jax_platform_name", "cpu")  # "gpu"

def f1(i, x):
    y = jnp.zeros(x.shape)
    mask = jnp.full(x.shape, True)
    return jnp.where(mask, x, y) + 1

def f2(i, x):
    y = jnp.zeros(x.shape)
    mask = jnp.full(x.shape, True).astype(int)
    return x * mask + y * (1 - mask) + 1

@jit
def g1(x):
    return lax.fori_loop(0, 10000, f1, x)

@jit
def g2(x):
    return lax.fori_loop(0, 10000, f2, x)

for shape in [100, 1000, 10000, 100000]:
    x = jnp.arange(float(shape))
    g1(x)
    g2(x)
    print("Test for shape: {}".format(shape))
    %timeit z = g1(x).block_until_ready()
    %timeit z = g2(x).block_until_ready()
```
In CPU, `where` is faster
```
Test for shape: 100
227 µs ± 1.02 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
273 µs ± 7.21 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
Test for shape: 1000
932 µs ± 50 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
1.34 ms ± 17 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
Test for shape: 10000
7.4 ms ± 29.2 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
10.5 ms ± 162 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
Test for shape: 100000
72.9 ms ± 799 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)
109 ms ± 586 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
In GPU, they are similar
```
Test for shape: 100
97.8 ms ± 632 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
97.7 ms ± 439 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
Test for shape: 1000
97.2 ms ± 396 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
99.8 ms ± 3.06 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
Test for shape: 10000
107 ms ± 383 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
107 ms ± 202 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
Test for shape: 100000
107 ms ± 346 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
104 ms ± 3.74 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

OlaRonning: I expected the effect to be larger (and not in favor of `jnp.where`). Thanks for checking! I'll make the update to `jnp.where`.

numpyro/infer/hmc_gibbs.py
```diff
@@ -367,18 +368,26 @@ def subsample_gibbs_fn(model, model_args=(), model_kwargs={}):
...
+        Tran, M.-N., Kohn, R., Quiroz, M. Villani, M. (2017)
 
-    :param callable model: a callable with NumPyro primitives. This should be the same model
+    :param callable model: A callable with NumPyro primitives. This should be the same model
         as the one used in the `inner_kernel` of :class:`HMCGibbs`.
+
     :param tuple model_args: Arguments provided to the model.
     :param dict model_kwargs: Keyword arguments provided to the model.
-    :return: a callable `gibbs_fn` to be used in :class:`HMCGibbs`
+    :param dict num_blocks: Number of blocks to partition subsample sites into.
```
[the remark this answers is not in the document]
OlaRonning: I had a contrived example with an HMM, which emits multiple distributions in mind; suppose it's a minor change if it's ever requested. I'll make the update to a single value.

fehiepsi (approved):

PR #872, opened 2021-01-13
Add default values to constraints to make grad propogate properly

### LINKED ISSUE #871: Gradient propagation failure for masked invalid values
I am working on a model that requires implementing a plate over arrays with variable sizes. My understanding is that the conventional solution to this problem is padding shorter arrays and then masking the padded values. 

I initially implemented this with `nan` pads, but found that that yielded completely divergent samples. 

With @fehiepsi's [help](https://forum.pyro.ai/t/hierarchical-bayesian-model-with-data-of-varying-length/1588), I was able to solve this problem by padding with a valid value instead of `nan`. They suggested that even though the padding values are masked, the gradient is still propagated incorrectly. 

I can confirm that parameter estimates seem to be invariant across masked valid values; however, masked invalid values (e.g. including `nan` and values excluded from the support of the relevant probability distribution) cause 100% divergence.

@fehiepsi asked me to open an issue and suggested a fix that automatically replaces invalid values with valid values.  

I included a reproducible example below. 

```


# Generate Data

# specify number of observations for each group
n1 = 2000
n2 = 1000

# Specify pad value
pad = -1

# Sample observations from different beta distributions for each group, given specified sample size
np.random.seed(seed=123)
y1 = beta.rvs(.5, .5, size=n1)
np.random.seed(seed=321)
y2 = beta.rvs(3, 1, size=n2)

pads = numpy.ones((n1-n2)) * pad
y2 = np.concatenate([y2, pads])

# Combine samples
y = np.array([y1, y2], order='K').T

sample_sizes =  np.array([n1, n2])

def model(sample_sizes, y = None):
    
    # Define the number of groups from which we've attained samples
    n_groups =  sample_sizes.shape[0]
    
    # Initialize a plate for each group
    with numpyro.plate("n_groups", n_groups):
        
        # Set priors on a and b within each group
        a = numpyro.sample('a', dist.Gamma(.5, .5))
        b =  numpyro.sample('b', dist.Gamma(3, 3))
        
        # Calculate maximum group size
        I = sample_sizes.max().item()
        # Create a range of 0:I
        i = torch.arange(I).unsqueeze(-1).numpy()
        
        # Initialize a plate for each observation
        with numpyro.plate('data', I):
            # Mask observations that exceed sample size in the respective plate
            with numpyro.handlers.mask(mask = i < sample_sizes): 
                # Estimate y_hat
                y_hat = numpyro.sample('y_hat', dist.Beta(a, b), obs=y)

rng_key = random.PRNGKey(0)
rng_key, rng_key_ = random.split(rng_key)

num_warmup, num_samples = 500, 2000

# Run NUTS.
kernel = NUTS(model)
mcmc = MCMC(kernel, num_warmup, num_samples, num_chains = 3)

start = time.time()
mcmc.run(rng_key_, sample_sizes=sample_sizes, y = y)
end = time.time()
mcmc.print_summary()
```

fehiepsi:
Fixes  #871.

In https://github.com/pyro-ppl/pyro/issues/2447, @fritzo proposed to implement a safe_mask function, but it seems to me that this is an easier to implement solution.

### DIFF
--- a/numpyro/distributions/constraints.py
+++ b/numpyro/distributions/constraints.py
@@ -73,9 +73,9 @@ class Constraint(object):
         """
         return self(value)
 
-    def _default_value(self, shape):
+    def feasible_like(self, prototype):
         """
-        Get default value with shape specified in `shape`.
+        Get a feasible value which has the same shape as dtype as `prototype`.
         """
         raise NotImplementedError
 
@@ -84,7 +84,7 @@ class _Boolean(Constraint):
     def __call__(self, x):
         return (x == 0) | (x == 1)
 
-    def default_like(self, prototype):
+    def feasible_like(self, prototype):
         return jax.numpy.zeros_like(prototype)
 
 
@@ -98,7 +98,7 @@ class _CorrCholesky(Constraint):
         unit_norm_row = jnp.all((x_norm <= 1) & (x_norm > 1 - 1e-6), axis=-1)
         return lower_triangular & positive_diagonal & unit_norm_row
 
-    def default_like(self, prototype):
+    def feasible_like(self, prototype):
         return jax.numpy.broadcast_to(jax.numpy.eye(prototype.shape[-1]), prototype.shape)
 
 
@@ -113,7 +113,7 @@ class _CorrMatrix(Constraint):
         unit_variance = jnp.all(jnp.abs(jnp.diagonal(x, axis1=-2, axis2=-1) - 1) < 1e-6, axis=-1)
         return symmetric & positive & unit_variance
 
-    def default_like(self, prototype):
+    def feasible_like(self, prototype):
         return jax.numpy.broadcast_to(jax.numpy.eye(prototype.shape[-1]), prototype.shape)
 
 
@@ -133,7 +133,7 @@ class _GreaterThan(Constraint):
     def __call__(self, x):
         return x > self.lower_bound
 
-    def default_like(self, prototype):
+    def feasible_like(self, prototype):
         return jax.numpy.broadcast_to(self.lower_bound + 1, jax.numpy.shape(prototype))
 
 
@@ -144,7 +144,7 @@ class _LessThan(Constraint):
     def __call__(self, x):
         return x < self.upper_bound
 
-    def default_like(self, prototype):
+    def feasible_like(self, prototype):
         return jax.numpy.broadcast_to(self.upper_bound - 1, jax.numpy.shape(prototype))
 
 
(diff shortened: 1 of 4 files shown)


numpyro/distributions/distribution.py
```diff
@@ -687,6 +687,11 @@ def log_prob(self, value):
             return jnp.zeros(shape)
         if self._mask is True:
             return self.base_dist.log_prob(value)
+        try:
+            default_value = self.base_dist.support._default_value(self.base_dist.event_shape)
+            value = jnp.where(self._mask, value, default_value)
+        except NotImplementedError:
+            pass
```
fritzo, line 694: nit: restrict scope of potential errors:
```py
try:
    default_value = self.base_dist.support._default_value(self.base_dist.event_shape)
except NotImplementedError:
    pass
else:
    value = jnp.where(self._mask, value, default_value)
```

fritzo (commented):
This is a great idea, especially since constraints do the checking.

Could I suggest a different interface, so as to be compatible with PyTorch, where we might need to know precision and device placement?
```
class Constraint:
    def default_like(self, prototype):
        raise NotImplementedError

class Positive(Constraint):
    def default_like(self, prototype):
        return np.ones(prototype.shape, dtype=prototype.dtype)
```
This is similar to `torch.zeros_like()` and similar functions. If you're ok with this, I'd like to copy this pattern in PyTorch/Pyro.

fehiepsi:
Thanks, @fritzo! I'll update the api. It looks reasonable to me.

numpyro/distributions/constraints.py
```diff
@@ -73,11 +73,20 @@ def check(self, value):
         """
         return self(value)
 
+    def _default_value(self, shape):
+        """
+        Get default value with shape specified in `shape`.
+        """
```
fritzo, line 79: Update api

fritzo (changes requested):
Hmm I think I prefer `.feasible_like()` over `.default_like()`, since the former is evokes the language of constraints. @fehiepsi do you have a preference?

numpyro/distributions/constraints.py
fehiepsi: Thanks for the catch! I don't have preference. :D

fritzo (approved):
Looks great!  Mind if I port these to PyTorch, or would you like to?

fehiepsi:
Thanks for reviewing and valuable feedbacks, @fritzo! I would appreciate if you port those to PyTorch. :)

PR #873, opened 2021-01-14
Revise API of gibbs functions

fehiepsi:
This PR revises the API of discrete_gibbs_fn and subsample_gibbs_fn for both:
+ improve flexibility: currently, a single `gibbs_fn` together with its restricted arguments is not flexible enough to initialize necessary information for the metropolis update other than the HMC (e.g. the behavior `use_unconstrained_gibbs_fn`). In addition, this enables further enhancements such as: passing `potential_energy`, `potential_energy_gen`,... between gibbs_fn and the main kernel `sample` method, and getting other diagnostics rather than the current one in HMCGibbs.
+ clean up the user API: avoid unnecessary users' specifications such as `model`, `model_args`, `model_kwargs`, `gibbs_fn`, `gibbs_sites`.

To do that, I inherited the class `HMCGibbs` (NB: MixedHMC and MixedHMCGibbs in #826 also inherited from this class).

Things are changed in this PR:
+ the api of the above two functions (I probably move things around a bit but the implementation should be the same)
+ add `self._use_constrained_gibbs_fn` to tell `gibbs_fn` if `z_hmc` is constrained or unconstrained
+ add `self._prototype_trace` to avoid recomputing the trace
+ marginalization in `DiscreteHMCGibbs` requires an explicit `infer={"enumerate": "parallel"}` specification in `sample` statement, rather than automatically as before.

cc @OlaRonning I think this change is also more appropriate for HMCECS, where it might need a `rng_key` to initialize some thing (I seem to me that proxy stuff there is sensitive to the seed of the prototype trace)

### DIFF
--- a/numpyro/infer/hmc_gibbs.py
+++ b/numpyro/infer/hmc_gibbs.py
@@ -315,8 +315,8 @@ class DiscreteHMCGibbs(HMCGibbs):
 
     .. note:: The site update order is randomly permuted at each step.
 
-    .. note:: This class supports mixing with enumeration. To marginalize out a discrete
-        latent site, we can specify `infer={'enumerate': 'parallel'}` keyword in its
+    .. note:: This class supports enumeration of discrete latent variables. To marginalize out a
+        discrete latent site, we can specify `infer={'enumerate': 'parallel'}` keyword in its
         corresponding :func:`~numpyro.primitives.sample` statement.
 
     :param inner_kernel: One of :class:`~numpyro.infer.hmc.HMC` or :class:`~numpyro.infer.hmc.NUTS`.
@@ -354,7 +354,10 @@ class DiscreteHMCGibbs(HMCGibbs):
         >>> kernel = DiscreteHMCGibbs(NUTS(model), modified=True)
         >>> mcmc = MCMC(kernel, 1000, 100000, progress_bar=False)
         >>> mcmc.run(random.PRNGKey(0), probs, locs)
-        >>> mcmc.print_summary()  # doctest: +SKIP
+        >>> mcmc.print_summary()
+        >>> samples = mcmc.get_samples()["x"]
+        >>> assert abs(jnp.mean(samples) - 1.3) < 0.1
+        >>> assert abs(jnp.var(samples) - 4.36) < 0.5
 
     """
 
@@ -412,13 +415,13 @@ class HMCECS(HMCGibbs):
 
     HMC with Energy Conserving Subsampling.
 
-    A subclass of :class:`HMCGibbs` to perform HMC-within-Gibbs for model with subsampling
-    statements using :class:`~numpyro.plate` primitive. This implements the Algorithm 1
+    A subclass of :class:`HMCGibbs` for performing HMC-within-Gibbs for models with subsample
+    statements using the :class:`~numpyro.plate` primitive. This implements Algorithm 1
     of reference [1] but uses a naive estimation (without control variates) of log likelihood,
     hence might incur a high variance.
 
-    The function can partition named subsample statements and update only one block in the parition
-    to improve acceptance rate of proposed subsamples as detailed in [3].
+    The function can divide subsample indices into blocks and update only one block at each
+    MCMC step to improve the acceptance rate of proposed subsamples as detailed in [3].
 
     .. note:: New subsample indices are proposed randomly with replacement at each MCMC step.
 
@@ -455,7 +458,7 @@ class HMCECS(HMCGibbs):
         >>> mcmc = MCMC(kernel, 1000, 1000)
         >>> mcmc.run(random.PRNGKey(0), data)
         >>> samples = mcmc.get_samples()["x"]
-        >>> assert abs(jnp.mean(samples).copy() - 1.) < 0.1
+        >>> assert abs(jnp.mean(samples) - 1.) < 0.1
 
     """
     def __init__(self, inner_kernel, *, num_blocks=1):
--- a/numpyro/infer/hmc_gibbs.py
+++ b/numpyro/infer/hmc_gibbs.py
@@ -9,7 +9,6 @@ from jax import device_put, ops, random, value_and_grad
 import jax.numpy as jnp
 from jax.scipy.special import expit
 
-from numpyro.distributions import biject_to
 from numpyro.handlers import condition, seed, substitute, trace
 from numpyro.infer.hmc import HMC
 from numpyro.infer.mcmc import MCMCKernel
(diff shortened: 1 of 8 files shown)


OlaRonning (commented):
Hi @fehiepsi, this looks good! 

The proxies should not have any random behavior when initialized, however, the model wrapper will depend on the estimator and this will be simpler if it's just set at `init`.

Have a couple of comments, see above.

numpyro/infer/hmc_gibbs.py
```diff
@@ -315,12 +264,9 @@ def discrete_gibbs_fn(model, model_args=(), model_kwargs={}, *, random_walk=Fals
...
 
     def gibbs_fn(rng_key, gibbs_sites, hmc_sites):
-        # convert to unconstrained values
-        z_hmc = {k: biject_to(prototype_trace[k]["fn"].support).inv(v)
-                 for k, v in hmc_sites.items()
-                 if k in prototype_trace and prototype_trace[k]["type"] == "sample"}
+        z_hmc = hmc_sites
         use_enum = len(set(support_sizes) - set(gibbs_sites)) > 0
-        wrapped_model = _wrap_model(model)
+        wrapped_model = model
```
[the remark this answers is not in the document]
fehiepsi: Good point! I'll make the change.

numpyro/infer/hmc_gibbs.py
```diff
@@ -93,6 +92,8 @@ def __init__(self, inner_kernel, gibbs_fn, gibbs_sites):
         self.inner_kernel._model = _wrap_model(inner_kernel.model)
         self._gibbs_sites = gibbs_sites
         self._gibbs_fn = gibbs_fn
+        self._use_constrained_gibbs_fn = True
```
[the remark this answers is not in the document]
fehiepsi: I think we can just use the default behavior, to not populate the api, until users need the `False` pattern.

numpyro/infer/hmc_gibbs.py
```diff
@@ -359,35 +307,33 @@ def body_fn(i, val):
...
-    hence might incur a high variance.
+    A subclass of :class:`HMCGibbs` which performs Metropolis updates for discrete latent sites.
 
-    The function can partition named subsample statements and update only one block in the parition
-    to improve acceptance rate of proposed subsamples as detailed in [3].
-    .. note:: New subsample indices are proposed randomly with replacement at each MCMC step.
+    .. note:: The site update order is randomly permuted at each step.
 
-    **References:**
+    .. note:: This class supports mixing with enumeration. To marginalize out a discrete
```
martinjankowiak, line 318: nit: This class supports enumeration of discrete latent variables.

numpyro/infer/hmc_gibbs.py
```diff
@@ -397,35 +343,42 @@ def subsample_gibbs_fn(model, model_args=(), model_kwargs={}, *, num_blocks=1):
...
-        >>> gibbs_fn = subsample_gibbs_fn(model, (data,), num_blocks={'N': 10})
-        >>> kernel = HMCGibbs(NUTS(model), gibbs_fn, gibbs_sites=["N"])
-        >>> mcmc = MCMC(kernel, 1000, 1000)
-        >>> mcmc.run(random.PRNGKey(0), data)
-        >>> samples = mcmc.get_samples()["x"]
-        >>> assert abs(jnp.mean(samples).copy() - 1.) < 0.1
+        >>> probs = jnp.array([0.15, 0.3, 0.3, 0.25])
+        >>> locs = jnp.array([-2, 0, 2, 4])
+        >>> kernel = DiscreteHMCGibbs(NUTS(model), modified=True)
+        >>> mcmc = MCMC(kernel, 1000, 100000, progress_bar=False)
```
martinjankowiak, line 355: why do this many samples if there's no `assert`?

numpyro/infer/hmc_gibbs.py
```diff
@@ -441,13 +394,90 @@ def gibbs_fn(rng_key, gibbs_sites, hmc_sites):
...
     return gibbs_fn
+
+
+class HMCECS(HMCGibbs):
+    """
+    [EXPERIMENTAL INTERFACE]
+
+    HMC with Energy Conserving Subsampling.
+
+    A subclass of :class:`HMCGibbs` to perform HMC-within-Gibbs for model with subsampling
```
martinjankowiak, line 415: nit: for performing

martinjankowiak: nit: for models with subsample statements using the

numpyro/infer/hmc_gibbs.py
```diff
@@ -441,13 +394,90 @@ def gibbs_fn(rng_key, gibbs_sites, hmc_sites):
...
+    statements using :class:`~numpyro.plate` primitive. This implements the Algorithm 1
```
martinjankowiak, line 416: nit: This implements Algorithm 1

numpyro/infer/hmc_gibbs.py
```diff
@@ -441,13 +394,90 @@ def gibbs_fn(rng_key, gibbs_sites, hmc_sites):
...
+    of reference [1] but uses a naive estimation (without control variates) of log likelihood,
+    hence might incur a high variance.
+
+    The function can partition named subsample statements and update only one block in the parition
```
martinjankowiak, line 420: typo: partition

numpyro/infer/hmc_gibbs.py
```diff
@@ -441,13 +394,90 @@ def gibbs_fn(rng_key, gibbs_sites, hmc_sites):
...
+    to improve acceptance rate of proposed subsamples as detailed in [3].
```
martinjankowiak, line 421: nit: improve the acceptance rate

numpyro/infer/hmc_gibbs.py
```diff
@@ -441,13 +394,90 @@ def gibbs_fn(rng_key, gibbs_sites, hmc_sites):
...
+        ...     with numpyro.plate("N", data.shape[0], subsample_size=100):
+        ...         batch = numpyro.subsample(data, event_dim=0)
+        ...         numpyro.sample("obs", dist.Normal(x, 1), obs=batch)
+        ...
+        >>> data = random.normal(random.PRNGKey(0), (10000,)) + 1
+        >>> kernel = HMCECS(NUTS(model), num_blocks=10)
+        >>> mcmc = MCMC(kernel, 1000, 1000)
+        >>> mcmc.run(random.PRNGKey(0), data)
+        >>> samples = mcmc.get_samples()["x"]
+        >>> assert abs(jnp.mean(samples).copy() - 1.) < 0.1
```
martinjankowiak, line 458: why copy?

numpyro/infer/hmc_gibbs.py
fehiepsi: Oops, probably this losts during a commit. I'll reverse the assertations.

numpyro/infer/hmc_gibbs.py
fehiepsi: There is no reason. I'll remove it.

fehiepsi:
Thanks for reviewing, @OlaRonning and @martinjankowiak!

martinjankowiak (approved):

Issue #874, opened 2021-01-14
Update distribution api with new upstream changes


fehiepsi:
This tracks some enhancemences in PyTorch/Pyro distributions. Some are required by `funsor`. Some are changed for consistency.
- [x] `.infer_shapes()` https://github.com/pyro-ppl/pyro/pull/2739
- [x] `Transform.inv`  #164
- [x] Move `is_discrete` to constraints (deprecate `Distribution.is_discrete`?)
- [x] Add `event_dim` to constraint (remove transforms' input_event_dim/output_event_dim?)

fritzo:
@fehiepsi I'll try to submit a `Transforms` PR to PyTorch tomorrow, where we can discuss the Transforms api. I like your idea of getting rid of `.input_event_dim` and `.output_event_dim`, and I think we can even deprecate `.event_dim`.

fehiepsi:
Thanks, @fritzo, for resolving this issue!

PR #876, opened 2021-01-15
Add constraint event_dim

fehiepsi:
Addresses #874.

In this PR, I copied @fritzo's effort at https://github.com/pytorch/pytorch/pull/50547 and added his suggestion for IndependentTransform. Also removing `input_event_dim/output_event_dim` and deprecating `Transform.event_dim`.

### DIFF
--- a/numpyro/distributions/distribution.py
+++ b/numpyro/distributions/distribution.py
@@ -35,7 +35,7 @@ import numpy as np
 from jax import lax, tree_util
 import jax.numpy as jnp
 
-from numpyro.distributions.constraints import is_dependent, real
+from numpyro.distributions.constraints import independent, is_dependent, real
 from numpyro.distributions.transforms import ComposeTransform, Transform
 from numpyro.distributions.util import lazy_property, promote_shapes, sum_rightmost, validate_sample
 from numpyro.util import not_jax_tracer
@@ -582,7 +582,7 @@ class Independent(Distribution):
 
     @property
     def support(self):
-        return self.base_dist.support
+        return independent(self.base_dist.support, self.reinterpreted_batch_ndims)
 
     @property
     def has_enumerate_support(self):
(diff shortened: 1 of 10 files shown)


numpyro/distributions/distribution.py
```diff
@@ -797,11 +798,13 @@ def rsample(self, key, sample_shape=()):
...
-            t.domain = domain
-            domain = t.codomain
-        return domain
+        codomain = self.transforms[-1].codomain
+        codomain_event_dim = codomain.event_dim
+        assert self.event_dim >= codomain_event_dim
+        if self.event_dim == codomain_event_dim:
+            return codomain
+        else:
+            return independent(codomain, self.event_dim - codomain_event_dim)
```
fritzo, line 807: I believe this needs more complex logic as in https://github.com/pytorch/pytorch/pull/50581/files#diff-8b108de0fae81bfc4f939e40322885653fc3c957b2f8ad0ff25aafc16bf2c639

numpyro/distributions/transforms.py
```diff
@@ -168,18 +152,28 @@ def codomain(self):
...
+    for part in parts[len(parts) - 1::-1]:
+        input_event_dim = part.domain.event_dim + max(input_event_dim - part.codomain.event_dim, 0)
+    return input_event_dim
+
+
+def _get_compose_transform_output_event_dim(parts):
+    output_event_dim = parts[0].codomain.event_dim
+    for part in parts[1:]:
+        output_event_dim = part.codomain.event_dim + max(output_event_dim - part.domain.event_dim, 0)
+    return output_event_dim
```
fritzo, line 176: This is a bit trickier in case any parts change event dim. I believe there are correct implementations in https://github.com/pytorch/pytorch/pull/50581/files#diff-9c38272c807be20a5832ca58bc3b1f3454c70d30e28ecbd39ca40191219f7d81

fritzo (changes requested):
@fehiepsi could you also take a look at the updated PR https://github.com/pytorch/pytorch/pull/50581?

numpyro/distributions/distribution.py
fehiepsi: Yes, I added a comment there. Here we fixed `self.event_dim` directly from the constructor.

numpyro/distributions/transforms.py
fehiepsi: I just saw that we are making duplicated effort (they are equivalent, I did comment about the reason in your PR). ;)

numpyro/distributions/distribution.py
fritzo: Ok, it seems reasonable to defer `.forward_shape()` to a separate PR.

fritzo (approved):

test/test_distributions.py
```diff
@@ -230,81 +230,85 @@ def _is_batched_multivariate(jax_dist):
...
-    elif isinstance(constraint, constraints._IntegerGreaterThan):
+    elif isinstance(constraint, constraints.integer_greater_than):
         return constraint.lower_bound + random.poisson(key, np.array(5), shape=size)
-    elif isinstance(constraint, constraints._Interval):
+    elif isinstance(constraint, constraints.interval):
         lower_bound = jnp.broadcast_to(constraint.lower_bound, size)
         upper_bound = jnp.broadcast_to(constraint.upper_bound, size)
         return random.uniform(key, size, minval=lower_bound, maxval=upper_bound)
-    elif isinstance(constraint, (constraints._Real, constraints._RealVector)):
+    elif constraint in (constraints.real, constraints.real_vector):
```
neerajprad, line 247: FYI - we may expose these classes in the API (https://github.com/pytorch/pytorch/issues/50616) so that this `ininstance/is` distinction can be avoided.

PR #878, opened 2021-01-16
Support forward-mode differentiation

### LINKED ISSUE #877: Support forward mode differentiation in HMC
Currently, we use reverse mode differentiation. But forward-mode is also useful in case users' data shapes change and they don't want to recompile the model.

fehiepsi:
Fixes #877.

This adds support for forward-mode differentiation, hence users can use `fori_loop` or `while_loop` in their models.

@eb8680 With this, it is pretty straightforward to support nested loops with only observed variable (it might be more complicated to have latent variables in body_fn but might be doable) in `contrib.control_flow.fori_loop/while_loop`. :)

@OlaRonning I think after this, we can implement block poisson hmcecs.

### DIFF
--- a/numpyro/infer/hmc.py
+++ b/numpyro/infer/hmc.py
@@ -147,6 +147,7 @@ def hmc(potential_fn=None, potential_fn_gen=None, kinetic_fn=None, algo='NUTS'):
     max_treedepth = None
     wa_update = None
     wa_steps = None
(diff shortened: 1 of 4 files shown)


neerajprad:
LGTM, you'll need to merge changes from the latest master.

neerajprad (approved):

PR #884, opened 2021-01-21
Enhance DiscreteHMCGibbs and HMCECS

fehiepsi:
This PR reimplemented `sample` method of `HMCGibbs.sample` to make `gibbs_fn` access to HMC potential_fn and save some computations for computing the current potential energy and the new potential energy. By using HMC potential_fn, we can remove all enum stuffs in `discrete_gibbs_fn`, and no need to use `log_likelihood` in `subsample_gibbs_fn` (hence enable using `enum` in HMCECS).

In addition, I enabled `forward_mode_differentiation` and stored `_discrete_proposal_fn` inside the class `DiscreteHMCGibbs` because I would like to reuse it in [MixedHMC implementation](https://github.com/pyro-ppl/numpyro/pull/826), where MixedHMC is a subclass of DiscreteHMCGibbs.


numpyro/contrib/funsor/enum_messenger.py
```diff
@@ -475,7 +475,27 @@ def process_message(self, msg):
...
+                        if msg["type"] == "param":
+                            statement = "numpyro.param({}, ..., event_dim={})".format(msg["name"], event_dim)
+                        else:
+                            statement = "numpyro.subsample(..., event_dim={})".format(event_dim)
+                        raise ValueError(
+                            "Inside numpyro.plate({}, {}, dim={}) invalid shape of {}: {}"
+                            .format(self.name, self.size, self.dim, statement, shape))
+                    if self.subsample_size < self.size:
+                        value = msg["value"]
+                        new_value = jnp.take(value, self.indices, dim)
```
fehiepsi, line 497: @eb8680 This is inconsistent between the origin plate and here. There `self._indices` is an ndarray while here it is a tuple. So I copied the implementation there to here and change `self._indices` to `self.indices`.

martinjankowiak (approved):
lgtm but defer to @eb8680 on changes in enum_messenger.py

eb8680 (approved):
Sorry, missed this. Changes LGTM.

fehiepsi:
Thanks for reviewing, @martinjankowiak and @eb8680!

PR #885, opened 2021-01-21
Memoize Transform.inv

fritzo:
Fixes behavior mismatch blocking https://github.com/pyro-ppl/funsor/pull/427 identified by @eb8680 

This makes NumPyro's `Transform.inv` property act more like PyTorch's.

## Tested
- [x] added a new assertion
- [x] verified funsor tests pass locally

### DIFF
--- a/test/test_distributions.py
+++ b/test/test_distributions.py
@@ -1038,6 +1038,7 @@ def test_bijective_transforms(transform, event_shape, batch_shape):
     z = transform.inv(y)
     assert_allclose(x, z, atol=1e-6, rtol=1e-6)
     assert transform.inv.inv is transform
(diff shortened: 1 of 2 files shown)


test/test_distributions.py
```diff
@@ -1038,6 +1038,7 @@ def test_bijective_transforms(transform, event_shape, batch_shape):
     z = transform.inv(y)
     assert_allclose(x, z, atol=1e-6, rtol=1e-6)
     assert transform.inv.inv is transform
+    assert transform.inv is transform.inv
```
fehiepsi, line 1041: Thanks for the fix, @fritzo!

fehiepsi (approved):

PR #886, opened 2021-01-21
Make TransformReparam compatible with .to_event()

fritzo:
Port of https://github.com/pyro-ppl/pyro/pull/2746
Fixes a bug [on the forum](https://forum.pyro.ai/t/using-transformreparam-with-to-event/2553).

This PR:
- refactors `._unexpand()` into `._unwrap()` to handle arbitrary interleaved `Independent` and `ExpandedDistribution`;
- uses `._unwrap()` in `TransformedDistribution`
- updates `TransformDistribution` computation of `event_dim` to support batched transforms applied to scalar base distributions.

## Tested
- [x] added new test cases

### DIFF
--- a/numpyro/infer/reparam.py
+++ b/numpyro/infer/reparam.py
@@ -4,6 +4,7 @@
 from abc import ABC, abstractmethod
 
 import jax.numpy as jnp
+from jax import lax
 
 import numpyro
 import numpyro.distributions as dist
@@ -36,10 +37,18 @@ class Reparam(ABC):
             fn = fn.base_dist
         return fn, batch_shape, event_dim
 
-    def _wrap(self, fn, event_dim):
+    def _wrap(self, fn, batch_shape, event_dim):
         """
-        Wrap in Independent distributions.
+        Wrap in Independent and ExpandedDistribution distributions.
         """
+        # Match batch_shape.
+        assert fn.event_dim <= event_dim
+        fn_batch_shape = batch_shape + (1,) * (event_dim - fn.event_dim)
+        fn_batch_shape = lax.broadcast_shapes(fn_batch_shape, fn.batch_shape)
+        if fn.batch_shape != fn_batch_shape:
+            fn = fn.expand(fn_batch_shape)
+
+        # Match event_dim.
         if fn.event_dim < event_dim:
             fn = fn.to_event(event_dim - fn.event_dim)
         assert fn.event_dim == event_dim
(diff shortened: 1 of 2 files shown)


numpyro/infer/reparam.py
```diff
@@ -127,14 +120,15 @@ class TransformReparam(Reparam):
...
-        fn, batch_shape = self._unexpand(fn)
+        fn, batch_shape, event_dim = self._unwrap(fn)
         assert isinstance(fn, dist.TransformedDistribution)
 
         # Draw noise from the base distribution.
-        # We need to make sure that we have the same batch_shape
-        reinterpreted_batch_ndims = fn.event_dim - fn.base_dist.event_dim
+        base_event_dim = event_dim
+        for t in reversed(fn.transforms):
+            base_event_dim += t.domain.event_dim - t.codomain.event_dim
```
fritzo, line 129: This is important in cases like
```py
TransformedDistribution(Normal(np.zeros(10), 1),  biject_to(constraits.simplex))
```
where `fn.base_dist.event_dim == 0` but `fn.event_dim == 1`.

numpyro/infer/reparam.py
```diff
@@ -100,11 +91,13 @@ def __call__(self, name, fn, obs):
                                      constraint=constraints.unit_interval)
         params["loc"] = fn.loc * centered
         params["scale"] = fn.scale ** centered
-        decentered_fn = type(fn)(**params).expand(batch_shape)
+        decentered_fn = type(fn)(**params)
+        decentered_fn = self._wrap(decentered_fn, event_dim)
+        decentered_fn = decentered_fn.expand(batch_shape)
```
fehiepsi, line 96: Agreed! We should correct event_dim first, then doing `expand`. However, this might raise an issue for the case:
`dist.Normal(0, 1).expand([10]).to_event(1)`, where `decentered_fn = dist.Normal(0, 1)` and it can't be expanded. I think we can do something like
```
decentered_fn = self._wrap(decentered_fn.expand(batch_shape + event_shape), event_dim)
```

numpyro/infer/reparam.py
fehiepsi: I think we need to use the logic that we recently added to ComposedTransform here.
```python
base_event_dim = event_dim
t = ComposedTransform(fn.transforms)
base_event_dim += t.domain.event_dim - t.codomain.event_dim
```

fehiepsi: Hmm, it seems to not needed because going backward, event_dim is always guaranteed to greater than `domain.event_dim`. Please ignore my comment.

numpyro/infer/reparam.py
fritzo: Thanks, it took me a while to understand. I just pushed an update to the `._wrap()` method which seems clean 😄

fehiepsi: Thanks for the fix! It looks great!

fehiepsi (approved):

PR #887, opened 2021-01-22
Add transforms' forward_shape and inverse_shape

fehiepsi:
This PR steals @fritzo implementation in https://github.com/pytorch/pytorch/pull/50581 and resolve various issues that couldn't be addressed previously. Thanks for your effort, @fritzo!

### DIFF
--- a/numpyro/contrib/tfp/distributions.py
+++ b/numpyro/contrib/tfp/distributions.py
@@ -87,10 +87,16 @@ class BijectorTransform(Transform):
         return self.bijector.forward_log_det_jacobian(x, self.domain.event_dim)
 
     def forward_shape(self, shape):
-        return self.bijector.forward_event_shape(shape)
+        out_shape = self.bijector.forward_event_shape(shape)
+        in_event_shape = self.bijector.inverse_event_shape(out_shape)
+        batch_shape = shape[:len(shape) - len(in_event_shape)]
+        return batch_shape + out_shape
 
     def inverse_shape(self, shape):
-        return self.bijector.inverse_event_shape(shape)
+        in_shape = self.bijector.inverse_event_shape(shape)
+        out_event_shape = self.bijector.forward_event_shape(in_shape)
+        batch_shape = shape[:len(shape) - len(out_event_shape)]
+        return batch_shape + in_shape
 
 
 @biject_to.register(BijectorConstraint)
--- a/numpyro/contrib/tfp/distributions.py
+++ b/numpyro/contrib/tfp/distributions.py
@@ -86,6 +86,12 @@ class BijectorTransform(Transform):
     def log_abs_det_jacobian(self, x, y, intermediates=None):
         return self.bijector.forward_log_det_jacobian(x, self.domain.event_dim)
 
+    def forward_shape(self, shape):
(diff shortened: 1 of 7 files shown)


numpyro/distributions/distribution.py
```diff
@@ -774,16 +762,10 @@ def __init__(self, base_distribution, transforms, validate_args=None):
...
-        if event_dim <= base_ndim:
-            batch_shape = shape[:base_ndim - event_dim]
-            event_shape = shape[base_ndim - event_dim:]
-        else:
-            event_shape = (-1,) * event_dim
-            batch_shape = ()
+        shape = transform.forward_shape(shape)
+        cut = len(shape) - event_dim
+        batch_shape = shape[:cut]
+        event_shape = shape[cut:]
```
fritzo, line 768: Nice, this is so much simpler that the PyTorch version! I think I'll copy your use of `ComposeTransform` upstream in https://github.com/pytorch/pytorch/pull/50581 😄

fritzo (approved):
Thanks for porting this! My only suggestion is to add a comment about `.forward_shape()` versus `.forward_event_shape()`.

numpyro/distributions/distribution.py
fritzo: @fehiepsi I've made some more changes to the upstream version, including wrapping `base_dist` in `Independent` to ensure `.sample(sample_shape)` is correct. Can you PTAL at https://github.com/pytorch/pytorch/pull/50581?

numpyro/contrib/tfp/distributions.py
```diff
@@ -86,6 +86,12 @@ def _inverse(self, y):
     def log_abs_det_jacobian(self, x, y, intermediates=None):
         return self.bijector.forward_log_det_jacobian(x, self.domain.event_dim)
 
+    def forward_shape(self, shape):
+        return self.bijector.forward_event_shape(shape)
+
+    def inverse_shape(self, shape):
+        return self.bijector.inverse_event_shape(shape)
+
```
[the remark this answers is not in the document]
fehiepsi: You are right. Bijectors like `Affine(zeros(3), ones(3))` do not align with Pyro transforms. How do you know?

I just added a fix for it.

numpyro/distributions/distribution.py
fehiepsi: You are right. We have assumed too much on base distribution. Your change is pretty clean. Thanks!

numpyro/contrib/tfp/distributions.py
```diff
@@ -86,6 +86,18 @@ def _inverse(self, y):
     def log_abs_det_jacobian(self, x, y, intermediates=None):
         return self.bijector.forward_log_det_jacobian(x, self.domain.event_dim)
 
+    def forward_shape(self, shape):
+        out_shape = self.bijector.forward_event_shape(shape)
+        in_event_shape = self.bijector.inverse_event_shape(out_shape)
+        batch_shape = shape[:len(shape) - len(in_event_shape)]
+        return batch_shape + out_shape
```
fritzo, line 93: Nice trick!

fritzo (approved):
Thanks for porting this!

PR #888, opened 2021-01-23
Bump to version 0.5.0

### LINKED ISSUE #838: Broken install under Python 3.9
When I try to install numpyro in a Python 3.9 environment, I get a messed up installation of incompatible packages. The following have been my steps:
```
conda create --name numpyro-py39 python=3.9
conda activate numpyro-py39
pip install numpyro --no-cache
```
The outcome of this are the following installed packages
```
jax==0.1.35
jaxlib==0.1.57
numpyro==0.1.0
```
Now, this version of jaxlib is incompatible with that version of jax and, of course, I would like to have the most recent version of numpyro to work with.

It seems the following is what happens: The jax authors have not provided versions of jaxlib previous to 0.1.57 for Python 3.9 on the package index. The latest version of numpyro (0.4.1) requires jaxlib 0.1.56, which is thus not available. Pip goes through the available versions, not finding possible configuration until it reaches numpyro version 0.1.0, which does not fix a specific version of jaxlib (but is fine with `jaxlib>=0.1.14`, and requires `jax==0.1.35`, which itself does not constrain the jaxlib version).

As a quick fix, would you consider bundling up a 0.4.2 version with updated jax and jaxlib dependencies that installs on Python 3.9?

fehiepsi:
Close #838.

Blocked by
- [x] #887
- [x] #886
- [x] funsor release


numpyro/infer/hmc_gibbs.py
```diff
@@ -365,7 +365,7 @@ def __init__(self, inner_kernel, *, random_walk=False, modified=False):
         super().__init__(inner_kernel, lambda *args: None, None)
         self._random_walk = random_walk
         self._modified = modified
-        self._use_unconstrained_gibbs_fn = True
+        self._use_constrained_gibbs_fn = False
```
fehiepsi, line 368: This is a typo. This flag will be removed in #884.

fritzo (approved):

eb8680 (approved):

fehiepsi:
Thanks, @fritzo @eb8680!

Issue #454, opened 2019-11-21
Create a test suite for performance testing


neerajprad:
To get around perf issues, we have had to put in optimizations in certain parts of the codebase. Sometimes these optimizations add to code complexity and can be removed once the perf issues are resolved (either upstream or within numpyro). 

It will be nice if we had a set of perf tests capturing different kind of models (ideally something that runs within a couple of minutes):
 - Anytime we have to put in a hack in the code / interface to get good performance on a model, we can add a corresponding perf test.
 - Then when the issue upstream is resolved or we feel confident that the hack isn't needed, we can refactor and check against the perf test.

PR #890, opened 2021-01-24
Optimized test

loopylangur:
Hi,

I was able to reduce the running time of `test_functional_beta_bernoulli_x64` test from 37  seconds to about 6 seconds on my local machine by changing some hyper-parameters.

I also ran the test several times to ensure that the test is not flaky.

Environment:
```
python 3.6
jax==0.2.8
numpy==1.19.5
```
Is this something that you guys will be interested in? If yes, then I can help you optimize some other tests in this repo as well. If you have any other suggestions/edits I will be happy to integrate them as well. Please let me know.

Thanks!

### DIFF
--- a/test/test_mcmc.py
+++ b/test/test_mcmc.py
@@ -484,7 +484,7 @@ def test_extra_fields():
 
 @pytest.mark.parametrize('algo', ['HMC', 'NUTS'])
 def test_functional_beta_bernoulli_x64(algo):
-    warmup_steps, num_samples = 500, 20000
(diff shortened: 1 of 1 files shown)


fehiepsi:
It would be very helpful. In the early days, we used high numbers there to benchmark some of the changes. Because things look stable now, we should create a different [test suite](https://github.com/pyro-ppl/numpyro/issues/454) for performance. Please feel free to adjust other slow tests. Thanks, @loopylangur!

fehiepsi (approved):

PR #893, opened 2021-01-26
Move the warning of using the sequential chain method to the constructor

fehiepsi:
Tried to address memory leak as reported by @PaoloRanzi81 in https://github.com/pyro-ppl/numpyro/issues/539 for `chain_method='sequential'` in 1 GPU but haven't found the root problem yet. But I think it would be nice to raise the warning about not enough devices to run parallel method as early as possible.

Also use `jax.local_device_count` instead of `jax.lib.xla_bridge.device_count` to calculate the number of available devices, as mentioned in docs of `pmap`.

### DIFF
--- a/numpyro/infer/mcmc.py
+++ b/numpyro/infer/mcmc.py
@@ -252,6 +252,13 @@ class MCMC(object):
         if chain_method not in ['parallel', 'vectorized', 'sequential']:
             raise ValueError('Only supporting the following methods to draw chains:'
                              ' "sequential", "parallel", or "vectorized"')
+        if chain_method == 'parallel' and xla_bridge.device_count() < self.num_chains:
+            chain_method = 'sequential'
+            warnings.warn('There are not enough devices to run parallel chains: expected {} but got {}.'
+                          ' Chains will be drawn sequentially. If you are running MCMC in CPU,'
+                          ' consider to use `numpyro.set_host_device_count({})` at the beginning'
+                          ' of your program.'
+                          .format(self.num_chains, xla_bridge.device_count(), self.num_chains))
         self.chain_method = chain_method
         self.progress_bar = progress_bar
         # TODO: We should have progress bars (maybe without diagnostics) for num_chains > 1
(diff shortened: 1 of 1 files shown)


numpyro/infer/mcmc.py
```diff
@@ -499,17 +497,12 @@ def run(self, rng_key, *args, extra_fields=(), init_params=None, **kwargs):
...
             states = tree_map(lambda x: x[jnp.newaxis, ...], states_flat)
         else:
-            if chain_method == 'sequential':
-                if self.progress_bar:
-                    states, last_state = _laxmap(partial_map_fn, map_args)
-                else:
-                    states, last_state = lax.map(partial_map_fn, map_args)
-            elif chain_method == 'parallel':
+            if self.chain_method == 'sequential':
+                states, last_state = _laxmap(partial_map_fn, map_args)
```
fehiepsi, line 501: Using `lax.map` here is a bit faster but expensive in term of memory requirement so I switched to for loop. See the benchmark
```python
%%time
%%memit

import jax

import numpyro; numpyro.set_platform("gpu")

def model():
    numpyro.sample("x", dist.Normal(0, 1).expand([100]))

mcmc = MCMC(NUTS(model), 1000, num_samples=10000, progress_bar=False, num_chains=8)
mcmc.run(jax.random.PRNGKey(0))
samples = mcmc.get_samples(group_by_chain=True)["x"]
print(samples.shape)
```
#### using lax.map
```
(8, 10000, 100)
peak memory: 5462.81 MiB, increment: 5415.39 MiB
CPU times: user 9min 23s, sys: 6.47 s, total: 9min 29s
Wall time: 9min 25s
```
while using for loop `_laxmap`:
```
(8, 10000, 100)
peak memory: 1795.53 MiB, increment: 1748.04 MiB
CPU times: user 9min 41s, sys: 5.9 s, total: 9min 47s
Wall time: 9min 44s
```

neerajprad: This seems more than acceptable for the memory savings!

numpyro/infer/mcmc.py
```diff
@@ -252,6 +251,14 @@ def __init__(self,
         if chain_method not in ['parallel', 'vectorized', 'sequential']:
             raise ValueError('Only supporting the following methods to draw chains:'
                              ' "sequential", "parallel", or "vectorized"')
+        if chain_method == 'parallel' and local_device_count() < self.num_chains:
+            chain_method = 'sequential'
+            warnings.warn('There are not enough devices to run parallel chains: expected {} but got {}.'
+                          ' Chains will be drawn sequentially. If you are running MCMC in CPU,'
+                          ' consider to use `numpyro.set_host_device_count({})` at the beginning'
```
neerajprad, line 258: nit: `consider using`...`devices are available`...

neerajprad (approved):

PR #900, opened 2021-01-29
Pbar jitted mcmc

jeremiecoullon:
Here's a first attempt a progress bar for compiled MCMC (based on issue #309). It still has issues (see this [comment](https://github.com/pyro-ppl/numpyro/issues/309#issuecomment-769807184), and probably more).

The main idea is that it uses the module `host_callback` to invoke the python print function so that the progress of a chain within `fori_loop` is printing. Note that this module is still experimental in JAX so it might change, so it might actually not be a good idea!


jeremiecoullon:
And sorry for the broken tests!

jeremiecoullon:
Also, I've been playing around with this over the weekend: it's possible to use tqdm here using `tqdm.update()` in the `_print_consumer()` function, so this can be improved.

fehiepsi:
Thanks for your patient, @jeremiecoullon! This is a really nice feature to have. I'll play with this a bit to understand the behavior and will get back to you soon.

fehiepsi:
Woo, I can confirm that this works for multi chains and seems to not affect the performance!

> it's possible to use tqdm here using tqdm.update()

Yeah, that would be ideal! Please let me know if you need me clarify any numpyro details.

Because this is an experimental feature in JAX, I think we can add an experimental argument `host_callback=False` (by default) to MCMC class and `fori_collect` so everyone can play with this option for a while before we make this a default behavior when `progress_bar=True`.

jeremiecoullon:
So tqdm now seems to work when using 1 chain.

However for multiple chains the progress bar goes on for too long. I think what’s happening is that each chain updates the same progress bar. A solution might be to either define the progress bar to run for `num_samples*num_chains` iterations, or define the print rate to be `usual_print_rate/num_chains`.
I’ve tried this manually and it works, but the `fori_collect` function in `util.py` doesn’t have access to `num_chains` so it can’t do this.

jeremiecoullon:
Another potential issue: the tqdm progress is defined outside JAX and the `fori_loop` then updates it. 
That means that if the `fori_loop` isn't computed straight away (and is only computed after some other code is run) then you get an empty progress bar defined first, and then another progress bar appears when the JAX loop is finally computed.

If this happens in the numpyro sampler (I've only seen this happen elsewhere) then a solution could be to have `host_callback.id_tap` call a python function at iteration 0 of the loop to define a tqdm progress bar. That means it's defined at run time.  
There would still be the issue of each of these python function needing to have access to the tqdm variable (this is currently solved by defining tqdm in a closure). A simple solution is to have the progress bar defined as a global variable, but there might be something better to do.

fehiepsi:
Hi @jeremiecoullon, it seems that we can use `tap_with_device=True` to detect the chain that invokes the id_tap. I agree that id_tap needs to access an outside tqdm object, but we might put it locally inside the body_fn wrapper. Something likes this (based on all of your suggestions)

```python
from jax import lax, jit, pmap
from jax.experimental.host_callback import id_tap, id_print
import jax.numpy as jnp
import numpyro
numpyro.set_host_device_count(4)
from tqdm.auto import tqdm
import time

def power3(chain, x):

    def body_fn(val, i):
        val = val * 2
        tqdm_bars = {}

        def tap_fn(tap, transforms, device):
            i, val = tap
            chain = int(str(device)[-1])
            if i == 0:
                tqdm_bars[chain] = tqdm(total=20, position=chain)
            if i % 2 == 0:
                tqdm_bars[chain].set_description(str(i), refresh=True)
                tqdm_bars[chain].update(2)
            if i == 19:
                tqdm_bars[chain].close()
            return

        id_tap(tap_fn, (i, val + i), tap_with_device=True)
        return val, None

    y, _ = lax.scan(body_fn, x, jnp.arange(20))
    return y

results = pmap(power3)(jnp.arange(4, dtype=jnp.uint32), jnp.arange(4) + 10.)
time.sleep(0.1)
print(results)
```

What do you think?

jeremiecoullon:
@fehiepsi : great idea about using `tap_with_device` I didn't realise we could do that!

I tried your suggestion on numpyro: it does indeed work (see latest commit). But the issue is that when one device finishes on a chain it starts working on another, and therefore adds another progress bar. However when everything is finished you usually end up with the correct number of progress bars (though not always!).

From trying out a few times, it seems that what happens is that if you print something after running `mcmc.run(...)` (such as `print("Done!")` or `mcmc.print_summary()`) you might end up with too many progress bars even after everything's done. But if you don't print anything after you seem to always end up with the correct number of progress bars once everything's finished. This is from trying it out by running a python script with the "8 schools" example. 
I also tried it in a notebook and the printing behaviour is a worse unfortunately!

Any thoughts on what to do here? It essentially works but some UI glitches. I'm guessing that the weird printing behaviour described above is to do with JAX's asynchronous dispatch.

numpyro/util.py
```diff
@@ -166,58 +165,50 @@ def _wrapped(fn):
     return _wrapped
 
 
-def _print_consumer(arg, transrorm):
-    i, n_iter = arg
-    print(f"Iteration {i}/{n_iter}")
-
-
-@jit
```
jeremiecoullon: I just realised that I had left this bit in!

numpyro/util.py
```diff
@@ -164,6 +165,70 @@ def _wrapped(fn):
...
 
+def progress_bar_factory(num_samples):
+    """Factory that builds a progress bar decorator along
+    with the `set_tqdm_description` and `close_tqdm` functions
+    """
+
+    tqdm_bars = {}
+
+    def _define_tqdm(arg, transform, device):
+        chain = int(str(device)[4:])
```
fehiepsi, line 176: yeah, we should do `[4:]`, thanks!

numpyro/util.py
```diff
@@ -164,6 +165,70 @@ def _wrapped(fn):
...
+        message = f"Running chain {chain}"
+        tqdm_bars[chain].set_description(message, refresh=False,)
+
+    def _update_tqdm(arg, transform, device):
+        chain = int(str(device)[4:])
+        tqdm_bars[chain].update(arg)
+
+    def _close_tqdm(arg, transform, device):
+        chain = int(str(device)[4:])
+        tqdm_bars[chain].close()
```
fehiepsi, line 187: what happens if `num_samples = 1010` and `print_rate=50`? will this close at 99%?

numpyro/util.py
```diff
@@ -164,6 +165,70 @@ def _wrapped(fn):
...
+        _ = lax.cond(
+            iter_num % print_rate == 0,
+            lambda _: host_callback.id_tap(_update_tqdm, print_rate, result=iter_num, tap_with_device=True),
+            lambda _: iter_num,
+            operand=None,
+        )
+
+        _ = lax.cond(
+            iter_num == num_samples-1,
+            lambda _: host_callback.id_tap(_close_tqdm, print_rate, result=iter_num, tap_with_device=True),
```
fehiepsi, line 211: I guess we need to update `num_samples % print_rate` here before closing.

numpyro/util.py
```diff
@@ -221,7 +286,9 @@ def _body_fn(i, vals):
 
     collection = jnp.zeros((collection_size,) + init_val_flat.shape)
     if not progbar:
```
fehiepsi, line 288: With `num_chains=1` and `progress_bar=False`, I guess we want to use the old `_body_fn`. You can add an option to `progbar_opts` to control the behavior I guess, e.g. adding `num_chains` to [progbar_opts](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/infer/mcmc.py#L354), removing [this branch](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/infer/mcmc.py#L264), and revising the logic here into
```python
if not progbar:
    last_val, collection, _, _ = fori_loop(0, upper, _body_fn, (init_val, collection, start_idx, thinning))
elif num_chains > 1:
    progress_bar_fori_loop = progress_bar_factory(upper)
    _body_fn_pbar = progress_bar_fori_loop(_body_fn)
    last_val, collection, _, _ = fori_loop(0, upper, _body_fn_pbar, (init_val, collection, start_idx, thinning))
```

numpyro/util.py
```diff
@@ -164,6 +165,70 @@ def _wrapped(fn):
...
+def progress_bar_factory(num_samples):
+    """Factory that builds a progress bar decorator along
+    with the `set_tqdm_description` and `close_tqdm` functions
+    """
+
+    tqdm_bars = {}
+
+    def _define_tqdm(arg, transform, device):
+        chain = int(str(device)[4:])
+        tqdm_bars[chain] = tqdm.tqdm(range(num_samples))
```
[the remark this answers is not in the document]
fehiepsi: By the way, I guess we need to use `from tqdm.auto import tqdm` here. Using `tqdm.tqdm` does not display the bars well in my notebook. But probably it is just my issue.

fehiepsi (commented):
The change looks great, @jeremiecoullon! I don't find a good solution for the `printing` issue but probably adding
```
host_callback.barrier_wait()
```
after `fori_loop` would do the job. I don't know about its side-effect though. :(

numpyro/util.py
```diff
@@ -164,6 +165,70 @@ def _wrapped(fn):
...
+
+    def _update_tqdm(arg, transform, device):
+        chain = int(str(device)[4:])
+        tqdm_bars[chain].update(arg)
+
+    def _close_tqdm(arg, transform, device):
+        chain = int(str(device)[4:])
+        tqdm_bars[chain].close()
+
+    @jit
```
fehiepsi, line 189: you don't need to jit here I guess. `fori_loop` already does the job.

jeremiecoullon:
Thanks for these! I'll have a try and get back to you

numpyro/util.py
```diff
@@ -8,7 +8,7 @@
 import re
 
 import numpy as np
-import tqdm
+from tqdm.auto import tqdm
```
jeremiecoullon, line 11: This completely fixes the issue for notebooks for me; great suggestion @fehiepsi !

numpyro/util.py
```diff
@@ -172,9 +172,16 @@ def progress_bar_factory(num_samples):
 
     tqdm_bars = {}
 
+    if num_samples > 20:
```
jeremiecoullon, line 175: This (along with the change below) fixes 2 bugs:
1. if `num_samples` < 20 you now don't get a modulo zero error.
2. The progress bar now works if 20 doesn't divide the `num_samples` (before it overshot the total number of iterations)

numpyro/util.py
```diff
@@ -200,15 +206,22 @@ def _update_progress_bar(iter_num, print_rate):
...
+            (iter_num % print_rate == 0) & (iter_num != num_samples - remainder),
             lambda _: host_callback.id_tap(_update_tqdm, print_rate, result=iter_num, tap_with_device=True),
             lambda _: iter_num,
             operand=None,
         )
 
         _ = lax.cond(
-            iter_num == num_samples-1,
-            lambda _: host_callback.id_tap(_close_tqdm, print_rate, result=iter_num, tap_with_device=True),
+            iter_num == num_samples - remainder,
```
jeremiecoullon, line 216: This deals with the case of `num_samples` not divisible by 20

numpyro/util.py
fehiepsi: thanks, the logic looks great!

numpyro/util.py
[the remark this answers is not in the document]
jeremiecoullon: I tried adding `host_callback.barrier_wait()` just after the `fori_loop()` in `fori_collect` but the sampler simply doesn't run anymore.. :(. 

The issue seems fixed for notebook (at least when I tried it) though which is good. When running it in the terminal it can still look a bit messy though

numpyro/util.py
jeremiecoullon: Yes, good suggestion! The progress bar actually overshoots rather than undershoots. I fixed this in the latest commit

numpyro/util.py
fehiepsi: Haha, this is a bit different from what I suggested. ;) I meant to only invoke `tqdm.auto` when running multiple chains. I would prefer to have the old progress_bar for [reasons here](https://github.com/pyro-ppl/numpyro/pull/813#issuecomment-726487003).

jeremiecoullon: Ah oops! Ok I'll have a look at that and change it :)

numpyro/util.py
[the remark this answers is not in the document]
fehiepsi: How about adding `host_callback.barrier_wait()` at the end of mcmc.run (when num_chains > 1, chain_method="parallel", and progress_bar=True)? I guess it might do the trick.

numpyro/util.py
jeremiecoullon: Sorry for the delay!

I don't understand what the API should be for the new progress bar.

Should `progress_bar` in `MCMC` still control the old progress bar (for un-compiled loops)? You had suggested in an earlier comment that we could add `host_callback=False` in the MCMC class and pass that to `fori_collect` so that people can test it at first. 

I don’t quite get how `num_chains` should control the behaviour of the compiled progress bar. Do you mean that the progress bar for complied sampling should be on if `num_chains>1`?

numpyro/util.py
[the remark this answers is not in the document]
jeremiecoullon: Sorry for takes ages to get to this!

I don't quite get this bit: `progress_bar=True` is for the progress bar for the un-compiled chain, no?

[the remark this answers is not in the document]
fehiepsi: Yes, that position would work IMO. I think as long as it is outside of `jax.pmap`, it will be fine. Probably, after [this line](https://github.com/pyro-ppl/numpyro/blob/ce6cac87ec715ea467a353160dcb56a1827b28a9/numpyro/infer/mcmc.py#L504) is better...

numpyro/util.py
fehiepsi: Currently, with `num_chains > 1` and `chain_method="parallel"`, we [switch progress_bar to False](https://github.com/pyro-ppl/numpyro/blob/ce6cac87ec715ea467a353160dcb56a1827b28a9/numpyro/infer/mcmc.py#L266) because we don't have support for it yet. And IIUC, your PR tries to address that issue. Back then, I thought you are going to do host callback for every MCMC step, so I wondered about the overhead. That was the reason that I suggested having a new argument to test things out first. For now, with your strategy of updating progressing only after each 5% of the total progress, I think the overhead of using host_callback would be pretty small. So you can go ahead and make it work for the default behavior of `num_chains > 1, progress_bar=True, chain_method="parallel"`.

jeremiecoullon: Unfortunately, I’m still confused about what the API should be.

To summarise, in this PR we currently have the following behaviour:

• `num_chains=1`, `progress_bar=False` -> compiled sampler, and uses the new progress bar 
• `num_chains=1`, `progress_bar=True` -> “python loop” sampler, and uses the old progress bar 
• `num_chains>1`, `progress_bar=False` -> compiled sampler, and uses the new progress bar. 
• `num_chains>1`, `progress_bar=True` -> compiled sampler, and uses the new progress bar. This has the same behaviour as above because of that [line](https://github.com/pyro-ppl/numpyro/blob/ce6cac87ec715ea467a353160dcb56a1827b28a9/numpyro/infer/mcmc.py#L266) in`run()`

So my questions are:
1. Which of these behaviours should change? 
2. It might be good to be able to have the compiled sampler without the progress bar (for both 1 or several chains). How should we do this? The problem is that `progress_bar=True/False` controls that behaviour. Ideally there would be another argument `compiled=True/False` that controls whether it’s compiled. Though I get that changing the API might not be a good idea.

fehiepsi: > num_chains=1, progress_bar=False -> compiled sampler, and uses the new progress bar
> num_chains>1, progress_bar=False -> compiled sampler, and uses the new progress bar.

I think those are unexpected.

> It might be good to be able to have the compiled sampler without the progress bar

I think if users want progress bars, they can set `progress_bar=True`. Otherwise, they can set `progress_bar=False`. We can have a new argument for users to test the case `num_chains>1, progress_bar=False` but as I mentioned above, it seems to me that the overhead is likely small, so we might not need to introduce that new argument.

jeremiecoullon: To be clear on what definitions I using:
- _“compiled”_: the sampler that uses `fori_loop(body_fn)` 
- _“non-compiled”_: the sampler that uses `python_loop(jit(body_fn))`
- _“the new progress bar”_: the tqdm progress bar that is added in this PR. This runs for the compiled sampler
- _“the old progress bar”_ the tqdm progress bar that is in the current master branch of numpyro. This runs for the un-compiled sampler.

----

> I think those are unexpected.

Which bit specifically is unexpected, and what should then be the correct behaviour? 

> Regarding compiled sampler, I guess you meant to display the progress bar even when users set `progress_bar=False`

This is what is currently happening on this branch. See below for my issues with the argument `progress_bar`


> We can have a new argument for users to test the case num_chains>1, progress_bar=False but as I mentioned above, it seems to me that the overhead is likely small, so we might not need to introduce that new argument. What do you think?

I wasn't actually worried about the overhead as I agree that it's very small. I was more thinking about turning off the progress bar for compiled samplers in general (See addition _A1_ below)


> Regarding compiled sampler, I guess you meant to display the progress bar even when users set progress_bar=False, i.e. compiled means fori_loop(body_fn) vs non-compiled means python_loop(jit(body_fn))?

That's indeed what I mean by compiled vs non-compiled (see definitions above). I do think we should have the progress bar for a compiled sampler with 1 chain. I thought that was always the point of this PR. Though from re-reading the original [issue](https://github.com/pyro-ppl/numpyro/issues/309), I realised it was originally about `num_chains>1` (this is probably why I was getting confused :) ). Do you think that the compiled sampler for 1 chain should _not_ display a progress bar?

----

The current behaviour and API of this branch is given by the 4 cases in my [previous comment](https://github.com/pyro-ppl/numpyro/pull/900#discussion_r585462473). I think this behaviour seems good. Note that all samplers (compiled, non-compiled, 1 chain, many chains) have a progress bar.

My thought was that it might be good to have an argument in `MCMC` allowing you to run the compiled sampler without the progress bar (let’s call this addition _A1_). This would apply for both single chain and many chain samplers. This is for when you run MCMC within a larger workflow and want it to be fast but don’t want to see the progress bar. However I get that this is a specific use-case. So perhaps it’s not necessary!

In any case (ie: whether or not we add addition _A1_), the argument `progress_bar` now controls whether or not the sampler is compiled, and not the progress bar. This is the main reason why I suggested added `compiled=True/False` to control whether or not the sampler is compiled. 
However I see your point that `single_chain_compiled` might be better as only a single chain can be non-compiled. And I agree that this might be a confusing message to users!

fehiepsi: > Which bit specifically is unexpected, and what should then be the correct behaviour?

I think the correct behavior is to not have progress bar when setting `progress_bar=False` and having progress bar when setting `progress_bar=True`. Currently, NumPyro does not have progress bar for multiple chains and that is also unexpected. That is why issue #309 is still open for a long time.

> Do you think that the compiled sampler for 1 chain should not display a progress bar?

It depends on the API I guess. With the current API, the compiled sampler for 1 chain (equivalently, progress_bar=False, num_chains=1) should not display a progress bar because `progress_bar=False`. Let's discuss having a different API in another thread. What do you think?

It seems to me that we agree on having progress bar on num_chains>1 is nice. How about creating a new issue/feature request for always displaying progress bar, so we can address it in another PR? I think we need to discuss first if it is necessary to introduce a new argument for that feature. Some specific usage cases, like when MCMC is very fast to sample, a lot of samples are required while users only have 1 device to run a single chain, and users want to see the 5%-bin progress, might be important. I guess the dispatching cost of the current `progress_bar=True` approach (for single chain, in master branch) will be large in that case but it is not currently clear to me why we need to display progress bar if it is so fast to sample, so would like to have more discussion for that usage. There might be other usage cases that I overlooked, so having a separate thread would be very helpful too. :)

jeremiecoullon: Ah I think I understand now, thanks for explaining!

So I'll make some fixes to make sure the `progress_bar=True/False` maps to whether there is a progress bar. 

And I agree that a separate PR would be good to discuss whether to always show the progress bar (ex: for 1 compiled chain)

fehiepsi: Thank you, @jeremiecoullon! Sorry for making confusion. I'm happy that the path ahead is clear now. :)

numpyro/util.py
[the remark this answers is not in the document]
jeremiecoullon: I tried added it where you suggested and the extra progress bars still appear (when running a script in the terminal). Do you get the same behaviour?

numpyro/util.py
jeremiecoullon: I just changed this in the latest pull request: `tqdm.auto` is now only used for `num_chains>1`. For `num_chains=1`: the usual tqdm progress bar is used (on the non-compiled sampler).

Is this what you meant?

numpyro/util.py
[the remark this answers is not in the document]
fehiepsi: It seems that the following code also works in my system
```python
def progress_bar_factory(num_samples, num_chains):
    ...
    finished_chains = []
    for chain in range(num_chains):
        tqdm_bars[chain] = tqdm_auto(range(num_samples), position=chain)
        message = f"Running chain {chain}"
        tqdm_bars[chain].set_description(message, refresh=False,)
        tqdm_bars[chain].update(0)

    def _close_tqdm(arg, transform, device):
        chain = int(str(device)[4:])
        tqdm_bars[chain].update(arg)
        finished_chains.append(chain)
        if len(finished_chains) == num_chains:
            for chain in range(num_chains):
                tqdm_bars[chain].close()

    def _update_progress_bar(iter_num):
        _ = lax.cond(
            iter_num % print_rate == 0,
            lambda _: host_callback.id_tap(_update_tqdm, print_rate, result=iter_num, tap_with_device=True),
            lambda _: iter_num,
            operand=None,
        )
        _ = lax.cond(
            iter_num == num_samples,
            lambda _: host_callback.id_tap(_close_tqdm, remainder, result=iter_num, tap_with_device=True),
            lambda _: iter_num,
            operand=None,
        )

    def progress_bar_fori_loop(func):
        def wrapper_progress_bar(i, vals):
            result = func(i, vals)
            _update_progress_bar(i + 1)
            return result
```
The changes that I made are:
+ start the chains in order: this is achieved by moving `_define_tqdm` out of the fori loop. Without this, the progress bars do not appear in order.
+ close the chains in order: this is achieved by adding a variable `finished_chains` to record which chains have been finished. After all chains are finished, we close all progress bars. Without this, in the terminal, the progress bars swap together in an arbitrary order (probably the chain that finished first will be on the top - I don't know). 
+ report the progress only after the current iteration finished its job. This is a bit different from the current behavior in your code, where progress bars are updated at positions 0, print_rate, 2 * print_rate (while I think it should be print_rate, 2*print_rate,...). Can we discuss a bit more on what is a better behavior here?

numpyro/util.py
fehiepsi: Thanks, this looks correct to me.

numpyro/infer/mcmc.py
```diff
@@ -351,7 +349,8 @@ def _single_chain_mcmc(self, init, args, kwargs, collect_fields):
                                     thinning=self.thinning,
                                     collection_size=collection_size,
                                     progbar_desc=partial(_get_progbar_desc_str, lower_idx, phase),
-                                    diagnostics_fn=diagnostics)
+                                    diagnostics_fn=diagnostics,
+                                    num_chains=self.num_chains)
```
fehiepsi, line 353: could you change this to: `num_chains=self.num_chains if self.chain_method == "parallel" else 1` (if you prefer clarity, we can change `num_chains` to `num_parallel_chains`)

jeremiecoullon: done! (in the recent commit)

numpyro/util.py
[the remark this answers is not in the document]
jeremiecoullon: I added the code you suggested:
- Updating the progress bar after running the iterations makes sense!
- On my system: your fix means that the chains now appear in order. I also now don’t have an extra progress bar appearing in the middle of a run and then disappearing.
- I tested different number of samples/chains/warmup: this seems to work fine
- The `barrier_wait()` is not needed on my system when running your MH example. Which bits of texts were badly positioned for you?
- For me the only small point left is: The chains now start at 0% and wait for the sampler to compile before running. Perhaps a message `print(“Compiling..”)` at the beginning of the progress bar would be useful to make it clear that everything is working fine. However this is not really a problem to be honest, and perhaps printing too much stuff isn't great either.

> Could you provide the script so I can test it in my system?

I've been using the 8 schools example to test my code throughout this PR:

```
import numpy as np
import time

J = 8
y = np.array([28.0, 8.0, -3.0, 7.0, -1.0, 1.0, 18.0, 12.0])
sigma = np.array([15.0, 10.0, 16.0, 11.0, 9.0, 11.0, 10.0, 18.0])

import numpyro
numpyro.set_host_device_count(4)

def eight_schools(J, sigma, y=None):
    mu = numpyro.sample('mu', dist.Normal(0, 5))
    tau = numpyro.sample('tau', dist.HalfCauchy(5))
    with numpyro.plate('J', J):
        theta = numpyro.sample('theta', dist.Normal(mu, tau))
        numpyro.sample('obs', dist.Normal(theta, sigma), obs=y)


from jax import random

print("Define sampler")
nuts_kernel = NUTS(eight_schools)

mcmc = MCMC(nuts_kernel, num_warmup=502, num_samples=100000, progress_bar=True, num_chains=4)
rng_key = random.PRNGKey(0)

starttime = time.time()
mcmc.run(rng_key, J, sigma, y=y, extra_fields=('potential_energy',))
endtime = time.time()
print(f"Running time: {endtime-starttime:.1f}sec")


mcmc.print_summary()
```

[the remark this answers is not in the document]
fehiepsi: > Which bits of texts were badly positioned for you?

Nevermind, in my system, sometimes the line `abccc` appears in the middle of the bars (in terminal). It is pretty rare and only happens for very fast `mcmc.run`. I would prefer to skip adding `barrier_wait`, which I don't fully understand its effect.

> The chains now start at 0% and wait for the sampler to compile before running

Understood. We can start the bar with "Compiling..." (also sample the first step) and switch to the current sampling text when `i == 1`. Please feel free to adjust this if you want.

numpyro/infer/mcmc.py
fehiepsi: thanks!

fehiepsi (approved):
Awesome work, @jeremiecoullon! The failing test is irrelevant. Could you help me change the tolerance [here](https://github.com/pyro-ppl/numpyro/blob/master/test/infer/test_reparam.py#L127) to `assert_allclose(pe_transformed, pe - log_det_jacobian, rtol=2e-7)` to make CI happy? Thanks!

numpyro/util.py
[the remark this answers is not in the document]
jeremiecoullon: I added the "Compiling.." messages in the tqdm progress bars which then get overridden when sampling starts. Is this what you meant?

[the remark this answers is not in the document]
fehiepsi: Sorry, I meant to add
```python
    def _update_tqdm(arg, transform, device):
        chain = int(str(device)[4:])
        tqdm_bars[chain].set_description(f"Running chain {chain}", refresh=False,)
        tqdm_bars[chain].update(arg)

    def _update_progress_bar(iter_num):
        _ = lax.cond(
            iter_num == 1,
            lambda _: host_callback.id_tap(_update_tqdm, 0, result=iter_num, tap_with_device=True),
            lambda _: iter_num,
            operand=None,
        )
```
and starting with `Compiling..` as you did. It seems that `python examples/hmm.py --num-chains 4` gives expected behavior.

fehiepsi:
It seems that we are getting conflicting files. Could you merge with master?

jeremiecoullon:
> It seems that we are getting conflicting files. Could you merge with master?

Ah yeah oops! I had forgotten to update my branch so there was a bunch of stuff that's been updated since I started (including the `rtol=2e-7` bit you mentioned earlier!)

numpyro/util.py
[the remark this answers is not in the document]
jeremiecoullon: I tried this solution and it seems that both methods give the same behaviour (for both the hmm.py example and the 8 schools example). 

Is this the same for you? And do you have a preference for which one we use?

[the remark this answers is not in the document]
fehiepsi: IIUC this will keep Compiling.. until 5% samples are collected. I think it is better to keep Compiling.. until the first sample is collected. WDYT?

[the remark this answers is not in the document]
jeremiecoullon: Ah yeah you're right!
I just fixed it; the message now appears at the start of sampling.

PR #904, opened 2021-02-05
Mention that progress_bar=False can require more memory than progress_bar=True

### LINKED ISSUE #899: OOM error when using "sequential" chain method on a GPU
As reported by @PaoloRanzi81 in https://github.com/pyro-ppl/numpyro/issues/539, RAM might not free its resources after a chain finished its run, which leads to OOM for complicated models trained in hours.

Some observations so far:
+ This seems to not happen on simple models
+ This seems to not happen in CPU and TPU
+ This seems to not happen when sampling in short period of time (says less than 1 hour)
+ This happens when `progress_bar=False`

In progress of testing
+ The default behavior `progress_bar=True`
+ Using only one chain with large number of samples
+ Simplifying the model to isolate which operator, handler,... causes the issue

fehiepsi:
Resolves #899.

### DIFF
--- a/numpyro/infer/mcmc.py
+++ b/numpyro/infer/mcmc.py
@@ -200,7 +200,8 @@ class MCMC(object):
 
     .. note:: `chain_method` is an experimental arg, which might be removed in a future version.
 
-    .. note:: Setting `progress_bar=False` will improve the speed for many cases.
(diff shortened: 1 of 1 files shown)


OlaRonning (approved):
LGTM

martinjankowiak (approved):

PR #908, opened 2021-02-08
Fix DiscreteHMCGibbs to work with multiple chains

### LINKED ISSUE #897: Raise errors when there are no gibbs sites in DiscreteHMCGibbs and HMCECS
Currently, if there is no discrete latent site, DiscreteHMCGibbs will throw `OverflowError: cannot convert float infinity to integer` (see #889), which is misleading. We should raise a better error message when there is no gibbs sites in those classes.

### LINKED ISSUE #907: UnexpectedTracerError with DiscreteHMCGibbs, a latent discrete variable with more than 1 dim, and num_chains>1
The combination of 
1) A latent discrete variable with event_shape > 1x
2) num_chains>1

Throws the following error
#UnexpectedTracerError: Encountered an unexpected tracer. Perhaps this tracer escaped through global state from a previously traced function.
#The functions being transformed should not save traced values to global state. Detail: tracer created on line /home/skynet3/numpyro/numpyro/infer/hmc_gibbs.py:356 (<dictcomp>).

Minimally reproducible example
```
import os
import numpyro
import jax
from jax import random
import jax.numpy as jnp
from numpyro.infer import DiscreteHMCGibbs, MCMC, NUTS
numpyro.set_platform("gpu") 
numpyro.set_host_device_count(2) 
print(jax.__version__) #0.2.8
print(numpyro.__version__) #0.5.0
print(jax.config.FLAGS.jax_backend_target) #local
print(jax.lib.xla_bridge.get_backend().platform) #gpu

#Fails
#UnexpectedTracerError: Encountered an unexpected tracer. Perhaps this tracer escaped through global state from a previously traced function.
#The functions being transformed should not save traced values to global state. Detail: tracer created on line /home/skynet3/numpyro/numpyro/infer/hmc_gibbs.py:356 (<dictcomp>).
def model():
  C = numpyro.sample("C", dist.Bernoulli(probs=0.5 * jnp.ones(5 )  ))
kernel = DiscreteHMCGibbs(NUTS(model), modified=True)
mcmc   = MCMC(kernel, num_warmup=100, num_samples=100, num_chains=2)
mcmc.warmup(random.PRNGKey(0))
mcmc.run(random.PRNGKey(1))
```

The error does not appear if the discrete variable is size 1 or if the number of chains=1. Only the combination of both throws the error

```
#Works
def model():
  C = numpyro.sample("C", dist.Bernoulli(probs=0.5  ))
kernel = DiscreteHMCGibbs(NUTS(model), modified=True)
mcmc   = MCMC(kernel, num_warmup=100, num_samples=100, num_chains=2)
mcmc.warmup(random.PRNGKey(0))
mcmc.run(random.PRNGKey(1))

#Works
def model():
  C = numpyro.sample("C", dist.Bernoulli(probs=0.5 * jnp.ones(5 )  ))
kernel = DiscreteHMCGibbs(NUTS(model), modified=True)
mcmc   = MCMC(kernel, num_warmup=100, num_samples=100, num_chains=1)
mcmc.warmup(random.PRNGKey(0))
mcmc.run(random.PRNGKey(1))
```

fehiepsi:
Fixes #907. The issue is when we `pmap` the `DiscreteHMCGibbs.init` method, with `jnp.broadcast_to`, the `support_sizes` will be a tracer, rather than a concrete array. It raises some errors as reported in #907.

Also fixes #897 to raise errors when there are no gibbs sites.

### DIFF
--- a/numpyro/infer/hmc_util.py
+++ b/numpyro/infer/hmc_util.py
@@ -415,7 +415,9 @@ def warmup_adapter(num_adapt_steps, find_reasonable_step_size=None,
 
         if adapt_step_size:
             step_size = find_reasonable_step_size(step_size, inverse_mass_matrix, z_info, rng_key_ss)
-            ss_state = ss_init(jnp.log(10 * step_size))
+            # NB: when step_size is large, say 1e38, jnp.log(10 * step_size) will be inf
+            # and jnp.log(10) + jnp.log(step_size) will be finite
+            ss_state = ss_init(jnp.log(10) + jnp.log(step_size))
 
         return HMCAdaptState(step_size, inverse_mass_matrix, mass_matrix_sqrt,
                              ss_state, mm_state, window_idx, rng_key)
@@ -475,6 +477,8 @@ def _is_turning(inverse_mass_matrix, r_left, r_right, r_sum):
     elif inverse_mass_matrix.ndim == 1:
         v_left = jnp.multiply(inverse_mass_matrix, r_left)
         v_right = jnp.multiply(inverse_mass_matrix, r_right)
+    else:
+        raise ValueError("invese_mass_matrix should have 1 or 2 dimensions.")
 
     # This implements dynamic termination criterion (ref [2], section A.4.2).
     r_sum = r_sum - (r_left + r_right) / 2
@@ -737,6 +741,8 @@ def euclidean_kinetic_energy(inverse_mass_matrix, r):
         v = jnp.matmul(inverse_mass_matrix, r)
     elif inverse_mass_matrix.ndim == 1:
         v = jnp.multiply(inverse_mass_matrix, r)
+    else:
+        raise ValueError("invese_mass_matrix should have 1 or 2 dimensions.")
 
     return 0.5 * jnp.dot(v, r)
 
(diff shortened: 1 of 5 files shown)


numpyro/infer/hmc_util.py
```diff
@@ -475,6 +477,8 @@ def _is_turning(inverse_mass_matrix, r_left, r_right, r_sum):
     elif inverse_mass_matrix.ndim == 1:
         v_left = jnp.multiply(inverse_mass_matrix, r_left)
         v_right = jnp.multiply(inverse_mass_matrix, r_right)
+    else:
+        raise ValueError("invese_mass_matrix should have 1 or 2 dimensions.")
```
martinjankowiak, line 481: typo

numpyro/infer/hmc_util.py
```diff
@@ -737,6 +741,8 @@ def euclidean_kinetic_energy(inverse_mass_matrix, r):
         v = jnp.matmul(inverse_mass_matrix, r)
     elif inverse_mass_matrix.ndim == 1:
         v = jnp.multiply(inverse_mass_matrix, r)
+    else:
+        raise ValueError("invese_mass_matrix should have 1 or 2 dimensions.")
```
martinjankowiak, line 745: ditto here and below

martinjankowiak (approved):

fehiepsi:
Thanks for reviewing, @martinjankowiak!

martinjankowiak (approved):

PR #909, opened 2021-02-10
Enabling sampling with intermediates for `ExpandedDistribution`

lumip:
Currently, `ExpandedDistribution` blocks intermediate values from the base distribution in `sample_with_intermediates`, i.e., intermediate values cannot be retrieved from sampling and traces when `ExpandedDistribution` and, by extension, the `plate` primitive is used.
The pull request implements `ExpandedDistribution.sample_with_intermediates` under the assumption that batch_shape of intermediate values expands in the same way as for the main sampled value.

As of yet, there are not additional tests to verify correctness of the intermediates, though.

### DIFF
--- a/numpyro/distributions/distribution.py
+++ b/numpyro/distributions/distribution.py
@@ -447,18 +447,38 @@ class ExpandedDistribution(Distribution):
         interstitial_sizes = tuple(self._interstitial_sizes.values())
         expanded_sizes = tuple(self._expanded_sizes.values())
         batch_shape = expanded_sizes + interstitial_sizes
-        samples = sample_fn(key, sample_shape=sample_shape + batch_shape)
-        return samples.reshape(sample_shape + self.batch_shape + self.event_shape)
+        samples, intermediates = sample_fn(key, sample_shape=sample_shape + batch_shape)
+
+        def reshape_sample(x):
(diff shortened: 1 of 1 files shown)


numpyro/distributions/distribution.py
```diff
@@ -444,28 +444,41 @@ def has_rsample(self):
...
-            samples = jnp.swapaxes(samples, dim1, dim2)
-        return samples.reshape(sample_shape + self.batch_shape + self.event_shape)
+        samples, intermediates = sample_fn(key, sample_shape=sample_shape + batch_shape)
+
+        def reshape_sample(x):
+            """ Reshapes samples and intermediates to ensure that the output
+                shape is correct: This implicitly replaces the interstitial dims
+                of size 1 in the original batch_shape of base_dist with those
+                in the expanded dims. While it somewhat 'shuffles' over batch
+                dimensions, we don't care because they are considered independent."""
```
fehiepsi, line 457: Makes sense to me. The change looks great but let me think a bit more to see if I miss something.

fehiepsi (approved):
Thanks, @lumip! The change looks great. I believe this would be pretty useful for costly transformed distributions under `plate`.

Issue #906, opened 2021-02-06
Stan like constraints


fehiepsi:
Move the issue https://github.com/pyro-ppl/pyro/issues/2756 to here because the problem has been resolved there. Currently, NumPyro don't those constraints
- [x] #910 sphere - probably this is the only continuous constraint so far that does not have a bijective transform. We might need to raise errors if users use `sphere` in ImproperUniform distribution.
- [x] #913 positive_ordered_vector
- [x] #913 rename `InvCholeskyTransform` to `CholeskyTransform` and `CorrMatrixCholeskyTransform`

fritzo:
I'm working on `sphere` (with `ProjectedNormal` and `ProjectedNormalReparam`)

fehiepsi:
All missing constraints have been implemented. Thanks for addressing this issue, @fritzo!

PR #910, opened 2021-02-10
Add constraints.sphere, dist.ProjectedNormal, ProjectedNormalReparam

fritzo:
Ports https://github.com/pyro-ppl/pyro/pull/2736
Addresses #906

## Tested
- [x] constraint tests
- [x] distribution tests
- [x] gof tests
- [x] reparam tests


numpyro/distributions/continuous.py
```diff
@@ -1320,3 +1320,118 @@ def tree_flatten(self):
     @classmethod
     def tree_unflatten(cls, aux_data, params):
         return cls(batch_shape=aux_data)
+
+
+class ProjectedNormal(TorchDistribution):
```
fehiepsi, line 1325: Probably `directional.py` is a better place for this distribution. Just thinking about this today, this is a nice distribution to have and the approach here looks very reasonable to me.

Just a note for myself: tell users to use this distribution and reparam (probably registering `biject_to` for `sphere`, but raise an error there) instead of using ImproperUniform. `concentration=0` means uniform over the sphere. No need to have `scale` parameter because, under projection, the location of `concentration` is enough to determine the distribution over sphere.

test/test_distributions.py
```diff
@@ -187,6 +187,10 @@ def sample(self, key, sample_shape=()):
     T(dist.VonMises, 2., 10.),
     T(dist.VonMises, 2., jnp.array([150., 10.])),
     T(dist.VonMises, jnp.array([1 / 3 * jnp.pi, -1.]), jnp.array([20., 30.])),
+    T(dist.ProjectedNormal, jnp.array([0., 0.])),
+    T(dist.ProjectedNormal, jnp.array([2., 3.])),
+    T(dist.ProjectedNormal, jnp.array([0., 0., 0.])),
+    T(dist.ProjectedNormal, jnp.array([-1., 2., 3.])),
```
fehiepsi, line 193: It will be better if we convert a test for (2,) shape and a test for (3,) shape to (1, 2) and (1, 3) shapes.

numpyro/infer/reparam.py
```diff
@@ -145,6 +145,30 @@ def __call__(self, name, fn, obs):
...
+
+    This reparameterization works only for latent variables, not likelihoods.
+    """
+    def __call__(self, name, fn, obs):
+        assert obs is None, "ProjectedNormalReparam does not support observe statements"
+        fn, batch_shape, event_dim = self._unwrap(fn)
+        assert isinstance(fn, dist.ProjectedNormal)
+
+        # Draw parameter-free noise.
+        new_fn = dist.Normal(jnp.zeros(fn.concentration.shape), 1)
```
fehiepsi, line 161: +1 (it is easy to have a bug when the base dist has an incorrect event shape)

fehiepsi (commented):
Looks great to me! I just have a few nits below. Thanks, @fritzo!

numpyro/distributions/directional.py
```diff
@@ -59,3 +60,122 @@ def variance(self):
...
+
+class ProjectedNormal(Distribution):
+    """
+    Projected isotropic normal distribution of arbitrary dimension.
+
+    This distribution over directional data is qualitatively similar to the von
+    Mises and von Mises-Fisher distributions, but permits tractable variational
+    inference via reparametrized gradients.
+
+    To use this distribution with autoguides, use ``poutine.reparam`` with a
```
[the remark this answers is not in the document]
fehiepsi: `autoguides` -> `autoguides` and `HMC`

numpyro/distributions/directional.py
```diff
@@ -59,3 +60,122 @@ def variance(self):
...
+    r2 = _dot(concentration, concentration) - t2
+    perp_part = (-0.5) * r2 - 0.5 * math.log(2 * math.pi)
+
+    # This is the log of a definite integral, computed by mathematica:
+    # Integrate[x/(E^((x-t)^2/2) Sqrt[2 Pi]), {x, 0, Infinity}]
+    # = (t + Sqrt[2/Pi]/E^(t^2/2) + t Erf[t/Sqrt[2]])/2
+    para_part = jnp.log((jnp.exp((-0.5) * t2) * ((2 / math.pi) ** 0.5)
+                         + t * (1 + erf(t * 0.5 ** 0.5))) / 2)
+
+    return para_part + perp_part
```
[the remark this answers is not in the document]
fritzo: Yeah, this tricky math was my main motivation for adding `gof`, and indeed it took me a few tries to get the formulae correct 😄

fritzo:
Thanks for reviewing, @fehiepsi !

fehiepsi (approved):

Issue #914, opened 2021-02-12
[FR] Two-sided TruncatedNormal distribution


njjiang:
Currently the TruncatedNormal distribution only supports setting the low bound, and adding the high bound is in [a TODO comment](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/distributions/continuous.py#L1155). Is there a plan to add this soon?

fehiepsi:
Hi @njjiang, please follow-up with #895. Looking like it is a useful feature so I'll set a higher priority for that issue.

Issue #894, opened 2021-01-26
Support icdf and cdf methods in current distributions


fehiepsi:
It would be nice to have `icdf` and `cdf` methods implemented in some popular NumPyro distributions. There are several motivations for this
+ `icdf` will be useful in nested sampling
+ `cdf` will be useful for truncated distributions
+ some [models](https://forum.pyro.ai/t/beta-cdf-in-numpyro/2520) require `cdf` method

PR #915, opened 2021-02-12
Add one-sided/two-sided truncated distribution and cdf/icdf method for univariate symmetric distributions

### LINKED ISSUE #895: Support for TruncatedDistribution
Currently, we have one-side truncated normal and cauchy implemented. But it would be nice to add a general class, which uses `cdf` method, for other distributions. There is a [nice discussion here](https://github.com/probtorch/pytorch/pull/121) for context on relevant numerical issues.

cc @fritzo

fehiepsi:
Resolves #895, #894, #914. The implementation only supports symmetric distributions because we can leverage the symmetry to address numerical issues that are discussed at [probtorch](https://github.com/probtorch/pytorch/pull/121) for the lower bound > 5. The issue is cdf/icdf suffers from precision errors at the right tail (e.g. `Normal().cdf(6.) = 1.`). Thanks to the symmetry, we can transform the right tail to the left tail, which has better precision (e.g. `Normal().cdf(-6.) = 9.865896e-10`).

### TODO
- [x] Add student T cdf/icdf
- [x] Support truncated from above
- [x] Add tests for cdf/icdf
- [x] Add tests for TruncatedDistribution
- [x] Make pytree test pass

### DIFF
--- a/numpyro/distributions/continuous.py
+++ b/numpyro/distributions/continuous.py
@@ -117,7 +117,7 @@ class Cauchy(Distribution):
         return jnp.arctan(scaled) / jnp.pi + 0.5
 
     def icdf(self, q):
-        return self.loc + self.scale * jnp.tan(jnp.pi(q - 0.5))
+        return self.loc + self.scale * jnp.tan(jnp.pi * (q - 0.5))
 
 
 class Dirichlet(Distribution):
@@ -428,7 +428,7 @@ class Laplace(Distribution):
 
     def icdf(self, q):
         a = q - 0.5
-        return self.loc - self.scale * self.sign(a) * jnp.log1p(-2 * jnp.abs(a))
+        return self.loc - self.scale * jnp.sign(a) * jnp.log1p(-2 * jnp.abs(a))
 
 
 class LKJ(TransformedDistribution):
@@ -1082,9 +1082,9 @@ class StudentT(Distribution):
         return 0.5 * (1 + jnp.sign(scaled) * (1 - betainc(0.5 * self.df, 0.5, beta_value)))
 
     def icdf(self, q):
+        # scipy.special.betaincinv is not avaiable yet in JAX
         # upstream issue: https://github.com/google/jax/issues/2399
-        raise NotImplementedError("Not implemented until scipy.special.betaincinv is"
-                                  " avaiable in JAX.")
+        raise NotImplementedError
 
 
 class LeftTruncatedDistribution(Distribution):
@@ -1096,7 +1096,8 @@ class LeftTruncatedDistribution(Distribution):
         assert base_dist.support is constraints.real, \
             "The base distribution should be univariate and have real support."
         batch_shape = lax.broadcast_shapes(base_dist.batch_shape, jnp.shape(low))
-        self.low = promote_shapes(low, batch_shape)
+        self.base_dist = base_dist
+        self.low, = promote_shapes(low, shape=batch_shape)
         self._support = constraints.greater_than(low)
         super().__init__(batch_shape, validate_args=validate_args)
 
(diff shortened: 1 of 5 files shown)


fehiepsi:
@fritzo Could you help me review this PR? Pls consider this as a low priority PR. A nice thing is your `gof` helps me fix several mistakes in the `sample` method. ;)

numpyro/distributions/continuous.py
```diff
@@ -1049,6 +1071,240 @@ def variance(self):
...
+        base_dist = base_cls.tree_unflatten(base_aux, base_flatten)
+        return cls(base_dist, high=high)
+
+
+class TwoSidedTruncatedDistribution(Distribution):
+    arg_constraints = {"low": constraints.dependent, "high": constraints.dependent}
+    reparametrized_params = ["low", "high"]
+
+    def __init__(self, base_dist, low=0., high=1., validate_args=None):
+        assert isinstance(base_dist, (Cauchy, Laplace, Logistic, Normal, StudentT))
```
fritzo, line 1210: Consider declaring types in a class-level tuple
```py
class TwoSidedTrincatedDistribution(Distribution):
    supported_types = (Cauchy, Laplace, Logistic, Normal, StudentT)
    def __init__(...):
        assert isinstance(base_dist, self.supported_types)
        ...
```
That way users can extend with custom distributions
```py
class MySymmetricDistribution(Distribution):
    def cdf(self, value): ...
    def icdf(self, value): ...

TwoSidedTruncatedDistribution.supported_types += (MySymmetricDistribution,)
```

test/test_distributions.py
```diff
@@ -519,6 +548,28 @@ def test_log_prob(jax_dist, sp_dist, params, prepend_shape, jit):
     assert_allclose(jit_fn(jax_dist.log_prob)(samples), expected, atol=1e-5)
 
 
+@pytest.mark.parametrize('jax_dist, sp_dist, params', CONTINUOUS)
+def test_cdf_and_icdf(jax_dist, sp_dist, params):
+    if not sp_dist:
+        pytest.skip('no corresponding scipy distn.')
```
fritzo, line 554: Are all the `.cdf()` and `.icdf()` methods actually tested? You might want to add some simple algebraic tests that are independent of scipy:
- `d.cdf(d.icdf(u)) = u` for `u ~ uniform(0,1)`
- `d.icdf(d.cdf(x)) = x` for `x ~ d`
- `d/dx d.cdf(x) = d.log_prob(x).exp()` for `x ~ d`
- `d/du d.icdf(u) = 1 / d.log_prob(d.icdf(u)).exp()` for `u ~ uniform(0,1)`

numpyro/distributions/continuous.py
```diff
@@ -1049,6 +1071,240 @@ def variance(self):
...
+        loc = self.base_dist.loc
+        sign = jnp.where(loc >= self.low, 1., -1.)
+        return self.base_dist.cdf(loc - sign * (loc - self.low))
+
+    @lazy_property
+    def _tail_prob_at_high(self):
+        # if low < loc, returns cdf(high) = 1; otherwise returns 1 - cdf(high) = 0
+        loc = self.base_dist.loc
+        sign = jnp.where(loc >= self.low, 1., -1.)
+        return 0.5 * (1 + sign)
```
fritzo, line 1120: why not simply
```py
return jnp.where(self.low < self.base_dist.loc, 1., 0.)
```

numpyro/distributions/continuous.py
```diff
@@ -1049,6 +1071,240 @@ def variance(self):
...
+        # if low < loc, returns cdf(high); otherwise returns 1 - cdf(high)
+        loc = self.base_dist.loc
+        sign = jnp.where(loc >= self.low, 1., -1.)
+        return self.base_dist.cdf(loc - sign * (loc - self.high))
+
+    def sample(self, key, sample_shape=()):
+        assert is_prng_key(key)
+        u = random.uniform(key, sample_shape + self.batch_shape)
+
+        # NB: we use a more numerical formula for a symmetric base distribution
```
fritzo, line 1242: nit: "more numerical" -> "more numerically stable"

fritzo (commented):
Interface looks great!  I haven't checked math of `.cdf()`,`.icdf()` but I would feel more confident if we added some simple algebraic tests independent of SciPy, testing with a large grid of random values.

numpyro/distributions/continuous.py
fehiepsi: yeah, thanks

test/test_distributions.py
fehiepsi: Thanks, those tests are nice! I just added them.

> Are all the .cdf() and .icdf() methods actually tested?

Yes, the test exercises all implemented methods.

test/test_distributions.py
```diff
@@ -519,6 +548,40 @@ def test_log_prob(jax_dist, sp_dist, params, prepend_shape, jit):
...
+@pytest.mark.parametrize('jax_dist, sp_dist, params', CONTINUOUS)
+def test_cdf_and_icdf(jax_dist, sp_dist, params):
+    d = jax_dist(*params)
+    if d.event_dim > 0:
+        pytest.skip('skip testing cdf/icdf methods of multivariate distributions')
+    samples = d.sample(key=random.PRNGKey(0), sample_shape=(100,))
+    quantiles = random.uniform(random.PRNGKey(1), (100,) + d.shape())
+    try:
+        if d.shape() == ():
+            rtol = 1e-3 if jax_dist is dist.StudentT else 1e-5
```
fehiepsi, line 560: `rtol=1e-5` fails for StudentT. I guess the precision of betainc/its grad is not good enough or samples got extreme values.

fritzo (approved):

PR #916, opened 2021-02-12
Add goodness of fit helpers for testing distributions

fritzo:
Addresses https://github.com/pyro-ppl/pyro/issues/2658
Ports https://github.com/pyro-ppl/pyro/pull/2738 (this PR is nearly identical, only differing in `torch`/`numpy` syntax)

This adds a new `numpyro.distributions.gof` module and uses it in some new distribution tests.

I've changed `update_headers.py` to match the version in funsor, which is the most up-to-date.

## Tested
- [x] added a `test_gof()` to test_distributions for continuous distributions

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,8 +2,12 @@ all: test
 
 lint: FORCE
 	flake8
+	python scripts/update_headers.py --check
 
-format: FORCE
+license: FORCE
+	python scripts/update_headers.py
+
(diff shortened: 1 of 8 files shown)


fehiepsi (approved):
Looks great to me! Thanks, @fritzo! I'll use `gof` to test the truncated implementation. :)

PR #917, opened 2021-02-14
Example of how to use mask

RaulPL:
I added a simple example in the documentation of how to use mask in order to select which data to use in the likelihood. Let me know what you think.


fehiepsi (approved):
Thank you, @RaulPL! The example looks great to me!

PR #918, opened 2021-02-16
Fix small typo

RaulPL:
Fix small typo


fehiepsi (approved):
Thanks for the fix, @RaulPL!

PR #919, opened 2021-02-16
adding Softplus transform

vitkl:
Hi

This PR just adds Softplus transform class without any changes to the registry


fritzo (changes requested):
Thanks for splitting this out! Could you 
- add tests to test/test_distributions.py
- add docs to docs/source/distributions.rst

vitkl:
Just realised - is it better to port complex TFP-jax softplus transform rather than adding this?
https://www.tensorflow.org/probability/api_docs/python/tfp/substrates/jax/bijectors/Softplus

fritzo:
@vitkl I think this is simple enough to create independently. Also tfp bijectors have a slightly different shape interface, so it would be good to have this new version.

numpyro/distributions/transforms.py
```diff
@@ -450,6 +450,34 @@ def _inverse(self, y):
...
+            return constraints.interval(self.__call__(self.domain.lower_bound),
+                                        self.__call__(self.domain.upper_bound))
+        else:
+            raise NotImplementedError
+
+    def __call__(self, x):
+        return softplus(x)
+
+    def inv(self, x):
+        return jnp.log(jnp.expm1(x))
```
fehiepsi, line 477: A [more stable version](https://github.com/pytorch/pytorch/pull/52300#discussion_r578900954) can be
```
jnp.log(-jnp.expm1(-y)) - y
```

numpyro/distributions/transforms.py
```diff
@@ -450,6 +450,34 @@ def _inverse(self, y):
...
+        elif isinstance(self.domain, constraints.interval):
+            return constraints.interval(self.__call__(self.domain.lower_bound),
+                                        self.__call__(self.domain.upper_bound))
+        else:
+            raise NotImplementedError
+
+    def __call__(self, x):
+        return softplus(x)
+
+    def inv(self, x):
```
fehiepsi, line 476: you need to use `_inverse` instead of `inv` here. Since the last NumPyro release, `Transform.inv` will return an InversedTransform.

fehiepsi:
@vitkl You can add a test case for this transform [here](https://github.com/pyro-ppl/numpyro/blob/c19e7bf3bcc73f2795c736ef5b4a5ad848dce9d4/test/test_distributions.py#L1168).

numpyro/distributions/transforms.py
vitkl: I will update soon.

This got me thinking, what are other differences between jax and pytorch from this transform. Looks like jax uses more naive implementation (https://jax.readthedocs.io/en/latest/_modules/jax/_src/nn/functions.html#softplus) of softplus than pytorch (that reverts to linear after x=20 and has beta parameter).

Interestingly, this less numerically stable version still works for cell2location.

docs/source/distributions.rst
```diff
@@ -637,6 +637,14 @@ ExpTransform
     :undoc-members:
     :show-inheritance:
     :member-order: bysource
+    
+SoftplusTransform
+------------
```
fehiepsi, line 642: nit: missing some `-`

numpyro/distributions/transforms.py
```diff
@@ -451,6 +452,36 @@ def log_abs_det_jacobian(self, x, y, intermediates=None):
         return x
 
 
+class SoftplusTransform(Transform):
+    r"""
+    Transform from unconstrained space to positive domain via softplus :math:`y = \log(1 + \exp(x))`.
+    The inverse is computed as :math:`y = \log(\exp(x) - 1)`.
+    """
+    def __init__(self):
+        self.domain = constraints.real
```
fehiepsi, line 461: I think you can simply define default attributes
```
class SoftplusTransform(Transform):
    domain = constraints.real
    codomain = constraints.positive
```
Otherwise, it is better to have the `domain` argument in `__init__` (as in your initial commit).

vitkl: Changed to a simpler case.

test/test_distributions.py
```diff
@@ -1165,6 +1171,7 @@ def inv_vec_transform(y):
 @pytest.mark.parametrize('transform, event_shape', [
     (PermuteTransform(jnp.array([3, 0, 4, 1, 2])), (5,)),
     (PowerTransform(2.), ()),
+    (SoftplusTransform, ()),
```
fehiepsi, line 1174: Should be `SoftplusTransform()` I guess

vitkl: Yes, this worked! I can add more complex `codomain` stuff now if that would be useful.

fehiepsi: I think this is fine. :)

fehiepsi (approved):
LGTM, thanks @vitkl!

fritzo (approved):
Looks great, thanks for splitting this into simple PRs!

PR #922, opened 2021-02-19
add BarkerMH kernel

martinjankowiak:
pair-coded with @fehiepsi 

[The Barker proposal: combining robustness and efficiency in gradient-based MCMC](https://arxiv.org/abs/1908.11812)


test/test_mcmc.py
```diff
@@ -36,6 +36,9 @@ def potential_fn(z):
     init_params = jnp.array(0.)
     if kernel_cls is SA:
         kernel = SA(potential_fn=potential_fn, dense_mass=dense_mass)
+    elif kernel_cls is BarkerMH:
+        # TODO: fix dense_mass once BarkerMH supports it
```
martinjankowiak, line 40: i'll add dense_mass support soon so i guess it's ok if we have duplicate tests for a few days

numpyro/infer/barker.py
```diff
@@ -0,0 +1,209 @@
...
+                model_args=model_args,
+                model_kwargs=model_kwargs)
+            init_params = params_info[0]
+            model_kwargs = {} if model_kwargs is None else model_kwargs
+            self._potential_fn = potential_fn_gen(*model_args, **model_kwargs)
+        return init_params
+
+    def init(self, rng_key, num_warmup, init_params, model_args, model_kwargs):
+        self._num_warmup = num_warmup
+        # TODO (low-priority): support chain_method="vectorized", i.e. rng_key is a batch of keys
```
martinjankowiak, line 148: do we need to raise errors to catch this?

numpyro/infer/hmc_util.py
```diff
@@ -431,7 +432,10 @@ def update_fn(t, accept_prob, z_info, state):
         :return: new state of the adapt scheme.
         """
         step_size, inverse_mass_matrix, mass_matrix_sqrt, ss_state, mm_state, window_idx, rng_key = state
-        rng_key, rng_key_ss = random.split(rng_key)
+        if rng_key is not None:
```
martinjankowiak, line 435: why was this necessary?

numpyro/infer/barker.py
fehiepsi: Yeah, that is a good idea! the assertation is `assert rng_key.shape == (2,)`

numpyro/infer/hmc_util.py
fehiepsi: We don't use `find_reasonable_step_size` stuff, so `adapt_state.rng_key` is set to `None` in Barker.

numpyro/infer/barker.py
martinjankowiak: how does this look?

fehiepsi (dismissed):
LGTM apart from following nits.

@martinjankowiak I just fixed all logistic regression tests that target wrongs expected values. The true_coefs is not the mean of posterior samples. This is verified by
+ running AutoDelta with a large number of steps
+ running NUTS with a large number of samples

fehiepsi:
Hmm, CI / test-modeling (3.6) (pull_request) Failing after 69m ... <- looking into this

fehiepsi (approved):

martinjankowiak:
thanks for your help with this @fehiepsi 💯

PR #924, opened 2021-02-20
Improving compiling time in MCMC samplers

### LINKED ISSUE #923: test_compile fails locally but the error is silent in CI
For some reason, the model is traced more than needed. This will a bit affect large models.

```python
FAILED test/test_compile.py::test_mcmc_one_chain[True-True] - assert 8 == 6
FAILED test/test_compile.py::test_mcmc_one_chain[True-False] - assert 7 == 5
FAILED test/test_compile.py::test_mcmc_one_chain[False-True] - assert 5 == 4
FAILED test/test_compile.py::test_mcmc_one_chain[False-False] - assert 4 == 3
```

I have no idea why some errors (like in doctest or this one) are silent on CI...

fehiepsi:
Fixes #923. For some reason, the compiling tests pass on CI so I guess this is a system-dependent issue - probably related to the new C++ device array stuff in JAX. The change in this PR makes the compiling tests pass locally. As an effect, compiling time is 2x faster than previously. :D


martinjankowiak (approved):
👍

PR #934, opened 2021-02-28
Fix student-t.cdf numerical issue

fehiepsi:
If we use `1 - x` where x is a very small positive number, we will lose precision. This PR fixes that issue.


numpyro/distributions/continuous.py
```diff
@@ -1154,7 +1155,7 @@ def _tail_prob_at_low(self):
     @lazy_property
     def _tail_prob_at_high(self):
         # if low < loc, returns cdf(high) = 1; otherwise returns 1 - cdf(high) = 0
-        return jnp.where(self.low < self.base_dist.loc, 1., 0.)
+        return jnp.where(self.low <= self.base_dist.loc, 1., 0.)
```
fehiepsi, line 1158: this is a typo: we used `loc >= low` everywhere in the implementation.

martinjankowiak (approved):

PR #937, opened 2021-03-03
Higgs example for HMCECS

OlaRonning:
This PR introduces a minimal example for using HMCECS with the relatively large [Higgs](https://archive.ics.uci.edu/ml/datasets/HIGGS).

Loading the Higgs dataset takes a while, so I introduced `num_datapoints` to `load_dataset` to avoid loading the entire dataset into memory. 

The default values are the smallest such that HMCECS outperforms HMC in runtime (with compile-time) on my machine using a Quadro RTX 6000.

Note: `jaxlib==0.1.61` breaks examples in Numpyro if they use `numpyro.plate` and the data comes from type `numpy.ndarray`, both when I use `jnp.asarray(data)`,  `jnp.array(data)` and `jax.put_device(data)`. Not an issue to with `jaxlib==0.1.60`.


examples/hmcecs.py
```diff
@@ -0,0 +1,147 @@
...
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser("Hamiltonian Monte Carlo with Energy Conserving Subsampling")
+    parser.add_argument('--subsample_size', type=int, default=1300)
+    parser.add_argument('--num_svi_steps', type=int, default=5000)
+    parser.add_argument('--num_blocks', type=int, default=100)
+    parser.add_argument('--num_warmup', type=int, default=500)
+    parser.add_argument('--num_samples', type=int, default=500)
+    parser.add_argument('--num_datapoints', type=int, default=1_500_000)
+    parser.add_argument('--inner_kernel', type=str, default='nuts')
+    parser.add_argument('--device', default='cpu', type=str, help='use "cpu" or "gpu".')
```
martinjankowiak, line 140: nit: use `choices=['nuts', 'hmc']` instead of `lower` etc

test/test_examples.py
```diff
@@ -24,6 +23,7 @@
     'capture_recapture.py --num-samples 4 --num-warmup 1 -m 5',
     'covtype.py --algo HMC --num-samples 10 --num-warmup 10',
     'gp.py --num-samples 10 --num-warmup 10 --num-chains 2',
+    'hmcecs.py --subsample_size 5 --num_svi_steps 1 --num_blocks 1 --num_warmup 1 --num_samples 5 --num_datapoints 100',
```
martinjankowiak, line 26: does that mean the huge dataset gets downloaded in CI? can you instead add a `dataset=['mock', 'higgs']` and create a simple fake dataset for testing on the fly?

examples/hmcecs.py
```diff
@@ -0,0 +1,147 @@
...
+    data, obs = fetch()
+
+    hmcecs_key, hmc_key = random.split(random.PRNGKey(args.rng_seed))
+
+    # choose inner_kernel
+    if args.inner_kernel.lower() == 'hmc':
+        inner_kernel = HMC(model)
+    elif args.inner_kernel.lower() == 'nuts':
+        inner_kernel = NUTS(model)
+    else:
```
martinjankowiak, line 84: get rid of this?

examples/hmcecs.py
```diff
@@ -0,0 +1,147 @@
...
+import numpyro
+import numpyro.distributions as dist
+from numpyro.examples.datasets import HIGGS, load_dataset
+from numpyro.infer import HMC, HMCECS, MCMC, NUTS, SVI, Trace_ELBO, autoguide
+from numpyro.infer.hmc_gibbs import taylor_proxy
+
+
+def model(data, obs, subsample_size):
+    n, m = data.shape
+    theta = numpyro.sample('theta', dist.continuous.Normal(jnp.zeros(m), .5 * jnp.ones(m)))
```
martinjankowiak, line 36: nit: remove continuous?

examples/hmcecs.py
```diff
@@ -0,0 +1,147 @@
+# Copyright Contributors to the Pyro project.
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+Example: Hamiltonian Monte Carlo with Energy Conserving Subsampling
+=========================
+
+This example illustrates the use of subsampling of HMC using Energy Conserving Subsampling when the likelihood is
```
martinjankowiak, line 8: maybe something like this?

This example illustrates the use of data subsampling in HMC using Energy Conserving Subsampling, which is applicable when the likelihood factorizes as a product of N terms.

examples/hmcecs.py
```diff
@@ -0,0 +1,147 @@
...
+    svi_key, mcmc_key = random.split(hmcecs_key)
+
+    # find reference parameterization
+    optimizer = numpyro.optim.Adam(step_size=1e-3)
+    guide = autoguide.AutoDelta(model)
+    svi = SVI(model, guide, optimizer, loss=Trace_ELBO())
+    params, losses = svi.run(svi_key, args.num_svi_steps, data, obs, args.subsample_size)
+    ref_params = {'theta': params['theta_auto_loc']}
+
+    # setup proxy
```
martinjankowiak, line 53: expand comment?

examples/hmcecs.py
```diff
@@ -0,0 +1,147 @@
...
+    with numpyro.plate('N', n, subsample_size=subsample_size):
+        batch_feats = numpyro.subsample(data, event_dim=1)
+        batch_obs = numpyro.subsample(obs, event_dim=0)
+        numpyro.sample('obs', dist.Bernoulli(logits=theta @ batch_feats.T), obs=batch_obs)
+
+
+def run_hmcecs(hmcecs_key, args, data, obs, inner_kernel):
+    svi_key, mcmc_key = random.split(hmcecs_key)
+
+    # find reference parameterization
```
martinjankowiak, line 46: expand comment?

examples/hmcecs.py
```diff
@@ -0,0 +1,147 @@
+# Copyright Contributors to the Pyro project.
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+Example: Hamiltonian Monte Carlo with Energy Conserving Subsampling
+=========================
```
fehiepsi, line 6: Nice example! Could you add a png plot to [this folder](https://github.com/pyro-ppl/numpyro/tree/master/docs/source/_static/img/examples) and add this example to the Applications section of this [index file](https://github.com/pyro-ppl/numpyro/blob/master/docs/source/index.rst)?

test/test_examples.py
OlaRonning: Yes, I did not think of that. Would you prefer adding a parameter to the example over mocking `_load_higgs` in the test?

examples/hmcecs.py
OlaRonning: If `inner_kernel` is `None`, HMCECS becomes naive subsampling which Betancourt showed to samples from a bias trajectory, but the option is in NumPyro, which is why I figured it would be worth displaying. I could change the choice to `naive` instead.

examples/hmcecs.py
OlaRonning: Make sense; I'll update it.

test/test_examples.py
martinjankowiak: whatever you think is simplest and most transparsent?

examples/hmcecs.py
martinjankowiak: gotcha yeah just add `naive` to `choices` 👍

test/test_examples.py
OlaRonning: I forgot how cumbersome `patching` is in python; I'll stick to your suggestion.

examples/hmcecs.py
OlaRonning: Yes, np, and I'll fix the equals.

fehiepsi (approved):
Looks great to me! Thanks, @OlaRonning! Could you also add a test for the `naive` case?

> jaxlib==0.1.61 breaks examples in Numpyro if they use numpyro.plate and the data comes from type numpy.ndarray, both when I use jnp.asarray(data), jnp.array(data) and jax.put_device(data)

It is strange. Could you create a separate issue for this? Currently, the master branch tracks the latest jaxlib version, but CI does not raise any issues yet.

fehiepsi:
@OlaRonning Not related to this PR - I just think that it might be better to make `taylor_proxy` a static method of HMCECS class. With that, users won't have to do `from numpyro.infer.hmc_gibbs import taylor_proxy`. What do you think?

OlaRonning:
@fehiepsi @martinjankowiak, thanks for the reviews and feedback! 

> Looks great to me! Thanks, @OlaRonning! Could you also add a test for the naive case?

Ups, nice catch! That's what you get for assuming nothing changes. The naive case raises a [ValueRrror](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/infer/hmc_gibbs.py#L90-L91). I'll update the PR.

> It is strange. Could you create a separate issue for this? Currently, the master branch tracks the latest jaxlib version, but CI does not raise any issues yet.

I'll open an issue with a minimal example to reproduce the problem.

> @OlaRonning Not related to this PR - I just think that it might be better to make taylor_proxy a static method of HMCECS class. With that, users won't have to do from numpyro.infer.hmc_gibbs import taylor_proxy. What do you think?

Your proposal is consistent with how the `perturbed_method` is [hidden](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/infer/hmc_gibbs.py#L545) from the user. As I believe `taylor_proxy` is a fine default, I think what you suggest would be a better solution.

Regarding adding it as a static method to `HMCECS`, I'm torn; on the one hand, it would reduce modularity, and on the other, `taylor_proxy` is currently written explicitly for `HMCECS`. For the user to interact with `taylor_proxy` would be 
``` python
from numpyro.infer.hmc_gibbs import taylor_proxy  # I prefer this one
ref_params = ..
proxy_constructor = taylor_proxy(ref_params)
...
```
vs
``` python
from numpyro.infer import HMCECS
ref_params = ..
proxy_constructor = HMCECS.taylor_proxy(ref_params)
...
```
The refactor in either case is trivial; let me know what you think.

fehiepsi:
> it would reduce modularity

Probably we can keep `taylor_proxy` there but add a static method like this?
```python
class HMCECS:
     @static_method
     def taylor_proxy(reference_params):
          return taylor_proxy(reference_params)
```

I think it is nice to make `taylor_proxy` more exposed to users, rather than hidden behind the import statement. In addition, it reduces the number of imports for using this sampler
```
from numpyro.infer.hmc_gibbs import taylor_proxy  # unnecessary/less exposed import
from numpyro.infer import MCMC, NUTS, HMCECS

OlaRonning:
> Probably we can keep taylor_proxy there but add a static method like this?

Yes, that would do.

> I think it is nice to make taylor_proxy more exposed to users, rather than hidden behind the import statement. In addition, it reduces the number of imports for using this sampler.

Ahh, I misunderstood your intentions, I thought you wanted to hide the `taylor_proxy` call completely and just make the signature `HMCECS(inner_kernel, reference_params, *, ...)`. I'll make the update directly in this PR (tomorrow morning).

fehiepsi (approved):
LGTM!

martinjankowiak (approved):

OlaRonning:
@fehiepsi cannot reproduce the problem I mentioned.

PR #939, opened 2021-03-04
Expose covtype example to docs

### LINKED ISSUE #927: Expose covtype example to docs
Originally, this example is used to benchmark. Since then, we have added new sampling methods that are suitable for large data. The inference is fast and the result is pretty good. So it would be nice to expose it to our current gallery of examples.

fehiepsi:
Resolves #927.

I also fixed random issues in docs and imports.

### DIFF
--- a/numpyro/infer/hmc.py
+++ b/numpyro/infer/hmc.py
@@ -401,12 +401,13 @@ class HMC(MCMCKernel):
         self._model = model
         self._potential_fn = potential_fn
         self._kinetic_fn = kinetic_fn if kinetic_fn is not None else euclidean_kinetic_energy
-        self._step_size = step_size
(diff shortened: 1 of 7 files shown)


numpyro/infer/hmc.py
```diff
@@ -401,12 +401,13 @@ def __init__(self,
         self._model = model
         self._potential_fn = potential_fn
         self._kinetic_fn = kinetic_fn if kinetic_fn is not None else euclidean_kinetic_energy
-        self._step_size = step_size
+        self._step_size = float(step_size) if isinstance(step_size, int) else step_size
```
fehiepsi, line 404: this resolves issues when users provide wrong dtypes such as `step_size=1` or `trajectory_length=6`.

fritzo (approved):

fehiepsi:

PR #940, opened 2021-03-04
Clean TruncatedNormal/Cauchy and Uniform

fehiepsi:
Previously, those classes are implemented as transformed distributions to address dynamic support issues. But that support has been dropped for a while, in favor of Reparm API. This PR removes those unnecessary layers.

### DIFF
--- a/numpyro/distributions/continuous.py
+++ b/numpyro/distributions/continuous.py
@@ -31,7 +31,7 @@ import jax.nn as nn
 import jax.numpy as jnp
 import jax.random as random
 from jax.scipy.linalg import cho_solve, solve_triangular
-from jax.scipy.special import betainc, expit, gammaln, logit, log_ndtr, logsumexp, multigammaln, ndtr, ndtri
+from jax.scipy.special import betainc, expit, gammaln, logit, logsumexp, multigammaln, ndtr, ndtri
 
 from numpyro.distributions import constraints
 from numpyro.distributions.distribution import Distribution, TransformedDistribution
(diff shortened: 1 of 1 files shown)


numpyro/distributions/continuous.py
```diff
@@ -1415,61 +1378,24 @@ def tree_unflatten(cls, aux_data, params):
...
-        return self.base_loc - ndtri(ndtr(self.base_loc) * (1 - u))
-
-    @validate_sample
-    def log_prob(self, value):
-        # log(cdf(high) - cdf(low)) = log(1 - cdf(low)) = log(cdf(-low))
-        return self._normal.log_prob(value) - log_ndtr(self.base_loc)
-
-
-class TruncatedNormal(TransformedDistribution):
+class TruncatedNormal(LeftTruncatedDistribution):
```
fritzo, line 1381: Hmm, I would expect `TruncatedNormal` to be a dispatcher (and similar for `TruncatedCauchy`):
```py
def TruncatedNormal(loc, scale, low=None, high=None, validate_args=None):
    if low is None:
        if high is None:
            return Normal(loc, scale, validate_args=validate_args)
        else:
            return RightTruncatedNormal(loc, scale, high, validate_args=validate_args)
    else:
        if high is None:
            return LeftTruncatedNormal(loc, scale, low, validate_args=validate_args)
        else:
            return BiTruncatedNormal(loc, scale, low, high, validate_args=validate_args)
```
I guess that's out of scope for this PR though...

test/infer/test_autoguide.py
```diff
@@ -183,7 +183,8 @@ def test_uniform_normal():
     def model(data):
         alpha = numpyro.sample('alpha', dist.Uniform(0, 1))
         with numpyro.handlers.reparam(config={'loc': TransformReparam()}):
-            loc = numpyro.sample('loc', dist.Uniform(0, alpha))
+            loc = numpyro.sample('loc', dist.TransformedDistribution(
+                dist.Uniform(0, 1), transforms.AffineTransform(0, alpha)))
```
fritzo, line 187: Will this break existing users?  If so, will the breakage emit an informative error message?

fritzo (approved):
LGTM, though we may want to revisit the interface to allow two-sided truncated distributions.

test/infer/test_autoguide.py
fehiepsi: Yes. I'll raise a more informative message in `TransformReparam`.

numpyro/distributions/continuous.py
fehiepsi: At first, I made a dispatcher and added a deprecation warning message. But some tests require the old interface (e.g. to test mean/var), so I would like to defer doing so in future PRs, if needed.

> revisit the interface to allow two-sided truncated distributions

What is a good way to do this without breaking existing code? I am not sure why I decided to use `low, loc, scale` order back then... Probably we can add a deprecation/warning message in this version, and change the API in the next version.

fritzo: Future PRs sounds good :+1:

> What is a good way to do this without breaking existing code?

Well I'm unsure how to avoid _breaking_ code, but you can avoid _silently_ breaking by requiring kwargs rather than args:
```py
def TruncatedNormal(*, loc=None, scale=None, low=None, high=None, validate_args=None):
    ...
```

fehiepsi: Haha, nice trick!!! I'll keep this in mind. Currently, I'm not sure if it is better to remove `TruncatedNormal`, `TruncatedCauchy` apis (in favor of `TruncatedDistribution` API), so I'll defer the change to after the next release.

numpyro/infer/reparam.py
```diff
@@ -30,23 +29,23 @@ def __call__(self, name, fn, obs):
...
         """
         Wrap in Independent and ExpandedDistribution distributions.
         """
         # Match batch_shape.
         assert fn.event_dim <= event_dim
-        fn_batch_shape = batch_shape + (1,) * (event_dim - fn.event_dim)
-        fn_batch_shape = lax.broadcast_shapes(fn_batch_shape, fn.batch_shape)
-        if fn.batch_shape != fn_batch_shape:
-            fn = fn.expand(fn_batch_shape)
+        fn = fn.expand(expand_shape)  # no-op if expand_shape == fn.batch_shape
```
fehiepsi, line 48: @fritzo This is a bit subtle. Currently, when we unwrap `dist.Normal(0, 1).expand([10]).to_event()`, the `fn, batch_shape, event_dim` is `dist.Normal(0, 1), (), 1`. So we are losing the shape `(10,)` information. I have adjusted the logic a bit such that `_unwrap` will return `fn, expand_shape, event_dim` such that we can always recover the original distribution with `fn.expand(expand_shape).to_event(event_dim - fn.event_dim)`. With that, it seems to me that we don't need this broadcasting logic. As a trade-off, we have to make sure that regardless how we reparam the base `fn`, event dimensions need to be consistent, that is
```
fn, expand_shape, event_dim = _unwrap(fn)
reparam_fn = foo(fn)
assert foo_inv(reparam_fn).event_dim == fn.event_dim
reparam_event_dim = event_dim - fn.event_dim + reparam_fn.event_dim
base_value = sample(name, _wrap(reparam_fn, expand_shape, reparam_event_dim))
```
For example, in ProjectedNormal, reparam_fn should be `dist.Normal(...).to_event(1)`. Similarly, for TransformReparam, we need to make sure that `transform(base_dist).event_dim == transformed_dist.event_dim`, i.e. an atom event of base_dist should be transformed exactly to an atom event of transformed_dist. This is guaranteed given our recent change in TransformedDistribution.

fritzo (approved):

numpyro/infer/reparam.py
fritzo: Indeed that's subtle, thanks for explaining.

Issue #855, opened 2020-12-31
Softplus transform as a more numerically stable way to enforce positive constraint


vitkl:
Happy New Year Numpyro team!

I observe a very substantial improvement in inference stability when I replace `exp` transformation with `softplus` for constraining `site_scale` (https://github.com/vitkl/cell2location_numpyro/blob/main/cell2location_numpyro/distributions/AutoNormal.py#L142). Without this change the ELBO tends to become very variable and shoot up (see plot). 
![image](https://user-images.githubusercontent.com/22567383/103389266-bcd61b00-4b05-11eb-9cc3-f863ea6ec98b.png) 

This also agrees with Softplus being used in pymc3 and this discussion: https://github.com/tensorflow/probability/issues/751.

So I am wondering if you could add softplus transformation to the library as well as use it as a default way to enforce the positive constraint.

I tried implementing it, starting from TFP implementation (https://github.com/tensorflow/probability/blob/v0.11.1/tensorflow_probability/python/bijectors/softplus.py#L61-L173) but I am confused by the difference in the meaning of `log_det_jacobian` between  TFP and Numpyro.

fehiepsi:
Happy new year @vitkl! Your suggestion looks reasonable to me. I think we can start with an implementation of SoftplusTransform (probably without `hinge_softness` for simplicity). Would you like to submit a PR? Regarding `log_det_jacobian`, I think both frameworks use the same terminology: that is, `log_abs_det_jacobian(x, y) = -softplus(-x)`.

After that, there might be several approaches to do. The simplest one to me is to add a context manager to switch from ExpTransform to SoftplusTransform for `biject_to(constraints.positive)` globally (i.e. modify the behavior of [this register](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/distributions/transforms.py#L616)).

vitkl:
Hi again, 

Tested whether replacing `Exp` with `Softplus` improves the behaviour of our model (https://github.com/vitkl/cell2location_numpyro). As I mentioned before, using `Softplus` for transforming `site_scale` to positive substantially improves stability. However, I do not see additional improvement in the stability (or accuracy) after adding `Softplus` as a default positive transformation (https://github.com/vitkl/numpyro/commit/2369cae8c1b4ffb05f7473f4c24eadc13bf86ecd). So I am not sure whether `Softplus` should be used as a default transformation - using it for `site_scale` is more important. What do you think?

fehiepsi:
[...]

PR #941, opened 2021-03-04
Use softplus transform in autoguide

fehiepsi:
This ports @fritzo's PR https://github.com/pyro-ppl/pyro/pull/2767, which can be helpful for the training process, to address #855.

### DIFF
--- a/numpyro/distributions/constraints.py
+++ b/numpyro/distributions/constraints.py
@@ -375,9 +375,13 @@ class _SoftplusPositive(_GreaterThan):
     def __init__(self):
         super().__init__(lower_bound=0.0)
 
+    def feasible_like(self, prototype):
+        return jax.numpy.full(jax.numpy.shape(prototype), np.log(2))
+
 
 class _SoftplusLowerCholesky(_LowerCholesky):
-    pass
+    def feasible_like(self, prototype):
+        return jax.numpy.broadcast_to(jax.numpy.eye(prototype.shape[-1]) * np.log(2), prototype.shape)
 
 
 class _Sphere(Constraint):
--- a/numpyro/distributions/constraints.py
+++ b/numpyro/distributions/constraints.py
@@ -47,6 +47,8 @@ __all__ = [
     'real_vector',
     'simplex',
     'sphere',
+    'softplus_lower_cholesky',
+    'softplus_positive',
     'unit_interval',
     'Constraint',
 ]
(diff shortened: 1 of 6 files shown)


fritzo (approved):
Thanks for porting this!

fritzo:
I'm not sure what's causing the errors. Maybe we should preserve the old default
```
scale_constraint = constraints.positive
```
until @vitkl demonstrates improved performance?  I also saw some new failures in https://github.com/pyro-ppl/pyro/pull/2767 and needed to tweak some learning rates.  I'm ok changing also Pyro to preserve the old default until experiments show improvement.

numpyro/distributions/constraints.py
```diff
@@ -369,6 +371,19 @@ def feasible_like(self, prototype):
         return jax.numpy.full_like(prototype, 1 / prototype.shape[-1])
 
 
+class _SoftplusPositive(_GreaterThan):
+    def __init__(self):
+        super().__init__(lower_bound=0.0)
+
+    def feasible_like(self, prototype):
+        return jax.numpy.full(jax.numpy.shape(prototype), np.log(2))
```
fehiepsi, line 379: nit: should we scale softplus transform so that 0 -> 1 (instead of log(2))?

vitkl: pymc3 initialises to 0, such that `scales = softplus(0)`

https://discourse.pymc.io/t/how-to-choose-initial-values-for-meanfield-sigma-and-transform-rho-sigma/6343/2?u=vitkl

vitkl:
> I also saw some new failures in [pyro-ppl/pyro#2767](https://github.com/pyro-ppl/pyro/pull/2767) and needed to tweak some learning rates. 

Could that be because of initial values `scales = softplus(log(2))` rather than because of learning rates?

fehiepsi:
> Could that be because of initial values scales = softplus(log(2))

The initial value of scale is `0.1` by default. We can change init using `init_scale=...` argument in the constructor of auto guides.

numpyro/distributions/constraints.py
fehiepsi: Interesting! Currently, the behavior in Pyro/NumPyro is to init unconstrained scale parameter to `softplus_inv(0.1)` or `exp_inv(0.1)`. This feasible_like method is only useful to initialize latent variables (e.g. in HMC). I think having 0 -> 1 here is not important...

fritzo (approved):
LGTM

fehiepsi:

Issue #942, opened 2021-03-06
Beta Distribution samples wrong for a=b--->0


GUIpsp:
For small values of a and b, the sample method tends to produce the value 0.5, instead of the expected values near 0 and 1. This issue comes from the underlying Dirichlet distribution.

This issue is shared with the corresponding pytorch dist.
https://github.com/pytorch/pytorch/issues/15738

Tensorflow probability and scipy stats do not have these issues.
Here's tensorflow's implementation:
https://github.com/tensorflow/probability/blob/v0.12.1/tensorflow_probability/python/distributions/dirichlet.py#L232

Here's a colab notebook showing the behaviour: https://colab.research.google.com/drive/1j4LuadlQfGzycsOYpYSKUeMCihEBngrS?usp=sharing

This is as far as I got debugging it. Hope it helps.

fehiepsi:
Interesting. Seems like we can add a quick fix for this based on [this stackexchange thread](https://stats.stackexchange.com/questions/7969/how-to-quickly-sample-x-if-expx-gamma).

fehiepsi:
Thank you for reporting the issue and giving us a clear context!! I think we have fixed the issue in #943. If you still face numerical issues, please feel free the reopen this.

PR #943, opened 2021-03-06
Improve precision for dirchlet sampler

fehiepsi:
Addresses https://github.com/pyro-ppl/numpyro/issues/942

Current tests for `beta` and `dirichlet` should exercise the change.

image

### DIFF
--- a/test/test_distributions.py
+++ b/test/test_distributions.py
@@ -450,6 +450,7 @@ def test_sample_gradient(jax_dist, sp_dist, params):
         "Gamma": ["concentration"],
         "Beta": ["concentration1", "concentration0"],
         "Chi2": ["df"],
+        "Dirichlet": ["concentration"],
         "InverseGamma": ["concentration"],
         "LKJ": ["concentration"],
         "LKJCholesky": ["concentration"],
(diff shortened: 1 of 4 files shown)


fritzo (dismissed):

test/test_distributions.py
```diff
@@ -450,6 +450,7 @@ def test_sample_gradient(jax_dist, sp_dist, params):
         "Gamma": ["concentration"],
         "Beta": ["concentration1", "concentration0"],
         "Chi2": ["df"],
+        "Dirichlet": ["concentration"],
```
fehiepsi, line 453: previously, this test passed for Dirichlet is just an artifact (e.g. due to some specific parameters/random seeds); we should skip it here.

fritzo: IIUC the reparametrized gradients for Gamma, Beta, Dirichlet, Chi2, and StudentT are all stochastic in the sense that gradients are correct only in expectation.  You might instead add a weaker test that drew many samples and estimated the Jacobian `d/dtheta E_p(z|theta)[ f(z) ]` for the parameter `theta` and some collection of functions `f` like `[1, z, z**2, z**3, z**4]`.

fritzo: See e.g. Figure 2 in https://arxiv.org/pdf/1806.01851.pdf

fehiepsi: Thanks, currently we have a TODO for them at test_pathwise_gradient. Let me create a separate issue instead.

fehiepsi:
Thanks for reviewing, @fritzo! I just fixed an import lint issue since your last approval. Could you help me approve this again? Thanks!

fritzo (approved):

PR #946, opened 2021-03-07
Fix AutoDelta with deterministic sites

### LINKED ISSUE #945: AutoDelta raises duplication error
Currently, using AutoDelta on some models generates sort of `AssertionError: all sites must have unique names but got `sd_dip_rnde` duplicated` error. More information to come...

fehiepsi:
Fixes #945 by adding `handlers.block()` in `setup_prototype`. When there are deterministic sites, the `postprocess_fn` traces the model (we can skip tracing in `postprocess_fn` but it is beyond the scope of this PR), hence creates duplicated site during `svi.init(...)`.

Also added small fixes for some warnings in CI.

### DIFF
--- a/docs/source/distributions.rst
+++ b/docs/source/distributions.rst
@@ -638,14 +638,6 @@ ExpTransform
     :show-inheritance:
     :member-order: bysource
     
-SoftplusTransform
------------------
-.. autoclass:: numpyro.distributions.transforms.SoftplusTransform
-    :members:
-    :undoc-members:
-    :show-inheritance:
-    :member-order: bysource
-
 IdentityTransform
 -----------------
 .. autoclass:: numpyro.distributions.transforms.IdentityTransform
--- a/examples/neutra.py
+++ b/examples/neutra.py
@@ -116,28 +116,28 @@ def main(args):
     ax1.set_title('Autoguide training loss\n(after 1000 steps)')
 
     ax2.contourf(X1, X2, P, cmap='OrRd')
-    sns.kdeplot(guide_samples[:, 0], guide_samples[:, 1], n_levels=30, ax=ax2)
+    sns.kdeplot(x=guide_samples[:, 0], y=guide_samples[:, 1], n_levels=30, ax=ax2)
     ax2.set(xlim=[-3, 3], ylim=[-3, 3],
             xlabel='x0', ylabel='x1', title='Posterior using\nAutoBNAFNormal guide')
 
-    sns.scatterplot(guide_base_samples[:, 0], guide_base_samples[:, 1], ax=ax3,
+    sns.scatterplot(x=guide_base_samples[:, 0], y=guide_base_samples[:, 1], ax=ax3,
                     hue=guide_trans_samples[:, 0] < 0.)
     ax3.set(xlim=[-3, 3], ylim=[-3, 3],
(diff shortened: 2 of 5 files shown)


docs/source/distributions.rst
```diff
@@ -638,14 +638,6 @@ ExpTransform
...
     :member-order: bysource
     
-SoftplusTransform
------------------
-.. autoclass:: numpyro.distributions.transforms.SoftplusTransform
-    :members:
-    :undoc-members:
-    :show-inheritance:
-    :member-order: bysource
-
```
fehiepsi, line 648: this entry is duplicated

examples/neutra.py
```diff
@@ -116,28 +116,28 @@ def main(args):
     ax1.set_title('Autoguide training loss\n(after 1000 steps)')
 
     ax2.contourf(X1, X2, P, cmap='OrRd')
-    sns.kdeplot(guide_samples[:, 0], guide_samples[:, 1], n_levels=30, ax=ax2)
+    sns.kdeplot(x=guide_samples[:, 0], y=guide_samples[:, 1], n_levels=30, ax=ax2)
```
fehiepsi, line 119: seaborn raises FutureWarning to specify explicitly keyword arguments

test/pyroapi/test_pyroapi.py
```diff
@@ -8,7 +8,7 @@
 pytestmark = pytest.mark.filterwarnings("ignore::numpyro.compat.util.UnsupportedAPIWarning")
 
 
-@pytest.yield_fixture
+@pytest.fixture
```
fehiepsi, line 11: both are the same, but the former is deprecated

numpyro/infer/autoguide.py
```diff
@@ -287,15 +287,15 @@ class AutoDelta(AutoGuide):
...
         super().__init__(model, prefix=prefix, init_loc_fn=init_loc_fn, create_plates=create_plates)
 
     def _setup_prototype(self, *args, **kwargs):
         super()._setup_prototype(*args, **kwargs)
-        self._init_locs = {
-            k: v for k, v in self._postprocess_fn(self._init_locs).items() if k in self._init_locs
-        }
+        with numpyro.handlers.block():
+            self._init_locs = {
+                k: v for k, v in self._postprocess_fn(self._init_locs).items() if k in self._init_locs
```
fritzo, line 297: Looks reasonable. What does `self._postprocess_fn()` do? Should it always be blocked, or only here? I guess an alternative place to block would be in `.__call__()`:
```py
    if self.prototype_trace is None:
        with numpyro.handlers.block():
            self._setup_prototype(*args, **kwargs)
```
then you could remove the `block` from `_setup_prototype`...

fritzo (approved):
LGTM

numpyro/infer/autoguide.py
fehiepsi: > What does self._postprocess_fn() do? Should it always be blocked, or only here?

Only here I think. `postprocess_fn` transforms unconstrained values to constrained values and also collects deterministic sites.

> I guess an alternative place to block would be in .__call__():

You are right. Do you want to move `block` there? Both places look good to me.

fritzo: Your original design looks good.

PR #947, opened 2021-03-07
Port get_mask to NumPyro

fehiepsi:
Port https://github.com/pyro-ppl/pyro/pull/2751 to NumPyro to skip log density computation in NeuTra, AutoContinuous, and HMCECS (cc @OlaRonning).

### DIFF
--- a/numpyro/primitives.py
+++ b/numpyro/primitives.py
@@ -206,6 +206,49 @@ def deterministic(name, value):
     return msg['value']
 
 
+def _inspect():
+    """
+    EXPERIMENTAL Inspect the Pyro stack.
+
+    .. warning:: The format of the returned message may change at any time and
+        does not guarantee backwards compatibility.
+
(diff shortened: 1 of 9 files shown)


numpyro/primitives.py
```diff
@@ -206,6 +206,49 @@ def deterministic(name, value):
...
+    """
+    # NB: this is different from Pyro that in Pyro, all effects applied.
+    # Here, we only apply mask effect handler.
+    msg = {
+        "type": "inspect",
+        "fn": lambda: True,
+        "args": (),
+        "kwargs": {},
+        "value": None,
+        "mask": None,
```
fehiepsi, line 227: I can't anticipate the future usage of `_inspect`, so I have simplified it a bit.

fritzo (approved):
Thanks for porting this @fehiepsi !

fehiepsi:

Issue #949, opened 2021-03-08
Automatically rendering probabilistic graphical models


kpj:
I was wondering whether there's any way of automatically rendering defined models, similar to [`pymc3.model_graph.model_to_graphviz`](https://docs.pymc.io/api/model_graph.html).

If there's currently no way of doing this, I'd be happy to submit a PR (if anyone else is also interested in this functionality).
One could either go for the pygraphviz approach or use [daft](https://github.com/daft-dev/daft) instead.

I'd be very happy if someone could point me towards a part of numpyro's codebase which would make extracting the underlying network the easiest. Rendering only a subset of numpyro's primitives would be an acceptable start for me. I'd probably go for `numpyro.sample` and `numpyro.plate` first.

fritzo:
See also https://github.com/pyro-ppl/pyro/issues/1502

fehiepsi:
This would be a very nice feature. Currently, I don't know how to extract the graphical structure of a Pyro/NumPyro model yet. Currently, you can obtain `sample <-> plate` relationship using trace:
```python
import jax
import jax.numpy as jnp

import numpyro.distributions as dist
import numpyro
from numpyro import handlers

def model(data):
    x = numpyro.sample("x", dist.Normal(0, 1))
    sd = numpyro.sample("sd", dist.LogNormal(0, 1))
    with numpyro.plate("N", len(data)):
        numpyro.sample("obs", dist.Normal(x, sd), obs=data)

data = jnp.ones(10)
trace = handlers.trace(handlers.seed(model, 0)).get_trace(data)
trace
```
which gives us
```
...
             ('obs',
              {'type': 'sample',
               'name': 'obs',
               'fn': <numpyro.distributions.distribution.ExpandedDistribution at 0x7f53a4412970>,
               'args': (),
               'kwargs': {'rng_key': None, 'sample_shape': ()},
               'value': DeviceArray([1., 1., 1., 1., 1., 1., 1., 1., 1., 1.], dtype=float32),
               'scale': None,
               'is_observed': True,
               'intermediates': [],
               'cond_indep_stack': [CondIndepStackFrame(name='N', dim=-1, size=10)],
               'infer': {}})])
```
As you can see, the `obs` site has the field `cond_indep_stack`, which is a list of plates that `obs` belongs to.

That is a good start I guess. I'll think a bit more about how to extract `sample <-> sample` relationship. One way is to use grad information.
```python
def log_probs(sample):
[...]

PR #952, opened 2021-03-11
Automatic rendering of probabilistic graphical models

kpj:
This PR implements the feature discussed in #949.

In short, it automagically transforms
```python
def model(data):
    sd = numpyro.sample("sd", dist.LogNormal(x, 1))
    with numpyro.plate("N", len(data)):
        numpyro.sample("obs", dist.Normal(x, sd), obs=data)
```
into
image


kpj:
Should I add `graphviz` to `install_requires`?

fehiepsi:
@kpj Checkout this [gist](https://nbviewer.jupyter.org/gist/fehiepsi/1b2abcb8f073d0a555aa9264e3dcc8ef) to see how your PR generates beautiful graphics for annotation models. :) Especially, the last figure is even better than the customized image.

kpj:
> You can use [try/catch](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/contrib/module.py#L37) pattern for now.

I did this for now. However, the automated tests will still fail because `graphviz` is not installed.

>  My first suggestion is to add tests on some prototype models with different structures to see if `generate_graph_specification` generates the expected output. Those tests will be very helpful for detecting bugs when we are going to modify the code in the future. (for example, I saw `class` plate is lost in `dawid_skene` model)

That's a great idea, I went ahead and implemented some tests (more are welcome!).

> Re having distribution on the side: I guess it is easy to extract the distribution class from `type(site["fn"]).__name__` field from the sites of the trace. But it is hard to get information for the parameters of those distributions though (e.g. it is hard to do something like `c ~ Normal(a ** 2, b * 10)` but it is easy to get something like `c ~ Normal`). If you think it is helpful, I'll update the `get_relations` function.

While it would be super cool to also have the parameters for each distribution, having `c ~ Normal` would already be useful, I think.

> You can use the "rename" trick I guess. I meant to have two separate plates side-by-side as in [this comment](https://github.com/pyro-ppl/numpyro/issues/949#issuecomment-793549236) and rename `MM` back to `M` when drawing.

Do you think it would be the easiest to assign an `id` and `label` to each plate in `get_model_relations` where in your case the `id`s would be `M` and `MM` but the labels `M` and `M` respectively?
Otherwise I could try to detect these plate relations from RV overlaps in `generate_graph_specification`.

numpyro/util.py
```diff
@@ -309,3 +311,190 @@ def soft_vmap(fn, xs, batch_ndims=1, chunk_size=None):
...
+                rv, label=rv, shape='circle', style='filled', fillcolor=color
+            )
+
+    for plate, data in plate_data.items():
+        parent_plate = data['parent']
+
+        if parent_plate is None:
+            graph.subgraph(plate_graph_dict[plate])
+        else:
+            plate_graph_dict[parent_plate].subgraph(plate_graph_dict[plate])
```
fehiepsi, line 470: I think you can resolve the TODOs in this function by
```python
    # add leaf nodes first
    while len(plate_data) >= 1:
        for plate, data in plate_data.items():
            parent_plate = data['parent']
            is_leaf = True
            for plate2, data2 in plate_data.items():
                if plate == data2['parent']:
                    is_leaf = False
                    break
            if is_leaf:
                plate_graph_dict[parent_plate].subgraph(plate_graph_dict[plate])
                plate_data.pop(plate)
                break
```

numpyro/util.py
```diff
@@ -309,3 +311,190 @@ def soft_vmap(fn, xs, batch_ndims=1, chunk_size=None):
...
+        plate: graphviz.Digraph(name=f'cluster_{plate}')
+        for plate in plate_groups
+        if plate is not None
+    }
+    for plate, plate_graph in plate_graph_dict.items():
+        plate_graph.attr(label=plate, labeljust='r', labelloc='b')
+
+    # add nodes
+    for plate, rv_list in plate_groups.items():
+        cur_graph = graph if plate is None else plate_graph_dict[plate]
```
fehiepsi, line 456: you can simplify this logic (and the while loop in the comment below) with
```python
    plate_graph_dict[None] = graph
    for plate, rv_list in plate_groups.items():
        cur_graph = plate_graph_dict[plate]
```

numpyro/util.py
```diff
@@ -309,3 +311,190 @@ def soft_vmap(fn, xs, batch_ndims=1, chunk_size=None):
...
+
+    # add edges
+    for source, target in edge_list:
+        graph.edge(source, target)
+
+    # return whole graph
+    return graph
+
+
+def render_model(model, *args, filename=None, **kwargs):
```
fehiepsi, line 480: I think it is better to use the pattern `render_model(model, model_args=(), model_kwargs={}, filename=None, num_tries=10)`, so we can freely add more options in the future.

fehiepsi (commented):
@kpj I think you can add `sudo apt install -y graphviz` to [action file](https://github.com/pyro-ppl/numpyro/blob/master/.github/workflows/ci.yml#L62) and add graphviz to [setup file]((https://github.com/pyro-ppl/numpyro/blob/master/setup.py#L52)) (then we can remove `graphviz` in `examples` section there).

Re drawing `y ~ Normal(...)`: I just thought that it is better for users to look at model definition instead. What do you think?

notebooks/source/model_rendering.ipynb
```diff
@@ -0,0 +1,149 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "happy-temperature",
+   "metadata": {},
+   "source": [
+    "# Automatic rendering of NumPyro models\n",
+    "\n",
+    "In this tutorial we will demonstrate how to create beautiful visualizations of your probabilistic graphical models."
```
fehiepsi, line 10: Having a notebook is a great idea. I'll craft several examples which you can adapt for the tutorial. :)

numpyro/util.py
```diff
@@ -309,3 +311,190 @@ def soft_vmap(fn, xs, batch_ndims=1, chunk_size=None):
...
+        plate_graph.attr(label=plate, labeljust='r', labelloc='b')
+
+    # add nodes
+    for plate, rv_list in plate_groups.items():
+        cur_graph = graph if plate is None else plate_graph_dict[plate]
+
+        for rv in rv_list:
+            color = 'grey' if node_data[rv]['is_observed'] else 'white'
+            cur_graph.node(
+                rv, label=rv, shape='circle', style='filled', fillcolor=color
```
fehiepsi, line 461: probably it is better to use the default `ellipse` shape, or switching between `circle` and `ellipse` depending on the length of `rv`. WDYT?

kpj:
> @kpj I think you can add `sudo apt install -y graphviz` to [action file](https://github.com/pyro-ppl/numpyro/blob/master/.github/workflows/ci.yml#L62) and add graphviz to setup file (then we can remove `graphviz` in `examples` section there).

I have done this for now. Maybe we should think about whether it makes sense to have the "big" graphviz package as a hard dependency for the whole numpyro package. Maybe one could just install the extra requirements for testing?

> Re drawing `y ~ Normal(...)`: I just thought that it is better for users to look at model definition instead. What do you think?

That's certainly true, but one could probably make the same argument for the whole visualization :-D
If retrieving the distribution definitions is too difficult, we can of course skip it for now!

notebooks/source/model_rendering.ipynb
kpj: Awesome :-)

numpyro/util.py
kpj: I first thought that most other visualizations use circles, but after checking again even daft uses ellipses sometimes.
And this would definitely be helpful for longer variable names.
So I think we can probably just switch to ellipses completely!

numpyro/util.py
kpj: Good point!

I set the default values for `model_*` to `None` (see https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments).

numpyro/util.py
kpj: This looks like a good idea!

I added the code, let's see when our tests complain :-)

numpyro/util.py
```diff
@@ -309,3 +311,190 @@ def soft_vmap(fn, xs, batch_ndims=1, chunk_size=None):
...
+
+    # infer plate structure
+    # TODO: subset relation might not always hold, raise exception if not
+    plate_data = {}
+    for plate1, plate2 in list(itertools.permutations(plate_groups, 2)):
+        if plate1 is None or plate2 is None:
+            continue
+
+        if set(plate_groups[plate1]) < set(plate_groups[plate2]):
+            plate_data[plate1] = {'parent': plate2}
```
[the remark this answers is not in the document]
kpj: I assume you are referring to a model such as
```python
def plate_improper_subsets():
    with numpyro.plate('N', 10):
        with numpyro.plate('M', 10):
            numpyro.sample('x', dist.Normal(0, 1))
```

By just looking at the RV lists, we would never be able to figure out which plate is the parent and which is the child, would we?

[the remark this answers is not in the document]
fehiepsi: You are right. We are using a convention that the order follows what appears first in the trace (in case two plates have the same content). Could you add a comment here in case we look at this again in the future?

numpyro/util.py
fehiepsi: agreed!

fehiepsi:
> retrieving the distribution definitions is too difficult

Oh, it is pretty simple. I'll make a comment on it. :)

martinjankowiak:
this is really cool! 

so as you know heuristics will never be a foolproof way of determining all conditional independencies.

the way we address this in inference algorithms that actually make use of conditional independency information (e.g. [`TraceGraph_ELBO`](http://docs.pyro.ai/en/stable/inference_algos.html?highlight=tracegraph#pyro.infer.tracegraph_elbo.TraceGraph_ELBO)) is to be maximally conservative. in particular if sample site `b` follows sample site `a` in a model then we assume that `b` depends on `a` unless conditional independence is guaranteed by something explicit like a `plate`. 

on the other hand your heuristic probably will catch most cases of dependence. 

still, i would argue that ideally you should be able to have a `conservative=True` mode in which the visualization doesn't trust the heuristics and instead only uses site ordering and plates to reason about dependency.

so for example the following model (slightly tweaked from the PR example)

```
def model(data):
    sd = numpyro.sample("sd", dist.LogNormal(0.0, 1)) # no longer depends on x
    with numpyro.plate("N", len(data)):
        numpyro.sample("obs", dist.Normal(x, sd), obs=data)
```

would generate the same visualization as

```
def model(data):
    sd = numpyro.sample("sd", dist.LogNormal(x, 1)) # depends on x
    with numpyro.plate("N", len(data)):
        numpyro.sample("obs", dist.Normal(x, sd), obs=data)
```

if `conservative=True` is activated.

numpyro/util.py
```diff
@@ -309,3 +311,227 @@ def soft_vmap(fn, xs, batch_ndims=1, chunk_size=None):
...
+    # TODO: put import in more sensible location
+    from numpyro import handlers
+
+    model_args = model_args or ()
+    model_kwargs = model_kwargs or {}
+
+    trace = handlers.trace(handlers.seed(model, 0)).get_trace(*model_args, **model_kwargs)
+    obs_sites = [name for name, site in trace.items()
+                 if site['type'] == 'sample' and site['is_observed']]
+
```
fehiepsi, line 332: @kpj This will give you names of the distributions
```python

def _get_dist_name(fn):
    if isinstance(fn, (dist.Independent, dist.ExpandedDistribution, dist.MaskedDistribution)):
        return _get_dist_name(fn.base_dist)
    return type(fn).__name__

sample_dist = {name: _get_dist_name(site["fn"]) for name, site in trace.items() if site["type"] == "sample"}
```

numpyro/util.py
kpj: The egg shape is also great :-D
image

We are now using ellipses.

numpyro/util.py
kpj: Awesome! Was planning to do exactly that :-)

And because positioning the text node and aligning the text properly seems difficult, I'll leave it off by default.

numpyro/util.py
```diff
@@ -309,3 +311,254 @@ def soft_vmap(fn, xs, batch_ndims=1, chunk_size=None):
...
+    model,
+    model_args=None,
+    model_kwargs=None,
+    filename=None,
+    render_distributions=False,
+    num_tries=10,
+):
+    """
+    Wrap all functions needed to automatically render a model.
+
```
fehiepsi, line 542: Could you add two warnings:
+ This utility has not support the :func:`~numpyro.contrib.control_flow.scan` primitive yet.
+ Currently, the utility uses a heuristic approach, which will work for most cases, to detect dependency in a NumPyro model.

notebooks/source/model_rendering.ipynb
fehiepsi: This might be beyond the purpose of this PR: I think you can port some plots in [this gist](https://nbviewer.jupyter.org/gist/fehiepsi/1b2abcb8f073d0a555aa9264e3dcc8ef) to this tutorial. It would be nice to say a bit about `unflatten` feature, `render_distributions` feature, and the difference in using `reparam` in `logistic_random_effects` model.

fehiepsi (commented):
Looks great to me overall. I just have some small nits:

fehiepsi:
> have a conservative=True mode in which the visualization doesn't trust the heuristics and instead only uses site ordering and plates to reason about dependency.

This mode will likely create confusing plate interpretation for a model. What is your motivation to have a monotone graph of where nodes/plates follow the order in the trace? Plotting the trace might be tricky for models where we define `plate` before using it as a context manager:
```python
def model():
    N_plate = numpyro.plate("N", 10, dim=-2)
    M_plate = numpyro.plate("M", 5, dim=-1)
    with N_plate:
        with M_plate:
            y = numpyro.sample("y", dist.Normal(0, 1))
    with M_plate:
        z = numpyro.sample("z", dist.Normal(0, 1))
```
I would prefer to add a note that we are using a heuristic approach.

fehiepsi:
@martinjankowiak @fritzo @eb8680 @neerajprad Could you put your opinion on this? Here are some [plots](https://nbviewer.jupyter.org/gist/fehiepsi/1b2abcb8f073d0a555aa9264e3dcc8ef) for your reference. :)

numpyro/util.py
```diff
@@ -309,3 +311,254 @@ def soft_vmap(fn, xs, batch_ndims=1, chunk_size=None):
...
+    :param model: Model to render.
+    :param str filename: File to save rendered model in.
```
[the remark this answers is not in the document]
kpj: The format can already be influenced via the filename (e.g., `model.pdf` vs `model.png`).
Would it be enough to simply add a note for this in the documentation?

notebooks/source/model_rendering.ipynb
kpj: I'd be happy to do so!
Do you think the `unflatten` feature is useful enough to include its usage with an extra argument to `render_model`?

numpyro/util.py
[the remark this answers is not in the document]
fehiepsi: Oh, great! Then I think we don't need format argument. In the notebook, could you display `png` format so that the diff is small between different commits?

notebooks/source/model_rendering.ipynb
fehiepsi: My impression so far is it is tricky to choose a good `unflatten` value. It might be better to let users decide what they want. We might introduce it in the tutorial. It is also helpful to show users how to save the graph to a file there (probably a link to the graphviz API docs for saving is enough) - personally, I would like to play with the graph and save it in various formats without having to rerun `render_model`. :)

eb8680:
This is great!  @martinjankowiak and @fehiepsi make some good points about the need to be conservative when inferring dependence with heuristics. 

A more direct approach we could explore in the future is to write a [custom Jaxpr interpreter](https://jax.readthedocs.io/en/latest/notebooks/Writing_custom_interpreters_in_Jax.html) that computes dependencies by propagating them through Jax primitives, as in e.g. [this paper](https://proceedings.neurips.cc/paper/2011/hash/0d7de1aca9299fe63f3e0041f02638a3-Abstract.html).

notebooks/source/model_rendering.ipynb
kpj: Agreed, that makes sense!

numpyro/util.py
```diff
@@ -309,3 +311,263 @@ def soft_vmap(fn, xs, batch_ndims=1, chunk_size=None):
...
+def get_model_relations(model, model_args=None, model_kwargs=None, num_tries=10):
+    """
+    Infer relations of RVs and plates from given model and optionally data.
+    See issue #949 on pyro-ppl/numpyro for more details.
+
+    :param int num_tries: times to trace model to detect discrete -> continuous dependency.
+    """
+    # TODO: put import in more sensible location
+    from numpyro import handlers
+    import numpyro.distributions as dist
```
fehiepsi, line 325: @kpj Per discussion with other devs, it is better to move the implementation to a separate file `numpyro/render.py` or `numpyro/contrib/render.py`. This will resolve the circular import issue and preserve the independence of `numpyro.util`.

notebooks/source/index.rst
```diff
@@ -16,6 +16,7 @@ Welcome to NumPyro Tutorials!
    ordinal_regression
    bayesian_hierarchical_linear_regression
    discrete_imputation
+   model_rendering
```
fehiepsi, line 19: Could you also add the tutorial to the end of the section "Introductory Tutorials" in this [index file](https://github.com/pyro-ppl/numpyro/blob/master/docs/source/index.rst)?

numpyro/util.py
```diff
@@ -309,3 +311,263 @@ def soft_vmap(fn, xs, batch_ndims=1, chunk_size=None):
...
+    model_args=None,
+    model_kwargs=None,
+    filename=None,
+    render_distributions=False,
+    num_tries=10,
+):
+    """
+    Wrap all functions needed to automatically render a model.
+
+    .. warning:: This utility does not support the :func:~numpyro.contrib.control_flow.scan primitive yet.
```
fehiepsi, line 544: nit:
```
:func:`~numpyro.contrib.control_flow.scan`
```

numpyro/util.py
```diff
@@ -309,3 +311,263 @@ def soft_vmap(fn, xs, batch_ndims=1, chunk_size=None):
...
+
+    .. warning:: Currently, this utility uses a heuristic approach, which will work for most cases, to detect dependencies in a NumPyro model.
```
fehiepsi, line 546: nit: probably this line is too long?

notebooks/source/model_rendering.ipynb
```diff
@@ -0,0 +1,558 @@
...
+    "data = jnp.ones(10)\n",
+    "numpyro.render_model(model, model_args=(data,))"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "false-qualification",
+   "metadata": {},
+   "source": [
+    "The visualization can be saved to a file by providing `filename='path'` to `numpyro.render_model`. You can use different formats such as PDF or PNG by changing the filename's suffix."
```
fehiepsi, line 133: Could you also mention that if `filename` is not specified, we can also change the format by using e.g. `graph.format = "png"`

fehiepsi (commented):
Sorry for the late response, @kpj! Thanks for addressing my comments and making the tutorial better. I just took the final reviewing pass: this is ready to merge modulo minor comments above. :)

numpyro/util.py
kpj: Either option is fine by me. I assume the general consensus is to prefer the second one (moving everything to `numpyro.contrib.render` and keeping the `graphviz` dependency in `extras_require.dev`)?

numpyro/util.py
kpj: Out of curiosity, how difficult would it be to also support rendering `scan`?

fehiepsi: To inspect the model and extract time information, it might be not so complicated. But it will be trickier to render. It is not clear to me how to represent the loops (which can have a long historical dependency: e.g. x_{t} depends on x_{t-1} and x_{t-2} in ARIMA) in graphics. Do you know which plate graphical notation is used to represent the loops? Probably we can start with history=1 and mimic some scan formats in [hmm enum example](http://num.pyro.ai/en/stable/examples/hmm_enum.html):
+ put all variables inside `scan` in a dashed-border rectangular (with label `t=0`)
+ create the same block with label `t=1` and put it side-to-side with block `t=0`
+ add three dots beside `t=1` block
+ add edges from `t=0` block to `t=1` block
+ add edges from outside-of-time-block nodes to inside-of-time-block nodes

Do you think it is feasible to automatically render such a graph? I can try to extract time information from the model if you want. Rendering the time plate would be truly neat, but it is complicated I guess. :)

numpyro/util.py
fehiepsi: You are right!

numpyro/util.py
kpj: In the general case, I'd just render the whole chain without the `...` notation.
Or is it clear that the "sub-models" created by each scan iteration will always have the same structure?
Could a numpyro model potentially do something like this:
```python
def model():
    def foo(carry, i):
        # doesn't work, I know :-)
        if i == 6:
            numpyro.sample('special', dist.Beta(1, 1))
        else:
            numpyro.sample('default', dist.Beta(0.1, 0.1))

        return carry, i
        
    scan(foo, 0.01, jnp.arange(10))
```

Besides rendering dots, I also don't know of any better graphical representation.
Ignoring special spatial arrangements, we could probably treat `scan` as a "special plate" and surround it with a differently styled box.

fehiepsi (approved):
Awesome work, @kpj! I'm happy to merge this. :100: 

**edit:** could you resolve the merge conflict?

kpj:
That's wonderful to hear :-)

fehiepsi (approved):

fehiepsi:
Yup, let's address this in #949. :)

kpj:
Will do! Thank you so much for the merge and your help :-)

fritzo:
@kpj any chance you could add a docstring to `get_model_relations()`?  I'd like to copy the interface in Pyro, but it's unclear whether e.g. `sample_sample` is organized as upstream -> downstreams or downstream -> upstreams.

fehiepsi:
@fritzo For
```python
def model(data):
    m = numpyro.sample('m', dist.Normal(0, 1))
    sd = numpyro.sample('sd', dist.LogNormal(m, 1))
    with numpyro.plate('N', len(data)):
        numpyro.sample('obs', dist.Normal(m, sd), obs=data)
```
the relation is
```
{'sample_sample': {'m': [], 'sd': ['m'], 'obs': ['m', 'sd']},
 'sample_dist': {'m': 'Normal', 'sd': 'LogNormal', 'obs': 'Normal'},
 'plate_sample': {'N': ['obs']},
 'observed': ['obs']}
```

fritzo:
@fehiepsi thanks for explaining! I'll submit a docstring improvement shortly.

kpj:
@fritzo Great :-) Is there anything else that needs to be done?

fritzo:
@kpj nothing remains. Thanks again for adding this function!

PR #959, opened 2021-03-15
Bump to 0.6.0

fehiepsi:
TODO
- [x] Write a release note -  see the [draft](https://github.com/pyro-ppl/numpyro/releases) - please feel free to edit it
- [x] Also fixes some issues in docs during collecting commits for the release note
- [x] Using `jnp.result_type` instead of `jax.dtypes.canonicalize_dtype`
- [x] Fix compiling issues in `mcmc` and `svi`:
    + using `jnp.zeros(())` instead of `jnp.array(0.)` (I guess this is a regression of jax - but this works)
    + convert init_params to strong types
- [x] merge with MixedHMC


numpyro/infer/svi.py
```diff
@@ -116,6 +116,9 @@ def init(self, rng_key, *args, **kwargs):
                 params[site['name']] = transform.inv(site['value'])
 
         self.constrain_fn = partial(transform_fn, inv_transforms)
+        # we convert weak types like float to float32/float64
+        # to avoid recompiling body_fn in svi.run
+        params = tree_map(lambda x: lax.convert_element_type(x, jnp.result_type(x)), params)
```
fehiepsi, line 121: @martinjankowiak Your observation is right: compiling time is slower in the recent jax release due to this issue.

setup.py
```diff
@@ -33,9 +33,9 @@
     author='Uber AI Labs',
     install_requires=[
         # TODO: pin to a specific version for the release (until JAX's API becomes stable)
-        'jax>=0.2.8',
+        'jax==0.2.10',
         # check min version here: https://github.com/google/jax/blob/master/jax/lib/__init__.py#L26
-        'jaxlib>=0.1.59',
+        'jaxlib==0.1.62',
```
martinjankowiak, line 38: why ==

martinjankowiak: maybe time to change `author='Uber AI Labs'`?

martinjankowiak (approved):

setup.py
fehiepsi: We can discuss this in the meeting and change it across pyro-ppl repo.

Re `==`: we pinned jax/jaxlib version for each release until jax is stable enough (I thought it was stable but recent regressions changed my mind - we might wait a bit more to relax this restriction)

fehiepsi:
Thanks for reviewing, @martinjankowiak! I'll release this soon.

PR #971, opened 2021-03-23
Using black in numpyro

### LINKED ISSUE #962: Use black for numpyro
Follows https://github.com/pyro-ppl/funsor/pull/449.

(we will do this when all pending PRs merged to avoid huge merge conflicts for contributors)

fehiepsi:
Resolves #962. Using `black` gives more consistent code in `funsor`, so it would be nice to do the same thing in numpyro.

The changes in this PR are:
+ Add `black --check .` and `isort --check .` to `make lint` to guarantee consistency in PRs
+ Add `black .` and `isort .` to `make format`
+ Add `black` to extras_require in `setup.py` and rearrange a few entries there to be more consistent
+ Add
```
[flake8]
ignore = W503,E203

[isort]
profile = black
combine_as_imports = true
```
to `setup.cfg` file. `W503` and `E203` are common flake issues after using black. From what I understand from various threads on the internet, black behavior is right. `combine_as_imports` is used to shorten imports like at [this line](https://github.com/pyro-ppl/numpyro/blob/32fad82846a753e5a9e4b8b09910e6de4d8af91c/numpyro/compat/pyro.py#L8)
```
from numpyro.primitives import module, param as _param, plate, sample
```
vs without it
```
from numpyro.primitives import module
from numpyro.primitives import param as _param
from numpyro.primitives import plate, sample
```

The only nontrivial change that I made is to change the [annotations](https://github.com/pyro-ppl/numpyro/compare/master...fehiepsi:black?expand=1#diff-74d943755080fb95f58171371b2b1bb535d05b9b93badce1a79f985ca39dc1e6R60) array format from `7 x 45` to `45 x 7`. Without this, every item stays in a line and this single array becomes very lengthy.


fritzo (approved):
LGTM, but there are already many merge conflicts 😅

fehiepsi:
@fritzo I just fixed some merge conflicts. Thank you for reviewing!

fritzo (approved):

PR #972, opened 2021-03-23
Fix a regression bug for `ExpandedDistribution`

fehiepsi:
The bug is reported in [forum](https://forum.pyro.ai/t/multivariate-observed-dimensions-get-mixed-up-in-the-posterior-predictive/2725/2)

```python
from jax import random
import numpy as np

dist.Poisson(np.array([[1], [10], [100]])).expand([3, 10]).sample(random.PRNGKey(0))
```
returns
```
DeviceArray([[  2,  12,  88,   1,  12,  93,   0,  10, 113,   3],
             [  9, 117,   1,  10,  95,   2,   8,  88,   1,  11],
             [ 99,   1,   8,  95,   0,  12, 112,   1,   9, 105]],            dtype=int32)
```

This regression issue is introduced in https://github.com/pyro-ppl/numpyro/pull/909 where some simplifications are performed and the tests do not cover the change.

### DIFF
--- a/numpyro/distributions/distribution.py
+++ b/numpyro/distributions/distribution.py
@@ -37,7 +37,12 @@ from jax import lax, tree_util
 import jax.numpy as jnp
 
 from numpyro.distributions.transforms import ComposeTransform, Transform
-from numpyro.distributions.util import lazy_property, promote_shapes, sum_rightmost, validate_sample
+from numpyro.distributions.util import (
+    lazy_property,
+    promote_shapes,
+    sum_rightmost,
+    validate_sample,
+)
 from numpyro.util import not_jax_tracer
 
 from . import constraints
@@ -120,6 +125,7 @@ class Distribution(metaclass=DistributionMeta):
        >>> d.event_shape
        (4,)
     """
+
     arg_constraints = {}
     support = None
     has_enumerate_support = False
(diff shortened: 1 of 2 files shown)


numpyro/distributions/distribution.py
```diff
@@ -501,17 +501,28 @@ def _sample(self, sample_fn, key, sample_shape=()):
...
-            subshape = x.shape[len(sample_shape) + len(batch_shape):]
-            # subshape == base_dist.batch_shape + event_shape of x (latter unknown for intermediates)
-            event_shape = subshape[len(self.base_dist.batch_shape):]
+            """
+            Reshapes samples and intermediates to ensure that the output
+            shape is correct: This implicitly replaces the interstitial dims
+            of size 1 in the original batch_shape of base_dist with those
+            in the expanded dims.
+            """
+            for dim1, dim2 in zip(interstitial_dims, interstitial_sample_dims):
```
fritzo, line 523: seems reasonable, but this might be cheaper using a single [np.transpose](https://numpy.org/doc/stable/reference/generated/numpy.transpose.html#numpy.transpose) rather than a loop of `np.swapaxes` calls.

fritzo (dismissed):

fehiepsi:
Thanks for reviewing and the helpful comment, @fritzo!

PR #973, opened 2021-03-25
Fix model rendering tutorial headers

kpj:
The model rendering notebook uses `#` for sub-headings which makes the ToC confusing:
image

This PR makes it use `##` instead.


fehiepsi (approved):
Thanks for the fix, @kpj!

PR #975, opened 2021-03-25
Raise errors on CI for doctest

fehiepsi:
Currently, `make doctest` catches the errors/warnings but does not raise them. This PR fixes some docs errors and makes doctest working properly again.

I also make some JAX changes:
+ import `tree_map` from `jax.tree_util` because `from jax import tree` is going to be depreciated
+ use `jax.default_backend()` instead of `jax.lib.xla_bridge.get_backend().platform`

Some other small changes are commented below

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -2,7 +2,7 @@
 #
 
 # You can set these variables from the command line.
-SPHINXOPTS    =
+SPHINXOPTS    = -W
 SPHINXBUILD   = sphinx-build
 SPHINXPROJ    = numpyro
 SOURCEDIR     = source
@@ -17,5 +17,7 @@ help:
 # Catch-all target: route all unknown targets to Sphinx using the new
 # "make mode" option.  $(O) is meant as a shortcut for $(SPHINXOPTS).
 %: Makefile
-	-@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
-	git clean -dfx source/
+	@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
+	git clean -dfx source/examples
+	git clean -dfx source/tutorials
+	git clean -f source/getting_started.rst
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -3,7 +3,7 @@ flax
 funsor
 jax>=0.1.65
 jaxlib>=0.1.45
-nbsphinx
+nbsphinx==0.8.1
 sphinx-gallery
-tfp_nightly  # TODO: change this to stable release
+tensorflow-probability
 tqdm
--- a/numpyro/examples/datasets.py
+++ b/numpyro/examples/datasets.py
@@ -154,7 +154,6 @@ def _load_sp500():
         for row in csv_reader:
             date.append(row['DATE'])
             value.append(float(row['VALUE']))
-    date = np.stack(date)
     value = np.stack(value)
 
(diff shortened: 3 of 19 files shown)


docs/Makefile
```diff
@@ -2,7 +2,7 @@
 #
 
 # You can set these variables from the command line.
-SPHINXOPTS    =
+SPHINXOPTS    = -W
```
fehiepsi, line 5: this raises error when there are warnings

docs/Makefile
```diff
@@ -17,5 +17,7 @@ help:
 # Catch-all target: route all unknown targets to Sphinx using the new
 # "make mode" option.  $(O) is meant as a shortcut for $(SPHINXOPTS).
 %: Makefile
-	-@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
-	git clean -dfx source/
+	@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
```
fehiepsi, line 20: removing `-` to actually raise an error (otherwise, the error is stripped)

docs/Makefile
```diff
@@ -17,5 +17,7 @@ help:
...
+	git clean -dfx source/examples
+	git clean -dfx source/tutorials
+	git clean -f source/getting_started.rst
```
fehiepsi, line 23: be more explicit about what we want to clean. Otherwise, this will clean any new file in `source` folder

docs/requirements.txt
```diff
@@ -3,7 +3,7 @@ flax
 funsor
 jax>=0.1.65
 jaxlib>=0.1.45
-nbsphinx
+nbsphinx==0.8.1
```
fehiepsi, line 6: `nbsphinx==0.8.2` does not render correctly (e.g. see `github_url` at [Bayesian regression tutorial](http://num.pyro.ai/en/stable/tutorials/bayesian_regression.html). I don't know how to fix the issue so would like to pin `nbsphinx` to a working version instead.

numpyro/examples/datasets.py
```diff
@@ -90,7 +89,7 @@ def train_test_split(file):
                 train.append(np.array([int(at_bats), int(hits)]))
                 season_at_bats, season_hits = row['SeasonAt-Bats'], row['SeasonHits']
                 test.append(np.array([int(season_at_bats), int(season_hits)]))
-        return np.stack(train), np.stack(test), np.array(player_names)
+        return np.stack(train), np.stack(test), player_names
```
fehiepsi, line 92: keep `player_names` a list of strings, rather than a ndarray string object.

numpyro/examples/datasets.py
```diff
@@ -307,7 +306,7 @@ def init():
 
     def get_batch(i=0, idxs=idxs):
         ret_idx = lax.dynamic_slice_in_dim(idxs, i * batch_size, batch_size)
-        return tuple(lax.index_take(a, (ret_idx,), axes=(0,)) if isinstance(a, DeviceArray)
-                     else np.take(a, ret_idx, axis=0) for a in arrays)
+        return tuple(np.take(a, ret_idx, axis=0) if isinstance(a, list)
```
fehiepsi, line 309: using `np.take` when a is a list of strings (in BASEBALL dataset).

docs/Makefile
fritzo: Good idea, should we also do this in Pyro and Funsor?

fritzo (approved):

fehiepsi:
@fritzo Thanks for reviewing and helping me resolve the merge conflict!

Issue #889, opened 2021-01-23
OverflowError: cannot convert float infinity to integer using HMCGibbs


gcgibson:
Hi all,

It's so cool that you got marginalizing discrete parameters out! Very nice. I'm wondering if there is any way to extract some summary metrics like there is in pyro. My discrete latent parameter is interpretable so it would be awesome to get some info on it.

Thanks!

fehiepsi:
Hi @gcgibson, please follow the issue https://github.com/pyro-ppl/numpyro/issues/770. In the meantime, you can use [DiscreteHMCGibbs](http://num.pyro.ai/en/latest/mcmc.html#numpyro.infer.hmc_gibbs.DiscreteHMCGibbs), which does HMC-with-Gibbs (instead of enumeration) and will give discrete samples when you run MCMC.

gcgibson:
Hi @fehiepsi ,

I tried the DiscreteHMCGibbs and I'm getting, 

OverflowError: cannot convert float infinity to integer


On the exact same model that works with marginalizing out the discrete parameter.

Any insight?


Casey

fehiepsi:
Interesting! I couldn't find a suspected line of code. Could you make a reproducible code?

gcgibson:
@fehiepsi Hi!

Here is the smallest nugget of my model that I could pull out. 

https://github.com/gcgibson/cfr_numpyro/blob/main/test.py.ipynb

Hopefully that works for you. If it doesn't can you lmk and I will attempt a smaller example?

fehiepsi:
@gcgibson It seems that there is a typo in your code at the discrete site `offset`. There, you set `obs=20` by accidentially I believe. FYI, JAX does not support dynamic size in a model, so something like `B_local[offset:]` will not work when `offset` is not a concrete scalar. A solution is to keep the array as-is and use `mask` handler, like this
```python
        offset = numpyro.sample('offset', dist.Categorical(probs=jnp.repeat(1./20, 20)))  # typo
        y_hat_lagged = jnp.roll(y_hat, offset)
        d_hat = y_hat_lagged + jnp.dot(jnp.array(B_local), b_raw)
        mask = jnp.arange(d_hat.shape[0]) >= offset
        numpyro.sample('d_obs', dist.Normal(d_hat, sigma2).mask(mask), obs=obs_deaths)
```

gcgibson:
Hi @fehiepsi , 

[...]

Issue #857, opened 2021-01-02
invalid prng key error when converting to arviz


ananis25:
I was trying to understand the numpyro API by writing the examples from the `Bayesian methods for hackers` text. Below is one of the exercises where the model sampling apparently succeeds, but it throws when converting to arviz for visualization. 

I'm sorry I don't really know how the conversion is done but I expected it to only involve data reshaping, curious why data sampling is done again?

* library version
```txt
numpyro==0.4.1
arviz==0.10.0
```

* repro
```py
import jax
from jax.config import config; config.update("jax_enable_x64", True)
import jax.numpy as jnp
import numpy as np
import arviz as az
import numpyro
import numpyro.distributions as dist
from numpyro.infer import MCMC, NUTS

def model(n_students, n_reports):
    p_cheat = numpyro.sample('p_cheat', dist.Uniform(0, 1))
    true_responses = numpyro.sample('true_responses', dist.Bernoulli(p_cheat), sample_shape=(n_students,))
    first_flips = numpyro.sample('first_flips', dist.Bernoulli(0.5), sample_shape=(n_students,))
    second_flips = numpyro.sample('second_flips', dist.Bernoulli(0.5), sample_shape=(n_students,))
    
    responses = numpyro.deterministic('responses', true_responses * first_flips + (1 - first_flips) * second_flips)
    fraction = numpyro.deterministic('fraction', responses.sum()/n_students)
    return numpyro.sample('num_reports', dist.Binomial(n_students, fraction), obs=jnp.int64(n_reports))

mcmc = MCMC(NUTS(model), 1000, 1000)
mcmc.run(jax.random.PRNGKey(42), 500, 120)
inferred = az.from_numpyro(mcmc)
```

* Error traceback
```txt
/usr/local/lib/python3.6/dist-packages/jax/random.py in _uniform(key, shape, dtype, minval, maxval)
    399     raise TypeError("uniform only accepts 32- or 64-bit dtypes.")
    400 
--> 401   bits = _random_bits(key, nbits, shape)
    402 
    403   # The strategy here is to randomize only the mantissa bits with an exponent of

/usr/local/lib/python3.6/dist-packages/jax/random.py in _random_bits(key, bit_width, shape)
    304   """Sample uniform random bits of given width and shape using PRNG key."""
    305   if not _is_prng_key(key):
--> 306     raise TypeError("_random_bits got invalid prng key.")
    307   if bit_width not in (8, 16, 32, 64):
[...]

Issue #770, opened 2020-10-01
Add a utility to get posterior samples of discrete latent variables


fehiepsi:
Per @martinjankowiak suggestion, after we marginalize discrete latent variables to run HMC and get posterior for continuous variables, it would be nice to have a utility to get posterior samples for discrete latent variables. That is we have a model with `p(data | discrete, continuous)`, we marginalize and run MCMC to get `p(continuous | data)`. This utility will get samples from `p(discrete | data, continuous)`.

References:
+ Pyro's epidemiological models
+ @fritzo's answer for a question in [forum](https://forum.pyro.ai/t/can-i-use-infer-discrete-to-get-the-posterior-distribution-of-discrete-latent-variables/2149)

fehiepsi:
This is achieved through `infer_discrete`.

Issue #744, opened 2020-09-15
Log probability calculation with arviz conversion goes wrong with enumeration


vanAmsterdam:
Not sure if I should post this here or at the arviz github, but converting an MCMC object to arviz data goes wrong when using enumeration in the model. 
A invalid _random_bits error get's thrown when calculating the log likelihoods
Check e.g. notebooks/source/discrete_imputation.ipynb and try `az.from_numpyro(impmcmc)` on cell 9

fehiepsi:
Thanks for the report, @vanAmsterdam! I have thought about this a bit. We need to tell [log_likelihood](http://num.pyro.ai/en/stable/utilities.html#numpyro.infer.util.log_likelihood) to use `enum` if the model has discrete latent variable. I think we can add a flag to `log_likelihood` to control a behavior and after the next release of numpyro, we can make a PR in arviz to enable the flag when the model has discrete latent variables.

fehiepsi:
@vanAmsterdam Do you need this feature soon? I have thought a lot about this FR but couldn't find a simple solution yet. As I understand, we need to filter out all latent sites but keep those discrete sites that the observation site depends on, and marginalize them. That seems doable to me. When there are multiple observation sites, my only solution is to loop over them and do the above steps for each of them.

A much simpler solution is to skip marginalizing those discrete sites. For that, we only need to seed the model before feeding into the likelihood calculation. For example, you can add to [this line](https://github.com/pyro-ppl/numpyro/blob/cce00d1b0e976909d686fb7866b404ea751c0f44/numpyro/infer/util.py#L581):
```python
substituted_model = seed(substituted_model, rng_seed=0)
```

It will take me a bit of time to implement the first solution (but I will try to do it before the next release). In the mean time, you can use the second solution if you don't do any diagnostics with likelihood in arviz.

vanAmsterdam:
I have no immediate use for this feature, but obviously we will need to sort it out. Once a solution is in place, we could add it to the discrete imputation example

fehiepsi:
I think this is achieved via `infer_discrete`. We'll illustrate its usage in the `annotation` example.

PR #977, opened 2021-03-27
Add infer_discrete

fehiepsi:
Addresses #744 #770 #857 #889 and related questions on forum.

This ports Pyro implementation at https://github.com/pyro-ppl/pyro/pull/2781 to numpyro.

Pair-coded with @eb8680 @fritzo @ordabayevy 

### TODO:
- [ ] Make all tests pass

### DIFF
--- a/numpyro/contrib/funsor/discrete.py
+++ b/numpyro/contrib/funsor/discrete.py
@@ -1,0 +1,140 @@
+# Copyright Contributors to the Pyro project.
+# SPDX-License-Identifier: Apache-2.0
+
+from collections import OrderedDict
+import functools
+
+import funsor
+
+import numpyro
+from numpyro.contrib.funsor import enum, plate_to_enum_plate, trace
+# from pyro.contrib.funsor.handlers.enum_messenger import _get_support_value
+# from pyro.contrib.funsor.infer.traceenum_elbo import terms_from_trace
+from numpyro.distributions.util import is_identically_one
+from numpyro.handlers import block, replay
+# from pyro.poutine.util import site_is_subsample
+
+
+@functools.singledispatch
+def _get_support_value(funsor_dist, name, **kwargs):
+    raise ValueError("Could not extract point from {} at name {}".format(funsor_dist, name))
+
+
+@_get_support_value.register(funsor.cnf.Contraction)
+def _get_support_value_contraction(funsor_dist, name, **kwargs):
+    delta_terms = [v for v in funsor_dist.terms
+                   if isinstance(v, funsor.delta.Delta) and name in v.fresh]
+    assert len(delta_terms) == 1
+    return _get_support_value(delta_terms[0], name, **kwargs)
+
+
+@_get_support_value.register(funsor.delta.Delta)
+def _get_support_value_delta(funsor_dist, name, **kwargs):
+    assert name in funsor_dist.fresh
+    return OrderedDict(funsor_dist.terms)[name][0]
+
+
+# @_get_support_value.register(funsor.Tensor)
+# def _get_support_value_tensor(funsor_dist, name, **kwargs):
+#    assert name in funsor_dist.inputs
+#    return funsor.Tensor(
+#        funsor.ops.new_arange(funsor_dist.data, funsor_dist.inputs[name].size),
+#        OrderedDict([(name, funsor_dist.inputs[name])]),
+#        funsor_dist.inputs[name].size
+#    )
+
+
+def terms_from_trace(tr):
+    """Helper function to extract elbo components from execution traces."""
+    log_factors = {}
+    log_measures = {}
+    sum_vars, prod_vars = frozenset(), frozenset()
+    history = 1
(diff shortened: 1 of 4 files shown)


fehiepsi:
@eb8680 @fritzo The tests pass nicely. :) The only nontrivial bug lies at indexing `arr[idx]` where `arr` is a numpy ndarray and `idx` is a DeviceArray:
```python
import torch
import jax.numpy as jnp

p = torch.tensor([[0.25, 0.75], [0.1, 0.9]])
z = torch.tensor([[[0]], [[1]]])
assert p[z].shape == (2, 1, 1, 2)
assert p.numpy()[z.numpy()].shape == p[z].shape
assert jnp.array(p)[jnp.array(z)].shape == p[z].shape
assert jnp.array(p)[z.numpy()].shape == p[z].shape
assert p.numpy()[jnp.array(z)].shape == p[z].shape  # fail
```

numpyro/contrib/funsor/discrete.py
```diff
@@ -0,0 +1,133 @@
...
+        sum_op, prod_op = funsor.ops.logaddexp, funsor.ops.add
+        rng_key, sub_key = random.split(rng_key)
+        approx = funsor.montecarlo.MonteCarlo(rng_key=sub_key)
+    else:
+        raise ValueError("temperature must be 0 (map) or 1 (sample) for now")
+
+    if first_available_dim is None:
+        with block():
+            model_trace = trace(seed(model, rng_key)).get_trace(*args, **kwargs)
+        first_available_dim = -_guess_max_plate_nesting(model_trace) - 1
```
fehiepsi, line 84: I added this logic to guess `first_available_dim` when it is not provided.

fritzo (dismissed):
LGTM.  Thanks for porting this @fehiepsi!

numpyro/contrib/funsor/discrete.py
```diff
@@ -0,0 +1,133 @@
...
+        else:
+            log_measure = approx_factors[terms["log_measures"][name]]
+            sample_subs[name] = _get_support_value(log_measure, name)
+            node["value"] = funsor.to_data(sample_subs[name], name_to_dim=node["infer"]["name_to_dim"])
+
+    with replay(guide_trace=sample_tr):
+        return model(*args, **kwargs)
+
+
+def infer_discrete(model, first_available_dim=None, temperature=1, rng_key=None):
```
fritzo, line 130: Do you want to add a docstring? Here are the [pyro docs](https://github.com/pyro-ppl/pyro/blob/9e1fd393eeac510238c8cf9ab7aa62193d182ed5/pyro/infer/discrete.py#L163) for reference:

<details>

```
"""
    A poutine that samples discrete sites marked with
    ``site["infer"]["enumerate"] = "parallel"`` from the posterior,
    conditioned on observations.

    Example::

        @infer_discrete(first_available_dim=-1, temperature=0)
        @config_enumerate
        def viterbi_decoder(data, hidden_dim=10):
            transition = 0.3 / hidden_dim + 0.7 * torch.eye(hidden_dim)
            means = torch.arange(float(hidden_dim))
            states = [0]
            for t in pyro.markov(range(len(data))):
                states.append(pyro.sample("states_{}".format(t),
                                          dist.Categorical(transition[states[-1]])))
                pyro.sample("obs_{}".format(t),
                            dist.Normal(means[states[-1]], 1.),
                            obs=data[t])
            return states  # returns maximum likelihood states

    .. warning: The ``log_prob``s of the inferred model's trace are not
        meaningful, and may be changed future release.

    :param fn: a stochastic function (callable containing Pyro primitive calls)
    :param int first_available_dim: The first tensor dimension (counting
        from the right) that is available for parallel enumeration. This
        dimension and all dimensions left may be used internally by Pyro.
        This should be a negative integer.
    :param int temperature: Either 1 (sample via forward-filter backward-sample)
        or 0 (optimize via Viterbi-like MAP inference). Defaults to 1 (sample).
    :param bool strict_enumeration_warning: Whether to warn in case no
        enumerated sample sites are found. Defalts to True.
    """
```

</details>

eb8680:
@fehiepsi FYI you might want to test this locally against the latest `master` commit of Funsor before merging, and in a separate followup PR you could bump the dependency in `setup.py`.  You'll also need to cut a new Funsor release before the next NumPyro release.

numpyro/contrib/funsor/discrete.py
eb8680: You should also mention that this does not (yet) support `scan`

fehiepsi:
@fritzo @eb8680, thanks for reviewing, recommendations, and especially for making this possible!

numpyro/contrib/funsor/discrete.py
fehiepsi: Thanks, I have added the docs and mentioned that this does not yet support `scan`. :)

fritzo (approved):
🎉

PR #984, opened 2021-04-01
Allow setting max tree depth during warmup phase

### LINKED ISSUE #960: Allow to set an additional max tree depth during warmup phase
This could be useful (needs some experiments though) in cases NUTS trajectories are long and it is slow to collect samples to estimate mass matrix during warmup phase. The side effect can be warmup samples are more correlated, so the estimated mass matrix might be not good.

fehiepsi:
Resolves #960


martinjankowiak (approved):
great thanks! 👍

PR #986, opened 2021-04-01
[breaking change] not setting platform to CPU by default

### LINKED ISSUE #727: NumPyro disables GPU on import

fehiepsi:
Resolves #727.

I think for those who don't use GPU/TPU, CPU is still the default device (because there are no other devices and most of those people will install jaxlib through `pip install jaxlib`). For those who install GPU-supported jaxlib version
```
pip install --upgrade jaxlib==0.1.64+cuda110 -f https://storage.googleapis.com/jax-releases/jax_releases.html
```
they will want to use GPU for some reason and they will know that JAX uses GPU device by default. So it might be a better idea to not change the default behavior of JAX.

This PR also relaxes jax/jaxlib versions and cleans up `ravel_pytree` utility, which is available on jax 0.2.11.

### DIFF
--- a/numpyro/infer/hmc.py
+++ b/numpyro/infer/hmc.py
@@ -257,9 +257,10 @@ def hmc(potential_fn=None, potential_fn_gen=None, kinetic_fn=None, algo="NUTS"):
 
         """
         step_size = lax.convert_element_type(step_size, jnp.result_type(float))
-        trajectory_length = lax.convert_element_type(
-            trajectory_length, jnp.result_type(float)
-        )
+        if trajectory_length is not None:
+            trajectory_length = lax.convert_element_type(
+                trajectory_length, jnp.result_type(float)
+            )
         nonlocal wa_update, max_treedepth, vv_update, wa_steps, forward_mode_ad
         forward_mode_ad = forward_mode_differentiation
         wa_steps = num_warmup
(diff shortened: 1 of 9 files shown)


numpyro/infer/hmc.py
```diff
@@ -257,9 +257,10 @@ def init_kernel(
 
         """
         step_size = lax.convert_element_type(step_size, jnp.result_type(float))
-        trajectory_length = lax.convert_element_type(
-            trajectory_length, jnp.result_type(float)
-        )
+        if trajectory_length is not None:
+            trajectory_length = lax.convert_element_type(
+                trajectory_length, jnp.result_type(float)
+            )
```
fehiepsi, line 263: this is a bug - somehow, the old jax is smart to make this conversion pass

test/infer/test_autoguide.py
```diff
@@ -244,7 +244,7 @@ def model():
     class _AutoGuide(AutoDiagonalNormal):
         def __call__(self, *args, **kwargs):
             return substitute(
-                super(_AutoGuide, self).__call__, {"_auto_latent": x_init}
+                super(_AutoGuide, self).__call__, {"_auto_latent": x_init[None]}
```
fehiepsi, line 247: the old jax is smart here too: this latent value should have shape `(1,)`, rather than `()`.

fritzo (approved):
LGTM, this seems like the right behavior, to agree with JAX's default.

Issue #979, opened 2021-03-28
[FR] Integrate with new Flax linen


gustavehug:
It seems that the Flax [integration](http://num.pyro.ai/en/stable/primitives.html?highlight=flax#random-flax-module) leverages the now [deprecated](https://flax.readthedocs.io/en/latest/flax.nn.html) `flax.nn` module.

This is a feature request to integrate with the new [`flax.linen`](https://github.com/google/flax/tree/master/flax/linen#linen-a-comfortable-evolution-of-flax) module.

fehiepsi:
Hi @gustavehug, do you want to take a stab at this? I'll be busy for the next few weeks and also would like to use `linen` for an LDA model after that. At my first impression, its API looks closer to PyTorch `nn.Module` than `flax.nn`.

gustavehug:
Hi @fehiepsi , yes I can take a look in the coming weeks !

fehiepsi:
This is integrated into the master branch. Thanks, @gustavehug !

PR #989, opened 2021-04-01
Support new Flax linen modules

gustavehug:
Closes https://github.com/pyro-ppl/numpyro/issues/979

Note that users now have to prefix e.g. priors with `params` with linen.

### DIFF
--- a/numpyro/contrib/module.py
+++ b/numpyro/contrib/module.py
@@ -25,7 +25,7 @@ def flax_module(name, nn_module, *, input_shape=None, **kwargs):
     :func:`~numpyro.primitives.param` statements.
 
     :param str name: name of the module to be registered.
-    :param flax.nn.Module nn_module: a `flax` Module which has .init and .apply methods
+    :param flax.linen.Module nn_module: a `flax` Module which has .init and .apply methods
     :param tuple input_shape: shape of the input taken by the
         neural network.
     :param kwargs: optional keyword arguments to initialize flax neural network
@@ -49,12 +49,16 @@ def flax_module(name, nn_module, *, input_shape=None, **kwargs):
         args = (jnp.ones(input_shape),) if input_shape is not None else ()
         # feed in dummy data to init params
         rng_key = numpyro.prng_key()
-        _, nn_params = nn_module.init(rng_key, *args, **kwargs)
+        nn_params = nn_module.init(rng_key, *args, **kwargs)
+
+        # make a mutable copy
+        nn_params = flax.core.unfreeze(nn_params)
+
         # make sure that nn_params keep the same order after unflatten
         params_flat, tree_def = tree_flatten(nn_params)
         nn_params = tree_unflatten(tree_def, params_flat)
         numpyro.param(module_key, nn_params)
-    return partial(nn_module.call, nn_params)
+    return partial(nn_module.apply, nn_params)
 
 
 def haiku_module(name, nn_module, *, input_shape=None, **kwargs):
@@ -142,25 +146,27 @@ def random_flax_module(name, nn_module, prior, *, input_shape=None, **kwargs):
         the module `B` defined as follows::
 
             class A(nn.Module):
-                def apply(self, x):
-                    return nn.Dense(x, 1, bias=False, name='dense')
+                @nn.compact
+                def __call__(self, x):
+                    return nn.Dense(1, use_bias=False, name='dense')(x)
 
             class B(nn.Module):
-                def apply(self, x):
-                    return A(x, name='inner')
+                @nn.compact
+                def __call__(self, x):
+                    return A(name='inner')(x)
 
-        has parameters `{'inner': {'dense': {'kernel': param_value}}}`. In the argument
+        has parameters `{'params': {'inner': {'dense': {'kernel': param_value}}}}`. In the argument
         `prior`, to specify `kernel` parameter, we join the path to it using dots:
-        `prior={"inner.dense.kernel": param_prior}`.
+        `prior={"params.inner.dense.kernel": param_prior}`.
 
     :param str name: name of NumPyro module
-    :param flax.nn.Module: the module to be registered with NumPyro
+    :param flax.linen.Module: the module to be registered with NumPyro
     :param prior: a NumPyro distribution or a Python dict with parameter names as keys and
         respective distributions as values. For example::
 
             net = random_flax_module("net",
-                                     flax.nn.Dense.partial(features=1),
-                                     prior={"bias": dist.Cauchy(), "kernel": dist.Normal()},
+                                     flax.linen.Dense(features=1),
+                                     prior={"params.bias": dist.Cauchy(), "params.kernel": dist.Normal()},
                                      input_shape=(4,))
 
     :type param: dict or ~numpyro.distributions.Distribution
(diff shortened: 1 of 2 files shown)


numpyro/contrib/module.py
```diff
@@ -49,12 +49,16 @@ def flax_module(name, nn_module, *, input_shape=None, **kwargs):
...
         # feed in dummy data to init params
         rng_key = numpyro.prng_key()
-        _, nn_params = nn_module.init(rng_key, *args, **kwargs)
+        nn_params = nn_module.init(rng_key, *args, **kwargs)
+
+        # make a mutable copy
+        nn_params = flax.core.unfreeze(nn_params)
+
         # make sure that nn_params keep the same order after unflatten
         params_flat, tree_def = tree_flatten(nn_params)
```
fehiepsi, line 58: I guess we only need to extract `params` field for optimizer:
```python
    params_flat, tree_def = tree_flatten(nn_params["params"])
    ...
return partial(nn_module.apply, {"params": nn_params})
```
This will keep the params ingestion the same as before (i.e. no prefix `params`). However, this will throw away important mutable variables like `batch_stats` in [nn.BatchNorm](https://flax.readthedocs.io/en/latest/notebooks/flax_basics.html#Variables-and-collections-of-variables).

I'm not sure what's a good solution here. Probably it is better to stick with the old pattern (of setting priors) and think about how to address extra mutable stats in another PR. What do you think?

fehiepsi: One solution to address extra mutable variables is to define inside this `flax_module` primitive something like this
```python
def apply_fn(*args, **kwargs_)
    variables = {"params": nn_params}
    mutable_vars = kwargs.get("mutable", [])
    if mutable_vars:
        for var in mutable_vars:
            # get the current mutable value
            mutable = numpyro.param_like(name + "$" + var)
            variables[var] = mutable  # jax.lax.stop_gradient(mutable)
        out, variables = nn_module.apply(variables, *args, **kwargs_)
        for var, mutable in variables.items():
            # update the current mutable value
            numpyro.param_like(name + "$" + var, mutable)
    else:
        out = nn_module.apply(variables, *args, **kwargs_)
    return out

return apply_fn
```
Here `SVI` will recognize those variables defined in `param_like` primitives and store it in `SVIState`.

fehiepsi:
Hi @gustavehug, I'm going to make some enhancements to address some issues with BatchNorm and Dropout layers. Would you mind resolving the new pattern for param names in this PR?

numpyro/contrib/module.py
```diff
@@ -49,12 +49,16 @@ def flax_module(name, nn_module, *, input_shape=None, **kwargs):
         args = (jnp.ones(input_shape),) if input_shape is not None else ()
         # feed in dummy data to init params
         rng_key = numpyro.prng_key()
-        _, nn_params = nn_module.init(rng_key, *args, **kwargs)
+        nn_params = nn_module.init(rng_key, *args, **kwargs)
+
+        # make a mutable copy
+        nn_params = flax.core.unfreeze(nn_params)
```
fehiepsi, line 55: You can capture parameters with `nn_params = flax.core.unfreeze(nn_params)["params"]`

numpyro/contrib/module.py
```diff
@@ -49,12 +49,16 @@ def flax_module(name, nn_module, *, input_shape=None, **kwargs):
...
+
         # make sure that nn_params keep the same order after unflatten
         params_flat, tree_def = tree_flatten(nn_params)
         nn_params = tree_unflatten(tree_def, params_flat)
         numpyro.param(module_key, nn_params)
-    return partial(nn_module.call, nn_params)
+    return partial(nn_module.apply, nn_params)
```
fehiepsi, line 61: With `params` captured as above, we can recover it here:
```
return partial(nn_module.apply, {"params": nn_params})
```

numpyro/contrib/module.py
```diff
@@ -142,25 +146,27 @@ def random_flax_module(name, nn_module, prior, *, input_shape=None, **kwargs):
...
 
             class B(nn.Module):
-                def apply(self, x):
-                    return A(x, name='inner')
+                @nn.compact
+                def __call__(self, x):
+                    return A(name='inner')(x)
 
-        has parameters `{'inner': {'dense': {'kernel': param_value}}}`. In the argument
+        has parameters `{'params': {'inner': {'dense': {'kernel': param_value}}}}`. In the argument
```
fehiepsi, line 158: no need to change the pattern here

numpyro/contrib/module.py
```diff
@@ -142,25 +146,27 @@ def random_flax_module(name, nn_module, prior, *, input_shape=None, **kwargs):
...
         `prior`, to specify `kernel` parameter, we join the path to it using dots:
-        `prior={"inner.dense.kernel": param_prior}`.
+        `prior={"params.inner.dense.kernel": param_prior}`.
```
fehiepsi, line 160: you can revert the change

numpyro/contrib/module.py
```diff
@@ -142,25 +146,27 @@ def random_flax_module(name, nn_module, prior, *, input_shape=None, **kwargs):
...
-    :param flax.nn.Module: the module to be registered with NumPyro
+    :param flax.linen.Module: the module to be registered with NumPyro
     :param prior: a NumPyro distribution or a Python dict with parameter names as keys and
         respective distributions as values. For example::
 
             net = random_flax_module("net",
-                                     flax.nn.Dense.partial(features=1),
-                                     prior={"bias": dist.Cauchy(), "kernel": dist.Normal()},
+                                     flax.linen.Dense(features=1),
+                                     prior={"params.bias": dist.Cauchy(), "params.kernel": dist.Normal()},
```
fehiepsi, line 169: you can revert the change here

test/contrib/test_module.py
```diff
@@ -85,13 +85,15 @@ def test_flax_module():
 
     with handlers.trace() as flax_tr, handlers.seed(rng_seed=1):
         flax_model_by_shape(X, Y)
-    assert flax_tr["nn$params"]["value"]["kernel"].shape == (100, 100)
-    assert flax_tr["nn$params"]["value"]["bias"].shape == (100,)
+    params = flax_tr["nn$params"]["value"]["params"]
```
fehiepsi, line 88: you can either revert the change or removing `["params"]` here

test/contrib/test_module.py
```diff
@@ -85,13 +85,15 @@ def test_flax_module():
...
+    assert params["kernel"].shape == (100, 100)
+    assert params["bias"].shape == (100,)
 
     with handlers.trace() as flax_tr, handlers.seed(rng_seed=1):
         flax_model_by_kwargs(X, Y)
-    assert flax_tr["nn$params"]["value"]["kernel"].shape == (100, 100)
-    assert flax_tr["nn$params"]["value"]["bias"].shape == (100,)
+    params = flax_tr["nn$params"]["value"]["params"]
```
fehiepsi, line 94: same as above

test/contrib/test_module.py
```diff
@@ -154,9 +156,9 @@ def test_random_module__mcmc(backend, init):
     if backend == "flax":
         import flax
 
-        linear_module = flax.nn.Dense.partial(features=1)
-        bias_name = "bias"
-        weight_name = "kernel"
+        linear_module = flax.linen.Dense(features=1)
+        bias_name = "params.bias"
+        weight_name = "params.kernel"
```
fehiepsi, line 161: you can use the old param names here

fehiepsi:
@gustavehug In #1016, I just add supports for mutable states. When this is merged, I'll add an LDA example that utilizes `flax.linen`. :)

fehiepsi:
Hi @gustavehug, can I push some commits to this PR to address the above API issues?

fehiepsi:
Thank you, @gustavehug, for the update. We have incorporated part of your changes in the master branch. :)

PR #991, opened 2021-04-03
Make infer_discrete work with scan

fehiepsi:
Continue of #977. Pair-coded with @eb8680 @fritzo @ordabayevy 

### TODO

- [ ] fix substituting issues of `enum_scan`: under `enum`, latent variable names will be `_PREV_z` and `z` while without it, there is only one latent variable names `z`. One solution is to merge `_PREV_z` and `z` in `infer_discrete`
- [ ] fix adjoint issue with length=10

Follow-up PR: add test for history > 1

Blocked by https://github.com/pyro-ppl/funsor/pull/517

### DIFF
--- a/numpyro/contrib/funsor/discrete.py
+++ b/numpyro/contrib/funsor/discrete.py
@@ -7,10 +7,9 @@ import functools
 from jax import random
 
 import funsor
-from numpyro.contrib.funsor.enum_messenger import enum, trace as packed_trace
-from numpyro.contrib.funsor.infer_util import plate_to_enum_plate
-from numpyro.distributions.util import is_identically_one
-from numpyro.handlers import block, replay, seed, trace
+from numpyro.contrib.funsor.enum_messenger import enum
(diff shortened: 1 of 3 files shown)


numpyro/contrib/funsor/discrete.py
```diff
@@ -97,27 +57,14 @@ def _sample_posterior(
...
-        log_prob = funsor.optimizer.apply_optimizer(log_prob)
+    with funsor.adjoint.AdjointTape() as tape:
+        with block(), enum(first_available_dim=first_available_dim):
+            log_prob, model_tr, log_measures = _enum_log_density(
+                model, args, kwargs, {}, sum_op, prod_op
+            )
 
     with approx:
-        approx_factors = funsor.adjoint.adjoint(sum_op, prod_op, log_prob)
+        approx_factors = tape.adjoint(sum_op, prod_op, log_prob)
```
fritzo, line 67: this is the line where `Scatter` and `Delta`s are introduced

PR #998, opened 2021-04-08
Add a flag to control regularize mass matrix behavior

### LINKED ISSUE #994: Make BarkerMH pickable
The PR #968 missed this kernel. It would be nice to support this for completeness.

### LINKED ISSUE #997: HMC fails to sample posterior properly while PyMC3 works fine
I'm having trouble with sampling from a posterior using NUTS for a relatively simple linear model. The chains fail to converge. When I implement the same model with the same priors in PyMC3, the sampler converges easily. [Here](https://github.com/fbartolic/numpyro_vs_pymc3_hmc) is a notebook which reproduces the issue for me. 

I can't figure out if there's a problem with the model or the sampler.  I've used pretty much the same settings for the sampler in PyMC3 and numpyro and the two models look identical to me. I would print out the initial log density for both models to check that they're returning the same numerical value but I'm not sure how to do that in numpyro. I've also tried different versions of numpyro and that didn't change anything.

Can anyone else reproduce this issue?

fehiepsi:
Resolves #997.

I also fix #994 to make BarkerMH pickable. And add a flag `validate_grad` to `initialize_model` to make SA an actual gradient-free kernel.

### DIFF
--- a/numpyro/infer/hmc.py
+++ b/numpyro/infer/hmc.py
@@ -194,6 +194,7 @@ def hmc(potential_fn=None, potential_fn_gen=None, kinetic_fn=None, algo="NUTS"):
         max_tree_depth=10,
         find_heuristic_step_size=False,
         forward_mode_differentiation=False,
+        regularize_mass_matrix=True,
         model_args=(),
         model_kwargs=None,
         rng_key=random.PRNGKey(0),
@@ -252,6 +253,9 @@ def hmc(potential_fn=None, potential_fn_gen=None, kinetic_fn=None, algo="NUTS"):
             `d2` is the max tree depth during post warmup phase.
         :param bool find_heuristic_step_size: whether to a heuristic function to adjust the
             step size at the beginning of each adaptation window. Defaults to False.
+        :param bool regularize_mass_matrix: whether or not to regularize the estimated mass
+            matrix for numerical stability during warmup phase. Default to True. This flag
+            does not take effect if ``adapt_mass_matrix == False``.
         :param tuple model_args: Model arguments if `potential_fn_gen` is specified.
         :param dict model_kwargs: Model keyword arguments if `potential_fn_gen` is specified.
         :param jax.random.PRNGKey rng_key: random key to be used as the source of
@@ -298,6 +302,7 @@ def hmc(potential_fn=None, potential_fn_gen=None, kinetic_fn=None, algo="NUTS"):
             dense_mass=dense_mass,
             target_accept_prob=target_accept_prob,
             find_reasonable_step_size=find_reasonable_ss,
+            regularize_mass_matrix=regularize_mass_matrix,
         )
 
         rng_key_hmc, rng_key_wa, rng_key_momentum = random.split(rng_key, 3)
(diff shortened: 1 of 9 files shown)


numpyro/infer/hmc.py
```diff
@@ -194,6 +194,7 @@ def init_kernel(
         max_tree_depth=10,
         find_heuristic_step_size=False,
         forward_mode_differentiation=False,
+        regularize_mass_matrix=True,
```
fritzo, line 197: nit: consider clearly separating positional from kwarg-only arguments everywhere, e.g.
```diff
  def init_kernel(
      init_params,
      num_warmup,
+     *,
      step_size=1.0,
      ...
```
This ensures backwards compatibility when you insert a new kwarg like `regularize_mass_matrix` in the middle.

test/infer/test_mcmc.py
```diff
@@ -897,6 +900,16 @@ def model():
     mcmc.run(random.PRNGKey(0))
 
 
+def test_SA_gradient_free():
```
martinjankowiak, line 903: how does this test anything?

numpyro/infer/hmc.py
```diff
@@ -252,6 +253,9 @@ def init_kernel(
             `d2` is the max tree depth during post warmup phase.
         :param bool find_heuristic_step_size: whether to a heuristic function to adjust the
             step size at the beginning of each adaptation window. Defaults to False.
+        :param bool regularize_mass_matrix: whether or not to regularize the estimated mass
```
martinjankowiak, line 256: do you think it'd be more useful to set the regularization scale instead? (so e.g. `0.0` corresponds to no regularization?)

test/infer/test_mcmc.py
fehiepsi: This is just to make sure that no grad call is performed. Here the model has `while_loop`, which we can't take grad (we need to use jacfwd for this).

numpyro/infer/hmc.py
fehiepsi: Currently, the regularization does
```python
var = var * (n / (n + 5)) + 1e-3 * (5 / (n + 5))
```
There are a few numbers 5, 1e-3, n. Do you suggest
```
var = var + scale * (1e-3 - var) * (5 / (n + 5))
```
so that `scale=0` means no regularization, `scale=1` means the old regularization? I feel like this regularization is not needed and might only be useful for dense mass in early warmup phase. So a boolean flag here is enough I guess.

martinjankowiak: hmm i meant the `1e-3` term but i guess setting that to zero doesn't mean there's no regularization... maybe there are too many parameters to expose meaningfully

numpyro/infer/hmc.py
```diff
@@ -252,6 +253,9 @@ def init_kernel(
...
+            matrix for numerical stability during warmup phase. Default to True. This flag
```
martinjankowiak, line 257: here and elsewhere: 'Defaults to...'

martinjankowiak (dismissed):

numpyro/util.py
```diff
@@ -75,9 +75,9 @@ def set_host_device_count(n):
 
     :param int n: number of CPU devices to use.
     """
-    xla_flags = os.getenv("XLA_FLAGS", "").lstrip("--")
+    xla_flags = os.getenv("XLA_FLAGS", "")
     xla_flags = re.sub(
-        r"xla_force_host_platform_device_count=.+\s", "", xla_flags
+        r"--xla_force_host_platform_device_count=\S+", "", xla_flags
```
fehiepsi, line 80: just updated a correct regex to match any non-space character

martinjankowiak (approved):

PR #1002, opened 2021-04-08
Implement SoftLaplace distribution

fritzo:
Port of https://github.com/pyro-ppl/pyro/pull/2791, also adding properties `.mean` and `.variance`.

This is an infinitely-differentiable distribution with maximally heavy tails while remaining log-convex:
image

## Tested
- [x] added to standard unit tests


fehiepsi (commented):
Thanks for porting this distribution! Could you add an entry to [the docs](https://github.com/pyro-ppl/numpyro/blob/master/docs/source/distributions.rst) and expand `supported_types` in LeftTruncated, RightTruncated, and TwoSidedTruncated distributions?

fehiepsi (approved):

PR #1003, opened 2021-04-12
Implement sparse Poisson.log_prob()

fritzo:
This adds an `is_sparse` kwarg to `dist.Poisson` triggering a cheaper `.log_prob()` computaiton, following the idiom of Pyro's `DirichletMultinomial` https://github.com/pyro-ppl/pyro/pull/1740

If this works well, we may want to add `is_sparse` options to other discrete distributions used as likelihoods, e.g. `GammaPoisson` aka negative binomial, `DirichletMutlinomial`, and maybe `Multinomial`.

## Questions for reviewers
- [x] Can you help me with the failing tests?  I am new to JAX reshaping logic.
- [x] Is the `is_sparse` constructor arg ok?  I slightly prefer this to a separate `SparsePoisson` subclass, since I've seen classes with multiple such algorithmic flags and it seems cleaner to avoid a giant n-cube class hierarchy of interacting variants.

## Tested
- [x] refactored test_distributions.py and added a `SparsePoisson` test

cc @phylyc


numpyro/distributions/discrete.py
```diff
@@ -620,6 +622,16 @@ def sample(self, key, sample_shape=()):
     def log_prob(self, value):
         if self._validate_args:
             self._validate_sample(value)
+        if self.is_sparse and value.shape:
+            rate, value = promote_shapes(self.rate, value)
+            nonzero = value > 0
+            sparse_value = value[nonzero]
+            sparse_rate = rate[nonzero]
```
fehiepsi, line 629: Unfortunately, this won't work (the shape of sparse_value is dynamic). I think you'll need to use `vmap(cond)` here. Something like
```python
shape = jax.lax.broadcast_shape(self.batch_shape, jnp.shape(value))
rate = jnp.broadcast_to(self.rate, shape).reshape(-1)
value = jnp.broadcast_to(value, shape).reshape(-1)
log_prob = jax.vmap(lambda r, v: jax.lax.cond(
    v > 0, lambda _: jnp.log(r) * v - gammaln(v + 1) - r, lambda _: -r, operand=None))(rate, value)
return log_prob.reshape(shape)
```

fehiepsi:
> Is the is_sparse constructor arg ok? I slightly prefer this to a separate SparsePoisson subclass

yup, both look good to me. Please go with the one you prefer. :)

numpyro/distributions/discrete.py
fritzo: Hmm, I worry this will be expensive.  In @phylyc's application (and in my applications in the past) this is used as a likelihood with static data `value`.  Would it be possible to implement this assuming `value` and `nonzero` are static constants?  I think a highly-optimized limited case would actually be most useful.

fehiepsi: Yup, it is possible then. We can add a small check here for that. Let me do a bit of benchmarking...

fritzo: @fehiepsi can you recommend any reading to help me understand when `jax.vmap(... jax.lax.cond(...))` is needed versus when `jax.obs.index_add(..)` will suffice?

fehiepsi: I think [this section](https://jax.readthedocs.io/en/latest/notebooks/thinking_in_jax.html#to-jit-or-not-to-jit) explains why `y[x>0]` does not work when `x` is a [tracer](https://jax.readthedocs.io/en/latest/glossary.html#term-Tracer). I benchmarked and found that `vmap(cond(...))` does not give any improvement for sparse arrays (actually, it is even slower than the non-sparse version). 

```python
import jax
import jax.numpy as jnp
from jax.scipy.special import gammaln
import numpyro
import numpy as np

numpyro.set_platform("cpu")


@jax.jit
def fn(rate, value):
    return (jnp.log(rate) * value) - gammaln(value + 1) - rate


@jax.jit
def fn_index_add(rate):
    nonzero = value > 0
    sparse_value = value[nonzero]
    sparse_rate = rate[nonzero]
    return jax.ops.index_add(
        -rate, nonzero,
        jnp.log(sparse_rate) * sparse_value - gammaln(sparse_value + 1))


@jax.jit
def fn_vmap_cond(rate, value):
    return jax.vmap(lambda r, v: jax.lax.cond(
        v > 0, lambda _: jnp.log(r) * v - gammaln(v + 1) - r,
        lambda _: -r, operand=None))(rate, value)

N = int(1e7)
rate = jnp.exp(jax.random.normal(jax.random.PRNGKey(0), (N,)))
value = np.zeros(N, dtype=np.int32)
value[:10] = np.arange(1, 11, dtype=np.int32)
x = fn(rate, value)
x1 = fn_index_add(rate)
x2 = fn_vmap_cond(rate, value)
np.testing.assert_allclose(x, x1, atol=1e-5)
np.testing.assert_allclose(x, x2, atol=1e-5)

%timeit x = fn(rate, value).copy()
%timeit x1 = fn_index_add(rate).copy()
%timeit x2 = fn_vmap_cond(rate, value).copy()
```

So in this case, we can do
```python
value = jax.device_get(value)
if self.is_sparse and not isinstance(value, jax.core.Tracer):
    ...
```

fritzo: Thanks for the help @fehiepsi !  It looks like this is working now.

test/test_distributions.py
```diff
@@ -652,6 +665,29 @@ def test_pathwise_gradient(jax_dist, sp_dist, params):
         assert_allclose(actual_grad[i], expected_grad, rtol=0.005)
 
 
+@pytest.mark.parametrize(
+    "jax_dist, sp_dist, params", CONTINUOUS + DISCRETE + DIRECTIONAL
+)
+def test_jit_log_likelihood(jax_dist, sp_dist, params):
```
fritzo, line 671: Whereas jitting in `test_log_prob` fixes params and varies the `value`, jitting in this test fixes the `value` and varies params, as would be the case in an observe statement `numpyro.sample(..., obs=data)`.

numpyro/distributions/discrete.py
```diff
@@ -620,6 +623,23 @@ def sample(self, key, sample_shape=()):
...
+        value = jax.device_get(value)
+        if (
+            self.is_sparse
+            and not isinstance(value, jax.core.Tracer)
+            and jnp.size(value) > 1
+        ):
+            shape = lax.broadcast_shapes(self.batch_shape, jnp.shape(value))
+            rate = jnp.broadcast_to(self.rate, shape).reshape(-1)
+            nonzero = jnp.broadcast_to(value > 0, shape).reshape(-1).nonzero()
+            value = jnp.broadcast_to(value, shape).reshape(-1)
```
fehiepsi, line 635: nit: I guess we can swap the order of those two lines:
```
value = jnp.broadcast_to(value, shape).reshape(-1)
nonzero = (value > 0).nonzero()
```
Btw, do we need to call `.nonzero()` here?

fehiepsi (approved):
LGTM. Thanks for supporting this and adding more useful tests!

numpyro/distributions/discrete.py
fritzo: Done. I think `.nonzero()` was required for tests to pass in an earlier commit, but now things pass without it, so I've removed.

fehiepsi (approved):
LGTM, thank you @fritzo!

Issue #1004, opened 2021-04-13
Support TruncatedZeroInflatedPoisson


fehiepsi:
I would like to use this distribution, so it would be nice to add some helper classes to make it easier to construct this distribution

Tasks
- [x] Refactor truncated distributions into a file
- ~[ ] Support other univariate truncated distributions through `TruncatedDistribution` api~ replaced by a tutorial instead
- [x] Add `Poisson.cdf` method
- [x] Add `ZeroInflatedDistribution` for univariate count distributions
- [ ] Write a tutorial on how to construct a truncated distribution. Relevant: https://forum.pyro.ai/t/truncated-binomial-how-to-implement/2808

cc @cameronraysmith

fehiepsi:
Close in favor of #1268

PR #1006, opened 2021-04-13
Add Poisson cdf method

fehiepsi:
Addresses #1004 

This implements Poisson.cdf and tests against scipy

### DIFF
--- a/numpyro/distributions/discrete.py
+++ b/numpyro/distributions/discrete.py
@@ -33,7 +33,7 @@ from jax import lax
 from jax.nn import softmax, softplus
 import jax.numpy as jnp
 import jax.random as random
-from jax.scipy.special import expit, gammaln, logsumexp, xlog1py, xlogy
+from jax.scipy.special import expit, gammainc, gammaln, logsumexp, xlog1py, xlogy
 
 from numpyro.distributions import constraints
 from numpyro.distributions.distribution import Distribution
@@ -629,6 +629,9 @@ class Poisson(Distribution):
     @property
     def variance(self):
         return self.rate
+
+    def cdf(self, value):
+        return 1 - gammainc(value + 1, self.rate)
 
 
 class ZeroInflatedPoisson(Distribution):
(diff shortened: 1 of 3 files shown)


test/test_distributions.py
```diff
@@ -745,6 +749,8 @@ def test_cdf_and_icdf(jax_dist, sp_dist, params):
         assert_allclose(d.icdf(d.cdf(samples)), samples, atol=1e-5, rtol=1e-5)
     except NotImplementedError:
         pass
+    except TypeError:
+        pass
```
fritzo, line 753: Where might a `TypeError` come from?  Can you confirm that this part actually runs?  I would expect it to fail with `NotImplementedError` because `Poisson.icdf()` is not implemented.

fehiepsi: I'll skip this part instead. `TypeError` comes from taking grad w.r.t. integer input. We haven't defined a definition for `icdf` of discrete distributions yet.

numpyro/distributions/discrete.py
```diff
@@ -630,6 +630,9 @@ def mean(self):
     def variance(self):
         return self.rate
 
+    def cdf(self, value):
+        return 1 - gammainc(value + 1, self.rate)
```
phylyc, line 634: This implementation relies on value to be of type int. To be consistent with the definition of the cdf, and also with the scipy implementation, I'd use
 ```
   def cdf(self, value):
        k = jnp.floor(value) + 1
        return gammaincc(k, self.rate)
```

fehiepsi: Thanks, this is a good point!

(in the future, out-of-support values will raise a warning.)

phylyc: Well, even though the pmf only has integer support, the cdf has real values support. You should keep that in mind when adding the out-of-support warnings!

fehiepsi: Yup, but I think it is better to raise a warning and let users convert the value, rather than supporting a generic definition of cdf_support (though it makes sense to say that `Exponential.cdf(-1.) = 0`).

PR #1008, opened 2021-04-13
Fix an erroneous incantation in documentation

jatentaki:
As in title: the [docs](http://num.pyro.ai/en/0.6.0/distributions.html#numpyro.distributions.directional.ProjectedNormal) suggest to use
```
@handlers.reparam(args...)
def f():
   ...
```

which doesn't work since `@handlers.reparam` is a decorator, rather than return a decorator (which would support partial initialization), resulting in `f == None` which gives rather cryptic error messages. I updated the documentation to the working incantation

```
@partial(reparam, args...)
def f():
    ...
```

Note the change from `handlers.reparam` to just `reparam` in order to fit in PEP80. I assume the reader can infer from the context what namespace `reparam` is in (it's mentioned above).

If that interpretation violates your intents, feel free to reject.

### DIFF
--- a/numpyro/distributions/directional.py
+++ b/numpyro/distributions/directional.py
@@ -87,7 +87,7 @@ class ProjectedNormal(Distribution):
     with a :class:`~numpyro.infer.reparam.ProjectedNormalReparam`
     reparametrizer in the model, e.g.::
 
-        @handlers.reparam(config={"direction": ProjectedNormalReparam()})
+        @partial(reparam, config={"direction": ProjectedNormalReparam()})
         def model():
             direction = numpyro.sample("direction",
                                        ProjectedNormal(zeros(3)))


fritzo:
In Pyro the standard notation is `@poutine.my_poutine(**kwargs)` which accomplishes the `functools.partial` internally. @fehiepsi do we want to document `@handlers.my_handler(**kwargs)` as the standard notation in NumPyro?

fehiepsi:
> do we want to document @handlers.my_handler(**kwargs) as the standard notation in NumPyro?

I think so. It would be nice to have `handlers` as a decorator. I guess we need to rewrite our handlers as
```python
class SeedMessenger(...):
    ...

def seed(fn=None, ...):
    if fn is None:
        ...
    else:
        ...
```
Do you think we can get this work without such modification?

fehiepsi:
> that interpretation violates your intents

Thank you! We would like to support both behaviors. Let us figure out a solution for this.

fritzo:
> Do you think we can get this work without such modification?

How about #1009 ?

fehiepsi:
Yeah, it is beautiful, out of my expectation. :D 

@jatentaki Thanks for bringing this up! The decorator syntax will be supported in #1009.

fritzo:
@jatentaki thanks for pointing this out!  I hope you don't mind if we upgrade the behavior rather than downgrading the docs 😉

PR #1009, opened 2021-04-13
Support decorator syntax for handlers

fritzo:
Aims to support the syntax that @jatentaki noticed was missing.
Replaces #1008 

This aims to support Pyro-style syntax like
```py
@handlers.reparam(config={"x": LocScaleReparam})
def model():
    ...
```

## Tested
- [x] added a test of all sorts of syntax

### DIFF
--- a/test/infer/test_reparam.py
+++ b/test/infer/test_reparam.py
@@ -37,6 +37,59 @@ def get_moments(x):
     return jnp.stack([m1, m2, m3, m4])
 
 
+def test_syntax():
+    loc = np.random.uniform(-1.0, 1.0, ())
+    scale = np.random.uniform(0.5, 1.5, ())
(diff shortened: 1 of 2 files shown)


eb8680 (approved):
LGTM

test/infer/test_reparam.py
```diff
@@ -37,6 +37,59 @@ def get_moments(x):
...
+    trace = numpyro.handlers.trace()
+    reparam = numpyro.handlers.reparam(config=config)
+
+    @reparam
+    @trace
+    @seed
+    def m():
+        return model()
+
+    tr4 = trace.trace
```
fehiepsi, line 88: I just guess that you need to call `m()` above this line.

fehiepsi (approved):
LGTM, thanks for making this nice syntax.

Issue #768, opened 2020-10-01
[FR] add a thompson sampling example


martinjankowiak:
add an example script that that uses thompson sampling, either in the context of gaussian process based bayesian optimization (see e.g. [here](http://gpss.cc/gpss15/talks/Nado.pdf)) or gaussian process or logistic regression based contextual binary bandits (see e.g. [here](https://gdmarmerola.github.io/ts-for-contextual-bandits/)).

for a basic GP implementation in numpyro see [here](https://github.com/pyro-ppl/numpyro/blob/master/examples/gp.py).
for a basic logistic regression implementation in numpyro see [here](https://github.com/pyro-ppl/numpyro/blob/master/examples/covtype.py).

fehiepsi:
I think this is resolved in #1014. Thanks, @dominikstrb!

PR #1014, opened 2021-04-20
Example: Thompson sampling for Bayesian Optimization with GPs

dominikstrb:
This PR addresses issue #768. I implemented a basic version of Thompson sampling for GPs based on [this tutorial](https://gdmarmerola.github.io/ts-for-bayesian-optim/). I am looking forward to improve this PR based on your feedback!


examples/thompson_sampling.py
```diff
@@ -0,0 +1,302 @@
...
+    out = (
+        -20 * jnp.exp(-0.2 * jnp.sqrt(0.5 * (x ** 2 + y ** 2)))
+        - jnp.exp(0.5 * (jnp.cos(2 * jnp.pi * x) + jnp.cos(2 * jnp.pi * y)))
+        + jnp.e
+        + 20
+    )
+    return out
+
+
+# matern kernel with nu = 1.5
```
martinjankowiak, line 40: this does not appear to be the 3/2 kernel?

examples/thompson_sampling.py
```diff
@@ -0,0 +1,302 @@
...
+
+        # store moments of training y (to normalize)
+        self.y_mean = jnp.mean(Y)
+        self.y_std = jnp.std(Y)
+
+        # normalize y
+        Y = (Y - self.y_mean) / self.y_std
+
+        # setup optimizer and SVI
+        optim = numpyro.optim.Adam(step_size=0.005)
```
martinjankowiak, line 83: i think it's usually good to use a smaller `b1` in this context, e.g. `b1=0.5`

examples/thompson_sampling.py
```diff
@@ -0,0 +1,302 @@
...
+            Y=Y,
+        )
+
+        params, _ = svi.run(rng_key, n_step)
+
+        # get kernel parameters from guide with proper names
+        self.kernel_params = svi.guide.median(params)
+
+        # store inverted prior covariance
+        self.K_xx_inv = jnp.linalg.inv(self.kernel(X, X, **self.kernel_params))
```
martinjankowiak, line 100: given that this code is most of the way towards a complete implementation would it make sense to use `jax.scipy.linalg.cholesky` and related functions instead of `inv`, which tends to be less stable?

examples/thompson_sampling.py
```diff
@@ -0,0 +1,302 @@
...
+        + 20
+    )
+    return out
+
+
+# matern kernel with nu = 1.5
+def matern32_kernel(X, Z, var=1.0, length=0.5, jitter=1.0e-6):
+    d = jnp.sqrt(0.5) * jnp.sqrt(jnp.power((X[:, None] - Z), 2.0)) / length
+    k = var * (1 + d + (d ** 2) / 3) * jnp.exp(-d)
+    if jitter:
```
martinjankowiak, line 44: make it clear you are assuming noise-free functions?

examples/thompson_sampling.py
```diff
@@ -0,0 +1,302 @@
...
+        # draw one sample
+        return jax.random.multivariate_normal(rng_key, mean=y_mean, cov=y_cov)
+
+
+# our TS-GP optimizer
+class ThompsonSamplingGP:
+
+    # initialization
+    def __init__(
+        self, gp, n_random_draws, objective, x_bounds, interval_resolution=1000
```
martinjankowiak, line 141: instead of `interval_resolution` perhaps `num_thompson_samples` would be more informative?

examples/thompson_sampling.py
```diff
@@ -0,0 +1,302 @@
+# Copyright Contributors to the Pyro project.
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+Example: Gaussian Process
+=========================
+
+In this example we show how to implement Thompson sampling for Gaussian processes.
```
martinjankowiak, line 8: please make it clear you're doing minimization

martinjankowiak (commented):
this is great @dominikstrb ! thanks for contributing.

i didn't look at the CI output so i don't know why it's failing but let us know if you have trouble getting tests to pass

examples/thompson_sampling.py
dominikstrb: You're right, it's the 5/2 kernel. I fixed the comment.

examples/thompson_sampling.py
dominikstrb: It does seem to help with stability of the optimization. Could you tell me why?

examples/thompson_sampling.py
dominikstrb: Added a comment to clarify.

examples/thompson_sampling.py
dominikstrb: `interval_resolution` was indeed ambiguous. It is not the number of Thompson samples but rather the resolution of the grid. I clarified that in the code.

examples/thompson_sampling.py
dominikstrb: I made it clear. But maybe I should add even more documentation and a plot to illustrate the idea of Bayesian optimization and Thompson sampling, right?

examples/thompson_sampling.py
dominikstrb: I'll need to look up the equations again on the weekend, but that's a good idea. Thanks!

dominikstrb:
Hi @martinjankowiak, thanks for all the suggestions! I incorporated most of them and I will also add the equations using Cholesky decompositions when I have more time.

examples/thompson_sampling.py
martinjankowiak: well in this case there is no stochasticity of any kind. `b1` controls the momentum. momentum is most useful when things are stochastic because of data-subsampling etc. here large `b1` will tend to oversmooth the objective.

examples/thompson_sampling.py
martinjankowiak: ok but i don't think this is right either? see e.g. https://docs.pymc.io/api/gp/cov.html#pymc3.gp.cov.Matern52

examples/thompson_sampling.py
```diff
@@ -63,9 +64,7 @@ def model(X, Y, kernel=matern32_kernel):
 
 
 class GP:
-    """ Adapted to numpyro from https://gdmarmerola.github.io/ts-for-bayesian-optim/ """
```
martinjankowiak: why was this removed?

examples/thompson_sampling.py
```diff
@@ -206,32 +213,19 @@ def choose_next_sample(self, n_step=2_000):
         self.X = np.append(self.X, next_sample)
         self.y = np.append(self.y, next_observation)
 
-        # return everything if possible
```
martinjankowiak: what is your logic here? why not return nothing if it errors and revert to picking a random point in the domain?

examples/thompson_sampling.py
martinjankowiak: i think you can keep it pretty simple. you can always link to papers/tutorials/etc for more info

examples/thompson_sampling.py
dominikstrb: I put this into the ThompsonSamplingGP class, where it makes more sense

examples/thompson_sampling.py
dominikstrb: This is in there so that the plots still make sense even when we plot the samples during the random exploration phase.

fehiepsi (commented):
@dominikstrb Thanks for your contribution! Could you add this example to [example page](https://github.com/pyro-ppl/numpyro/blob/master/docs/source/index.rst) and the corresponding image to [this folder](https://github.com/pyro-ppl/numpyro/tree/master/docs/source/_static/img/examples), replace the title "Gaussian Process" with the PR title, add a few introduction lines with reference to the docstring, and also embed the image like [gp example](https://github.com/pyro-ppl/numpyro/blob/master/examples/gp.py).
```
.. image:: ../_static/img/examples/thompson_sampling.png
    :align: center
```

dominikstrb:
Hi @fehiepsi, I hope I did everything right and the PR is okay to merge :) Thanks for your support and for all your work on the `numpyro` package!

fehiepsi (commented):
Thanks, @dominikstrb! LGTM pending further comments by @martinjankowiak.

martinjankowiak (approved):
lgtm thanks @dominikstrb  !

PR #1016, opened 2021-04-21
Support for mutable params

### LINKED ISSUE #990: Allow propagate mutable states in SVI
While reviewing #989, I see that there are situations that we need to maintain extra variables, like moving average stats in BatchNorm. So it would be nice to introduce an additional primitive for this. Alternatively, we can add some options into `deterministic` primitive so that SVI knows and store it.

@martinjankowiak also requested something like this last year
> it’ll just be easier to pipe through the stats i’m interested in and store them in numpyro.deterministic

Note that this is not an issue for a stateful framework like PyTorch.

Here are solutions from other jax nn libraries:
+ [flax](https://flax.readthedocs.io/en/latest/notebooks/flax_basics.html#Variables-and-collections-of-variables)  introduces `mutable` argument to decide which fields in `variables` needed to be updated when evaluates the output. Here `variables` do not only contain `params` but also other mutable states like `'batch_stats'`.
```
y, updated_state = model.apply(variables, x, mutable=['batch_stats'])
```
+ [haiku](https://dm-haiku.readthedocs.io/en/latest/api.html#haiku.transform_with_state) introduces `transform_with_state`
> This function is equivalent to transform(), however it allows you to maintain and update internal state (e.g. ExponentialMovingAverage in BatchNorm) via get_state() and set_state():
```
counter, state = f.apply(params, state, None)
```

Those frameworks also provide additional ways to deal with `rng` states in stochastic layers like Dropout. It is easy to deal with: using `numpyro.prng_key()` primitive. But for stats states like above, currently, we don't have a solution yet.

fehiepsi:
Close #990.

This PR introduces an additional `mutable_state` field in SVIState, which stores parameters marked by `infer={"mutable": True}`. Those parameters will be skipped in the optimizer and it is the job of the model/guide to update them, e.g.
```python
def model(...):
    p = numpyro.param("p", {"a": 1}, infer={"mutable": True}) 
    p["a"] = 10  # update p
```

- [x] Support mutable params in haiku
- [x] Add new primitive `numpyro.mutable`
- [x] Add a few assertations (e.g. params is None == mutable is None)
- [x] Consider to replicate methods with new input/output signatures


fritzo (commented):
Hi @fehiepsi, could you provide some documentation somewhere (maybe in the `numpyro.param` docstring?) to explain what mutable params are and why they might be useful?  I'm struggling to imagine a use case from the interface alone.

fehiepsi:
Hi @fritzo, I just add some explanation in `param` docstring. I hope that it motivates the usage of the mutable flag:
+ param + mutable=False: this param is updated by optimizers
+ deterministic: execute the model and store the intermediate values (information of its previous value is not recorded)
+ param + mutable=True: this param is not updated by optimizers, rather than that, it is updated by users' model execution

fritzo (dismissed):
LGTM, thanks for your patience and for explaining the code!

fritzo (approved):
LGTM.

I hav reviewed general design and skimmed code. I have not meticulously reviewed all logic (e.g. destructured assignment) and assume tests will catch type errors.

fehiepsi:
Thanks for reviewing, @fritzo! Right now we don't have a good set of tests but we can exercise this pattern in PR #1056.

PR #1017, opened 2021-04-21
Raise error for SVI if model/guide has discrete latent values

fehiepsi:


martinjankowiak (approved):

fehiepsi:
Thanks for reviewing, Martin!

PR #1026, opened 2021-04-26
Fix issue where scan would error out if _PYRO_STACK is empty.

### LINKED ISSUE #1024: Scan fails with empty PYRO_STACK
When `scan` is taken outside the context of a non-empty `PYRO_STACK`, `scan` produces an error.

The following code produces the error. I don't think it matters much what the scan actually does, but in this case it's interative multiplication of a vector by a matrix with the addition of an second vector at each step.

```python
def matrix_rollout(K, c_in):
    
    def iteration(c_prev, c_in):
        c_next = jnp.dot(c_prev, K) + c_in
        return c_next, (c_next,)
    
    _, (ys,) = scan(iteration, init=jnp.asarray([1.0, 0.0]), xs=c_in)

    return ys

matrix_rollout(
    jnp.asarray([[0.7, 0.3],
                 [0.3, 0.7]]),
    c_in=jnp.asarray([
        [1.0, 0.0],
    ])
```

Produces

```
UnboundLocalError                         Traceback (most recent call last)
<ipython-input-409-2df128ef5191> in <module>
     12     return ys
     13 
---> 14 matrix_rollout(
     15     jnp.asarray([[0.7, 0.3],
     16                  [0.3, 0.7]]),

<ipython-input-409-2df128ef5191> in matrix_rollout(K, c_in)
      8         return c_next, (c_next,)
      9 
---> 10     _, (ys,) = scan(iteration, init=jnp.asarray([1.0, 0.0]), xs=c_in)
     11 
     12     return ys

numpyro/contrib/control_flow/scan.py in scan(f, init, xs, length, reverse, history)
    397         (length, rng_key, carry), (pytree_trace, ys) = msg['value']
    398 
--> 399     if not msg["kwargs"].get("enum", False):
    400         for msg in pytree_trace.trace.values():
    401             apply_stack(msg)

UnboundLocalError: local variable 'msg' referenced before assignment
```

The relevant source code is [here](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/contrib/control_flow/scan.py#L468). I don't really understand how messengers work, but it looks like it might be an easy fix. I'd be happy to do it myself, but I'm not sure what all this messengers code does...

justinrporter:
Fixes #1024 by returning results of `scan_wrapper` before encountering messages code.


fehiepsi (approved):
Thanks, @justinrporter! Could you add a small regression test in [this file](https://github.com/pyro-ppl/numpyro/blob/master/test/contrib/test_control_flow.py) so we can be notified early if things are not working in the future?

fehiepsi:
@justinrporter Could you update `black` and run `make format` to fix the lint issue?

fehiepsi (approved):

Issue #879, opened 2021-01-16
Support other control flows fori_loop, while_loop, and cond to work with numpyro primitives


fehiepsi:
Currently, we have `scan` implemented in [contrib.control_flow](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/contrib/control_flow/scan.py). After #878, it should be doable to use other control flows in HMC/NUTS. The implementation should mimic what we currently have in `scan` and can be addressed step-by-step:

- [ ] Support while_loop
    - We can assume no primitive in `cond_fn` because any primitive in `cond_fn` can be moved to `body_fn`
    - We can assume body_fn only contains observed nodes because while_loop is not meant to be used to collect stuffs.
    - Because body_fn only contains observed nodes, no need to worry about enumeration.
    - Because body_fn does not have latent variables, no need to worry about substitute/conditon stuffs.
 - [ ] Support fori_loop
    - Any fori_loop can be rewrite as a while_loop. So supporting this should be straightforward.
 - [ ] Support cond?
 - [ ] Support nested while_loop
 - [ ] Support nested scan with no latent variables (but we can collect stuffs)

I think this is an interesting work and not as complicated as `scan`. This is a good chance to get familiar with JAX's control flow. If anyone is interested in addressing this issue, please let me know any question that you have.

tcbegley:
Hey @fehiepsi 

I would be interested in working on some of these as an excuse to learn more about JAX and NumPyro. Do you have an example of what a model using `while_loop` might look like? I feel like that would help give me a bit more context to understand what's required. 

In particular when you say things like "We can assume `body_fn` only contains observed nodes" I'm interested to know the reasoning behind a statement like that. Is that a design choice or does it actually not make sense to collect latent variables like this?

Thanks in advance for any pointers!

fehiepsi:
Hi @tcbegley, back then, the motivation for this PR is to support [compiling some Stan models](https://arxiv.org/abs/1810.00873), which mainly uses `for` loop. But it turns out that #878 seems to be enough for the purpose. For body functions that have latent variables, it is better to use `scan` because we want to collect them (`while_loop`, `fori_loop` are not used to collect values). Back then, I forgot that to make predictions, we also need to collect values. So it does not make sense to use `while_loop` or `fori_loop` for unobserved/observed random variables. Sorry for this!

I think only supporting `cond` makes sense for now. But I'm not sure if there is a need for it. Do you think it is an interesting problem to pursue?

[...]

PR #1028, opened 2021-05-01
Add `cond` primitive

tcbegley:
Hey @fehiepsi,

I had an initial go at implementing a `cond` primitive following the implementation of `scan`. I didn't think about enumeration or anything like that yet, just wanted to get something simple working.

I'm getting an `UnexpectedTracerError` whenever I try to return the execution trace and so far I'm struggling to diagnose the issue. As far as I can tell I _think_ it's the seed handler modifying the `rng_key` that's causing problems, but I'm not entirely sure, and I don't understand why it's causing a problem here when it doesn't appear to in the `scan` primitive.

I'm happy to work on this more and spend more time debugging, but I think I need some pointers so I know where to start looking. If you have any advice or observations that's much appreciated! Also let me know if me getting confused trying to implement this is actually just creating more work for you than implementing it yourself, I don't want to waste your time 🙂 

Here's a minimal example that will produce the error:

```python
import numpyro
import numpyro.handlers as handlers
from numpyro.contrib.control_flow import cond

if __name__ == "__main__":
    with handlers.seed(rng_seed=0):
        res = cond(
            True,
            lambda _: numpyro.sample("x", dist.Normal(20.)),
            lambda _: numpyro.sample("x", dist.Normal(0.)),
            None,
        )
```

See #879


numpyro/contrib/control_flow/cond.py
```diff
@@ -0,0 +1,80 @@
...
+                subs_fn = partial(_subs_wrapper, subs_map)
+                if subs_type == "condition":
+                    seeded_fn = handlers.condition(seeded_fn, condition_fn=subs_fn)
+                elif subs_type == "substitute":
+                    seeded_fn = handlers.substitute(seeded_fn, substitute_fn=subs_fn)
+
+            with handlers.trace() as trace:
+                value = seeded_fn(operand)
+
+        return value, PytreeTrace(trace)
```
tcbegley, line 40: Returning only `value` here and not the `trace` eliminates the error, but I don't understand why yet... 😕

fehiepsi: This seems like a leak in PytreeTrace implementation, where we consider `rng_key` as an auxiliary (non traced) static data. Could you help me add the following logic at [this line](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/contrib/control_flow/scan.py#L39)?
```python
                        if key == "kwargs":
                            kwargs = site["kwargs"].copy()
                            if "rng_key" in kwargs:
                                kwargs["rng_key"] = None
                            aux_trace[name][key] = kwargs
                        else:
                            aux_trace[name][key] = site[key]
```
The reason that I didn't want to trace it is, under `scan`, those `rng_key`s will be collected, hence the output trace will store a batch of `rng_key`s, which does not make sense for a NumPyro `sample` site (which only requires 1 single key rather than a batch of keys). But I didn't anticipate that it leaks to some error like this. To be safe, the above code set `rng_key` to None in the collected trace. :)

fehiepsi:
> about enumeration or anything like that

Yup, please don't worry about it. In case you wanted to address, I think you can add `enum` and `first_available_dim` to `cond_wrapper` and replace
```python
with handlers.block():
    ...
```
by
```python
with handlers.block(), enum(first_available_dim=first_available_dim):
    ...
```
in your `wrap_fn` function. To test enum, you can import `enum` and `config_enumerate` in `numpyro.contrib.funsor` and run
```python
with enum(first_available_dim=-1), config_enumerate(), seed(...):
    out = model()
```

numpyro/contrib/control_flow/cond.py
tcbegley: Thanks! This makes sense.

numpyro/contrib/control_flow/util.py
```diff
@@ -0,0 +1,45 @@
+# Copyright Contributors to the Pyro project.
+# SPDX-License-Identifier: Apache-2.0
+
+from jax.tree_util import register_pytree_node_class
+
+
+@register_pytree_node_class
+class PytreeTrace:
```
tcbegley, line 8: Since this was no longer specific to `scan` I moved it into a separate file, what do you think?

numpyro/contrib/control_flow/cond.py
```diff
@@ -0,0 +1,131 @@
...
+            if rng_key is not None
+            else subs_map
+        )
+        return subs_map(site)
+    return None
+
+
+def wrap_fn(fn, substitute_stack):
+    def wrapper(wrapped_operand):
+        rng_key, operand = wrapped_operand
```
tcbegley, line 29: In `scan_wrapper` the `rng_key` is split inside `body_fn`. I understood that that was because you are iterating and you don't want to use the same key on each iteration. In this case we just need to pass the single key down to the wrapped function, is that correct?

numpyro/contrib/control_flow/cond.py
```diff
@@ -0,0 +1,131 @@
...
+       ...         numpyro.sample("x", dist.Normal(m1, s1))
+       ...
+       ...     def false_fun(_):
+       ...         numpyro.sample("x", dist.Normal(m2, s2))
+       ...
+       ...     cluster = numpyro.sample("cluster", dist.Normal())
+       ...     cond(cluster > 0, true_fun, false_fun, None)
+       >>>
+       >>> svi = SVI(model, guide, numpyro.optim.Adam(1e-2), Trace_ELBO(num_particles=100))
+       >>> params, losses = svi.run(random.PRNGKey(0), num_steps=2500)
```
tcbegley, line 99: Used the example from the linked issue, what do you think?

numpyro/contrib/control_flow/util.py
fehiepsi: Sure, please go ahead.

numpyro/contrib/control_flow/cond.py
fehiepsi: I believe it is correct.

numpyro/contrib/control_flow/cond.py
fehiepsi: Looks great to me. Probably adding some `return` statement to make it more like a generative model, rather than an `effect-handled` model.

numpyro/contrib/control_flow/cond.py
```diff
@@ -0,0 +1,131 @@
...
+
+        .. warning:: This is an experimental utility function that allows users to use
```
fehiepsi, line 101: I think you need to remove 1 block of indentation here and below.

numpyro/contrib/control_flow/cond.py
```diff
@@ -0,0 +1,131 @@
...
+    wrapped_true_fun = wrap_fn(true_fun, substitute_stack)
+    wrapped_false_fun = wrap_fn(false_fun, substitute_stack)
+    wrapped_operand = device_put((rng_key, operand))
+    return lax.cond(pred, wrapped_true_fun, wrapped_false_fun, wrapped_operand)
+
+
+def cond(pred, true_fun, false_fun, operand):
+    """
+    This primitive conditionally applies `true_fun` or `false_fun`. See
+    :func:`jax.lax.cond` for more information.
```
fehiepsi, line 63: Could you add a `note` section to mention that all `sample` sites need to belong to the same distribution class?

fehiepsi (commented):
Awesome work, @tcbegley! Looks great and clean to me! I just have some recommendation for testing:
+ I think we need to test if `substitute` plays nicely with `cond`. You can perform MCMC inference for the current test and see if the collection of samples `x` follows a mixture Normal distribution (by comparing mean/variance).
+ Perform HMC for the case `cluster` is a Bernoulli site. It might work...
+ It is a bit ambiguous to decide dimensions of `scan`'s sites under `plate` but for `cond`, it is clear which dimensions we want to take care of. So it would be nice support (probably in a follow-up PR)
```
with plate(...):
    return cond(...)
```

numpyro/contrib/control_flow/cond.py
tcbegley: Good idea, added!

numpyro/contrib/control_flow/cond.py
tcbegley: Good catch, thanks! Have checked this locally now and I think it's fixed.

numpyro/contrib/control_flow/cond.py
tcbegley: Good suggestion, have added something.

tcbegley:
> I think we need to test if substitute plays nicely with cond. You can perform MCMC inference for the current test and see if the collection of samples x follows a mixture Normal distribution (by comparing mean/variance).

I've included inference with MCMC / NUTS in the test. Did I understand correctly that because MCMC / NUTS uses `condition` internally this is an adequate test of those handlers?

One thing I noticed is that because the target distribution is very bimodal, each chain finds one of the modes and then gets stuck there when doing MCMC. For this reason my test (which looks at mean and std of samples less than 10 and greater than 10 separately) could fail if all four chains get stuck in the same mode. I could either adjust the parameters so that the two normal distributions aren't quite so separated, or I could manually initialise the chains to make it a bit more robust, what do you think?

> Perform HMC for the case cluster is a Bernoulli site. It might work...

This didn't work. Initially it failed because `enum` got added to the kwargs (presumably by an `enum` handler inside the MCMC logic). I made some changes along the lines of what you suggested above, but I think it's going to take a bit more work. Specifically I tried this model

```python
def model():
    def true_fun(_):
        numpyro.sample("x", dist.Normal(20.0))

    def false_fun(_):
        numpyro.sample("x", dist.Normal(0.0))

    cluster = numpyro.sample("cluster", dist.Bernoulli(0.5))
    cond(cluster == 1, true_fun, false_fun, None)
```

With enumeration `cluster == 1` becomes `DeviceArray([False  True], dtype=bool)` which causes an error because `lax.cond` expects a scalar argument to `pred`. I think supporting enumeration will require a separate `cond_enum`, which perhaps could check if the shape of `pred` is in agreement with the shape of the enumeration dimensions?

I think this is going to require a bit more thought. Happy to work on it, but up to you whether we leave it for a follow-up PR?

> It is a bit ambiguous to decide dimensions of scan's sites under plate but for cond, it is clear which dimensions we want to take care of. So it would be nice support (probably in a follow-up PR)

That would be both interesting to implement and useful I think, but I agree that perhaps it's best to leave that for a separate PR.

fehiepsi:
> Did I understand correctly that because MCMC / NUTS uses condition internally this is an adequate test of those handlers?

You are right. It might be better to test the `substitute` handler directly but not necessary

> because the target distribution is very bimodal
> adjust the parameters so that the two normal distributions aren't quite so separated

Good catch! I like the idea of bringing the modes closer to each other. 

> enumeration cluster == 1 becomes DeviceArray([False True], dtype=bool) 

Oops... this is tricky. I don't think there's an easy way to resolve this issue. :( Could you add a warning to the docs and leave it for a future PR? I think `plate` also suffers from the same issue when `cond` condition is a batch of booleans...

tcbegley:
Have made the suggested changes:
- Less separated modes in the MCMC + `cond` test to ensure (some) mixing.
- Added a warning to the `cond` docstring. 

I also added `enum` and `first_available_dim` as keyword arguments to `cond_wrapper` and raise a `RuntimeError` with a more informative error message if `enum` is `True`.

Since enumeration and plate support are being left for subsequent PRs, I think that means this is ready for a final review.

test/contrib/test_control_flow.py
```diff
@@ -132,3 +132,65 @@ def iteration(c_prev, c_in):
...
+        num_chains=4,
+        chain_method="sequential",
+    )
+    mcmc.run(random.PRNGKey(0))
+
+    x = mcmc.get_samples()["x"]
+    assert x.shape == (10_000,)
+    assert_allclose(
+        [x[x > 2.5].mean(), x[x > 2.5].std(), x[x < 2.5].mean(), x[x < 2.5].std()],
+        [5.0, 1.0, 0.0, 1.0],
```
fehiepsi, line 194: Could you also check for mixture mean and mixture std: `x.mean() vs 2.5`, `x.std() vs sqrt(2)`?

fehiepsi (dismissed):
Looks great, thanks for addressing my comments @tcbegley!

test/contrib/test_control_flow.py
tcbegley: Yep, added, though I made the mixture std `sqrt(29 / 4)`.

```
E(X^2) = 0.5 * (E(X^2 | c = 0) + E(X^2 | c = 1)) = 0.5 * (1 + (1 + 25)) = 27 / 2
```

Then subtract off `2.5^2`. Do you agree? I checked this empirically with NumPy.

fehiepsi: You are right! I calculated that term wrongly. Thanks!

tcbegley:
Had to move the modes a bit closer again as mixing was still not great and samples were biased towards one mode leading to a bad estimate of the mixture mean. All seems to be ok now.

fehiepsi (approved):

fehiepsi:
@tcbegley The master branch is broken after this merged. I guess testing for the mean is a bit unstable. Could we separate it into another assertation with higher tolerate?

tcbegley:
Ah sorry about that, yes I can split that out and increase the tolerance. Will submit a PR shortly.

Btw, the test worked for me locally, and also worked on the branch. Since the RNGKeys are fixed I didn't think the results could vary like that? Have I misunderstood something?

fehiepsi:
I think that's because jax devs just released a new version :D

PR #1029, opened 2021-05-04
Make cond test more robust

tcbegley:
The target distribution is bimodal which means MCMC chains sometimes get stuck in a particular mode and bias the mean. This PR increases the tolerance to reduce the chance of the test failing.


fehiepsi (approved):

fehiepsi:
Thank you, @tcbegley !

PR #1031, opened 2021-05-07
[WIP] adding init_to_mean

vitkl:
Hi! This PR adds init_to_mean function - also not sure where to add tests. What do you think?

### DIFF
--- a/numpyro/infer/initialization.py
+++ b/numpyro/infer/initialization.py
@@ -35,6 +35,26 @@ def init_to_median(site=None, num_samples=15):
         except NotImplementedError:
             return init_to_uniform(site)
 
+def init_to_mean(site=None):
+    """
+    Initialize to the prior mean; fallback to median if mean is undefined.
+    """
+    if site is None:
+        return partial(init_to_mean)
+
+    try:
+        # Try .mean() method.
+        if site['type'] == 'sample' and not site['is_observed'] and not site['fn'].is_discrete:
+            value = site["fn"].mean
+            # if jnp.isnan(value):
+            #    raise ValueError
+            if hasattr(site["fn"], "_validate_sample"):
+                site["fn"]._validate_sample(value)
+            return np.array(value)
+    except (NotImplementedError, ValueError):
+        # Fall back to a median.
+        # This is required for distributions with infinite variance, e.g. Cauchy.
+        return init_to_median(site)
 
 def init_to_sample(site=None):
     """


numpyro/infer/initialization.py
```diff
@@ -35,6 +35,26 @@ def init_to_median(site=None, num_samples=15):
...
+    """
+    Initialize to the prior mean; fallback to median if mean is undefined.
+    """
+    if site is None:
+        return partial(init_to_mean)
+
+    try:
+        # Try .mean() method.
+        if site['type'] == 'sample' and not site['is_observed'] and not site['fn'].is_discrete:
+            value = site["fn"].mean
```
fehiepsi, line 48: Could you also add
```python
sample_shape = site["kwargs"].get("sample_shape")
if sample_shape:
    value = jnp.broadcast_to(value, sample_shape + jnp.shape(value))
```

numpyro/infer/initialization.py
```diff
@@ -35,6 +35,26 @@ def init_to_median(site=None, num_samples=15):
...
+            # if jnp.isnan(value):
+            #    raise ValueError
```
fehiepsi, line 50: this is not needed because it is addressed below

numpyro/infer/initialization.py
```diff
@@ -35,6 +35,26 @@ def init_to_median(site=None, num_samples=15):
...
+            if hasattr(site["fn"], "_validate_sample"):
+                site["fn"]._validate_sample(value)
```
fehiepsi, line 52: in numpyro, we raise a warning for invalid values, so I guess you can copy the logic [here](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/distributions/distribution.py#L291) and instead of raising a warning, you can raise a ValueError, so it will be captured in the `except` statement below.

numpyro/infer/initialization.py
```diff
@@ -35,6 +35,26 @@ def init_to_median(site=None, num_samples=15):
...
+            return np.array(value)
```
fehiepsi, line 53: I guess you can simply return `value`. Using `np.array` here might cause some jax issues if value is traced under `jit`.

fehiepsi (commented):
Thanks, @vitkl! For testing, you can add `init_to_mean` [here](https://github.com/pyro-ppl/numpyro/blob/master/test/infer/test_infer_util.py#L216) and the test below it.

fehiepsi:
Hi @vitkl, just a head up, we are going to make the new release. Do you want to incorporate this feature?

vitkl:
Sorry, swamped with other projects and I need some time to understand your suggestions.

fehiepsi:
Yup, there is no rush at all. :)

fehiepsi:
@vitkl Can I push a commit on the top of this PR to have this feature in the next release?

vitkl:
Hi @fehiepsi 

Yes absolutely!

fehiepsi:
Closed in favor of https://github.com/pyro-ppl/numpyro/pull/1550.

Issue #1023, opened 2021-04-23
feature request: please wrap optax optimizers


murphyk:
The [optax library from Deepmind](https://github.com/deepmind/optax) is the future for jax ecosystem (e.g. it will be adopted by Flax team), and will likely replace [jax.experimental.optimizers](https://jax.readthedocs.io/en/latest/jax.experimental.optimizers.html). It would be good to update https://num.pyro.ai/en/stable/optimizers.html accordingly.

fehiepsi:
Nice suggestion! I think we can add a new `numpyro.contrib.optax` module which "automatically" wraps `optax` optimizers and make them compatible with SVI. We can also add some helper methods in SVI or ELBO to make them compatible with `optax` directly.

fehiepsi:
This is addressed in #1034. @murphyk @tcbegley thanks for requesting this feature and resolving it!

PR #1034, opened 2021-05-08
Add support for Optax optimizers

tcbegley:
This PR adds support for Optax optimizers / gradient transformations by adding a simple wrapper class to make them compatible with `SVI`. The wrapper class is very closely related to `numpyro.optim._NumPyroOptim`, with some minor differences. This doesn't require any changes to `SVI`, but I did add a line to the constructor that will automatically wrap any `optax.GradientTransformation` that is passed in as the optimizer.

Currently I've only implemented the wrapper, but if you think it would be useful we could create Optax versions of classes like `numpyro.optim.Adam` etc.?

Let me know what you think of the approach I took, and whether the test coverage is sufficient. Probably the main thing to think about is whether Optax should be a dependency of NumPyro, and if not how to make sure it's only imported if installed etc.

Here's a very small example

```python
import numpyro
import optax
from jax import random
from numpyro.infer.autoguide import AutoNormal

def model():
    numpyro.sample("mu", dist.Normal(3, 2))

svi = SVI(
    model, 
    AutoNormal(model), 
    optax.chain(optax.clip(10.0), optax.adam(1e-3)), 
    Trace_ELBO(num_particles=100),
)
params, losses = svi.run(random.PRNGKey(0), num_steps=5000)```
```

See #1023


fehiepsi:
This work is awesome, @tcbegley! I would like to share some thoughts for further discussion:
+ my first impression is the `_OptaxWrapper` class can be (heuristically) simplified to
```python
def optax_to_numpyro(transformation):
    def init(params):
        return (params, transformation.init(params))

    def update(params):
        updates, opt_state = self.transformation.update(g, opt_state, params)
        updated_params = optax.apply_updates(params, updates)
        return ...

    def get_params(state):
        return state[0]

    return _NumPyroOptim(lambda x, y, z: (x, y, z), init, update, get_params)
```
This will reduce the maintenance overhead if we want to change something in `_NumPyroOptim`.

+ It might be not a good idea to modify the input `optim` in SVI constructor. Do you think we can detect if the `optim` input is an optax object, without having to import the module? Otherwise, we can add a helper function to `numpyro.optim`, and call
```
svi = SVI(
    model, 
    AutoNormal(model), 
    numpyro.optim.from_optax(optax.chain(optax.clip(10.0), optax.adam(1e-3))), 
    Trace_ELBO(num_particles=100),
)
```
or `numpyro.contrib.optim.from_optax` (to be more explicit that this is an optional functionality).

tcbegley:
> my first impression is the _OptaxWrapper class can be simplified

Wow that's much neater. 🤯  Let's do that instead!

> Do you think we can detect if the optim input is an optax object, without having to import the module?

I was thinking a bit about this. One idea is that inside `numpyro.infer.svi` you do something like

```python
try:
    import optax
except ImportError:
    optax = None
```

Then later we could do something like

```python
if isinstance(optim, numpyro.optim._NumPyroOptim):
    self.optim = optim
elif optax is not None and isinstance(optim, optax.GradientTransformation):
    self.optim = optax_to_numpyro(optim)
else:
    raise ValueError(...)
```

The alternative would be inspecting `type(optim)` but I think that is a bit hacky / brittle.

tcbegley:
Oh, I also had a question, where is the best place to put this in the docs do you think?

tcbegley:
I implemented your simplified wrapper and had a go at making the optax import in SVI module optional. There's actually two versions of it on the last two commits, in case you have a preference for one or the other. I think latest version is probably a bit neater.

I initially thought that this was responsible for the failing build but just saw that sphinx 4.0.0 was released a few hours ago so now I think that's too blame, the docs are building fine for me locally with an older version of Sphinx.

test/contrib/test_optax.py
```diff
@@ -0,0 +1,106 @@
...
+import pytest
+
+from jax import grad, jit, partial, random
+from jax.lax import fori_loop
+import jax.numpy as jnp
+from jax.test_util import check_close
+import optax
+
+import numpyro
+from numpyro.contrib.optax import optax_to_numpyro
```
fehiepsi, line 14: Could you move imports `optax, optax_to_numpyro` to the content of the tests, like [this](https://github.com/pyro-ppl/numpyro/blob/673880c8c2c50f1bc291af9115fe46e8876554d5/test/contrib/test_tfp.py#L23). This will be helpful when collecting tests for those who do not have `optax`. A better solution is to add
```
try:
    import optax
    from numpyro.contrib.optim import optax_to_numpyro
except ImportError:
    pytestmark = pytest.mark.skip()
```

numpyro/contrib/optax.py
```diff
@@ -0,0 +1,47 @@
...
+
+import optax
+
+from numpyro.optim import _NumPyroOptim
+
+_Params = TypeVar("_Params")
+_OptState = TypeVar("_OptState")
+_State = Tuple[_Params, _OptState]
+
+
```
fehiepsi, line 19: nit: probably renaming the file from `optax` -> `optim`

numpyro/infer/svi.py
```diff
@@ -80,7 +81,8 @@ class SVI(object):
     :param model: Python callable with Pyro primitives for the model.
     :param guide: Python callable with Pyro primitives for the guide
         (recognition network).
-    :param optim: an instance of :class:`~numpyro.optim._NumpyroOptim`.
+    :param optim: an instance of :class:`~numpyro.optim._NumpyroOptim` or an Optax
+        ``GradientTransformation``.
```
fehiepsi, line 85: I think you can add a small snippet here. :) Like
```
from optax import ...

svi = SVI(model, guide, ..., loss)
```

fehiepsi (commented):
The current version is pretty clean to me. Regarding sphinx, I think you can pin its version to <4. readthedocs still uses sphinx 1 (or 2) so we don't need to update.

tcbegley:
Thanks for the comments @fehiepsi, I think I've addressed them all.

One more thing to check with you: I added an entry to the docs but I'm not convinced I added it in the right location. In particular it appears in the sidebar but the entry itself is at the bottom of the page. I think I probably want to either move it into its own file or change the heading level or something? Anyway, let me know where you think it belongs and I'll happily move it.

fehiepsi (approved):
LGTM! The api looks so cool. Regarding the doc section, I think you can change it to follows

numpyro/contrib/optim.py
```diff
@@ -0,0 +1,47 @@
...
+"""
+
+from typing import Tuple, TypeVar
+
+import optax
+
+from numpyro.optim import _NumPyroOptim
+
+_Params = TypeVar("_Params")
+_OptState = TypeVar("_OptState")
```
fehiepsi, line 16: I guess you can use `optax.OptState` for this.

tcbegley:
Thanks for reviewing again @fehiepsi, have addressed your comments.

The docs build was failing again, this time it was a new version of Jinja2 (version 3.0.0 of Jinja2 to be precise... 🤔), I pinned the version in the requirements as with Sphinx, but maybe a longer term solution is needed at some point.

fehiepsi (approved):
This is pretty cool! I have been thinking about how to support new optimizers in jax without having to check the source occasionally and you gave me the solution. Thanks!

PR #1040, opened 2021-05-25
explicit num_warmup, num_samples arguments for MCMC

### LINKED ISSUE #1038: order of args to mcmc is inconsistent between pyro and numpyro
I see that [pyro.mcmc](https://num.pyro.ai/en/stable/mcmc.html) uses (kernel, num_samples, warmup_steps)
but [numpyro.mcmc](https://num.pyro.ai/en/stable/mcmc.html) uses (kernel, warmup_steps, num_samples), which is a bit confusing.

fehiepsi:
Resolves #1038.

This might break things but in the long run, I think this requirement is good. Using only the numbers `MCMC(kernel, 500, 500)` is a bit confusing and should be discouraged.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -4,7 +4,7 @@ funsor
 jax>=0.1.65
 jaxlib>=0.1.45
 optax==0.0.6
-nbsphinx>=0.8.4
+nbsphinx==0.8.1  # newer nbsphinx version breaks github links in tutorial pages
 sphinx-gallery
 tfp-nightly  # TODO: change this to tensorflow-probability when it is stable
 tqdm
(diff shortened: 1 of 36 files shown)


fritzo (approved):
LGTM, indeed this seems like the most conservative resolution to #1038.

fehiepsi:
Thanks for reviewing! :)

alexander-maier:
Is this right? Calling
```python
compat.infer.MCMC(kernel, num_warmup=100, num_samples=1000)
```
throws an error since no kwargs are handed to `mcmc.MCMC`. Also, shouldnt the correct `compat` call be
```python
compat.infer.MCMC(kernel, warmup_steps=100, num_samples=1000)?
```
Thanks

fehiepsi:
Oh, we haven't updated the compat module for this change. It should have the later signatures. Could you make a separate issue for this @alexander-maier ?

alexander-maier:
#1321

PR #1042, opened 2021-05-27
Add some png thumnail images for examples/tutorials

### LINKED ISSUE #847: Add images to examples/tutorials
Currently, the [tutorial page](http://num.pyro.ai/en/latest/) only has thumbnail images for 13 among 21 examples/tutorials. It would be nice to add thumbnail images for the remaining examples. The images do not need to be generated by the scripts. We can use something relevant to the content. For example, [plate notation](https://en.wikipedia.org/wiki/Plate_notation) of one of the models. Or for `annotation` example, some image relevant to crowdsourcing would be nice.

The following tasks track thumbnail images for the remaining examples

- [ ] Baseball
- [ ] VAE
- [ ] Annotation
- [ ] Enum HMM
- [ ] Capture recapture
- [ ] Bayesian imputation
- [ ] Sparse regression
- [ ] Proportion test

fehiepsi:
Resolves #847. I don't have good ideas for some examples like baseball or sparse regression, so I just use texts for those thumbnail images.


martinjankowiak (approved):
👍

PR #1043, opened 2021-05-27
Illustrate the usage of infer_discrete in the annotation example

fehiepsi:
It seems that we have some `infer_discrete` bug in the master branch. I merged changes in #991 and it works for models in this example.

Changes in this PR:
- refactor infer_discrete following #991
- deprecate `replace` handler's `guide_trace` argument (to be consistent with Pyro)

Questions for reviewers: I found that it is tricky for users to perform `infer_discrete` correctly to get extra posterior samples for discrete latent sites. Should we make some helpers like `MCMC.infer_discrete` or `Predictive.infer_discrete` that does the job in this `annotation` example under the hood?
```python
    def infer_discrete_model(rng_key, samples):
        conditioned_model = handlers.condition(model, data=samples)
        infer_discrete_model = infer_discrete(
            config_enumerate(conditioned_model), rng_key=rng_key
        )
        with handlers.trace() as tr:
            infer_discrete_model(*data)

        return {
            name: site["value"]
            for name, site in tr.items()
            if site["type"] == "sample" and site["infer"].get("enumerate") == "parallel"
        }

    posterior_samples = mcmc.get_samples()
    discrete_samples = vmap(infer_discrete_model)(
        random.split(random.PRNGKey(1), args.num_samples), posterior_samples
    )
```
I imagine that it would be convenient if we have something like
```
mcmc = MCMC(...)
mcmc.run()
mcmc.infer_discrete(rng_key, *args, first_available_dim=None, temperature=1, **kwargs)
posterior_samples = mcmc.get_samples()  # this will include discrete latent sites
```


numpyro/contrib/funsor/discrete.py
```diff
@@ -97,27 +57,14 @@ def _sample_posterior(
...
-    with funsor.interpretations.lazy:
-        log_prob = funsor.sum_product.sum_product(
-            sum_op,
-            prod_op,
-            list(terms["log_factors"].values()) + list(terms["log_measures"].values()),
-            eliminate=terms["measure_vars"] | terms["plate_vars"],
-            plates=terms["plate_vars"],
-        )
-        log_prob = funsor.optimizer.apply_optimizer(log_prob)
+    with funsor.adjoint.AdjointTape() as tape:
```
eb8680, line 60: Does using `lazy` here as in the old version and `funsor.adjoint.adjoint` below not work? I see we made this change in #991 but can't recall why.

numpyro/contrib/funsor/discrete.py
```diff
@@ -97,27 +57,14 @@ def _sample_posterior(
...
+        with block(), enum(first_available_dim=first_available_dim):
+            log_prob, model_tr, log_measures = _enum_log_density(
+                model, args, kwargs, {}, sum_op, prod_op
+            )
 
     with approx:
-        approx_factors = funsor.adjoint.adjoint(sum_op, prod_op, log_prob)
+        approx_factors = tape.adjoint(sum_op, prod_op, log_prob)
```
eb8680, line 67: Does using the more functional interface `funsor.adjoint.adjoint` not work here?

eb8680 (approved):
Looks good to me.

> Should we make some helpers like MCMC.infer_discrete or Predictive.infer_discrete that does the job in this annotation example under the hood?

Yes - in fact, this should probably be the default behavior of `Predictive` for HMC/NUTS when `infer_discrete` is finished.

numpyro/contrib/funsor/discrete.py
fehiepsi: Sorry for the late response, @eb8680! Yes, for some reason there is an error happens at this line.
```
  File "/home/fehiepsi/numpyro/numpyro/contrib/funsor/discrete.py", line 120, in _sample_posterior
    approx_factors = funsor.adjoint.adjoint(sum_op, prod_op, log_prob)
  File "/home/fehiepsi/funsor/funsor/adjoint.py", line 135, in adjoint
    return tape.adjoint(sum_op, bin_op, root)
  File "/home/fehiepsi/funsor/funsor/adjoint.py", line 115, in adjoint
    in_adjs = adjoint_ops(fn, sum_op, bin_op, adjoint_values[output], *inputs)
  File "/home/fehiepsi/funsor/funsor/registry.py", line 106, in __call__
    return self[key](*args)
  File "/home/fehiepsi/funsor/funsor/registry.py", line 63, in __call__
    return self.partial_call(*args)(*args)
  File "/home/fehiepsi/funsor/funsor/adjoint.py", line 209, in adjoint_contract_generic
    assert len(terms) == 1 or len(terms) == 2
AssertionError
```

fehiepsi:
Thanks for helping me review this PR, @eb8680!!

PR #1044, opened 2021-05-27
Add TraceGraph_ELBO

### LINKED ISSUE #896: [FR] Implement a surrogate loss for discrete, e.g. categorical, distributions
As discussed in [this thread](https://forum.pyro.ai/t/using-transformreparam-with-to-event/2553/11) right now using SVI in combination with categorical distributions, e.g. used in gaussian mixture models, leads to a wrong calculation of ELBO gradients, and thus unexpected results might be the consequence. 

In order to fix this, one approach as explained in the [NumPyro's SVI docs](http://pyro.ai/examples/svi_part_iii.html#Tricky-Case:-Non-reparameterizable-Random-Variables) is using a surrogate loss.

fehiepsi:
Resolves #896, comparing to [Pyro implementation](https://github.com/pyro-ppl/pyro/blob/dev/pyro/infer/tracegraph_elbo.py)

TODO
- [x] implement `_compute_downstream_costs`
- [x] use the downstream cost to compute the surrogate loss
- [x] add tests


numpyro/infer/elbo.py
```diff
@@ -280,3 +282,266 @@ def single_particle_elbo(rng_key):
...
+    for name, node in trace.items():
+        if node["type"] == "sample":
+            succ[name] = set()
+    for name, node in trace.items():
+        if node["type"] == "sample":
+            for past_name, past_node in trace.items():
+                if past_node["type"] == "sample":
+                    if past_name == name:
+                        break
+                    # XXX: different from Pyro, we always add edge past_name -> name
```
fehiepsi, line 363: This is the only difference comparing to [Pyro implementation](https://github.com/pyro-ppl/pyro/blob/4a61ef2f9050ef81d1b0aa148d14ecc876f24a51/pyro/poutine/trace_messenger.py#L26).

numpyro/infer/util.py
```diff
@@ -68,6 +69,49 @@ def log_density(model, model_args, model_kwargs, params):
...
+    Stop gradient for samples at latent sample sites that has_rsample=False.
+    """
+
+    def postprocess_message(self, msg):
+        if (
+            msg["type"] == "sample"
+            and (not msg["is_observed"])
+            and (not msg["fn"].has_rsample)
+        ):
+            msg["value"] = lax.stop_gradient(msg["value"])
```
fehiepsi, line 83: We can distinguish `sample` and `rsample` in Distribution but that might require many changes. So I made this handler to stop gradient for samples at sites with `has_rsample=False`.

martinjankowiak:
@fehiepsi i'll try to review this in the next few days. it's a big PR so please have some patience : )

fehiepsi:
Thanks, @martinjankowiak! There is no rush! :P

numpyro/infer/elbo.py
```diff
@@ -280,3 +281,266 @@ def single_particle_elbo(rng_key):
...
+    :class:`TraceGraph_ELBO`.
+    """
+    return {
+        name: [f for f in node["cond_indep_stack"]]
+        for name, node in trace.items()
+        if node["type"] == "sample"
+    }
+
+
+class MultiFrameTensor(dict):
```
martinjankowiak, line 300: can you please confirm which chunks of code were borrowed verbatim apart from torch/jnp edits?

numpyro/infer/elbo.py
martinjankowiak: can you please clarify?

numpyro/infer/elbo.py
```diff
@@ -280,3 +281,266 @@ def single_particle_elbo(rng_key):
...
+    used to reduce variance:
+
+    - the sequential order of samples (z is sampled after y => y does not depend on z)
+    - :class:`~numpyro.plate` generators
+
+    References
+
+    [1] `Gradient Estimation Using Stochastic Computation Graphs`,
+        John Schulman, Nicolas Heess, Theophane Weber, Pieter Abbeel
+    [2] `Neural Variational Inference and Learning in Belief Networks`
```
martinjankowiak, line 479: you can remove ref [2] since no baseline support

numpyro/infer/elbo.py
```diff
@@ -280,3 +281,266 @@ def single_particle_elbo(rng_key):
...
+            model_trace, guide_trace = get_importance_trace(
+                seeded_model, seeded_guide, args, kwargs, param_map
+            )
+
+            # XXX: different from Pyro, we don't support baseline_loss here
+            non_reparam_nodes = {
+                name
+                for name, site in guide_trace.items()
+                if site["type"] == "sample"
+                and (site["is_observed"] or not site["fn"].has_rsample)
```
martinjankowiak, line 515: shouldn't this be `not site["is_observed"] and not site["fn"].has_rsample`? that is the defn in pyro

numpyro/infer/util.py
```diff
@@ -68,6 +68,52 @@ def log_density(model, model_args, model_kwargs, params):
...
+    (EXPERIMENTAL) Returns traces from the guide and the model that is run against it.
+    The returned traces also store the log probability at each site.
+
+    .. note:: Gradients are blocked at latent sites which do not have reparametrized samplers.
+    """
+    guide = substitute(guide, data=params)
+    with _without_rsample_stop_gradient():
+        guide_trace = trace(guide).get_trace(*args, **kwargs)
+    model = substitute(replay(model, guide_trace), data=params)
+    with _without_rsample_stop_gradient():
```
martinjankowiak, line 97: is this second `_without_rsample_stop_gradient` necessary?

test/infer/test_svi.py
```diff
@@ -232,3 +232,59 @@ def guide():
...
+    data.append(jnp.array([0.0, 0.4]))
+    data.append(jnp.array([0.2, 0.5]))
+    data.append(jnp.array([0.1, 0.7]))
+    n_data = len(data)
+    sum_data = data[0] + data[1] + data[2] + data[3]
+    analytic_lam_n = lam0 + n_data * lam
+    analytic_log_sig_n = -0.5 * jnp.log(analytic_lam_n)
+    analytic_loc_n = sum_data * (lam / analytic_lam_n) + loc0 * (lam0 / analytic_lam_n)
+
+    class Normal(dist.Normal):
```
martinjankowiak, line 254: call this `FakeNormal` for clarity?

test/infer/test_svi.py
```diff
@@ -232,3 +232,59 @@ def guide():
     svi = SVI(model, guide, optim.Adam(1), Trace_ELBO())
     with pytest.warns(UserWarning, match="SVI does not support models with discrete"):
         svi.run(random.PRNGKey(0), 10)
+
+
+def test_normal_normal():
```
martinjankowiak, line 237: is this a sufficient integration test? are there more tests that can be borrowed from pyro apart from the complex pyramid test? maybe the chain test? wdyt?

numpyro/infer/elbo.py
fehiepsi: I don't actually understand what's the role of `counter` in [CondIndepStackFrame](https://github.com/pyro-ppl/pyro/blob/4a61ef2f9050ef81d1b0aa148d14ecc876f24a51/pyro/poutine/indep_messenger.py#L15). I just guess that it is used for `iplate` or something like that. I assume `query.counter == target.counter == 0`, so the `if` branch in Pyro implementation does not trigger. Do I understand correctly?

numpyro/infer/elbo.py
```diff
@@ -280,3 +281,266 @@ def single_particle_elbo(rng_key):
...
+
+    def __init__(self, *items):
+        super().__init__()
+        self.add(*items)
+
+    def add(self, *items):
+        """
+        Add a collection of (cond_indep_stack, tensor) pairs. Keys are
+        ``cond_indep_stack``s, i.e. tuples of :class:`CondIndepStackFrame`s.
+        Values are :class:`torch.Tensor`s.
```
fehiepsi, line 323: typo: ndarray

numpyro/infer/elbo.py
fehiepsi: yes, it is just the same here: only torch -> jnp

numpyro/infer/elbo.py
martinjankowiak: yes counter is only for non-vectorized plates. i see you're saying you don't need to accommodate within-non-vectorized-plate-independence

numpyro/infer/util.py
fehiepsi: You are right, it seems to not required here.

test/infer/test_svi.py
fehiepsi: There are some other tests for [trace graph elbo](https://github.com/pyro-ppl/pyro/blob/dev/tests/integration_tests/test_tracegraph_elbo.py) but they require baseline. I think without baseline things will still work, so I'll a couple more.

numpyro/infer/elbo.py
```diff
@@ -280,3 +281,266 @@ def single_particle_elbo(rng_key):
...
+                    elbo = elbo + jnp.sum(site["log_prob"])
+            for name, site in guide_trace.items():
+                if site["type"] == "sample":
+                    log_prob_sum = jnp.sum(site["log_prob"])
+                    if name in non_reparam_nodes:
+                        surrogate = jnp.sum(
+                            site["log_prob"] * stop_gradient(downstream_costs[name])
+                        )
+                        log_prob_sum = (
+                            stop_gradient(log_prob_sum + surrogate) - surrogate
```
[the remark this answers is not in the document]
fehiepsi: I was confused too. Here we want to return the (not surrogate) elbo loss so this is either
```
                            stop_gradient(log_prob_sum - surrogate) + surrogate
```
or
```
                            stop_gradient(log_prob_sum + surrogate) - surrogate
```
According to [Pyro implementation](https://github.com/pyro-ppl/pyro/blob/dev/pyro/infer/tracegraph_elbo.py#L196), we need to add this `surrogate` term to the elbo, so I think 
```python
                    log_prob_sum = stop_gradient(log_prob_sum + surrogate) - surrogate
                    elbo = elbo - log_prob_sum
```
is correct and has verified that if I switch the signs here, normal normal test will fail (actually, I can confirm that losses match Pyro implementation with the same initial params in both backends).

numpyro/infer/elbo.py
fehiepsi: Good catch! Here is [pyro implementation](https://github.com/pyro-ppl/pyro/blob/0decbe54f8660cbf453ef4915495d77f0d7fb454/pyro/poutine/trace_struct.py#L293). I guess what I wanted to code is `and not (site["is_observed"] or site["fn"].has_rsample)`. Thanks for pointing this out.

test/infer/test_svi.py
martinjankowiak: yeah these low dimensional examples should work fine without baselines (although they will require more steps)

martinjankowiak: https://github.com/pyro-ppl/pyro/blob/d7687ae0f738bd81a792dabbb18a53c0fce73765/tests/integration_tests/test_conjugate_gaussian_models.py

martinjankowiak: maybe this is a good test to port: https://github.com/pyro-ppl/pyro/blob/0decbe54f8660cbf453ef4915495d77f0d7fb454/tests/integration_tests/test_tracegraph_elbo.py#L331

fehiepsi:
Thanks for helping me review this large PR, @martinjankowiak! I have added a couple of tests, including the gaussian chain. For RaoBlackwellizationTests, I would like to skip because the Pyro version uses iplate.

test/infer/test_svi.py
```diff
@@ -232,3 +233,267 @@ def guide():
...
+    assert_allclose(alpha_error, 0, atol=0.04)
+    assert_allclose(beta_error, 0, atol=0.04)
+
+
+@pytest.mark.parametrize(
+    "num_latents,num_steps,step_size,atol,difficulty",
+    [
+        (3, 5000, 0.003, 0.05, 0.6),
+        (5, 6000, 0.003, 0.05, 0.6),
+        (7, 8000, 0.003, 0.05, 0.6),
```
fehiepsi, line 386: The numbers of steps are higher than the Pyro version because we don't use baseline here.

numpyro/infer/elbo.py
```diff
@@ -280,3 +281,265 @@ def single_particle_elbo(rng_key):
         renyi_elbo = avg_log_exp / (1.0 - self.alpha)
         weighted_elbo = jnp.dot(stop_gradient(weights), elbos) / self.num_particles
         return -(stop_gradient(renyi_elbo - weighted_elbo) + weighted_elbo)
+
+
+def _get_plate_stacks(trace):
+    """
+    This builds a dict mapping site name to a set of plate stacks. Each
+    plate stack is a list of :class:`CondIndepStackFrame`s corresponding to
+    an :class:`plate`. This information is used by :class:`Trace_ELBO` and
```
martinjankowiak, line 290: not typo: an -> a

numpyro/infer/util.py
```diff
@@ -68,6 +68,51 @@ def log_density(model, model_args, model_kwargs, params):
     return log_joint, model_trace
 
 
+class _without_rsample_stop_gradient(numpyro.primitives.Messenger):
+    """
+    Stop gradient for samples at latent sample sites that has_rsample=False.
```
martinjankowiak, line 73: nit: that => for which

numpyro/infer/util.py
martinjankowiak: what are `intermediates` here and elsewhere?

test/infer/test_svi.py
```diff
@@ -232,3 +233,267 @@ def guide():
...
+    ],
+)
+def test_tracegraph_gaussian_chain(num_latents, num_steps, step_size, atol, difficulty):
```
martinjankowiak, line 389: 👍 for porting. this is tricky code so it's great to have a mix of unit and integration tests for extra confidence

martinjankowiak (approved):

numpyro/infer/util.py
fehiepsi: Those are intermediate values of transformed distributions, performed [here](https://github.com/pytorch/pytorch/blob/7949fdd2b64faef1d90510784cb57190dcf87c69/torch/distributions/transformed_distribution.py#L117). In Pyro, they are cached in transforms class. In NumPyro, we store them in `intermediates` field. We stop gradient for them to match PyTorch behavior (but I'm not sure if PyTorch behavior is correct - i.e. I'm not sure if we need to stop gradient for intermediate values).

martinjankowiak:
for example try running this:

```
import torch
from torch.distributions import transforms
from pyro.distributions.testing.fakes import NonreparameterizedNormal
from pyro.distributions import Normal, TransformedDistribution


def print_grad(base_dist):
    torch.manual_seed(0)
    mean = torch.tensor(0.0, requires_grad=True)
    d = TransformedDistribution(base_dist(mean, 1.0), transforms=transforms.ExpTransform())
    z = d.rsample() if d.has_rsample else d.sample()
    z.pow(2.0).backward()
    print("mean.grad", mean.grad)

print_grad(Normal)
print_grad(NonreparameterizedNormal)
```

numpyro/infer/util.py
fehiepsi: @fritzo I think that you made some changes in [this PR](https://github.com/pytorch/pytorch/pull/5086/files#diff-8b108de0fae81bfc4f939e40322885653fc3c957b2f8ad0ff25aafc16bf2c639L56-R60). What do you think?

fritzo: > What do you think?

Exactly what question are you asking?  I believe [the PR you mentioned](https://github.com/pytorch/pytorch/pull/5086/files#diff-8b108de0fae81bfc4f939e40322885653fc3c957b2f8ad0ff25aafc16bf2c639L56-R60) ensures that Pyro's `TransformedDistribution.sample()` is not accidentally reparametrized when the base dist is nonreparametrized but the transforms are differentiable.  I don't know enough about this PR #1044 or NumPyro's rsample design to say any more.

fehiepsi: Let me give an example. Assume the distribution that we are interested in is: `TransformedDistribution(FakeNormal(0, 1), [AffineTransform(0, scale), AffineTransform(0, 2)])`, where `scale` is a learnable parameter. To draw a sample of that distribution, currently in Pyro, we return
```
x = FakeNormal().sample().detach()
y = 2 * scale.detach() * x
return y
```
Before that PyTorch PR, we return `2 * scale * x`.

I guess we are interested in `log_prob(y)`, which is calculated by
```python
FakeNormal().log_prob(x) + log(2) + log(scale)
```
in both situations. Now, assume that, `AffineTransform(0, scale).log_abs_det_jacobian(x, y)` is implemented by `return log(y / x)` rather than `return scale`. Then `log_prob(y)` will be calculated by
```
FakeNormal().log_prob(x) + log(2) + log(scale)
```
if `y = 2 * scale * x`, and will be calculated by
```
FakeNormal().log_prob(x) + log(2) + log(scale.detach())
```
if `y = 2 * scale.detach() * x`. Now two implementations give two different gradients w.r.t. `scale` parameter. I was a bit confused about that. _I'm not sure_ if that change is intended or not. That's why I asked. It is not relevant to NumPyro implementation.

fritzo: well if you do find a bug in `.log_abs_det_jacobian()` then https://github.com/pytorch/pytorch/issues/new seems like a good place for extended discussion

fehiepsi: Hmm, I didn't expect to find a bug. I thought there would be a simple interpretation here. Let me think about it more. Creating a PyTorch issue is not necessary because I couldn't find a real transform that has the pattern `Transform(some_params).logdet(x,y) = y/x`. Let's discuss in some lunch instead. :D 
@martinjankowiak How about we defer this discussion to the future? I don't actually understand much about those different interpretations.

fehiepsi:
This is blocked until we decide if we should remove all transforms' cached values. This is achieved by setting `Transform.cache_size=0` in Pyro or clearing `site["intermediates"]` in NumPyro.

martinjankowiak (approved):

martinjankowiak:
@fehiepsi lgtm you'll need to resolve merge conflicts though

martinjankowiak (approved):

PR #1051, opened 2021-05-31
Add tests for pathwise derivative

### LINKED ISSUE #944: Add more tests for sample gradient of Beta/StudentT/Dirichlet
Quoted from @fritzo 

> IIUC the reparametrized gradients for Gamma, Beta, Dirichlet, Chi2, and StudentT are all stochastic in the sense that gradients are correct only in expectation. You might instead add a weaker test that drew many samples and estimated the Jacobian `d/dtheta E_p(z|theta)[ f(z) ]` for the parameter theta and some collection of functions `f` like `[1, z, z**2, z**3, z**4]`. See e.g. Figure 2 in https://arxiv.org/pdf/1806.01851.pdf

fehiepsi:
Resolves #944

@fritzo I'm not sure if the tests I added are non-trivial. Could you take a look?

### DIFF
--- a/test/test_distributions.py
+++ b/test/test_distributions.py
@@ -665,48 +665,39 @@ def test_sample_gradient(jax_dist, sp_dist, params):
 
 
 @pytest.mark.parametrize(
-    "jax_dist, sp_dist, params",
+    "jax_dist, params",
     [
-        (dist.Gamma, osp.gamma, (1.0,)),
-        (dist.Gamma, osp.gamma, (0.1,)),
-        (dist.Gamma, osp.gamma, (10.0,)),
-        (dist.Chi2, osp.chi2, (1.0,)),
-        (dist.Chi2, osp.chi2, (0.1,)),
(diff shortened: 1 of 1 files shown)


martinjankowiak (approved):
yup gradient of a MC expectation versus gradient of an analytic expectation should suffice

fritzo (approved):
LGTM

fritzo:
I suppose another test that avoided relying on `.mean` and `.variance` (e.g. unavailable in Cauchy) might use importance weighting, something like this:
```py
def probe(sample):
    # ... an arbitrary scalar function of sample ...

samples = d.sample(*params)
actual_probe = np.mean(probe(samples), axis=0)

log_weights = d.log_prob(block(samples))
dice_weights = np.exp(log_weights - block(log_weights))
expected_probe = np.dot(dice_weights, probe(block(samples)))

assert_grads_equal(actual_probe, expected_probe, params)
```

Issue #1033, opened 2021-05-08
Add LDA example to illustrate some usage of `flax_module`/`haiku_module`


fehiepsi:
After #1016 and #989, it would be nice to port [ProdLDA](http://pyro.ai/examples/prodlda.html#ProdLDA:-Latent-Dirichlet-Allocation-with-Product-of-Experts) example in Pyro to illustrate some usage of JAX nn frameworks `flax`/`haiku` in NumPyro. We might consider replacing `SoftmaxNormal` there with `Dirichlet` distribution. This is a good first issue for those who are interested in both NLP and Bayesian inference.

tcbegley:
I'm interested in this and happy to start working on it.

tcbegley:
Hi @fehiepsi, I have a question about the Pyro version of this notebook.

Both encoder and decoder have dropout layers, is there a way to know inside the model / guide if they are being run in a training context by e.g. `SVI` versus being used in a generative / predictive context? In the latter case I think you would want to ensure that dropout is turned off right?

From what I can tell in the Pyro / ProdLDA example it doesn't really matter, because the encoder / decoder only ever do a forward pass during training / inference and then the weight vector of the decoder are directly extracted once training is complete. Is that right?

fehiepsi:
You are right, only the weight vector is important to extract topics' word distributions. To let `model/guide` know about training context, you can add an argument `training=True/False` to them:
```python
def Decoder(x, training=...):
    dropout_rate = ... if training else 0.
    ...

def model(..., training=...):
    decoder = haiku_module("encoder", hk.transform(Decoder), apply_rng=True, ...)
    out = decoder(numpyro.prng_key(), x, False)
```

fehiepsi:
Thanks, @tcbegley!

PR #1056, opened 2021-06-05
ProdLDA example with Haiku + Flax

tcbegley:
First version of NumPyro port of [ProdLDA notebook from Pyro docs](http://pyro.ai/examples/prodlda.html) as requested in #1033. It's branched off #1016.

Having some issues with unexpected tracers inside SVI when differentiating the loss function. I think it's to do with the batchnorm mutable state, but it's a bit beyond my understanding atm. Any pointers would be appreciated.


notebooks/source/prodlda.ipynb
```diff
@@ -0,0 +1,794 @@
...
+   },
+   "outputs": [],
+   "source": [
+    "import os\n",
+    "\n",
+    "assert numpyro.__version__.startswith(\"0.6.0\")\n",
+    "# Enable smoke test - run the notebook cells on CI.\n",
+    "smoke_test = \"CI\" in os.environ"
+   ]
+  },
```
tcbegley, line 99: Adding

```
{
    "nbsphinx": "hidden",
    "tags": []
}
```

to the cell metadata allows you to suppress cells from the rendered output. I think this could be a nice pattern for removing smoke test configuration from the docs, but still allowing notebooks to be run in CI. What do you think?

fehiepsi (commented):
Hi @tcbegley, I just fixed a bug in #1016 that caused the issue. Sorry for that!

I think it is better to make an example for prodlda that links to the original tutorial for a few reasons:
+ only code is changed I guess - the content seems to be similar
+ the main attribute should be given to the Pyro tutorial
+ we can add [comments like this](https://github.com/pyro-ppl/numpyro/blob/master/examples/capture_recapture.py#L57); sphinx-gallery will render the text well if some explanation for haiku/flax is needed. You can also add the result image, like in [this example](https://github.com/pyro-ppl/numpyro/blob/4ad6b98ae3127e0d509ca7ff425cbf07185fbdad/examples/funnel.py#L25)
+ for comparison - it might be better to look at an end-to-end code in numpyro, without the need to follow a similar training pattern in Pyro
+ it is easier to add flax code :D

 About the implementation, most look great to me. I just have a few comments about the API:
+ making `HaikuEncoder` seems to be unnecessary - I guess you can just default
```python
def HaikuEncoder(x, is_training, num_topics, hidden, dropout_rate):
    # follow the same implementation of HaikuEncoder.__call__

def model(docs, hypers, is_training=False):
    # hyperparameters can be accessed through hypers["vocal_size"],...
    encoder = haiku_module(
        "encoder",
        hk.transform_with_state(partial(HaikuEncoder, num_topics=...)),
        ...
```

The data is not large so you can use [svi.run](https://num.pyro.ai/en/latest/svi.html#numpyro.infer.svi.SVI.run) to train, just make sure that you use `numpyro.subsample` correctly
```python
def model(docs, ...):
    with numpyro.plate("documents", docs.shape[0], subsample_size=hyper["batch_size"]):
        logtheta = numpyro.sample(
            "logtheta", dist.Normal(0, 1).expand([num_topics]).to_event(1)
        )
        theta = jax.nn.softmax(logtheta)
        ...
        batch_docs = numpyro.subsample(docs, event_dim=1)
        numpyro.sample("obs", ..., obs=batch_docs)
```

tcbegley:
Thanks for the review!

Been a bit busy lately so progress is a little slow, but have addressed your comments:
- Drop notebook in favour of script
- Not subclassing `hk.Module` in Encoder / Decoder implementations
- Ditching extra functional wrappers such as `build_model_and_guide`.
- Use `svi.run` with `numpyro.subsample`.

The script runs without error, but in my testing the model doesn't actually learn anything useful. Will need to do some experimentation to figure out why exactly. If you see anything obviously wrong let me know!

TODO: 
- [ ] Resolve training issues
- [ ] Allow variables to be overridden from the commandline (i.e. use `argparse` for setting hyperparameters)
- [ ] Add Flax
- [ ] Investigate using Dirichlet distribution instead of softmax normal as suggested in #1033 (I figured I'd try and match outputs of the Pyro implementation before trying this)

fehiepsi:
> Resolve training issues

Thanks, @tcbegley! That's a great plan and the script is pretty clean!! I'll look at this issue today.

examples/prodlda.py
```diff
@@ -0,0 +1,207 @@
...
+
+class HaikuDecoder:
+    def __init__(self, vocab_size, dropout_rate):
+        self._dropout_rate = dropout_rate
+        self._vocab_size = vocab_size
+
+    def __call__(self, inputs, is_training):
+        dropout_rate = self._dropout_rate if is_training else 0.0
+        h = hk.dropout(hk.next_rng_key(), dropout_rate, inputs)
+        h = hk.Linear(self._vocab_size)(h)
```
fehiepsi, line 76: I think we need `with_bias=False` here.

examples/prodlda.py
```diff
@@ -0,0 +1,207 @@
...
+        h = hk.BatchNorm(create_scale=True, create_offset=True, decay_rate=0.1)(
```
fehiepsi, line 77: You can set `create_scale=False` and `create_offset=False`

examples/prodlda.py
```diff
@@ -0,0 +1,207 @@
...
+    )
+
+    with numpyro.plate(
+        "documents", docs.shape[0], subsample_size=hyperparams["batch_size"]
+    ):
+        batch_docs = numpyro.subsample(docs, event_dim=1)
+        logtheta_loc = jnp.zeros((batch_docs.shape[0], hyperparams["num_topics"]))
+        logtheta_scale = jnp.ones((batch_docs.shape[0], hyperparams["num_topics"]))
+        logtheta = numpyro.sample(
+            "logtheta", dist.Normal(logtheta_loc, logtheta_scale).to_event(1)
```
fehiepsi, line 104: You can use `dist.Normal(0, 1).expand([hyperparams["num_topics"]]).to_event(1)` for simplicity.

examples/prodlda.py
```diff
@@ -0,0 +1,207 @@
...
+        )
+        theta = jax.nn.softmax(logtheta)
+
+        count_param = decoder(numpyro.prng_key(), theta, is_training)
+
+        total_count = batch_docs.sum(-1).max().astype(int)
```
fehiepsi, line 110: This should be `total_count = batch_docs.sum(-1)`

examples/prodlda.py
```diff
@@ -0,0 +1,207 @@
...
+    print(f"Dictionary size: {len(vocab)}")
+    print(f"Corpus size: {docs.shape}")
+
+    # TODO: allow these to be overridden with command-line arguments
+    rng_key = random.PRNGKey(0)
+    num_topics = 20
+    docs = device_put(docs)
+    batch_size = 32
+    learning_rate = 1e-3
+    num_steps = 200_000
```
fehiepsi, line 181: I think 30,000 is enough (similar to the number in Pyro).

fehiepsi (commented):
@tcbegley Could you double-check if the following fixes solve the issue? I am seeing similar losses w.r.t. Pyro tutorial (note that in Pyro, the reported running loss is scaled down by `batch_size`).

examples/prodlda.py
```diff
@@ -0,0 +1,207 @@
...
+        num_topics=num_topics,
+        hidden=hidden,
+        dropout_rate=0.2,
+        batch_size=batch_size,
+    )
+
+    optimizer = numpyro.optim.Adam(learning_rate)
+    svi = SVI(model, guide, optimizer, loss=TraceMeanField_ELBO())
+
+    params, losses = svi.run(rng_key, num_steps, docs, hyperparams, is_training=True)
```
fehiepsi, line 195: new api: `svi_result = svi.run(...)`

examples/prodlda.py
```diff
@@ -0,0 +1,207 @@
...
+
+    plt.plot(losses)
```
fehiepsi, line 197: this is `svi_result.losses`

examples/prodlda.py
```diff
@@ -0,0 +1,207 @@
...
+
+    beta = params["decoder$params"]["haiku_decoder/linear"]["w"]
```
fehiepsi, line 199: this is `beta = svi_result.params["decoder$params"]["linear"]["w"]

examples/prodlda.py
```diff
@@ -0,0 +1,207 @@
...
+        batch_docs = numpyro.subsample(docs, event_dim=1)
+        logtheta_loc, logtheta_scale = encoder(
+            numpyro.prng_key(), batch_docs, is_training
+        )
+        numpyro.sample(
+            "logtheta", dist.Normal(logtheta_loc, logtheta_scale).to_event(1)
+        )
+
+
+def plot_word_cloud(b, ax, n):
```
fehiepsi, line 146: the arguments are `plot_word_cloud(b, ax, vocab, n)`

fehiepsi:
I can confirm that with the above change, we get the desired result.

examples/prodlda.py
tcbegley: Thanks! I suspected I had got something wrong with my batchnorms >_<

examples/prodlda.py
tcbegley: Very slick 🙂

examples/prodlda.py
tcbegley: I think I bumped this up to see what would happen when it wasn't learning anything in my initial testing and forgot I had done that 😳

examples/prodlda.py
tcbegley: Oops, good catch!

examples/prodlda.py
```diff
@@ -0,0 +1,332 @@
...
+"""
+Example: ProdLDA
+================
+In this example, we will follow [1] to implement the ProdLDA topic model from
+Autoencoding Variational Inference For Topic Models by Akash Srivastava and Charles
+Sutton [2]. This model returns consistently better topics than vanilla LDA and trains
+much more quickly. Furthermore, it does not require a custom inference algorithm that
+relies on complex mathematical derivations. This example also serves as an
+introduction to Flax and Haiku modules in NumPyro.
+
```
fehiepsi, line 13: Could you add a paragraph "For the interested readers, a nice extension of this model is the CombinedTM model (reference [3]) which utilizes a pre-trained sentence transformer (like https://www.sbert.net/) to generate a better representation of the encoded latent vector." ?

fehiepsi (commented):
Awesome work, @tcbegley! The implementation is so clean (I can quickly learn the difference between flax and haiku :D). This is a great example even for users who are not familiar with those nn frameworks. +1

examples/prodlda.py
```diff
@@ -0,0 +1,332 @@
...
+    def __call__(self, inputs, is_training):
+        h = nn.Dropout(self.dropout_rate, deterministic=not is_training)(inputs)
+        h = nn.Dense(self.vocab_size, use_bias=False)(h)
+        h = nn.BatchNorm(
+            use_bias=False, use_scale=False, use_running_average=not is_training
+        )(h)
+        return nn.softmax(h)
+
+
+def model(docs, hyperparams, is_training=False, use_haiku=False):
```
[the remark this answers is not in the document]
tcbegley: I went with `nn_framework` rather than `nn_module` because I thought the latter might get confused with `nn.Module`? Probably doesn't matter but let me know if you would prefer `nn_module`.

examples/prodlda.py
```diff
@@ -0,0 +1,332 @@
...
+        hyperparams,
+        is_training=True,
+        progress_bar=not args.disable_progbar,
+        use_haiku=args.use_haiku,
+    )
+
+
+def plot_word_cloud(b, ax, vocab, n):
+    indices = jnp.argsort(b)[::-1]
+    df = pd.DataFrame(indices[:100], columns=["index"])
```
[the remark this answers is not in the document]
tcbegley: I tried 20 and it looked a little sparse so I went with 30. What do you think? (See results [here](https://github.com/pyro-ppl/numpyro/blob/45716595bb856879a51d181527e51bebc1dade08/docs/source/_static/img/examples/prodlda.png))

examples/prodlda.py
```diff
@@ -0,0 +1,332 @@
...
+    fig.savefig("wordclouds.png")
+
+
+if __name__ == "__main__":
+    assert numpyro.__version__.startswith("0.6.0")
+    parser = argparse.ArgumentParser(
+        description="Probabilistic topic modelling with Flax and Haiku"
+    )
+    parser.add_argument("-n", "--num-steps", nargs="?", default=30_000, type=int)
+    parser.add_argument("-t", "--num-topics", nargs="?", default=20, type=int)
```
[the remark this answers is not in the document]
tcbegley: Went with 12 as the default but adjusted the plotting code so it can handle any number 💪

examples/prodlda.py
```diff
@@ -0,0 +1,360 @@
...
+
+        h = jax.nn.softplus(hk.Linear(self._hidden)(inputs))
+        h = jax.nn.softplus(hk.Linear(self._hidden)(h))
+        h = hk.dropout(hk.next_rng_key(), dropout_rate, h)
+
+        # NB: here we set `create_scale=False` and `create_offset=False` to reduce
+        # the number of learning parameters
+        h_mu = hk.Linear(self._num_topics)(h)
+        logtheta_loc = hk.BatchNorm(
+            create_scale=False, create_offset=False, decay_rate=0.99
```
tcbegley, line 67: I think this is the same as the default in flax (though their parameter is called `momentum` rather than `decay_rate`).

Haiku requires that you set this, should I set it in the flax layers too to be more explicit? Also is `0.99` a good choice for this problem? I don't have that much intuition for batch norm...

tcbegley:
Thanks for reviewing again @fehiepsi!

I think I've addressed all your comments. Still have the Dirichlet distribution to try out, but I think then I'll be more or less done.

examples/prodlda.py
fehiepsi: Good point! Luckily we didn't perform any prediction in this example, so the running mean/scale is not used anywhere. 

> intuition for batch norm

Currently, `momentum=0.1` is used by default in PyTorch [BatchNorm1d](https://pytorch.org/docs/stable/generated/torch.nn.BatchNorm1d.html#torch.nn.BatchNorm1d) (also in [ctm model](https://github.com/MilaNLProc/contextualized-topic-models/blob/master/contextualized_topic_models/networks/decoding_network.py#L82)) to put less weights on the new input value when updating. The `momentum` notion in PyTorch is equivalent to `1 - momentum` notion in Flax/Haiku, so I think it is fine to use `0.9` here. I guess `0.99` also works but changing the default value in flax will make that hyperparameter more explicit to reader :D

examples/prodlda.py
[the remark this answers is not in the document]
fehiepsi: Yeah, agree. I like `nn_framework`. :D

examples/prodlda.py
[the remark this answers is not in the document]
fehiepsi: That looks great, thanks for trying out! (I just thought that 100 is a bit noisy)

fehiepsi (commented):
Yeah, I think we can use Dirichlet distribution directly in future iterations. Could you add this example to the applications section of [this index file](https://github.com/pyro-ppl/numpyro/blob/master/docs/source/index.rst)? Could you also add tests for both `flax` and `haiku` [here](https://github.com/pyro-ppl/numpyro/blob/master/test/test_examples.py#L39) (with small `num_steps` and `hidden`)? Then I think this is ready. Fantastic work, @tcbegley!!!

tcbegley:
> I think we can use Dirichlet distribution directly in future iterations.

To clarify, do you mean I should not try that on this PR and leave that for a future update?

setup.py
```diff
@@ -65,7 +65,14 @@
             # TODO: relax this restriction when we revise tfp wrapper
             "tfp-nightly<=0.14.0.dev20210608",
         ],
-        "examples": ["arviz", "jupyter", "matplotlib", "pandas", "seaborn"],
+        "examples": [
+            "arviz",
```
fehiepsi, line 69: I think you'll also need `scikit-learn` to get data.

test/test_examples.py
```diff
@@ -37,6 +37,7 @@
     "minipyro.py",
     "neutra.py --num-samples 100 --num-warmup 100",
     "ode.py --num-samples 100 --num-warmup 100 --num-chains 1",
+    "prodlda.py --num-steps 10 --hidden 10",
```
fehiepsi, line 40: Could you test for both flax and haiku?
```
"prodlda.py --num-steps 10 --hidden 10 --nn_framework='flax'",
"prodlda.py --num-steps 10 --hidden 10 --nn_framework='haiku'",
```

fehiepsi:
> should not try that on this PR and leave that for a future update

Sorry, I thought you wanted to defer it to a future PR. It would be nice to try in this PR. To make the training process more stable, could you register KL divergence for [Dirichlet-Dirichlet](https://github.com/pytorch/pytorch/blob/05c1e5b65571d2fa14b0f06c37ba970fd7cc43d0/torch/distributions/kl.py#L232) pair [here](https://github.com/pyro-ppl/numpyro/blob/f8f482ad7b2f3f658d198c99f76fdda12cff84ca/numpyro/distributions/kl.py#L198)?

tcbegley:
> It would be nice to try in this PR.

Oh great, I misunderstood. I'll take a more careful look into it then 🙂

examples/prodlda.py
```diff
@@ -0,0 +1,369 @@
...
+    def __call__(self, inputs, is_training):
+        h = nn.Dropout(self.dropout_rate, deterministic=not is_training)(inputs)
+        h = nn.Dense(self.vocab_size, use_bias=False)(h)
+        h = nn.BatchNorm(
+            use_bias=False,
+            use_scale=False,
+            momentum=0.9,
+            use_running_average=not is_training,
+        )(h)
+        return nn.softmax(h)
```
fehiepsi, line 137: It is more stable to return `h` here and use `Multinomial(..., logits=...)`.

examples/prodlda.py
```diff
@@ -0,0 +1,369 @@
...
+        logtheta = numpyro.sample(
+            "logtheta",
+            dist.Normal(0, 1).expand((hyperparams["num_topics"],)).to_event(1),
+        )
+        theta = jax.nn.softmax(logtheta)
+
+        if nn_framework == "flax":
+            count_param = decoder(
+                theta, is_training, rngs={"dropout": numpyro.prng_key()}
+            )
```
fehiepsi, line 183: It is more stable to define `logits = decoder(...)`

examples/prodlda.py
```diff
@@ -0,0 +1,369 @@
...
+        elif nn_framework == "haiku":
+            count_param = decoder(numpyro.prng_key(), theta, is_training)
+
+        total_count = batch_docs.sum(-1)
+        numpyro.sample(
+            "obs", dist.Multinomial(total_count, count_param), obs=batch_docs
```
fehiepsi, line 189: use `Multinomial(total_count, logits=logits)`

examples/prodlda.py
```diff
@@ -0,0 +1,369 @@
...
+
+    with numpyro.plate(
+        "documents", docs.shape[0], subsample_size=hyperparams["batch_size"]
+    ):
+        batch_docs = numpyro.subsample(docs, event_dim=1)
+        logtheta = numpyro.sample(
+            "logtheta",
+            dist.Normal(0, 1).expand((hyperparams["num_topics"],)).to_event(1),
+        )
+        theta = jax.nn.softmax(logtheta)
```
fehiepsi, line 178: This can be changed into
```diff
-        logtheta = numpyro.sample(
-            "logtheta",
-            dist.Normal(0, 1).expand((hyperparams["num_topics"],)).to_event(1),
+        theta = numpyro.sample(
+            "theta",
+            dist.Dirichlet(jnp.ones(hyperparams["num_topics"]))
,
         )
-        theta = jax.nn.softmax(logtheta)
```

examples/prodlda.py
```diff
@@ -0,0 +1,369 @@
...
+            logtheta_loc, logtheta_scale = encoder(
+                batch_docs, is_training, rngs={"dropout": numpyro.prng_key()}
+            )
+        elif nn_framework == "haiku":
+            logtheta_loc, logtheta_scale = encoder(
+                numpyro.prng_key(), batch_docs, is_training
+            )
+
+        numpyro.sample(
+            "logtheta", dist.Normal(logtheta_loc, logtheta_scale).to_event(1)
```
fehiepsi, line 248: My dirichlet version
```diff
         if nn_framework == "flax":
-            logtheta_loc, logtheta_scale = encoder(
+            concentration = encoder(
                 batch_docs, is_training, rngs={"dropout": numpyro.prng_key()}
         ...
-            "logtheta", dist.Normal(logtheta_loc, logtheta_scale).to_event(1)
+            "theta", dist.Dirichlet(concentration)
```

fehiepsi (commented):
Here are my changes to use Dirichlet distribution directly. It seems to work but you might try a higher `num_steps` or other hyperparameters.

examples/prodlda.py
```diff
@@ -0,0 +1,369 @@
...
+            use_running_average=not is_training,
+        )(h_mu)
+        logtheta_logvar = nn.BatchNorm(
+            use_bias=False,
+            use_scale=False,
+            momentum=0.9,
+            use_running_average=not is_training,
+        )(h_sigma)
+        logtheta_scale = jnp.exp(0.5 * logtheta_logvar)
+        return logtheta_loc, logtheta_scale
```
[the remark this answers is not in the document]
tcbegley: Thanks! I think when I tried briefly the other day I forgot to exponentiate the log concentration 🤦‍♂️

tcbegley:
Thanks for the pointers @fehiepsi! Registering the KL like that is really neat.

Tested a few times with different hyperparameters. I don't think the `num_steps` needs increasing, if anything I think you could get away with decreasing it! I did increase `hidden` as the topics seemed a bit better defined when doing so. When using GPU for inference this had no noticeable impact on runtime.

I'm not sure if the examples tests are going to pass, I ran them locally in a fresh virtual environment, running only the commands that get run in CI and it was fine, I also [ran the prodlda tests on my fork](https://github.com/tcbegley/numpyro/pull/1/files), with the other examples tests disabled so I could read the logs more easily, and in both cases they passed, but they've failed here a couple times in a row. If they fail again I'll take a more careful look into why.

examples/prodlda.py
```diff
@@ -0,0 +1,346 @@
...
+    num_topics: int
+    hidden: int
+    dropout_rate: float
+
+    @nn.compact
+    def __call__(self, inputs, is_training):
+        h = nn.softplus(nn.Dense(self.hidden)(inputs))
+        h = nn.softplus(nn.Dense(self.hidden)(h))
+        h = nn.Dropout(self.dropout_rate, deterministic=not is_training)(h)
+
```
fehiepsi, line 97: nit: no empty space

examples/prodlda.py
```diff
@@ -0,0 +1,346 @@
...
+    def __call__(self, inputs, is_training):
+        dropout_rate = self._dropout_rate if is_training else 0.0
+
+        h = jax.nn.softplus(hk.Linear(self._hidden)(inputs))
+        h = jax.nn.softplus(hk.Linear(self._hidden)(h))
+        h = hk.dropout(hk.next_rng_key(), dropout_rate, h)
+
+        # NB: here we set `create_scale=False` and `create_offset=False` to reduce
+        # the number of learning parameters
+        h = hk.Linear(self._num_topics)(h)
```
fehiepsi, line 65: nit: move this to the line 61.

examples/prodlda.py
```diff
@@ -0,0 +1,346 @@
...
+"""
+Example: ProdLDA
+================
+In this example, we will follow [1] to implement the ProdLDA topic model from
+Autoencoding Variational Inference For Topic Models by Akash Srivastava and Charles
+Sutton [2]. This model returns consistently better topics than vanilla LDA and trains
+much more quickly. Furthermore, it does not require a custom inference algorithm that
+relies on complex mathematical derivations. This example also serves as an
+introduction to Flax and Haiku modules in NumPyro.
+
```
fehiepsi, line 13: Could you mention the change for using Dirichlet directly, instead of the Softmax-Normal approximation here?

fehiepsi (approved):
LGTM! Could you help me add a test for KL [here](https://github.com/pyro-ppl/numpyro/blob/f8f482ad7b2f3f658d198c99f76fdda12cff84ca/test/test_distributions.py#L2035)? The code should be similar to NormalNormal pair but you might need a higher number of samples than `10000`.

test/test_distributions.py
```diff
@@ -2033,3 +2033,13 @@ def test_kl_normal_normal(shape):
     x = p.sample(random.PRNGKey(0), (10000,)).copy()
     expected = jnp.mean((p.log_prob(x) - q.log_prob(x)), 0)
     assert_allclose(actual, expected, rtol=0.05)
+
+
+@pytest.mark.parametrize("shape", [(), (4,), (2, 3)], ids=str)
+def test_kl_dirichlet_dirichlet(shape):
+    p = dist.Normal(np.exp(np.random.normal(size=shape)))
+    q = dist.Normal(np.exp(np.random.normal(size=shape)))
```
fehiepsi, line 2041: Normal -> Dirichlet

test/test_distributions.py
```diff
@@ -2033,3 +2033,13 @@ def test_kl_normal_normal(shape):
     x = p.sample(random.PRNGKey(0), (10000,)).copy()
     expected = jnp.mean((p.log_prob(x) - q.log_prob(x)), 0)
     assert_allclose(actual, expected, rtol=0.05)
+
+
+@pytest.mark.parametrize("shape", [(), (4,), (2, 3)], ids=str)
```
fehiepsi, line 2038: For dirichlet, you should remove `()` because Dirichlet is a multivariate distribution.

fehiepsi (approved):

PR #1058, opened 2021-06-08
Add warning when initializing a substituted model

### LINKED ISSUE #1057: raise a better error/warning for init strategies with substituted models
Right now, init strategy with a ```substituted``` model breaks, and the error message showed is not straight forward. It could be clarified that, in order to work, ```substitute``` requires that the substituted site is ```blocked```, and that anyway ```condition``` is preferred to ```substitute```  for modelling purposes.

fehiepsi:
Resolves #1057. For a substituted site, `rng_key` is not available (because `seed` handler knows that this site already has some value) so we can't generate a new init value. This PR adds a warning and in that case, the init strategy will return the substituted value.

I also fixed some doc formatting issues and added a note to `substitute` docstring regarding conditioning.

cc @llaurabat91


numpyro/infer/initialization.py
```diff
@@ -58,9 +65,14 @@ def init_to_uniform(site=None, radius=2):
...
     ):
+        if site["value"] is not None:
+            warnings.warn(
+                f"Site '{site['name']}' already stores a value - skip initializing this site."
+            )
+            return site["value"]
+
         rng_key = site["kwargs"].get("rng_key")
         sample_shape = site["kwargs"].get("sample_shape")
-        rng_key, subkey = random.split(rng_key)
```
fehiepsi, line 63: the subkey is not used so we don't need this line

numpyro/handlers.py
```diff
@@ -604,16 +604,16 @@ class scope(Messenger):
...
+       >>> import numpyro
+       >>> import numpyro.distributions as dist
+       >>> from numpyro.handlers import scope, seed, trace
+       >>>
+       >>> def model():
+       ...     with scope(prefix="a"):
+       ...         with scope(prefix="b", divider="."):
+       ...             return numpyro.sample("x", dist.Bernoulli(0.5))
+       ...
+       >>> assert "a/b.x" in trace(seed(model, 0)).get_trace()
```
fehiepsi, line 616: Using 3 spaces following [doctest docs](https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html#confval-doctest_test_doctest_blocks) and to be consistent w.r.t. to other docstrings in this file. 3 or 4 spaces do not matter I believe.

numpyro/infer/initialization.py
```diff
@@ -25,6 +25,12 @@ def init_to_median(site=None, num_samples=15):
         and not site["is_observed"]
         and not site["fn"].is_discrete
     ):
+        if site["value"] is not None:
+            warnings.warn(
+                f"Site '{site['name']}' already stores a value - skip initializing this site."
```
fritzo, line 30: nit: I think it would be clearer to warn

> init_to_median() skipping initialization of site {} which already stores a value.

since 'skip' could be interpreted as a command (hey you should skip this) rather than a statement (hey I'm skipping this).

fritzo:
Seems like a good idea; we should do this in Pyro too!

numpyro/infer/initialization.py
fehiepsi: Thanks, I have made the change.

fehiepsi:
@fritzo Could you help me review this pr?

fritzo (approved):

fehiepsi:
Thanks, Fritz!

PR #1059, opened 2021-06-08
fix loc scale reparam with center=1

fehiepsi:
Caught and fixed by @fritzo in https://github.com/pyro-ppl/pyro/pull/2863.


fritzo (approved):
Thanks for porting this!

PR #1055, opened 2021-06-03
Sine Skewed distribution

OlaRonning:
Translating Sine Skewed distribution from [Pyro](https://github.com/pyro-ppl/pyro/pull/2826).

PR #1063, opened 2021-06-10
Porting Sine Bivaraite von Mises from Pyro

OlaRonning:
Translating Sine Bivariate von Mises Distribution from Pyro; phi marginal is sampled similarly to von Mises distribution to accommodate for immutable arrays.


test/test_distributions.py
```diff
@@ -333,6 +341,31 @@ def get_sp_dist(jax_dist):
     T(dist.VonMises, 2.0, 10.0),
     T(dist.VonMises, 2.0, jnp.array([150.0, 10.0])),
     T(dist.VonMises, jnp.array([1 / 3 * jnp.pi, -1.0]), jnp.array([20.0, 30.0])),
+    T(
+        SineBivariateVonMises,
+        jnp.array([0.0]),
+        jnp.array([0.0]),
+        jnp.array([5.0]),
+        jnp.array([6.0]),
+        jnp.array([2.0]),
```
fehiepsi, line 350: Is it required that those are at least 1D?

numpyro/distributions/directional.py
```diff
@@ -75,6 +132,246 @@ def variance(self):
...
+            psi_loc.shape,
+            phi_concentration.shape,
+            psi_concentration.shape,
+            correlation.shape,
+        )
+        super().__init__(batch_shape, (2,), validate_args)
+
+    @lazy_property
+    def norm_const(self):
+        corr = self.correlation.reshape(1, -1) + 1e-8
```
fehiepsi, line 230: use `jnp.reshape` for scalars

numpyro/distributions/directional.py
```diff
@@ -75,6 +132,246 @@ def variance(self):
...
+            self.phi_loc,
+            self.psi_loc,
+            self.phi_concentration,
+            self.psi_concentration,
+            self.correlation,
+        ) = promote_shapes(
+            phi_loc, psi_loc, phi_concentration, psi_concentration, correlation
+        )
+        batch_shape = lax.broadcast_shapes(
+            phi_loc.shape,
```
fehiepsi, line 220: use `jnp.shape(...)` for scalars

numpyro/distributions/directional.py
```diff
@@ -75,6 +132,246 @@ def variance(self):
...
+            * jnp.sin(value[..., 1] - self.psi_loc)
+        )
+        return indv + corr - self.norm_const
+
+    def sample(self, key, sample_shape=()):
+        """
+        ** References: **
+            1. A New Unified Approach for the Simulation of aWide Class of Directional Distributions
+               John T. Kent, Asaad M. Ganeiber & Kanti V. Mardia (2018)
+        """
```
fehiepsi, line 267: assert key is prng_key

numpyro/distributions/directional.py
```diff
@@ -75,6 +132,246 @@ def variance(self):
...
+        b0 = self._bfind(eig)
+
+        total = _numel(sample_shape)
+        phi_den = log_I1(0, conc[1]).squeeze(0)
+        phi_shape = (total, 2, _numel(self.batch_shape))
+        phi_state = SineBivariateVonMises._phi_marginal(
+            phi_shape, phi_key, conc, corr, eig, b0, eigmin, phi_den
+        )
+
+        phi = lax.atan2(phi_state.phi[:, 1:], phi_state.phi[:, :1])
```
fehiepsi, line 286: `lax.atan2` -> `jnp.arctan2`

numpyro/distributions/directional.py
```diff
@@ -75,6 +132,246 @@ def variance(self):
...
+
+        alpha = jnp.sqrt(conc[1] ** 2 + (corr * jnp.sin(phi)) ** 2)
+        beta = lax.atan(corr / conc[1] * jnp.sin(phi))
```
fehiepsi, line 289: ditto: use `jnp.arctan2`

numpyro/distributions/directional.py
```diff
@@ -75,6 +132,246 @@ def variance(self):
...
+
+        fs = (
+            lbinoms.reshape(-1, 1)
+            + 2 * m * jnp.log(corr)
+            - m * jnp.log(4 * jnp.prod(conc, axis=-1))
+        )
+        fs += log_I1(49, conc, terms=51).sum(-1)
+        mfs = fs.max()
+        norm_const = 2 * jnp.log(jnp.array(2 * pi)) + mfs + logsumexp(fs - mfs, 0)
+        return norm_const.reshape(self.phi_loc.shape)
```
fehiepsi, line 247: ditto: use `jnp.shape`

docs/source/distributions.rst
```diff
@@ -477,6 +477,13 @@ ZeroInflatedNegativeBinomial2
 
 Directional Distributions
 =========================
+SineBivariateVonMises
```
fehiepsi, line 480: nit: add an empty line above. This should be after `ProjectedNormal`

numpyro/distributions/directional.py
```diff
@@ -75,6 +132,246 @@ def variance(self):
...
+    """
+
+    arg_constraints = {
+        "phi_loc": constraints.real,
+        "psi_loc": constraints.real,
+        "phi_concentration": constraints.positive,
+        "psi_concentration": constraints.positive,
+        "correlation": constraints.real,
+    }
+    support = constraints.independent(constraints.real, 1)
```
fehiepsi, line 188: I'm not sure if we need to be consistent with VonMises here. Could you add # TODO at those supports to revisit the issue?

fehiepsi (commented):
Thanks, @OlaRonning! I haven't checked the math because this is your port from Pyro. I just have some comments on shape sematics.

numpyro/distributions/directional.py
OlaRonning: I see, I'll update this.

numpyro/distributions/directional.py
OlaRonning: same

numpyro/distributions/directional.py
OlaRonning: That makes sense, thanks.

numpyro/distributions/directional.py
OlaRonning: I believe we do, I'll add a TODO and revisit once #1080 is settled.

test/test_distributions.py
OlaRonning: Yes, however, with your suggested changes, this may be resolved. I'll check and update the test cases appropriately.

docs/source/distributions.rst
OlaRonning: Yes, you are right. I'll fix this.

test/test_distributions.py
fehiepsi: If this is required, you should change the `arg_constraints` in the distribution appropriately (using `independent`). But from the implementation, it seems to be that those parameters are scalars (modulo batch_shape).

numpyro/distributions/directional.py
```diff
@@ -75,6 +132,246 @@ def variance(self):
...
+
+    .. note:: The correlation and weighted_correlation params are mutually exclusive.
+
+    .. note:: In the context of :class:`~numpyro.infer.SVI`, this distribution can be used as a likelihood but not for
+        latent variables.
+
+    ** References: **
+        1. Probabilistic model for two dependent circular variables Singh, H., Hnizdo, V., and Demchuck, E. (2002)
+
+    :param jnp.Tensor phi_loc: location of first angle
```
fehiepsi, line 172: Is there jnp.Tensor?

numpyro/distributions/directional.py
```diff
@@ -75,6 +132,246 @@ def variance(self):
...
+    .. note:: Sample efficiency drops as
+
+        .. math::
+            \frac{\rho}{\kappa_1\kappa_2} \rightarrow 1
+
+        because the distribution becomes increasingly bimodal.
+
+    .. note:: The correlation and weighted_correlation params are mutually exclusive.
+
+    .. note:: In the context of :class:`~numpyro.infer.SVI`, this distribution can be used as a likelihood but not for
```
fehiepsi, line 166: Nit: use full path for SVI (missing svi. part)

numpyro/distributions/directional.py
```diff
@@ -75,6 +132,246 @@ def variance(self):
...
+    .. math::
+        C = (2\pi)^2 \sum_{i=0} {2i \choose i}
+        \left(\frac{\rho^2}{4\kappa_1\kappa_2}\right)^i I_i(\kappa_1)I_i(\kappa_2),
+
+    where :math:`I_i(\cdot)` is the modified bessel function of first kind, mu's are the locations of the distribution,
+    kappa's are the concentration and rho gives the correlation between angles :math:`x_1` and :math:`x_2`.
+    This distribution is helpful for modeling coupled angles such as torsion angles in peptide chains.
+
+    To infer parameters, use :class:`~numpyro.infer.NUTS` or :class:`~numpyro.infer.HMC` with priors that
+    avoid parameterizations where the distribution becomes bimodal; see note below.
```
fehiepsi, line 155: Use full path for NUTS and HMC

numpyro/distributions/directional.py
```diff
@@ -75,6 +132,246 @@ def variance(self):
...
+        super().__init__(batch_shape, (2,), validate_args)
+
+    @lazy_property
+    def norm_const(self):
+        corr = self.correlation.reshape(1, -1) + 1e-8
+        conc = jnp.stack(
+            (self.phi_concentration, self.psi_concentration), axis=-1
+        ).reshape(-1, 2)
+        m = jnp.arange(50).reshape(-1, 1)
+        num = lax.lgamma(2 * m + 1.0)
```
fehiepsi, line 235: Use special.gammaln

numpyro/distributions/directional.py
fehiepsi: Yes, sorry

numpyro/distributions/directional.py
```diff
@@ -75,6 +132,246 @@ def variance(self):
...
+        eigmin = jnp.broadcast_to(eigmin, shape)
+        phi_den = jnp.broadcast_to(phi_den, shape)
+
+        def update_fn(curr):
+            i, done, phi, key = curr
+            phi_key, key = random.split(key)
+            accept_key, acg_key, phi_key = random.split(phi_key, 3)
+
+            x = jnp.sqrt(1 + 2 * eig / b0) * random.normal(acg_key, shape)
+            x /= jnp.linalg.norm(x, axis=1)[
```
fehiepsi, line 317: I guess you wanted to use keepdims here?

numpyro/distributions/directional.py
```diff
@@ -75,6 +132,246 @@ def variance(self):
...
+            - m * jnp.log(4 * jnp.prod(conc, axis=-1))
+        )
+        fs += log_I1(49, conc, terms=51).sum(-1)
+        mfs = fs.max()
+        norm_const = 2 * jnp.log(jnp.array(2 * pi)) + mfs + logsumexp(fs - mfs, 0)
+        return norm_const.reshape(self.phi_loc.shape)
+
+    def log_prob(self, value):
+        if self._validate_args:
+            self._validate_sample(value)
```
fehiepsi, line 251: Use validate_sample decorator instead

numpyro/distributions/directional.py
```diff
@@ -75,6 +132,246 @@ def variance(self):
...
+        return PhiMarginalState(
+            phi_state.i, phi_state.done, phi_state.phi, phi_state.key
+        )
+
+    @property
+    def mean(self):
+        """Computes circular mean of distribution. NOTE: same as location when mapped to support [-pi, pi]"""
+        return (jnp.stack((self.phi_loc, self.psi_loc), axis=-1) + jnp.pi) % (
+            2.0 * jnp.pi
+        ) - jnp.pi
```
fehiepsi, line 366: I think you want to broadcast this to batch_shape

numpyro/distributions/directional.py
```diff
@@ -1,22 +1,79 @@
...
+    orders = orders + 1
+    if value.ndim == 0:
+        vshape = jnp.shape([1])
+    else:
+        vshape = value.shape
+    value = value.reshape(-1, 1)
+    flat_vshape = _numel(vshape)
+
+    k = jnp.arange(terms)
+    lgammas_all = lax.lgamma(jnp.arange(1.0, terms + orders + 1))
```
fehiepsi, line 51: Use special gammaln

numpyro/distributions/directional.py
```diff
@@ -1,22 +1,79 @@
...
+    assert lgammas_all.shape == (orders + terms,)  # lgamma(0) = inf => start from 1
+
+    lvalues = lax.log(value / 2) * k.reshape(1, -1)
```
fehiepsi, line 54: Use jnp.log. Basically, we will try to avoid lax api as much as possible.

fehiepsi (commented):
Hi @OlaRonning, I missed several points in the first pass. Here are some of the new ones. :)

OlaRonning:
Thanks for the review. I see I missed some left

numpyro/distributions/directional.py
OlaRonning: Ups; I thought I fixed this. It May have been on  #1055... I'll fix this to `jnp.array`

numpyro/distributions/directional.py
OlaRonning: Will this make linking work properly? I will fix this throughout and in #1055.

numpyro/distributions/directional.py
OlaRonning: Yes, thanks!

numpyro/distributions/directional.py
OlaRonning: Will do, I'll also fix it in #1055.

numpyro/distributions/directional.py
OlaRonning: I see; I'll remove `lax` throughout.

test/test_distributions.py
OlaRonning: It looks like the scalars work with your latest suggested changes.

fehiepsi: Could you add one test that invokes scalars?

OlaRonning: Sure, I’ll update once I back from holidays next week (forgot my laptop C:).

numpyro/distributions/directional.py
```diff
@@ -75,6 +133,259 @@ def variance(self):
...
+    arg_constraints = {
+        "phi_loc": constraints.real,
+        "psi_loc": constraints.real,
+        "phi_concentration": constraints.positive,
+        "psi_concentration": constraints.positive,
+        "correlation": constraints.real,
+    }
+    support = constraints.independent(
+        constraints.real, 1
+    )  # TODO: @OlaRonning update to circular constraint @1080
```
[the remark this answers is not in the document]
OlaRonning: Makes sense; I'll make the update :+1:

numpyro/distributions/directional.py
```diff
@@ -92,6 +150,256 @@ def variance(self):
...
+    :param np.ndarray phi_concentration: concentration of first angle
+    :param np.ndarray psi_concentration: concentration of second angle
+    :param np.ndarray correlation: correlation between the two angles
+    :param np.ndarray weighted_correlation: set correlation to weigthed_corr * sqrt(phi_conc*psi_conc)
+        to avoid bimodality (see note).
+    """
+
+    arg_constraints = {
+        "phi_loc": constraints.circular,
+        "psi_loc": constraints.circular,
```
fehiepsi, line 201: I feel that we can relax this constraint - even with real constraints, we still have valid distributions. What do you think?

fehiepsi (approved):

numpyro/distributions/directional.py
OlaRonning: I like the circular constraint because it is explicit; what is the advantage of using real constraint?

fehiepsi: It is just unnecessary imo (e.g. we can allow users to put in angles from pi to 2pi or even 10pi). But it is up to you.

PR #1064, opened 2021-06-11
Refactor TFP wrappers to avoid new metadata logic

### LINKED ISSUE #1062: Redesign TFP wrappers
The new changes in meta stuff of TFP distribution just break our CI. I have spent a day resolving the issue but couldn't make it work :(. To avoid similar issues in the future, I think it is better to make wrappers like in funsor.distributions, rather than creating Mixin class/metaclass. Making wrappers also help us resolving the issues like `Distribution.shape` is a method in pyro but it is a property in tfp.

Currently, this issue is blocking all PRs.

fehiepsi:
Resolves #1062.

In a recent nightly release, TFP distributions are a subclass of so-called `AutoCompositeTensorDistribution`, which has a new meta type. It is pretty complicated to resolve relevant issues to make CI pass, so I revise the TFP wrappers so that they are more agnostic to the upstream changes. This way, we can resolve some conflicts like `mean` is a property in NumPyro while it is a method in TFP.

### DIFF
--- a/numpyro/contrib/tfp/distributions.py
+++ b/numpyro/contrib/tfp/distributions.py
@@ -130,8 +130,6 @@ class TFPDistribution(NumPyroDistribution, metaclass=_TFPDistributionMeta):
 
         d = TFPDistribution[tfd.Normal](0, 1)
 
-    .. note:: Some special distributions like TransformedDistribution is not yet supported.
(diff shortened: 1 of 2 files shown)


martinjankowiak (approved):

fehiepsi:
Thanks for reviewing, @martinjankowiak!

PR #1065, opened 2021-06-11
Port LKJ example from Pyro

irustandi:
Port of the LKJ example (lkj.py) from Pyro


fehiepsi:
Hi @irustandi, thanks for the first PR!! We love to have more examples/tutorials that cover the practice usage of the framework. It would be nice to have some content that is not covered in the Pyro example (otherwise, maybe we move part of the code to LKJ docstring). Some of my suggestions
+ Find a good dataset/problem that you found LKJ is useful
+ Add some plots to interpret the result

irustandi:
The reason I created this example was because I was looking for such an example myself, and I saw there's one available in Pyro. I like the example because it's quite isolated. Definitely fine if this is put into the LKJ docstring.

fehiepsi:
I see. I think putting the modeling code to LKJ docstring would be nice, probably with a more interesting dataset (something that is not drawn from diagonal normal).

By the way, I do wish we have a tutorial for LKJ distribution. Do you want to take a stab at that? :)

irustandi:
Noted. Let me also think about doing the tutorial. Where is the source code for tutorials in general?

fehiepsi:
Yup, that's right! :D

numpyro/distributions/continuous.py
```diff
@@ -606,6 +625,26 @@ class LKJCholesky(Distribution):
...
+        def model(y): # y has dimension N x d
+            d = y.shape[1]
+            N = y.shape[0]
+            # Vector of variances for each of the d variables
+            theta = numpyro.sample("theta", dist.HalfCauchy(jnp.ones(d)))
+            # Lower cholesky factor of a correlation matrix
+            concentration = jnp.ones(1)  # Implies a uniform distribution over correlation matrices
+            L_omega = numpyro.sample("L_omega", dist.LKJCholesky(d, concentration))
+            # Lower cholesky factor of the covariance matrix
+            L_Omega = jnp.matmul(jnp.diag(jnp.sqrt(theta)), L_omega)
```
fehiepsi, line 639: Comment for a faster implementation would be useful
```
sigma = jnp.sqrt(theta)
# we can also use a faster formula `L_Omega = sigma[..., None] * L_omega`
L_Omega = jnp.matmul(jnp.diag(sigma), L_omega)
```

fehiepsi (commented):
Awesome! Those sample codes are very helpful! I just have a few small comments to pass CI.

irustandi:
Incorporated the comments, let me know if there is anything else.

fehiepsi (approved):
Very nice examples! Thanks a lot, @irustandi!

PR #1079, opened 2021-06-30
Easier installation for GPU/TPU

lumip:
Currently, numpyro installation instructions for GPU support ask the user to "first install `jax` and `jaxlib` with CUDA support." This is inconvenient and prone to lead to errors as the user then needs to watch out to install compatible versions of `jax` and `jaxlib`. Similar for TPU support.

Lately, JAX has improved its installation scripts by adding optional targets for GPU/TPU support that automatically install a CUDA/TPU version of `jaxlib` that is compatible with the installed version of `jax`, starting from version `0.2.11`. This PR adds the same targets to numpyro (and internally simply relies on the identical jax targets).

Installing numpyro for CUDA is now possible with
```
pip install numpyro.[cuda101] -f https://storage.googleapis.com/jax-releases/jax_releases.html
```
assuming CUDA 10.1 (similar for 10.2, 11.1, ..)
and for TPU
```
pip install numpyro.[tpu] -f https://storage.googleapis.com/jax-releases/libtpu_releases.html
```

The additional package index URLs are currently required by JAX and are a bit unfortunate, but overall this is still simpler and less error prone than the current approach.

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,7 +197,11 @@ pip install numpyro[cudaXXY] -f https://storage.googleapis.com/jax-releases/jax_
 where `XXY` is your CUDA XX.Y version number, e.g., for CUDA 10.2 use `cuda[102]` in the above.
 If you need further guidance, please have a look at the [JAX GPU installation instructions](https://github.com/google/jax#pip-installation-gpu-cuda).
 
-To run NumPyro on Cloud TPUs, you can use pip to install NumPyro as above and setup the TPU backend as detailed [here](https://github.com/google/jax/tree/master/cloud_tpu_colabs).
+To run NumPyro on Cloud TPUs, you need to setup the TPU backend as detailed [here](https://github.com/google/jax/tree/master/cloud_tpu_colabs)
+and then install NumPyro using the following pip command:
+```
+pip install numpyro[tpu] -f https://storage.googleapis.com/jax-releases/libtpu_releases.html
+```
 
 > **Default Platform:** JAX will use GPU by default if CUDA-supported `jaxlib` package is installed. You can use [set_platform](http://num.pyro.ai/en/stable/utilities.html#set-platform) utility `numpyro.set_platform("cpu")` to switch to CPU at the beginning of your program.
 
--- a/README.md
+++ b/README.md
@@ -184,21 +184,28 @@ Pyro users will note that the API for model specification and inference is large
 
 > **Limited Windows Support:** Note that NumPyro is untested on Windows, and might require building jaxlib from source. See this [JAX issue](https://github.com/google/jax/issues/438) for more details. Alternatively, you can install [Windows Subsystem for Linux](https://docs.microsoft.com/en-us/windows/wsl/) and use NumPyro on it as on a Linux system. See also [CUDA on Windows Subsystem for Linux](https://developer.nvidia.com/cuda/wsl) and [this forum post](https://forum.pyro.ai/t/numpyro-with-gpu-works-on-windows/2690) if you want to use GPUs on Windows.
 
-To install NumPyro with a CPU version of JAX, you can use pip:
+To install NumPyro with the latest CPU version of JAX, you can use pip:
 
 ```
 pip install numpyro
 ```
 
-To use NumPyro on the GPU, you need to install CUDA first and then use the following pip command:
+In case of compatibility issues arise during execution of the above ocmmand, you can instead force the installation of a known
+compatible CPU version of JAX with
+
+```
+pip install numpyro[cpu]
+```
+
+To use **NumPyro on the GPU**, you need to install CUDA first and then use the following pip command:
 ```
 pip install numpyro[cudaXXY] -f https://storage.googleapis.com/jax-releases/jax_releases.html
 ```
 where `XXY` is your CUDA XX.Y version number, e.g., for CUDA 10.2 use `cuda[102]` in the above.
 If you need further guidance, please have a look at the [JAX GPU installation instructions](https://github.com/google/jax#pip-installation-gpu-cuda).
 
-To run NumPyro on Cloud TPUs, you need to setup the TPU backend as detailed [here](https://github.com/google/jax/tree/master/cloud_tpu_colabs)
-and then install NumPyro using the following pip command:
+To run **NumPyro on Cloud TPUs**, you need to setup the TPU backend as detailed in the [Cloud TPU VM JAX Quickstart Guide](https://cloud.google.com/tpu/docs/jax-quickstart-tpu-vm). After you have verified that the TPU backend is properly set up,
+you can install NumPyro using the following pip command:
 ```
 pip install numpyro[tpu] -f https://storage.googleapis.com/jax-releases/libtpu_releases.html
 ```
--- a/README.md
+++ b/README.md
@@ -205,10 +205,7 @@ where `XXY` is your CUDA XX.Y version number, e.g., for CUDA 10.2 use `cuda[102]
 If you need further guidance, please have a look at the [JAX GPU installation instructions](https://github.com/google/jax#pip-installation-gpu-cuda).
 
 To run **NumPyro on Cloud TPUs**, you need to setup the TPU backend as detailed in the [Cloud TPU VM JAX Quickstart Guide](https://cloud.google.com/tpu/docs/jax-quickstart-tpu-vm). After you have verified that the TPU backend is properly set up,
-you can install NumPyro using the following pip command:
-```
-pip install numpyro[tpu] -f https://storage.googleapis.com/jax-releases/libtpu_releases.html
-```
+you can install NumPyro using the `pip install numpyro` command.
 
 > **Default Platform:** JAX will use GPU by default if CUDA-supported `jaxlib` package is installed. You can use [set_platform](http://num.pyro.ai/en/stable/utilities.html#set-platform) utility `numpyro.set_platform("cpu")` to switch to CPU at the beginning of your program.
 
--- a/README.md
+++ b/README.md
@@ -190,7 +190,7 @@ To install NumPyro with the latest CPU version of JAX, you can use pip:
 pip install numpyro
 ```
 
-In case of compatibility issues arise during execution of the above ocmmand, you can instead force the installation of a known
+In case of compatibility issues arise during execution of the above command, you can instead force the installation of a known
 compatible CPU version of JAX with
 
 ```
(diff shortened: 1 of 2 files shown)


fehiepsi:
Hi @lumip, this is great! I just wonder if https://github.com/google/jax/pull/7146 is relevant before merging. Currently, I don't have a GPU system to test this yet. :(

lumip:
Hey @fehiepsi : I opened the linked PR in an attempt to make this even nicer by being able to drop the additional package index url given to pip. However, due to the PyPI policy of not allowing packages to install dependencies that are not listed in a package index, the changes made by that PR were unfortunately not possible. That is, unless the CUDA versions of jaxlib become available on PyPI, this is currently the best we can do..

setup.py
```diff
@@ -9,6 +9,13 @@
 from setuptools import find_packages, setup
 
 PROJECT_PATH = os.path.dirname(os.path.abspath(__file__))
+_available_cuda_versions = [
+    "101",
+    "102",
+    "110",
+    "111",
+]  # TODO: align these with what's available in JAX before release
+_jax_version_constraints = ">= 0.2.11"
```
fehiepsi, line 18: nit: no space after `">=`

Does the latest jax release support `jax[cpu]` syntax? If not, I think we can wait until after the next numpyro release.

README.md
```diff
@@ -190,7 +190,12 @@ To install NumPyro with a CPU version of JAX, you can use pip:
...
 
-To use NumPyro on the GPU, you will need to first [install](https://github.com/google/jax#installation) `jax` and `jaxlib` with CUDA support.
+To use NumPyro on the GPU, you need to install CUDA first and then use the following pip command:
+```
+pip install numpyro[cudaXXY] -f https://storage.googleapis.com/jax-releases/jax_releases.html
+```
+where `XXY` is your CUDA XX.Y version number, e.g., for CUDA 10.2 use `cuda[102]` in the above.
+If you need further guidance, please have a look at the [JAX GPU installation instructions](https://github.com/google/jax#pip-installation-gpu-cuda).
 
 To run NumPyro on Cloud TPUs, you can use pip to install NumPyro as above and setup the TPU backend as detailed [here](https://github.com/google/jax/tree/master/cloud_tpu_colabs).
```
fehiepsi, line 200: Maybe we can rephrase this to `... setup the TPU backend as detailed here and use pip to install NumPyro as above`?

setup.py
```diff
@@ -32,8 +38,7 @@
     url="https://github.com/pyro-ppl/numpyro",
     author="Uber AI Labs",
     install_requires=[
-        "jax>=0.2.11",
-        "jaxlib>=0.1.62",
+        f"jax[cpu]{_jax_version_constraints}",
```
fehiepsi, line 41: Do you know what happens if `jax`, `jaxlib` GPU version is available and we want to update NumPyro
```
pip install -U numpyro
```
Will CPU versions replace the available GPU versions?

fehiepsi:
Understood! Thanks for explaining, @lumip!

setup.py
lumip: The `[cpu]` and `[tpu]` targets for jax are consistently available since jax version 0.2.11. However, I just noted that prior to 0.2.13 the GPU targets are `[cuda-xxx]` and since 0.2.13 its `[cudaxxx]`, which is a bit annoying. It might make sense to wait with this until we the jax dependency is bumped to 0.2.13 or above.

setup.py
lumip: This is indeed what happens (provided there was a change in the jax version NumPyro depends on). Updates would need to also specify the cuda option (and the jax release repo URL) in order not to replace with the CPU version.
One option would be to not install jaxlib unless the user specifies the target platform (which apparently what the folks at jax are going for). This means that jaxlib does not get replaced unintentionally but carries the risk of ending up with updated versions of numpyro and jax and an outdated (incompatible) jaxlib from the initial installation..

README.md
lumip: I have rephrased this now

lumip:
One further thing to consider with installation in general is that jaxlib does not maintain backward compatibility, so only constraining the version from below (as done previously in numpyro and would be the case through the jax[whatever] installation from this PR) means that users installing older versions will run into issues with high probability. I have raised an [issue](https://github.com/google/jax/issues/7174) with the folks at JAX about this, let's see what they think.

setup.py
fehiepsi: So I think we can remove this `cpu` option and set
```
f"jax>={_jax_version_constraints}",
"jaxlib>=0.1.62",
```
as previously. Since the next release, we will relax `jax/jaxlib` version - when people use numpyro in colab, they won't need to worry about jax/jaxlib installation (colab environment already has those versions)

README.md
fehiepsi: Thanks!

setup.py
fehiepsi: Please feel free to pump the minimum jax version to `0.2.13` and jaxlib to `0.1.65`. Currently, colab is using jax 0.2.13 and jaxlib 0.1.66 so its users won't have to reinstall (which potentially leads to conflicts) jax/jaxlib there.

fehiepsi:
> users installing older versions will run into issues with high probability

I think that if the new versions are available, we can just inform them to update the packages (if we suspect that the issue is due to an old release of jaxlib).

README.md
```diff
@@ -190,9 +190,18 @@ To install NumPyro with a CPU version of JAX, you can use pip:
...
+```
+where `XXY` is your CUDA XX.Y version number, e.g., for CUDA 10.2 use `cuda[102]` in the above.
+If you need further guidance, please have a look at the [JAX GPU installation instructions](https://github.com/google/jax#pip-installation-gpu-cuda).
 
-To run NumPyro on Cloud TPUs, you can use pip to install NumPyro as above and setup the TPU backend as detailed [here](https://github.com/google/jax/tree/master/cloud_tpu_colabs).
+To run NumPyro on Cloud TPUs, you need to setup the TPU backend as detailed [here](https://github.com/google/jax/tree/master/cloud_tpu_colabs)
+and then install NumPyro using the following pip command:
+```
+pip install numpyro[tpu] -f https://storage.googleapis.com/jax-releases/libtpu_releases.html
+```
```
fehiepsi, line 204: I think this command can be confusing. If I understand correctly, those people who want to run `pip install numpyro[tpu] ...` are using Cloud TPU VM (not Colab) and they want to skip [this step](https://cloud.google.com/tpu/docs/jax-quickstart-tpu-vm) in the Cloud TPU VM JAX setup process. I'm not sure we should encourage such simplification. In my opinion, we should encourage users to follow the upstream quickstart, test the environment, then install numpyro with `pip install numpyro`.

lumip:
> I think that if the new versions are available, we can just inform them to update the packages (if we suspect that the issue is due to an old release of jaxlib).

The problem is exactly the other way around: An installation of, say, numpyro v0.6.0, in the future may pull any newer jaxlib (or even jax) version which may well not be compatible, as neither make any guarantees about backward compatibility. I think it would be prudent to make sure that if the user wants to install a specific version of numpyro, they get it with working dependencies instead of being told to upgrade.
This is now fixed in jax for installation through the platform targets (`cpu`, `cuda`, `tpu`) by pinning a known compatible jaxlib version. However, numpyro installation might still pull an incompatible jax version (because it only constraints that from below).

README.md
lumip: Alright, I'll revisit this.

fehiepsi:
I see your point. In the past, we pinned jax/jaxlib versions to avoid breaking changes. But we think that jax/jaxlib has been more stable and requiring users to install `master` branch of numpyro is a bit inconvenient now. I'm not sure what jaxlib does after all so I can't tell how much its new releases will affect numpyro users. How about having `numpyro[cpu]` option to install a jaxlib version that is "assumed-by-jax-devs" compatible with the latest jax release? For the default installation, I think it is best, for now, to only check for jax/jaxlib requirements (and, as a pip policy, pip will try to install the latest versions if these dependencies are not available). Users can choose to `pip install numpyro[cpu]` to pin jaxlib if they want. What do you think?

README.md
lumip: Alright, maybe this is fine now?

fehiepsi (commented):
Yeah, let's see how things go. We might skip edge cases for now (e.g. some breaking changes that happen with any software or someone wants to upgrade jaxlib but keep the old jax version,...) CI can help us to detect the issues early so personally, I don't worry too much about them. Regarding `cpu`, I don't have a better name. It is fine I guess (at least it is concise :D).

README.md
```diff
@@ -184,15 +184,28 @@ Pyro users will note that the API for model specification and inference is large
...
 
-To install NumPyro with a CPU version of JAX, you can use pip:
+To install NumPyro with the latest CPU version of JAX, you can use pip:
 
 ```
 pip install numpyro
 ```
 
-To use NumPyro on the GPU, you will need to first [install](https://github.com/google/jax#installation) `jax` and `jaxlib` with CUDA support.
+In case of compatibility issues arise during execution of the above ocmmand, you can instead force the installation of a known
```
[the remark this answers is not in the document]
lumip: fixed

README.md
```diff
@@ -184,15 +184,28 @@ Pyro users will note that the API for model specification and inference is large
...
+compatible CPU version of JAX with
 
-To run NumPyro on Cloud TPUs, you can use pip to install NumPyro as above and setup the TPU backend as detailed [here](https://github.com/google/jax/tree/master/cloud_tpu_colabs).
+```
+pip install numpyro[cpu]
+```
+
+To use **NumPyro on the GPU**, you need to install CUDA first and then use the following pip command:
+```
+pip install numpyro[cudaXXY] -f https://storage.googleapis.com/jax-releases/jax_releases.html
```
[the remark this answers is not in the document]
lumip: done

README.md
```diff
@@ -184,15 +184,28 @@ Pyro users will note that the API for model specification and inference is large
...
+```
+where `XXY` is your CUDA XX.Y version number, e.g., for CUDA 10.2 use `cuda[102]` in the above.
+If you need further guidance, please have a look at the [JAX GPU installation instructions](https://github.com/google/jax#pip-installation-gpu-cuda).
+
+To run **NumPyro on Cloud TPUs**, you need to setup the TPU backend as detailed in the [Cloud TPU VM JAX Quickstart Guide](https://cloud.google.com/tpu/docs/jax-quickstart-tpu-vm). After you have verified that the TPU backend is properly set up,
```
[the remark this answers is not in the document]
lumip: done

fehiepsi (approved):
This is extremely helpful, thanks @lumip! Please let me know if you have any concerns that need further discussion or this is ready to merge. :)

lumip:
I think I'm happy with it now. I can squash it into a single commit still..

lumip:
Thanks for the review and the discussion to make improve this!

Issue #1070, opened 2021-06-22
VonMises distribution behaves poorly at the boundary


alexlyttle:
While converting my current work to `numpyro`, I noticed that the `VonMises` distribution doesn't sample well, especially when μ (or `loc`) is near ±π. I believe this is because the `support` for the `VonMises` distribution is the interval constraint from -π to +π. This does not allow the sampler to step from -π to +π or +π to -π by going over the boundary. I show the problem and a solution in my Gist [here](https://gist.github.com/alexlyttle/9510e10e0951fec356b9fc5bdd205f27) with plots below.

I would love to submit a PR with this fix if you agree that it improves/fixes the current `VonMises` distribution?

### Solution

<details>
<summary>Click to expand</summary>

See [here](https://gist.github.com/alexlyttle/9510e10e0951fec356b9fc5bdd205f27)  for code. My solution is to register a circular constraint as the `support` for the `VonMises` distribution by using a `CircularTransform` to allow the sampler to go over the boundary. The circular transform looks like this:

![circular_transform](https://user-images.githubusercontent.com/43786145/122927915-3fc38a00-d361-11eb-9a27-6b87821ccc00.png)

In the following plot, I show the trace for two variables sampled across 10 chains using the `numpyro` `NUTS` sampler: `phi_old` for the current VonMises distribution, and `phi_new` for my suggested fix.

![trace](https://user-images.githubusercontent.com/43786145/122928107-7699a000-d361-11eb-93f6-a516f8af497b.png)

</details>

fritzo:
Hi @alexlyttle your solution makes sense to me!  I think we'll want to add that as a reparametrizer and register the new circular constraint and transform with `transform_to()` but not `biject_to()`, and I think we'll want to add a new improper distribution called say `Sinusoidal` or `ImproperSinusoidal` or something.

[...]

PR #1080, opened 2021-07-01
Circular Reparameterization

alexlyttle:
Add circular reparameterization to improve the sampling of the `VonMises` distribution re. issue #1070.

To Do:

- [x] Add `CircularReparam` class
- [x] Add `circular = _Circular()` constraint
    - Must be in the range -pi < x < pi
- [x] Change support for `VonMises` to circular constraint
- [x] Modify existing tests for `VonMises` distribution
- [x] Add tests for `CircularReparam`
- [x] Update docs

Issues so far:

- `VonMises` does not work without using `CircularReparam` and so `test_distribution_constraints` fails with `NotImplementedError: <numpyro.distributions.constraints._Circular object at 0x13129f520> not implemented.`.
    - Do we want to change these tests and change the docs so that it tells the user to use `VonMises` with  `CircularReparam`? 
    - I understand there are plans to make reparameterization automatic in the future, so this shouldn't be an issue?


fritzo:
> Do we want to change these tests and change the docs so that it tells the user to use VonMises with CircularReparam?

Yes I think so. I believe you can follow the pattern for the `ProjectedNormal` distribution and `constraints.sphere`, including a new case in [helpful_support_errors()](https://github.com/pyro-ppl/numpyro/blob/ca6811bfc7f4fae9d93f012e13b252acacb1213b/numpyro/infer/util.py#L904).

> I understand there are plans to make reparameterization automatic in the future, so this shouldn't be an issue?

Yes, our vague vision is for reparametrizers to handle cases where bijections don't suffice.  We started out with bijections registered with `biject_to()`, but those don't handle many-to-one transforms.  It's still unclear to me how much functionality should be in each of `biject_to` and `handlers.reparam`, but I would like to make `handlers.reparam` as nearly-automatic as our constraints registered with `biject_to`.

alexlyttle:
> Yes I think so. I believe you can follow the pattern for the `ProjectedNormal` distribution and `constraints.sphere`, including a new case in [helpful_support_errors()](https://github.com/pyro-ppl/numpyro/blob/ca6811bfc7f4fae9d93f012e13b252acacb1213b/numpyro/infer/util.py#L904).

Great, thanks! I have added a `VonMises` docstring and to `helpful_support_errors()` (following `ProjectedNormal`) and added to `test_distributions` to account for the new `circular` constraint.

> Yes, our vague vision is for reparametrizers to handle cases where bijections don't suffice. We started out with bijections registered with `biject_to()`, but those don't handle many-to-one transforms. It's still unclear to me how much functionality should be in each of `biject_to` and `handlers.reparam`, but I would like to make `handlers.reparam` as nearly-automatic as our constraints registered with `biject_to`.

This sounds good!

I still need to come up with a test for `CircularReparam` but this shouldn't take too long.

numpyro/distributions/directional.py
```diff
@@ -20,9 +20,22 @@
...
+    """
+    The von Mises distribution, also known as the circular normal distribution.
+
+    To use this distribution with autoguides and HMC, use ``handlers.reparam``
+    with a :class:`~numpyro.infer.reparam.CircularReparam` reparametrizer in
+    the model, e.g.::
+
+        @handlers.reparam(config={"direction": CircularReparam()})
+        def model():
+            direction = numpyro.sample("direction", VonMises(0.0, 4.0)
```
fritzo, line 32: nit: add closing paren

numpyro/distributions/constraints.py
```diff
@@ -96,6 +98,14 @@ def feasible_like(self, prototype):
         return jax.numpy.zeros_like(prototype)
 
 
+class _Circular(Constraint):
+    def __call__(self, x):
+        return (x >= -math.pi) & (x < math.pi)
+
+    def feasible_like(self, prototype):
+        return jax.numpy.broadcast_to(math.pi, jax.numpy.shape(prototype))
```
fritzo, line 106: maybe use 0.0 instead, since `math.pi` doesn't strictly satisfy the constraint, and `-math.pi` is at a boundary

fritzo:
@alexlyttle looks great, do you mind if I port this to Pyro quick for our upcoming 1.7 release?

numpyro/distributions/directional.py
alexlyttle: Good spot, thanks will do

numpyro/distributions/constraints.py
alexlyttle: That makes sense!

alexlyttle:
> @alexlyttle looks great, do you mind if I port this to Pyro quick for our upcoming 1.7 release?

Thanks, go for it!

fritzo:
Hmm on closer inspection it looks like PyTorch simply uses `constraints.real` for the `VonMises` constraint, which ends up working fine during inference (both HMC and SVI).  I think that's actually a decent solution, and since changing `pyro.distributions.VonMises.support = circular` would break backwards compatibility, I think it's better to just leave the Pyro/PyTorch version as-is.

@fehiepsi do you think it would break anyone's code to simply relax
```py
numpyro.distributions.VonMises.support = constraints.real
``` as an improper distribution, thereby addressing @alexlyttle's original issue in one line?

**Comment by fehiepsi** [comment]
I'm not sure what's the best action here. Matching PyTorch behavior is nice but I think we need to discuss this a bit more. After reading through the previous discussions, here is what I understand (assuming that posterior is single-modal on the circle):

### using VonMises with real support

+ the distribution is improper
+ posterior is multi-modal. MCMC chain diagnostics will be off because the chains will likely stay at different 2pi periods
+ seems to fit well for 1-chain MCMC

### using VonMises with interval(-pi, pi) support

+ IIUC, posterior is 2-modal if it concentrates at the boundary (this is the original issue.) A "non-automatic" solution for this is to shift the center by `pi` (or simply by an estimated value of `loc`) and then shift the samples back. I guess it is a fair trade-off.

### using VonMises with circular support

+ posterior is multi-modal. This seems to face similar issues as VonMises with real support
+ reparam is needed

I feel that using real support is more suitable for SVI and interval support is more support for MCMC.

**Comment by OlaRonning** [comment]
@fehiepsi, regarding #1063 the support would become `constraints.independent(constraints.circular, 1)`, and a reparam is needed regardless. For #1055 a circular support would make it easy to check the antecedent distribution has the correct support.

My personal opinion is a circular support is the best solution; neither `real` nor `interval(-pi, pi)` encodes the periodic boundary.

**Comment by alexlyttle** [comment]
Apologies for slow progress. From the discussion above it sounds like this idea is still good to go? I'm working on the tests and changes today.

@OlaRonning

> My personal opinion is a circular support is the best solution; neither real nor interval(-pi, pi) encodes the periodic boundary.

If people are happy with this that's great! Technically, the `circular` support I have put in doesn't really encode the periodic boundary, it's using the `CircularReparam` which does that. The circular support is more representative of the distribution being circular (similar to `sphere` constraint for `ProjectedNormal`) and bound by -pi and +pi. In #1070 I discussed this with @fritzo and we decided to define the `circular` constraint as similar to `interval(-pi, +pi)`. I hope that makes sense!

**Comment by fehiepsi** [comment]
I just have a random thought that we can have `circular` support that behaves as `constraints.interval(...)`. This allows users to control the behavior as they want without having to redefine the class with some default behavior:
+ using `circular` without reparam: this uses `biject_to(interval)` for transformation and behaves as `interval(-pi, pi)` as previously. I think we can achieve this by simply changing the `raise ValueError` to a warning, so users get informed about the behavior under the hood.
+ using `circular` with reparam: as the current state of this PR

**Comment by alexlyttle** [comment]
@fehiepsi so have `circular` inherit `constraints.interval`? I think I had it like that first, but @fritzo suggested that the sigmoid transform in `biject_to(interval)` might interfere with `CircularReparam` or do an unnecessary transformation. What do you think?

**Comment by alexlyttle** [comment]
I've been looking into this more. Attempting to recreate the `ProjectedNormalReparam` test but for `CircularReparam`, it doesn't produce the same distribution. This is because in `CircularReparam` we sample N(0, 1) noise and then make the transformation. So when reparam is used in the following way we just get the N(0, 1) distribution.

```python 
import numpy as np
import numpyro
import arviz as az

from numpyro.infer.reparam import CircularReparam

def model(loc, concentration, n):
    with numpyro.plate_stack("plates", shape):
        with numpyro.plate("particles", n):
            numpyro.sample("x", dist.VonMises(loc, concentration))

shape = ()
loc = np.zeros(shape) + 2.0
conc = np.ones(shape)

with numpyro.handlers.trace() as trace_exp:
        model(loc, conc, 10000)

with numpyro.handlers.reparam(config={"x": CircularReparam()}):
    with numpyro.handlers.trace() as trace_act:
            model(loc, conc, 10000)

az.plot_trace(np.array(trace_exp['x']['value']))
az.plot_trace(np.array(trace_act['x']['value']))

```

image

image

@fritzo why did you suggest sampling parameter-free noise in `CircularReparam` [here](https://github.com/pyro-ppl/numpyro/issues/1070#issuecomment-866197492) over something more like the following code? Using the code below with `VonMises.support = constraints.real`, when doing the above test you get the same distributions, and it works with MCMC sampler. This way reparam would only be required for MCMC. Or, is there something I am not understanding here?

```python
class CircularReparam(Reparam):
    def __call__(self, name, fn, obs):

        # Draw from VonMises with constraints.real support
        value = numpyro.sample(
            f"{name}_unwrapped",
            fn,
            obs=obs,
        )

        # Differentiably transform.
        value = jnp.remainder(value + math.pi, 2 * math.pi) - math.pi

        numpyro.factor(f"{name}_factor", fn.log_prob(value))
        return None, value
```

fritzo:
> have `circular` inherit `constraints.interval`?

I like @fehiepsi's suggestion, and this inheritance now seems like the right design choice.  I had been worried about allowing a default registration for `biject_to(circular)`, but it seems ok to allow that registration but emit a warning in `helpful_support_errors()`.

@alexlyttle the motivation for my suggestion of 
```py
numpyro.sample(f"{name}_unwrapped", dist.Normal(0,1).mask(False), obs=obs)
```
is that that sample statement gets the constraint correct and works during inference. If we tried to use the original `fn` as in
```py
numpyro.sample(f"{name}_unwrapped", fn, obs=obs)
```
then the original `fn.support` would be used and we'd get nowhere.  Note that when reparametrizing via `Normal(0,1).mask(False)`, the density and posterior distributions will be correct, but sampling from the prior will be incorrect (as in your snippet).  I'm not sure whether that matters, and if so I'm not sure how to fix it (@fehiepsi perhaps by providing an optional `sampler` arg to `ImproperUniform`?).

fehiepsi:
I don't have a better idea. Having `sampler` argument seems reasonable to me. For now, I think it is fine to just use ImproperUniform(constraints.real, ...) and by default, it will raise an error if a user wants to get samples from it. To get samples from priors, users can run the non-reparam model.

Alternatively, we can just create a new instance of `fn` with `support` is real. But I think it is unnecessary.

alexlyttle:
Hi @fritzo thanks for the explanation, it makes sense to me. I am not 100% sure on the meaning of `mask(False)` likely due to still not being experienced with `pyro`/`numpyro`.

@fehiepsi this sounds good. I have used ImproperUniform instead and the `reparam` still works fine.

I have drafted a test for `CircularReparam`. Since getting samples using the reparam now raises an error, I have written a test which compares making samples without to MCMC inference with `CircularReparam`. The `get_moments` function isn't reliable here for circular data so instead I compare the lengths of the first 4 circular moments in the test. Hopefully it will make sense in my next commits.

numpyro/infer/util.py
```diff
@@ -870,4 +870,11 @@ def helpful_support_errors(site):
                 "a reparameterizer, e.g. "
                 f"numpyro.handlers.reparam(config={{'{name}': ProjectedNormalReparam()}})."
             )
+        if "circular" in support_name:
+            raise ValueError(
```
fehiepsi, line 874: raise a warning instead. In VonMises tests, you can skip the warning

numpyro/infer/reparam.py
```diff
@@ -288,3 +289,27 @@ def transform_sample(self, latent):
...
+        fn, expand_shape, event_dim = self._unwrap(fn)
+
+        # Draw parameter-free noise.
+        # new_fn = dist.Normal(0, 1).mask(False)
+        new_fn = dist.ImproperUniform(
+            constraints.real, original_fn.batch_shape, original_fn.event_shape
+        )
+        value = numpyro.sample(
+            f"{name}_unwrapped",
+            self._wrap(new_fn, expand_shape, event_dim),
```
fehiepsi, line 306: I think this is not needed. You can remove the above unwrapped logic and this wrap logic and use the new_fn directly here.

numpyro/infer/reparam.py
```diff
@@ -288,3 +289,27 @@ def transform_sample(self, latent):
         """
         x_unconstrained = self.transform(latent)
         return self.guide._unpack_and_constrain(x_unconstrained, self.params)
+
+
+class CircularReparam(Reparam):
+    def __call__(self, name, fn, obs):
+        original_fn = fn
```
fehiepsi, line 296: No need for this, you can use `fn` directly in the below logic.

numpyro/distributions/constraints.py
```diff
@@ -53,6 +53,8 @@
     "Constraint",
 ]
```
fehiepsi, line 55: add circular to __all__ dictionary

fehiepsi (commented):
The mask just simply zeroes out the log probability. It has the same effect as ImproperUniform in term of density calculation. Could you add circular and CircularReparam to `docs/source/distributions or reparms.rst` file?

numpyro/infer/reparam.py
```diff
@@ -288,3 +289,27 @@ def transform_sample(self, latent):
...
+        fn, expand_shape, event_dim = self._unwrap(fn)
```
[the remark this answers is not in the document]
alexlyttle: > The mask just simply zeroes out the log probability. It has the same effect as ImproperUniform in term of density calculation. Could you add circular and CircularReparam to `docs/source/distributions or reparms.rst` file?

Thanks for the review, I'll give this a go!

numpyro/infer/util.py
alexlyttle: I've changed this to a UserWarning and reworded slightly. I presumed I could revert my changes to `test_distributions.py` in commit a3ca41e because `isinstance(constraint.circular, constraint.interval) == True` so it just defaults to generating values in that interval for the tests. The tests pass when I do this.

I haven't found instances in the tests where we'd need to skip the new warning. But then I haven't fetched and merged from `pyro-ppl/numpyro/master` recently so there may be changes to the tests which I'll make sure pass.

numpyro/distributions/directional.py
```diff
@@ -20,9 +20,22 @@
 
 
 class VonMises(Distribution):
+    """
+    The von Mises distribution, also known as the circular normal distribution.
+
+    To use this distribution with autoguides and HMC, use ``handlers.reparam``
```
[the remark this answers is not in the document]
alexlyttle: It now reads

```python
"""

...

This distribution is supported by a circular constraint from -pi to +pi. By
default, the circular support behaves like 
``constraints.interval(-math.pi, math.pi)``. To avoid issues at the
boundaries of this interval during sampling, you should reparameterize this
distribution using ``handlers.reparam`` with a 
:class:`~numpyro.infer.reparam.CircularReparam` reparametrizer in
the model, e.g.::

...

"""
```

numpyro/infer/reparam.py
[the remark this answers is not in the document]
alexlyttle: I have added this

```python
support = fn.support
if isinstance(support, constraints._IndependentConstraint):
    support = fn.support.base_constraint
assert isinstance(fn.support, constraints._Circular)
```

numpyro/infer/reparam.py
alexlyttle: Makes sense, I've done this too.

numpyro/distributions/constraints.py
alexlyttle: Done

alexlyttle:
I have updated the docs and will push soon. I am getting less confident in the `remainder` solution for the reparam. When testing `CircularReparam` in my work I found that `arctan2(sin(x), cos(x))` performed better during inference than the `remainder` method. By running better, I mean for the same model, data, random key, samples, warmup, chains etc. `remainder` resulted in more divergences and higher r_hat statistics. I looked into it and found that 

`jnp.remainder(x + math.pi, 2 * math.pi) - math.pi` is slightly different to `jnp.arctan2(jnp.sin(x), jnp.cos(x))`. See the plots below for the difference.

This difference is tiny, but could be affecting inference. I'm not sure which method is right, my instinct says `arctan` because it performs better. Comparing the speed of the two in a `jit` compiled function they are as fast as each other, so I don't see much speed-advantage to choosing `remainder` over `arctan`.

(32-bit precision)
image

(64-bit precision)
image

numpyro/distributions/directional.py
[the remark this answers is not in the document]
fehiepsi: Thanks, that's a great description!

fehiepsi:
Hi @alexlyttle, for now, I don't have much intuition for the difference between `(a + pi) % 2pi - pi` and `arctan2(sin, cos)`. I'll take a closer look sometime this weekend and will let you know.

fehiepsi:
@alexlyttle It seems to me that we lost precision with `arctan`
```python
import jax.numpy as jnp
import numpy as np

def f1(x):
    return (x + jnp.pi) % (2 * jnp.pi) - jnp.pi

def f2(x):
    return jnp.arctan2(jnp.sin(x), jnp.cos(x))

def f3(x):
    return (x.copy() + np.pi) % (2 * np.pi) - np.pi

x = jnp.linspace(-10., 10., 1000)
y1 = f1(x)
y2 = f2(x)
y3 = f3(x)

print("numpy vs remainder:", abs(y1 - y3).max())
print("numpy vs arctan:", abs(y2 - y3).max())
plt.plot(x, y2 - y3, "o");
```
```
numpy vs remainder: 0.0
numpy vs arctan: 7.1525574e-07
```

> more divergences and higher r_hat statistics

I am a bit curious. Does this happen persistently across random seeds? How much different the results are? Maybe MCMC is happier with imprecise computation in this case. >__<

alexlyttle:
@fehiepsi Thanks for investigating, I found the same thing and couldn't understand why arctan seemed to produced better inference. I will check in case it was to do with the random seed.

> Btw, you can find and resolve lint issues by running make format or make lint.

Thanks! I forgot to do this last time round!

alexlyttle:
@fehiepsi It seems it is the choice of random key that was the issue, I should remember to vary it more often! The remainder method is perfectly fine.

Just to summary changes I am about to push:

When `CircularReparam` was not used, MCMC didn't just fall back to using Interval(-pi, +pi) support as I thought. We want this to happen because no reparam is still useful in some cases. To get it to do this, I changed the circular constraint to `circular = _Interval(-math.pi, math.pi)` rather than its own class. `VonMises` now samples without reparam -- e.g. this can be useful if your VonMises prior is ~ uniform and you want to do a prior=predictive check.

We wanted a warning to encourage the user to consider `CircularReparam`, so I added this to `helpful_support_errors`. Since `VonMises.support` no longer throws a `NotImplementedError`, I added the warning before the try/except. I'm not sure this is the perfect solution, because this warning shows up several times if running MCMC with many chains. We probably only want the warning to occur on model initialisation, so I could add a separate `helpful_support_warnings` context manager which is only called once?

Otherwise, this work should be nearly done!

numpyro/infer/util.py
```diff
@@ -962,10 +962,21 @@ def single_loglik(samples):
 
 @contextmanager
 def helpful_support_errors(site):
+    # Warnings
+    name = site["name"]
+    support = getattr(site["fn"], "support", None)
```
fehiepsi, line 967: for completeness, you can add
```
if isinstance(support, constraints.independent):
    support = support.base_constraint
```

fehiepsi (dismissed):
LGTM, thanks @alexlyttle! Regarding the multiple warnings, one way is to add a boolean argument `raise_warnings` to `helpful_support_errors` and raise it [here](https://github.com/pyro-ppl/numpyro/blob/b33f808baede1c7278dd74c13866c9c6c933fdc3/numpyro/infer/util.py#L400).

numpyro/infer/util.py
alexlyttle: Makes sense, I've added that now

alexlyttle:
I've added the `raise_warnings` flag and the warning now only occurs once as hoped. It looks like we're good to go! Thanks for everyone's help on this one 😊

fehiepsi (approved):
Happy to merge this! Thanks so much for addressing this tricky issue.

PR #1082, opened 2021-07-02
Checklist for 0.7 release

fehiepsi:
### TODO

- [x] Bump versions in examples/tutorials
- [x] Bump funsor version
- [x] Update docker file
- [x] Collect changelogs.
- [x] Merge LDA example PR
- [x] Support infer discrete in Predictive
- [x] Some python blocks do not [rendered well](http://num.pyro.ai/en/stable/getting_started.html#frequently-asked-questions) in README
- [x] Review setup.py and docs/requirements files to make sure that the dependencies are stated correctly
- ~[ ] (optional) Add some examples to Predictive  #1084~ Defer to a later release
- ~[ ] Get #1058 merged~ (optional)
- ~[ ] (optional) add tutorial for truncated distribution~ Defer to a later release

PR #1084, opened 2021-07-04
DOC Add Predictive examples

MarcoGorelli:
xref https://github.com/pyro-ppl/numpyro/pull/1082#issuecomment-872914309

I thought it might be clearer to slightly edit the existing examples in SVI and MCMCKernel, which already have runnable code, and then to refer to them in the Predictive kernel

Here's the part added to Predictive:

image

SVI:

image

MCMCKernel:

image

MCMC.run:

image

---

EDIT
----

some doctest failure...I'll fix this later, the example in `numpyro/infer/util.py` needs to just be code, but without doctest running on it


numpyro/infer/svi.py
```diff
@@ -101,6 +101,8 @@ class SVI(object):
         >>> svi = SVI(model, guide, optimizer, loss=Trace_ELBO())
         >>> svi_result = svi.run(random.PRNGKey(0), 2000, data)
         >>> params = svi_result.params
+        >>> predictive = Predictive(model, guide=guide, params=params, num_samples=1000)
+        >>> samples = predictive(random.PRNGKey(0), data)
```
fehiepsi, line 105: This code does not return any useful information unless you set `data=None` (which will give you posterior predictive samples). If you want to get posterior samples, then you can move those codes to the end and add a comment
```python
>>> # get posterior samples
>>> predictive = Predictive(guide, params=params, num_samples=1000)
>>> samples = predictive(random.PRNGKey(1), data)
```

numpyro/infer/mcmc.py
```diff
@@ -64,7 +64,9 @@ class MCMCKernel(ABC):
         >>> kernel = MetropolisHastings(f)
         >>> mcmc = MCMC(kernel, num_warmup=1000, num_samples=1000)
         >>> mcmc.run(random.PRNGKey(0), init_params=jnp.array([1., 2.]))
-        >>> samples = mcmc.get_samples()
+        >>> posterior_samples = mcmc.get_samples()
+        >>> predictive = Predictive(model, posterior_samples=posterior_samples)
+        >>> samples = predictive(rng_key1, *model_args, **model_kwargs)
```
fehiepsi, line 69: I don't think we have a model here. I think you can add a small code snippet to `get_samples` method of `MCMC`.

numpyro/infer/mcmc.py
```diff
@@ -509,6 +511,8 @@ def run(self, rng_key, *args, extra_fields=(), init_params=None, **kwargs):
         """
         Run the MCMC samplers and collect samples.
 
+        See :class:`~numpyro.infer.util.Predictive` for how to use them to collect posterior predictive samples.
```
fehiepsi, line 514: I think `get_samples` is a better place for this. Also maybe changing `them` to `those samples`?

numpyro/infer/util.py
```diff
@@ -743,6 +743,21 @@ class Predictive(object):
           argument is not None, its value should be equal to `num_chains x N`.
 
     :return: dict of samples from the predictive distribution.
+
+    **Example:**
+
+    Given a model, you can sample from the prior predictive:
+
+        >>> predictive = Predictive(model, num_samples=num_samples)
```
fehiepsi, line 751: nit `num_samples=1000`

fehiepsi (commented):
Thanks, @MarcoGorelli! I think you can remove `>>> ` to skip doctest.

MarcoGorelli:
Thanks for your feedback! I'll address your comments at the weekend (won't have time this week before then unfortunately)

fehiepsi:
Please take your time. We can always defer doc improvements to later releases. :)

numpyro/infer/mcmc.py
```diff
@@ -593,6 +593,14 @@ def get_samples(self, group_by_chain=False):
...
             but can be any :func:`jaxlib.pytree`, more generally (e.g. when defining a
             `potential_fn` for HMC that takes `list` args).
+
+        **Example:**
+
+        You can then pass those samples to :class:`~numpyro.infer.util.Predictive`:
+
+            posterior_samples = mcmc.get_samples()
+            predictive = Predictive(model, posterior_samples=posterior_samples)
+            samples = predictive(rng_key1, *model_args, **model_kwargs)
```
fehiepsi, line 603: I think you need an empty string after this to make lint pass.

MarcoGorelli: I think it might be the `*` - will try making it a `r` string

MarcoGorelli: OK, fixed, it just needed to be put into a code block, so with the previous line ending with `::` rather than `:`

MarcoGorelli: now it looks like this:

Screenshot 2021-07-11 at 21 17 21

fehiepsi: yeah, good catch!

fehiepsi (approved):
Awesome, thanks @MarcoGorelli!

PR #1086, opened 2021-07-08
Support infer_discrete for Predictive

fehiepsi:
Follow-up #1043, this PR adds arguments `infer_discrete` and `temperature` to `Predictive` to make it easier for users to get samples for discrete latent variables (illustrated in the annotation example).

### DIFF
--- a/numpyro/contrib/funsor/discrete.py
+++ b/numpyro/contrib/funsor/discrete.py
@@ -118,8 +118,7 @@ def _sample_posterior(
             values = [v.reshape((-1,) + prototype_shape[1:]) for v in values]
             data[root_name] = jnp.concatenate(values)
 
-    with substitute(data=data):
-        return model(*args, **kwargs)
+    return data
 
 
 def infer_discrete(fn=None, first_available_dim=None, temperature=1, rng_key=None):
@@ -169,6 +168,12 @@ def infer_discrete(fn=None, first_available_dim=None, temperature=1, rng_key=Non
             temperature=temperature,
             rng_key=rng_key,
         )
-    return functools.partial(
-        _sample_posterior, fn, first_available_dim, temperature, rng_key
-    )
+
+    def wrap_fn(*args, **kwargs):
+        samples = _sample_posterior(
+            fn, first_available_dim, temperature, rng_key, *args, **kwargs
+        )
+        with substitute(data=samples):
+            return fn(*args, **kwargs)
+
+    return wrap_fn
--- a/numpyro/infer/util.py
+++ b/numpyro/infer/util.py
@@ -674,14 +674,13 @@ def _predictive(
     posterior_samples,
     batch_shape,
     return_sites=None,
-    infer_discrete=False,
-    temperature=1,
+    infer_discrete_temperature=None,
     parallel=True,
     model_args=(),
     model_kwargs={},
 ):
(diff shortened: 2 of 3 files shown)


numpyro/infer/util.py
```diff
@@ -673,17 +674,47 @@ def _predictive(
     posterior_samples,
     batch_shape,
     return_sites=None,
+    infer_discrete=False,
+    temperature=1,
```
fritzo, line 678: nit: maybe call this `infer_discrete_temperature` to distinguish from sampling from annealed posteriors?

fehiepsi: Good idea! I'll make the change.

numpyro/contrib/funsor/discrete.py
```diff
@@ -118,8 +118,7 @@ def _sample_posterior(
             values = [v.reshape((-1,) + prototype_shape[1:]) for v in values]
             data[root_name] = jnp.concatenate(values)
 
-    with substitute(data=data):
-        return model(*args, **kwargs)
+    return data
```
fehiepsi, line 121: I revise this a bit to use this function in Predictive (and avoid tracing the model 2 times)

numpyro/infer/util.py
```diff
@@ -673,17 +674,47 @@ def _predictive(
...
-            *model_args, **model_kwargs
-        )
+        if infer_discrete:
+            from numpyro.contrib.funsor import config_enumerate
+            from numpyro.contrib.funsor.discrete import _sample_posterior
+
+            model_trace = prototype_trace
+            temperature = 1
+            pred_samples = _sample_posterior(
+                config_enumerate(condition(model, samples)),
```
fritzo, line 705: I'm surprised you're automatically configuring the model for enumeration here.  In Pyro we let the user decide which variables are enumerated.  That way a guide might sample some discrete latents, and the model might enumerate others.  I would expect `Predictive` to use the guide's samples if provided, and only enumerate sites already marked for enumeration by the user.

I guess if in NumPyro you automatically wrap models with `@config_enumerate` inside MCMC, then it would also make sense to automatically wrap them here.  Still, it would be nice to support SVI with guides that sample some or all discrete latent variables.

fritzo (approved):
LGTM as long as you're ok wrapping models with `@config_enumerate`.

numpyro/infer/util.py
fehiepsi: > let the user decide which variables are enumerated

We'll aim for this except for algorithms that do not work with discrete latent sites (in this case, it makes sense to enable enumeration by default and raise errors for invalid models). Currently, DiscreteHMCGibbs will perform Gibbs update for discrete latent sites that are not marked "enumerated". Then posterior samples will include all latent variables except for those marked "enumerated".

When `infer_discrete` is used, I assumed that the latent sites that are not available in the `posterior_samples` or `guide` are enumerated (those latent sites are the `samples` variable in the above code). So it makes sense to me to `config_enumerate` them by default (`config_enumerate` will skip observed sites, including those sites in `samples`). What do you think? (it seems to not contradict with the usage case in your comment)

fehiepsi:
Thanks for reviewing, @fritzo! I thought about it for a while and I think we can merge this and revisit the API later if needed (especially when TraceEnumELBO is available). We can check the model and raise a deprecation warning if `config_enuemerate` is required in the future.

PR #1087, opened 2021-07-08
Example/ssbvm mixture

OlaRonning:
A mixture model for protein dihedral angle prediction. The model uses directional likelihoods.


examples/ssbvm_mixture.py
```diff
@@ -0,0 +1,297 @@
...
+    "P",
+    "A",
+    "V",
+    "Q",
+    "W",
+    "C",
+]
+
+
+@config_enumerate
```
fehiepsi, line 83: `config_enumerate` is not needed because you already have `infer={"enumerate": "parallel"}`

examples/ssbvm_mixture.py
```diff
@@ -0,0 +1,297 @@
...
+@numpyro.handlers.reparam(
+    config={"phi_loc": CircularReparam(), "psi_loc": CircularReparam()}
```
fehiepsi, line 85: Maybe add a note for why we use reparam here? It would be helpful for other readers.

numpyro/examples/datasets.py
```diff
@@ -313,9 +326,18 @@ def iter_dataset(dset, batch_size=None, split="train", shuffle=True):
 
 
 def load_dataset(
-    dset, batch_size=None, split="train", shuffle=True, num_datapoints=None
+    dset,
+    batch_size=None,
+    split="train",
+    shuffle=True,
+    num_datapoints=None,
+    subset_key=None,
```
fehiepsi, line 334: Would it be simpler to use `split` for this?
```
def _load_9mers():
    _download(NINE_MERS)
    file_path = os.path.join(DATA_DIR, "9mers_data.pkl")
    return pickle.load(open(file_path, "rb"))
```

fehiepsi (commented):
Awesome example! Thanks, Ola. I just have a few comments

examples/ssbvm_mixture.py
```diff
@@ -0,0 +1,297 @@
...
+Example: Sine-skewed sine (bivariate von Mises) mixture
+=======================================================
+
+This example models the dihedral angles that occur in the backbone of a protein as a mixture of skewed
+directional distributions. The backbone angle pairs, called :math:`\phi` and :math:`\psi`, are a canonical
+representation for the fold of a protein. In this model, we fix the third dihedral angle (omega) as it usually only
+takes angles 0 and pi radian, with the latter being the most common. We model the angle pairs as a distribution on
+the torus using the sine distribution [1] and break point-wise (toroidal) symmetry using sine-skewing [2].
+
+.. image:: ../_static/img/examples/torus_top.png
```
fehiepsi, line 14: Maybe `ssbvm_mixture_torus_top.png` is a better name.

examples/ssbvm_mixture.py
OlaRonning: I see, I'll take it out.

examples/ssbvm_mixture.py
OlaRonning: Make sense; I'll update it.

numpyro/examples/datasets.py
OlaRonning: Makes sense, and is much simpler. I'll update it.

fehiepsi (approved):

PR #1090, opened 2021-07-08
Add L1 ball constraint and its corresponding L1BallTransform bijective

fehiepsi:
This constraint is useful for modelling the skewness parameter of SineSkewed distribution in #1055.

### DIFF
--- a/numpyro/distributions/transforms.py
+++ b/numpyro/distributions/transforms.py
@@ -33,9 +33,9 @@ __all__ = [
     "CorrCholeskyTransform",
     "CorrMatrixCholeskyTransform",
     "ExpTransform",
-    "SoftplusTransform",
     "IdentityTransform",
     "InvCholeskyTransform",
+    "L1BallTransform",
     "LowerCholeskyTransform",
     "LowerCholeskyAffine",
     "PermuteTransform",
@@ -404,8 +404,7 @@ class CorrCholeskyTransform(Transform):
             matrix_to_tril_vec(z1m_cumprod_shifted, diagonal=-1)
         )
         # inverse of tanh
-        x = jnp.log((1 + t) / (1 - t)) / 2
-        return x
+        return jnp.arctanh(t)
 
     def log_abs_det_jacobian(self, x, y, intermediates=None):
         # NB: because domain and codomain are two spaces with different dimensions, determinant of
@@ -581,6 +580,50 @@ class InvCholeskyTransform(Transform):
             return jnp.sum(
                 order * jnp.log(jnp.diagonal(x, axis1=-2, axis2=-1)), axis=-1
             )
+
+
+class L1BallTransform(Transform):
+    r"""
+    Transforms a uncontrained real vector :math:`x` into the unit L1 ball.
+    """
+    domain = constraints.real_vector
+    codomain = constraints.l1_ball
+
+    def __call__(self, x):
+        t = jnp.tanh(x)
+        # make sure that t in (-1, 1)
+        eps = jnp.finfo(t.dtype).eps
+        t = jnp.clip(t, a_min=(-1 + eps), a_max=(1 - eps))
+
+        # apply stick-breaking transform
+        remainder = jnp.cumprod(1 - jnp.abs(t[..., :-1]), axis=-1)
+        pad_width = [(0, 0)] * (t.ndim - 1) + [(1, 0)]
+        remainder = jnp.pad(remainder, pad_width, mode="constant", constant_values=1.0)
+        return t * remainder
+
+    def _inverse(self, y):
+        # inverse stick-breaking
+        remainder = 1 - jnp.cumsum(jnp.abs(y[..., :-1]), axis=-1)
+        pad_width = [(0, 0)] * (y.ndim - 1) + [(1, 0)]
+        remainder = jnp.pad(remainder, pad_width, mode="constant", constant_values=1.0)
+        t = y / remainder
+
+        # inverse of tanh
+        return jnp.arctanh(t)
+
+    def log_abs_det_jacobian(self, x, y, intermediates=None):
+        # compute stick-breaking logdet
+        #   t1 -> t1
+        #   t2 -> t2 * (1 - abs(t1))
+        #   t3 -> t3 * (1 - abs(t1)) * (1 - abs(t2))
+        # hence jacobian is triangular and logdet is the sum of the log
+        # of the diagonal part of the jacobian
+        one_minus_remainder = jnp.cumsum(jnp.abs(y[..., :-1]), axis=-1)
+        # log(remainder) = log1p(1 - remainder)
+        stick_breaking_logdet = jnp.sum(jnp.log1p(-one_minus_remainder), axis=-1)
+
+        tanh_logdet = -2 * jnp.sum(x + softplus(-2 * x) - jnp.log(2.0), axis=-1)
+        return stick_breaking_logdet + tanh_logdet
 
 
 class LowerCholeskyAffine(Transform):
@@ -979,6 +1022,11 @@ def _transform_to_interval(constraint):
             ),
         ]
     )
+
+
+@biject_to.register(constraints.l1_ball)
+def _transform_to_l1_ball(constraint):
+    return L1BallTransform()
 
 
 @biject_to.register(constraints.lower_cholesky)
--- a/numpyro/distributions/transforms.py
+++ b/numpyro/distributions/transforms.py
@@ -590,10 +590,8 @@ class L1BallTransform(Transform):
     codomain = constraints.l1_ball
 
     def __call__(self, x):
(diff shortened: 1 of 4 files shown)


numpyro/distributions/transforms.py
```diff
@@ -583,6 +582,50 @@ def log_abs_det_jacobian(self, x, y, intermediates=None):
...
+        eps = jnp.finfo(t.dtype).eps
+        t = jnp.clip(t, a_min=(-1 + eps), a_max=(1 - eps))
+
+        # apply stick-breaking transform
+        remainder = jnp.cumprod(1 - jnp.abs(t[..., :-1]), axis=-1)
+        pad_width = [(0, 0)] * (t.ndim - 1) + [(1, 0)]
+        remainder = jnp.pad(remainder, pad_width, mode="constant", constant_values=1.0)
+        return t * remainder
+
+    def _inverse(self, y):
```
OlaRonning, line 604: If there is a $|t_i| = 1$, the inverse will become `NaN` for all positions after i. With clipping, it becomes `inf`. 

``` python
# make sure that t in (-1, 1)
t = jnp.array([-.8, .9, -1., -.5])  # jnp.tanh(-10.) will produce -1.

eps = jnp.finfo(t.dtype).eps
t = jnp.clip(t, a_min=(-1 + eps), a_max=(1 - eps))

# apply stick-breaking transform
remainder = jnp.cumprod(1 - jnp.abs(t[..., :-1]), axis=-1)

pad_width = [(0, 0)] * (t.ndim - 1) + [(1, 0)]
remainder = jnp.pad(remainder, pad_width, mode="constant", constant_values=1.0)
y = t * remainder
remainder = 1 - jnp.cumsum(jnp.abs(y[..., :-1]), axis=-1)

pad_width = [(0, 0)] * (y.ndim - 1) + [(1, 0)]
remainder = jnp.pad(remainder, pad_width, mode="constant", constant_values=1.0)
t = y / remainder

# inverse
print(t)  # [-0.8, 0.8999999, -1.0000011,  -inf]
```
If you clip the `x` param to `[-6,6]` in the forward transformation (`__call__`),`tanh(x).abs() <= 0.9999876`  so it will not  produce zeros in the `remainder`.  Something like
``` python
 def __call__(self, x):
        t = jnp.tanh(jnp.clip(x,a_min=-6., a_max=6.))
        ...
```

Let me know what you think.

OlaRonning (commented):
@fehiepsi, thanks for implementing this transformation! I checked the math, and there is a corner case I think we should discuss. See the comment above. Otherwise, it looks good!

numpyro/distributions/transforms.py
```diff
@@ -583,6 +582,50 @@ def log_abs_det_jacobian(self, x, y, intermediates=None):
...
+    Transforms a uncontrained real vector :math:`x` into the unit L1 ball.
+    """
+    domain = constraints.real_vector
+    codomain = constraints.l1_ball
+
+    def __call__(self, x):
+        t = jnp.tanh(x)
+        # make sure that t in (-1, 1)
+        eps = jnp.finfo(t.dtype).eps
+        t = jnp.clip(t, a_min=(-1 + eps), a_max=(1 - eps))
```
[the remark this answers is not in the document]
fehiepsi: The tolerance there is not relevant. I put it there (following the pattern in sphere/simplex constraints) to guard some edge cases like `0.5 + 0.5 > 1.` due to computer precision.

numpyro/distributions/transforms.py
fehiepsi: Thanks, I think that during accumulation `1 - jnp.cumsum(jnp.abs(y[..., :-1]), axis=-1)`, the remainder can be `0`. I will add a clip for it.

numpyro/distributions/transforms.py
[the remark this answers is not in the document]
fehiepsi: I have removed this clip to avoid confusion.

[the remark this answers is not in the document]
OlaRonning: I see, thanks for clarifying.

numpyro/distributions/transforms.py
```diff
@@ -583,6 +582,53 @@ def log_abs_det_jacobian(self, x, y, intermediates=None):
...
 
 
+class L1BallTransform(Transform):
+    r"""
+    Transforms a uncontrained real vector :math:`x` into the unit L1 ball.
+    """
+    domain = constraints.real_vector
+    codomain = constraints.l1_ball
+
+    def __call__(self, x):
```
fritzo, line 592: Does this have nicer properties than the simpler
```py
norm = jnp.sum(jnp.abs(x), -1, True)
return x * ((1 - np.exp(-norm)) / norm)
```

fehiepsi: Nice idea. By transforming the norm from R+ to (0, 1), we get a desired transform. Do you think jacobian can be computed easily? It look a bit non-trivial to me. The stick breaking process behaves like shear transform, so jacobian is easily computed.

fritzo: I believe that since the change of measure depends only on the first derivative (the Jacobian), the Jacobian determinant should be simply
'''py
(1 - np.exp(-norm)) / norm
'''

fehiepsi: yeah, it seems that log Jacobian determinant has [simple formula according to wolfram](https://www.wolframalpha.com/input/?i=log+jacobian+determinant+x+%2F+%7Cx%7C+*+%281+-+e%5E%28-%7Cx%7C%29%29+with+respect+to+x): `-norm`. I'll switch to this which is both faster and more numerical stable.

fehiepsi: I fail to derive a nice formula for jacobian of x / (1 + norm(x)) transform :(

fritzo: What about `x/sqrt(1+norm(x)**2)` which is smoother?

fehiepsi: Do you have some tricks to calculate the determinant of jacobian? I can derive the jacobian matrix for a `x * f(norm(x))` transform. But the formula for determinant in the case `len(x) == 2` is complicated to generalize to e.g. `len(x) == 3`.

fritzo: How about this argument:
Let `t(x) = x * f(norm1(x))` be an angle-preserving transformation. Then the Jacobian in polar coordinates is an `f`-scaled identity matrix for directions orthogonal to `x` and in the direction tangent to `x` is
```
d/dradius norm2(x * f(norm1(x)))
  = f(norm1(x)) + norm2(x) * f'(norm1(x)) * d/dradius norm1(x)
  = f(norm1(x)) + norm1(x) * f'(norm1(x))
```
where I have used `d/dradius norm1(x) = norm1(x)/norm2(x)`. Thus the Jacobian determinant is
```
det_J = f(norm1(x)) ** x.shape[-1] * (1 + norm1(x) * f'(norm1(x)) * f(norm1(x)))
```
For example in the case `t(x) = x/(1+norm1(x))` the log determinant is
```
log_abs_det_jacobian
  = -log1p(norm1(x)) * x.shape[-1] + log1p(-norm1(x) / (1 + norm1(x)) ** 3))
```
(I probably have some math errors above, but I believe the orthogonal + tangential d/dradius argument will work)

fehiepsi: Thanks. I have been thinking about this but failed to follow the math and couldn't make the tests pass. :( Though it seems to be slower (if `x.shape[-1]` is large), the stick-breaking process is the same as what Ola constructed in [his example](https://github.com/pyro-ppl/numpyro/pull/1055/files#diff-c9303fefc2543090c64d22f3114e87b5a612220eb9e8b7431ba9836355cafff5R49). On a 2D plane, I guess to generate a pair (a, b) with |a| + |b| < 1, it is easier to use stick-breaking process than to use norm transform. So for now, I am in favor of keeping the transform as-is.

OlaRonning (approved):
@fehiepsi, thanks for adding clipping. With clipping `t=[-.8, .9, -1., -.5])` no longer produces `NaN`'s for `trans.inv o trans`, but everything subsequent the -1 will be zero. I'll find some time to work through @fritzo suggestion once I'm back at work on Monday, but think the current transform adequetly resolves the motivating issue from [my example](https://github.com/pyro-ppl/numpyro/pull/1055/files#diff-c9303fefc2543090c64d22f3114e87b5a612220eb9e8b7431ba9836355cafff5R49).

fehiepsi:
I think that is fine. You can use float64 if you want to work with extreme values (like in ExpTransform). Please let me know if you get some working version of norm transforms and want to make a switch. We then need to choose which norm transform is good for your distribution. Personally, I like your stick breaking approach.

fehiepsi:
I'll go ahead and merge this. We can revisit other transforms in the future, if needed.

PR #1093, opened 2021-07-11
Bump version to 0.7.1

### LINKED ISSUE #1092: import numpyro error (version 0.7)
Hello,

I installed the most recent version of numpyro (GPU) and am getting the following error when importing:

ImportError: cannot import name 'get_dtype' from 'numpyro.distributions.util'

fehiepsi:
Resolves #1092

For some reason, when creating the wheel file with `python setup.py sdist bdist_wheel` in the last release, `.whl` file has content different from `.tar.gz` file (some scripts in `.whl` are from one of the old versions). I'm not sure what happens. :( I'll check two files for consistency this time.

List of old scripts in `numpyro-0.7.0-py3-none-any.whl` file:
+ __init__.py
+ contrib/funsor/__init__.py
+ contrib/funsor/infer_util.py
+ distributions/__init__.py
+ distributions/directional.py
+ distributions/discrete.py

I think I haven't deleted the old `build` folder and got the same issue as this https://stackoverflow.com/questions/66017277/bdist-using-old-files-to-create-built-distribution

### DIFF
--- a/examples/annotation.py
+++ b/examples/annotation.py
@@ -327,7 +327,7 @@ def main(args):
 
 
 if __name__ == "__main__":
-    assert numpyro.__version__.startswith("0.7.0")
+    assert numpyro.__version__.startswith("0.7.1")
     parser = argparse.ArgumentParser(description="Bayesian Models of Annotation")
     parser.add_argument("-n", "--num-samples", nargs="?", default=1000, type=int)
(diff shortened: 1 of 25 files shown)


martinjankowiak (approved):

fehiepsi:
Thanks, @martinjankowiak !

PR #1099, opened 2021-07-14
add fix and test for model rendering of only discrete variables

bdatko:
Found an edge case within model rendering where empty inputs were being passed to [jax.jacobian](https://github.com/pyro-ppl/numpyro/blob/cf85229edab08144108623338f31c78a1a16d5d5/numpyro/contrib/render.py#L120), see [discussion](https://forum.pyro.ai/t/numpyro-render-model-valueerror-too-few-leaves-for-pytreedef/3113/2?u=bdatko) for more details.

Added fix for `render.py` along with a simple model for testing

### DIFF
--- a/numpyro/contrib/render.py
+++ b/numpyro/contrib/render.py
@@ -117,7 +117,10 @@ def get_model_relations(model, model_args=None, model_kwargs=None, num_tries=10)
         and not site["is_observed"]
         and not site["fn"].is_discrete
     }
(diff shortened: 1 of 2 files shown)


fehiepsi (approved):
Awesome! Also thanks for adding the small and clean test, @bdatko!

bdatko:
You are very welcome =) Thanks, @fehiepsi. First every PR 😲

PR #1101, opened 2021-07-17
Relax matching device pattern

fehiepsi:
I think this is causing some issues in [this forum thread](https://forum.pyro.ai/t/sampling-for-numpyro-breaks-running-in-parallel-show-progress-mode/3140/2). The pattern there assumes the device has the form `..._1` but in my system in recent jax release, it is `cpu:1`.

I have tested the gp examples locally. Chain MCMC tests also pass on CI.


OlaRonning (approved):
LGTM

fehiepsi:
Thanks for reviewing, @OlaRonning!

PR #1109, opened 2021-07-28
Support direct use of TFP distributions in `numpyro.sample`

### LINKED ISSUE #1106: Using TFP distributions directly in `sample` primitive
Quoted from [the forum](https://forum.pyro.ai/t/sample-from-the-mixture-same-family-distribution/3178/6): The current wrappers in `numpyro.contrib.tfp` seems to not work for special distributions like MixtureSameFamily. For now, to make it work, we'll need to do
```python
from numpyro.contrib.tfp.distributions import TFPDistribution
import tensorflow_probability...distributions as tfd

def model():
    x = sample("x", TFPDistribution(tfd.MixtureSampleFamily(...)))
```
In the long term, I think it would be nice to use TFP distributions directly in the `sample` primitive:
```python
import tensorflow_probability...distributions as tfd

def model():
    x = sample("x", tfd.MixtureSampleFamily(...))
```

I believe it is enough to modify [`sample` primitive](https://github.com/pyro-ppl/numpyro/blob/e55f0d41c9eba48a10e88fb403a5e016f18857e6/numpyro/primitives.py#L104) such that if `fn` is not a NumPyro distribution, we will check if `tensorflow_probability` is available, then trying to check if `fn` is a TFP distribution. If yes, then we replace `fn` by `numpyro.contrib.tfp.distributions.TFPDistribution[type(fn)](**fn.parameters)` or by
```python
new_fn = TFPDistribution()
new_fn.tfp_dist = fn
new_fn.tfd_class = type(fn)  # and extra repr if needed
```

tcbegley:
This PR automatically wraps instances of `tfd.Distribution` inside `numpyro.sample` so that distributions can be imported directly from `tfd` rather than `numpyro.contrib.tfp.distributions`.

Added a couple of tests, but let me know if you think something more extensive is needed.

Fixes #1106

### DIFF
--- a/numpyro/contrib/tfp/distributions.py
+++ b/numpyro/contrib/tfp/distributions.py
@@ -245,7 +245,7 @@ class TFPDistribution(NumPyroDistribution, metaclass=_TFPDistributionMeta):
         fn = jax.tree_util.tree_unflatten(aux_data, params)
         with warnings.catch_warnings():
             warnings.simplefilter("ignore", category=FutureWarning)
-            return cls[fn.__class__](**fn.parameters)
+            return TFPDistribution[fn.__class__](**fn.parameters)
 
 
 __all__ = ["BijectorConstraint", "BijectorTransform", "TFPDistribution"]
--- a/numpyro/contrib/tfp/distributions.py
+++ b/numpyro/contrib/tfp/distributions.py
@@ -7,6 +7,7 @@ import warnings
 
 import numpy as np
 
+import jax
 import jax.numpy as jnp
 from tensorflow_probability.substrates.jax import bijectors as tfb, distributions as tfd
 
@@ -241,6 +242,16 @@ class TFPDistribution(NumPyroDistribution, metaclass=_TFPDistributionMeta):
     def is_discrete(self):
         # XXX: this should cover most cases
         return self.support is None
+
+    def tree_flatten(self):
+        return jax.tree_util.tree_flatten(self.tfp_dist)
+
+    @classmethod
+    def tree_unflatten(cls, aux_data, params):
+        fn = jax.tree_util.tree_unflatten(aux_data, params)
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore", category=FutureWarning)
+            return cls[fn.__class__](**fn.parameters)
 
 
 __all__ = ["BijectorConstraint", "BijectorTransform", "TFPDistribution"]
--- a/numpyro/contrib/tfp/distributions.py
+++ b/numpyro/contrib/tfp/distributions.py
@@ -136,7 +136,6 @@ class _TFPDistributionMeta(type(NumPyroDistribution)):
         init.__signature__ = inspect.signature(tfd_class.__init__)
 
         _PyroDist = type(tfd_class_name, (TFPDistribution,), {})
-        _PyroDist.tfd_class = tfd_class
         _PyroDist.__init__ = init
 
         if tfd_class is tfd.InverseGamma:
@@ -187,9 +186,6 @@ class TFPDistribution(NumPyroDistribution, metaclass=_TFPDistributionMeta):
         d = TFPDistribution[tfd.Normal](0, 1)
 
     """
-
-    tfd_class = None
-
     def __getattr__(self, name):
         # return parameters from the constructor
         if name in self.tfp_dist.parameters:
--- a/numpyro/contrib/tfp/distributions.py
+++ b/numpyro/contrib/tfp/distributions.py
@@ -154,12 +154,9 @@ class _TFPDistributionMeta(type(NumPyroDistribution)):
             if hasattr(numpyro_dist, tfd_class_name):
                 numpyro_dist_class = getattr(numpyro_dist, tfd_class_name)
                 # resolve FooProbs/FooLogits namespaces
-                if type(numpyro_dist_class).__name__ == "function":
-                    if not hasattr(numpyro_dist, f"{tfd_class_name}Logits"):
-                        raise ValueError
-                    numpyro_dist_class = getattr(
-                        numpyro_dist, f"{tfd_class_name}Logits"
-                    )
+                numpyro_dist_class = getattr(
+                    numpyro_dist, f"{tfd_class_name}Logits", numpyro_dist_class
+                )
                 _PyroDist.arg_constraints = numpyro_dist_class.arg_constraints
                 _PyroDist.has_enumerate_support = (
                     numpyro_dist_class.has_enumerate_support
@@ -247,10 +244,7 @@ for _name, _Dist in tfd.__dict__.items():
     if _Dist is tfd.Distribution:
         continue
 
-    try:
-        _PyroDist = TFPDistribution[_Dist]
-    except ValueError:
-        continue
+    _PyroDist = TFPDistribution[_Dist]
     _PyroDist.__module__ = __name__
     locals()[_name] = _PyroDist
 
--- a/numpyro/contrib/tfp/distributions.py
+++ b/numpyro/contrib/tfp/distributions.py
@@ -2,7 +2,6 @@
 # SPDX-License-Identifier: Apache-2.0
 
 from functools import lru_cache
-import inspect
 import warnings
 
 import numpy as np
@@ -133,8 +132,6 @@ class _TFPDistributionMeta(type(NumPyroDistribution)):
             )
             self.tfp_dist = tfd_class(*args, **kwargs)
 
-        init.__signature__ = inspect.signature(tfd_class.__init__)
-
         _PyroDist = type(tfd_class_name, (TFPDistribution,), {})
         _PyroDist.__init__ = init
 
@@ -186,6 +183,7 @@ class TFPDistribution(NumPyroDistribution, metaclass=_TFPDistributionMeta):
         d = TFPDistribution[tfd.Normal](0, 1)
 
     """
+
     def __getattr__(self, name):
         # return parameters from the constructor
         if name in self.tfp_dist.parameters:
--- a/numpyro/contrib/tfp/distributions.py
+++ b/numpyro/contrib/tfp/distributions.py
@@ -282,6 +282,6 @@ __doc__ = "\n\n".join(
     """.format(
             _name
         )
-        for _name in __all__[:_len_all] + sorted(__all__[_len_all:])
+        for _name in __all__[:_len_all]
     ]
 )
--- a/numpyro/contrib/tfp/distributions.py
+++ b/numpyro/contrib/tfp/distributions.py
@@ -1,6 +1,7 @@
 # Copyright Contributors to the Pyro project.
 # SPDX-License-Identifier: Apache-2.0
 
+from functools import lru_cache
 import inspect
 import warnings
 
@@ -116,6 +117,7 @@ def _onehot_enumerate_support(self, expand=True):
 
 
 class _TFPDistributionMeta(type(NumPyroDistribution)):
+    @lru_cache(maxsize=None)
     def __getitem__(cls, tfd_class):
         assert issubclass(tfd_class, tfd.Distribution)
 
--- a/numpyro/contrib/tfp/distributions.py
+++ b/numpyro/contrib/tfp/distributions.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import inspect
+import warnings
 
 import numpy as np
 
@@ -121,6 +122,12 @@ class _TFPDistributionMeta(type(NumPyroDistribution)):
         tfd_class_name = tfd_class.__name__
 
         def init(self, *args, **kwargs):
+            warnings.warn(
+                "Importing distributions from numpyro.contrib.tfp.distributions is "
+                "deprecated. You should import distributions directly from "
+                "tensorflow_probability.substrates.jax.distributions instead.",
+                FutureWarning,
+            )
             self.tfp_dist = tfd_class(*args, **kwargs)
 
         init.__signature__ = inspect.signature(tfd_class.__init__)
--- a/numpyro/contrib/tfp/distributions.py
+++ b/numpyro/contrib/tfp/distributions.py
@@ -105,18 +105,67 @@ def _transform_to_bijector_constraint(constraint):
     return BijectorTransform(constraint.bijector)
 
 
+def _onehot_enumerate_support(self, expand=True):
+    n = self.event_shape[-1]
+    values = jnp.identity(n, dtype=jnp.result_type(self.dtype))
+    values = values.reshape((n,) + (1,) * len(self.batch_shape) + (n,))
+    if expand:
+        values = jnp.broadcast_to(values, (n,) + self.batch_shape + (n,))
+    return values
+
+
 class _TFPDistributionMeta(type(NumPyroDistribution)):
     def __getitem__(cls, tfd_class):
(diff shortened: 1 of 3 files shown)


numpyro/primitives.py
```diff
@@ -131,6 +131,29 @@ def sample(
...
+            "tensorflow_probability, make sure it is correctly installed."
+        )
+        try:
+            from tensorflow_probability.substrates.jax import distributions as tfd
+
+            from numpyro.contrib.tfp.distributions import TFPDistribution
+        except ImportError:
+            # if tensorflow_probability fails to import, then fn can't have been a tf
+            # distribution in the first place, so returning TypeError is ok
+            raise type_error
```
tcbegley, line 148: Are you happy with this reasoning? Error message includes a note to make sure tfp is installed correctly if the user is trying to use it but is still getting an `ImportError`.

numpyro/primitives.py
```diff
@@ -131,6 +131,29 @@ def sample(
         + "_unobserved"`` which should be used by guides.
     :return: sample from the stochastic `fn`.
     """
+    if not isinstance(fn, numpyro.distributions.Distribution):
+        type_error = TypeError(
+            "It looks like you tried to use a fn that isn't an instance of "
+            "numpyro.distributions.Distribution, or "
+            "tensorflow_probability.distributions.Distribution. If you're using "
+            "tensorflow_probability, make sure it is correctly installed."
+        )
```
tcbegley, line 140: Looks like I also need to account for some `Funsor` stuff, which I think is why the tests are failing. Will take a closer look later.

fehiepsi:
@tcbegley It would be nice to raise DeprecationError when users import distributions from `numpyro.contrib.tfp.distributions`. :D Could you also move [this logic](https://github.com/pyro-ppl/numpyro/blob/e55f0d41c9eba48a10e88fb403a5e016f18857e6/numpyro/contrib/tfp/distributions.py#L242-L251) and [this logic](https://github.com/pyro-ppl/numpyro/blob/e55f0d41c9eba48a10e88fb403a5e016f18857e6/numpyro/contrib/tfp/distributions.py#L192-L224) (using `if `tfd_class` is `tfd.InverseGamma: add_attributes` ...) to `_TFPDistributionMeta.__getitem__`?

numpyro/primitives.py
fehiepsi: I like this pattern. :+1:

numpyro/primitives.py
fehiepsi: Yeah, maybe we can try to import `funsor` first, if success then checks if it is a funsor distribution. Otherwise, following the below logic.

numpyro/contrib/tfp/distributions.py
```diff
@@ -235,21 +248,11 @@ def _onehot_enumerate_support(self, expand=True):
...
-            # resolve FooProbs/FooLogits namespaces
-            if type(numpyro_dist_class).__name__ == "function":
-                if not hasattr(numpyro_dist, _name + "Logits"):
-                    continue
-                numpyro_dist_class = getattr(numpyro_dist, _name + "Logits")
-            _PyroDist.arg_constraints = numpyro_dist_class.arg_constraints
-            _PyroDist.has_enumerate_support = numpyro_dist_class.has_enumerate_support
-            _PyroDist.enumerate_support = numpyro_dist_class.enumerate_support
-        locals()[_name] = _PyroDist
+    except ValueError:
```
tcbegley, line 252: Can this ever actually be hit? I included it to replace the `continue` statement in the original code, but I think the preceding if statements mean it's unnecessary?

tcbegley:
> Could you also move this logic and this logic

I did this in `f02b8fa`, let me know if that's what you had in mind.

>  It would be nice to raise DeprecationError when users import distributions from `numpyro.contrib.tfp.distributions`

I had a think about this. Are there legitimate reasons why someone would want to still import `BijectorConstraint`, `BijectorTransform`, or `TFPDistribution` or should they also be deprecated? I think you can raise a deprecation warning when the module as a whole is imported (and catch it inside `numpyro.sample`), but I'm not sure if it's possible to raise the warning only when certain classes are imported? Maybe I'm misunderstanding?

numpyro/contrib/tfp/distributions.py
fehiepsi: Because you moved all the dispatching logic to the `__getitem__`, this is not required. Please feel free to remove this.

test/contrib/test_tfp.py
```diff
@@ -251,3 +251,45 @@ def test_sample_tfp_distributions():
...
+    "dist,args",
+    [
+        ["Bernoulli", (0,)],
+        ["Beta", (1, 1)],
+        ["Binomial", (10, 0)],
+        ["Categorical", ([0, 1, -1],)],
+        ["Cauchy", (0, 1)],
+        ["Dirichlet", ([1, 2, 0.5],)],
+        ["Exponential", (1,)],
+        ["Normal", (0, 1)],
```
fehiepsi, line 268: Could you add this [one](https://forum.pyro.ai/t/sample-from-the-mixture-same-family-distribution/3178/3?u=fehiepsi) to exercise that mixture distribution?

fehiepsi:
> raise the warning only when certain classes are imported

I think we can raise the warning in the `__init__` method. For those distributions in `contrib.tfp.distributions`, you can overwrite the init method:

```python
_PyroDist = TFPDistribution[_Dist]

def init(self, *args, **kwargs):
    warnings.warn("...", FutureWarning)
    self.tfp_dist = _Dist(*args, **kwargs)

init.__signature__ = inspect.signature(_Dist.__init__)
_PyroDist.__init__ = init
```

test/contrib/test_tfp.py
tcbegley: Had to make a separate test for this as I can't use `tfd` arguments in the `parametrize`, but this is now added.

test/contrib/test_tfp.py
```diff
@@ -42,6 +43,7 @@ def f(x):
 
 
 @pytest.mark.filterwarnings("ignore:can't resolve package")
+@pytest.mark.filterwarnings("ignore:Importing distributions")
```
fehiepsi, line 46: nit: this test can be removed I think

test/contrib/test_tfp.py
```diff
@@ -53,6 +55,7 @@ def test_independent():
...
```
fehiepsi, line 58: This test can be removed I think

tcbegley: Updated the test to check the pattern you describe.

test/contrib/test_tfp.py
```diff
@@ -32,6 +32,7 @@ def test_api_consistent():
                 assert p in inspect.getfullargspec(tfp_dist.__init__)[0]
 
 
+@pytest.mark.filterwarnings("ignore:Importing distributions")
```
fehiepsi, line 35: This one is interesting. For completeness, it would be nice if you could add methods `tree_flatten`, `tree_unflatten` to `TFPDistribution`. Some sort of
```python
def tree_flatten(self):
    return jax.tree_util.tree_flatten(self.tfp_dist)

@classmethod
    fn = jax.tree_util.tree_unflatten(aux_data, params)
    return TFPDistribution[fn.__class__](**fn.parameters)
```

fehiepsi: While looking at this, I am a bit worried that we will keep registering the [pytree](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/distributions/distribution.py#L140) each time we have a `sample` statement. This might populate JAX pytree caches. I'll look at this in more detail later.

tcbegley: I'm not sure how to check pytree caches, but I can confirm that the call to `register_pytree_node` is currently happening with every `numpyro.sample` statement.

Perhaps the wrapped distributions should be constructed in the module scope as before, and we can instantiate the correct one inside `numpyro.sample` using `getattr` on the model?

fehiepsi: It can be fixed by creating a cache `_TFD_CACHE = {}` in `contrib.tfp.distributions` and in the `__getitem__` method, return the cache if the tfd class is available there, otherwise caching the wrapped class at the end of `__getitem__`.

tcbegley: I think we can just use `functools.lru_cache` for this with `maxsize=None`, what do you think? Have pushed a commit with this change. Tests are passing locally and I've verified that each distribution is only registered once.

tcbegley: I tried adding the tree_flatten methods, running your snippet I get `TypeError: Value 'x' with type <class 'str'> is not a valid JAX type`. Same problem if I replace `tfd.Normal` with `numpyro.distributions.Normal`. 

Did you mean to trace inside the jitted function or jit _then_ trace perhaps?

fehiepsi: I guess the trace holds some non-device types such as `trace["type"] == "samples"`. You might just return `tr["fn"]` then :).

fehiepsi: > we can just use functools.lru_cache for this with maxsize=None

Nice solution! I didn't know this Python helper. Thanks!

numpyro/primitives.py
```diff
@@ -131,6 +131,44 @@ def sample(
...
+                # numpyro Distribution or a Funsor, and it can't have been a tfp
+                # distribution either, so raising TypeError is ok
+                raise type_error
+
+            if isinstance(fn, tfd.Distribution):
+                with warnings.catch_warnings():
+                    # ignore FutureWarnings when instantiating TFPDistribution
+                    warnings.simplefilter("ignore", category=FutureWarning)
+                    # if fn is a tfp distribution we need to wrap it
+                    fn = TFPDistribution[fn.__class__](**fn.parameters)
```
fehiepsi, line 166: I think we can simplify this into
```python
fn = TFPDistribution[fn.__class__](lambda: fn)
```
so we can avoid reinitializing the tfp instance

numpyro/contrib/tfp/distributions.py
```diff
@@ -105,18 +107,71 @@ def _transform_to_bijector_constraint(constraint):
...
+                "tensorflow_probability.substrates.jax.distributions instead.",
+                FutureWarning,
+            )
             self.tfp_dist = tfd_class(*args, **kwargs)
 
         init.__signature__ = inspect.signature(tfd_class.__init__)
 
-        _PyroDist = type(tfd_class.__name__, (TFPDistribution,), {})
+        _PyroDist = type(tfd_class_name, (TFPDistribution,), {})
         _PyroDist.tfd_class = tfd_class
```
fehiepsi, line 138: I think `tfd_class` is not used elsewhere, could you help me remove this and the one in TFPDistribution? Thanks!

numpyro/contrib/tfp/distributions.py
```diff
@@ -105,18 +107,71 @@ def _transform_to_bijector_constraint(constraint):
...
         def init(self, *args, **kwargs):
+            warnings.warn(
+                "Importing distributions from numpyro.contrib.tfp.distributions is "
+                "deprecated. You should import distributions directly from "
+                "tensorflow_probability.substrates.jax.distributions instead.",
+                FutureWarning,
+            )
             self.tfp_dist = tfd_class(*args, **kwargs)
 
         init.__signature__ = inspect.signature(tfd_class.__init__)
```
fehiepsi, line 135: We don't need this anymore I think. Users should use TFP distributions directly.

numpyro/primitives.py
fehiepsi, line 166: I think you can simplify this into
```
old_fn = fn
fn = TFPDistribution[fn.__class__](lambda: old_fn)

File: numpyro/contrib/tfp/distributions.py (hunk 10 above)
Comment (tcbegley) [reply numpyro/contrib/tfp/distributions.py:135]: Done! I removed `test_api_consistent` since that's no longer relevant. Is that correct?

File: numpyro/contrib/tfp/distributions.py [hunk 11]
```diff
@@ -187,41 +237,15 @@ def is_discrete(self):
...
-Pareto.arg_constraints = {
-    "concentration": constraints.positive,
-    "scale": constraints.positive,
-}
+    @classmethod
+    def tree_unflatten(cls, aux_data, params):
+        fn = jax.tree_util.tree_unflatten(aux_data, params)
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore", category=FutureWarning)
+            return cls[fn.__class__](**fn.parameters)
```
fehiepsi, line 248: nit: to trigger the cache, we need to use `TFPDistribution` instead of `cls` (otherwise, a new cache will be stored).

numpyro/contrib/tfp/distributions.py
```diff
@@ -234,22 +258,9 @@ def _onehot_enumerate_support(self, expand=True):
...
-                if not hasattr(numpyro_dist, _name + "Logits"):
-                    continue
-                numpyro_dist_class = getattr(numpyro_dist, _name + "Logits")
-            _PyroDist.arg_constraints = numpyro_dist_class.arg_constraints
-            _PyroDist.has_enumerate_support = numpyro_dist_class.has_enumerate_support
-            _PyroDist.enumerate_support = numpyro_dist_class.enumerate_support
-        locals()[_name] = _PyroDist
+    _PyroDist = TFPDistribution[_Dist]
+    _PyroDist.__module__ = __name__
+    locals()[_name] = _PyroDist
```
fehiepsi, line 263: nit: could you also help me remove the docstring below
```
for name in __all__[:_len_all]   # this is not required: + sorted(__all__[_len_all:])
```

fehiepsi (dismissed):
Great improvement, @tcbegley! Thanks!

numpyro/contrib/tfp/distributions.py
tcbegley: As in, keep the docstring, but don't document any of the generated wrapped distributions? Believe I have done this but do check to make sure I understood correctly.

test/contrib/test_tfp.py
```diff
@@ -17,71 +16,50 @@
...
+    @jit
+    def f(x):
+        with numpyro.handlers.trace() as tr:
+            numpyro.sample("x", tfd.Normal(x, 1))
+        return tr["x"]["fn"]
 
-@pytest.mark.filterwarnings("ignore:can't resolve package")
-def test_independent():
-    from numpyro.contrib.tfp import distributions as tfd
+    with numpyro.handlers.seed(rng_seed=0):
```
fehiepsi, line 30: nit: I think it is a better practice to move seed handler inside the `@jit`. Otherwise, if you run
```
with numpyro.handlers.seed(rng_seed=1):
    res = f(0.0)
```
you will get the same result as `rng_seed=0`.

test/contrib/test_tfp.py
```diff
@@ -240,6 +220,7 @@ def potential_fn(z):
 # test if sampling from tfp distributions works as expected using
 # numpyro sample function: numpyro.sample("name", dist) (bug)
 @pytest.mark.filterwarnings("ignore:can't resolve package")
+@pytest.mark.filterwarnings("ignore:Importing distributions")
```
fehiepsi, line 223: I guess we need to test
```
from tensorflow_probability.substrates.jax import distributions as tfd
from numpyro.contrib.tfp.distributions import TFPDistribution

d = TFPDistribution[tfd.Normal](0, 1)
```

fehiepsi (commented):
Sorry, a couple of nits

fehiepsi (approved):

PR #1113, opened 2021-08-01
Add sphinx-search to docs and pin version numbers

### LINKED ISSUE #1111: [Docs] Syntax highlighting broken for notebook examples / tutorials
Syntax highlighting doesn't seem to work on any notebook based examples tutorials in the deployed documentation. See e.g. http://num.pyro.ai/en/stable/tutorials/bayesian_regression.html

Interestingly I don't have the same problems if I build the documentation locally (following steps in a fresh virtual environment)

```
pip install .[doc]
pip install -r docs/requirements.txt
make docs
python -m http.server --directory docs/build/html
```

I can't see from the repo how the docs are built / deployed, so I suspect there's something in that process that disagrees with the above steps? In particular I think the `ipython` dependency is key, is that maybe not getting picked up in the build of the deployed docs?

tcbegley:
This PR pins some of the dependencies in `docs/requirements.txt`. The goal is to add support for `readthedocs-sphinx-search`, and also fix syntax highlighting issues in examples / tutorials rendered from notebooks.

Docs build and look ok for me in a fresh virtual environment installing only from `docs/requirements.txt` (I think this is what happens in the build on readthedocs itself?). I added `ipython` after initially getting an error `Pygments lexer name 'ipython3' is not known` when it was not included as a dependency.

Note that the search feature apparently only works when the documentation is actually hosted on readthedocs itself, so I haven't been able to verify that it's configured correctly locally. (The following is logged to the browser console).

```sh
[INFO] Docs are not being served on Read the Docs, readthedocs-sphinx-search will not work.
```

Fixes #1111

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,11 +1,15 @@
 dm-haiku
 flax
 funsor>=0.4.1
+ipython
 jax>=0.2.11
 jaxlib>=0.1.62
 jaxns>=0.0.7
 optax>=0.0.6
-nbsphinx>=0.8.5
+nbsphinx==0.8.6
+readthedocs-sphinx-search==0.1.0
+sphinx==4.1.2
 sphinx-gallery
+sphinx_rtd_theme==0.5.2
 tensorflow_probability>=0.13
 tqdm
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -73,6 +73,7 @@ extensions = [
     "sphinx.ext.mathjax",
     "sphinx.ext.viewcode",
     "sphinx_gallery.gen_gallery",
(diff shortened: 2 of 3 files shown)


fehiepsi (approved):

fehiepsi:
Thanks a lot, @tcbegley! Let's see how things look like when merged

fehiepsi:
@tcbegley Could you pin `sphinx` to `4.0.3` instead? The build gets [this issue](https://github.com/spatialaudio/nbsphinx/issues/584).

tcbegley:
> Could you pin sphinx to 4.0.3 instead?

Sure thing, have opened #1114 for this.

PR #1114, opened 2021-08-01
Pin Sphinx for readthedocs

tcbegley:
Follow-up to #1113, pinning older version of sphinx to avoid [this issue](https://github.com/spatialaudio/nbsphinx/issues/584)


fehiepsi (approved):

fehiepsi:
@tcbegley Looks like the latest version is working well. Thanks!

Issue #1104, opened 2021-07-19
[FR] add format_shapes method to trace


bdatko:
Having `trace.format_shapes()` as in [Pyro](https://docs.pyro.ai/en/stable/poutine.html?highlight=format_shapes#pyro.poutine.Trace.format_shapes) would help with debugging and verifying models, especially with enumeration, as shown in the [Pyro tutorial](https://pyro.ai/examples/tensor_shapes.html#Broadcasting-to-allow-parallel-enumeration).

tcbegley:
@fehiepsi unlike Pyro the trace object in NumPyro is an `OrderedDict` rather than a custom class, so seems to me there's a choice to be made here:

* Create a utility function that takes a trace as input and returns the table of shapes
* Create a custom trace object (probably subclassing `OrderedDict` in the first instance to minimise changes?) and define a `format_shapes` method.

Do you have a preference out of these two, or even a preferred third option?

fritzo:
👍for a utility function.

fehiepsi:
Yes, a utility function would be nice. A third option is to have a handler that adds `print` effect to each sample site. :)

PR #1116, opened 2021-08-03
Add format_shapes utility

tcbegley:
This is a proposed solution to #1104. It's a pretty direct port of `.format_shapes` from [Pyro](https://github.com/pyro-ppl/pyro/blob/b7ba915b851d51c61d805da741bf7a74fcf9319d/pyro/poutine/trace_struct.py#L435-L534).

I've created [a gist](https://gist.github.com/tcbegley/70cb8e34a681c8324533b88cc5d99b1d) with examples from the Pyro docs.

Would love some feedback on a few things in particular:
- Function name + location. I just put it in `numpyro.util` for now.
- The main difference between this and Pyro that I'm aware of is that in Pyro plates are traced as `"sample"` sites, which means they show up in the result of `format_shapes`. I can add some logic to add `"plate"` sites from NumPyro traces to the result. Should they be mixed in with the sample sites like in Pyro or should they get their own section?
- Added a `log_prob` argument to mimic the effect of running `trace.compute_log_prob()` before `trace.format_shapes()` in Pyro. Is this a reasonable solution?
- Couldn't see any tests in the Pyro repo for this to replicate. Is it worth writing tests for?

TODO:

- [x] Update docstring

### DIFF
--- a/numpyro/util.py
+++ b/numpyro/util.py
@@ -462,6 +462,12 @@ def format_shapes(trace, *, title="Trace Shapes:", last_site=None, log_prob=Fals
                     + [str(size) for size in batch_shape]
                     + ["|", None]
                 )
+        elif site["type"] == "plate":
+            shape = getattr(site["value"], "shape", ())
+            rows.append(
+                [f"{name} plate", None] + [str(size) for size in shape] + ["|", None]
+            )
+
         if name == last_site:
             break
 
--- a/numpyro/util.py
+++ b/numpyro/util.py
@@ -408,3 +408,108 @@ def soft_vmap(fn, xs, batch_ndims=1, chunk_size=None):
         lambda y: jnp.reshape(y, (-1,) + jnp.shape(y)[map_ndims:])[:batch_size], ys
     )
     return tree_map(lambda y: jnp.reshape(y, batch_shape + jnp.shape(y)[1:]), ys)
+
+
+def format_trace_shapes(trace, title="Trace Shapes:", last_site=None, log_prob=False):
+    """
+    Returns a string showing a table of the shapes of all sites in the
+    trace.
+    """
+    if not trace.keys():
+        return title
+    rows = [[title]]
+
+    rows.append(["Param Sites:"])
+    for name, site in trace.items():
+        if site["type"] == "param":
+            rows.append(
+                [name, None]
+                + [str(size) for size in getattr(site["value"], "shape", ())]
+            )
+        if name == last_site:
+            break
+
+    rows.append(["Sample Sites:"])
+    for name, site in trace.items():
+        if site["type"] == "sample":
+            # param shape
+            batch_shape = getattr(site["fn"], "batch_shape", ())
+            event_shape = getattr(site["fn"], "event_shape", ())
+            rows.append(
+                [f"{name} dist", None]
+                + [str(size) for size in batch_shape]
+                + ["|", None]
+                + [str(size) for size in event_shape]
+            )
+
+            # value shape
+            event_dim = len(event_shape)
+            shape = getattr(site["value"], "shape", ())
+            batch_shape = shape[: len(shape) - event_dim]
+            event_shape = shape[len(shape) - event_dim :]
+            rows.append(
+                ["value", None]
+                + [str(size) for size in batch_shape]
+                + ["|", None]
+                + [str(size) for size in event_shape]
+            )
+
+            # log_prob shape
+            if log_prob:
+                batch_shape = getattr(site["fn"].log_prob(site["value"]), "shape", ())
+                rows.append(
+                    ["log_prob", None]
+                    + [str(size) for size in batch_shape]
+                    + ["|", None]
+                )
+        if name == last_site:
+            break
+
+    return _format_table(rows)
+
+
+def _format_table(rows):
+    """
+    Formats a right justified table using None as column separator.
+    """
+    # compute column widths
+    column_widths = [0, 0, 0]
+    for row in rows:
+        widths = [0, 0, 0]
+        j = 0
+        for cell in row:
+            if cell is None:
+                j += 1
+            else:
+                widths[j] += 1
+        for j in range(3):
+            column_widths[j] = max(column_widths[j], widths[j])
+
+    # justify columns
+    for i, row in enumerate(rows):
+        cols = [[], [], []]
+        j = 0
+        for cell in row:
+            if cell is None:
+                j += 1
+            else:
+                cols[j].append(cell)
+        cols = [
+            [""] * (width - len(col)) + col
+            if direction == "r"
+            else col + [""] * (width - len(col))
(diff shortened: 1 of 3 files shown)


numpyro/util.py
```diff
@@ -408,3 +408,108 @@ def soft_vmap(fn, xs, batch_ndims=1, chunk_size=None):
         lambda y: jnp.reshape(y, (-1,) + jnp.shape(y)[map_ndims:])[:batch_size], ys
     )
     return tree_map(lambda y: jnp.reshape(y, batch_shape + jnp.shape(y)[1:]), ys)
+
+
+def format_trace_shapes(trace, title="Trace Shapes:", last_site=None, log_prob=False):
```
fehiepsi, line 413: nit: I think `format_shapes` is fine here. Could you add `*` after `trace` so that we can have clearer user codes?

fehiepsi:
> Is it worth writing tests for?

I think you can compare the output with expected strings

> Should they be mixed in with the sample sites like in Pyro

I think this is the way to go and I like all other resolutions. :)

numpyro/util.py
```diff
@@ -408,3 +408,121 @@ def soft_vmap(fn, xs, batch_ndims=1, chunk_size=None):
...
+            event_dim = len(event_shape)
+            shape = getattr(site["value"], "shape", ())
+            batch_shape = shape[: len(shape) - event_dim]
+            event_shape = shape[len(shape) - event_dim :]
+            rows.append(
+                ["value", None]
+                + [str(size) for size in batch_shape]
+                + ["|", None]
+                + [str(size) for size in event_shape]
+            )
```
tcbegley, line 462: Will these two blocks ever actually produce different results?

fritzo: yes, e.g. you can manually observe a non-batched tensor as `value` even if `fn` is batched. Also during enumeration, `value` or `fn` may have extra enumeration dimensions, and they may differ.

test/test_util.py
```diff
@@ -101,3 +103,35 @@ def f(x):
...
+            mean_batch = mean[ind]
+            numpyro.sample("x", dist.Normal(mean_batch, 1), obs=batch)
+
+    with numpyro.handlers.seed(rng_seed=0), numpyro.handlers.trace() as t:
+        model_test()
+
+    assert (
+        format_shapes(t)
+        == "Trace Shapes:       \n Param Sites:       \n         mean    100\n"
+        "Sample Sites:       \n   data plate 10   |\n       x dist 10   |\n        "
```
fehiepsi, line 124: I think it is clearer rearrange those lines into
```python
"Trace Shapes: ...\n" \
" Param Site: ...\n" \
"              ... mean ... 100\n" \
```

fehiepsi (approved):
Looks great to me! Thank you, this utility is really helpful!

numpyro/util.py
tcbegley: Got it, thanks!

test/test_util.py
```diff
@@ -101,3 +103,44 @@ def f(x):
...
+    )
+    assert (
+        format_shapes(t, log_prob=True) == "Trace Shapes:       \n"
+        " Param Sites:       \n"
+        "         mean    100\n"
+        "Sample Sites:       \n"
+        "   data plate 10   |\n"
+        "       x dist 10   |\n"
+        "        value 10   |\n"
+        "     log_prob 10   |"
```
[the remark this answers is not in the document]
tcbegley: I recreated the model from the test in Pyro and added it to the end of [the gist](https://gist.github.com/tcbegley/70cb8e34a681c8324533b88cc5d99b1d). Looks consistent with Pyro to me, let me know what you think?

Will add a scale parameter later to test the event shape is rendered correctly 👍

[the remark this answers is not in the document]
tcbegley: Expanded the test to check event shape formatting.

numpyro/util.py
```diff
@@ -408,3 +408,121 @@ def soft_vmap(fn, xs, batch_ndims=1, chunk_size=None):
         lambda y: jnp.reshape(y, (-1,) + jnp.shape(y)[map_ndims:])[:batch_size], ys
     )
     return tree_map(lambda y: jnp.reshape(y, batch_shape + jnp.shape(y)[1:]), ys)
+
+
+def format_shapes(trace, *, title="Trace Shapes:", last_site=None, log_prob=False):
```
fehiepsi, line 413: It would be nice to expose this utility, at least to the docstring so that users know where to find this and know how to get the trace (especially for models with enumeration). What do you think is a good place for this? I'm thinking of renaming `Visualization` section and adding an entry for this utility there.

I think we can rename `log_prob` to [compute_log_prob](https://docs.pyro.ai/en/stable/_modules/pyro/poutine/trace_struct.html#Trace.compute_log_prob) which supports True/False/callable where callable can be used to filter the sites.

numpyro/util.py
```diff
@@ -408,3 +408,121 @@ def soft_vmap(fn, xs, batch_ndims=1, chunk_size=None):
...
+    """
+    Returns a string showing a table of the shapes of all sites in the
+    trace.
+
+    :param dict trace: The model trace to format. Use ``numpyro.handlers.trace`` to
+        produce the trace.
+    :param str title: Title for the table of shapes.
+    :param str last_site: Name of a site in the model. If supplied, subsequent sites
+        are not displayed in the table.
+    :param bool log_prob: Display shapes of log probabilities for each sample site.
```
fehiepsi, line 423: I think we can rename `log_prob` to [compute_log_prob](https://docs.pyro.ai/en/stable/_modules/pyro/poutine/trace_struct.html#Trace.compute_log_prob) which supports True/False/callable where callable can be used to filter the sites.

numpyro/util.py
```diff
@@ -408,3 +408,121 @@ def soft_vmap(fn, xs, batch_ndims=1, chunk_size=None):
...
     return tree_map(lambda y: jnp.reshape(y, batch_shape + jnp.shape(y)[1:]), ys)
+
+
+def format_shapes(trace, *, title="Trace Shapes:", last_site=None, log_prob=False):
+    """
+    Returns a string showing a table of the shapes of all sites in the
+    trace.
+
+    :param dict trace: The model trace to format. Use ``numpyro.handlers.trace`` to
+        produce the trace.
```
fehiepsi, line 419: How about rephrasing the last sentence to: "Use :class:`~numpyro.handlers.trace` handler (or funsor :class:`~numpyro.contrib.funsor.enum_messenger.trace` handler for enumeration) to produce the trace."

numpyro/util.py
tcbegley: Done, let me know if I did as you were expecting or if you wanted a whole new section.

fehiepsi (approved):

numpyro/util.py
fehiepsi: Thanks! LGTM.

PR #1118, opened 2021-08-05
Recommend AutoNormal guide when hessian in AutoLaplace is singular

### LINKED ISSUE #1001: Use Gauss-Newton hessian matrix in laplace approximation
Currently, in AutoLaplaceApproximation, we use `jax.hessian` to compute Hessian matrix. But the hessian matrix is not always positive definite. So it would be nice to use Gaussian-Newton hessian matrix when [scale_tril = Nan](https://github.com/pyro-ppl/numpyro/blob/8bb94f170de3f6c276fe61e4c92cd4e21de70a4b/numpyro/infer/autoguide.py#L898).

freddyaboulton:
Fixes #1001 

To be honest, I'm not 100% sure what the gauss-newton hessian is so I'm taking my best guess from what I read [here](https://math.stackexchange.com/questions/1105214/difference-between-newtons-method-and-gauss-newton-method/2687949) 😅

### DIFF
--- a/numpyro/infer/autoguide.py
+++ b/numpyro/infer/autoguide.py
@@ -886,8 +886,12 @@ class AutoLaplaceApproximation(AutoContinuous):
         scale_tril = cholesky_of_inverse(precision)
         if not_jax_tracer(scale_tril):
             if np.any(np.isnan(scale_tril)):
-                jacobian = jacfwd(loss_fn)(loc)
-                scale_tril = jnp.outer(jacobian.T, jacobian)
+                warnings.warn(
+                    "Hessian of log posterior at the MAP point is singular. Posterior"
+                    " samples from AutoLaplaceApproxmiation will be constant (equal to"
+                    " the MAP point). Please consider using an AutoNormal guide."
+                )
+            scale_tril = jnp.where(jnp.isnan(scale_tril), 0.0, scale_tril)
         return LowerCholeskyAffine(loc, scale_tril)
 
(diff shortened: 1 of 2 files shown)


fehiepsi:
@freddyaboulton Really sorry, I didn't find references for that feature request. I found [this article](https://andrew.gibiansky.com/blog/machine-learning/gauss-newton-matrix/#Gauss-Newton-Matrix) is helpful and it seems that this method is called "outer product of gradients (OPG)" or BHHH estimator in literature. As I understand, this estimator assumes that the likelihood is `f(latent) = (f_1(latent), ..., f_n(latent))` - then the precision matrix can be approximated by the sum/average of the outer product of gradients of those f_i. This seems to be a bit different from the formula that you use: outer product of gradients of `sum(f_i)`. What do you think?

For now, I'm not sure what is a good way to get a covariance matrix for Laplace approximation in general cases. I guess we can defer users to AutoNormal guides in case the Hessian is singular.

freddyaboulton:
@fehiepsi Thanks for the feedback! Do you want me to update the warning message to recommend the AutoNormal guide when the Hessian is singular or would you rather I close out the PR until we do more research?

fehiepsi:
> recommend the AutoNormal guide when the Hessian is singular

Sure, please go ahead. It is great to have a better warning message. Thanks!

fehiepsi (approved):

fehiepsi:
Thanks, @freddyaboulton!

Issue #1124, opened 2021-08-17
SVI with discrete latent variables


quattro:
Hi all, thanks again for developing such a fantastic inference library. It's been a joy to use for simulations and inference procedures so far.

I noticed the recently implemented `TraceGraph_ELBO` and decided to try it out for inference of discrete latent variables and was met with `UserWarning: Currently, SVI does not support models with discrete latent variables`. Is this a regression bug, or am I mistaken in that `TraceGraph_ELBO` in numpyro cannot handle discrete latent variable?

If it's the former, I'm happy to add a small fork + pull request that adds a simple check on the `loss` member variable within an `SVI` instance.

I'm using `numpyro` version 0.7.1.

fehiepsi:
Hi @quattro, you're right, we can remove that warning now. Please feel free to submit a PR. :)

PR #1125, opened 2021-08-19
Svi discrete

quattro:
Addresses #1124 

Added an attribute to the base ELBO class reflecting the ability to infer discrete latent rvs. Default is to return `False` unless individual subclasses override. 

Might be overkill, but leaves open the opportunity for other ELBO losses supporting discrete inference downstream.


quattro:
Leaked pushes across branches. I'll start a clean branch and re-pull.

PR #1126, opened 2021-08-19
smarter warning about discrete inference in SVI models (#1124)

quattro:
Addresses #1124 

Added class method `can_infer_discrete` to base ELBO class indicating whether or not an ELBO class supports inference of discrete latent variables. Default is `False`. Base classes which support inference should override and return `True`.

This makes room for downstream subclasses to support discrete inference without going into type-checking instances during the SVI warning check. Might be overkill.

### DIFF
--- a/numpyro/infer/elbo.py
+++ b/numpyro/infer/elbo.py
@@ -27,7 +27,7 @@ class ELBO:
 
     """
     Determines whether the ELBO objective can support inference of discrete latent variables.
-    
+
     Subclasses that are capable of inferring  discrete latent variables should override to `True`
     """
     can_infer_discrete = False
--- a/numpyro/infer/elbo.py
+++ b/numpyro/infer/elbo.py
@@ -25,6 +25,13 @@ class ELBO:
         (gradient) estimators.
     """
 
+    """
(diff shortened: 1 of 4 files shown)


numpyro/infer/elbo.py
```diff
@@ -67,6 +67,14 @@ def loss_with_mutable_state(
         """
         raise NotImplementedError("This ELBO objective does not support mutable state.")
 
+    def can_infer_discrete(self):
```
fritzo, line 70: why not a property instead of a method?
```py
can_infer_discrete = False
```

quattro: That works. I can also shorten the name to something that works a bit better in 'not' situations. Maybe `supports_discrete` ?

fehiepsi: maybe `has_discrete_support`? I like `can_infer_discrete` though.

fehiepsi:
@quattro You'll need to revise the warning filter [here](https://github.com/pyro-ppl/numpyro/blob/935d4f5b0d7a9759c44caf3409be9b95e407d956/test/pyroapi/test_pyroapi.py#L10) to make CI pass.

fehiepsi (approved):

PR #1127, opened 2021-08-19
Added Gompertz distribution class, sampling utilities

jackpot-nfer:
Added the ability to sample from the Gompertz distribution by implementing a `Gompertz` class under `numpyro.dist.continuous`

https://www.tandfonline.com/doi/abs/10.1080/03461238.2012.687697


cameronraysmith:
There are approximations for the expectation and variance in

> Lenart A. [The moments of the Gompertz distribution and maximum likelihood estimation of its parameters](https://www.researchgate.net/profile/A-Lenart-2/publication/254355912_The_moments_of_the_Gompertz_distribution_and_maximum_likelihood_estimation_of_its_parameters/links/56ea7fc008ae3a5b48cddafb/The-moments-of-the-Gompertz-distribution-and-maximum-likelihood-estimation-of-its-parameters.pdf). Scand Actuar J. 2014;2014: 255–277. doi:10.1080/03461238.2012.687697

that may be useful.

numpyro/distributions/continuous.py
```diff
@@ -490,6 +490,51 @@ def icdf(self, q):
...
 
+class Gompertz(Distribution):
+    arg_constraints = {
+        "loc": constraints.positive,
+        "shape": constraints.positive,
+    }
+    support = constraints.positive
+    reparametrized_params = ["loc", "shape"]
+
+    def __init__(self, loc=1.0, shape=1.0, validate_args=None):
```
fehiepsi, line 501: In NumPyro, it is more common to use names `mean` and `concentration` (loc is typically used for loc-scale families) -see e.g. [BetaProportion](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/distributions/continuous.py#L1556). Is the default value 1., 1. commonly used (I guess we don't need to specify default values for those parameters.)? You can use `concentration` and `rate` parameterizations to match Gamma distribution (which is related to this distribution) API. Then you can set the default value `rate=1.`.

numpyro/distributions/continuous.py
```diff
@@ -490,6 +490,51 @@ def icdf(self, q):
...
+        self.loc, self.shape = promote_shapes(loc, shape)
+        # Alias parameters as they appear in this article
+        # https://www.tandfonline.com/doi/abs/10.1080/03461238.2012.687697
+        self.__a, self.__b = self.loc, self.shape
```
fehiepsi, line 505: not against but this pattern is not used in the codebase. I guess using the original names in the implementation would be clearer?

numpyro/distributions/continuous.py
```diff
@@ -490,6 +490,51 @@ def icdf(self, q):
...
+    def sample(self, key, sample_shape=()):
+        assert is_prng_key(key)
+        random_shape = sample_shape + self.batch_shape + self.event_shape
+        unifs = random.uniform(key, shape=random_shape)
+        return self.icdf(unifs)
+
+    @validate_sample
+    def log_prob(self, value):
+        # https://www.wolframalpha.com/input/?i=log%28a+e%5E%28-%28a+%28-1+%2B+e%5E%28b+x%29%29%29%2Fb+%2B+b+x%29%29
+        return jnp.log(self.__a) + self.__b * value - self.__a / self.__b * (jnp.exp(self.__b * value) - 1)
```
fehiepsi, line 520: nit: you can store `concentration * value` and use `jnp.expm1` for better numerical stability.

numpyro/distributions/continuous.py
```diff
@@ -490,6 +490,51 @@ def icdf(self, q):
...
+
+    @property
+    def mean(self):
+        # Using approximation from corollary 1 in aforementioned paper
+        return jnp.exp(self.__a / self.__b) * (self.__a / self.__b - jnp.log(self.__a / self.__b) - EULER_MASCHERONI) / self.__b
```
fehiepsi, line 525: nit: avoid recomputing `a/b`

numpyro/distributions/continuous.py
```diff
@@ -490,6 +490,51 @@ def icdf(self, q):
...
+
+    @property
+    def variance(self):
+        raise NotImplementedError("Gompertz variance not yet implemented")
+
+    def cdf(self, value):
+        return 1 - jnp.exp(-1 * self.__a / self.__b * (jnp.exp(self.__b * value) - 1))
```
fehiepsi, line 532: nit: use 2 `expm1`: `-expm1(-a/b* expm1(b*value))`

numpyro/distributions/continuous.py
```diff
@@ -490,6 +490,51 @@ def icdf(self, q):
...
+
+    def icdf(self, q):
+        return jnp.log(1 - jnp.log(1 - q) * self.__b / self.__a) / self.__b
```
fehiepsi, line 535: Using `log1p(-log1p(-q) * b / a) / b`

fehiepsi (commented):
Awesome work, @jackpot-nfer! Thanks for your first contribution and I hope that more will be coming.

To test the implementation, you will need to add some params [here](https://github.com/pyro-ppl/numpyro/blob/935d4f5b0d7a9759c44caf3409be9b95e407d956/test/test_distributions.py#L201).

Could you also expose the distribution to [init file](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/distributions/__init__.py) and add an entry to [the docs](https://github.com/pyro-ppl/numpyro/blob/master/docs/source/distributions.rst)?

numpyro/distributions/continuous.py
fritzo: As a courtesy to users who want to debug their models, we should avoid double underscores and the implied [name mangling](https://stackoverflow.com/questions/1301346).

numpyro/distributions/continuous.py
jackpot-nfer: I went with `rate` and `concentration`.

numpyro/distributions/continuous.py
jackpot-nfer: Removed

numpyro/distributions/continuous.py
jackpot-nfer: Resolved

numpyro/distributions/continuous.py
jackpot-nfer: Resolved

numpyro/distributions/continuous.py
jackpot-nfer: Resolved

numpyro/distributions/continuous.py
jackpot-nfer: Resolved

jackpot-nfer:
will add tests / docs shortly

numpyro/distributions/continuous.py
fehiepsi: I like those parameters. Could you swap the order and set `rate=1.` by default? (this is to match Gamma distribution)

numpyro/distributions/continuous.py
```diff
@@ -490,6 +490,66 @@ def icdf(self, q):
         return self.loc - self.scale * jnp.log(-jnp.log(q))
 
 
+class Gompertz(Distribution):
+    arg_constraints = {
+        "rate": constraints.positive,
+        "concentration": constraints.positive,
+    }
+    support = constraints.positive
+    reparametrized_params = ["loc", "shape"]
```
fehiepsi, line 499: nit: `concentration`, `rate`

numpyro/distributions/continuous.py
```diff
@@ -511,7 +571,7 @@ def sample(self, key, sample_shape=()):
 
     @validate_sample
     def log_prob(self, value):
-        normalize_term = jnp.log(2 * self.scale)
+        normalize_term = jnp.log1p(2 * self.scale)
```
fehiepsi, line 574: Is this a bug?

jackpot-nfer: it was on accident! undone

numpyro/distributions/continuous.py
jackpot-nfer: taken care of

jackpot-nfer:
hi, I fixed the usage of `log1p` and `expm1`. I also undid my accidental edit to `Laplace. I swapped the order of `rate` and `concentration` now to reflect the gamma distribution args, and set `rate = 1.0` by default.

How can I get started with adding tests and docs?

jackpot-nfer:
> hi, I fixed the usage of `log1p` and `expm1`. I also undid my accidental edit to `Laplace`. I swapped the order of `rate` and `concentration` now to reflect the gamma distribution args, and set `rate = 1.0` by default.
> 
> How can I get started with adding tests and docs?

I'm confused by the `T(<Class>, <Vals>` idiom and don't write to write a trivial test suite

fehiepsi:
Hi @jackpot-nfer, for testing, you will need to add `Gompertz` to [init](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/distributions/__init__.py) (and also [docs](https://github.com/pyro-ppl/numpyro/blob/master/docs/source/distributions.rst)), and add a couple of entries to [here](https://github.com/pyro-ppl/numpyro/blob/master/test/test_distributions.py#L235):
```
    T(dist.Gompertz, jnp.array([1.7]), jnp.array([[2.0], [3.0]])),
    T(dist.Gompertz, jnp.array([0.5, 1.3]), jnp.array([[1.0], [3.0]])),
```

numpyro/distributions/continuous.py
```diff
@@ -490,6 +490,69 @@ def icdf(self, q):
...
+        "concentration": constraints.positive,
+        "rate": constraints.positive,
+    }
+    support = constraints.positive
+    reparametrized_params = ["concentration", "rate"]
+
+    def __init__(self, concentration, rate=1.0, validate_args=None):
+        """
+        The CDF is
+        ..math
```
fehiepsi, line 504: I think you need `.. math::`

numpyro/distributions/continuous.py
```diff
@@ -490,6 +490,69 @@ def icdf(self, q):
...
+            * (rate_over_conc - jnp.log(rate_over_conc) - EULER_MASCHERONI)
+            / self.concentration
+        )
+
+    @property
+    def variance(self):
+        raise NotImplementedError("Gompertz variance not yet implemented")
+
+    def cdf(self, value):
+        return 1 - jnp.exp(
```
fehiepsi, line 545: nit: use `expm1`

cameronraysmith:
@fehiepsi if you would please take another look I believe you will find @jackpotrykus has addressed the existing comments.

jackpot-nfer:
@fehiepsi I believe all the necessary tests and docs edits have been made.

numpyro/distributions/continuous.py
```diff
@@ -448,6 +448,85 @@ def tree_flatten(self):
...
+
+    All derivations can be found in this paper (Lenart, 2012):
+        https://doi.org/10.1080/03461238.2012.687697
+    with the exception that we call the parameters `concentration` and `rate`,
+    as opposed to their "location" and "scale" terms.
+
+    The CDF is
+
+    .. math::
+        F(x) = 1 - \exp \left\{ \frac{\text{rate}}{\text{concentration}} * \left [ \exp\{\text{concentration} \cdot x\} - 1 \right ] \right\}
```
fehiepsi, line 462: Hi @jackpot-nfer, the implementation looks correct to me, but it seems to me that the names of the parameters are a bit strange. E.g. by using `rate`, I expected that `F(x)` should depend on `x / rate`, rather than `concentration * x`. Could you use the [wikipedia](https://en.wikipedia.org/wiki/Gompertz_distribution) derivation instead? I believe that you just need to change:
+ concentration -> 1 / rate
+ rate_over_conc -> concentration
+ rate -> concentration * rate
across the implementation.

fehiepsi:
@jackpot-nfer There are also a couple of lint issues. You can run `make lint` locally to see them. I guess you need two `\\` like [this example](https://github.com/pyro-ppl/numpyro/blob/master/examples/sparse_regression.py#L15).

numpyro/distributions/continuous.py
cameronraysmith: @fehiepsi would the following work for you?

```latex
\begin{align}
\text{concentration} &\rightarrow \frac{1}{\text{rate}}\\
\text{rate} &\rightarrow \frac{\text{concentration}}{\text{rate}}
\end{align}

$$F(x) = 1 - \exp \left\{ -\text{concentration} * \left [ \exp \left\{\frac{x}{\text{rate}}\right\} - 1 \right ] \right\}$$
```

If we return precisely to the wikipedia definition for [Gompertz](https://en.wikipedia.org/wiki/Gompertz_distribution) I think we might violate the analogy with the definition of the `Gamma` distribution in terms of concentration and rate parameters (as opposed to shape and scale) that [you mentioned above](https://github.com/pyro-ppl/numpyro/pull/1127#discussion_r692634197).

Alternatively, we could note that the current parameterization is quite similar to the form appearing in the [Gompertz-Makeham law of mortality](https://en.wikipedia.org/wiki/Gompertz%E2%80%93Makeham_law_of_mortality) where $\alpha \rightarrow \text{rate}$ and $\beta \rightarrow \text{concentration}$. Of course the Makeham term is absent here.

fehiepsi: > $$F(x) = 1 - \exp \left\{ -\text{concentration} * \left [ \exp \left\{\frac{x}{\text{rate}}\right\} - 1 \right ] \right\}$$

Yep, I vote for this. This matches wikipedia definition of shape and scale. In NumPyro Gamma.concentration = Wikipedia's Gamma.shape, and Gamma.rate = Wikipedia's Gamma.rate=1/Gamma.scale) - so we are using the same pattern here for Gompertz. In Gompertz-Makeham law of mortality, I saw `x` go along with `beta`, so beta should be the scale parameter (using `concentration` in place of `scale:=1/rate` is strange to me).

fehiepsi: I don't know why the "The Gompertz distribution and Maximum Likelihood Estimation of its parameters" paper uses the name "shape" for the scale parameter - this is not conventional I believe.

jackpot-nfer:
Hey @fehiepsi, I've reparametrized the distribution as requested. It now takes `concentration` and `rate=1.0` as args. I also added a `scale` argument and attribute. If it's provided (`None` by default), it sets `rate = 1 / scale`. If not, the `scale` parameter is set to `1 / rate` as you'd expect. I figured this was a good compromise between sticking to the wikipedia derivation and also matching the `Gamma` distribution.

I'm having trouble to get the tests to run right now -- tried running from a fresh `venv` and still didn't work. Not sure what is causing that. Could you try running them?

I manually `black`ened and `isort`ed the change files, so all should be good to go on that front.

numpyro/distributions/continuous.py
```diff
@@ -451,32 +451,45 @@ def tree_flatten(self):
 class Gompertz(Distribution):
     """Gompertz Distribution sampling utilities.
 
-    All derivations can be found in this paper (Lenart, 2012):
-        https://doi.org/10.1080/03461238.2012.687697
-    with the exception that we call the parameters `concentration` and `rate`,
-    as opposed to their "location" and "scale" terms.
+    This implementation is based off the derivations from Wikipedia page for the Gompertz distribution.
```
fehiepsi, line 454: Could you make the doc max line length shorter? Currently, some lines are long and `lint` does not pass.

test/test_distributions.py
```diff
@@ -257,6 +257,8 @@ def get_sp_dist(jax_dist):
     T(dist.Gamma, jnp.array([0.5, 1.3]), jnp.array([[1.0], [3.0]])),
     T(dist.GaussianRandomWalk, 0.1, 10),
     T(dist.GaussianRandomWalk, jnp.array([0.1, 0.3, 0.25]), 10),
+    T(dist.Gompertz, jnp.array([1.7]), jnp.array([[2.0], [3.0]])),
+    T(dist.Gompertz, jnp.array([0.5, 1.3]), jnp.array([[1.0], [3.0]])),
```
fehiepsi, line 261: Could you add pytest.skip in `test_mean_var` of this file?

fehiepsi (commented):
Thanks for updating and adding detailed documentation, @jackpot-nfer! For testing, I think you can do `pip install -e .[dev, test]` (at the numpyro directory) to install all dependencies. Looking like most tests passed. This is almost good to go. I just have a few minor comments below.

fehiepsi:
@jackpot-nfer It has been a while but I guess we're only blocked by missing some tests. Can I add a commit for it in this PR?

cameronraysmith:
@fehiepsi I don't want to speak for @jackpotrykus , but I'm quite certain this would be fine.

fehiepsi:
We just merged #1551. Thanks @jackpot-nfer and @cameronraysmith !!

Issue #1129, opened 2021-08-22
Adding Dirichlet prior for Ordinal Regression case study


svilupp:
I'd like to contribute an implementation of a more principled prior based on [Ordinal Regression Case Study from M.Betancourt (2019)](https://betanalpha.github.io/assets/case_studies/ordinal_regression.html) (in particular section 2.2). It requires adding a Simplex2Ordered Transform.

Branch location: https://github.com/svilupp/numpyro/tree/ordinal-regr-prior

Changes have been made in the following files (so far):
- distributions/transforms.py - adding the necessary transform
- docs/source/distributions.rst - adding the reference to be displayed in the docs
- notebooks/source/ordinal_regression.ipynb - showcasing how to use it (model4)

I have a few **questions**:
- Am I right to assume that log_abs_det_jacobian() method of Transforms Class is defined as additive in the *transformed space* (ie, log_abs_det_jacobian+log_transformed_density = log_base_density). I have noticed that TransformedDistribution deducts the value from the prob density of the base distribution + I have confirmed empirically. 
For that reason, I have *flipped the sign* compared to the log_abs_det_jacobian in [1]
- I haven't noticed any transform-specific tests (only the class itself). Have I missed it?
- Is the verbosity of the `model4` in Ordinal Regression example appropriate?
- Is there any other file that should be changed?
- All tests pass, until doctest where I get: 
```
Warning, treated as error:
/home/ec2-user/SageMaker/Other/Contrib/numpyro/docs/source/tutorials/bayesian_hierarchical_linear_regression.ipynb:7:Unexpected indentation.
``` 
I haven't made any changes to this file. I have seen this issue before with docstrings missing new lines and I can see a cell, which could be problematic (download URL for the dataset). However, changing it to one long line has no impact on the error. Has anyone encountered this error as well?

***
**TODOs**:

- [x] ensure transform call() and inverse() can handle batch dimensions (currently, it works on vector--> vector basis)
- [x] ensure log_abs_det of the transform can handle batch dimensions 

***
PS: I think there might be some breaking changes in the latest version of JAX, as I had to install jax==0.2.13 jaxlib==0.1.65 otherwise tests would not pass (failing test was hmm_enum.py and some simple array slicing, so the changes seem to be around handling ellipsis) 

***
UPDATE: 
- transform should now fully support batch dimensions

fehiepsi:
This is awesome! Waiting for your PR, @svilupp!

[...]

PR #1130, opened 2021-08-23
Adding Dirichlet prior for Ordinal Regression case study

svilupp:
Details in [Issue #1129](https://github.com/pyro-ppl/numpyro/issues/1129)

Changes have been made in the following files (so far):
- distributions/transforms.py - adding the necessary transform
- docs/source/distributions.rst - adding the reference to be displayed in the docs
- notebooks/source/ordinal_regression.ipynb - showcasing how to use it (model4)

Tests:
- lint, format, test have passed
- doctest fails on warning around "Unexpected indentation" in a file I haven't changed

There are a few outstanding changes now based on comments from Fehiepsi


numpyro/distributions/transforms.py
```diff
@@ -790,6 +791,72 @@ def log_abs_det_jacobian(self, x, y, intermediates=None):
...
+        J=J.at[...,0].set(1.0)
+        # partial derivation of sigma
+        rho=jnp.multiply(sigma,1-sigma)
+        i,j=jnp.diag_indices(K-1) 
+        # diagonal element
+        J=J.at[..., i+1, j+1].set(-rho)
+        # off-diagonal element
+        J=J.at[..., i, j+1].set(rho)
+
+        _,log_det=jnp.linalg.slogdet(J)
```
fehiepsi, line 857: lint: you need 2 spaces around ` = `. I guess you can run `make format` to automatically perform that job.

numpyro/distributions/transforms.py
```diff
@@ -790,6 +791,72 @@ def log_abs_det_jacobian(self, x, y, intermediates=None):
         x_abs = jnp.abs(x)
         return -x_abs - 2 * jnp.log1p(jnp.exp(-x_abs))
 
+class Simplex2OrderedTransform(Transform):
```
fehiepsi, line 794: nit: 2 empty lines before class definition.

numpyro/distributions/transforms.py
```diff
@@ -41,6 +41,7 @@
     "PermuteTransform",
     "PowerTransform",
     "SigmoidTransform",
+    "Simplex2OrderedTransform",
```
fehiepsi, line 44: Could you think of another name for this transform? In our forum, I use this name just to clarify the idea. For end-user, I think that `SimplexToOrderedTransform` might be a better name.

numpyro/distributions/transforms.py
```diff
@@ -790,6 +791,72 @@ def log_abs_det_jacobian(self, x, y, intermediates=None):
...
+    
+    def __call__(self, x):        
+        def scan_fn(expit_term, x):
+            y=self.anchor_point-logit(expit_term-x)        
+            next_expit_term=expit(self.anchor_point-y)
+            return next_expit_term, y
+        
+        # the first term defies the general map formula
+        # hence, the most succinct implementation is to parametrize the changing term
+        expit_init=jnp.broadcast_to(1.0,x.shape)[...,0]
```
fehiepsi, line 819: nit, use `jnp.ones(x.shape[:-1])`

numpyro/distributions/transforms.py
```diff
@@ -790,6 +791,72 @@ def log_abs_det_jacobian(self, x, y, intermediates=None):
...
+     
+    **References:**
+
+    1. *Ordinal Regression Case Study, section 2.2*,
+       M. Betancourt, https://betanalpha.github.io/assets/case_studies/ordinal_regression.html
+    """    
+    domain = constraints.simplex
+    codomain = constraints.ordered_vector
+    
+    def __init__(self, anchor_point=0.0):
```
fehiepsi, line 808: Could you add docs to this anchor_point? I guess we assume it is scalar for simplicity?

numpyro/distributions/transforms.py
```diff
@@ -790,6 +791,72 @@ def log_abs_det_jacobian(self, x, y, intermediates=None):
...
+            next_expit_term=expit(self.anchor_point-y)
+            return next_expit_term, y
+        
+        # the first term defies the general map formula
+        # hence, the most succinct implementation is to parametrize the changing term
+        expit_init=jnp.broadcast_to(1.0,x.shape)[...,0]
+        # scan must run along the last dimension, so we need to move that dimension to be the leading
+        _, y = lax.scan(scan_fn, expit_init,jnp.moveaxis(x[...,:-1],-1,0))
+        # reverse the move for last dimension to represent the event shape
+        y=jnp.moveaxis(y,0,-1)
```
fehiepsi, line 823: whoa, thanks for supporting batching!

numpyro/distributions/transforms.py
```diff
@@ -790,6 +791,72 @@ def log_abs_det_jacobian(self, x, y, intermediates=None):
...
+        # partial derivation of sigma
+        rho=jnp.multiply(sigma,1-sigma)
+        i,j=jnp.diag_indices(K-1) 
+        # diagonal element
+        J=J.at[..., i+1, j+1].set(-rho)
+        # off-diagonal element
+        J=J.at[..., i, j+1].set(rho)
+
+        _,log_det=jnp.linalg.slogdet(J)
+        # NB: sign has to be the opposite of [1] due to Numpyro's implementation of TransformedDistribution
```
fehiepsi, line 858: here we compute log_det of the transform from x to y, but the above lines compute log_det of the inverse transform from y to x, so we need to take the minus sign here. This is not related to the implementation of TransformedDistribution.

test/test_distributions.py
```diff
@@ -1651,9 +1652,10 @@ def inv_vec_transform(y):
             ),
             (2,),
         ),
+        (Simplex2OrderedTransform(0.0),(5,))
     ],
 )
-@pytest.mark.parametrize("batch_shape", [(), (1,), (3,), (6,), (3, 1), (1, 3), (5, 3)])
+@pytest.mark.parametrize("batch_shape", [pytest.param((), marks=pytest.mark.xfail(reason="Simplex2Ordered cannot run with batch_shape=()")), (1,), (3,), (6,), (3, 1), (1, 3), (5, 3)])
```
fehiepsi, line 1658: To resolve this issue, I guess you can test for
```
ComposedTransform([biject_to(constraints.simplex), Simplex2OrderedTransform(0.0), biject_to(constraints.ordered_vector).inv])
```
This one will test for the transform: `x_unconstrained -> (x -> y ->) y_unconstrained` instead of `x -> y`.

numpyro/distributions/transforms.py
svilupp: Agreed!

numpyro/distributions/transforms.py
svilupp: SimplexToOrderedTransform sounds good. I'll reflect it in all deps.

numpyro/distributions/transforms.py
svilupp: Silly me! I forgot to run black after the latest commits.

numpyro/distributions/transforms.py
svilupp: Black!

numpyro/distributions/transforms.py
svilupp: I've updated the comment.

numpyro/distributions/transforms.py
svilupp: I've added a dedicated one-liner instead of the brief mention that was there before. Is that what you had in mind?

":param anchor_point: Anchor point is a nuisance parameter to improve the identifiability of the transform. For simplicity, we assume it is a scalar value."

test/test_distributions.py
svilupp: Would you mind saying a bit more about this proposal?

It seems similar to what `test_bijective_transforms` does, ie, Normal--> biject to domain (=x) --> SimplexToOrdered (=y) and then we test.

I've tried to plug it into parameterized but it also fails in the jacobian step - this time not on the square shape, but on the value itself (`actual J != excepted J`)

fehiepsi: Sorry, I should be clearer. I meant to replace
```python
(Simplex2OrderedTransform(0.0),(5,))
```
by
```python
(ComposedTransform([biject_to(constraints.simplex), Simplex2OrderedTransform(0.0), biject_to(constraints.ordered_vector).inv]), (5,))
```
The test will compute log det jacobian of the composed transform implementation, and compare it with the autodiff one.

numpyro/distributions/transforms.py
fehiepsi: I think you can link it to the reference. As a user, it is still not clear for me how to interpret it. If we allow non-scalar values, what is the expected shape? Should it be the same as `x`, or broadcastable to `x`,... It also depends on your implementation. Based on the implementation, I guess `anchor_point.shape` should be broadcastable to `x.shape[:-1]`?

test/test_distributions.py
fehiepsi: > actual J != excepted J

Your implementation looks like the one in Stan's implementation. I'm not sure what's going wrong. The derivation of jacobian seems not trivial to me. I'll look into details later in the week. One thing I'm not sure in the Stan implementation is why Jacobian has shape 5 x 5. Shouldn't it have shape 4 x 4?

numpyro/distributions/transforms.py
svilupp: So my interpretation is that it's like a "baseline affinity" for the map between proba and cutpoints to be one-to-one. Practically, we could exclude it (because it would be implicitly 0.0, ie, centered around 0.0 anchor point), but we need to make sure there is no intercept in your model (or a fixed covariate). Otherwise, we would introduce the non-identifiability again. I think it's just MB being very precise and explicit rather than implicit.

***
I think you're right with the shape `x.shape[:-1]` - it absolutely must be fixed over the event_shape (axis=-1). 
However, if we are modelling cutpoints that should come from the same distribution (same `true` model, just batched) then it must be fixed (assuming it's the "baseline affinity" that effectively calibrates/anchors the cutpoints). 
Ie, I don't see the need to support anything but scalar - WDYT?

test/test_distributions.py
svilupp: Yes, that's exactly what I've done and I was confused by the error as well.

As for the Jacobian - that puzzled me too at first! 
It's a square matrix because the map is `p` of dim 5 --> (`c` of dim 4, S ), where S is a variable representing the normalization (sum of p = 1); That's why the left column in Jacobian is ones -- because dp\dS is 1
Together it's a 5x5 map.

> Case Study 2.2: ... This allows us to transform the K probabilities (p1,…,pK) to a new variable that encodes the normalization, S=∑Kk=1pk=1 and the K−1 cut points, (S,c1,…,cK−1).... 

It doesn't matter for the density afterwards, because it's marginalized out

*** 
Not sure what to do with the autodiff - I have verified the Jacobian earlier by some simply density simulations (on samples), but it's a fairly coarse method, so it wouldn't catch small differences.

numpyro/distributions/transforms.py
```diff
@@ -791,6 +792,78 @@ def log_abs_det_jacobian(self, x, y, intermediates=None):
...
+        def scan_fn(expit_term, x):
+            y = self.anchor_point - logit(expit_term - x)
+            next_expit_term = expit(self.anchor_point - y)
+            return next_expit_term, y
+
+        # the first term defies the general map formula
+        # hence, the most succinct implementation is to parametrize the changing term
+        expit_init = jnp.ones(x.shape[:-1])
+        # scan must run along the last dimension, so we need to move that dimension to be the leading
+        _, y = lax.scan(scan_fn, expit_init, jnp.moveaxis(x[..., :-1], -1, 0))
```
fehiepsi, line 826: I think `scan` is unnecessary. The transform from cutpoints to probs is
```
x = x - self.anchor_point
s = expit(x)
# I prefer to thinking this way, rather than p0 = 1 - expit(-x0) as in Stan thread
# p0 = s0, p1 = s1 - s0, p2 = s2 - s1,..., pn = 1 - s[n-1]
# add two boundary points 0 and 1
pad_width = [(0, 0)] * (jnp.ndim(s) - 1) + [(1, 1)]
s = jnp.pad(s, pad_width, constant_values=(0, 1))
p = s[..., 1:] - s[..., :-1]
```
and the inverse
```
s = jnp.cumsum(p[..., :-1], axis=-1)
x = logit(s) + self.anchor_point
```

test/test_distributions.py
fehiepsi: Based on the derivation in my below comment, it seems that Jacobian has an analytical form:
```python
# |dp/dc| = prod(ds/dx) = prod(expit'(x))
# we know log derivative of expit(x) is `-softplus(x) -softplus(-x)`
J = -(softplus(x) + softplus(-x)).sum(-1)
return J
```
where `x` is the cutpoints. Could you give it a try?

numpyro/distributions/transforms.py
fehiepsi: Thanks for the clear explanation! Now, I understand what anchor_point is and see why you wanted to define `OrderedToSimplexTransform` (because `anchor_point` is more related to cut_points than the probability). Would it make more sense to construct it instead of `SimplexToOrderedTransform`?

Regarding the shape of `anchor_point`, I just meant to let users know what's its expected shape. I think you can either mentioned that `anchor_point` shape should be broadcastable to `x.shape[:-1]` or add some assertation in the call method.

test/test_distributions.py
fehiepsi: It is `cutpoint -> probability`. The Jacobian matrix (which has diagonal + off-diagonal bands) is lower triangular, so its determinant is the product of entries on the diagonal.

> So if anything, maybe we could skip the S=sum(p_i) variable trick (and its column of 1s) and just have a square Jacobian by defining only the p_0, ..., p_{k-1}

Yes, I think this is a right approach. Let's verify it empirically to see if tests pass (i.e. using the Jacobian term in my last comment - note that `x` is the cutpoints in that derivation - and rerun the test).

fehiepsi: I just did some tests to verify that the above calculation is correct. :)
```python
import numpyro
from numpyro.distributions import constraints, transforms
import jax
import jax.numpy as jnp

# TODO: use this transform in OrderedLogistic
class OrderedToSimplexTransform(transforms.Transform):
    domain = constraints.ordered_vector
    codomain = constraints.simplex

    def __init__(self, anchor_point=0.0):
        self.anchor_point = anchor_point

    def __call__(self, x):
        x = x - jnp.expand_dims(self.anchor_point, -1)
        sigma = jax.nn.sigmoid(x)
        pad_width = [(0, 0)] * (jnp.ndim(sigma) - 1) + [(1, 1)]
        sigma = jnp.pad(sigma, pad_width, constant_values=(0, 1))
        p = sigma[..., 1:] - sigma[..., :-1]
        return jnp.clip(p, a_min=jnp.finfo(p.dtype).tiny)

    def _inverse(self, y):
        sigma = jnp.cumsum(y[..., :-1], axis=-1)
        x = jax.scipy.special.logit(sigma)
        return x + jnp.expand_dims(self.anchor_point, -1)

    def log_abs_det_jacobian(self, x, y, intermediates=None):
        x = x - jnp.expand_dims(self.anchor_point, -1)
        return -(jax.nn.softplus(x) + jax.nn.softplus(-x)).sum(-1)

    def forward_shape(self, shape):
        if len(shape) < 1:
            raise ValueError("Too few dimensions on input")
        return shape[:-1] + (shape[-1] + 1,)

    def inverse_shape(self, shape):
        if len(shape) < 1:
            raise ValueError("Too few dimensions on input")
        return shape[:-1] + (shape[-1] - 1,)


t = transforms.ComposeTransform([transforms.OrderedTransform(),
                                 OrderedToSimplexTransform(),
                                 StickBreakingTransform().inv])
x = jax.random.normal(jax.random.PRNGKey(1), (5,))
expected = jnp.linalg.slogdet(jax.jacobian(t)(x))[1]
actual = t.log_abs_det_jacobian(x, t(x))
print(expected, actual)
# 1.7702796 1.7702808
```

numpyro/distributions/transforms.py
svilupp: I've added the comment into docs. As for OrderedToSimplex - I'd prefer to stick to the current implementation, as it feels more natural to transform Dirichlet in call() direction.

test/test_distributions.py
svilupp: As always, you're right - I forgot that we lose the column of 1s, so the zeros under the diagonal will wipe the off-diagonal band.

I've plugged it in as per your suggestion (with `-J_logdet`) given the direction it works in.

numpyro/distributions/transforms.py
svilupp: Very ellegant! Updated in the latest push

numpyro/distributions/transforms.py
fehiepsi: Thanks, yep let's stick to OrderedToSimplex. :)

numpyro/distributions/transforms.py
```diff
@@ -791,6 +792,50 @@ def log_abs_det_jacobian(self, x, y, intermediates=None):
...
+
+    domain = constraints.simplex
+    codomain = constraints.ordered_vector
+
+    def __init__(self, anchor_point=0.0):
+        self.anchor_point = anchor_point
+
+    def __call__(self, x):
+        s = jnp.cumsum(x[..., :-1], axis=-1)
+        y = logit(s) + self.anchor_point
```
fehiepsi, line 819: use `jnp.expand_dims(self.anchor_point, -1)`

numpyro/distributions/transforms.py
```diff
@@ -791,6 +792,50 @@ def log_abs_det_jacobian(self, x, y, intermediates=None):
...
+        return y
+
+    def _inverse(self, y):
+        y = y - self.anchor_point
```
fehiepsi, line 823: ditto: `jnp.expand_dims`

numpyro/distributions/transforms.py
```diff
@@ -791,6 +792,50 @@ def log_abs_det_jacobian(self, x, y, intermediates=None):
...
+        pad_width = [(0, 0)] * (jnp.ndim(s) - 1) + [(1, 1)]
+        s = jnp.pad(s, pad_width, constant_values=(0, 1))
+        x = s[..., 1:] - s[..., :-1]
+        return x
+
+    def log_abs_det_jacobian(self, x, y, intermediates=None):
+        # |dp/dc| = |dx/dy| = prod(ds/dy) = prod(expit'(y))
+        # we know log derivative of expit(y) is `-softplus(y) - softplus(-y)`
+        J_logdet = -(softplus(y) + softplus(-y)).sum(-1)
+        return -J_logdet
```
fehiepsi, line 836: nit: no need for two `-` signs

fehiepsi (approved):
Thanks, @svilupp! Could you help me replace [those lines](https://github.com/pyro-ppl/numpyro/blob/87ff265a0845fef194b230906081d8509376902a/numpyro/distributions/discrete.py#L435-L439) by
```
probs = SimplexToOrderedTransform(self.predictor).inv(self.cutpoints)
```
Regarding the precision, I think you can add some tolerant `rtol`. In the other PR, I just wanted to address an issue with large event_shape, where `log_det` returns `nan`/`inf` when using StickBreakingTransform.

test/test_distributions.py
fehiepsi: Sure, you can use the following derivation
```
softplus(x) = -log(sigmoid(-x))
sigmoid'(x) = sigmoid(x)*sigmoid(-x)
=> log (sigmoid'(x)) = -softplus(-x) - softplus(x)
```

fehiepsi (approved):
This work is awesome! Thanks for your contribution, @svilupp!

PR #1132, opened 2021-08-24
add AutoDAIS guide

martinjankowiak:
autoguide that uses hmc inside of variational inference.


test/infer/test_autoguide.py
```diff
@@ -74,14 +76,16 @@ def body_fn(i, val):
...
     # test .sample_posterior method
     posterior_samples = guide.sample_posterior(
         random.PRNGKey(1), params, sample_shape=(1000,)
     )
-    assert_allclose(jnp.mean(posterior_samples["beta"], 0), true_coefs, atol=0.05)
+    posterior_mean = jnp.mean(posterior_samples["beta"], 0)
+    assert_allclose(posterior_mean, true_coefs, atol=0.05)
 
-    if auto_class not in [AutoDelta, AutoIAFNormal, AutoBNAFNormal]:
+    if auto_class not in [AutoDAIS, AutoDelta, AutoIAFNormal, AutoBNAFNormal]:
```
martinjankowiak, line 88: currently the final assert in this test fails although the rest of the test is ok. not really sure why

numpyro/infer/autoguide.py
```diff
@@ -608,6 +610,108 @@ def sample_posterior(self, rng_key, params, sample_shape=()):
...
+            log_factor = log_factor + momentum_dist.log_prob(v_hat) - momentum_dist.log_prob(v_prev)
+            return (theta, v, log_factor), None
+
+        (theta, _, log_factor), _ = jax.lax.scan(scan_body, (theta_0, v_0, 0.), eps)
+
+        numpyro.factor("factor", -log_factor)
+
+        return theta
+
+    def sample_posterior(self, rng_key, params, sample_shape=()):
```
martinjankowiak, line 701: @fehiepsi is this why the last part of the test is failing? not really sure if i did this right (even for `sample_shape=(N,)`)

numpyro/infer/autoguide.py
```diff
@@ -608,6 +610,150 @@ def sample_posterior(self, rng_key, params, sample_shape=()):
...
+        def _single_sample(_rng_key):
+            latent_sample = handlers.substitute(
+                handlers.seed(self._sample_latent, _rng_key), params
+            )(sample_shape=())
+            return self._unpack_and_constrain(latent_sample, params)
+
+        if sample_shape:
```
martinjankowiak, line 746: @fehiepsi so i think this now supports arbitrary `sample_shape` correctly. but is `lax.map` really what i want on line 748? this won't be vectorized right?

fehiepsi: Using `lax.map` is fine I think. `vmap` over all `num_samples` can be expensive.

numpyro/infer/autoguide.py
fehiepsi: There was a bug in `soft_vmap` to handler empty arrays (appears at `factor` sites). It is addressed in #1011.

numpyro/util.py
```diff
@@ -382,7 +382,7 @@ def soft_vmap(fn, xs, batch_ndims=1, chunk_size=None):
 
     # we'll do map(vmap(fn), xs) and make xs.shape = (num_chunks, chunk_size, ...)
     num_chunks = batch_size = int(np.prod(batch_shape))
-    prepend_shape = (-1,) if batch_size > 1 else ()
+    prepend_shape = (batch_size,) if batch_size > 1 else ()
```
martinjankowiak, line 385: @fehiepsi i guess these changes are leading to some failing tests, for example:

test/infer/test_infer_util.py::test_log_likelihood[batch_shape0] FAILED
test/test_util.py::test_soft_vmap[None-batch_shape0] FAILED
test/test_util.py::test_soft_vmap[None-batch_shape1] FAILED

fehiepsi: fixed, `np.prod(...)` returns a float value?? I just convert it to `int`.

PR #1140, opened 2021-08-30
Make Sigmoid and StickBreaking transforms more stable

fehiepsi:
This PR is motivated during the discussion in #1130. I got some numerical issues when using those transforms so would like to fix them.

Current bijective tests should cover the changes.


fritzo (approved):

fehiepsi:
Thanks, Fritz!

PR #1141, opened 2021-09-01
add basic horseshoe regression example

martinjankowiak:
@fehiepsi what other changes do i need to make to .rsts etc?


fehiepsi (approved):
LGTM. Could you add the example to [this file](https://github.com/pyro-ppl/numpyro/blob/master/docs/source/index.rst)?

martinjankowiak:
@fehiepsi i updated this to include a binary response model

fehiepsi (approved):

fehiepsi (approved):
Thanks, Martin! This is a very nice example.

PR #1142, opened 2021-09-01
add notebook that explores bad posterior geometries and tips for dealing with them

martinjankowiak:
@fehiepsi for some time i've wanted to add a notebook that goes over some tips for dealing with bad geometries. here's a start at such a tutorial:  [updated nbviewer link](https://github.com/pyro-ppl/numpyro/blob/cfe6bf4fabd99017bc2c34256503de2f7d5566b5/notebooks/source/bad_posterior_geometry.ipynb)

any other ideas? @fritzo


fritzo:
Great idea @martinjankowiak. Actually this morning I was thinking of creating a Pyro tutorial about `pyro.reparam` and geometry, but then decided it would be better to wait until we have the first few automatic reparametrization strategies https://github.com/pyro-ppl/pyro/pull/2884

Minor suggestions:
- [ ] Emphasize the difference between _manually_ reparametrizing a model versus using using the reparametrization strategies in `handlers.reparam`. Actually I'm not sure how easy this is in NumPyro, as some of the reparametrizers have learnable parameters and require an initial SVI phase to calibrate those parameters.
- [ ] Mention Neutra
- [ ] nit: In the manual reparam model, I'd add some newlines and specifically call out the one or two lines that differ between models

fehiepsi:
Oh nice, I intended to mention the structure mass matrix but you already added the example 3. 👍

martinjankowiak:
@fehiepsi can you please check that i'm using `TransformReparam` correctly?

fehiepsi:
Yep, that's right. I think it would be cleaner to define a model, I think you can also do
```
_rep_hs_model2 = numpyro.handlers.reparam(_rep_hs_model1, config={"betas": LocScaleReparam(0)})
```

fehiepsi (approved):

martinjankowiak:
@fehiepsi tried to use `LocScaleReparam` but it doesn't seem to work as expected... can you take a look?

fehiepsi:
It is a bit strange. Seems like a bug to me. I tried various assumptions but still do not understand why yet.

fehiepsi:
@martinjankowiak There are a couple of bugs in the models:
```python
def _unrep_hs_model(X, Y):
    lambdas = numpyro.sample("lambdas", dist.HalfCauchy(jnp.ones(X.shape[1])))
    tau = numpyro.sample("tau", dist.HalfCauchy(jnp.ones(1)))
    betas = numpyro.sample("betas", dist.Normal(tau * lambdas))  # should be scale=tau * lambdas
    mean_function = jnp.dot(X, betas)
    numpyro.sample("Y", dist.Normal(mean_function, 0.05), obs=Y)
```
```python
def _rep_hs_model1(X, Y):
    lambdas = numpyro.sample("lambdas", dist.HalfCauchy(jnp.ones(X.shape[1])))
    tau = numpyro.sample("tau", dist.HalfCauchy(jnp.ones(1)))
    unscaled_betas = numpyro.sample("unscaled_betas", dist.Normal(jnp.ones(X.shape[1])))  # should be jnp.zeros
    scaled_betas = numpyro.deterministic("betas", tau * lambdas * unscaled_betas)
    mean_function = jnp.dot(X, scaled_betas)
    numpyro.sample("Y", dist.Normal(mean_function, 0.05), obs=Y) 
```

martinjankowiak:
@fehiepsi thanks nice catch!

fehiepsi (approved):

PR #1146, opened 2021-09-06
add ScaledUnitLowerCholeskyTransform and change default AutoMultivariateNormal parameterization

martinjankowiak:
joint work with @fehiepsi 

instead of parameterizing a lower cholesky factor as an unconstrained strictly lower triangular piece and a positive diagonal we instead parameterize as

`L = unit_scale_tril @ scale_diag` 

where `unit_scale_tril` is lower triangular with ones along the diagonal and `scale_diag` is a positive diagonal matrix.

not surprisingly (consider e.g. the [analogous parameterization](https://github.com/pyro-ppl/numpyro/blob/f787ff946413338890fb67e941e5daef48b51d70/numpyro/infer/autoguide.py#L1009) in `AutoLowRankMultivariateNormal`) this seems to lead to consistently better performance (all results use `AutoMultivariateNormal`):

### logistic regression dataset with N=50k, D=28
- elbo: `-32 430.37`          [this PR]
- elbo: `-32 590.92`          [before this PR]

### logistic regression dataset with N=50k, D=18
- elbo: `-25 112.28`          [this PR]
- elbo: `-25 280.20`          [before this PR]

### logistic regression dataset with N=37k, D=15
- elbo: `-12 576.18`          [this PR]
- elbo: `-12 592.29`          [before this PR]

### sparse FITC GP classifier with N =  37k,  D = 14, and latent dim = 128 (inducing points)
- elbo:   `9 897.12`  [this PR]
- elbo:   `9 765.09 `   [before this PR]

### sparse FITC GP classifier with N =  37k,  D = 14, and latent dim = 64 (inducing points)
- elbo:  `9 925.11`    [this PR]
- elbo:  `9 902.51`    [before this PR]

### sparse FITC GP classifier with N =  37k,  D = 14, and latent dim = 32 (inducing points)
- elbo:  `9 896.23`     [this PR]
- elbo:  `9 884.37`     [before this PR]

### sparse FITC GP classifier with N =  37k,  D = 14, and latent dim = 16 (inducing points)
- elbo:    `9 320.48`  [this PR]
- elbo:    `9 311.96`  [before this PR]

### sparse FITC GP classifier with N =  100k,  D = 27, and latent dim = 64 (inducing points)
- elbo:    `-55 858.99`  [this PR]
- elbo:    `-55 947.08`  [before this PR]

### sparse FITC GP classifier with N =  100k,  D = 17, and latent dim = 64 (inducing points)
- elbo:    `-43 956.38`  [this PR]
- elbo:    ` -44 015.16`  [before this PR]

(note: inducing points have the same initialization for each comparison; not surprisingly i was getting noisy results before i made sure this was the case; also note the N=37k GP elbos are missing a log pi term that shifts them by large amounts)

there appears to be a clear winner but do we need more experiments? @fehiepsi what do you think?

(note: whether positivity is enforced via exponential or softplus tends to be much less important)

### DIFF
--- a/numpyro/distributions/transforms.py
+++ b/numpyro/distributions/transforms.py
@@ -681,6 +681,12 @@ class LowerCholeskyAffine(Transform):
 
 
 class LowerCholeskyTransform(Transform):
+    """
+    Transform a real vector to a lower triangular cholesky
+    factor, where the strictly lower triangular submatrix is
+    unconstrained and the diagonal is parameterized with an
+    exponential transform.
+    """
     domain = constraints.real_vector
     codomain = constraints.lower_cholesky
 
@@ -709,6 +715,17 @@ class LowerCholeskyTransform(Transform):
 
 
 class ScaledUnitLowerCholeskyTransform(LowerCholeskyTransform):
+    """
+    Like `LowerCholeskyTransform` this `Transform` transforms
+    a real vector to a lower triangular cholesky factor. However
+    it does so via a decomposition
+
+    :math:`y = loc + unit\_scale\_tril\ @\ scale\_diag\ @\ x`.
+
+    where :math:`unit\_scale\_tril\` has ones along the diagonal
+    and :math:`scale\_diag\` is a diagonal matrix with all positive
+    entries that is parameterized with an exponential transform.
+    """
     domain = constraints.real_vector
     codomain = constraints.scaled_unit_lower_cholesky
 
--- a/numpyro/distributions/transforms.py
+++ b/numpyro/distributions/transforms.py
@@ -710,13 +710,25 @@ class LowerCholeskyTransform(Transform):
 
 class ScaledUnitLowerCholeskyTransform(LowerCholeskyTransform):
     domain = constraints.real_vector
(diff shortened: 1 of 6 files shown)


numpyro/distributions/transforms.py
```diff
@@ -707,6 +715,38 @@ def inverse_shape(self, shape):
...
+        diag = jnp.exp(x[..., -n:])
+        return (z + jnp.identity(n)) * diag[..., None]
+
+    def _inverse(self, y):
+        diag = jnp.diagonal(y, axis1=-2, axis2=-1)
+        z = matrix_to_tril_vec(y / diag[..., None], diagonal=-1)
+        return jnp.concatenate([z, jnp.log(diag)], axis=-1)
+
+    def log_abs_det_jacobian(self, x, y, intermediates=None):
+        # the jacobian is diagonal, so logdet is the sum of diagonal `exp` transform
```
martinjankowiak, line 745: i guess we should remove this comment

fehiepsi (approved):

fehiepsi:
LGTM - you might need to double-check if using softplus is still good for your experiments.

PR #1147, opened 2021-09-06
Support autoguide subsample with deterministic sites

fehiepsi:


fritzo (approved):

numpyro/infer/autoguide.py
```diff
@@ -146,6 +146,10 @@ def _setup_prototype(self, *args, **kwargs):
                 model_args=args,
                 model_kwargs=kwargs,
             )
+        # We apply a fixed seed just in case postprocess_fn requires
+        # a random key to generate subsample indices. It does not matter
+        # because we only collect deterministic sites.
```
martinjankowiak, line 151: hmm what do you mean by 'does not matter'? what if the deterministic site depends on the subsample statement?

martinjankowiak: oh gotcha this is in `_setup_prototype`

PR #1162, opened 2021-09-24
Use black[jupyter] in notebooks

MarcoGorelli:
Currently, `black` is only applied to the Python files.

But since `21.8b0`, it can also be run on Jupyter Notebooks, so that's what I'm suggesting be done here


notebooks/source/bad_posterior_geometry.ipynb
```diff
@@ -49,7 +49,8 @@
     "from numpyro.diagnostics import summary\n",
     "\n",
     "from numpyro.infer import MCMC, NUTS\n",
-    "assert numpyro.__version__.startswith('0.7.2')\n",
+    "\n",
+    "assert numpyro.__version__.startswith(\"0.7.2\")\n",
```
fehiepsi, line 53: Hi @MarcoGorelli, could you verify if the [update_version](https://github.com/pyro-ppl/numpyro/blob/master/scripts/update_version.py) script is still work? You can modify [this line](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/version.py#L4) and run `python scripts/update_version.py`.

MarcoGorelli: Sure - that script isn't run in CI by the way, should it be?

MarcoGorelli: Would you be open to using nbQA for this? Then you could already just do
```console
$ nbqa scripts.update_version notebooks/
updating /home/marco/numpyro-dev/notebooks/source/time_series_forecasting.ipynb
```
and have
```diff
$ git diff
diff --git a/notebooks/source/time_series_forecasting.ipynb b/notebooks/source/time_series_forecasting.ipynb
index 4ecb365..01360c5 100644
--- a/notebooks/source/time_series_forecasting.ipynb
+++ b/notebooks/source/time_series_forecasting.ipynb
@@ -48,7 +48,7 @@
     "    set_matplotlib_formats(\"svg\")\n",
     "\n",
     "numpyro.set_host_device_count(4)\n",
-    "assert numpyro.__version__.startswith(\"0.5.0\")"
+    "assert numpyro.__version__.startswith(\"0.7.2\")"
    ]
   },
   {
```

fehiepsi: > that script isn't run in CI by the way, should it be?

We have been making PRs to bump the versions.

> using nbQA for this

Whoa, `nbqa` is neat! I guess it is somehow overkilled. Could you replace [those lines](https://github.com/pyro-ppl/numpyro/blob/master/scripts/update_version.py#L24-L27) with
```
pattern1 = re.compile('assert numpyro.__version__.startswith\\("[^"]*"\\)')
pattern2 = re.compile('assert numpyro.__version__.startswith\\(\\\\"[^"]*\\\\"\\)')
text1 = f"assert numpyro.__version__.startswith({new_version})"
text2 = text1.replace('"', '\\"')
```

MarcoGorelli: Yes, that also works, thanks for the suggestion!

I then tried running `time_series_forecasting.ipynb`, and it still executes fine

fehiepsi (approved):
This is great, thanks so much @MarcoGorelli!

PR #1168, opened 2021-09-25
Use parse_known_args instead of parse_args

### LINKED ISSUE #1103: Use parse_known_args across all examples
Currently, when executing the example script in jupyter notebook, we will get an error due to `argparse` could not recognize additional jupyter argument:
```
error: unrecognized arguments: -f /home/fehiepsi/.local/share/jupyter/runtime/kernel-b4867b12-6bd4-4ec2-9adf-a328899062d1.json
```
s (through the call `parser.parse_args()`). A solution is to use `parse_known_args` as suggested by @GMCobraz in [forum](https://forum.pyro.ai/t/error-in-gaussian-process-example/3142/3)

fehiepsi:
Resolves #1103

This allows users to run example scripts in jupyter notebook, without having to modify the argparse code.

### DIFF
--- a/examples/annotation.py
+++ b/examples/annotation.py
@@ -341,7 +341,7 @@ if __name__ == "__main__":
         ' "id" (item_difficulty), "lre" (logistic_random_effects)',
     )
     parser.add_argument("--device", default="cpu", type=str, help='use "cpu" or "gpu".')
-    args = parser.parse_args()
+    args, _ = parser.parse_known_args()
 
     numpyro.set_platform(args.device)
     numpyro.set_host_device_count(args.num_chains)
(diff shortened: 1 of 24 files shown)


fritzo (commented):
Hmm, I think it's safer to error loudly when an arg is not recognized. It's very easy to misspell an argument, and if `parse_known_args` is used, then that misspelling will be silently dropped. I foresee that causing lots of forum questions like "my model isn't converging even when I set ``--num-eopchs=100000``?!?"

Anyway, my two cents against defensive programming 😄

fehiepsi:
That's a good point! I don't know if there is a workaround in jupyter but let's close this pr.

PR #1169, opened 2021-09-25
Allow pickle autoguide

### LINKED ISSUE #1160: AutoGuides do not permit sampling using only learned parameters
## The Goal
Our setting requires that we separate sampling from a learned variational posterior some time after and in a separate script/program than the inference and *without having access to the training data* any more.

## The Problem
We use NumPyro's SVI algorithm with the `AutoDiagonalNormal` guide to infer a variational approximation of our model posterior. We then store the learned parameters of the guide.

At a later time, to sample from the learned posterior, we load the stored parameters, set up our model and the `AutoDiagonalNormal` guide and then call `Predictive` to sample from the data. However, rather than giving us samples from our approximated posterior, this results in one of several possible errors (depending on the internals of the model).

Example code:
```Python
import numpy as np
import jax
import numpyro
import numpyro.distributions as dists
from numpyro.infer import SVI, Predictive
from numpyro.infer.autoguide import AutoDiagonalNormal
from numpyro.primitives import sample, plate

d = 2
np.random.seed(12345)

x = np.random.poisson(1.0, size=(100,))

def model(x, N):
    rate = sample('param', dists.Gamma(1., 1.))
    batch_size = len(x) if x is not None else None
    with plate('batch', N, batch_size):
        sample('x', dists.Poisson(rate), obs=x)

# run inference
optim = numpyro.optim.Adam(1e-2)
svi = SVI(model, guide, optim, numpyro.infer.Trace_ELBO())
state = svi.init(jax.random.PRNGKey(0), x, len(x))
state = jax.lax.fori_loop(0, 10000, lambda i, state: svi.update(state, x, len(x))[0], state)

# get params
params = svi.get_params(state)

## sampling directly after inference works fine
samples = Predictive(model, guide=guide, params=params, num_samples=1, return_sites=['param', 'x'])(
    jax.random.PRNGKey(1), None, 1
)
print(samples)

## sampling with fresh guide does not work
guide = AutoDiagonalNormal(model) # fresh setup of guide to emulate script without inference

samples = Predictive(model, guide=guide, params=params, num_samples=1, return_sites=['param', 'x'])(
    jax.random.PRNGKey(1), None, 1
)
print(samples)
```
The second call to `Predictive` raises a `RuntimeError: MCMC only supports continuous sites or discrete sites with enumerate support, but got ExpandedDistribution.` (but really it has issues with the `Poisson`). Using other distributions will results in different errors.

## A bit of digging into this
We did a bit of digging and have come to the conclusion that this is caused by how the autoguides work: When they are first used, they will trace through the model to determine how to convert samples from an unconstrained domain to whatever is required in the model. It does this for all sample sites that are not observations. This is problematic in our case since we only want to sample from the learned posterior and therefore do not at any point fix observations in the model. The autoguide therefore treats the final sample sites corresponding to observations as something it needs to optimise (even though it will never have to do that).

## Ways forward?
We believe this is quite an oversight: Since the learned parameters fully describe the variational posterior mathematically, saving and reloading those should suffice to sample from the posterior in the API as well, without having to pass data into the model again (which is also a promise in the docstring of `AutoDiagonalNormal`': "The guide does not depend on the model's ``*args, **kwargs``.")

A workaround for us currently is to simply pass in some dummy data to initialize the guide, then go forward to sample from it using the learned parameters. However, this is somewhat cumbersome and unintuitive.

We think that ideally `Predictive` or the autoguide should be made smart enough to detect this case (maybe with an additional input for autoguide on which sites should be considered observations even when no data is given) and handle it.

At least it should be possible to store and load the state of the autoguide alongside the parameters. We tried to see whether this is possible at the moment. It didn't seem to be the case, but maybe we missed something.

The error messages for this should also be improved, as they don't really point to the actual underlying problem and are not consistent (see below for more examples).

What do you think? Or is there already a way to achieve our purpose using NumPyro that we have not thought of?

### Categorical distribution:
```
x = np.random.multinomial(1, [.5, .25, .125, .125], size=(100,)) * np.array([1,2,3,4])
x = np.sum(x, axis=-1)

# Categorical
def model(x, N):
    ps = sample('param', dists.Dirichlet(np.ones(4,)))
    batch_size = len(x) if x is not None else None
    with plate('batch', N, batch_size):
        sample('x', dists.Categorical(ps), obs=x)
```
This raises a `ImportError: Looking like you want to do inference for models with discrete latent variables. This is an experimental feature. You need to install funsor to be able to use this feature. It can be installed with pip install funsor.`

### Normal distribution:
```
x = np.random.normal(10, 3, size=(100, 2))
x[:,-1] *= -1

# Normal
def model(x, N):
    loc = sample('param', dists.Normal(0, 100), sample_shape=(d,))
    scale = np.array([2., 2.])

    batch_size = len(x) if x is not None else None
    with plate('batch', N, batch_size):
        sample('x', dists.Normal(loc, scale).to_event(1), obs=x)
```
This raises a `TypeError: reshape total size must be unchanged, got new_sizes (1, 2) for shape (0,).`.

fehiepsi:
Resolves #1160

TODO:
- [ ] allow pickling unravel_fn
- [ ] allow pickling postprocess_fn
- [ ] add tests

### DIFF
--- a/numpyro/infer/autoguide.py
+++ b/numpyro/infer/autoguide.py
@@ -4,6 +4,7 @@
 # Adapted from pyro.infer.autoguide
 from abc import ABC, abstractmethod
 from contextlib import ExitStack
+from functools import partial
 import warnings
 
 import numpy as np
@@ -450,6 +451,29 @@ class AutoDelta(AutoGuide):
         return locs
 
 
+def unravel_pytree(x_flat, treedef):
+    assert jnp.shape(x_flat) == 1
+    x = {}
+    curr_pos = next_pos = 0
+    for name, shape in treedef.items():
+        next_pos = curr_pos + np.prod(shape)
+        x[name] = x_flat[curr_pos:next_pos].reshape(shape)
+        curr_pos = next_pos
+    assert next_pos == x_flat.shape[0]
+    return x
+
+
+def ravel_pytree(x):
+    assert isinstance(x, dict)
+    treedef = {}
+    x_flat = []
+    for name, value in x.items():
+        treedef[name] = jnp.shape(value)
+        x_flat.append(value.reshape(-1))
+    x_flat = jnp.concatenate(x_flat) if x_flat else jnp.zeros((0,))
+    return x_flat, treedef
+
+
 class AutoContinuous(AutoGuide):
     """
     Base class for implementations of continuous-valued Automatic
@@ -474,7 +498,8 @@ class AutoContinuous(AutoGuide):
 
     def _setup_prototype(self, *args, **kwargs):
         super()._setup_prototype(*args, **kwargs)
-        self._init_latent, unpack_latent = ravel_pytree(self._init_locs)
+        self._init_latent, treedef = ravel_pytree(self._init_locs)
+        unpack_latent = partial(unravel_pytree, treedef=treedef)
         # this is to match the behavior of Pyro, where we can apply
         # unpack_latent for a batch of samples
         self._unpack_latent = UnpackTransform(unpack_latent)


numpyro/distributions/transforms.py
```diff
@@ -106,6 +106,16 @@ def inverse_shape(self, shape):
         """
         return shape
 
+    # Allow for pickle serialization of transforms.
```
fehiepsi, line 109: This follows [Pyro](https://github.com/pyro-ppl/pyro/blob/43dbb2e67331bd3aa6c73bb5483825432a8b7145/pyro/distributions/torch_patch.py#L39).

fehiepsi:
@lumip This supposes to fix your pickling issue. Could you double-check? Thanks!

numpyro/infer/autoguide.py
```diff
@@ -104,6 +104,11 @@ def _create_plates(self, *args, **kwargs):
                 )
         return self.plates
 
+    def __getstate__(self):
+        state = self.__dict__.copy()
+        state.pop("plates", None)
```
fehiepsi, line 109: `plates` is generated during execution, so we don't need to cache. Actually, we typically call the guide under `jit`, so `plates` will hold abstract values which cannot be pickle-able.

fritzo (approved):
Test looks good :+1:  It would be nice to be able to unpickle and use a guide without the model (via `guide()`), but I guess this is a separate issue.

fehiepsi:
Thanks for reviewing, @fritzo! You can use `guide` to sample posteriors by using either
```
guide.sample_posterior(rng_key, svi_result.params, sample_shape=(num_samples,))
```
or
```
Predictive(guide, params=svi_result.params, num_samples=num_samples)(rng_key)
```
We need it because we need `params`, `rng_key` to be able to call `guide()`.

Using `Predictive(model, guide, params)` is unnecessary. It is just a shortcut for
```
posterior_samples = guide.sample_posterior(...)
x = Predictive(model, posterior_samples, ...)(rng_key, ...)['x']
```
cc @lumip

test/test_pickle.py
```diff
@@ -59,3 +63,31 @@ def test_pickle_hmcecs():
...
+
+    predictive = Predictive(
+        poisson_regression,
+        guide=pickled_guide,
+        params=svi_result.params,
+        num_samples=1,
+        return_sites=["param", "x"],
+    )
+    samples = predictive(random.PRNGKey(1), None, 1)
+    assert set(samples.keys()) == {"param", "x"}
```
lumip, line 93: Maybe better to assert that `pickled_guide` returns the same samples as `guide` instead of just verifying that all sites are present?

lumip:
This seems to work well in some simple tests I ran. Unfortunately I can't right now verify that in our more elaborate settings but I may get around to that tomorrow.

Unfortunately I'm not all too familiar with the code that was changed, so I cannot review those changes in detail.

fehiepsi:
Thanks for reviewing, @lumip! I think we can just remove model from a pickled guide (if the guide already has prototype trace). Let me address that. For now, you can just set guide.model = None or delete that attribute before pickling.

PR #1171, opened 2021-09-26
WIP: Initial port of check_model_guide_match to numpyro/util.py

### LINKED ISSUE #1119: Check for shape consistency between model and guide
In SVI latent random variables sampled from the guide completely mask those sampled in the model during inference. However, nothing prevents us from specifying different shapes for such a sample site in model and guide respectively. This makes it easy to introduce confusing bugs when code in the model expects a certain shape that is different than what guide provides. This can easily happen if the model needs an adaptation but the user forgets to reflects that in the guide.

The problem then is not as much that this results in an error, but where that error occurs: In the code following the sampling of the RV in the model (when the actual error might have been sampling a wrong shape in the guide). Furthermore, the model function itself is sane, i.e., if the user searches for the error there, they will not find anything, thus leading to much confusion.

Long story short, I think it would be beneficial to check that shapes of sample sites referring to the same random variable in guide and model are consistent and at least issue a warning (if not outright raise an error) if this is not the case, as that would make it much easier to spot this type of error. Thoughts?

maw501:
Hey @fehiepsi, sorry for the delay, I have now started to take a look at this. I think I'm broadly okay with what is going on but am unsure about how to integrate it into the wider NumPyro library. I'd love to get your feedback and apologise in advance if I've done anything stupid, I'm still getting to grips with everything :grimacing:. 

### Summary

This will close #1119. Port of [check_model_guide_match](https://github.com/pyro-ppl/pyro/blob/7e293c5627458948db100c6a12d1e566f9f45327/pyro/util.py#L238) from Pyro to check consistency between model and guide RV shapes.

No tests have been added yet but will be once we get a bit closer to agreeing the changes. 

### Changes vs. Pyro [check_model_guide_match](https://github.com/pyro-ppl/pyro/blob/7e293c5627458948db100c6a12d1e566f9f45327/pyro/util.py#L238)

1. Removed `_Subsample` conditional statements as I couldn't see this referenced anywhere in NumPyro. E.g. removing the:

```python
if type(site["fn"]).__name__ != "_Subsample"
```

statements when getting the `model_vars`, `guide_vars` and `enum_vars`.

2. Removed `_enumerate_dim` conditional statements as I couldn't see this referenced anywhere in NumPyro. So the:

```python
if site["infer"].get("_enumerate_dim
```

statement was removed when getting `enum_vars`.

3. Obtaining the `model_shape` and `guide_shape` doesn't pass `*args` and `**kwargs` to `.shape()` [as is the case in Pyro](https://github.com/pyro-ppl/pyro/blob/7e293c5627458948db100c6a12d1e566f9f45327/pyro/util.py#L316) as `.shape()` complains about `rng_key` not being a keyword argument. Instead `sample_shape` is the only thing passed to `.shape()`. So, for example, we do: 

```python
model_shape = model_site["fn"].shape(model_site["kwargs"]["sample_shape"])
```

instead. 

### Questions

**1. Where to add the calls to `check_model_guide_match` in NumPyro?** 

As a check I tried adding a call to `check_model_guide_match` in `numpyro/infer/elbo.py` and this seems to work as expected when the shapes don't match. 

However, I had to grab the `model_trace` again **before** the call to `log_density` (which is where it's currently returned from). If I don't grab the `model_trace` before the call to `log_density` then this function errors when the shapes cannot be broadcast and I assume the point of `check_model_guide_match` is to catch the error before this happens and raise something more informative. 

I don't know what to do here so any input is welcome, and on the wider point of where to add calls to `check_model_guide_match`. 

**2. What to do about the `max_plate_nesting` argument?** 

I don't think I understand how this is used (i.e. from where it would be passed to `check_model_guide_match`) at the moment but have left it as a keyword argument for now.

### DIFF
--- a/numpyro/infer/elbo.py
+++ b/numpyro/infer/elbo.py
@@ -13,6 +13,7 @@ from numpyro.distributions.kl import kl_divergence
 from numpyro.distributions.util import scale_and_mask
 from numpyro.handlers import replay, seed, substitute, trace
 from numpyro.infer.util import get_importance_trace, log_density
+from numpyro.util import check_model_guide_match
 
 
 class ELBO:
@@ -121,6 +122,8 @@ class Trace_ELBO(ELBO):
             }
             params.update(mutable_params)
             seeded_model = replay(seeded_model, guide_trace)
+            model_trace = trace(seeded_model).get_trace(*args, **kwargs)
+            check_model_guide_match(model_trace, guide_trace)
             model_log_density, model_trace = log_density(
                 seeded_model, args, kwargs, params
             )
--- a/numpyro/util.py
+++ b/numpyro/util.py
@@ -3,6 +3,7 @@
 
 from collections import OrderedDict
 from contextlib import contextmanager
+from itertools import zip_longest
 import os
 import random
 import re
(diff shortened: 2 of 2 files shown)


fehiepsi:
Hi @maw501, thank you for taking a stab at this. The changes that you made sound reasonable to me. Regarding `enum_vars`, those are discrete latent variables but currently, we don't have TraceEnum_ELBO support so you can just leave it `enum_vars = set()` and add a note to address it when TraceEnum_ELBO is available. 

> Where to add the calls to check_model_guide_match in NumPyro?

It is better to add this check after you have `model_trace` and `guide_trace`. Making another trace will be costly. You can add a warning [here](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/handlers.py#L213) to early inform that something might be wrong. Something like
```
if msg["type"] == "sample":
    new_shape = jnp.shape(msg["value"])
    old_shape = msg["fn"].shape(msg["kwargs"]["sample_shape"])
    try:
        jax.lax.broadcast_shapes(new_shape, old_shape)
    except:
        warnings.warn(...)  # two shapes are not compatible.
```
I'm not sure how useful it is and what's a proper check there though. :(

> What to do about the max_plate_nesting argument?

Please just ignore it. We don't use it in NumPyro.

maw501:
Thanks @fehiepsi.

I can definitely add the snippet you suggest into `replay` but all this means is that a warning will be raised and then the error will still happen inside the call to `log_density` [here](https://github.com/pyro-ppl/numpyro/blob/632a13ad37a5fb9c47664fde3659161697abd6d4/numpyro/infer/elbo.py#L124) i.e. we'll never reach the call to `check_model_guide_match` to actually perform the check we want to.

It seems like the problem here is coming the fact that `log_density` makes the trace. I can think of two other solutions, both of which involve modifying `log_density`: 

1. Allow `log_density` to optionally accept the trace object and if `None` is passed it grabs it itself like it currently does? At the moment `log_density` is really doing two things by setting up the trace and then calculating the log density and it would be cleaner IMO to separate this out. 
2. Indeed, as far as I can tell `model`, `model_args`, `model_kwargs` and `params` are all passed to `log_density` just to construct `model_trace` - we could also just change `log_density` to solely accept a trace object? Then all `log_density` is doing is calculating the log density.

Thanks for your input!

fehiepsi:
I see. But I would like to argue that the errors typically happen during tracing, rather than computing log_density.

I guess you want to raise errors before some log_prob error might happen? If so, you can add a check inside `log_density` at [this line](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/infer/util.py#L63). You can check if the `value.shape` and `site["fn"].shape()` are compatible (i.e. if they are broadcastable), then raise an error if that happens. I guess this is the best solution.

fehiepsi (commented):
Thank you @maw501! Could you add some tests for `check_model_guide_match` in [this file](https://github.com/pyro-ppl/numpyro/blob/master/test/test_util.py)? I believe you will know a lot more of NumPyro by playing with examples, that violate the checks that you added. You can create a violated pair of (model, guide), run SVI with Trace_ELBO loss, and catch the warnings.

maw501:
I _think_ I've addressed most of the issues now but there are quite a few tests failing (see image). 

I'm going to work through them but am posting here for visbility. I think the main issue is other tests using bare `float`s and not `jnp.array`s which we require now in order to be able to check shapes. Also, for `tfp` distributions I think a call to `.shape()` doesn't return a `tuple` but a `TensorShape` object so I've fixed this. 

The rest I will look into.

image

fehiepsi:
Oh, from the CI log, I just notice that `vae` example is failing. That shows that the checks that you added are super helpful. You can modify the model there with
```
def model(batch, hidden_dim=400, z_dim=100):
    batch = jnp.reshape(batch, (batch.shape[0], -1))
    batch_dim, out_dim = jnp.shape(batch)
    decode = numpyro.module("decoder", decoder(hidden_dim, out_dim), (batch_dim, z_dim))
    with numpyro.plate("batch", batch_dim):
        z = numpyro.sample("z", dist.Normal(0, 1).expand([z_dim]).to_event(1))
        img_loc = decode(z)
        return numpyro.sample("obs", dist.Bernoulli(img_loc), obs=batch)


def guide(batch, hidden_dim=400, z_dim=100):
    batch = jnp.reshape(batch, (batch.shape[0], -1))
    batch_dim, out_dim = jnp.shape(batch)
    encode = numpyro.module("encoder", encoder(hidden_dim, z_dim), (batch_dim, out_dim))
    z_loc, z_std = encode(batch)
    with numpyro.plate("batch", batch_dim):
        return numpyro.sample("z", dist.Normal(z_loc, z_std).to_event(1))
```

maw501:
> Oh, from the CI log, I just notice that `vae` example is failing. That shows that the checks that you added are super helpful. You can modify the model there with
> 
> ```
> def model(batch, hidden_dim=400, z_dim=100):
>     batch = jnp.reshape(batch, (batch.shape[0], -1))
>     batch_dim, out_dim = jnp.shape(batch)
>     decode = numpyro.module("decoder", decoder(hidden_dim, out_dim), (batch_dim, z_dim))
>     with numpyro.plate("batch", batch_dim):
>         z = numpyro.sample("z", dist.Normal(0, 1).expand([z_dim]).to_event(1))
>         img_loc = decode(z)
>         return numpyro.sample("obs", dist.Bernoulli(img_loc), obs=batch)

I don't think we want the model return statement inside `plate`?

fehiepsi:
> I don't think we want the model return statement inside plate?

We observed a batch of images, so that `obs` should be inside a plate. This type of model only involves local random variables (see [here](https://pyro.ai/examples/svi_part_ii.html#Subsampling-when-there-are-only-local-random-variables)). :)

Issue #969, opened 2021-03-22
Truncated Gamma


quattro:
It would be great to have a [truncated gamma distribution](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3898536/) implemented in NumPyro in order to cover lower-bounded variances (or upper bounded precisions) in probabilistic programs. 

I'm happy to code this up and issue a pull-req after some testing.

A nice example would be in Mendelian Randomization from summary statistics (ie two-stage least squares). Residual variance greater than 1 is consistent with heterogeneity across studies, but inferred variance should be bounded below by 1.

fehiepsi:
@quattro I didn't use Truncated Gamma previously but I think this would be nice to have. FYI, we have [TruncatedDistribution](https://github.com/pyro-ppl/numpyro/blob/7493e35ed921108d884a88f58ab92ec5565f845d/numpyro/distributions/continuous.py#L1335) API with base distributions can be Cauchy, Laplace, Logistic, Normal, or StudentT. Probably it would be cleaner to follow that setting and have LeftTruncatedGamma, RightTruncatedGamma, TwoSidedTruncatedGamma implemented, then dispatching `TruncatedGamma` to the corresponding class (based on `low=None` or `high=None`).

quattro:
Great, thanks for the advice @fehiepsi. I'll likely be busy with some other things for a little while but plan on coming back to this soon.

quattro:
Hi @fehiepsi , I made some progress on this, but stopped at the sampling implementation. I think there are two paths forward, and would appreciate your thoughts on the matter:

1. Use a `uniform` -> `invCDF(gamma)` sampling based approach
2. Use rejection sampling

In terms of implementation:
1. Requires `gammaincinv` in order to compute the quantiles of the `gamma` distribution, which is currently not supported by JAX (as you pointed out here: https://github.com/google/jax/issues/5350). It looks like TFP has `gammaincinv` implemented (along with gradients), but admittedly, I don't have the bandwidth to initiate a PR and port it over to JAX.
2. There are a few papers describing sampling approaches to either the [Left/Right truncated Gamma](https://link.springer.com/article/10.1023/A:1018534102043), or the [more general case using latent variables](https://www.jstor.org/stable/1391008), that might be worthwhile investigating.

I'm hoping to move forward with this soon, as it looks like there is other support for a truncated gamma distribution, as mentioned in this thread: https://github.com/google/jax/issues/552 .

[...]

PR #1187, opened 2021-10-12
Truncated gamma

quattro:
PR for issue #969 . Contains initial implementation that performs uniform sampling + inverse CDF of Left/Right/Doubly truncated Gamma. Relies on tensorflow functionality for igammainv function, which is not yet implemented at the lax/jax level (see https://github.com/google/jax/issues/5350).

There is a test that fails, but it is not clear to me if this is purely a numerical issue with the uniform + iCDF sampling, or a larger issue that I missed at the time I implemented things.

### DIFF
--- a/numpyro/distributions/__init__.py
+++ b/numpyro/distributions/__init__.py
@@ -81,15 +81,15 @@ from numpyro.distributions.mixtures import MixtureSameFamily
 from numpyro.distributions.transforms import biject_to
 from numpyro.distributions.truncated import (
     LeftTruncatedDistribution,
+    LeftTruncatedGamma,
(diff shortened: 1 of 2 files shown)


fehiepsi:
@quattro Looking the the PR is is the good shape - just have small comments above. Any chance we can have this in the next numpyro release?

quattro:
Will try my best. Should have some time closer to Thanksgiving holidays, does that fall before next release schedule?

fehiepsi:
Absolutely, there is no plan for the release date yet. Thank you!

disadone:
Will we have this feature in the future?

PR #1194, opened 2021-10-20
Option to use state of previous SVI run as starting state

bjeffrey92:
Edited the method `SVI.run()` to allow the user to supply the state object of a previous SVI run which will be used as the starting state rather than calling `SVI.init()` at the start of every call to `SVI.run()` (this is still the default behaviour).

Example usage:
```
svi = SVI(model, guide, optimizer, loss=Trace_ELBO())
svi_result = svi.run(random.PRNGKey(0), 2000, data) 

# upon inspection of svi_result the user decides that the model has not converged

# continue from the end of the previous svi run rather than beginning again from iteration 0
svi_result = svi.run(random.PRNGKey(0), 2000, data, svi_initial_state=svi_result.state)
```

This is analogous to passing `mcmc.last_state` to the `init_params` argument of `mcmc.run()`

### DIFF
--- a/numpyro/infer/svi.py
+++ b/numpyro/infer/svi.py
@@ -309,10 +309,10 @@ class SVI(object):
             final state of previous SVI run. Usage::
 
                 svi = SVI(model, guide, optimizer, loss=Trace_ELBO())
-                svi_result = svi.run(random.PRNGKey(0), 2000, data) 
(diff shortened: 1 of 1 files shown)


fehiepsi:
Sounds reasonable to me. Could you help me add your usage case to the docs? And maybe change the name to `init_state`?

PR #1200, opened 2021-10-26
Add multivariate student-t distribution

### LINKED ISSUE #1190: MultivariateStudentT for numpyro
Port **https://docs.pyro.ai/en/stable/distributions.html#multivariatestudentt** to numpyro.

maw501:
This will close #1190. 

There are a few nuances where the treatment is different to SciPy, highlighted below. All feedback welcome. 🙏 

Tests are passing locally.

### Differences to SciPy

Passing an arbitrary tuple as `size` to the `.rvs` method doesn't work for as per MVN in SciPy. 

E.g
 
```
mvn = stats.multivariate_normal(mean=np.ones(2), cov=np.eye(2))
mvn.rvs(size=(2, 5))    # this works
mvn.rvs(size=())        # this works
```
but 

```
mvt = stats.multivariate_t(loc=np.ones(2), shape=np.eye(2), df=6)
mvt.rvs(size=(2, 5))    # fails with ValueError
mvn.rvs(size=())        # fails with IndexError
```

This is the reason for the extra logic in `test_dist_shape`.

### TODO

- [x] Check batched `df` is supported, I _think_ this is the case in Pyro.
- [x] Add to Sphinx docs

### Nice to have

- [x] Add extra check [here](https://github.com/pyro-ppl/numpyro/blob/bc47ff025b947fba530bec7e428eea66c102a459/test/test_distributions.py#L679) as per Multivariate Normal?
- [x] Skip tests if `scipy.__version__ < "1.6.0"`


fehiepsi:
Looking like scipy 1.6 requires python 3.7. `jax` [requires python 3.7](https://github.com/google/jax/blob/d403b6d03c95ca4e61078478a9f70d8f0ebb0c8f/setup.py#L40) so please feel free to pump python versions in [ci](https://github.com/pyro-ppl/numpyro/blob/master/.github/workflows/ci.yml) file.

maw501:
Thanks @fehiepsi - is the docs build failing related to #1204 by any chance?

fehiepsi (commented):
Sorry @maw501, I just recognize that CI has been testing on Python 3.6 with an old version of jax (the latest version that supports python 3.6). When you pump the version to 3.7, further issues happen. Currently, the last tfp release is not compatible with the last jax release (due to some deprecation in jax), so could you help me changes `tensorflor_probability` requirements in [setup](https://github.com/pyro-ppl/numpyro/blob/master/setup.py) and [requirements](https://github.com/pyro-ppl/numpyro/blob/master/docs/requirements.txt) files to
```
tfp-nightly  # TODO: change to tensorflow_probability when it is stable
```
By the way, could you also add MVT to [sphinx](https://github.com/pyro-ppl/numpyro/blob/master/setup.py) file? Thanks!

fehiepsi:
Hi @maw501, sorry for the trouble. I made a separate PR #1209 to fix some issues. Please feel free to merge it if you want to address all failing tests in your CL.

maw501:
Thanks @fehiepsi - I've already addressed some of them but will bring across your changes as you've resolved 2 I hadn't got to.

numpyro/distributions/gof.py
```diff
@@ -186,7 +186,7 @@ def density_goodness_of_fit(samples, probs, plot=False):
     if len(samples) <= 100:
         raise InvalidTest("imprecision; use more samples")
 
-    index = np.argsort(samples, 0, kind="quicksort")
+    index = np.argsort(samples, 0, kind="stable")
```
fehiepsi, line 189: This seems to be unnecessary. I think you can ping scipy version to 1.6.3. Newer versions seem to have bug (see https://github.com/scipy/scipy/issues/14953).

setup.py
```diff
@@ -52,7 +52,7 @@
             "isort>=5.0",
             "pytest>=4.1",
             "pyro-api>=0.1.1",
-            "scipy>=1.1",
+            "scipy>=1.6",
```
fehiepsi, line 55: I think you can change this to `scipy>=1.6<1.7`. Newer versions seem to have bug (see https://github.com/scipy/scipy/issues/14953).

numpyro/distributions/gof.py
fehiepsi: To suppress a new deprecation warning in CI, could you help me change `np.int` in line 148 to `int`. Thanks!

fehiepsi:
Not sure what happens but it seems that the tests have been greatly slowed down when we pump the version (e.g. test-modeling took 22m [previously](https://github.com/pyro-ppl/numpyro/runs/4061200271?check_suite_focus=true) but now took 30m). :D Seems like a performance regression in jax to me. But it is a separate issue. I'm also seeing some flaky errors like
```
    def bernoulli(key: KeyArray,
                  p: RealArray = np.float32(0.5),
                  shape: Optional[Union[Sequence[int], NamedShape]] = None) -> jnp.ndarray:
      """Sample Bernoulli random values with given shape and mean.
    
      Args:
        key: a PRNG key used as the random key.
        p: optional, a float or array of floats for the mean of the random
          variables. Must be broadcast-compatible with ``shape``. Default 0.5.
        shape: optional, a tuple of nonnegative integers representing the result
          shape. Must be broadcast-compatible with ``p.shape``. The default (None)
          produces a result shape equal to ``p.shape``.
    
      Returns:
        A random array with boolean dtype and shape given by ``shape`` if ``shape``
        is not None, or else ``p.shape``.
      """
      key, _ = _check_prng_key(key)
      dtype = dtypes.canonicalize_dtype(lax.dtype(p))
      if shape is not None:
        shape = core.as_named_shape(shape)
      if not jnp.issubdtype(dtype, np.floating):
        msg = "bernoulli probability `p` must have a floating dtype, got {}."
        raise TypeError(msg.format(dtype))
      p = lax.convert_element_type(p, dtype)
>     return _bernoulli(key, p, shape)  # type: ignore
E     ValueError: static arguments should be comparable using __eq__.The following error was raised during a call to '_bernoulli' when comparing two objects of types <class 'NoneType'> and <class 'jax.core.NamedShape'>. The error was:
E     TypeError: NamedShape doesn't support comparisons with <class 'NoneType'>
```
in the last CI result of this PR. I can't reproduce it locally so but it's worth reaching out to jax devs regarding this issue if it happens again.

fehiepsi (dismissed):
Look great to me! Also thanks for addressing the tricky versioning issue, @maw501! Let me know if you want to address other TODOs in the PR description. Btw, could you help me suppress another CI warning at [this line](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/examples/datasets.py#L235) (changing `np.long` to `int`)?

maw501:
Nice @fehiepsi - thank you for helping out with getting the version issues resolved! 

Re. the remaining TODOs, I think they are all nice to have except perhaps the first as currently using a batched `df` doesn't return the correct shape (i.e. `batch_shape`) from `log_prob`. 

Test case which fails:

```python
T(
    dist.MultivariateStudentT,
    jnp.arange(20, 26, dtype=jnp.float32).reshape((3, 2)),
    jnp.ones(2),
    jnp.array([[1.0, 0.0], [0.5, 1.0]]),
),
```
where the arguments are `df`, `loc` and `scale_tril`.

numpyro/distributions/continuous.py
```diff
@@ -1044,6 +1044,109 @@ def infer_shapes(
...
+        loc=0.0,
+        scale_tril=None,
+        validate_args=None,
+    ):
+        if jnp.ndim(loc) == 0:
+            (loc,) = promote_shapes(loc, shape=(1,))
+        batch_shape = lax.broadcast_shapes(
+            jnp.shape(df), jnp.shape(loc)[:-1], jnp.shape(scale_tril)[:-2]
+        )
+        (self.df,) = promote_shapes(jnp.array(df), shape=batch_shape)
```
fehiepsi, line 1068: I think we don't need to use `jnp.array` here (just to be more consistent with other distributions).

numpyro/distributions/continuous.py
```diff
@@ -1044,6 +1044,109 @@ def infer_shapes(
...
+    def log_prob(self, value):
+        n = self.scale_tril.shape[-1]
+        sample_shape = value.shape[:-2] if self.batch_shape != () else value.shape[:-1]
+        a_shape = sample_shape + self.batch_shape + (n, n)
+        b_shape = sample_shape + self.batch_shape + (n,)
+        y = solve_triangular(
+            jnp.broadcast_to(self.scale_tril, a_shape),
+            jnp.broadcast_to(value - self.loc, b_shape),
+            lower=True,
+        )
```
fehiepsi, line 1104: You can use `M = _batch_mahalanobis(self.scale_tril, value - self.loc)` for the term `jnp.power(y, 2).sum(-1)` below. It is more scalable and might fix the shape errors that you faced. The logic for `sample_shape` here is a bit confusing to me (e.g. why sample shape is changing when batch_shape is non-trivial,...)

numpyro/distributions/continuous.py
```diff
@@ -1044,6 +1044,109 @@ def infer_shapes(
...
+        identity = jnp.broadcast_to(
+            jnp.eye(self.scale_tril.shape[-1]), self.scale_tril.shape
+        )
+        return cho_solve((self.scale_tril, True), identity)
+
+    @property
+    def mean(self):
+        # for df <= 1. should be jnp.nan (keeping jnp.inf for consistency with scipy)
+        return jnp.broadcast_to(
+            jnp.where(self.df[..., jnp.newaxis] <= 1, jnp.inf, self.loc),
```
fehiepsi, line 1131: Use `jnp.expand_dims(self.df, -1)` for a scalar df.

numpyro/distributions/continuous.py
```diff
@@ -1044,6 +1044,109 @@ def infer_shapes(
...
+            self.shape(),
+        )
+
+    @property
+    def variance(self):
+        df = self.df[..., jnp.newaxis]
```
fehiepsi, line 1137: ditto: `jnp.expand_dims(self.df, -1)`

numpyro/distributions/continuous.py
maw501: 🤯 well, I wish I'd known about this function sooner.

fehiepsi (approved):

maw501:
@fehiepsi I added a final check in so I think we are fine to merge now unless we wanted to add in a check for the SciPy version? I don't think we could use the `@pytest.mark.skipif` decorator actually to check this as it would need to be on a paramaterized function, e.g. `test_log_prob` which I don't think is supported?

test/test_distributions.py
```diff
@@ -67,6 +67,13 @@ def _mvn_to_scipy(loc, cov, prec, tril):
     return osp.multivariate_normal(mean=mean, cov=cov)
 
 
+def _multivariate_t_to_scipy(df, loc, tril):
```
fehiepsi, line 70: I think you can do:
```
if scipy.__version__ < "1.6.0":
    pytest.skip("Multivariate Student-T distribution is not available in scipy < 1.6")
```

fehiepsi (approved):

PR #1215, opened 2021-11-05
Add link to render_model() link in tutorial

fritzo:
not sure if i got the syntax right for numpyro's notebook to html renderer


fehiepsi (approved):
Thanks!

PR #1216, opened 2021-11-05
Add example of jax.lax.scan in AR2 model

MarcoGorelli:
Hello,

Following up on [this discussion](https://forum.pyro.ai/t/lax-scan-to-implement-arma-1-1/3203), I'm just adding a simple model (AR2) which serves as an easy example of `jax.lax.scan`

What's not clear to me, though, is - what's the process for generating the thumbnails in the examples? For some of the others I see e.g. `gp_plot.pdf`, but then I don't see the `pdf` file anywhere.


examples/ar2.py
```diff
@@ -0,0 +1,130 @@
+# Copyright Contributors to the Pyro project.
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+Example: AR2 process
+====================
+
```
fehiepsi, line 7: Maybe add a recursion formula for AR2?

examples/ar2.py
```diff
@@ -0,0 +1,130 @@
...
+    ax.plot(t, y, color="blue", label="True labels")
+    # plot mean prediction
+    # note that we can't make predictions for the first two points,
+    # because they don't have lagged values to use prediction.
+    ax.plot(t[2:], mean_prediction, color="orange", label="Mean predictions")
+    ax.set(xlabel="time", ylabel="y", title="AR2 process")
+    ax.legend()
+
+    plt.savefig("ar2_plot.pdf")
+    fig.savefig("docs/source/_static/img/examples/ar2.png")
```
fehiepsi, line 115: It is better to check if the folder exist before writing.

fehiepsi (approved):
Looks great @MarcoGorelli! To generate the thumbnails, I have to run the example and change output to png (it is a bit inconvenient).

MarcoGorelli:
Hey @fehiepsi , thanks for your review!

I've updated, here's what the top of the page looks like now:

Screenshot 2021-11-05 at 21 29 23

(also fixed some unrelated typos in a separate commit)

examples/ar2.py
```diff
@@ -0,0 +1,128 @@
...
+    mean_prediction = samples["mu"].mean(axis=0)
+
+    # make plots
+    fig, ax = plt.subplots(figsize=(8, 6), constrained_layout=True)
+
+    # plot training data
+    ax.plot(t, y, color="blue", label="True values")
+    # plot mean prediction
+    # note that we can't make predictions for the first two points,
+    # because they don't have lagged values to use prediction.
```
martinjankowiak, line 108: nit: use for?

martinjankowiak (commented):
looks great!

the `unroll_loop=False` code path is never tested. is it worth making this an arg and adding a corresponding entry to test_examples? or otherwise have `main` run both code paths for a timing comparison (XXX seconds versus YYY seconds)?

MarcoGorelli:
Thanks @martinjankowiak for your review! That's a good point - I've added this argument, and added it to the tests. As far as I can tell, the output isn't shown in the examples from the docs, so I've just made a note of the speed improvement in the introductory paragraph

fehiepsi (approved):

fehiepsi (commented):
Also thanks for fixing many typos, @MarcoGorelli!

PR #1220, opened 2021-11-10
Fixup Bayesian Hierarchical Stacking notebook

MarcoGorelli:
Hey,

Fixing up a few things:

- the `assert numpyro.__version__` was still showing an old version;
- `set_matplotlib_formats` was undefined; `w` is assigned to but unused; some unused imports (found this by running `flake8`)
- the stacking method being compared against in the paper is more involved than I'd made it out to be. As far as I can tell, they are comparing with the method described in https://arxiv.org/pdf/1704.02030.pdf , which is what's implemented in `arviz.compare`, so I've used that here

PR #1221, opened 2021-11-11
Bump JAX, update imports

MarcoGorelli:
xref https://github.com/pyro-ppl/numpyro/pull/1220#issuecomment-965978034 , updating as per latest JAX version

### DIFF
--- a/numpyro/infer/autoguide.py
+++ b/numpyro/infer/autoguide.py
@@ -8,12 +8,13 @@ from functools import partial
 import warnings
 
 import numpy as np
-from packaging import version
 
 import jax
 from jax import grad, hessian, lax, random, tree_map
 
-if version.parse(jax.__version__) >= version.parse("0.2.25"):
+from numpyro.util import _versiontuple
+
+if _versiontuple(jax.__version__) >= (0, 2, 25):
     from jax.example_libraries import stax
 else:
     from jax.experimental import stax
--- a/numpyro/infer/autoguide.py
+++ b/numpyro/infer/autoguide.py
@@ -8,10 +8,16 @@ from functools import partial
 import warnings
 
 import numpy as np
+from packaging import version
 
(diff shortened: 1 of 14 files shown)


numpyro/infer/autoguide.py
```diff
@@ -11,7 +11,7 @@
 
 import jax
 from jax import grad, hessian, lax, random, tree_map
-from jax.experimental import stax
+from jax.example_libraries import stax
```
fehiepsi, line 14: Hi @MarcoGorelli, how about checking jax version before importing? (e.g. colab is using an old jax version so it would be nice to still support it). Kind of
```
if jax.__version__ >= "...":
    from jax.example_libraries import stax
else:
    from jax.experimental import stax
```

MarcoGorelli: Yes, that's better, thanks!

Should there be a separate CI job to check the oldest supported jax version, or is a local check I've done OK?

fehiepsi: I think we might skip it. Btw, can we avoid using extra package `packaging`? `jax` follows standard names so we can just compare the strings.

MarcoGorelli: Wouldn't comparing strings bring problems if they released, say, `0.2.100`?
```python
>>> '0.2.100' > '0.2.13'
False
```

And sure, the `versiontuple` answer looks good enough for this use-case, it seems very unlikely that JAX would suddenly change the format of their version numbers

fehiepsi (approved):
Thanks so much, @MarcoGorelli!

numpyro/infer/autoguide.py
fehiepsi: I think we can safely remove those version checks after a few jax releases or the next numpyro release but I agree the safe solution is better.

PR #1223, opened 2021-11-12
Fix links in bad posterior geometry

MarcoGorelli:
I was reading through this notebook and noticed that the links don't render. This is how they look on the [website](https://numpyro.readthedocs.io/en/latest/tutorials/bad_posterior_geometry.html)  :

---

image

---

With this PR:

---

image

---

It seems that the pattern
```
[`...`](...)
```
works fine in Jupyter, but not when the notebook is rendered by Sphinx?


fehiepsi (approved):

fehiepsi:
Thanks for the fix!!

PR #1225, opened 2021-11-13
Color divergences in Neal's Funnel example

MarcoGorelli:
Just a suggestion for the docs, sorry for not having discussed this first

I really like the plots from the Stan docs where the divergences are highlighted: https://mc-stan.org/users/documentation/case-studies/divergences_and_bias.html

So, am just suggesting the same be done here - it also serves as an example of how to get the diverging transitions from the `mcmc` object (missing from current examples / notebooks)

Current docs:
---

image

---

With this PR:
---

image

---


fehiepsi (approved):
Awesome, thanks @MarcoGorelli!

PR #1226, opened 2021-11-13
Fix autocontinuous not compatible with non-array init values

fehiepsi:
Fix a bug reported in [forum](https://forum.pyro.ai/t/svi-optimizer-scheduling/3542/16): AutoContinuous raises errors when using init_to_value with non-array init values.


OlaRonning (approved):
LGTM

fehiepsi:
Thanks, @OlaRonning!

PR #1229, opened 2021-11-18
Fix example from bayesian imputation notebook

MarcoGorelli:
I was trying to make sense of Bayesian imputation and noticed these syntax errors


fehiepsi (approved):
Thanks @MarcoGorelli!!!! <3

PR #1243, opened 2021-12-03
DiscreteUniform distribution

### LINKED ISSUE #1242: Feature request: DiscreteUniform distribution
[As suggested in the forum](https://forum.pyro.ai/t/impute-missing-counts-statistical-rethinking-15h7/3589), I'm opening this up as a feature request

If welcome, I'd be interested in working on it

MarcoGorelli:
closes #1242


fehiepsi (commented):
Thanks for the great contribution, @MarcoGorelli! I have a few comments. Could you also add an entry to [sphinx](https://github.com/pyro-ppl/numpyro/blob/master/docs/source/distributions.rst)?

MarcoGorelli:
Thanks @fehiepsi for your review! I've updated accordingly

The JAX tracer part is opaque to me, I've tried imitating what's done in `BinomialLogits` for that part but I don't (yet) understand it

numpyro/distributions/discrete.py
```diff
@@ -405,6 +405,67 @@ def Categorical(probs=None, logits=None, validate_args=None):
...
+
+    @property
+    def variance(self):
+        return ((self.high - self.low + 1) ** 2 - 1) / 12.0
+
+    def enumerate_support(self, expand=True):
+        if not not_jax_tracer(self.high) or not not_jax_tracer(self.low):
+            raise NotImplementedError("Both `low` and `high` must not be a JAX Tracer.")
+        low = np.amax(self.low)
+        if np.amin(low) != self.low:
```
fehiepsi, line 450: In case low is an array, we need to check `np.any(np.amax(self.low) != self.low)`. Similar for `high`.

fehiepsi (commented):
Thanks, @MarcoGorelli! I just have a minor comment. Regarding tracer logic, I think you got it right. We need to check if low and high are constants, and return an enumerated version.

fehiepsi (approved):

fehiepsi (approved):

Issue #1180, opened 2021-10-08
Add an info warning for HMC with enumeration


fehiepsi:
Quoted by @clgarciga in https://github.com/pyro-ppl/numpyro/issues/1174
> Also wondering if maybe there is an error or warning that could be thrown when users attempt to use Discrete latents in an HMC/NUTS model?

fehiepsi:
Hi @clgarciga, you can use this link https://pyro.ai/examples/enumeration.html.

clgarciga:
Okay, so, here's where I'm at with this: some MCMC kernels (e.g. HMC,NUTS) don't support discrete latents, and some do (e.g. DiscreteHMCGibbs, Mixed HMC). So I need to:
1. Check if the user is using MCMC
2. If yes, check if there are discrete latents in the model
3. If yes, check which kernel they are using
4. Check if that kernel supports discretes or not
5. If not, throw a warning (error?) that they need use a different kernel that supports discretes.
Does that make sense at all?

fehiepsi:
I think you can add an info warning after [this line](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/infer/util.py#L608). I'm not sure how to check properly for each kernel.

PR #1244, opened 2021-12-06
Raise warning for the automatic enumerate behavior

### LINKED ISSUE #1235: Raise warning for the automatic enumerate behavior
Following a thread on NumPyro forum about inference with discrete latent variable in hmm, here is the github issue as requested: 
In the hmm example, the discrete latent variable are marginalized out automatically. 
To avoid confusion, in future versions, the "automatic" behavior could be removed, and the users would provide infer={"enumerate": "parallel"} explicitly.
Thank you.

fehiepsi:
Currently, for HMC/NUTS, we perform automatic enumerate without asking users to specify `{"enumerate": "parallel"}`. This is different from Pyro, where the `parallel` method needs to be specified explicitly. It turns out that several users are confused by that difference. In this PR, we raise a FutureWarning for that automatic behavior, in preparation to remove it in the future.

Resolves #1235 and #1180.


fritzo (approved):
LGTM, seems like a reasonable plan to converge on behavior.

fehiepsi:

PR #1245, opened 2021-12-06
Add a warning for missing plate statements in the model

### LINKED ISSUE #1131: TraceGraph_ELBO bug shape error
Hi, I've been running into a bug related to data with multiple likelihoods. It seems that when computing downstream costs in the TraceGraph, multiple children have an empty `frozenset` frames on line 389 of `elbo.py` resulting in shape mismatch due to lookup bugs in the `MultiFrameTensor`.

I have an example program [here](https://gist.github.com/quattro/c3cdf41bf52a2334aa2dd2851388d254) to recreate the issue. The example I provided is a trivial example where the data likelihoods are completely independent (and thus could be optimized separately), but this error also occurs in the situation when the data likelihoods are conditionally independent given some latent variable.

Ideally these types of situations would have another dimension to reflect different data settings with similarly structured likelihoods (and possible 'resolve' this issue naturally), but the shapes change per data type and don't work cleanly in a single tensor structure.

```
  File "/Users/nicholas/Projects/VAE_SuSiE/g_test.py", line 180, in <module>
    sys.exit(main(sys.argv[1:]))
  File "/Users/nicholas/Projects/VAE_SuSiE/g_test.py", line 166, in main
    results = svi.run(
  File "/Users/nicholas/miniforge3/lib/python3.9/site-packages/numpyro/infer/svi.py", line 327, in run
    svi_state, loss = jit(body_fn)(svi_state, None)
  File "/Users/nicholas/miniforge3/lib/python3.9/site-packages/numpyro/infer/svi.py", line 316, in body_fn
    svi_state, loss = self.stable_update(svi_state, *args, **kwargs)
  File "/Users/nicholas/miniforge3/lib/python3.9/site-packages/numpyro/infer/svi.py", line 276, in stable_update
    (loss_val, mutable_state), optim_state = self.optim.eval_and_stable_update(
  File "/Users/nicholas/miniforge3/lib/python3.9/site-packages/numpyro/optim.py", line 93, in eval_and_stable_update
    (out, aux), grads = value_and_grad(fn, has_aux=True)(params)
  File "/Users/nicholas/miniforge3/lib/python3.9/site-packages/numpyro/infer/svi.py", line 59, in loss_fn
    elbo.loss(
  File "/Users/nicholas/miniforge3/lib/python3.9/site-packages/numpyro/infer/elbo.py", line 604, in loss
    return -single_particle_elbo(rng_key)
  File "/Users/nicholas/miniforge3/lib/python3.9/site-packages/numpyro/infer/elbo.py", line 579, in single_particle_elbo
    downstream_costs, _ = _compute_downstream_costs(
  File "/Users/nicholas/miniforge3/lib/python3.9/site-packages/numpyro/infer/elbo.py", line 483, in _compute_downstream_costs
    downstream_costs[node].add(*downstream_costs[child].items())
  File "/Users/nicholas/miniforge3/lib/python3.9/site-packages/numpyro/infer/elbo.py", line 389, in add
    self[frames] = self[frames] + value
  File "/Users/nicholas/Projects/jax/jax/_src/numpy/lax_numpy.py", line 5847, in deferring_binary_op
    return binary_op(self, other)
  File "/Users/nicholas/Projects/jax/jax/_src/numpy/lax_numpy.py", line 431, in fn
    return lax_fn(x1, x2) if x1.dtype != bool_ else bool_lax_fn(x1, x2)
TypeError: add got incompatible shapes for broadcasting: (2,), (4,).
```

### LINKED ISSUE #1230: A way to notify shapes going wrong in log_prob computation in ELBO?
We've recently run into a subtle (modelling) bug with broadcasting in the ELBO loss.
The offending part in our model boiled down to something like this:
```
import numpyro.distributions as dists

from numpyro import plate, sample
from numpyro.handlers import trace, seed

import numpy as np

N = 100
d = 7
logits = np.random.randn(N)
ys = np.random.binomial(1, .5, size=(N, 1))
 
def model(logits, ys):
    with plate("batch", len(ys)):
        sample("ys", dists.Bernoulli(logits), obs=ys)

# [some SVI inference code with Trace_ELBO here]
 ```
 
Note that while `logits` are of shape `(N,)`, our `ys` had shape `(N,1)` from our preprocessing script. As a consequence, the ELBO-internal call to `Bernoulli(logits).log_prob(ys)`  broadcasts over `ys` and returns a `(N, N)` matrix, over which the ELBO / `log_density` then sums. The program does not produce any error but our inference obviously fails.

Now, this is an oversight by us and not strictly a bug in numpyro (there could be scenarios where this behavior could be desired by a user, I guess (although I find it unlikely)). However, due the log_prob being abstracted away so nicely, it is very difficult to track this down when it is not desired, as in our case (we have been struggling to solve our poor model performance for a long time before we actually figured out that this is happening). I therefore wanted to bring this up and ideally have a discussion on whether there is a way to make this easier to spot if it happens.

One thought I had off the top of my head was to add an optional `expected_log_prob_shape` kwargs to `sample` where the user can specify what shape they would expect coming out of `log_prob` and the program failing if this is not the case. There might even be a way to do this without the explicit argument based on the distributions batch and even shapes maybe? I think another useful thing that could help a bit would be to actually check given `obs` for shape consistency with `sample_shape` (if both are provided to `sample`) as a sanity check for the user.

### PR DESCRIPTION
**Pull request by fehiepsi**
Resolves #1230 & resolves #1131

From discussions at #1230, I think it is better to always raise warnings for missing plate statements in the model.

### CODE REVIEW

File: test/infer/test_autoguide.py [hunk 1]
```diff
@@ -104,12 +106,12 @@ def body_fn(i, val):
     # Predictive can be instantiated from posterior samples...
     predictive = Predictive(model, posterior_samples=posterior_samples)
     predictive_samples = predictive(random.PRNGKey(1), None)
-    assert predictive_samples["obs"].shape == (1000, 2)
+    assert predictive_samples["obs"].shape == (1000, N, 2)
```
martinjankowiak, line 109: does this mean predictive was generating incorrect samples before?

fehiepsi: That's right. Previously we missed the plate around the likelihood `numpyro.sample("obs", dist.Bernoulli(f), obs=data)` so if data is None, the likelihood will have the same shape as `f`, which is `(2,)`.

martinjankowiak (approved):

Issue #1247, opened 2021-12-09
Move contrib modules to the main library


fehiepsi:
Currently, some contrib modules have been stable for a while. It would be convenient to move them directly to the main library. Some candidates and their corresponding target folders are

- [x] `indexing` -> `ops`
- [ ] `control_flow` -> `ops` (scan + cond)
- [ ] `module` -> `nn`?
- [ ] `render` -> `ops/inspect`
- [ ] `optim` -> `optim`

fehiepsi:
It seems to be better to leave the remaining submodule in contrib for a while.
+ For `control_flow`, we might need to introduce "named" scan (similar to `plate`) so that we can identify all the shapes of a variable. This might require revising the implementation a bit (hence might enable us to support nested scan or render time dimensions...).
+ For `random_flax_module` and `random_haiku_module`, their APIs are a bit different from each other. In addition, there are no such primitives in Pyro so it is not clear to me what's the best place to merge them. Maybe we can wait until we can officially support one of those nn frameworks.
+ For `optim`, it is just a convenient utility. No one will want to import it so we might leave it as-is and clean it up in the future if needed.
+ For `render`, we will move it to `ops.inspect` to be aligned with Pyro. Because users do not need to import this utility from contrib (we can import `numpyro.render` directly), it is fine to leave this module as-is right now and clean up in the future.

PR #1248, opened 2021-12-09
Add ProvenanceArray to infer relational structure in a model

### LINKED ISSUE #1217: Support ProvenanceArray to infer latent variable dependencies
It would be nice to have a version of [ProvenanceTensor](https://github.com/pyro-ppl/funsor/pull/543) to [infer dependency structure](https://github.com/pyro-ppl/pyro/pull/2959) of a numpyro model.

For now, it is not clear to me how to do the same thing in JAX. :(

### LINKED ISSUE #1241: Need help: NUTS does't converge
I observed my numpyro implementation gives very different results from a JAGS model (working really well using 1000 warmups). The main issue is the chain does not converge regardless of warmup numbers. Here are some notes:

* I implement the numpyro model wrong。 
* Maybe I need to initiate the chain but I did not.
* Sampling looks a bit weird when I implement an exponential function or a power law. Right before the end, the acc.prob usually drops from .8 to .3. 
* I try to increase the warmups samples to 50,000, still does not converge

My code:

```python 

import numpy as np
import numpyro as npyro
import arviz as az
import jax.numpy as jnp

from jax import random 
from scipy.stats import norm, binom 

def hMF( obs, eps=.001):

    ## Populational level distributions
    mu_alpha  = npyro.sample( 'mu_alpha', dist.Uniform( 0, 1))
    tau_alpha = npyro.sample( 'tau_alpha', dist.Gamma( eps, eps))
    mu_a      = npyro.sample( 'mu_a', dist.Uniform( 0, 1))
    tau_a     = npyro.sample( 'tau_a', dist.Gamma( eps, eps))
    mu_b      = npyro.sample( 'mu_b', dist.Uniform( 0, 1))
    tau_b     = npyro.sample( 'tau_b', dist.Gamma( eps, eps))

    ## Individual distributions
    alpha = npyro.sample( 'alpha', dist.TruncatedDistribution(
                    dist.Normal( jnp.broadcast_to( mu_alpha, [obs['n_subj'], 1]), tau_alpha), low=0, high=1))
    a     = npyro.sample( 'a', dist.TruncatedDistribution( 
                    dist.Normal( jnp.broadcast_to( mu_a, [obs['n_subj'], 1]), tau_a), low=0, high=1))
    b     = npyro.sample( 'b', dist.TruncatedDistribution( 
                    dist.Normal( jnp.broadcast_to( mu_b, [obs['n_subj'], 1]), tau_b), low=0, high=1))

    ## Each time steps
    sub_id = np.arange( obs['n_subj'])
    t_lags = np.arange( obs['T']).reshape([1,-1])
    theta = a[sub_id] + (1 - a[sub_id]) * b[sub_id] *\
        jnp.exp(-alpha[sub_id] * t_lags)
    
    return npyro.sample( 'recall', dist.Binomial( obs['n_items'], theta
                                ), obs=obs['recalls'])

def sim_hMF( seed=1234):

    ## Generate data
    rng = np.random.RandomState( seed)
    tlags = [ 0, 1, 5, 10, 20, 50]
    n_subj, T, n_items = 4, len(tlags), 20
    recalls = np.zeros( [ n_subj, T]) + np.nan 
    for sub_i in range(n_subj):
        a = .2*rng.rand()
        b = .1*rng.rand() + .9
        alpha = .3*rng.rand() + .1
        print( f'Sub {sub_i}: a={a:.2f}, b={b:.2f}, alpha={alpha:.2f}')
        for j, t in enumerate(tlags):
            p = a + (1-a) * b * np.exp(-alpha*t)
            recalls[ sub_i, j] = binom( n=n_items, p=p).rvs()
    obs = { 'n_subj': n_subj, 'n_items': 20, 'T': T,
            'recalls': recalls}

    ## Fix the random seed
    rng_key = random.PRNGKey(seed)

    ## Sampling 
    kernel = NUTS( model)
    posterior = MCMC( kernel, num_chains=1, 
                              num_samples=5000, 
                              num_warmup=10000,)
    posterior.run( rng_key, obs)
    samples = posterior.get_samples()
    
    ## Summarize the sampling results
    print( posterior.print_summary())
    az.plot_trace( az.from_numpyro( posterior), compact=True)
```

My az figures

image

fehiepsi:
Resolves #1217 and closes #1241 

Motivation: Previously, we rely on `grad` trick to decide dependency. That trick does not apply to discrete latent variables. In addition, when some numerical issues happen, the rendered model is wrong.

Here we follow @ordabayevy and @fritzo design for ProvenanceTensor in [funsor](https://github.com/pyro-ppl/funsor/pull/543) and [pyro](https://github.com/pyro-ppl/pyro/pull/2959) and adapt it for JAX system. Unlike pytorch, where provenance can be tracked by a computation (e.g. `a + b + c`), in JAX we need to create a function `lambda a, b, c: a + b + c` and use `eval_provenance` to get the provenance output of that function. The API is pretty much similar to [jax.eval_shape](https://jax.readthedocs.io/en/latest/jax.html#jax.eval_shape). It is surprising that the implementation is not too complicated. The trick is to introduce a "provenance" named dimension for each array and handle it whenever a JAX operator/primitive is performed on provenance arrays.

### Design decisions
+ Create a new submodule `ops` that mimics Pyro. This will be a useful place when we move some contrib modules to the main library in the future #1247
+ Move `contrib/control_flow/util.py` to `ops/pytree.py` because I need to use PytreeTrace to get an "abstract" trace (i.e. a trace where all arrays only contain shape and dtype information).
+ Add `plate` messenger to the PytreeTrace. As a result, we need to skip it in the previous codes in `control_flow` that use PytreeTrace.

### Tests
+ Current tests in `test_render_model`
+ Add simple tests for `provenance`
+ Rerun the rendering notebook
+ Correctly render the model in #1241

### TODO
- [x] port `get_dependencies` and its tests

Also fixes some numerical tests and small bugs (like `set(DeviceArray)` does not work anymore) in the new jax/jaxlib release.


numpyro/contrib/render.py
```diff
@@ -43,31 +46,43 @@ def model(data):
...
+        # Work around an issue where jax.eval_shape does not work
+        # for distribution output (e.g. the function `lambda: dist.Normal(0, 1)`)
+        # Here we will remove `fn` and store its name in the trace.
+        for name, site in trace.items():
+            if site["type"] == "sample":
+                site["fn_name"] = _get_dist_name(site.pop("fn"))
+        return PytreeTrace(trace)
+
+    # We use eval_shape to avoid any array computation.
+    trace = jax.eval_shape(get_trace).trace
```
fritzo, line 78: :+1: abstract interpretation FTW!

numpyro/contrib/render.py
```diff
@@ -43,31 +46,43 @@ def model(data):
...
         return type(fn).__name__
 
+    def get_trace():
+        # We use `init_to_sample` to get around ImproperUniform distribution,
+        # which does not have `sample` method.
+        subs_model = handlers.substitute(
+            handlers.seed(model, 0),
+            substitute_fn=init_to_sample,
+        )
+        trace = handlers.trace(subs_model).get_trace(*model_args, **model_kwargs)
```
fritzo, line 68: In Pyro we [use an effect handler](https://github.com/pyro-ppl/pyro/blob/b134ae8ec2380ca86046afcff28b337a0bac86c9/pyro/infer/inspect.py#L48) to replace the provenance of each sample value by the singleton `site["name"]`, thereby tracking only direct dependencies rather than transitive dependencies.  Can you point me to where in this logic you replace transitive dependencies by the new sample site name?

fritzo (commented):
Nice, this is much cleaner and cheaper!

Can you confirm that provenance tracking is covered by existing unit tests with many different model structures?

numpyro/contrib/render.py
```diff
@@ -111,45 +127,16 @@ def get_log_probs(sample, seed=0):
         }
 
     samples = {
-        name: site["value"]
+        name: ProvenanceArray(site["value"], frozenset({name}))
```
fehiepsi, line 130: @fritzo This is the place we define provenance arrays, then substitute into get_log_probs function defined above to get provenance outputs.

fehiepsi:
@fritzo I just ported `get_dependencies` and its tests to NumPyro. There should be no difference w.r.t. to Pyro implementation except that I used `eval_provenance` with `get_log_probs` function rather than `TraceProvenance` messenger as in Pyro.

fritzo (approved):
🎉  This is great @fehiepsi, thanks for porting this work to NumPyro!

fehiepsi:
Thanks for reviewing, @fritzo!!

fehiepsi:
Thanks for reviewing, @fritzo!!

PR #1252, opened 2021-12-11
Use multipledispatch for kl_registry

### LINKED ISSUE #762: Use multipledispatch for `kl_divergence`
As @fritzo's suggestion in https://github.com/pyro-ppl/numpyro/pull/748, we can use `multipledispatch` to simplify dispatching work in `kl_divergence` (and possibly in `biject_to` too?). We can do this when `funsor` becomes the main dependency of numpyro.

fehiepsi:
Resolves #762

My motivation is to detect ambiguity in (Delta, ExpandedDistribution) early (which happened in #1245). But it turns out we can simplify much code here. Thanks for suggesting, @fritzo!


fritzo (dismissed):
Great simplification!  The historical reason we didn't use multipledispatch was that PyTorch wanted to avoid adding dependencies; we ended up building minimal functionality from scratch.

fritzo (approved):

PR #1256, opened 2021-12-12
Fix CI freezing

fehiepsi:
Recently, CI is freezing at `test/contrib/test_tfp.py::test_mcmc_kernels` `SliceSampler`. I'm not sure what's going on so I want to disable it on CI (we don't have much support and AFAIK, no one is using those helper kernel classes with numpyro models). In addition, the test `test/infer/test_reparam.py::test_circular` is quite flaky for high moments, so I modified the test to just test for the first and second moments.

Other than the above, I also made the following changes:
+ there are many places that we invoke JAX computation globally (e.g. creating a device array `jnp.array([2., 3.])`). In JAX, when we invoke a computation, global settings will be fixed. That makes it not possible to change numerical precision in a particular test. So I changed `jnp.array` to `np.array` in many places.
+ tests for `SineBivariateVonMises` are slow, so I removed duplicated ones in CONTINUOUS list (because they already appeared in the DIRECTIONAL list) and disabled some of them on CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,7 +71,7 @@ jobs:
         pip freeze
     - name: Test with pytest
       run: |
-        CI=1 pytest -vs -k "not test_example" --durations=100 --ignore=test/infer/
+        CI=1 pytest -vs -k "not test_example" --durations=100 --ignore=test/infer/ --ignore=test/contrib/
 
 
   test-inference:
@@ -102,6 +102,7 @@ jobs:
       run: |
         pytest -vs --durations=20 test/infer/test_mcmc.py
         pytest -vs --durations=20 test/infer --ignore=test/infer/test_mcmc.py
+        pytest -vs --durations=20 test/contrib
     - name: Test x64
       run: |
         JAX_ENABLE_X64=1 pytest -vs test/infer/test_mcmc.py -k x64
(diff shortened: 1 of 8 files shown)


martinjankowiak (approved):

martinjankowiak:
> In JAX, when we invoke a computation, global settings will be fixed. That makes it not possible to change numerical precision in a particular test.

can you please elaborate?

OlaRonning (approved):
I can't comment directly in the code, but why is [`_ImproperWrapper`](https://github.com/pyro-ppl/numpyro/pull/1256/files#diff-5f4563907d8780f49ec203153cbb2790be2214b62626143751a65896969ab95cR158-R165) (test_distributions line 159-166) using `jnp.array`? Otherwise, looks good. 

Tomorrow, I'll find more efficient parameterizations for the remaining `SineBivariateVonMises` tests.

fehiepsi:
Thanks for reviewing @martinjankowiak @OlaRonning!

Regarding the change from `jnp.array` to `np.array`, consider the following program
```python
import jax.numpy as jnp

x = jnp.ones(3)
def f(x):
    import numpyro
    numpyro.enable_x64()
    y = jnp.ones(3)
    assert y.dtype == jnp.float32
```
The last assertation will pass because when we call `x = jnp.ones(3)` in the global scope, we can no longer switch to float64, i.e. `numpyro.enable_x64()` won't take effect. It is fine to use `jnp.foo` as long as that `jnp.foo` does not belong to the global scope.

PR #1258, opened 2021-12-15
Extending Predicitive to allow SteinVI

OlaRonning:
To use the `Predictive` class with SteinVI, the [`batch_shape`](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/infer/util.py#L707) must be promoted to include an extra (right) dimension for the particles. Consider the following linear regression model: 
``` python
def model(x, y):
  n, m = x.shape
  prec = sample("prec", 
                Gamma(1.0,0.1))  # One per particle 
  feat = sample("feat",
                 Normal(zeros(m),
                 1/prec))  # broadcast to (m,)
  with plate('data',  n):
    sample("y", Normal(x @ feat, 1), obs=y)
```
Currently, `Predictive(model, guide=AutoDelta(model), params=results.params, num_samples=1)` will make the `batch_shape=(1,)` so the [soft_vmaps](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/infer/util.py#L771-L773) `batch_ndim=1`. The `Normal` will try to promote the shapes `(1,), (m,), (num_particles)` which will fail for arbitrary choice of particles.

This PR introduces `num_particles` as a parameter to `Predictive` that expands `batch_shape` with `num_particles` as a new right most dimension, i.e.
``` python
batch_shape = (1,) * (self.batch_ndims - 1) + (
                self.num_samples,
                self.num_particles,
            )
```
An alternative (suggested by @fehiepsi) is to allow tuples of integers for `num_samples` instead. For SteinVI, we would then have `num_sample=(num_sample, num_particles)`. This solution is more general than the one presented above. We may need to check the site shapes in the prototype against the `num_samples`. I'll sketch the solution to be sure.

Once we settle on a solution I'll add some test cases. PR related to #833.

### DIFF
--- a/numpyro/infer/util.py
+++ b/numpyro/infer/util.py
@@ -2,7 +2,6 @@
 # SPDX-License-Identifier: Apache-2.0
 
 from collections import namedtuple
-from collections.abc import Iterable
 from contextlib import contextmanager
 from functools import partial, reduce
 import operator
--- a/numpyro/infer/util.py
+++ b/numpyro/infer/util.py
@@ -3,15 +3,14 @@
 
 from collections import namedtuple
 from contextlib import contextmanager
-from functools import partial, reduce
-import operator
-from typing import Union
+from functools import partial
+from typing import Callable, Dict, List, Optional
 import warnings
 
 import numpy as np
 
 import jax
-from jax import device_get, jacfwd, lax, random, value_and_grad
+from jax import device_get, jacfwd, lax, random, tree_flatten, value_and_grad
 from jax.flatten_util import ravel_pytree
 from jax.lax import broadcast_shapes
 import jax.numpy as jnp
@@ -801,16 +800,24 @@ class Predictive(object):
         Note that this requires ``funsor`` installation.
     :param bool parallel: whether to predict in parallel using JAX vectorized map :func:`jax.vmap`.
         Defaults to False.
-    :param batch_ndims: the number of batch dimensions in posterior samples. Some usages:
+    :param batch_ndims: the number of batch dimensions in posterior samples or parameters. If `None` defaults
+        to 0 if guide is set (i.e. not `None`) and 1 otherwise. Usages for batched posterior samples:
 
         + set `batch_ndims=0` to get prediction for 1 single sample
 
         + set `batch_ndims=1` to get prediction for `posterior_samples`
-          with shapes `(num_samples x ...)`
+          with shapes `(num_samples x ...)` (same as`batch_ndims=None` with `guide=None`)
 
         + set `batch_ndims=2` to get prediction for `posterior_samples`
           with shapes `(num_chains x N x ...)`. Note that if `num_samples`
           argument is not None, its value should be equal to `num_chains x N`.
+
+        Usages for batched parameters:
+
+        + set `batch_ndims=0` to get 1 sample from the guide and parameters (same as `batch_ndims=None` with guide)
+
+        + set `batch_ndims=1` to get predictions from a one dimensional batch of the guide and parameters
+          with shapes `(num_samples x batch_size x ...)`
 
     :return: dict of samples from the predictive distribution.
 
@@ -838,16 +845,16 @@ class Predictive(object):
 
     def __init__(
         self,
-        model,
-        posterior_samples=None,
+        model: Callable,
+        posterior_samples: Optional[Dict] = None,
         *,
-        guide=None,
-        params=None,
-        num_samples: Union[int, None] = None,
-        return_sites=None,
-        infer_discrete=False,
-        parallel=False,
-        batch_ndims=None,
+        guide: Optional[Callable] = None,
+        params: Optional[Dict] = None,
+        num_samples: Optional[int] = None,
+        return_sites: Optional[List[str]] = None,
+        infer_discrete: bool = False,
+        parallel: bool = False,
+        batch_ndims: Optional[int] = None,
     ):
         if posterior_samples is None and num_samples is None:
             raise ValueError(
@@ -905,15 +912,6 @@ class Predictive(object):
         self._batch_shape = batch_shape
 
     def _call_with_params(self, rng_key, params, args, kwargs):
-        """
-        Returns dict of samples from the predictive distribution. By default, only sample sites not
-        contained in `posterior_samples` are returned. This can be modified by changing the
-        `return_sites` keyword argument of this :class:`Predictive` instance.
-
-        :param jax.random.PRNGKey rng_key: random key to draw samples.
-        :param args: model arguments.
-        :param kwargs: model kwargs.
-        """
         posterior_samples = self.posterior_samples
         if self.guide is not None:
             rng_key, guide_rng_key = random.split(rng_key)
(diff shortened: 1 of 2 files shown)


fehiepsi:
Hi @OlaRonning, could you add some tests for my better understanding?

> The Normal will try to promote the shapes (1,), (m,), (num_particles) which will fail for arbitrary choice of particles.

Does this happen only for SteinVI? Maybe we can talk about a simple example first. Given a `model`, `guide` with a latent `x` with shape (3,), what's do you want to get by using Predictive? A collection of samples with shape (num_samples, num_particles, 3)? If so, another option is to use Predictive with `num_samples=(num_samples x num_particles)` then reshaping the output.

OlaRonning:
@fehiepsi I'll add some test cases. 

> Does this happen only for SteinVI?

The problem only occurs with SteinVI (I have not checked with `hmc_gibbs`)

> Maybe we can talk about a simple example first. Given a model, guide with a latent x with shape (3,), what's do you want to get by using Predictive?

For your example Predictive works as is and produces samples with shape `(num_samples, num_particles, 3)` as expected. 
The params will look like 
``` python
steinvi = SteinVI(model, AutoDelta(model), Adam(.1), Trace_ELBO(), RBFKernel(), num_particles=5)
result = steinvi.run(rng_key, 1000)
print({k: v.shape for k,v in result.params.items()}  # {'x_auto_loc': (5, 3)}
```

Linear regression is the simplest example I can think of where a problem occurs.
``` python 
  def model(x, y=None):
  latent = numpyro.sample('x', Normal(0, jnp.ones(3, )).to_event(1))
  with numpyro.plate('data', x.shape[0]):
    numpyro.sample('y', Normal(x @ latent, .1), obs=y)  # x.shape = (n_data, 3) latent.shape = (n_particles, 3) during _predictive
```

fehiepsi:
Oh I see - the issue here is we have a "batch" of guide and a batch of params, that will generate a batch of posterior samples. Our current api only allows to define the number of batch of samples for the posterior_samples argument, not for the guide.

I'm not sure why the code does not perform vmap over guide to get num_samples x num_particles posterior samples. IIUC you can leverage batch_ndims for this new behavior. If it is 0, guide and params just give 1 sample. If it is 1, guide and params give a batch of samples. This has the same senmatics as posterior_samples. I guess we can let batch_ndims=None by default where it takes 1 if posterior samples are provided and takes 0 if guide is provided.

test/infer/test_infer_util.py
```diff
@@ -96,6 +96,35 @@ def guide(data):
...
+    def model(x, y=None):
+        latent = numpyro.sample("latent", dist.Normal(0.0, jnp.ones(fdim)).to_event(1))
+        with numpyro.plate("data", x.shape[0]):
+            numpyro.sample("y", dist.Normal(jnp.matmul(x, latent), 1.0), obs=y)
+
+    def guide(x, y=None):
+        latent_loc = numpyro.param(
+            "latent_loc", jnp.ones(fdim), constraint=constraints.real
+        )
+        numpyro.sample("latent", dist.Normal(latent_loc, 1.0).to_event(1))
```
fehiepsi, line 114: Maybe we can add an assertion here that `latent_loc.ndim == 1` (those assertations are typical in numpyro models). We need not to restrict users to write "batch-compatible" code. Is it a special requirement of SteinVI? Either way, we might not need to restrict ourselves in this more user-faced utility.

OlaRonning: I can add an assertion.

From the context I'm not sure what you mean by batch-compatible. Would you care to elaborate?

OlaRonning: I see. Thanks for clarifying. You are right; I'll have to vmap to produce the desired behavior. SteinVI vmaps over the particle dimension(s), so that shouldn't be a problem. I'll add a shape assert to one of the unittest to confirm in #833.

fehiepsi: Maybe we can rename the current `__call__` method into `_call_with_params(self, params)`. And in the `__call__` method, we can check
```
# NOTE: we also need to handler `rng_key`, `args`, `kwargs` correctly.
if self.guide is None or (batch_ndims == 0):
    return self._call_with_params(self.params)
elif batch_ndims == 1:
    tree_map(lambda x: jnp.swap_axes(x, 0, 1), vmap(self._call_with_params)(self.params))
else:
    raise NotImplementedError
```

OlaRonning: I was thinking to handle it with a similar conditional in the `__call__` instead. If we use your suggestions I can factor the `sample_shape` out the new `__call__`.

Could we so 
``` python            
jax.vmap(self._call_with_params, in_axes=self.batch_ndims, out_axes=-self.batch_ndims)(self.params)
```
instead of the `tree_map`?

fehiepsi: Yeah, using `in_axes` and `out_axes` are much cleaner. Maybe `in_axes=0` and `out_axes=1` for `batch_ndims==1`? What's your definition of `batch_ndims`? Currently, `batch_ndims` is used to denote the number of batch dimensions of `posterior_samples`. Here we don't have `posterior_samples` but having `guide`. My intuition is to let it denote the number of batch dimensions of `params`, but would be interested in hearing your intuition.

OlaRonning: I think of `batch_ndims` as the dimensionality of the particles, which is equivalent to what you describe. 

I sketched an implementation based on your suggestion. I have to compute  `param_batch_shape` to find the number of `rng_key’s needed. Not sure I can avoid this; what do you think? Your suggestion means we can avoid tuple `num_samples`. I’ll update to use `in_axes` and `out_axes` tomorrow morning.

numpyro/infer/util.py
```diff
@@ -935,6 +942,27 @@ def __call__(self, rng_key, *args, **kwargs):
...
+    def __call__(self, rng_key, *args, **kwargs):
+        if self.batch_ndims == 0 or self.params == {} or self.guide is None:
+            return self._call_with_params(rng_key, self.params, args, kwargs)
+        elif self.batch_ndims == 1:
+            param_batch_shape = None
+            for param in self.params.values():
+                if param_batch_shape is None:
+                    assert param.shape[: self.batch_ndims]
+                    param_batch_shape = param.shape[: self.batch_ndims]
+                assert param.shape[: self.batch_ndims] == param_batch_shape
```
fehiepsi, line 954: In numpyro, we allow `param` to have arbitrary pytree structure (e.g. the ones in flax/haiku modules) so `param.shape` will not be available. I think you can do
```
batch_size = jnp.shape(tree_flatten(self.params)[0][0])[0]
rng_keys = random.split(rng_key, batch_size)
```

test/infer/test_infer_util.py
```diff
@@ -96,6 +96,37 @@ def guide(data):
...
+    params = {"latent_loc": jnp.zeros((num_particles, fdim))}
+    x = dist.Normal(jnp.full(3, 0.2), 1.0).sample(random.PRNGKey(0), (num_data,))
+    predictions = Predictive(
+        model,
+        guide=guide,
+        params=params,
+        num_samples=num_samples,
+        batch_ndims=1,
+    )(random.PRNGKey(0), x)
+    print(predictions)
```
fehiepsi, line 126: todo: cleanup

test/infer/test_infer_util.py
```diff
@@ -128,6 +159,7 @@ def test_prior_predictive(batch_ndims):
 
     # check shapes
     batch_shape = (1,) * (batch_ndims - 1) + (100,)
+    print(batch_shape)
```
fehiepsi, line 162: todo: cleanup

fehiepsi (commented):
LGTM. I just have a few minor comments

OlaRonning:
Thanks for the review. I'll make the updates you suggest.

numpyro/infer/util.py
OlaRonning: Makes sense. I'll change it.

fehiepsi (approved):
LGTM, thanks @OlaRonning!

PR #1259, opened 2021-12-19
allow TruncatedNormal to take both low and high

### LINKED ISSUE #1112: Allow for `high` in TruncatedNormal, TruncatedCauchy
TruncatedDistribution has both `low` and `high`. Why do `TruncatedNormal` and `TruncatedCauchy` only have `low`?

MarcoGorelli:
xref #1112 

still need to do Cauchy - if this is alright, is it OK to leave that to a follow-up PR?

---

@fehiepsi @fritzo I've marked this as "ready for review", happy to receive feedback when you want (absolutely no hurry though :smile: !)


numpyro/infer/reparam.py
```diff
@@ -132,14 +132,13 @@ class TransformReparam(Reparam):
...
         fn, expand_shape, event_dim = self._unwrap(fn)
-        if isinstance(fn, (dist.Uniform, dist.TruncatedCauchy, dist.TruncatedNormal)):
+        if not isinstance(fn, dist.TransformedDistribution):
             raise ValueError(
                 "TransformReparam does not automatically work with {}"
                 " distribution anymore. Please explicitly using"
                 " TransformedDistribution(base_dist, AffineTransform(...)) pattern"
                 " with TransformReparam.".format(type(fn).__name__)
             )
-        assert isinstance(fn, dist.TransformedDistribution)
```
MarcoGorelli, line 142: couple of comments here:
- the `assert` wouldn't run when executing Python in optimized mode
- as `dist.TruncatedNormal` is a function, `isinstance(fn, dist.TruncatedNormal)` isn't allowed

Any objections to just raising if `fn` isn't an instance of `dist.TransformedDistribution`?

numpyro/distributions/truncated.py
```diff
@@ -325,45 +390,10 @@ def tree_unflatten(cls, aux_data, params):
...
-            aux_data = None
-        return (self.low, self.loc, self.scale), aux_data
-
-    @classmethod
-    def tree_unflatten(cls, aux_data, params):
-        d = cls(*params)
-        if aux_data is not None:
-            d._support = constraints.greater_than(aux_data)
-        return d
+def TruncatedNormal(*, low=None, high=None, loc=0.0, scale=1.0, validate_args=None):
```
fehiepsi, line 393: nit: maybe the order `loc`, `scale`, `low`, `high` is better?

test/test_distributions.py
```diff
@@ -425,13 +425,14 @@ def get_sp_dist(jax_dist):
         np.array([0.0, 1.0]),
         np.array([[1.0], [2.0]]),
     ),
-    T(dist.TruncatedNormal, -1.0, 0.0, 1.0),
-    T(dist.TruncatedNormal, 1.0, -1.0, np.array([1.0, 2.0])),
+    T(_TruncatedNormal, 0.0, 1.0, -1.0, None),
+    T(_TruncatedNormal, -1.0, np.array([1.0, 2.0]), 1.0, None),
```
fehiepsi, line 429: nit: I think you can use `dist.TruncatedNormal` and remove `_TruncatedNormal`.

numpyro/infer/reparam.py
fehiepsi: Those are good points! Thanks for making this enhancement!

fehiepsi (approved):
Look great to me overall!

test/test_distributions.py
MarcoGorelli: I'd tried this originally, but it makes the tests a lot more complicated because `dist.TruncatedNormal` takes keyword-only arguments, and so lots of tests needed adjusting to have both `params` and `kw_params` in the signature. IMO it's much simpler like this, but I'll change it if you prefer

MarcoGorelli (commented):
Thanks @fehiepsi for your review!

test/test_distributions.py
fehiepsi: You are right. Thanks for explaining!

fehiepsi (approved):

PR #1269, opened 2022-01-01
Add a warning for calling inverse of a non-bijective transform

### LINKED ISSUE #1253: Raise an exception (or warning) when calling the inverse of a non-invertible transform.
The `Transform` class in `numpyro.distributions.transforms` provides an `inv` property which is meant to return the inverse transform. However, this is available even for functions that are not invertible. In those cases, we silently return the identity function which seems weird.

### Example:
```python

f1 = dist.transforms.AbsTransform()
print(f1.inv(-1))       # Not defined, but returns -1
print(f1.inv(f1(-1))) # Returns +1
```

The issue is also present in transformations that are obtained via `ComposeTransform`.
When the transformation is invertible, it is very easy to obtain the CDF of the transformed distribution (which is something I often need). So I think providing an 'automatic' inverse function is a good idea, but maybe it'd be good to raise an error (or at least a warning) when trying to access the inverse of a non-invertible transform.

fehiepsi:
Resolves #1253

This PR also includes a couple of small doc fixes.

### DIFF
--- a/numpyro/distributions/transforms.py
+++ b/numpyro/distributions/transforms.py
@@ -159,6 +159,10 @@ class AbsTransform(Transform):
         return jnp.abs(x)
 
     def _inverse(self, y):
+        warnings.warn(
+            "AbsTransform is not a bijective transform."
+            "The inverse of `y` will be `y`."
+        )
         return y
 
 
--- a/numpyro/distributions/transforms.py
+++ b/numpyro/distributions/transforms.py
@@ -162,7 +162,8 @@ class AbsTransform(Transform):
     def _inverse(self, y):
         warnings.warn(
             "AbsTransform is not a bijective transform."
-            "The inverse of `y` will be `y`."
+            " The inverse of `y` will be `y`.",
+            stacklevel=find_stack_level(),
         )
         return y
 
--- a/numpyro/infer/reparam.py
+++ b/numpyro/infer/reparam.py
@@ -187,7 +187,7 @@ class NeuTraReparam(Reparam):
     """
     Neural Transport reparameterizer [1] of multiple latent variables.
 
-    This uses a trained :class:`~pyro.contrib.autoguide.AutoContinuous`
+    This uses a trained :class:`~numpyro.infer.autoguide.AutoContinuous`
     guide to alter the geometry of a model, typically for use e.g. in MCMC.
     Example usage::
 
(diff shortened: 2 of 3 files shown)


OlaRonning (dismissed):
Hi @fehiepsi, LGTM.

numpyro/distributions/transforms.py
```diff
@@ -159,6 +159,10 @@ def __call__(self, x):
         return jnp.abs(x)
 
     def _inverse(self, y):
+        warnings.warn(
+            "AbsTransform is not a bijective transform."
+            "The inverse of `y` will be `y`."
+        )
```
MarcoGorelli, line 165: Just a small suggestion (feel free to ignore of course :smile: ), so that the warning indicates where it's coming in the _user_'s code

```suggestion
        warnings.warn(
            "AbsTransform is not a bijective transform."
            "The inverse of `y` will be `y`.",
            stacklevel=3,
        )
```

MarcoGorelli: e.g. if I have a file `t.py` with
```python

f1 = dist.transforms.AbsTransform()
f1.inv(-1)
```

then the current code shows:
```console
$ python t.py 
/home/marco/numpyro-dev/numpyro/distributions/transforms.py:162: UserWarning: AbsTransform is not a bijective transform.The inverse of `y` will be `y`.
  warnings.warn(
```
while with this change, it shows
```console
$ python t.py 
t.py:4: UserWarning: AbsTransform is not a bijective transform.The inverse of `y` will be `y`.
  f1.inv(-1)
```

MarcoGorelli: ah, this also highlights the missing whitespace before "The" :laughing:

MarcoGorelli: actually, it seems that all warnings in NumPyro are without `stacklevel` set, so perhaps that's best left to a separate PR

I'll just keep the suggestion here to add the extra whitespace then

fehiepsi: Thank you! I just merged and addressed the issue.

OlaRonning (approved):
@MarcoGorelli nice catch, thanks! @fehiepsi looks good.

fehiepsi:
Thanks for reviewing, @OlaRonning and @MarcoGorelli!

PR #1270, opened 2022-01-01
Support kl divergence for TFP distributions

fehiepsi:
This PR enables us to use TraceMeanField_ELBO (or other inference methods that require `kl_divergence`) for models with TFP distributions.


fritzo (approved):
:+1:

fehiepsi:
Thanks, @fritzo!!

PR #1271, opened 2022-01-01
add stacklevel to warnings

MarcoGorelli:
xref https://github.com/pyro-ppl/numpyro/pull/1269#discussion_r777140387

This is so that warnings reference the line in the _user_'s code where the warning is emitted, rather than the line of code internal to NumPyro (this way users know what they need to change)


fehiepsi (approved):
I'm not sure what's the common practice here but the changes look good to me - it seems to be better to raise the warning at the user-level code. Thanks, @MarcoGorelli!

PR #1272, opened 2022-01-02
Tutorial for truncated distributions

omarfsosa:
The tutorial is not ready yet, but opening the PR so that anyone can give suggestions/comments while I work on it.


MarcoGorelli:
Nice, this is really well-written @omarfsosa , really useful+informative!

Just leaving a few comments / suggestions:

The table of contents could have links, see the [stacking](https://num.pyro.ai/en/latest/tutorials/bayesian_hierarchical_stacking.html) tutorial.

> the support of the Uniform(0, 1) distribution is the interval  (0,1)
Nitpitck, but I think the NumPyro one is [0, 1)? At least that's what it looks like from [jax.random.uniform](https://jax.readthedocs.io/en/latest/_autosummary/jax.random.uniform.html#jax-random-uniform)

```diff
- or sometimes the detectors fail if the signal exceed certain value
+ or sometimes the detectors fail if the signal exceeds certain value
```

```diff
- and implment a few basic ingredients:
- and implement a few basic ingredients:
```

```diff
- This were we defined the parameters of the distribution And we also use jax and lax to promote the parameters to shapes that are valid for broadcasting.
+ This is where we define the parameters of the distribution. And we also use jax and lax to promote the parameters to shapes that are valid for broadcasting.
```

> The brute force approach consists of simply scanning all non-negative integers, one by one, until the value of the cumulative density exceeds the argument  𝑢

Just out of interest, could binary search be used to speed this up?

fehiepsi:
Awesome tutorial, @omarfsosa! I'm looking forward to its final version.

Just a few comments for some of your todos:
+ `support` is used in some inference algorithms like MCMC, SVI with autoguides, where we need to perform the algorithm in the unconstrained space.
+ `is_discrete` support is used to tell the inference algorithms which variables are discrete latent variables
+ `support.event_dim` is used to tell us the event dimension of this distribution (e.g. MVN has event_dim=1). It is overlapping with the `self.event_dim` attribute of a distribution. We wanted this information to reduce overhead when using funsor distributions. Maybe you can skip it for those truncated distributions because they all have `event_dim=0`.
+ `reparametrized_params` is used to determine if a distribution is reparameterizable. Maybe you can link to [this tutorial](https://pyro.ai/examples/svi_part_iii.html#Tricky-Case:-Non-reparameterizable-Random-Variables) for interested readers
+ No need to use `jax.jit`

I'll try to see if we can make `host_callback` work.

omarfsosa:
Thank you for the comments @MarcoGorelli, I'll fix those :)
> Just out of interest, could binary search be used to speed this up?
I think so, yes! I'll revisit this once I get closer to finish the other parts of the notebook. I also have an approximate implementation here https://gist.github.com/omarfsosa/e418b63d5d4f093e393aeafc5076202d which what I personally use. It works for me and it's a lot faster but also a lot more tedious to explain. So maybe not good for a tutorial.

@fehiepsi  that is super helpful, thank you! Let me know if you have any luck using `host_callback`.

fehiepsi:
@omarfsosa It seems `host_callback` changed the signatures. You can use the following version
```python
from scipy.stats import poisson as sp_poisson
import jax
import numpy as np

def rv_truncated_poisson(args):
    rate, low, shape = args
    rate = np.asarray(rate)
    low = np.asarray(low)
    d = sp_poisson(rate)

    lower_cdf = d.cdf(low)
    upper_cdf = 1.
    normalizer = upper_cdf - lower_cdf
    sample = np.random.random(shape) * normalizer + lower_cdf
    return d.ppf(sample)

def sample(rate, low, shape):
    s = jax.experimental.host_callback.call(
        rv_truncated_poisson, (rate, low, shape),
        result_shape=jax.ShapeDtypeStruct(shape, jnp.result_type(float)))
    return s.astype(jnp.result_type(int))

jax.jit(sample, static_argnums=2)(1., 2, (100,))
```

notebooks/source/truncated_distributions.ipynb
```diff
@@ -0,0 +1,1383 @@
+{
```
fehiepsi, line 1: <p>typo: Numpyro -&gt; NumPyro or numpyro</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pyro-ppl/numpyro/pull/1272/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

fehiepsi, line 1: <pre class="ql-syntax" spellcheck="false">Line #15.    numpyro.set_host_device_count(4)
</pre><p>Maybe it is not necessary to use 4 devices and 4 chains?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pyro-ppl/numpyro/pull/1272/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

fehiepsi, line 1: <pre class="ql-syntax" spellcheck="false">Line #12.        _resolution = jnp.finfo(jnp.result_type(float)).resolution
</pre><p>When we implement a distribution, It is better to avoid invoking JAX ops globally (so modifies like <code>numpyro.enable_x64()</code>  work).</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pyro-ppl/numpyro/pull/1272/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='20'/>

martinjankowiak:
thanks @omarfsosa looks great!
random comments:
- "For this same reason, if the truncation point is known and we are going to do inference using MCMC, the normalization constant  can simply be dropped." this isn't generally true if some of the parameters of the distribution are random variables. 
- it may be worth noting that it's not necessary to implement `sample` if the distribution is only used as a prior in a model (i guess you do this later but maybe better to do so when you first list the main distribution methods?)
- "The reparametrized_params is used to determine if a distribution is reparameterizable. Reparametrizing is used to deal with complicated posterior geometries that make sampling very slow or biased" this isn't quite right. unfortunately the word "reparameterize" is overloaded with several meanings. the (svi-specific) usage in the distribution class is explained e.g. [here](http://pyro.ai/examples/svi_part_iii.html#Easy-Case:-Reparameterizable-Random-Variables)
- is this notebook runnable as is? e.g. where is `RNG` defined?
- add a link to the `host_callback` docs?
- maybe move "Built-in truncated distributions" before the custom stuff?

omarfsosa:
Thanks @martinjankowiak, I agree with all your comments! I'll do the changes and update the PR.
To answer your question, the notebook should be runnable as is. The `RNG` seed is defined in the first code cell right after the import statements.
About the `reparametrized_params`, I had completely misunderstood its purpose. Unfortunately, I don't know enough about how SVI is implemented to say anything useful here. Care to give me a sentence or two to put in there?

martinjankowiak:
maybe something like

> The `reparametrized_params`  attribute is used by variational inference algorithms when constructing gradient estimators. The parameters of many common distributions with continuous support (e.g. the Normal distribution) are reparameterizable, while the parameters of discrete distributions are not.  See [SVI Part III](http://pyro.ai/examples/svi_part_iii.html) for more details.

notebooks/source/truncated_distributions.ipynb
```diff
@@ -0,0 +1,1614 @@
...
```
fehiepsi, line 1: <p>Missing link to <a href="https://num.pyro.ai/en/stable/distributions.html#distribution" target="_blank">Distribution</a> at <code>A NumPyro distribution should subclass <code>Distribution</code> </code></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pyro-ppl/numpyro/pull/1272/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

fehiepsi, line 1: <p>Missing link to DiscreteHMCGibbs: it is possible to do so with <code>DiscreteHMCGibbs</code></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pyro-ppl/numpyro/pull/1272/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='22'/>

fehiepsi, line 1: <pre class="ql-syntax" spellcheck="false">Line #44.        @partial(jax.vmap, in_axes=(None, 0))
</pre><p>It is not clear to me why <code>in_axes=(None, 0)</code>. I think you can write a while loop with signature (n, cdf)</p><pre class="ql-syntax" spellcheck="false">def cond_fn(val):
    n, cdf = val
    return jnp.any(cdf &lt; u)

def body_fn(val):
    n, cdf = val
    n_new = jnp.where(cdf &lt; u, n + 1, n)
    return n_new, self.cdf(n_new)

</pre><p>which works for general shapes of <code>u</code>.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pyro-ppl/numpyro/pull/1272/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='24'/>

fehiepsi, line 1: <p>Missing link to TruncatedNormal</p><p> </p><p>You can link to https://num.pyro.ai/en/stable/distributions.html#truncated-distributions for <code>a bunch of truncated distributions</code></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pyro-ppl/numpyro/pull/1272/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='40'/>

fehiepsi, line 1: <p>Missing links to TruncatedNormal, SoftLaplace</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pyro-ppl/numpyro/pull/1272/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='44'/>

fehiepsi, line 1: <p>Missing links to <code>FoldedDistribution</code>.  The folded distribution class is useful even for non-symmetric base distributions, e.g. <code>FoldedDistribution(StudentT(df, 1.0, scale))</code>.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pyro-ppl/numpyro/pull/1272/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='49'/>

omarfsosa: Oh interesting, thanks for pointing this out @fehiepsi . I guess this means folding is an entirely different kind of "truncation". For example, `FoldedDistribution(Normal(loc=1.0))` is not the same as `TruncatedNormal(loc=1.0, low=0.0)`.
I'll elaborate a bit more in the tutorial. Do you know of any examples/applications where folding a non-symmetric base distribution is useful?

martinjankowiak: i think they are generally used as a way of creating heavy-tailed distributions. for example: https://www.sciencedirect.com/science/article/abs/pii/S0377042720306427#!

omarfsosa:
@fehiepsi , @martinjankowiak I've done some changes trying to include all of your comments/suggestions, but please let me know if I missed anything :)

martinjankowiak:
lgtm (note i didn't check code/math carefully)

martinjankowiak (approved):

notebooks/source/truncated_distributions.ipynb
```diff
@@ -0,0 +1,1916 @@
...
```
MarcoGorelli, line 1: <p>why is <code>x==x</code> necessary?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pyro-ppl/numpyro/pull/1272/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='31'/>

MarcoGorelli:
Just left a comment about the mysterious (to me, at least, so perhaps to other reads too :smile: ) `x==x`, but also wanted to point out that when built in Sphinx some of the formatting looks a bit off compared to when viewed in Jupyter, e.g.:

image

Likewise for the links

notebooks/source/truncated_distributions.ipynb
MarcoGorelli, line 1: <p>Perhaps move this explanation of <code>condition</code>  2 cells up, because that's the first place in the notebook where you use <code>condition</code>?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pyro-ppl/numpyro/pull/1272/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='49'/>

omarfsosa:
> what makes it non-smooth? i can appreciate that inferring the truncation point is going to be hard but is it impossible within a vanilla hmc framework?

@martinjankowiak , I mean that it is not smooth with respect to the `high` parameter. The posterior of `high` is defined piecewise. So if I take just one datapoint `y` for the sake of simplicity, the posterior has 2 pieces and the derivative does not exist at `high == y`. I suspect a similar logic follows when you have multiple observations `y`, though I haven't taken the time to check that.
I'm happy to be corrected, please let me know if you think I'm wrong here.

> nit: remove the last empty cell
Oops. Yes, I'll clean that up, thanks!

notebooks/source/truncated_distributions.ipynb
omarfsosa: Hmmm.. tbh I just copied the pattern from [here](https://github.com/pyro-ppl/numpyro/blob/ed498ab74a6dafdd37ea766e2e8653a51ba129cc/numpyro/distributions/constraints.py#L433). My guess is that it has something to do with NaN values, but not 100% sure.

omarfsosa:
@MarcoGorelli Thanks for the comments. I'll see if I can do anything to fix the formatting. 

> Perhaps move this explanation of condition 2 cells up, because that's the first place in the notebook where you use condition?

Yes, good catch! I'll move that :)

MarcoGorelli:
> @MarcoGorelli Thanks for the comments. I'll see if I can do anything to fix the formatting.
> 
> > Perhaps move this explanation of condition 2 cells up, because that's the first place in the notebook where you use condition?
> 
> Yes, good catch! I'll move that :)

Perhaps take a look at the "Bayesian Hierarchical Stacking" notebook? They render fine there when viewed on the website

martinjankowiak:
> I mean that it is not smooth with respect to the high parameter.

do you mean when it's used in an observe statement? i guess it should still be ok when used as a latent variable?

notebooks/source/truncated_distributions.ipynb
fehiepsi: <p>Here is a better link for DiscreteHMCGibbs: https://num.pyro.ai/en/stable/mcmc.html#discretehmcgibbs Typically we link to the docs in num.pyro.ai, rather than the source code.</p>

notebooks/source/truncated_distributions.ipynb
fehiepsi: <p>Not sure why the progress_bars in your MCMC run is so long. Maybe it depends on the screen resolution...</p>

notebooks/source/truncated_distributions.ipynb
fehiepsi: <p>That's right. NaN is an invalid value.</p>

fehiepsi, line 1: <pre class="ql-syntax" spellcheck="false">Line #24.    RNG = random.PRNGKey(0)
</pre><p>Maybe it is better to use different RNG keys for getting priors, getting posterior samples?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pyro-ppl/numpyro/pull/1272/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

fehiepsi, line 1: <p>Better link for truncated distributions: https://num.pyro.ai/en/stable/distributions.html#truncated-distributions</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pyro-ppl/numpyro/pull/1272/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

fehiepsi, line 1: <p>Not important: another link for TruncatedDistribution: https://num.pyro.ai/en/stable/_modules/numpyro/distributions/truncated.html#TruncatedDistribution</p><p> </p><p>Typo: Predictive (and its link: https://num.pyro.ai/en/stable/utilities.html#predictive)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pyro-ppl/numpyro/pull/1272/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='25'/>

fehiepsi, line 1: <p>Better link for FoldedDistribution: https://num.pyro.ai/en/stable/distributions.html#foldeddistribution</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pyro-ppl/numpyro/pull/1272/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='26'/>

fehiepsi, line 1: <pre class="ql-syntax" spellcheck="false">Line #2.    mcmc.run(RNG, num_observations, true_x)
</pre><p>We got a lot of divergences because the data can be out of the support. I think to perform inference, you need to set a prior for <code>high</code>  such that its minimum value should be greater than all observations. E.g.</p><pre class="ql-syntax" spellcheck="false">high = numpyro.sample("high", dist.HalfNormal()) + (x.max() if x is not None else 0)
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pyro-ppl/numpyro/pull/1272/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='42'/>

fehiepsi, line 1: <p>It is typically to get constant samples for <code>low</code> if its density is too large comparing to other values. For this model, I believe the posterior has a high density at low=4. However, the setting is a bit unusual to me. It is lucky that the data seems to have minimum is 4, hence prior for <code>low</code> is <code>Categorical(0.2 * jnp.ones(5))</code> make sense. If the data have minimum is 3, we should set a different prior. I think you can use the model</p><pre class="ql-syntax" spellcheck="false">def truncated_poisson_model(num_observations, x=None, infer_low=True):
    if infer_low:
  &nbsp; &nbsp;   low = numpyro.sample("low", dist.Categorical(0.2 * jnp.ones((5,))))
    else:
        low = 4
  &nbsp; rate = numpyro.sample("rate", dist.LogNormal(1, 1))
  &nbsp; with numpyro.plate("observations", num_observations):
  &nbsp; &nbsp; &nbsp; numpyro.sample("x", LeftTruncatedPoisson(rate, low), obs=x)
</pre><p>, then draw prior samples with <code>infer_low=False</code>. Or you can use the prior</p><pre class="ql-syntax" spellcheck="false">dist.Categorical(logits=jnp.zeros(int(x.min() + 1) if x is not None else 5)
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pyro-ppl/numpyro/pull/1272/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='66'/>

fehiepsi, line 1: <p>Use <code>discrete_distplot(thinned_samples);</code>  to remove AxesSubplot</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pyro-ppl/numpyro/pull/1272/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='73'/>

fehiepsi, line 1: <pre class="ql-syntax" spellcheck="false">Line #1.    plt.hist(true_x, bins=20)
</pre><p> </p><p>Not sure why this is not working in my system using the latest jax jaxlib seaborn matplotlib. My workaround is to use plt.plot(x.copy()) here and other plt.hist plots. :(</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pyro-ppl/numpyro/pull/1272/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

omarfsosa: Hi @fehiepsi , I couldn't reproduce this. What versions of jax/matplotlib do you have?

omarfsosa:
@fehiepsi @MarcoGorelli @martinjankowiak I think I've addressed all the comments now. I'd appreciate if you can have one more look to double check I didn't miss any :)

fehiepsi (approved):
Fantastic work!!! I love the details that you put into the tutorial - many (num)pyro users will appreciate that. Thanks for your contribution, @omarfsosa!

nit: could you add `!pip install -q numpyro@git+https://github.com/pyro-ppl/numpyro` to the top of your program?

notebooks/source/truncated_distributions.ipynb
fehiepsi: It also happens in CI (e.g. [this run](https://github.com/fehiepsi/rethinking-numpyro/runs/4696416337?check_suite_focus=true) that installs `matplotlib-3.5.1` `jax-0.2.26` `jaxlib-0.1.75`) and I fixed that issue by using [.copy](https://github.com/fehiepsi/rethinking-numpyro/commit/8115e954f1bb3226110c7e95a09d72eea42546aa#diff-9cae103a75e9ba92f55f71040a13b4b661f9862b55cb9225a98422f357570f0cR146). I got the same issues locally in my system with your tutorial.

You can test your tutorial in colab by adding
```
!pip install -q git+https://github.com/pyro-ppl/numpyro.git
!pip install -Uq jax jaxlib matplotlib
```
to the top of [your tutorial](https://colab.sandbox.google.com/github/omarfsosa/numpyro/blob/tutorial%2Ftruncated/notebooks/source/truncated_distributions.ipynb)

MarcoGorelli (approved):
> if you can have one more look to double check I didn't miss any :)

Yeah, no objections from my end, well done, I learned a lot from this!

notebooks/source/truncated_distributions.ipynb
omarfsosa: Awesome, thank you! I've just added this changes :)

martinjankowiak:
looks great!!!

> However, the log_prob method would work regardless

i guess you mean to say that all it requires is a cdf (which is more frequently available than the icdf)

omarfsosa:
> > However, the log_prob method would work regardless
> 
> i guess you mean to say that all it requires is a cdf (which is more frequently available than the icdf)

Yup, thank you! I've rephrased that to make it clear.

martinjankowiak (approved):

notebooks/source/truncated_distributions.ipynb
```diff
@@ -0,0 +1,1964 @@
...
```
fehiepsi, line 1: <pre class="ql-syntax" spellcheck="false">Line #2.    !pip install -Uq jax jaxlib matplotlib
</pre><p>This is unnecessary</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pyro-ppl/numpyro/pull/1272/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

fehiepsi, line 1: <pre class="ql-syntax" spellcheck="false">Line #5.        x, y = jnp.unique(samples, return_counts=True)
</pre><p>Maybe using <code>np.unique</code>?  I'm not sure if <code>ax.bar</code>  works with device arrays in the latest jax version.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pyro-ppl/numpyro/pull/1272/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='55'/>

omarfsosa: Ah, good point @fehiepsi thank you! Should be fixed now :)

fehiepsi (approved):

PR #1274, opened 2022-01-02
Holt-winters example

MarcoGorelli:
Hey, just suggesting adding an example of Holt-Winters smoothing, as it took me a while to figure out how to do this and I think it's much simpler than the current time series example (which is excellent BTW, learned a lot from it!)

Here's how the docs would look:

image

image

---

Of course, no offense taken if you think this doesn't belong in the docs


examples/holt_winters.py
```diff
@@ -0,0 +1,198 @@
...
+
+N_POINTS_PER_UNIT = 10  # number of points to plot for each unit interval
+
+
+def holt_winters(y, n_seasons, future=0):
+    T = y.shape[0]
+    level_smoothing = numpyro.sample("level_smoothing", dist.Beta(1, 1))
+    trend_smoothing = numpyro.sample("trend_smoothing", dist.Beta(1, 1))
+    seasonality_smoothing = numpyro.sample(
+        "seasonality_smoothing", dist.Uniform(0, 1 - level_smoothing)
```
fehiepsi, line 79: It is better to reparameterize this to have static supports. You can either do it manually or through a transformed distributions:
```python
    with numpyro.handlers.reparam(config=lambda _: numpyro.infer.reparam.TransformReparam()):
        seasonality_smoothing = numpyro.sample("seasonality_smoothing", dist.TransformedDistribution(
            dist.Uniform(0, 1), dist.transforms.AffineTransform(0, 1 - level_smoothing))
        "seasonality_smoothing", dist.Uniform(0, 1 - level_smoothing)
```

examples/holt_winters.py
```diff
@@ -0,0 +1,198 @@
...
+    )
+    noise = numpyro.sample("noise", dist.HalfNormal(1))
+    level_init = numpyro.sample("level_init", dist.Normal(0, 1))
+    trend_init = numpyro.sample("trend_init", dist.Normal(0, 1))
+    seasonality_init = numpyro.sample(
+        "seasonality_init", dist.Normal(0, 1).expand([n_seasons])
```
fehiepsi, line 85: nit: adding `dist.Normal(0, 1).expand([n_seasons]).to_event(1)` or using `plate` for this variable.

fehiepsi (approved):
LGTM! Thank you for the great example, @MarcoGorelli!

examples/holt_winters.py
MarcoGorelli: I've changed to use `plate`, but could you please explain why `.to_event(1)` would be better/necessary if I was using the `.expand` form?

examples/holt_winters.py
MarcoGorelli: OK, I've gone with the other formulation from [fpp3](https://otexts.com/fpp3/holt-winters.html), which has `gamma_star` between 0 and 1, and `gamma` is `gamma_star * (1-alpha)`, is this OK?

MarcoGorelli:
Thanks @fehiepsi for your review! I've addressed / responded inline

examples/holt_winters.py
fehiepsi: It is not necessary. Just a good practice for fancier algorithms like enumeration, where we need to use `plate` for all batch dimensions. Here, when you only use `expand([n_seasons])`, `seasonality_init` will have 1 batch dimension. We use `to_event` to make sure that it does not have any not-declared batch dimension.

examples/holt_winters.py
fehiepsi: Thanks, LGTM!

fehiepsi (approved):
Thanks, @MarcoGorelli! The changes look great to me.

examples/holt_winters.py
MarcoGorelli: Thanks for explaining!

PR #1275, opened 2022-01-04
Update fori_collect for new ravel_pytree behavior

fehiepsi:
After https://github.com/google/jax/pull/8951, `ravel_pytree` might return non-float dtype, so we need to make sure that the placeholder `collection` has the same dtype as `ravel_pytree` output.

Tested locally that the change passes the jax master branch.

### DIFF
--- a/numpyro/distributions/continuous.py
+++ b/numpyro/distributions/continuous.py
@@ -27,7 +27,7 @@
 
 
 from jax import lax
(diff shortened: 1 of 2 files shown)


OlaRonning (approved):
LGTM

fehiepsi:
Thanks, Ola!

PR #1277, opened 2022-01-07
Feature/batch ravel pytree

OlaRonning:
I will update the PR with test cases ~~tomorrow~~. It is related to #833.


test/test_util.py
```diff
@@ -92,6 +97,30 @@ def test_ravel_pytree(pytree):
...
+    [
+        {
+            "a": np.array([[1.0, 0.0], [1, 0.0]]),
+            "b": np.array([[1.0, 2.0], [3.0, 4.0]]),
+        },
+    ],
+)
+@pytest.mark.parametrize("nbatch_dims", [0, 1, 2])
+def test_ravel_pytree_batched(pytree, nbatch_dims):
+    flat, unravel_fn = ravel_pytree(pytree, batch_dims=nbatch_dims)
```
fehiepsi, line 111: I think for batch dim=1, you can also do
```
flat = vmap(lambda x: ravel_pytree(x)[0])(pytree)
unravel_fn = ravel_pytree(treemap(lambda x: x[0], pytree)[0]
```
For higher batch dims, you can reshape the batch dimensions to -1, use the above code, and reshape back the result.

OlaRonning: That makes sense. Thanks, @fehiepsi! I've updated to use your suggestion.

fehiepsi (approved):

PR #1279, opened 2022-01-09
improve performance of find_stack_level

MarcoGorelli:
This has been improved in pandas, so bringing the improvement over https://github.com/pandas-dev/pandas/pull/45247/files

On master:
```
In [1]: from numpyro.distributions.discrete import PRNGIdentity

In [2]: %%timeit
   ...: PRNGIdentity()
   ...: 
   ...: 
<magic-timeit>:1: FutureWarning: PRNGIdentity distribution is deprecated. To get a random PRNG key, you can use `numpyro.prng_key()` instead.
3.64 ms ± 70.8 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

Here:
```
In [1]: from numpyro.distributions.discrete import PRNGIdentity

In [2]: %%timeit
   ...: PRNGIdentity()
   ...: 
   ...: 
<magic-timeit>:1: FutureWarning: PRNGIdentity distribution is deprecated. To get a random PRNG key, you can use `numpyro.prng_key()` instead.
5.68 µs ± 58.8 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

### DIFF
--- a/numpyro/util.py
+++ b/numpyro/util.py
@@ -3,6 +3,7 @@
 
 from collections import OrderedDict
 from contextlib import contextmanager
+import inspect
 from itertools import zip_longest
 import os
 import random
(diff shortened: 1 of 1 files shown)


fehiepsi (approved):
Awesome!!! Thank you for the improvement!

PR #1283, opened 2022-01-11
Add Kumaraswamy and relaxed Bernoulli distributions

fehiepsi:
This PR adds Kumaraswamy and RelaxedBernoulli distributions. Those distributions are useful to approximate `Beta` and `Bernoulli` - they can be easily reparameterized.

For RelaxedBernoulli, I don't support `probs` argument to simplify the implementation.


numpyro/distributions/continuous.py
```diff
@@ -505,6 +508,60 @@ def icdf(self, q):
         return self.loc - self.scale * jnp.log(-jnp.log(q))
 
 
+class Kumaraswamy(TransformedDistribution):
```
OlaRonning, line 511: I checked math against wiki. Look good!

numpyro/distributions/continuous.py
```diff
@@ -1426,6 +1483,28 @@ def tree_flatten(self):
         return super(TransformedDistribution, self).tree_flatten()
 
 
+class RelaxedBernoulliLogits(TransformedDistribution):
```
OlaRonning, line 1486: I checked against `tfp` implementation. Looks good.

test/test_distributions.py
```diff
@@ -285,6 +285,8 @@ def get_sp_dist(jax_dist):
     T(_ImproperWrapper, constraints.positive, (), (3,)),
     T(dist.InverseGamma, np.array([1.7]), np.array([[2.0], [3.0]])),
     T(dist.InverseGamma, np.array([0.5, 1.3]), np.array([[1.0], [3.0]])),
+    T(dist.Kumaraswamy, 10.0, np.array([2.0, 3.0])),
+    T(dist.Kumaraswamy, np.array([1.7]), np.array([[2.0], [3.0]])),
```
OlaRonning, line 289: Could you add an example with `0. < conc0 or conc1 < 1.`?

OlaRonning:
@fehiepsi do you have an intuition about when to prefer `Kumaraswamy` over `beta`?

fehiepsi:
Because Kumaraswamy can be reparameterized, it might be useful in SVI. Personally I found that it is fine to use Beta directly.

test/test_distributions.py
fehiepsi: I just added the test.

fehiepsi (commented):
Thanks for reviewing, @OlaRonning!

OlaRonning (approved):
LGTM. Thanks @fehiepsi

PR #1284, opened 2022-01-11
Add various KL divergences for Gamma/Beta families

fehiepsi:
Blocked by #1283


numpyro/distributions/kl.py
```diff
@@ -126,6 +133,19 @@ def kl_divergence(p, q):
...
 
 
+@dispatch(Beta, Beta)
+def kl_divergence(p, q):
+    # From https://en.wikipedia.org/wiki/Beta_distribution#Quantities_of_information_(entropy)
+    a, b = p.concentration1, p.concentration0
+    alpha, beta = q.concentration1, q.concentration0
+    a_diff = alpha - a
+    b_diff = beta - b
+    t1 = betaln(alpha, beta) - betaln(a, b)
```
OlaRonning, line 143: When I compare with wiki I get `\alpha=alpha`, `\beta=beta`, `\alpha'=a` and `\beta'=b` so `t1 = betaln(a,b) - betaln(alpha,beta)`.

numpyro/distributions/kl.py
```diff
@@ -126,6 +133,19 @@ def kl_divergence(p, q):
...
+    t2 = a_diff * digamma(a) + b_diff * digamma(b)
+    t3 = (a_diff + b_diff) * digamma(a + b)
+    return t1 - t2 + t3
```
OlaRonning, line 146: How come you subtract `t2`?

numpyro/distributions/kl.py
```diff
@@ -136,3 +156,46 @@ def kl_divergence(p, q):
     t3 = p.concentration - q.concentration
     t4 = digamma(p.concentration) - digamma(sum_p_concentration)[..., None]
     return t1 - t2 + (t3 * t4).sum(-1)
+
+
+@dispatch(Gamma, Gamma)
+def kl_divergence(p, q):
```
OlaRonning, line 162: Looks good

numpyro/distributions/kl.py
```diff
@@ -136,3 +156,46 @@ def kl_divergence(p, q):
...
+
+@dispatch(Kumaraswamy, Beta)
+def kl_divergence(p, q):
+    # From https://arxiv.org/abs/1605.06197 Formula (12)
+    a, b = p.concentration1, p.concentration0
+    alpha, beta = q.concentration1, q.concentration0
+    b_reciprocal = jnp.reciprocal(b)
+    a_b = a * b
+    t1 = (alpha / a - 1) * (jnp.euler_gamma + digamma(b) + b_reciprocal)
+    t2 = jnp.log(a_b) + betaln(alpha, beta) + (b_reciprocal - 1)
```
OlaRonning, line 195: I believe you are using -E[log p/q], from which I get `-jnp.log(a_b)-betaln(alpha,beta) + (b_reciprocal -1)`.

numpyro/distributions/kl.py
```diff
@@ -136,3 +156,46 @@ def kl_divergence(p, q):
...
+    a_ = jnp.expand_dims(a, -1)
+    b_ = jnp.expand_dims(b, -1)
+    a_b_ = jnp.expand_dims(a_b, -1)
+    m = jnp.arange(1, p.KL_KUMARASWAMY_BETA_TAYLOR_ORDER + 1)
+    t3 = (beta - 1) * b * (jnp.exp(betaln(m / a_, b_)) / (m + a_b_)).sum(-1)
+    return t1 + t2 + t3
```
OlaRonning, line 201: If the above is holds, then t3 should be subtracted.

numpyro/distributions/kl.py
```diff
@@ -136,3 +156,46 @@ def kl_divergence(p, q):
...
+    t1 = gammaln(alpha) - gammaln(a)
+    t2 = (a - alpha) * digamma(a)
+    t3 = alpha * jnp.log(b_ratio)
+    t4 = a * (b_ratio - 1)
+    return t1 + t2 - t3 + t4
+
+
+@dispatch(Weibull, Gamma)
+def kl_divergence(p, q):
+    # From https://arxiv.org/abs/1401.6853 Formula (28)
```
OlaRonning, line 176: Not checked the math.

numpyro/distributions/kl.py
fehiepsi: Here I'm calculating E_p(log p / q) where p=Kuma(a, b), q=Beta(alpha,beta). I'm using the same formula as in the reference, i.e. same notations for a, b, alpha, beta, but swapping notations for p and q (i.e. p here is q in the reference, q here is p in the reference).

OlaRonning: Thanks for the clarification. Found the problem and the implementation is consistent with the paper.

OlaRonning (approved):
LGTM. Thanks @fehiepsi

fehiepsi:
Thanks, Ola!

PR #1288, opened 2022-01-13
Add Other Inference Algorithms section to docs

OlaRonning:
The Applications section is a bit long for now and needs refactoring. In Pyro, we have an "[Other Inference Algorithms](https://pyro.ai/examples/baseball.html)" section that we can mimic and use for SteinVI, nested sampling, hmcecs, MixedHMC, SA. Related to #833.


fehiepsi (approved):
Thank you, Ola!

PR #1289, opened 2022-01-16
Make hmc gibs algorithms work with improper distributions

fehiepsi:
Drawing prototype traces will raise errors for models with ImproperUniform. So we will rely on the `init_to_sample` strategy, which fallbacks to `init_to_uniform` if `sample` is not implemented like in ImproperUniform distributions.


OlaRonning (approved):
LGTM

fehiepsi:
Thanks, @OlaRonning!

PR #1291, opened 2022-01-17
Unittests einstein/util.py

OlaRonning:
Added unittest for `einstein/util.py` and updated to the latest version of the util. Related to #833.


fehiepsi (approved):
Thanks, Ola!

PR #1295, opened 2022-01-20
User contributed example from epidemiology

theorashid:
With reference to #189 and [this forum discussion](https://forum.pyro.ai/t/porting-a-model-from-bugs-to-numpyro/3695/16), I have created this model in NumPyro based on [this (my) paper](https://www.thelancet.com/journals/lanpub/article/PIIS2468-2667(21)00205-X/fulltext) which models death rates for small spatial areas in England for 19 age groups over the years 2002-19. The original model was written in [nimble](https://r-nimble.org), which uses BUGS-like syntax and is [here](https://github.com/theorashid/mortality-statsmodel/tree/master/Models/parametric).

@fehiepsi helped me rewrite this into NumPyro. I've made a slight change from the version in the forum, where I have changed the order of the plates: 
```diff
-space_plate = numpyro.plate("space", N_s2, dim=-3)
-age_plate = numpyro.plate("age_groups", N_age, dim=-2)
+age_plate = numpyro.plate("age_groups", N_age, dim=-3)
+space_plate = numpyro.plate("space", N_s2, dim=-2)
year_plate = numpyro.plate("year", N_t - 1, dim=-1)
```

which matches the maths notation described in the paper and the `[a,s,t]` order in the original model. The dimensions are `(19, 113, 18)` (`(age, space, time)`). Let me know if this change causes an issue.

There are a few issues with the current example, which I'll tick off when we find solutions:

- [ ] The model samples really slowly. I'm running one NUTS chain for a small number of samples, which leads to bad Rhats. It runs much slower than nimble (but I have not checked for efficiency – nimble is using Gibbs sampling). I haven't checked whether there is much speedup on a GPU because I don't have one. Either way, given that I intend to use this model in the future and scale it up from 113 spatial units to 6791 in the real dataset, this performance won't do. We have a couple of options to start:
  - Try a different algorithm. Standard HMC? SA? HMC Gibbs? Or even an SVI approach?
  - Better priors. We could use friendlier hyperpriors on the overall intercept like `Normal(0,10)` or narrower on the global intercept and slope, and `HalfNormal(1)` or narrower for the standard deviations. Although this isn't faithful to the original, it's probably a better prior choice, [according to the stan team](https://github.com/stan-dev/stan/wiki/Prior-Choice-Recommendations) at least.
- [ ] The dataset is currently in the main folder. I would like to use `load_dataset()` but the file needs to be in the cloud somewhere.
- [ ] I'm currently printing the model shape. It would be more instructive to use `render_model()` but that is _really_ slow for this model.
- [ ] Need to add a quick test to `test/test_examples.py` once the model is running nicely.
- [ ] Check the docs are rendering correctly. There's a lot of maths, I'm new to sphinx and I can't see an easy way of previewing.

Finally, if someone is feeling ambitious, this model could be adapted as an example for a CAR prior. This would be a _proper_ spatial model rather than the hierarchy of random effects I have used here. But it would require the development of a new sampler. On the plus side, it might be a gateway for more spatial statisticians to use NumPyro. PyMC have one running and [here is their example](https://docs.pymc.io/en/stable/pymc-examples/examples/case_studies/conditional-autoregressive-model.html).

Let me know where to start. I'm new here and I need a bit of help fixing this – especially the first issue.

Cheers,

Theo


examples/mortality.py
```diff
@@ -0,0 +1,276 @@
...
+    with age_plate:
+        alpha_age_drift_scale = jnp.pad(
+            jnp.broadcast_to(sigma_alpha_age, N_age - 1),
+            (1, 0),
+            constant_values=100.0,  # pad so first term is alpha0, the global intercept with prior N(0, 100)
+        )[:, jnp.newaxis, jnp.newaxis]
+        alpha_age_drift = numpyro.sample(
+            "alpha_age_drift", dist.Normal(0, alpha_age_drift_scale)
+        )
+        alpha_age = jnp.cumsum(alpha_age_drift, -2)
```
fehiepsi, line 187: I guess we need to cummulative sum over axis=-3 for `alpha_age` and `beta_age`?

fehiepsi:
> I would like to use load_dataset() but the file needs to be in the cloud

Can I upload it to [this repo](https://github.com/pyro-ppl/datasets)?

> Check the docs are rendering correctly.

I typically run `make docs` then open the file `docs/build/index.html` in the browser. :)

> It would be more instructive to use render_model() but that is really slow for this model.

I'll look into this. I think `render_model` will be much faster with https://github.com/pyro-ppl/numpyro/pull/1248

fehiepsi:
> leads to bad Rhats. It runs much slower than nimble

Could you [print summary](https://num.pyro.ai/en/stable/diagnostics.html#numpyro.diagnostics.print_summary) (e.g. r_hat, n_eff) of nimble samples for comparison? Did you try to simulate observations using some fixed parameters in both numpyro and nimble to see if they generate comparable results? For NUTS, you can set `max_tree_depth` to smaller values to make it run faster, but the samples will be more correlated - so I'm curious on diagnostics of nimble's samples for this model.

examples/mortality.py
```diff
@@ -0,0 +1,276 @@
...
+    parser.add_argument("-n", "--num-samples", nargs="?", default=100, type=int)
+    parser.add_argument("--num-warmup", nargs="?", default=50, type=int)
+    parser.add_argument("--num-chains", nargs="?", default=1, type=int)
+    parser.add_argument("--device", default="cpu", type=str, help='use "cpu" or "gpu".')
+    parser.add_argument(
+        "--rng_seed", default=21, type=int, help="random number generator seed"
+    )
+    args = parser.parse_args()
+
+    numpyro.set_platform(args.device)
```
fehiepsi, line 274: Maybe also set `numpyro.enable_x64()`? I think it is better to work in double precision for this model.

examples/mortality.py
theorashid: Very good spot. It's interesting that the model still ran with axis=-2. I guess the shape of `alpha_age` is `(19, 1, 1)` so summing over `-2` wouldn't sum anything. And in the case of the incorrect code above, `alpha_age_drift` and `alpha_age` are the same. So the code still ran but it wasn't correct.

examples/mortality.py
theorashid: Sure I'll add that. Is there a downside to `numpyro.enable_x64()`? Would it eat up a lot more memory? When I run this on the full, real dataset with nimble, it gets very memory-intensive (300+ GB for 4 parallel chains)

martinjankowiak:
@theorashid do you know how nimble does inference in detail? presumably it's not pure gibbs sampling because (while i haven't looked at your model in detail) it doesn't look like it's fully conjugate. does it automatically add auxiliary latent variables to deal with e.g. the likelihood? does it use some sort of random-walk metropolis hastings kernel for e.g. the parameters with uniform priors?

your best bet for improving inference would be to either use `HMCGibbs` or to integrate out (marginalize) some of the latent variables by hand. this is because it tends to be harder to get HMC to work well in large numbers of dimensions, in part because it becomes harder to adapt the mass matrix well using a reasonably modest number of warm-up samples. so anything you can do to reduce the dimensionality of the latent space should help. i would focus on the latent variables with the largest dimensionalities. i would expect (partial) marginalization to work best if it's computationally feasible, but i'm unsure if it is, since i haven't taken a closer look at your model.

theorashid:
> I typically run `make docs` then open the file `docs/build/index.html` in the browser. :)

For some reason, `make docs` spits out 
```
/.../numpyro/docs/source/tutorials/discrete_imputation.ipynb:15:Content block expected for the "raw" directive; none found.
make[1]: *** [html] Error 2
```
so I can't build an html file to check the docs. It also creates loads on new files in `docs/source/`

> Can I upload it to [this repo](https://github.com/pyro-ppl/datasets)?

Sure, please add the file! Then I'll implement the `load_dataset()` method.

> Did you try to simulate observations using some fixed parameters in both numpyro and nimble to see if they generate comparable results?

Yes, the data was simulated for the same model with the values `alpha0=-5.1, beta0=-0.04, sigma_alpha_age=1.0, sigma_alpha_s1=0.15, sigma_alpha_s2=0.15, sigma_beta_age=0.003, sigma_beta_s1=0.006, sigma_beta_s2=0.003, sigma_gamma=0.028, sigma_nu=0.063, sigma_xi=0.26, theta=1400`. I don't think NumPyro is recovering those values, although I haven't run enough samples to check fully yet as it would take many hours.

> Could you [print summary](https://num.pyro.ai/en/stable/diagnostics.html#numpyro.diagnostics.print_summary) (e.g. r_hat, n_eff) of nimble samples for comparison?
> Seems like nimble is using [RW](https://r-nimble.org/html_manual/cha-mcmc.html#default-assignment-of-sampler-algorithms) sampler for this model.

nimble uses conjugate samplers where possible and RW samplers are default where they are not. The sampling by nimble was extremely inefficient – I had to thin **a lot.** That's one of my main reasons for moving the model to NumPyro (as well as the potential GPU speedup, and jax neural net support). I actually haven't yet run nimble for this exact dataset.

theorashid:
> @theorashid do you know how nimble does inference in detail? presumably it's not pure gibbs sampling because (while i haven't looked at your model in detail) it doesn't look like it's fully conjugate. does it automatically add auxiliary latent variables to deal with e.g. the likelihood? does it use some sort of random-walk metropolis hastings kernel for e.g. the parameters with uniform priors?

I rebuilt the model above but in nimble. nimble prints out a handy receipt of which samplers they use for which node:
```
===== Monitors =====
thin = 1: mu
thin2 = 1: theta, alpha_age, beta_age, alpha0, beta0, sigma_alpha_age, sigma_beta_age, sigma_xi, sigma_gamma, sigma_nu, sigma_alpha_s1, sigma_beta_s1, sigma_alpha_s2, sigma_beta_s2
===== Samplers =====
RW sampler (4401)
  - sigma_alpha_s1
  - sigma_beta_s1
  - sigma_alpha_s2
  - sigma_beta_s2
  - sigma_alpha_age
  - sigma_beta_age
  - sigma_xi
  - sigma_nu
  - sigma_gamma
  - theta
  - gamma[]  (323 elements)
  - xi[]  (2147 elements)
  - nu[]  (1921 elements)
conjugate sampler (314)
  - alpha0
  - beta0
  - alpha_s1[]  (25 elements)
  - beta_s1[]  (25 elements)
  - alpha_age[]  (18 elements)
  - beta_age[]  (18 elements)
  - alpha_s2[]  (113 elements)
  - beta_s2[]  (113 elements)
```

> your best bet for improving inference would be to either use `HMCGibbs` or to integrate out (marginalize) some of the latent variables by hand.

This is a good point, and nimble has given us a clue as to which variables have conjugate relationships. Although I'm not sure I'm capable of doing them by hand myself... Does NumPyro use conjugate samplers automatically like nimble?

> * Better priors. We could use friendlier hyperpriors on the overall intercept like `Normal(0,10)` or narrower on the global intercept and slope, and `HalfNormal(1)` or narrower for the standard deviations. Although this isn't faithful to the original, it's probably a better prior choice, [according to the stan team](https://github.com/stan-dev/stan/wiki/Prior-Choice-Recommendations) at least.

We could also use friendlier priors which use even more conjugate relationships. Although I'm not sure HalfNormal is conjugate to Normal.

martinjankowiak:
@theorashid looks like nimble is largely relying on RW-MH with less than 10% of the latent variables (in terms of dimensionality) being inferred with gibbs sampling. kind of amazing it works at all... i imagine the sampler mixes very slowly and so the aggressive thinning you do is necessary to get halfway reasonable approximate posterior samples.

this looks like a tricky inference problem, especially if you want to scale up the dimensionality further. ideally you would integrate out `xi` and/or `nu` but it looks like that's "blocked" by the non-conjugate `BetaBinomial` likelihood. maybe there's a data augmentation strategy to deal with that but i'm not aware of any...

you could try incorporating Gibbs updates w.r.t. `alpha_s2` and `beta_s2` and see how much that helps. the corresponding posterior conditionals are presumably gaussian and relatively straightforward to derive. it's hard to say how much that would help a priori however....

you can also try using some of the tricks in the [bad geometry](https://github.com/pyro-ppl/numpyro/blob/master/notebooks/source/bad_posterior_geometry.ipynb) tutorial, in particular the structured mass matrix in example #3. but again i'm not sure how much this will help in practice....

fehiepsi:
> conjugate sampler (314)
```
  - alpha0
  - beta0
  - alpha_s1[]  (25 elements)
  - beta_s1[]  (25 elements)
  - alpha_age[]  (18 elements)
  - beta_age[]  (18 elements)
  - alpha_s2[]  (113 elements)
  - beta_s2[]  (113 elements)
```
Interesting! Currently, we don't support this out-of-box, but you can manually remove those variables from the model (thanks @fritzo for the suggestion). Basically, you can leverage some properties like
+ If `X ~ Normal(0, np.eye(10))` and `y ~ Normal(0, 1)` then `X + y ~ MVN(0, np.eye(10) + 1)`
+ If `X ~ Normal(0, np.eye(10))` then `cumsum(X) ~ MVN(0, scale_tril=np.tril(np.ones(10, 10)))`

Using those properties, you can construct giant covariance matrices for `xi`, `nu_drift`, and `gamma_drift` (with sizes `age x space`, `space x year`, `age x year` correspondingly) after integrating out `alpha_s1`, `beta_s1`, `alpha_s2`, `beta_s2`, `alpha_age_drift`, `beta_age_drift`.

Then you can perform NUTS/RW on the same variables as the `nimble` RW sampler.

Hopefully, in the future, we can leverage [funsor](https://github.com/pyro-ppl/funsor) to do such integrations automatically.

> I don't think NumPyro is recovering those values

It is better to check first if two models are equivalent by simulating two models and comparing observations (that will help you identify something like the wrong dimension of `cumsum` in one of my comments). For NumPyro, you can do
```
conditioned_model = numpyro.handler.condition(model, data=your_truth_values)
```
to fix latent values to some specific values specified in `your_truth_values`).

>  make docs spits out

Could you try to go to `docs` directory and run `make html`? Btw, here is the link to your data: `https://raw.githubusercontent.com/pyro-ppl/datasets/master/simulated_mortality.csv`

theorashid:
> Could you try to go to `docs` directory and run `make html`
This producs the same error as descibed before, creating a load of new files in `docs/source/` and I don't get an html file to open 


Screenshot 2022-01-27 at 10 09 21


Working on the model stuff now. Will update everyone soon

theorashid:
> this looks like a tricky inference problem, especially if you want to scale up the dimensionality further

I assume this means it's a tricky inference problem regardless of the algorithm (NUTS/other MCMC or SVI)? I'd like to use something more efficient in numpyro than nimble + aggressive thinning.

So I've discussed the modelling with some of my colleagues too. It's fine to rewrite the model by manually deriving conjugate where possible, but the main largest samplers are the interaction samplers, particularly `xi` and `nu`. We're going to look at whether there is a better way of formulating these interactions with a lower dimension or one that invokes more conjugate relationships.

I propose that, for this example, we simplify the model by removing the `xi` interaction and moving the random walks (`nu` and `gamma`) into a single global random walk `pi` and using a (normal) Binomial. The likelihood will be
```
...
# random walk
with year_plate:
    pi_drift = numpyro.sample("pi_drift", dist.Normal(0, sigma_pi))
    pi = jnp.pad(jnp.cumsum(pi_drift, -1), (1, 0))
...
latent_rate = (alpha_s2 + alpha_age) + (beta_s2 + beta_age) * t + pi
with numpyro.plate("N", N):
        mu_logit = latent_rate[space, age, time]
        mu = numpyro.deterministic("mu", expit(mu_logit))
        numpyro.sample(
            "deaths",
            dist.Binomial(mu, population),
            obs=deaths,
        )
```
Any advice on the most efficient way to add in the slopes`(beta_space + beta_age) * t`? `jnp.cumsum` in the `year_plate`?

I propose this model because:
1. I don't think it's instructive to have an example that takes several hours to run
1. I don't think it's instructive to have an example where the code is not easily read, as would be the case if we manually rewrote conjugate relationships

So the plan would be to use this simplified model as an example, and then discuss with colleagues improvements to the interaction model, and discuss on the forum the correct implementation of `HMCGibbs()` for the conjugate samplers. If the code for the `HMCGibbs` works nicely, we can update this example in the future.

Let me know what you think, and I will update the code accordingly. It would be good to have an example that ran with NUTS, HMC or SA (or maybe even SVI), depending on the argument the user picks.

fehiepsi:
> I don't think it's instructive to have an example where the code is not easily read, as would be the case if we manually rewrote conjugate relationships

Agreed! I think we should be better supporting this (I don't foresee much difficulty here, except that it might require a bit of time to derive the math and design a nice api, so it might not be available in a near future - something similar to [collapse](https://num.pyro.ai/en/stable/handlers.html#numpyro.handlers.collapse) or implementing some gaussian ops in `numpyro.ops.gaussian` (like [pyro.ops.gaussian](https://github.com/pyro-ppl/pyro/blob/dev/pyro/ops/gaussian.py#L397))
```
alpha_s1 = mvn_to_gaussian(dist.Normal(0, sigma_alpha_s1).expand([N_s1, 1]))
alpha_s2 = alpha_s1[lookup]  # new gaussian
alpha_age_drift = mvn_to_gaussian(dist.Normal(...))
alpha_age = alpha_age_drift.cumsum()  # new gaussian
xi_dist = gaussian_to_mvn(alpha_s2 + alpha_age + mvn_to_gaussian(dist.Normal(0, sigma_xi)))
with age_plate, space_plate:
        xi = numpyro.sample("xi", xi_dist)
```
or working directly with Normal/MVN distributions (which I prefer for now)
```
alpha_s1 = dist.Normal(0, sigma_alpha_s1).expand([N_s1, 1])
alpha_s2 = gaussian_lookup(alpha_s1, lookup)
alpha_age_drift = dist.Normal(...)
alpha_age = gaussian_cumsum(alpha_age_drift, dim=...)
xi_dist = gaussian_add(alpha_age, alpha_s2, dist.Normal(0, sigma_xi))
xi_base = numpyro.sample("xi",  dist.Normal(...))
xi = xi_dist.loc + xi_dist.scale_tril @ xi_base
```
I might help on writing those utilities later this week or next week. My main concern is even if we marginalize those variables, will NUTS behave well for this model and dataset? I'm not sure because there are still a lot of variables remaining...

> the plan would be to use this simplified model as an example

This sounds great!! Sadly I don't have much intuition on the modeling part to say whether using the simplified model is reasonable for this problem/dataset. :(

theorashid:
I'm still having issues with the model. I've reverted to a much simpler model by:
- changing `BetaBinomial` to `Binomial`
- removing `xi` – the big age-space sampler
- removing space slopes `beta_s2`
- removing `nu` and `gamma` and replacing by a single random walk `pi` to account for non-linearities in time. The age slopes are now summed as `beta_age_cum = jnp.outer(beta_age, jnp.arange(N_t))[:, jnp.newaxis, :]`

However, the model is not producing realistic values. e.g output:

```
                                        mean       std    median      5.0%     95.0%     n_eff     r_hat
 alpha_age_drift_decentered[0,0,0]     -0.57      0.00     -0.57     -0.57     -0.57      3.84      1.51
 alpha_age_drift_decentered[1,0,0]     -0.59      0.00     -0.59     -0.59     -0.59      5.49      1.28
 alpha_age_drift_decentered[2,0,0]     -0.05      0.00     -0.05     -0.05     -0.05      2.83      2.04
 alpha_age_drift_decentered[3,0,0]      0.53      0.00      0.53      0.53      0.53      5.86      1.25
 alpha_age_drift_decentered[4,0,0]     -0.06      0.00     -0.06     -0.06     -0.06      2.59      2.61
 alpha_age_drift_decentered[5,0,0]      0.22      0.00      0.22      0.22      0.22      4.63      1.47
 alpha_age_drift_decentered[6,0,0]     -0.17      0.00     -0.17     -0.17     -0.17      5.16      1.30
 alpha_age_drift_decentered[7,0,0]      0.15      0.00      0.15      0.15      0.15      3.34      1.62
 alpha_age_drift_decentered[8,0,0]     -0.75      0.00     -0.75     -0.75     -0.75      4.57      1.01
 alpha_age_drift_decentered[9,0,0]     -0.35      0.00     -0.35     -0.35     -0.35      3.67      1.99
alpha_age_drift_decentered[10,0,0]     -0.05      0.00     -0.05     -0.05     -0.05     10.63      1.00
alpha_age_drift_decentered[11,0,0]     -0.05      0.00     -0.05     -0.05     -0.05      4.52      1.37
alpha_age_drift_decentered[12,0,0]      0.19      0.00      0.19      0.19      0.19     17.97      1.00
```

I am surprised the much simpler model does not fit the data. I assume there are mistakes in the model that have led to the `std==0.00` everywhere. Do you have any ideas?

We could also try simulating data directly from this model and trying to fit based on that.

theorashid:
I've done a bit more digging on the model. The latest commits give some temporary code in `main()` for simulating deaths from this exact model using `numpyro.handlers.condition(model, data)` to condition on provided hyperparameter values. (Aside: I'm not entirely sure how to give the `alpha_age_drift[0]` or `beta_age_drift[0]` – the global intercept and slope – a data value. Is this a problem because of `numpyro.handlers.reparam`?).

However, when I use this simulated death data, the model is not recovering the true provided hyperparameters. The `std == 0` problem remains for all parameters – the inference is only finding a single value for all the parameters in the model. Any ideas on this?

theorashid:
@fehiepsi I have tried tweaking this model over the last week but without success. I think it's beyond my experience using these samplers. **Please** would you be able to **help** me on this. I do not understand the model output, given I am running the model on data simulated from that _exact_ model.

fehiepsi:
Hi @theorashid, sorry for the late response. I still don't have time to implement the gaussian logic to help you marginalize those gaussian conjugate priors. Even if we marginalize them, I'm still not sure if NUTS will be useful for this high dimensional model.

theorashid:
Hi @fehiepsi, an update on this:

I [reimplemented the model in pymc](https://github.com/theorashid/mortality-statsmodel/blob/pymc/Models/parametric/pymc-mod.ipynb) v4 and it seems to be running (much quicker) and sampling (a few divergent transitions, but I'll iron this out when not working on simulated data). It produces parameter estimates with reasonable distributions. I also used `sample_numpyro_nuts`, so that suggests that the numpyro model implementation has faults. Perhaps this pymc implementation might make the issues clearer.

Let me know if there's anything I can produce which will help.

fehiepsi:
@theorashid Could you render the model to see if there are differences in the implementation? I believe render_model is pretty fast now.

theorashid:
@fehiepsi See [here](https://github.com/theorashid/mortality-statsmodel/blob/pymc/Models/parametric/numpyro-mod.ipynb). It mostly looks equivalent. There are more terms in the pymc model but they're the same terms, just more drawn out because I've manually created a decentred parametrisation in pymc. I'm not sure why all terms have arrows towards the observation plate in the numpyro model. And more subtly, `sigma_alpha_s1` points towards both `alpha_s1` and `alpha_s2`.

Pymc:
Screenshot 2022-04-07 at 11 04 33

Numpyro:
Screenshot 2022-04-07 at 11 04 55
?

fehiepsi:
I think it is a bug, might be related to reparam handler. Let me take a look.

fritzo:
Note it is expected behavior for `LocScaleReparam` to add grandparent->child edges when applied to a parent node.  I think it makes most sense to visualize the graph before reparameterizing.  (This issue came up for me when implementing `AutoGaussian`'s structured precision matrix; reparameterizing breaks sparsity in that setting, so I avoided reparameterizing.)

theorashid:
To be clear, is this an issue with the visualisation or with the model itself? I guess there's still an issue with the model if it isn't working when the pymc version is

fehiepsi:
@theorashid Fritz is right, current loc scale reparam makes the visualization a bit tricky to follow. We can make an enhancement for that but it seems to not be the reason that makes NumPyro model not working. I'll look into NumPyro model later today. Can I just run `examples/mortality.py` script in this PR?

theorashid:
Yes, go ahead using the code on this PR. I used the same model and dataset in this PR to make the visualisation and the pymc model

examples/mortality.py
```diff
@@ -0,0 +1,268 @@
...
+    with year_plate:
+        pi_drift = numpyro.sample("pi_drift", dist.Normal(0, sigma_pi))
+        pi = jnp.pad(jnp.cumsum(pi_drift, -1), (1, 0))
+
+    # likelihood
+    latent_rate = alpha_age + beta_age_cum + alpha_s2 + pi
+    with numpyro.plate("N", N):
+        mu_logit = latent_rate[space, age, time]
+        mu = numpyro.deterministic("mu", expit(mu_logit))
+        numpyro.sample("deaths", dist.Binomial(population, mu), obs=deaths)
```
fehiepsi, line 180: There are two things that helped (no divergences but slow :( ):
+ a bug in the model: `mu_logit = latent_rate[age, space, time]`
+ a bug in Binomial??: `numpyro.sample("deaths", dist.Binomial(population, logits=mu_logit), obs=deaths)`

fehiepsi: @theorashid The second bug is fixed in #1387. In the meantime, you can use `dist.Binomial(population, logits=mu_logit)` (recommended) instead. Hope that those changes fix your issue.

theorashid: That's great, thanks. The numypro model takes 8 mins on my laptop. The pymc model takes 14 mins using the same numpyro sampler– but I'll post on their forum now in case the devs there have any speedups.

For some reason, they are producing different results. I'm not too worried as it isn't real data, but it is weird that they are converging to different things. `alpha0` in [pymc-mod.ipynb](https://github.com/theorashid/mortality-statsmodel/tree/pymc/Models/parametric) and `alpha_age_drift_decentered[0,0,0]` in [numpyro-mod.ipynb](https://github.com/theorashid/mortality-statsmodel/tree/pymc/Models/parametric) should be the same, unless the recentring means they aren't actually comparable.

fehiepsi: @theorashid Probably `alpha_s1` and `alpha_s2` are not reparameterized there?

theorashid: I did the reparametrisation myself as
```
alpha_age_innovations = pm.Normal("alpha_age_innovations", sigma=1.0, dims="age_minus_origin")
alpha_age_raw_rw = aet.cumsum(aet.concatenate([aet.zeros(shape=(1,)), alpha_age_innovations], axis=-1), axis=-1)
alpha_age = pm.Deterministic("alpha_age", alpha0 + alpha_age_raw_rw * sigma_alpha_age, dims="age")
```

So I suppose it depends whether pymc's `alpha_age[0]` and numpyro's `alpha_age_drift_decentered[0,0,0]` are equivalent, or whether the latter is actually pymc's `alpha_age_raw_rw[0]` or `alpha_age_innovations[0]`.

fehiepsi: > Would that make a difference to the values?

We have the same model, but using reparameterization might help the sampler.

theorashid: I'll run it, but I don't think normal(0, sd) -> sd * normal(0, 1) make much difference. One is just a scaled geometry of the other (I think). I agree normal(mean, sd) -> mean + normal will make a difference. But I'll run it and see what happens.

theorashid: I've dug further into this. The parameter estimates are roughly the same **except** for `alpha0`/`beta0` in pymc and their numpyro equivalents `alpha_age_drift_decentered[0,0,0]`/`beta_age_drift_decentered[0,0,0]`. I'm not quite sure why but I don't think it's important. So I'll update the model in this fork and move on

theorashid:
Great, so this is now much closer to where it needs to be. I've also added a small test.

One thing that will need updating is `assert numpyro.__version__.startswith("0.8.0")`. The problem is, my environment says `numpyro                   0.8.0                     dev_0    <develop>`. Whats the best way to fix this?  I'm guessing fetch upstream/master won't fix my environment

fehiepsi:
I think you need to use dev version and pump the version in your example to `"0.9.1"` (like other examples upstream). To make tests pass, could you help me change [this line](https://github.com/pyro-ppl/numpyro/blob/master/docs/requirements.txt#L7) to `jaxns==0.0.7`? Thanks!

theorashid:
Done, done and done and I've added an image from the paper

fehiepsi (approved):
Awesome work! Thanks for the contribution, @theorashid!

PR #1297, opened 2022-01-21
Example: Bayes NN with steinVI

OlaRonning:
See #1313. related to #833.


examples/stein_bnn.py
```diff
@@ -0,0 +1,177 @@
...
+    """
+
+    prec_nn = numpyro.sample(
+        "prec_nn", Gamma(1.0, 0.1)
+    )  # hyper prior for precision of nn weights and biases
+
+    n, m = x.shape
+
+    with numpyro.plate("l1_hidden", hidden_dim):
+        b1 = numpyro.sample(  # prior l1 bias term
```
fehiepsi, line 67: This is not formatted well

examples/stein_bnn.py
```diff
@@ -0,0 +1,177 @@
...
+            "nn_b1",
+            Normal(
+                0.0,
+                1.0 / prec_nn,
```
fehiepsi, line 71: Do you mean to take 1 / sqrt(precision)

examples/stein_bnn.py
```diff
@@ -0,0 +1,177 @@
...
+            ),
+        )
+        assert b1.shape == (hidden_dim,)
+
+        with numpyro.plate("l1_feat", m):
```
fehiepsi, line 76: It is a good practice to add dim to plate statements

examples/stein_bnn.py
```diff
@@ -0,0 +1,177 @@
...
+            w1 = numpyro.sample("nn_w1", Normal(0.0, 1.0 / prec_nn))  # prior l1 weights
+            assert w1.shape == (m, hidden_dim)
+
+    with numpyro.plate("l2_hidden", hidden_dim):
+        w2 = numpyro.sample("nn_w2", Normal(0.0, 1.0 / prec_nn))  # prior output weights
+
+    b2 = numpyro.sample("nn_b2", Normal(0.0, 1.0 / prec_nn))  # prior output bias term
```
fehiepsi, line 83: Ditto: sqrt(precision)?

examples/stein_bnn.py
```diff
@@ -0,0 +1,177 @@
...
+
+    prec_obs = numpyro.sample(  # precision prior on observations
+        "prec_obs", Gamma(1.0, 0.1)
+    )
+    with handlers.scale(scale=subsample_size), numpyro.plate(
```
fehiepsi, line 88: Could you comment on why we need to scale here?

numpyro/examples/datasets.py
```diff
@@ -14,6 +14,7 @@
 import zipfile
 
 import numpy as np
+import scipy
```
fehiepsi, line 17: Could you move scipy to local function? We don't require scipy by default.

numpyro/examples/datasets.py
```diff
@@ -70,6 +76,37 @@
     ["https://archive.ics.uci.edu/ml/machine-learning-databases/00280/HIGGS.csv.gz"],
 )
 
+LR_BANANA = dset(
+    "lr_banana", ["https://github.com/pyro-ppl/datasets/raw/master/benchmarks.mat"]
```
fehiepsi, line 80: The filename benchmarks is confusing. Could you rename the file? Also, if mat is not a convenient format (requires scipy) then it is better to store it as npz file

fehiepsi:
@OlaRonning While you are working on this, could you expose SteinVI and its utility to docs? Currently it seems to be missing. It would be nice to write up a little docs for this module (see e.g. Pyro's [forecasting](https://docs.pyro.ai/en/stable/contrib.forecast.html) module) and points to examples that you add.

OlaRonning:
I got a little sidetracked with another project.

> While working on this, could you expose SteinVI and its utility to docs? 

Sure, I'll expose the methods to docs in a separate PR and write module docs as well. 

> Then, I'll update the release docs for the 0.9.0 release.

Is the next release scheduled?

fehiepsi:
Yes, I just made the release. Sorry about the SteinVI docs, I just recognized it missing while collecting the release notes. When it is available, I'll update the release notes to help readers easier to navigate. Thank you!

examples/stein_bnn.py
OlaRonning: We don't need the scale. Performance is the same without scaling. If I recall, scaling mimics [temperature](https://docs.pymc.io/en/latest/api/inference.html#pymc.variational.inference.SVGD) in the pymc3 implementation.

examples/stein_bnn.py
OlaRonning: It makes sense; I've updated with dim params set.

examples/stein_bnn.py
OlaRonning: yeah, ups

numpyro/examples/datasets.py
OlaRonning: I'll fix the names and convert them to NumPy format.

numpyro/examples/datasets.py
OlaRonning: I'll convert the format as you suggest below and remove scipy.

numpyro/examples/datasets.py
OlaRonning: On second thought, I removed it. I'm not sure I'll add the example to numpyro, and if I do, I'll fix the problem then.

OlaRonning:
closed to fix rebase!

fehiepsi:
@OlaRonning it looks like black just released its stable version and changed x ** 2 into x**2 . Could you update black and run `make format`?

OlaRonning:
That makes sense. I'll fix it in a sec.

fehiepsi (approved):
Thanks, @OlaRonning! The example is easy to follow 💯

Issue #1296, opened 2022-01-21
Sparse Poisson different log_prob values depending on type


Vedranh13:
Using `numpyro-0.8.0`, depending on if you pass in an integer for the Poisson rate or a float, you get different values of `log_prob`. Was it intended that `log_prob` return an int value if you pass in an int? Or should it be cast?

To Reproduce:

```
import jax.numpy as jnp
import numpyro.distributions as dist
value = jnp.array([1.,2.,3.])
print(dist.Poisson(5, is_sparse=False).log_prob(value)) # [-3.390562  -2.4742706 -1.9634461]
print(dist.Poisson(5., is_sparse=True).log_prob(value)) # [-3.390562  -2.4742706 -1.9634461]
print(dist.Poisson(5, is_sparse=True).log_prob(value))  # [-3 -2 -1] 
```

fehiepsi:
I think we need to cast to float [this one](https://github.com/pyro-ppl/numpyro/blob/99c31327937cd752ac71fe99a2b7dd075a148eac/numpyro/distributions/discrete.py#L695): `jnp.asarray(rate, dtype=jnp.result_type(float))`. Do you want to submit the fix?

Vedranh13:
Sure, I would love to. I haven't contributed to numpyro before though so might have some questions around style, adding tests, etc during the review process

PR #1301, opened 2022-01-23
Cast rate parameter to float

Vedranh13:
Fix for #1296 

Casts rate parameter to float when calculating Poisson `log_prob`
Added a Poisson distribution with an `int` rate parameter to the test cases to cover this


fehiepsi (approved):
Thanks for the fix, @Vedranh13!!!

PR #1304, opened 2022-01-23
Add loose strategy for missing plates in MCMC

### LINKED ISSUE #1276: [FR] raise missing plate warning in MCMC inference
basically extend #1245 to be enforced more globally

fehiepsi:
Fixes #1276


martinjankowiak (approved):
lgtm. this has no effect on the current svi behavior, right?

fehiepsi:
Yes, but we can merge two implementations. The svi version is better because it also handles singleton dimensions. Let's me merge them.

martinjankowiak (dismissed):
lgtm. @fehiepsi is this still wip?

fehiepsi:
Thanks for reviewing, @martinjankowiak! I just fixed some tests due to new jaxlib/scikit-learn release:
```
/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/sklearn/utils/deprecation.py:87: 
FutureWarning: Function get_feature_names is deprecated; get_feature_names is deprecated in 1.0 and will be removed in 1.2. Please use get_feature_names_out instead.
```

numpyro/infer/elbo.py
```diff
@@ -569,6 +574,8 @@ def single_particle_elbo(rng_key):
             model_trace, guide_trace = get_importance_trace(
                 seeded_model, seeded_guide, args, kwargs, param_map
             )
+            check_model_guide_match(model_trace, guide_trace)
+            _validate_model(model_trace, plate_warning="strict")
```
martinjankowiak, line 578: why strict here?

martinjankowiak (approved):

numpyro/infer/elbo.py
fehiepsi: I'm seeing TraceGraphELBO leverages plate stacks etc to do inference. I'm not sure if strict is required. If you think using loose is enough, I can make the change.

fehiepsi: Let me merge this first, then would make the change if needed in #1310

PR #1305, opened 2022-01-23
Fix HMCECS for likelihoods with multiple plates

fehiepsi:
Currently, when the likelihood has multiple plate statements, HMCECS will raise errors. This PR fixes that bug.

### DIFF
--- a/numpyro/infer/hmc_gibbs.py
+++ b/numpyro/infer/hmc_gibbs.py
@@ -618,7 +618,9 @@ class HMCECS(HMCGibbs):
         self._subsample_plate_sizes = {
             name: site["args"]
             for name, site in self._prototype_trace.items()
-            if site["type"] == "plate" and site["args"][0] > site["args"][1]
+            if site["type"] == "plate"
+            and (site["args"][1] is not None)
+            and site["args"][0] > site["args"][1]
         }  # i.e. size > subsample_size
         self._gibbs_sites = list(self._subsample_plate_sizes.keys())
         assert self._gibbs_sites, "Cannot detect any subsample statements in the model."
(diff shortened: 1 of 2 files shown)


numpyro/infer/hmc_gibbs.py
```diff
@@ -787,7 +789,7 @@ def log_likelihood(params_flat, subsample_indices=None):
                             log_lik[frame.name] += _sum_all_except_at_dim(
                                 site["fn"].log_prob(site["value"]), frame.dim
                             )
-                        else:
+                        elif frame.name in subsample_indices:
```
fehiepsi, line 792: This is where the bug happens. We only want to collect likelihoods at subsample dimensions.

OlaRonning, line 792: That makes sense. 
I think you could merge the conditionals with a `defaultdict`. Something like
``` python
log_lik = defaultdict(float)
for site in tr.values():
  if site["type"] == "sample" and site["is_observed"]:
    for frame in site["cond_indep_stack"]:
      if frame.name in log_lik or frame.name in subsample_indices:
        log_lik[frame.name] += _sum_all_except_at_dim(
          site["fn"].log_prob(site["value"]), frame.dim)
```

OlaRonning (approved):
Thanks, @fehiepsi. The PR looks good. Have a minor comment.

numpyro/infer/hmc_gibbs.py
fehiepsi: I think this works but we need to maintain the type of the returned values. If we use defaultdict here, we will need to use defaultdict in other places like in gibbs_update function below.

PR #1306, opened 2022-01-24
Move contrib.render to infer.inspect

fehiepsi:
This PR follows up #1248 and moves `contrib.render` to `infer.inspect` to match Pyro api. Because we exposed `render_model` to `numpyro` namespace and used `numpyro.render_model` in tutorials so I don't maintain backward-compatibility for those importing `render_model` from `numpyro.contrib.render`.


fritzo:
Note I have no opinion as to whether `render_nodel()` belongs in `[num]pyro.render` or `[num]pyro.inspect`. Historically `numpyro.render` and `pyro.inspect` arose independently and only later did NumPyro acquire `get_dependencies()` and Pyro acquire `render_model()`.  It might make as much sense to move Pyro's `render_model()` into a `pyro.render` module. 🤷

fehiepsi:
Close caused this is unnecessary. Thanks for your input, Fritz!

PR #1310, opened 2022-01-28
Bump to 0.9.0

fehiepsi:
Blocked by #1304

This PR bumps numpyro to a new version. We already have a lot of stuffs since last release. In addition, jax/jaxlib just has new releases recently. Beside that, this PR also
+ Resolves deprecation warnings in examples, in particular we need to explicitly declare infer enumerate parallel in the models that requires enumeration.
+ Move contrib/optim.py to optim.py for (personal) convenience. This shouldn't affect any user.


martinjankowiak (approved):

fehiepsi:
Thanks, @martinjankowiak!

PR #1314, opened 2022-01-31
Stein VI docs

OlaRonning:
Expose documentation for SteinVI. Related #1297 #1298.

PR #1315, opened 2022-02-01
Example: Deep Markov Model with steinVI

OlaRonning:
related to #1314 (and #1298).


examples/stein_dmm.py
```diff
@@ -0,0 +1,379 @@
...
+    **Reference:**
+        1. Structured Inference Networks for Nonlinear State Space Models [arXiv:1609.09869]
+        Rahul G. Krishnan, Uri Shalit and David Sontag (2016)
+    """
+    gate_init_fun, gate_apply_fun = stax.serial(
+        stax.Dense(gate_hidden_dim), stax.Relu, stax.Dense(out_dim), stax.Sigmoid
+    )
+
+    prop_mean_init_fun, prop_mean_apply_fun = stax.serial(
+        stax.Dense(prop_mean_hidden_dim), stax.Relu, stax.Dense(out_dim)
```
fehiepsi, line 79: Not importance: just wonder if it is worth to use either flax or haiku in new tutorials/examples? We used stax in the old days. Maybe we can also use optax for new examples (especially those involve neural networks which is usually used for complicated problems hence needs the flexibility of optax) too. What do you think?

fehiepsi: It seems that stax has GRUcell that you can use. :D

examples/stein_dmm.py
```diff
@@ -0,0 +1,379 @@
+# Copyright Contributors to the Pyro project.
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+Example: Deep Markov Model inferred using SteinVI
+=================================================
+An implementation of a Deep Markov Model in NumPyro based on reference [1][2] and the Pyro DMM example.
+This is essentially the DKS variant outlined in the paper.
```
fehiepsi, line 8: Could you sketch a bit more the purpose of this model and this dataset (it is not clear to me what is the purpose of this fancy model based on the current description) and add link to the Pyro example to the reference section? FYI you can add thumbnail images to  the examples (in case you think it is worth).

OlaRonning: I'll update the model description and develop an image of the generated music and some inference/model diagnostics.

examples/stein_dmm.py
OlaRonning: I'll have a looks neat Optax, I haven't worked with it before. Haiku also has a [GRU cell](https://github.com/deepmind/dm-haiku/blob/main/haiku/_src/recurrent.py#L499) C:

examples/stein_dmm.py
```diff
@@ -0,0 +1,339 @@
...
+                "z",
+                dist.Normal(*combiner(hs, combiner_params))
+                .mask(jnp.expand_dims(masks, axis=-1))
+                .to_event(2),
+            )
+
+
+def vis_tune(i, tunes, lengths, name="stein_dmm.pdf"):
+    try:
+        from music21.chord import Chord
```
OlaRonning, line 270: Would it be preferable to use imshow (and not introduce a dependency)?

examples/stein_dmm.py
```diff
@@ -0,0 +1,340 @@
+# Copyright Contributors to the Pyro project.
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+Example: Deep Markov Model inferred using SteinVI
+=================================================
```
OlaRonning, line 6: Where do I add `stein_dmm.png` to make it a thumbnail?

examples/stein_dmm.py
```diff
@@ -0,0 +1,340 @@
...
+
+The model DMM based on reference [1][2] and the
+[Pyro DMM example](https://github.com/pyro-ppl/pyro/blob/f73df6c1c20bc7b9164d79ce4217557d0aa8e396/examples/dmm/dmm.py).
+
+
+**Reference:**
+    1. Pathwise Derivatives for Multivariate Distributions
+    Martin Jankowiak and Theofanis Karaletsos (2019)
+    2. Structured Inference Networks for Nonlinear State Space Models [arXiv:1609.09869]
+    Rahul G. Krishnan, Uri Shalit and David Sontag (2016)
```
fehiepsi, line 18: I think this is a better link for Deep Markov Model: https://pyro.ai/examples/dmm.html , which can be put into references sections

OlaRonning:
Hi @fehiepsi, I have a [CoLab version](https://colab.research.google.com/drive/17vAcyDXbcuDRj8IUym8zdP0OxN5kpLmD#scrollTo=PYd074KC4Mnh) as well. If you think it's relevant, I can update it to optax and add it.

examples/stein_dmm.py
fehiepsi: I think you put it into the right place.

examples/stein_dmm.py
fehiepsi: I think this `try except` pattern for optional dependency is fine.

examples/stein_dmm.py
```diff
@@ -0,0 +1,340 @@
...
+from jax import nn, numpy as jnp, random
+from optax import adam, chain
+
+import numpyro
+from numpyro.contrib.einstein import SteinVI
+from numpyro.contrib.einstein.kernels import RBFKernel
+import numpyro.distributions as dist
+from numpyro.examples.datasets import JSB_CHORALES, load_dataset
+from numpyro.infer import Predictive, Trace_ELBO
+from numpyro.optim import optax_to_numpyro
```
fehiepsi, line 34: I think you don't need this

fehiepsi (approved):
Looks great to me. Thank you, @OlaRonning!

examples/stein_dmm.py
fehiepsi: Oops, please ignore this.

fehiepsi:
> update it to optax

Maybe a small FR is to support optax directly in SteinVI, like in `SVI`. :) But it depends on you.

examples/stein_dmm.py
OlaRonning: Makes sense. I'll update it.

OlaRonning:
Thanks for the feedback, @fehiepsi. I'll look into supporting optax directly in a separate PR.

fehiepsi (approved):

PR #1300, opened 2022-01-22
Add some doctests to transforms 

wataruhashimoto52:
For some transformations that are not in `torch.distributions.transforms`, I left specific examples as doctests.

PR #1316, opened 2022-02-01
Example: Zero-Inflated Poisson (ZIP) regression model

wataruhashimoto52:
Hello,

I evaluated Zero-Inflated Poisson regression model with ~~MLE and~~ MAP, VI and MCMC in terms of their predictive performances of the number of fish caught by using ~~`statsmodels` and~~ NumPyro. 
I am looking forward to improve this PR based on your feedback!

Thank you!


examples/zero_inflated_poisson.py
```diff
@@ -0,0 +1,136 @@
...
+        num_chains=args.num_chains,
+        progress_bar=False if "NUMPYRO_SPHINXBUILD" in os.environ else True,
+    )
+    mcmc.run(random.PRNGKey(1), X, Y)
+    mcmc.print_summary()
+    return mcmc.get_samples()
+
+
+def fit_statsmodel_zip(args, X, Y):
+    zip_model = sm.ZeroInflatedPoisson(endog=Y, exog=X, exog_infl=X, inflation="logit")
```
martinjankowiak, line 61: i think it would be much nicer if instead of using `sm` you used a `AutoDelta` guide and did maximum likelihood estimation within numpyro

martinjankowiak (commented):
thanks for the contribution! always great to have more examples, especially simpler ones that are more appropriate for beginning users

examples/zero_inflated_poisson.py
```diff
@@ -0,0 +1,165 @@
+# Copyright Contributors to the Pyro project.
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+Example: Zero-Inflated Poisson regression model
+================================================
+
+We evaluate Zero-Inflated Poisson regression model with MAP, VI, and MCMC in terms of their predictive
+performances of the number of fish caught.
```
fehiepsi, line 9: Thanks, @wataruhashimoto52! Could you say a bit about the dataset here and define the problem here (e.g. Here we want to predict the number of fish caught given ... - I'm not sure)?

Looking like you have plots in your script, could you add one similar to [here](https://github.com/pyro-ppl/numpyro/blob/9719466a584af80683a7aa04401d4e065cc569ae/examples/funnel.py#L25) and [this folder](https://github.com/pyro-ppl/numpyro/tree/master/docs/source/_static/img/examples)? In addition, to have this example shown in the docs, you'll need to add an entry for it [in the applications section](https://github.com/pyro-ppl/numpyro/blob/master/docs/source/index.rst).

fehiepsi (approved):
Looks good to me pending further comments from @martinjankowiak!

examples/zero_inflated_poisson.py
```diff
@@ -0,0 +1,171 @@
+# Copyright Contributors to the Pyro project.
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+Example: Zero-Inflated Poisson regression model
+================================================
+
+In this example, we model and predict how many fish visitor groups to a state park catch.
```
martinjankowiak, line 8: nit:

In this example, we model and predict how many fish are caught by visitors to a state park.
Many groups of visitors catch zero fish, either because they did not fish at all or because they were unlucky. 
We would like to explicitly this bimodal behavior (zero versus non-zero) and ascertain which variables contribute to each behavior.

We answer this question by fitting a zero-inflated poisson regression model. We use MAP,
VI and MCMC as estimation methods. Finally, from the MCMC samples, we identify the variables that
contribute to the zero and non-zero components of the zero-inflated poisson likelihood.

examples/zero_inflated_poisson.py
```diff
@@ -0,0 +1,171 @@
...
+
+    def add_fig(var_name, title, ax):
+        ax.set_title(title)
+        ax.violinplot(
+            [posterior_samples[var_name][:, i].to_py() for i in range(len(cols))]
+        )
+        ax.set_xticks(np.arange(1, len(cols) + 1))
+        ax.set_xticklabels(cols, rotation=45, fontsize=10)
+
+    add_fig("b1", "coeffs for probability of catching fishes", axes[0])
```
martinjankowiak, line 150: nit: Coefficients for probability of catching fish

examples/zero_inflated_poisson.py
```diff
@@ -0,0 +1,171 @@
...
+    add_fig("b2", "coeffs for the number of fishes caught", axes[1])
```
martinjankowiak, line 151: nit: Coefficients for the number of fish caught

martinjankowiak (approved):
looks great thanks! added some suggested language changes

wataruhashimoto52:
@martinjankowiak I am sorry to trouble you, but could you edit this script for these suggestions I have received?
If I edit the code and push it, the approval will be canceled as shown in #1300 .

martinjankowiak (approved):

PR #1320, opened 2022-02-04
Fix provenance logic for jax 0.2.28

### LINKED ISSUE #1318: `numpyro.render_model` behaviour changed in 0.9.0
I noticed that the behaviour of `numpyro.render_model` appears to have changed since the latest release, specifically arrows signifying conditional dependence are not being rendered properly.

I haven't had time to dig into why this is yet, but here is a simple recreation

```python
import numpyro


def model():
    theta = numpyro.sample("theta", dist.Uniform())
    
    with numpyro.plate("N", size=10):
        numpyro.sample("y", dist.Bernoulli(theta))
        
        
if __name__ == "__main__":
    g = numpyro.render_model(model)
    g.render(f"model-{numpyro.__version__}", format="png")
```

Running this with 0.8.0 and 0.9.0 I get (respectively)

model-0 8 0

model-0 9 0

fehiepsi:
Fixes #1318

tldr; I have tested the change and it works for the old jax versions.

The recent jax release made provenance logic failing because the cached abstract value might be a different object w.r.t. the one stored in the output of `process_primitive`, while we only updated `provenance` for the output previously.

I think we'll need to make a patch release after this fix.

### DIFF
--- a/numpyro/ops/provenance.py
+++ b/numpyro/ops/provenance.py
@@ -26,6 +26,9 @@ class _ProvenanceJaxprTrace(partial_eval.DynamicJaxprTrace):
             out_tracers = out_tracers if primitive.multiple_results else [out_tracers]
             for t in out_tracers:
                 t.aval.named_shape["_provenance"] = out_provenance
+                # Also update provenance of the cached tracer -> aval dict.
+                aval_cache = self.frame.tracer_to_var[id(t)].aval
+                aval_cache.named_shape["_provenance"] = out_provenance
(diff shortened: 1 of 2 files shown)


fritzo (approved):
Nice, that's subtle!

Do you think it's worth adding a regression test, e.g. the model added in the issue?

fehiepsi:
Hi Fritz, all the tests involving provenance failed so we have to pin the jax version to make CI pass. In this pr, I just removed the restriction because the tests pass again.

fehiepsi:
Thanks for reviewing! :)

PR #1322, opened 2022-02-04
Fix stein docs rendering

fehiepsi:
Fixes some rendering issues of [SteinVI docs](https://num.pyro.ai/en/latest/contrib.html#stein-variational-inference).

### DIFF
--- a/notebooks/source/bayesian_hierarchical_stacking.ipynb
+++ b/notebooks/source/bayesian_hierarchical_stacking.ipynb
@@ -1231,13 +1231,6 @@
    "parameters": {},
    "start_time": "2021-09-18T10:50:40.734722",
    "version": "2.3.3"
-  },
-  "widgets": {
-   "application/vnd.jupyter.widget-state+json": {
-    "state": {},
-    "version_major": 2,
-    "version_minor": 0
-   }
   }
  },
  "nbformat": 4,
--- a/notebooks/source/bayesian_hierarchical_stacking.ipynb
+++ b/notebooks/source/bayesian_hierarchical_stacking.ipynb
@@ -21,7 +21,7 @@
     "  <figcaption>Photo by Belinda Fewings, https://unsplash.com/photos/6p-KtXCBGNw.</figcaption>\n",
     "</figure> \n",
     "\n",
-    "### Table of Contents\n",
+    "## Table of Contents\n",
     "\n",
     "* [Intro](#intro)\n",
     "* [1. Exploratory Data Analysis](#1)\n",
@@ -1204,9 +1204,9 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "good-bayesian",
+   "display_name": "Python 3 (ipykernel)",
    "language": "python",
-   "name": "good-bayesian"
+   "name": "python3"
   },
   "language_info": {
    "codemirror_mode": {
@@ -1218,7 +1218,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.8.5"
+   "version": "3.9.7"
   },
   "papermill": {
    "default_parameters": {},
@@ -1231,6 +1231,13 @@
    "parameters": {},
    "start_time": "2021-09-18T10:50:40.734722",
    "version": "2.3.3"
+  },
+  "widgets": {
+   "application/vnd.jupyter.widget-state+json": {
+    "state": {},
+    "version_major": 2,
+    "version_minor": 0
+   }
   }
  },
  "nbformat": 4,
--- a/numpyro/contrib/einstein/kernels.py
+++ b/numpyro/contrib/einstein/kernels.py
@@ -47,10 +47,12 @@ class SteinKernel(ABC):
         Computes the kernel function given the input Stein particles
 
         :param particles: The Stein particles to compute the kernel from
-        :param particle_info: A mapping from parameter names to the position in the particle matrix
+        :param particle_info: A mapping from parameter names to the position in the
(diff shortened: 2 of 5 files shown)


notebooks/source/bayesian_hierarchical_stacking.ipynb
```diff
@@ -21,7 +21,7 @@
     "  <figcaption>Photo by Belinda Fewings, https://unsplash.com/photos/6p-KtXCBGNw.</figcaption>\n",
     "</figure> \n",
     "\n",
-    "### Table of Contents\n",
+    "## Table of Contents\n",
```
fehiepsi, line 24: This one makes the tutorial sections not [arranged correctly](https://num.pyro.ai/en/latest/tutorials/bayesian_hierarchical_stacking.html#Table-of-Contents).

numpyro/infer/hmc.py
```diff
@@ -199,7 +199,7 @@ def init_kernel(
         regularize_mass_matrix=True,
         model_args=(),
         model_kwargs=None,
-        rng_key=random.PRNGKey(0),
+        rng_key=None,
```
fehiepsi, line 202: This change is used to avoid invoking jax operations when rendering the docs. This makes doc rendering faster.

numpyro/contrib/einstein/kernels.py
```diff
@@ -412,14 +425,16 @@ def kernel(x, y):
 
 class GraphicalKernel(SteinKernel):
     """
-    Calculates graphical kernel :math: `k(x,y) = diag({K^(l)(x,y)}_l)` from [1].
+    Calculates graphical kernel :math:`k(x,y) = diag({K^(l)(x,y)}_l)` (TODO: correct this formula) from [1].
```
fehiepsi, line 428: @OlaRonning Could you help me correct this formula? I couldn't find this kernel in the reference.

OlaRonning: @fehiepsi the corresponds is established (page 4 last paragraph before section 3.3) in 
```
     2. *Stein Variational Gradient Descent with Matrix-Valued Kernels* by Wang, Tang, Bajaj, and Liu
```

numpyro/infer/hmc.py
OlaRonning: I see.

numpyro/contrib/einstein/kernels.py
fehiepsi: I see, I think we need to modify :math:`k(x,y) = diag({K^(l)(x,y)}_l)` to `k(x,y) = diag(K_l(x,y))` (the current one does not rendered correctly so I'm not sure how to rewrite the formula into the one that you prefer? Maybe we need to say what is K and l too.

fehiepsi: I think I should make the comment clearer. In latex, `K^(l)` will not be the same as K^l

OlaRonning (commented):
Thanks for fixing the rendering, @fehiepsi! 

I have not checked the docs. When I built it, I got a TypeError.

Reproduce:

``` python
from numpyro.contrib.tfp import distributions

if __name__ == '__main__':
    print('hi')
```

numpyro/contrib/einstein/kernels.py
OlaRonning: `k(x,y) = diag(K_l(x,y))` is good. `K^(l)` should be `K^{(l)}` I suppose.

numpyro/contrib/einstein/kernels.py
```diff
@@ -66,18 +68,21 @@ def init(self, rng_key, particles_shape):
 class RBFKernel(SteinKernel):
     """
     Calculates the Gaussian RBF kernel function, from [1],
-    :math: `k(x,y) = \\exp(\\frac{1}{h} \\|x-y\\|^2)`,
+    :math:`k(x,y) = \\exp(\\frac{1}{h} \\|x-y\\|^2)`,
     where the bandwidth h is computed using the median heuristic
-    :math: `h = \\frac{1}{\\log(n)} \\med(\\|x-y\\|)`.
+    :math:`h = \\frac{1}{\\log(n)} \\med(\\|x-y\\|)`.
```
OlaRonning, line 73: `\med` does not render. Maybe just use `\text{med}` instead.

numpyro/contrib/einstein/kernels.py
OlaRonning: How about:
``` python
    """
    Calculates graphical kernel :math:`k(x,y) = diag({K_l(x_l,y_l)})` for local kernels :math:`K_l` from [1][2].

    **References:**

    1. *Stein Variational Message Passing for Continuous Graphical Models* by Wang, Zheng, and Liu
    2. *Stein Variational Gradient Descent with Matrix-Valued Kernels* by Wang, Tang, Bajaj, and Liu

    :param local_kernel_fns: A mapping between parameters and a choice of kernel
        function for that parameter (default to default_kernel_fn for each parameter)
    :param default_kernel_fn: The default choice of kernel function when none is
        specified for a particular parameter
    """
```

OlaRonning (dismissed):
I finally got it to render. Other than the `\\med` not rendering, it looks good. Thanks for fixing this again. I've posted a suggestion for the graphical kernel docstring.

numpyro/contrib/einstein/kernels.py
fehiepsi: Done, thanks for suggesting!

fehiepsi:
Thanks for reviewing and suggestions, Ola!

OlaRonning (approved):
LGTM

Issue #1323, opened 2022-02-04
Bayesian Hierarchical Stacking Example


Vinnie-Palazeti:
Hello,

The Bayesian Hierarchical Stacking example is not functioning. I tried in both the linked Google Colab & locally in jupyter lab.

Docs:
https://num.pyro.ai/en/latest/tutorials/bayesian_hierarchical_stacking.html

Link:
https://github.com/pyro-ppl/numpyro/tree/master/notebooks/source)/bayesian_hierarchical_stacking.ipynb

Error:

```

AssertionError                            Traceback (most recent call last)
[/usr/local/lib/python3.7/dist-packages/numpyro/handlers.py](https://localhost:8080/#) in postprocess_message(self, msg)
    157             msg["type"] in ("sample", "deterministic") and msg["name"] in self.trace
    158         ), "all sites must have unique names but got `{}` duplicated".format(
--> 159             msg["name"]
    160         )
    161         self.trace[msg["name"]] = msg.copy()

AssertionError: all sites must have unique names but got `w_test` duplicated

```

The issue stems from the test portion of the stacking() function

```
    if test:
        # test set stacking weights (in unconstrained space)
        f_test = jnp.hstack([X_test @ beta.T, jnp.zeros((N_test, 1))])
        # test set stacking weights (constrained to sum to 1)
        w_test = numpyro.deterministic("w_test", jax.nn.softmax(f_test, axis=1))
        numpyro.deterministic("w_test", w_test)

```

fehiepsi:
cc @MarcoGorelli  :)

MarcoGorelli:
Thanks for the ping! I'll take a look

Vinnie-Palazeti:
@MarcoGorelli 

I believe it is as simple as changing the variable name inside the numpyro.deterministic(). I am new to numpyro, so can't give a detailed reason why.

this `if test` section at bottom of stacking() function seems to work as intended:

```
def stacking(
    X,
    d_discrete,
    X_test,
    exp_lpd_point,
    tau_mu,
    tau_sigma,
    *,
    test,
):
    """
    Get weights with which to stack candidate models' predictions.

    Parameters
    ----------
    X
        Training stacking matrix: features on which stacking weights should depend, for the
        training set.
    d_discrete
        Number of discrete features in `X` and `X_test`. The first `d_discrete` features
        from these matrices should be the discrete ones, with the continuous ones coming
        after them.
    X_test
[...]

PR #1327, opened 2022-02-08
removed last line of stacking function

Vinnie-Palazeti:
This is in regard to Issue 1323

https://github.com/pyro-ppl/numpyro/issues/1323

Only change is last line of stacking function is removed. It is redundant and prevents the notebook from functioning due to duplicative 'w_test' name.


fehiepsi (approved):
Thanks for the fix, @Vinnie-Palazeti!

PR #1330, opened 2022-02-09
Vectorized sampling: Calculate var shape manually.

### LINKED ISSUE #1328: Vectorized sampling: error when var size is 0
JAX refuses to reshape an array using the `(-1, ...)` shape notation if its `size` (i.e. product of `shape`) is 0. This causes issues with vectorized sampling when samples from different chains are concatenated.

# Steps to reproduce
```python
import numpy as np
import numpyro

from jax import random

J = 8
n = 0
y = np.random.rand(n, J)
sigma = np.array([15.0, 10.0, 16.0, 11.0, 9.0, 11.0, 10.0, 18.0])


# Slightly modified Eight Schools example

def eight_schools(J, sigma, y=None):
    mu = numpyro.sample('mu', dist.Normal(0, 5))
    tau = numpyro.sample('tau', dist.HalfCauchy(5))
    with numpyro.plate('J', J):
        theta = numpyro.sample('theta', dist.Normal(mu, tau))
        numpyro.sample('obs', dist.Normal(theta, sigma), sample_shape=(n,))

nuts_kernel = NUTS(eight_schools)

mcmc = MCMC(nuts_kernel, num_warmup=50, num_samples=100, num_chains=2, chain_method='vectorized')

rng_key = random.PRNGKey(0)

mcmc.run(rng_key, J, sigma, y=y, extra_fields=('potential_energy',))
```

Sampling of individual chains works fine but a `[TypeError: reshape new_sizes must all be positive, got (-1, 0, 8).]()` is raised during final concatenation of samples.

# Possible fix
In `mcmc.py:612`

```python
states_flat = tree_map(
    lambda x: jnp.reshape(x, (-1,) + x.shape[2:]), states
)
```

simply calculating the dimension of the first axis is sufficient

```python
states_flat = tree_map(
    lambda x: jnp.reshape(x, (x.shape[0] * x.shape[1],) + x.shape[2:]), states
)
```

It seems unlikely that the issue can be "fixed" upstream in jax, since numpy has the same policy towards reshaping. Happy to open a PR with the above fix if it seems reasonable.

hessammehr:
Fixes #1328

### DIFF
--- a/test/infer/test_mcmc.py
+++ b/test/infer/test_mcmc.py
@@ -1080,3 +1080,29 @@ def test_discrete_site_without_infer_enumerate():
     mcmc = MCMC(NUTS(model), num_warmup=10, num_samples=10)
     with pytest.warns(FutureWarning, match="enumerated sites"):
         mcmc.run(random.PRNGKey(0))
+
+
+def test_vectorized_sampling_zero_sized():
+    J = 8
+    n = 0
+    sigma = np.array([15.0, 10.0, 16.0, 11.0, 9.0, 11.0, 10.0, 18.0])
+
+    # Modified Eight Schools example
+    def eight_schools(J, sigma):
+        mu = numpyro.sample("mu", dist.Normal(0, 5))
+        tau = numpyro.sample("tau", dist.HalfCauchy(5))
+        with numpyro.plate("J", J):
+            theta = numpyro.sample("theta", dist.Normal(mu, tau))
+            numpyro.sample("scores", dist.Normal(theta, sigma), sample_shape=(n,))
+
+    nuts_kernel = NUTS(eight_schools)
+    mcmc = MCMC(
+        nuts_kernel,
+        num_warmup=50,
+        num_samples=100,
+        num_chains=2,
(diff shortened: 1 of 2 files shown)


test/infer/test_mcmc.py
```diff
@@ -1080,3 +1080,30 @@ def model():
...
+    nuts_kernel = NUTS(eight_schools)
+    mcmc = MCMC(
+        nuts_kernel,
+        num_warmup=50,
+        num_samples=100,
+        num_chains=2,
+        chain_method="vectorized",
+    )
+    rng_key = random.PRNGKey(0)
+    mcmc.run(rng_key, J, sigma, y=y, extra_fields=("potential_energy",))
```
fehiepsi, line 1108: typo: no `y` and `y = np.random.rand(n, J)` above.

hessammehr: Sorry, I tried to simplify the example in the issue but forgot about `y` - fixed now.

fehiepsi:
Thanks @hessammehr!

Issue #1319, opened 2022-02-03
Working example for Bayesian Quantile Regression and AsymmetricLaplace Distribution


d-diaz:
I've prepared a simple working example of Bayesian Quantile Regression using an implementation of an AsymmetricLaplace distribution in a Colab Notebook [here](https://colab.research.google.com/drive/1h4yoaIYcDRHy6XYDtMgrUUCB0fJ1uf0v?usp=sharing). I'd like to offer this up as an example for numpyro documentation, but am not sure whether this needs to be converted to script or if it can be utilized as a notebook.

fehiepsi:
Hi @d-diaz I think it is best to have asymmetric laplace distribution in the library and a tutorial (which might be best to explain how quantile regression works and how to use numpyro to such problem) or an example script with current content. Using asymmetric parameter will match wikipedia article but you can make a wrapper like a function `make_lilelihood(..., quantile)` for your convenience in example/tutorial. Thanks for contributing!

fehiepsi:
It seems to me that the scale parameter lambda in the main wikipedia article should be the rate. Scipy also uses asymetry parameterization. If you find quantile argument useful, you can define a new distribution named AssymetryLaplaceQuantile or something like that. :)

d-diaz:
I think it seems most sensible to implement an AssymetricLaplace class following the Scipy and Pyro and Wikipedia pattern and a distinct AsymmetricLaplaceQuantile class following the Bayesian quantile regression pattern. I don’t know the use cases for the AsymmetricLaplace with asymmetry parameterization, but did run into a the need for this quantile regression version in my own work. I’ll expand the minimal working example in Colab into a notebook tutorial around quantile regression and add these two classes into the library. Will report back here when that’s done.

fehiepsi:
That's an awesome plan, thanks @d-diaz!

PR #1332, opened 2022-02-10
AsymmetricLaplace distributions

d-diaz:
Implementations of AsymmetricLaplace following the Pyro implementation, as well as an AsymmetricLaplaceQuantile class following alternative parameterization for Bayesian quantile regression. Related to #1319.

### DIFF
--- a/numpyro/distributions/continuous.py
+++ b/numpyro/distributions/continuous.py
@@ -124,86 +124,56 @@ class AsymmetricLaplace(Distribution):
             value <= self.loc,
             self.asymmetry**2
             / (1 + self.asymmetry**2)
-            * jnp.exp((self.scale / self.asymmetry) * (value - self.loc)),
+            * jnp.exp(self.right_scale * (value - self.loc)),
             1
             - 1
             / (1 + self.asymmetry**2)
-            * jnp.exp(-self.scale * self.asymmetry * (value - self.loc)),
+            * jnp.exp(-self.left_scale * (value - self.loc)),
+        )
+
+    def icdf(self, value):
+        temp = self.asymmetry**2 / (1 + self.asymmetry**2)
+        return jnp.where(
+            value <= temp,
+            self.loc + self.left_scale * jnp.log(value / temp),
+            self.loc
+            - self.right_scale * (jnp.log1p(self.asymmetry**2) + jnp.log1p(-value)),
         )
 
 
-class AsymmetricLaplaceQuantile(Distribution):
+class AsymmetricLaplaceQuantile(AsymmetricLaplace):
+    """An alternative parameterization of AsymmetricLaplace commonly applied in
+    Bayesian quantile regression.
+
+    Instead of the `asymmetry` parameter employed by AsymmetricLaplace, to
+    define the balance between left- versus right-hand sides of the
+    distribution, this class utilizes a `quantile` parameter, which describes
+    the proportion of probability density that falls to the left-hand side of
+    the distribution.
+
+    The `scale` parameter is also interpreted slightly differently than in
+    AsymmetricLaplce. When `loc=0` and `scale=1`, AsymmetricLaplace(0,1,1)
+    is equivalent to Laplace(0,1), while AsymmetricLaplaceQuantile(0,1,0.5) is
+    equivalent to Laplace(0,2).
+    """
+
     arg_constraints = {
         "loc": constraints.real,
         "scale": constraints.positive,
         "quantile": constraints.interval(0.0, 1.0),
     }
-    reparametrized_params = ["loc", "scale"]
+    reparametrized_params = ["loc", "qscale"]
     support = constraints.real
 
     def __init__(self, loc=0.0, scale=1.0, quantile=0.5, validate_args=None):
-        self.loc, self.scale, self.quantile = promote_shapes(loc, scale, quantile)
+        asymmetry = jnp.sqrt(1 / ((1 / quantile) - 1))
+        scale = scale * asymmetry / quantile
+        self.loc, self.scale, self.asymmetry = promote_shapes(loc, scale, asymmetry)
         batch_shape = lax.broadcast_shapes(
-            jnp.shape(loc), jnp.shape(scale), jnp.shape(quantile)
+            jnp.shape(loc), jnp.shape(scale), jnp.shape(asymmetry)
         )
-        super(AsymmetricLaplaceQuantile, self).__init__(
+        super(AsymmetricLaplace, self).__init__(
             batch_shape=batch_shape, validate_args=validate_args
-        )
-
-    def log_prob(self, value):
-        # following Yu and Moyeed (2001)
-        if self._validate_args:
-            self._validate_sample(value)
-
-        const = self.quantile * (1 - self.quantile) / self.scale
-        z = (value - self.loc) / self.scale
-        check = z * jnp.where(value <= self.loc, -(1 - self.quantile), self.quantile)
-        return jnp.log(const * jnp.exp(-check))
-
-    def sample(self, key, sample_shape=()):
-        # mixture of exponentials following Kozumi and Kobyashi (2009)
-        assert is_prng_key(key)
-        shape = (2,) + sample_shape + self.batch_shape + self.event_shape
-        u, v = random.exponential(key, shape=shape) * self.scale
-        e = u / self.quantile - v / (1 - self.quantile)
-        return self.loc + e
-
-    @property
-    def mean(self):
-        return (
-            self.loc
-            + (1 - 2 * self.quantile)
-            / (self.quantile * (1 - self.quantile))
-            * self.scale
-        )
-
-    @property
-    def variance(self):
-        return (
-            (1 - 2 * self.quantile + 2 * self.quantile**2)
-            / (self.quantile**2 * (1 - self.quantile) ** 2)
-            * self.scale**2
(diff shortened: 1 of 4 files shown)


numpyro/distributions/continuous.py
```diff
@@ -67,6 +66,147 @@
...
+            / (1 + self.asymmetry**2)
+            * jnp.exp((self.scale / self.asymmetry) * (value - self.loc)),
+            1
+            - 1
+            / (1 + self.asymmetry**2)
+            * jnp.exp(-self.scale * self.asymmetry * (value - self.loc)),
+        )
+
+
+class AsymmetricLaplaceQuantile(Distribution):
```
fehiepsi, line 135: Maybe we can add docstrings to clarify the difference between two distributions? Something like given p follows AsymmetricLaplace and q follows AsymmetricLaplaceQuantile, p = q if
```
# assume without miscalculation
q.loc = p.loc, q.scale = p.scale * q.quantile / p.asymmetry, q.quantile = 1 / (1 + 1 / p.asymmetry ** 2)
```
and mention that in this distribution, quantile `quantile` of this distribution is `loc` (not sure how to describe this properly).

fehiepsi (commented):
Thanks, @d-diaz! Could you add new entries to [docs](https://github.com/pyro-ppl/numpyro/blob/master/docs/source/distributions.rst) and [tests](https://github.com/pyro-ppl/numpyro/blob/master/test/test_distributions.py#L198) too?

numpyro/distributions/continuous.py
```diff
@@ -67,6 +66,147 @@
...
+        p = left / total
+        q = right / total
+        return p * left**2 + q * right**2 + p * q * total**2
+
+    def cdf(self, value):
+        return jnp.where(
+            value <= self.loc,
+            self.asymmetry**2
+            / (1 + self.asymmetry**2)
+            * jnp.exp((self.scale / self.asymmetry) * (value - self.loc)),
```
fehiepsi, line 127: You can use `left_scale` and `right_scale` here. I guess we can also derive similar `icdf` method given two parameterizations are equivalent.

numpyro/distributions/continuous.py
d-diaz: I'm not sure there is an intuitive translation between the two parameterizations, but can play around with sympy for a bit to see if there's a straightforward conversion of `quantile` into `asymmetry`. In both distributions, `location` and `scale` have the same usual intent. Both parameterizations are represented as a mixture of two exponential distributions, and `asymmetry` and `quantile` are just two different ways of defining how much probability density occurs under the left vs. right components of the mixture. 

At the very least, I think it would make sense to add a docstring to the `AsymmetricLaplaceQuantile` class to describe that this class is intended for Bayesian quantile regression, and that quantile refers to the desired quantile of the data against which the model is being fit.

fehiepsi: The scale parameters in two distributions are a bit difference. With AsymLaplace(0,1,1) is Laplace(0,1) while AsymLaplaceQuantile(0,1,0.5) is Laplace(0,2). Maybe it is also worth to mention that to avoid confusions.

fehiepsi: Looking at your sample method, we have
```
q.right_scale = q.scale / q.quantile = p.scale / p.asymetry = p.right_scale
q.left_scale = q.scale / (1 - q.quantile) = q.scale / q.quantile * q.quantile / (1 - q.quantile)
= (using the relation of quantile and asymmetry in my last comment)
= p.scale / p.asymetry * p.asymmetry ** 2 = p.left_scale
```
Hope it help.

d-diaz: Ok. Given the math you've done to show how to translate q (ALDQ) to p (ALD), I was able to redefine `AsymmetricLaplaceQuantile` as a child class of `AsymmetricLaplace`, calculating the conversion to `p.scale` from `q.scale` and calculating `p.asymmetry` from `q.scale` and `q.quantile`. 

I've also figured out the equation for `icdf` method for ALD, and have added that. Under the hood, ALDQ now just re-parameterizes the user input to use standard ALD parameterization, so no redundant code reproducing any of ALD's methods.

numpyro/distributions/continuous.py
```diff
@@ -67,6 +66,117 @@
 )
 
 
+class AsymmetricLaplace(Distribution):
+    arg_constraints = {
+        "loc": constraints.real,
+        "scale": constraints.positive,
+        "asymmetry": constraints.positive,
+    }
+    reparametrized_params = ["loc", "scale"]
```
fehiepsi, line 75: Looking at the sampler, I think all params are reparameterized (with base dists are exponentials).

numpyro/distributions/continuous.py
```diff
@@ -67,6 +66,117 @@
...
+        "loc": constraints.real,
+        "scale": constraints.positive,
+        "quantile": constraints.interval(0.0, 1.0),
+    }
+    reparametrized_params = ["loc", "scale"]
+    support = constraints.real
+
+    def __init__(self, loc=0.0, scale=1.0, quantile=0.5, validate_args=None):
+        asymmetry = jnp.sqrt(1 / ((1 / quantile) - 1))
+        scale = scale * asymmetry / quantile
```
fehiepsi, line 170: Do you think this will cause confusion - we init with some scale but the distribution stores another scale? I think it is best to keep scale as-is.

d-diaz: Makes sense. This was something I originally experimented with when trying to have ALDQ inhert from ALD, but I think that broke tests when attempting to call super init methods. Will revert back to keeping a `self._ald = AsymmetricLaplace(...)` and then calling `self._ald` methods for `log_prob`, `cdf`, etc. I've been struggling to install the development environment using pip, it keeps breaking on scipy installation, but I think I've got it fixed now, so should be able to run tests locally until they pass before pushing more changes.

d-diaz:
Alright! Finally figured out how to replicate batching dimensions and error checking from looking at other distributions like `StudentsT`, and fixed a couple bugs in the `cdf` and `icdf` methods, so all tests are now passing. 

I've gotten a better understanding of what is being tested in `test_distributions.py`, but want to make sure that the updates I made there are also appropriate. In particular, would like to confirm changes I made to `test_sample_gradient` to include the `quantile` and `asymmetry` params from ALDQ and ALD in the `gamma_derived_params` are justified.

numpyro/distributions/continuous.py
```diff
@@ -67,6 +66,143 @@
...
+class AsymmetricLaplace(Distribution):
+    arg_constraints = {
+        "loc": constraints.real,
+        "scale": constraints.positive,
+        "asymmetry": constraints.positive,
+    }
+    reparametrized_params = ["loc", "scale", "asymmetry"]
+    support = constraints.real
+
+    def __init__(self, asymmetry=1.0, loc=0.0, scale=1.0, validate_args=None):
```
fehiepsi, line 78: Can we use the same order as in Pyro, to avoid confusion?

test/test_distributions.py
```diff
@@ -866,6 +880,8 @@ def test_sample_gradient(jax_dist, sp_dist, params):
         "LKJ": ["concentration"],
         "LKJCholesky": ["concentration"],
         "StudentT": ["df"],
+        "AsymmetricLaplace": ["asymmetry"],
+        "AsymmetricLaplaceQuantile": ["quantile"],
```
fehiepsi, line 884: I think those parameters are reparametrizable. Did the test fail with those parameters? I can take a look if it is the case.

d-diaz: The distributions did fail this test earlier, but it might have been for another reason. Will comment out these lines and rerun the test to confirm.

numpyro/distributions/continuous.py
d-diaz: Sure. I wrote it this way to be consistent with `StudentT(df, loc, scale)` and with the scipy implementation `ALD(asym, loc, scale)`. I can change it back to follow Pyro if you prefer that.

fehiepsi: I see. Yeah, I think matching Pyro behavior is preferred. It will help users change their codes easier (and avoid hidden issues due to the wrong order)

test/test_distributions.py
d-diaz: When I comment out those lines, ALDQ fails with all three sets of parameters tested, but ALD still passes. I think it probably has to do with the calculation of `asymmetry` from reparameterized values of `scale` and/or `quantile`. The test shows two different error messages, one on the line for the calculation of `asymmetry`, giving a `ZeroDivisionError`, and the next two tests show an `AssertionError` where x has a nan value, and y has an imaginary complex number.

fehiepsi: > one on the line for the calculation of asymmetry, giving a ZeroDivisionError

Seems like this will happen at `asymmetry = (1 / ((1 / quantile) - 1)) ** 0.5` when quantile=0 or quantile=1. I will take a look.

d-diaz: Seems like it's passing with lines commented out now, so will remove these...

fehiepsi (approved):
Thanks, @d-diaz!

test/test_distributions.py
fehiepsi: Sorry for that, it's a problem of the test (which assumes we have the right order of reparam parameters).

Issue #1176, opened 2021-10-07
Scope handler doesn't add prefix to `cond_indep_stack` frames


amalvaidya:
Hi!

I'm working on building some tooling to help run model diagnostics and part of that involves extracting the dimensions/coordinates assoicated with each sample site. This is easy enough to do using the trace but I run into problems when using the scope handler.

The handler will add a prefix to the name of sample sites and the plate primitives in the model trace. What it doesn't do however is prepend the prefix to the name of the frame associated with the plate in the `cond_indep_stack`.

An example of the issue is below

```python
means = np.random.normal(size=(5,))
scales = np.abs(np.random.normal(size=(5,)))
data = np.random.normal(means[:, None] * np.ones(100), scales[:, None] * np.ones(100))

def model(y):
    n_groups = y.shape[0]
    n_trials = y.shape[1]
    
    group_plate =  numpyro.plate("group", n_groups, dim=-1)
    trial_plate = numpyro.plate("trial", n_trials, dim=-2)
    
    with group_plate:
        mu = numpyro.sample("mu", dist.Normal(0, 1))
        sigma = numpyro.sample("sigma", dist.HalfNormal(1))

    with group_plate, trial_plate:
        numpyro.sample("y", dist.Normal(mu, sigma), obs=y.T )

scope_model = numpyro.handlers.scope(model, prefix="scope")

trace = numpyro.handlers.trace(numpyro.handlers.seed(model, 0)).get_trace(data)
scoped_trace = numpyro.handlers.trace(numpyro.handlers.seed(scope_model, 0)).get_trace(data)

# Can identify the plate associated with site mu
assert trace['mu']['cond_indep_stack'][0].name in trace

# Can no longer run the same check when using the scope handler 
assert scoped_trace["scope/mu"]["cond_indep_stack"][0].name in scoped_trace
```

I think there's an easy way to ammend the scope handler to fix this, if this is a desired feature, and I'm happy to submit a PR. I've mainly been focused on MCMC so if this is a problem for running SVI I would be keen to learn more.

fehiepsi:
Nice catch! I think it is a bug. Happy to review your PR. :)

PR #1338, opened 2022-02-17
Do not rename plates under scope handler

### LINKED ISSUE #1335: Add option to exclude `plate` from `scope` handler
Currently, `scope` handler changes the name of all messages. I found that it will cause a bit confusing so it would be nice to have an option to exclude some message types. In the trace, we only raise issues when `sample` and `deterministic` messages are duplicated, so it makes sense to only rename them by default.

fehiepsi:
Fixes #1335

I think the current behavior is confusing and not originally intended. It is unnecessary to have `scope` for the plate (which has caused some [confusion](https://github.com/pyro-ppl/numpyro/issues/1176)). I'm not sure if anyone wants to apply `scope` for `param` but it might be good to allow such behavior (and to use a single parameter, we can simply call `param` once and pass that param around different scopes).

### DIFF
--- a/examples/hsgp.py
+++ b/examples/hsgp.py
@@ -221,7 +221,7 @@ def approx_se_ncp(x, alpha, length, L, M):
     """
     phi = eigenfunctions(x, L, M)
     spd = jnp.sqrt(diag_spectral_density(alpha, length, L, M))
-    with plate("basis", M):
+    with plate(f"basis_{M}", M):
         beta = sample("beta", dist.Normal(0, 1))
 
     f = deterministic("f", phi @ (spd * beta))
--- a/examples/hsgp.py
+++ b/examples/hsgp.py
@@ -221,7 +221,7 @@ def approx_se_ncp(x, alpha, length, L, M):
     """
     phi = eigenfunctions(x, L, M)
     spd = jnp.sqrt(diag_spectral_density(alpha, length, L, M))
-    with plate(f"basis_{M}", M):
+    with plate(f"basis", M):
         beta = sample("beta", dist.Normal(0, 1))
 
     f = deterministic("f", phi @ (spd * beta))
--- a/examples/hsgp.py
+++ b/examples/hsgp.py
@@ -221,7 +221,7 @@ def approx_se_ncp(x, alpha, length, L, M):
     """
     phi = eigenfunctions(x, L, M)
     spd = jnp.sqrt(diag_spectral_density(alpha, length, L, M))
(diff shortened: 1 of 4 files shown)


examples/hsgp.py
```diff
@@ -221,7 +221,7 @@ def approx_se_ncp(x, alpha, length, L, M):
     """
     phi = eigenfunctions(x, L, M)
     spd = jnp.sqrt(diag_spectral_density(alpha, length, L, M))
-    with plate("basis", M):
+    with plate(f"basis_{M}", M):
```
fehiepsi, line 224: This `basic` plate has different sizes in different scopes. So we need to add `M` here to fix the issue. This raises some questions around what would be a reasonable behavior of `scope`.

numpyro/handlers.py
```diff
@@ -625,25 +619,22 @@ class scope(Messenger):
     :param fn: Python callable with NumPyro primitives.
     :param str prefix: a string to prepend to sample names
     :param str divider: a string to join the prefix and sample name; default to `'/'`
+    :param bool skip_param: whether to apply scope for `param` sites. Defaults to True.
     """
 
-    def __init__(self, fn=None, prefix="", divider="/"):
+    def __init__(self, fn=None, prefix="", divider="/", *, skip_param=True):
```
fehiepsi, line 625: Not sure which argument we need to add here. Probably we can replace this with `optional_types` which allows users to specify additional types to apply scope to, e.g. `params` or `plate`.

eb8680 (approved):
This change seems fine because it brings numpyro in line with Pyro. I actually think the right behavior for `scope` is to apply to all named primitives (not just `sample` as in Pyro), but the implementations aren't ready to support that and there are other changes that would probably need to be made alongside that (e.g. adding an option to make parameter stores local to functions or modules).

fehiepsi:
My main motivation is to make it easier to inspect markov models [here](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/contrib/control_flow/scan.py#L156) but thanks to your input, I guess we can walk away with an optional argument like `hide_types` in Pyro's `block` handler. Let me sketch that.

fehiepsi:
Could you take a look, @eb8680! The new argument seems to provide the flexibility that I needed.

eb8680 (approved):
Seems like a reasonable interface change. Note that Pyro default behavior is equivalent to `hide_types=["plate", "param"]`

fehiepsi:
Thanks for reviewing, Eli!

PR #1343, opened 2022-02-20
Support multithreading for PYRO_STACK

fehiepsi:
Based on the discussion with @fritzo and @eb8680, it seems that the simplest way to support multithreading is to get PYRO_STACK from an intermediate function `get_pyro_stack`, where we can modify to return a stack for a particular thread.

TODOs:
- [ ] see how to do threading in jax
- [ ] finish the example

### DIFF
--- a/examples/threading.py
+++ b/examples/threading.py
@@ -1,0 +1,33 @@
+"""
+Examples to show how to use numpyro with multithreading.
+"""
+from collections import defaultdict
+import threading
+
+import jax
+import jax.numpy as jnp
+
+import numpyro
+from numpyro import handlers
+import numpyro.distributions as dist
+from numpyro.primitives import set_pyro_stack
+
+
+class _StackThreadDict(defaultdict):
+    def current_stack(self):
+        thread_id = threading.get_native_id()
+        return self[thread_id]
+
+
+_PYRO_THREAD_STACK = _StackThreadDict(list)
+set_pyro_stack(_PYRO_THREAD_STACK)
+
+
+def model():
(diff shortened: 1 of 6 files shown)


chkothe:
There's actually a much simpler way to accomplish basically the same, but without any code changes to either user code or any Messengers or other numpyro internals that have to interact with this variable. The trick is to use a subclass of `threading.local` as per Python's official documentation [here](https://docs.python.org/3/library/threading.html#thread-local-data) ("or a subclass") and their example [here](https://github.com/python/cpython/blob/3.13/Lib/_threading_local.py#L55-L62). Then every thread automatically sees a unique copy of the data structure.

You can also see the pattern in action for example in Haiku, where they have a similar global state stack ([here](https://github.com/google-deepmind/dm-haiku/blob/main/haiku/_src/data_structures.py#L83)). The numpyro equivalent would come down to pretty much this:
```
import threading


class MessengerStack(threading.local):
    """A stack of Messenger instances that is unique to the current thread."""

    def __init__(self):
        super().__init__()
        self.lst = []

    def __len__(self):
        return len(self.lst)

    def __iter__(self):
        return iter(self.lst)

    def __contains__(self, item):
        return item in self.lst

    def __getitem__(self, idx):
        return self.lst[idx]

    def append(self, item):
        self.lst.append(item)

    def index(self, item):
        return self.lst.index(item)

    def pop(self):
        return self.lst.pop()


_PYRO_STACK = MessengerStack()
```

I'd be happy to prepare a pull request for that since we do have a need for that sort of thread safety.

fehiepsi:
Thanks @chkothe! Looking forward to it!

PR #1351, opened 2022-02-28
Make factor an auxiliary variable

fehiepsi:
This is to match [Pyro behavior](https://github.com/pyro-ppl/pyro/blob/dev/pyro/primitives.py#L188).


OlaRonning (approved):
LGTM

fehiepsi:
Thanks, Ola! :)

PR #1358, opened 2022-03-10
Force validate_args to be keyword argument

### LINKED ISSUE #1348: Force validate_args to be keyword arguments
This will help some forum users avoid issues like using `HalfNormal(loc, scale)` for HalfNormal distribution. In such case, validate_args will take the value of `scale` (because HalfNormal does not have loc parameter) and raise some misleading errors.

tcbegley:
Closes #1348


fehiepsi (approved):
Awesome, thanks @tcbegley!

Issue #1309, opened 2022-01-27
Correct SVI API usage


ColdTeapot273K:
So I was playing around with SVI API and problems from Statistical Rethinking 2ed. And one thing - extracting posterior sample data - made me quite puzzled for a while.

The [doc on SVI](https://num.pyro.ai/en/stable/svi.html#numpyro.infer.svi.SVI) describes intended usage as:
```python
# ...
# get posterior samples
predictive = Predictive(guide, params=params, num_samples=1000)
samples = predictive(random.PRNGKey(1), data)
```

The result of that for my model was: 
1. Sample with deterministic estimands missing if I used `Predictive(guide, params=params, num_samples=1000)`
```python
>>> samples.keys()
# actual: dict_keys(['alpha', 'beta', 'sigma'])
# expected: dict_keys(['W', 'alpha', 'beta', 'mu', 'sigma'])
```
2. Empty sample if I used `Predictive(guide, params=params, num_samples=1000)`
```python
>>> samples.keys()
# actual: dict_keys([])
# expected: dict_keys(['W', 'mu'])
```
---
Changing a snippet like this solved the issue:
```python
# ...
# get posterior samples
predictive = Predictive(guide.model, params=params, num_samples=1000)
samples = predictive(random.PRNGKey(1), data)
```
I've found this solution while comparing documentation code to "Code 4.53" by @fehiepsi from [here](https://github.com/fehiepsi/rethinking-numpyro/blob/master/notebooks/04_geocentric_models.ipynb ) 

---

My full model code:
```python
import warnings

warnings.filterwarnings("ignore", category=DeprecationWarning)

import pandas as pd

import jax.numpy as jnp
from jax import random

import numpyro
import numpyro.distributions as dist
import numpyro.optim as optim
from numpyro.infer import SVI, Trace_ELBO, Predictive
from numpyro.infer.autoguide import AutoNormal

# INFO: load data for problems from the official repo
data_uri = "https://raw.githubusercontent.com/rmcelreath/rethinking/master/data/Howell1.csv"
df_dev = pd.read_csv(data_uri, sep=";")
df_dev = df_dev[df_dev["age"] >= 18]

def model(df_dev):
    alpha = numpyro.sample("alpha", dist.Normal(60, 10))
    beta = numpyro.sample("beta", dist.LogNormal(0, 1))
    sigma = numpyro.sample("sigma", dist.Uniform(0, 10))
[...]

PR #1363, opened 2022-03-13
Feature/svi docs accessibility

ColdTeapot273K:
Addresses SVI doc accessibility requests/suggestions from #1309


numpyro/infer/svi.py
```diff
@@ -86,33 +86,127 @@ class SVI(object):
...
-        >>> data = jnp.concatenate([jnp.ones(6), jnp.zeros(4)])
-        >>> optimizer = numpyro.optim.Adam(step_size=0.0005)
-        >>> svi = SVI(model, guide, optimizer, loss=Trace_ELBO())
-        >>> svi_result = svi.run(random.PRNGKey(0), 2000, data)
+        >>> import numpyro.optim as optim
+        >>> from numpyro.infer import SVI, Trace_ELBO, Predictive
+        >>> from numpyro.infer.autoguide import AutoNormal
+
+        >>> # Modeling person's weight from a 10-observation snippet
+        >>> # of 'Howell1' dataset from Statistical Rethinking 2nd ed. (R code 5.45)
```
fehiepsi, line 99: I think it is better to use the current example, which is concise, focuses on SVI inference (not modeling), and illustrates different aspects like initializing parameters. You can move this example to a standalone example/tutorial in https://github.com/pyro-ppl/numpyro/tree/master/examples .

fehiepsi: For the current docstring, I think we are missing prior/posterior predictive?

ColdTeapot273K: I wouldn't call the current doc example accessible to outsiders (compared to sklearn examples or compared to Pyro even: e.g. no `alpha0`, `beta0` variables from [reference](http://pyro.ai/examples/svi_part_i.html)). but my proposal is not very laconic i admit and i'm not the maintainer.

so yeah, if you want i can just secure all the 4 sampling showcase and channel the rest into examples section.

also sorry for being dilatory. i'm not abandoning conrtib efforts, i just have to do a lot of crisis management due to what's happening.

fehiepsi: I see. I guess a beta binomial model is not a good starting example. How about using a simple linear regression with example with scalar parameters instead? Like learning `y = w * x + b` where `x = jnp.arange(10)`, `y = 2 * x + 1 + random.normal(random.PRNGKey(0), shape=x.shape) * 0.1`. I think we need custom guide so it is better to avoid auto guide here. We also need `numpyro.param(..., lambda rng_key: ..., ...)` to initialize parameters.

Moving your example to the examples section sounds great to me. Please take your time. We have several PRs that take years to finish. ;) But we might occasionally check to see if you want to make those enhancements available for the next releases.

ColdTeapot273K: Haha thanks @fehiepsi 

@fehiepsi I've went through "SVI part 1 " Pyro example and decided that it's actually fine, just needs some more clarity in Numpyro adaptation. So I've tried to make it more coherent with Pyro implementation at the end of the tutorial so that it'd be easier to refer and compare differences of Numpyro SVI API to Pyro SVI API. Please check out the last commits

numpyro/infer/svi.py
```diff
@@ -93,26 +89,101 @@ class SVI(object):
...
+        ...     alpha0 = 10.
+        ...     beta0 = 10.
+        ...     # sample f from the Beta prior
+        ...     f = numpyro.sample("latent_fairness", dist.Beta(alpha0, beta0))
         ...     with numpyro.plate("N", data.shape[0]):
         ...         numpyro.sample("obs", dist.Bernoulli(f), obs=data)
 
         >>> def guide(data):
+        ...     # register the two variational parameters with Pyro
+        ...     # - both parameters will have initial value 15.0.
```
fehiepsi, line 106: nit: both parameters -> the first parameter

numpyro/infer/svi.py
```diff
@@ -93,26 +89,101 @@ class SVI(object):
...
+        ...     # - because we invoke constraints.positive, the optimizer
+        ...     # will take gradients on the unconstrained parameters
+        ...     # (which are related to the constrained parameters by a log)
         ...     alpha_q = numpyro.param("alpha_q", 15., constraint=constraints.positive)
+        ...     # variational parameter can be initialized stochastically
         ...     beta_q = numpyro.param("beta_q", lambda rng_key: random.exponential(rng_key),
-        ...                            constraint=constraints.positive)
+        ...                         constraint=constraints.positive)
```
fehiepsi, line 113: nit: missing 3 spaces before `constraint`

numpyro/infer/svi.py
```diff
@@ -93,26 +89,101 @@ class SVI(object):
...
+
+        >>> # here we use some facts about the Beta distribution
+        >>> # compute the inferred mean of the coin's fairness
         >>> inferred_mean = params["alpha_q"] / (params["alpha_q"] + params["beta_q"])
+        >>> # compute inferred standard deviation
+        >>> factor = params["beta_q"] / (params["alpha_q"] * (1.0 + params["alpha_q"] + params["beta_q"]))
+        >>> inferred_std = inferred_mean * math.sqrt(factor)
+
+        >>> # sampling from guide for quick diagnostics
+        >>> samples = guide.sample_posterior(random.PRNGKey(1), params, (1000,))
```
fehiepsi, line 135: I think you need to use Predictive here. `guide` is a callable which might not have `sample_posterior` method.

numpyro/infer/svi.py
```diff
@@ -93,26 +89,101 @@ class SVI(object):
...
+        >>> numpyro.diagnostics.print_summary(samples, prob=0.89, group_by_chain=False)
+
+        >>> # get posterior predictive samples, for arbitrary data size
+        >>> num_samples = 1000
+        >>> dist_posterior_predictive = Predictive(
+        ...     model=model,
+        ...     guide=guide,
+        ...     params=params,
+        ...     num_samples=num_samples,
+        ... )
```
fehiepsi, line 145: nit: it might be better to combine those lines into 1 single line

numpyro/infer/svi.py
```diff
@@ -93,26 +89,101 @@ class SVI(object):
...
+
+        >>> samples_posterior_predictive = dist_posterior_predictive(
+        ...     random.PRNGKey(1),
+        ...     data=data[:5],
+        ... )
```
fehiepsi, line 150: nit: it might be better to use 1 single line here

numpyro/infer/svi.py
```diff
@@ -93,26 +89,101 @@ class SVI(object):
...
+        >>> dist_prior_predictive = Predictive(
+        ...     model=model,
+        ...     params=None,
+        ...     num_samples=num_samples,
+        ... )
+
+        >>> samples_prior_predictive = dist_prior_predictive(
+        ...     random.PRNGKey(1),
+        ...     data=data[:5],
+        ... )
```
fehiepsi, line 163: ditto: using 1 single line and different random keys

numpyro/infer/svi.py
```diff
@@ -93,26 +89,101 @@ class SVI(object):
...
+        >>> dist_prior = Predictive(
+        ...     model=guide,
+        ...     params=None,
+        ...     num_samples=num_samples,
+        ... )
+
+        >>> samples_prior = dist_prior(
+        ...     random.PRNGKey(1),
+        ...     data=data,
+        ... )
```
fehiepsi, line 185: I think it is better to remove either `full data` part or partial data (should we use new data rather than partial data) part. Actually, I'm a bit confused of several notions here. What posterior predictive with arbitrary data size does? I think we need to set `data=None` for posterior predictive. Could we update the model for that pattern and only include code to show:
+ how to get posterior samples from the guide
+ (optional because it is not relevant to SVI and we already explain it in [bayesian regression](https://num.pyro.ai/en/stable/tutorials/bayesian_regression.html#Prior-Predictive-Distribution) tutorial) how to get prior samples (having data or not is not important to get prior samples - if we want to get prior predictive, it is necessary to set data=None)
+ how to use the learned parameters to draw predictive samples (with data=None): there are two ways to do this:
  + using posterior_samples above
  + using guide directly

fehiepsi (commented):
Thanks, @ColdTeapot273K! Just took another pass. :)

numpyro/infer/svi.py
fehiepsi: @ColdTeapot273K would you mind if I push a commit to revise the CL and merge this?

fehiepsi:
Closed in favor of https://github.com/pyro-ppl/numpyro/pull/1549. Thanks @ColdTeapot273K for taking a stab at this!

PR #1372, opened 2022-03-20
Fix/bvm sampling

OlaRonning:
Sampling currently fails if `batch_shape` or `event_shape` is set for `SineBivariateVonMises` (BVM).


fehiepsi (approved):
Do you want to add tests? Shouldn't event shape always be (2,)?

numpyro/distributions/directional.py
```diff
@@ -433,25 +433,30 @@ def sample(self, key, sample_shape=()):
...
-            jnp.reshape(eig, (2, batch_size)),
-            jnp.reshape(b0, (batch_size,)),
-            jnp.reshape(eigmin, (batch_size,)),
-            jnp.reshape(phi_den, (batch_size,)),
+            jnp.broadcast_to(jnp.reshape(conc, (2, -1)), (2, batch_size)),
+            jnp.broadcast_to(jnp.reshape(corr, -1), (batch_size,)),
+            jnp.broadcast_to(jnp.reshape(eig, (2, -1)), (2, batch_size)),
+            jnp.broadcast_to(jnp.reshape(b0, -1), (batch_size,)),
+            jnp.broadcast_to(jnp.reshape(eigmin, -1), (batch_size,)),
+            jnp.broadcast_to(jnp.reshape(phi_den, -1), (batch_size,)),
```
fehiepsi, line 441: We already broadcast (except for corr) in phimarginal so only reshape here is required I guess.

fehiepsi: It seems that the current approach is not valid if each parameter has its own batch shape. I think you need to do broadcasting in the construction, and revert changes in this PR.

fehiepsi (changes requested):
It is better to broadcast parameters in the construction.

OlaRonning:
Thanks for the review, @fehiepsi. I'll move broadcasting up to the constructor and add test cases.

Regarding the `event_shape,` it seems valid for `(S^1, S^1)^n` (S^1 is the circle); I'll double-check with Kent's paper.

numpyro/distributions/directional.py
```diff
@@ -357,22 +358,31 @@ def __init__(
...
+            self.correlation,
+        ) = map(
+            partial(jnp.broadcast_to, shape=batch_shape),
+            promote_shapes(
+                phi_loc,
+                psi_loc,
+                phi_concentration,
+                psi_concentration,
+                correlation,
+                shape=batch_shape,
```
OlaRonning, line 382: The promotion was missing the shape to promote to.

fehiepsi (approved):
Lgtm, thanks Ola!

Issue #1361, opened 2022-03-11
A question about generalizing into domain specific models (e.g. time series) 


hesenp:
I noticed that we already have [a few time series forecasting models](https://num.pyro.ai/en/stable/tutorials/time_series_forecasting.html) as part of the examples. For these models we mostly write the models from ground up.

On the other side of the spectrum, there's all these open source packages like Prophet, Orbit which just rely on a few models internally, but have more of a peripheral functions to make them easy to use for folks who are not modeling experts. Coincidentally, Orbit has [this LGT model](https://uber.github.io/orbit/tutorials/lgt.html) which is highly similar to the SGT model that Numpyro has used as an example, linked above. 

I wonder if there's any thoughts to building some domain specific models/packages based on numpyro? like `numpyro-ts` that uses numpyro as compute engine in the backend, but has utility functions in the front, so that people could do modeling easily? 

an example code would like: 

```
model = SGT() 
model.fit(data_frame)
model.predict(another_data_frame)
```

this could make the code / package very easy to get picked up by general practitioners.

fehiepsi:
There are efforts in the past https://github.com/pyro-ppl/numpyro/pull/544 https://github.com/pyro-ppl/numpyro/pull/551 see e.g. [the lynx example](https://github.com/pyro-ppl/numpyro/blob/36457e8e15cd898ded44a2a8d5fd038d429606a1/numpyro/contrib/forecast/lynx.ipynb). If anyone is interested, please feel free to reuse/modify the code there but I think it is better to start with some popular algorithms e.g. those in [Rob Hyndman's forecasting book](https://otexts.com/fpp3/ses.html).

hesenp:
oh interesting. what's your thought on these code in the [forecast folder](https://github.com/fehiepsi/numpyro/tree/pylgt/numpyro/contrib/forecast) @fehiepsi ? do you have plans to check them in some time?

fehiepsi:
Hi @hesenp, I don't have a plan for it. Currently, for time series problems, I just construct the model from scratch. But I believe having a module for basic forecasting models would be useful for many people.

fehiepsi:
That API sounds reasonable to me. We can have default methods to get init values/ to-be-scanned values for traditional models while still provides extensions for some derived models. Like to include global terms as in [Orbit](https://uber.github.io/orbit/tutorials/dlt.html)

hesenp:
[...]

PR #1377, opened 2022-03-25
[WIP] Time series modules 

hesenp:
Some time series wrapper as discussed in #1361 . Could appreciate any discussions and inputs as the effort is very nascent.

A few things this model tries to accomplish: 
 1. provide a building blocks for time series analysis. i think this should not be limited to forecasting only. it should have interpretation capability as well. 
 2. the building blocks should fit into the `fit`, `predict` pattern like sci-kit learn. 
 3. hide away sampling procedures. 
 4. hide away some of the numpyro handler functions, which could be a bit daunting to new users (including myself) 
 5. still enable flexibility to try out different patterns. 

With the above objectives in mind, this PR (so far) includes an implementation of a naive AR(1) model that is built on the following procedures: 

 * method 1: low level implementation of model, based on `NUTModelTrainer` only. 
 * method 2: higher level implementation of model, based on `StateSpaceModel`. 

Could appreciate any inputs and hole poking on whether these patterns in method 1 and 2 would break in certain situations. I might come back and add more text book examples later.

### DIFF
--- a/numpyro/contrib/time_series.py
+++ b/numpyro/contrib/time_series.py
@@ -1,0 +1,115 @@
+import jax 
+import jax.numpy as jnp
+import numpyro
+import numpyro.distributions as dist 
+from numpyro import sample
+from numpyro.contrib.control_flow import scan
+from numpyro.infer import Predictive, NUTS, MCMC
+
+
+class NUTModelTrainer(object):
+    """ 
+    basically a wrapper class that trains a model 
+    using NUT and MCMC. We can come back and add different 
+    types of trainers afterwards. 
+    """
+    def __init__(self, rng_seed = 1000):
+        self.rng_key = jax.random.PRNGKey(seed=rng_seed)
+        self.trained = False
+
+    def model(self, data=None):
+        raise NotImplementedError
+    
+    def fit(self, data, num_steps=1000, num_warmup_steps=1000):
+        self.sampler = NUTS(self.model)
+        self.mcmc = MCMC(self.sampler, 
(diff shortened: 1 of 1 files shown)


numpyro/contrib/time_series.py
```diff
@@ -0,0 +1,128 @@
...
+        return self.mcmc
+
+    def predict(self, data, num_samples=100): 
+        posterior = Predictive(self.model, num_samples=num_samples)
+        posterior_sample = posterior(self.rng_key)
+        return posterior_sample
+
+
+
+class StateSpaceModel(NUTModelTrainer):
```
hesenp, line 40: btw, is it correct to call this pattern state space model?

fehiepsi (commented):
Thanks, @hesenp! For low-level implementation, we should aim for flexibility while for high-level API, I prefer easy-to-use. The current designs look reasonable to me. I just wish that we can support both SVI and NUTS forecasters (personally, I would like to use SVI to quickly fit the model first). Maybe having a method `fit_svi` or `fit(..., guide=None)` to run svi?

fehiepsi:
Maybe it is fine to not support `SVI` in time-series module. We can have a separate `forecast` module that follows Pyro semantic: Forecaster and HMCForecaster.

hesenp:
ah sorry i'm very slow these days. will get back to it ha.

fehiepsi:
Closed because it is better (for api design and maintanance) to have a separate library for this. See also https://github.com/pyro-ppl/numpyro/issues/1361

PR #1352, opened 2022-03-03
Notebook/multilevel hierarchy

cstoafer:
These updates expand upon an existing notebook to illustrate how to add an intermediate level to hierarchical models.  

Closes #1336

PR #1380, opened 2022-03-29
update notebook headers to fix documentation menu

cstoafer:
I noticed a problem with my recent PR #1352 because some sections ("5. Add layer to model hierarchy: Smoking Status" and "Wrap-up") from the notebook appear in the [NumPyro documentation menu](https://num.pyro.ai/en/latest/index.html):

image


fehiepsi (approved):
Thanks, @cstoafer!

PR #1384, opened 2022-04-08
Enhance LocScaleReparam noncenter dependency

fehiepsi:
This enhances LocScaleReparam dependency for non-center reparameterization.

Current tests covered the case center=0. For rendering, the model at #1295 rendered as expected

image

Blocked by #1388.


numpyro/infer/reparam.py
```diff
@@ -102,8 +102,12 @@ def __call__(self, name, fn, obs):
                 jnp.full(event_shape, 0.5),
                 constraint=constraints.unit_interval,
             )
-        params["loc"] = fn.loc * centered
-        params["scale"] = fn.scale**centered
+        if self.centered in [0, 0.0]:
```
fritzo, line 105: Hmm in Pyro `self.centered` can be an array
```python
    def __init__(self, centered=None, shape_params=None):
        assert centered is None or isinstance(centered, (float, torch.Tensor))
```
in which case you would also need an `isinstance` test to avoid
```python
>>> numpy.zeros(3) in [0, 0.0]
  File "<stdin>", line 1, in <module>
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```
It's nice to allow arrays because we can differently center per element. Is that out of scope for this PR?

fehiepsi: The handler supports arrays but I didn't expect that we need `isinstance` here. Let me fix this, thanks!

fritzo (approved):

Issue #936, opened 2021-03-02
Excessive memory usage at the beginning of NUTS inference


d-caviedes:
Hello everyone,

asked by @fehiepsi I create the following issue regarding what seems "excessive" memory consumption at the beginning of NUTS inference (originally addressed in the discussion forum https://forum.pyro.ai/t/numpyro-memory-usage-at-the-beginning-of-nuts-inference/).

I am running inferences on a linear regression model such as

```
def model_sbl(data):
    obs = data["obs"]
    Nw = data["Nw"]
    Nt = data["Nt"]
    X = data["X"]
    b = data["b"]
    tau0 = data["tau0"]
    sigma = numpyro.sample("sigma", dist.LogNormal(0.0, 1.0))
    dist_var = dist.InverseGamma(1, 10 ** (-b) * jnp.ones(Nw * Nt))
    with numpyro.plate("coefficients", Nw * Nt):
        w_var = jnp.sqrt(numpyro.sample("w_var", dist_var))
        dist_w = dist.Normal(jnp.zeros((Nw * Nt)), tau0 * w_var)
        w = numpyro.sample("w", dist_w)
    mean = jnp.matmul(X, w)
    y = dist.Normal(mean, sigma)
    with numpyro.plate("observations", size=len(obs)):
        numpyro.sample("obs", y, obs=obs)
```

The parameter space is quite large, with _X_ being  a dictionary with dimensions (5236, 123200) in this case, and _w_ 123200 unknown coefficients. To run the inferences

```
data = dict(
      obs=jnp.array(h_fit.T.flatten()),
      X=jnp.array(D_fit),
      b=0,
      tau0=1e-1,
      sigma=0.1,
      w_var=0.1,
      Nw=Nw,
      Nt=Nt,
  )
rng_key = jrandom.PRNGKey(0)
rng_key, rng_key_ = jrandom.split(rng_key)
my_kernel = NUTS(model_sbl, max_tree_depth=7)
posterior = MCMC(my_kernel, num_samples=100, num_warmup=50, num_chains=1)
pdb.set_trace()
# run_posterior(posterior, rng_key_, data)
posterior.run(rng_key_, data=data)
```

The memory consumption tracked with memory_profiler is detailed in the following Figure. I annotated what happened while debugging. The first and second samples when `numpyro.sample("obs", y, obs=obs)` is called consume quite a lot of memory. But when these two first samples are computed, it stabilizes to a constant RAM, and keeps like that during the entire inference

[...]

PR #1386, opened 2022-04-08
Update for JAX 0.3.5 

fehiepsi:
After JAX 0.3.5, all MCMC codes are broken because we use `x.copy().tobytes()` to hash model's argument. Not only it is broken now, it was also my mistake to use that expensive operator for caching (this might be the reason for #936). I just tested that just simply using `id`, `test/infer/test_mcmc.py::test_compile_warmup_run` is still working as expected.

Also changing `tree_multimap` to `tree_map` because the former is just a duplicated version of the latter, and `tree_multimap` is deprecated now.

### DIFF
--- a/numpyro/infer/mcmc.py
+++ b/numpyro/infer/mcmc.py
@@ -13,7 +13,7 @@ from jax import jit, lax, local_device_count, pmap, random, vmap
 from jax.core import Tracer
 from jax.interpreters.xla import DeviceArray
 import jax.numpy as jnp
-from jax.tree_util import tree_flatten, tree_map, tree_multimap
+from jax.tree_util import tree_flatten, tree_map
 
 from numpyro.diagnostics import print_summary
 from numpyro.util import cached_by, find_stack_level, fori_collect, identity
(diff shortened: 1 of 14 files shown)


numpyro/infer/mcmc.py
```diff
@@ -187,13 +185,9 @@ def collect(x):
...
-    # When the arguments are JITed, ShapedArray is hashable.
-    if isinstance(x, Tracer):
-        return x
-    elif isinstance(x, DeviceArray):
-        return x.copy().tobytes()
-    elif isinstance(x, (np.ndarray, jnp.ndarray)):
-        return x.tobytes()
+    # NOTE: When the arguments are JITed, ShapedArray is hashable.
+    if isinstance(x, (np.ndarray, jnp.ndarray)):
+        return id(x)
```
fehiepsi, line 190: This is the only important change in this PR.

OlaRonning (approved):
LGTM @fehiepsi.

fehiepsi:
Thanks, Ola!

PR #1388, opened 2022-04-10
Update jaxns to 1.0.0

fehiepsi:
This PR updates jaxns to 1.0.0 and pins its version (which hopefully can reduce the maintaining effort).

**API change:** we defer specifying the constructor and termination configurations to users. There are many upstream configurations that keep changing so this will help reduce breakages. In addition, it is better for users to look at the docstring of jaxns to get the meaning of those configurations.

**Other changes:**
+ import locations
+ jaxns results signatures
+ remove some warnings logic given jaxns is better compatible with jax now
+ internally, jaxns requires loglikelihood to have specific signatures now (rather than **kwargs), so we use the following script to set up the signature of loglik_fn.
```
        local_dict = {}
        loglik_fn_def = """def loglik_fn({}):\n
        \tparams = dict({})\n
        \treturn log_density_(reparam_model, args, kwargs, params)[0]
        """.format(", ".join([f"{name}_base" for name in param_names]),
                   ", ".join([f"{name}_base={name}_base" for name in param_names]))
        exec(loglik_fn_def, locals(), local_dict)
        loglik_fn = local_dict["loglik_fn"]
```
I'm not sure if there is a cleaner solution for it.

**Tested:** current tests and `examples/gaussian_shells.py` passed.

cc @joshuaalbert


numpyro/contrib/nested_sampling.py
```diff
@@ -245,8 +233,17 @@ def run(self, rng_key, *args, **kwargs):
...
+        # Jaxns requires loglikelihood function to have explicit signatures.
+        local_dict = {}
+        loglik_fn_def = """def loglik_fn({}):\n
+        \tparams = dict({})\n
+        \treturn log_density_(reparam_model, args, kwargs, params)[0]
+        """.format(
+            ", ".join([f"{name}_base" for name in param_names]),
+            ", ".join([f"{name}_base={name}_base" for name in param_names]),
+        )
+        exec(loglik_fn_def, locals(), local_dict)
```
fritzo, line 245: oof, I guess it's not worth introducing a dependency on [makefun](https://pypi.org/project/makefun/), which is also used by funsor

fritzo (approved):

numpyro/contrib/nested_sampling.py
fehiepsi: Yeah using `makefun` would make the code much simpler. Thanks for pointing out!

Joshuaalbert:
Regarding the definition of log_likelihood. JAXNS now lets the user specify any signature, using args, and kwargs. It will take from the prior_chain the correct variables by name. See https://github.com/Joshuaalbert/jaxns/blob/master/jaxns/internals/maps.py#L48

fehiepsi:
In previous jaxns versions, we used **kwargs as signature of log likelihood didn't get the ignore warning in line 68 of that file. Now to bypass that warning, the signature contains keywords with names from prior chain.

Issue #1390, opened 2022-04-12
CAR sampler


theorashid:
This is a proposed **enhancement**.

Conditional autoregressive (CAR) models are very useful in spatial settings. Previously, our group has specified a model with a CAR as

```
tau = numpyro.sample('tau', dist.Gamma(14, 4))
alpha = numpyro.sample('alpha', dist.Uniform(low=0.0, high=1.0))

D = jnp.diag(d)
Q = tau*(D - alpha * A)
phi = numpyro.sample('phi', dist.MultivariateNormal(loc=jnp.zeros(N), precision_matrix=Q))
```

`d` is a vector of the number of neighbours of each spatial unit. `A`, is our adjacency matrix: specifying which spatial units share a boundary.

There are implementations in [pymc](https://github.com/pymc-devs/pymc/blob/fa015e3c1f7e087ea5577ba777ca8a631ed7eccd/pymc/distributions/multivariate.py#L2105) and [stan](https://mc-stan.org/users/documentation/case-studies/mbjoseph-CARStan.html), which make use of the sparseness of A, where possible, to improve efficiency in calculating logp. There are also variations like [ICAR](https://mc-stan.org/users/documentation/case-studies/icar_stan.html), which are popular in the spatial statistics field.

Is this something worth implementing or are the efficiency gains going to be very small relative to our current implementation? Is it easy to implement (just port from pymc)? Happy to have a go but I'd need a fair amount of direction and help

fehiepsi:
Hi @theorashid, there is a [jax.experimental.sparse](https://jax.readthedocs.io/en/latest/jax.experimental.sparse.html) module but I'm not sure if it helps here. I will need to dive into it a bit...

theorashid:
There's also [jax.scipy.sparse](https://jax.readthedocs.io/en/latest/jax.scipy.html#module-jax.scipy.sparse.linalg). Looking at the pymc code, we'd need equivalents of `aesara.sparse.dot` and `at.slinalg.eigvalsh` and some way of detecting whether A is sparse. Quite impressive that the aesara team built up sparse functions. I'm surprised jax don't have this functionality

fehiepsi:
Looking at the Stan post, I think you don't need to use JAX here. Just simply computing those quantities using numpy and when compiled, they will become constants, hence become pretty fast to compute.

theorashid:
Okay looking at the pymc implementation, which is equivalent to the stan one, comments inline:

```python
[...]

PR #1394, opened 2022-04-17
Add CAR distribution

theorashid:
Hi @fehiepsi,

This is for #1390. There's a lot going on in the numpyro code base, and I'm fairly new to writing good code, so if it's okay I'll open this draft PR and we can discuss together as I implement the distribution. I don't think it will be a lot of work, it's just tough to know where to start.

Perhaps we can start with `__init__`, `log_prob` and `sample`:

1. `__init__`. I followed the `MultivariateNormal`, using `loc` for the mean shape and `W` for the covariance matrix shape, as `W` (weighted adjacency matrix) has the same shape. One thing to note is that W should be symmetric with diagonal entries of 0. Is there some lower/upper diagonal format we can use to improve efficiency? Also, is it worth calculating other quantities, like `D = jnp.diag(W.sum(axis=-1))` (or just `W.sum(axis=-1)` will do), or the MVN's precision matrix `Q = tau*(D - alpha * W)`? Also, does `"alpha": constraints.unit_interval` include 0 and 1? I think we would need `alpha<1`, but it's something we'd have to test as the `alpha=1` case is very useful but known as ICAR and requires some further constraints. It might have to be added separately but we discuss this down the line.
2. `log_prob`. I've only implemented non-sparse for now. There are so many options for calculating `logquad`: `jnp.sum(phi * (D * phi - self.alpha * self.W @ phi))`, `jnp.sum(phi * D * phi) - self.alpha phi[np.newaxis, :] @ self.W @ phi))` and all sorts of other combinations with `*` or `@`. From your experience making these distributions, which is quickest and most robust?
3. `sample`. We should be able to draw samples from a MVN with precision matrix as described in 1. Is this easily done?

Once I've got this done, we can make our way down the list below. Feel free to add anything.

Cheers,

Theo

- [x] Add all the `@property`, `@lazy_property` and `infer_shapes`
- [x] Implement sparse
- [x] Make tests
- [x] Write distribution docs
- [ ] Implement ICAR (where `alpha=1` but have to deal with degeneracy) if there is scope for it. See bottom of [here](https://mc-stan.org/users/documentation/case-studies/mbjoseph-CARStan.html)


numpyro/distributions/continuous.py
```diff
@@ -1188,6 +1188,70 @@ def infer_shapes(
...
+        self.alpha = alpha
+        self.tau = tau
+        super(CAR, self).__init__(
+            batch_shape=batch_shape,
+            event_shape=event_shape,
+            validate_args=validate_args,
+        )
+
+    def sample(self, key, sample_shape=()):
+        # MVN with covariance matrix inv(tau * (D - alpha * W))
```
fehiepsi, line 1232: I think you can use:
```
D = jnp.diagonal(self.W.sum(axis=-1), axis1=-2, axis2=-1)
tau = jnp.expand_dims(self.tau, (-2, -1))
alpha = jnp.expand_dims(self.alpha, (-2, -1))
P = tau * (D - alpha * self.W))
mvn = dist.MultivariateNormal(self.loc, precision_matrix=P)
return mvn.sample(key, sample_shape=sample_shape)
```

docs/source/distributions.rst
```diff
@@ -249,6 +249,14 @@ MultivariateNormal
     :show-inheritance:
     :member-order: bysource
 
+CAR
+^^^^^^^^^^^^^^^^^^
```
fehiepsi, line 253: nit: redundant `^` symbols

numpyro/distributions/__init__.py
```diff
@@ -151,6 +152,7 @@
     "MultinomialLogits",
     "MultinomialProbs",
     "MultivariateNormal",
+    "CAR",
```
fehiepsi, line 155: nit: Could you put this in an alphabetical order?

numpyro/distributions/continuous.py
```diff
@@ -1188,6 +1188,70 @@ def infer_shapes(
         return batch_shape, event_shape
 
 
+class CAR(Distribution):
+    arg_constraints = {
+        "loc": constraints.real_vector,
+        "alpha": constraints.unit_interval,
+        "tau": constraints.positive,
+        "W": constraints.positive_definite,
```
fehiepsi, line 1196: I think you can skip the `W` constraint for now (so we can allow the sparse case and this distribution is fully reparametrized). We can revisit this in the future if it is worth to introduce a new Constraint object for `W`.

numpyro/distributions/continuous.py
```diff
@@ -1188,6 +1188,70 @@ def infer_shapes(
...
+        "alpha",
+        "tau",
+    ]
+
+    def __init__(
+        self,
+        loc=0.0,
+        alpha=None,
+        tau=None,
+        W=None,
```
fehiepsi, line 1210: It is better to just use `loc, alpha, tau, w, *, validate_args=None` here.

numpyro/distributions/continuous.py
```diff
@@ -1188,6 +1188,70 @@ def infer_shapes(
...
+        validate_args=None,
+    ):
+        if jnp.ndim(loc) == 0:
+            (loc,) = promote_shapes(loc, shape=(1,))
+        # temporary append a new axis to loc
+        loc = loc[..., jnp.newaxis]
+        if W is not None:
+            loc, self.W = promote_shapes(loc, W)
+
+        batch_shape = lax.broadcast_shapes(jnp.shape(loc)[:-2], jnp.shape(self.W)[:-2])
```
fehiepsi, line 1220: I think you can use
```
# no need: loc = loc[..., jnp.newaxis]
batch_shape = lax.broadcast_shapes(jnp.shape(loc)[:-1], jnp.shape(alpha), jnp.shape(tau), jnp.shape(self.W)[:-2])
event_shape = jnp.shape(self.W)[-1:]
self.loc = promote_shapes(loc, shape=batch_shape + event_shape)
self.alpha, self.tau = promote_shapes(alpha, tau, shape=batch_shape)
```

numpyro/distributions/continuous.py
```diff
@@ -1188,6 +1188,70 @@ def infer_shapes(
         return batch_shape, event_shape
 
 
+class CAR(Distribution):
+    arg_constraints = {
+        "loc": constraints.real_vector,
+        "alpha": constraints.unit_interval,
```
fehiepsi, line 1194: you can use `constraints.open_interval(0, 1)` for this

numpyro/distributions/continuous.py
```diff
@@ -1188,6 +1188,70 @@ def infer_shapes(
...
+        D = self.W.sum(axis=-1)
+
+        D_rsqrt = D ** (-0.5)
+        W_scaled = self.W * (D_rsqrt * D_rsqrt[:, jnp.newaxis])
+
+        lam = jnp.linalg.eigvalsh(W_scaled)
+
+        n = D.shape[-1]
+
+        logtau = n * jnp.log(self.tau).sum()
```
fehiepsi, line 1246: Why we need `.sum()` here?

numpyro/distributions/continuous.py
```diff
@@ -1188,6 +1188,70 @@ def infer_shapes(
...
+        logdet = jnp.log(1 - self.alpha * lam).sum()
```
fehiepsi, line 1247: I think you need `jnp.log1p(-jnp.expand_dims(self.alpha, -1) * lam).sum(-1)` here.

numpyro/distributions/continuous.py
```diff
@@ -1188,6 +1188,70 @@ def infer_shapes(
...
+        phi = value - self.loc
+
+        logquad = self.tau * jnp.sum(phi * (D * phi - self.alpha * self.W @ phi))
```
fehiepsi, line 1250: I think this is already robust. Just a note, for broadcasting, you will need
```
... jnp.sum(phi * (D * phi - jnp.expand_dims(self.alpha, -1) * (self.W @ phi)), -1)
```

test/test_distributions.py
```diff
@@ -366,6 +366,20 @@ def get_sp_dist(jax_dist):
...
+        2,
+        np.array(
+            [
+                [0.0, 1.0, 1.0, 0.0],
+                [1.0, 0.0, 0.0, 1.0],
+                [1.0, 0.0, 0.0, 1.0],
+                [0.0, 1.0, 1.0, 0.0],
+            ]
+        ),
+        None,
```
fehiepsi, line 381: nit: no need for the last `None`.

It is better to add various test cases with different batch shapes for `loc, alpha, tau`. It seems one parameter value is missing here? (I guess we need 4 parameters).

numpyro/distributions/continuous.py
theorashid: I'm confused. Are the users going to input a sparse W, or are we going to convert W to a sparse matrix within the log_p block? I thought the latter

fehiepsi: I thought we allow users to specify sparse W to save memory and time complexity.

Are you going to add a flag or a check for sparsity to convert a dense W to a sparse W? If we follow that path, I'm not sure what's a proper way to do such conversion. Seems like we need some check that W is a static numpy array, then use some scipy dense_to_sparse operator on the top of that (those operators should not operate on jax arrays because we need a static size of sparse form of W). Though if I'm a user of the sparse version, I'm not sure why I need to provide the dense version. (I think both stan and pymc allow users to provide the sparse version)

numpyro/distributions/continuous.py
theorashid: I've read a few more implementations. `alpha` can technically be between -1 and 1, but typically people put a prior on the value lying between 0 (no correlation, iid) and 1 (perfect correlation, although not mathematically possible). But I suppose there might be some cases where people want anticorrelation, so I'll use `constraints.open_interval(-1, 1)` pymc actually allow a vector alpha but I cannot think of a real use case so we'll stick to a single float.

numpyro/distributions/continuous.py
theorashid: true, this only makes sense if tau is a vector. and I think tau and alpha should both be floats

numpyro/distributions/continuous.py
theorashid: Hmm. I was thinking to put something like 

`self.sparse = True if (1.0 - np.count_nonzero(np.array(self.W))/self.W.size) > 0.5`. Or maybe there's another numpy/scipy function which can determine this. Basically something that determines if W is sparse. But if we want to allow the user to decide whether to use sparse computations regardless of whether W is actually sparse, we could have an extra argument sparse, and a flag if they input W as anything other than a sparse matrix when sparse=True.

Would this work with our implementation of sample though?

numpyro/distributions/continuous.py
theorashid: the shape stuff is very confusing but I'll rely on your experience. Let me know if I implemented your suggestion correctly when I push these commits

test/test_distributions.py
theorashid: I'll write some tests once we've finalised the CAR distribution class

numpyro/distributions/continuous.py
theorashid: `axis1=-2, axis2=-1` is very confusing. But I back your shape experience!

theorashid: We'll have to revisit if this works for sparse W

fehiepsi: We need to expand tau and alpha in case they have batch dimensions. E.g. you might want to have CAR priors for a batch of real vectors.

theorashid: `jnp.diagonal(W.sum(axis=-1), axis1=-2, axis2=-1)` with my test `W = jnp.array(
            [
                [0.0, 1.0, 1.0, 0.0],
                [1.0, 0.0, 0.0, 1.0],
                [1.0, 0.0, 0.0, 1.0],
                [0.0, 1.0, 1.0, 0.0],
            ]
        ) `

throws an error of `ValueError: axis 1 is out of bounds for array of dimension 1`. Any idea?

numpyro/distributions/continuous.py
fehiepsi: I see. Please go ahead if you want to use the dense matrix as the input and use some heuristic rule for the conversion. It is less flexible for users but it should work, just make sure to check that W is a numpy array before applying those numpy operators. In the future, we can add CARsparse or something like that if users want to use a sparse version. I dont know how to derive a sparse formula `sample` method - we might just use the dense version there.

I think stan also requires users to provide sparse W for the car prior. It does some sorts of data transformation from dense to sparse in advance, which is a reasonable approach for a tutorial (in practice, assume that a sparse W is stored in some csv file, that csv should only store the sparse version for memory and space efficient).

fehiepsi: > something that determines if W is sparse

I would like to control this by myself, rather than relying on some heuristic rule. It is hard to decide which threshold to use (given different performance on the shapes, the backends cpu tpu gpu, the memory,...) Even if I have a dense array from somewhere, I would like to call dense_to_sparse outside of mcmc code. The signature is the same with the cost of one additional line of code for that conversion (it would cost nothing if I already have a sparse array). As above, I can just simply create a new CARsparse version if I want such behavior. So please just go ahead with the heuristic approach.

theorashid: We can create a `CARsparse` separately if you think it's easier for the user? If we can work out how to sample from MVN with a sparse precision matrix, it won't be much more work. Up to you

fehiepsi: Actually, we can have a compromised version for all usage cases:
def init(W, is_sparse=False):
if is_sparse:
    if not scipy...issparse(W):
        assert isinstance(W, np.ndarray)
        W = scipy.sparse....from_dense(W)

theorashid: Got you. So let the user decide whether to use sparse computations, but also let the user decide whether to input sparse W or not. Perfect.

In `sample`, shall we just stick to dense computations?

numpyro/distributions/continuous.py
fehiepsi: Sorry, I wanted to convert a diagonal vector to a diagonal matrix there (the revert of the operator I used). I think we can use
```
D = self.W.sum(axis=-1, keepdims=True) * jnp.eye(self.W.shape[-1])
```

numpyro/distributions/continuous.py
```diff
@@ -1188,6 +1188,71 @@ def infer_shapes(
...
+        "alpha",
+        "tau",
+    ]
+
+    def __init__(self, loc, alpha, tau, W, *, validate_args=None):
+        if jnp.ndim(loc) == 0:
+            (loc,) = promote_shapes(loc, shape=(1,))
+
+        if W is not None:
+            loc, self.W = promote_shapes(loc, W)
```
fehiepsi, line 1210: This can be removed

theorashid: take out just lines 1209 and 1210 or 1206-10?

fehiepsi: just "lines 1209 and 1210"

numpyro/distributions/continuous.py
fehiepsi: > shall we just stick to dense computations

I think so. There should be a way to utilize the sparse form but I can't derive it now. Unless you want to perform variational inference on the top of this prior, otherwise, `sample` is not required for MCMC.

theorashid: There are a [bunch](https://docs.scipy.org/doc/scipy-1.8.0/html-scipyorg/reference/sparse.html) of sparse matrix formats. What's the best way of coping if a user inputs the wrong type?

theorashid: Also, what actually is scale_tril? I'm assuming it's not just the lower diagonal of a covariance matrix

fehiepsi: I think you can use `W = W.tocsr()` to make sure it has the desired sparse format to perform multiplication for `W_scaled`. `scale_tril` is the cholesky of covariance matrix.

theorashid: There's a [sparse cholesky from sci-kit](https://scikit-sparse.readthedocs.io/en/latest/cholmod.html), but is it usable, and is it worth using for `scale_tril`?

theorashid:
I have added some sparse methods. Perhaps they could be tidier. I'm not sure what happens if the adjacency matrix is passed as a jax array but we can specify that we want W to be a np.array in the docs

Re the check, for some reason when I run `black .`, it tries to reformat 10s of files changing ` ** ` to `**`. How do I fix this?

numpyro/distributions/continuous.py
fehiepsi: I'm not sure. I guess we are trying to compute cholesky of inverse of sparse matrix here.

theorashid: Scipy has a sparse inverse. At this stage it isn't worth it. When the spatial modellers start using the function in practice, we'll see if this is a pinch point and can evaluate whether it is worth implementing

numpyro/distributions/continuous.py
```diff
@@ -1188,6 +1191,99 @@ def infer_shapes(
...
+            if not sparse.issparse(self.W):
+                assert isinstance(self.W, np.ndarray)
+            W = sparse.csr_matrix(self.W)
+
+            D = np.asarray(W.sum(axis=-1)).squeeze()
+            D_rsqrt = np.diag(D ** (-0.5))
+
+            W_scaled = D_rsqrt.dot(W.dot(D_rsqrt))
+
+            Wphi = W.dot(phi)
```
fehiepsi, line 1264: It seems that using the later branch is much faster
```
from scipy import sparse
import numpy as np

size = 1000
a = np.zeros((size, size))
n = size // 10
a[np.random.randint(0, size, n), np.random.randint(0, size, n)] = np.random.rand(n) + 1
print(a.shape)
W = sparse.csr_matrix(a)
D = np.asarray(W.sum(axis=-1)).squeeze(-1)

%%timeit
D_rsqrt = np.diag(D ** (-0.5))
W_scaled = D_rsqrt.dot(W.dot(D_rsqrt))

%%timeit
D_rsqrt = D ** (-0.5)
W_scaled = W * (D_rsqrt * D_rsqrt[:, np.newaxis])
```
Here `W.dot(phi)` might not work (and not propagate gradients,...) if `phi` is a jax array. I think what we want is converting W to a jax sparse array then call `W @ phi` in both branches. In addition, it is beneficial to perform `np.linalg.eigvalsh(W_scaled)` in the sparse branch (because it is an expensive operator, and we would like to precompute the constant `lam`, as explained in the Stan article).

numpyro/distributions/continuous.py
fehiepsi: agreed!

fehiepsi:
> run black .,

Maybe checking the black version and try to update it? I think `black` changes its semantics in a recent version.

fehiepsi:
> what happens if the adjacency matrix is passed as a jax array

As long as it is not a scipy sparse matrix, I think the non-sparse branches work for it (because numpy/scipy operations only apply in the sparse branch).

numpyro/distributions/continuous.py
theorashid: > D_rsqrt = np.diag(D ** (-0.5))

This has some issues in the code above in the way `a` is defined because some entries of `D` are zero. We should specify in the docs that the graph definining the adjacency matrix `W` should be a single connected component to avoid this happening. The [code to deal with a disconnected component is quite specialised](https://connordonegan.github.io/geostan/reference/stan_icar.html). If users want it, we can implement it, but in the past I have just created links between the mainland and islands to avoid this happening.

theorashid: It's nice that the use of `multiply` keeps everything sparse – `dot` made it dense.

I was playing around with some sparse solvers for the line `lam`: `scipy.sparse.linalg.eigsh(W_scaled, k=W_scaled.shape[-1]-1, return_eigenvectors=False)`. But the output and the use of k aren't ideal. I couldn't get some version using [`jax.scipy.sparse.linalg.cg`](https://jax.readthedocs.io/en/latest/_autosummary/jax.scipy.sparse.linalg.cg.html) to do anything. So we'll stick to non-sparse solvers unless you think otherwise.

Do you have a preference between `np.linalg.eigvalsh(W_scaled)` and `scipy.linalg.eigvalsh(W_scaled, np.eye(W_scaled.shape[-1])`?

fehiepsi: Good point about the condition of D. I think you can add some checks on the sparse branch (like positive D and lam). And remove the positive definite constraint for W.

Re generate matrix: you can set nonzero entry on the first row together with some random entries, perform a + a.T, and fill diagonal by 0 as you did.

Re sparse solver: yeah, I'm not sure if there is a sparse version of it that gives all eigvals. Let's use with the dense version for now.

Difference of lam: in sparse branch, it will be a constant in every mcmc step while in nonsparse branch, it will be recomputed by jax in every mcmc step.

Re preference: It is up to you. I tend to use np operators more often than scipy operators

theorashid: Do we want to calculate `lam` at every mcmc step? I don't think we do. `W_scaled` is a constant. i don't think we want to repeat that calculation in the dense branch

theorashid:
I got black working (`pip install --upgrade pip` lol). I've added some properties and an infer_shapes method

numpyro/distributions/continuous.py
fehiepsi: Hmm, to avoid repeating that computation (in case you think it is expensive) in the non-sparse branch you can add a check that if W_scaled is a numpy array, then use numpy operator. You can benchmark with
```
import jax

@jax.jit
def f():
  return jax.numpy.linalg.eigvalsh(W_scaled)

@jax.jit
def g():
  return np.linalg.eigvalsh(W_scaled)
```

fehiepsi: To clarify: I suggested to use numpy operators on the top of numpy arrays to improve the performance. The current code repeats the computations. The benchmark code is used to confirm that it is actually repeating the computation if we use jax operators on the top of numpy array.

fehiepsi: If you run f, you will see that it is slow (because it is repeating the computations in later f() executions). The code is not intended to compare the speed of the jax operator and the numpy operator. In g, even if you add "very slow" numpy codes, the later execution of g is still very fast.

theorashid: For my understanding, _why_ does `jax.numpy.linalg.eigvalsh(W_scaled)` repeat the calculation at every step of the mcmc when it only needs to be calculated once? And why does `np.linalg.eigvalsh(W_scaled)` not?

fehiepsi: It is just how jax works. The compiled code performs all jax operators so it should perform jnp...eigvalsh again. The compiled code does not perform any non-jax operators so the value obtained from np.eigvalsh will be constant. NumPyro mcmc uses jax jit under the hood. You can try:
```
@jax.jit
def f():
    print("repeat")
    return 0.
```
to see that jax skips the print statement in the later executions. Same for numpy operators.

fehiepsi: Actually, I don't know how to make jax avoid repeating jnp.... operators, if that's what you are looking for. I thought it is not possible and the benchmark code seems to confirm so. If you think we can avoid repeating jnp... operators, I would love to try.

fehiepsi: Understood! It would be nice to treat outputs of jax ops on constant arrays as constant - it's not possible now at least for eigval but let's hope that xla (jax complier) can optimize the code better in the future.

numpyro/distributions/continuous.py
```diff
@@ -1188,6 +1192,150 @@ def infer_shapes(
...
+            W = self.W
+
+        D = W.sum(axis=-1, keepdims=True) * jnp.eye(W.shape[-1])
+        tau = jnp.expand_dims(self.tau, (-2, -1))
+        alpha = jnp.expand_dims(self.alpha, (-2, -1))
+        return tau * (D - alpha * W)
+
+    @staticmethod
+    def infer_shapes(loc, alpha, tau, W):
+        event_shape = jnp.shape(W)[-1:]
```
fehiepsi, line 1321: This already assumes the inputs are shapes, so you can do `event_shape = W[-1:]`

numpyro/distributions/continuous.py
```diff
@@ -1188,6 +1192,150 @@ def infer_shapes(
...
+
+        if self.is_sparse:
+            if not sparse.issparse(self.W):
+                assert isinstance(self.W, np.ndarray)
+            W = sparse.csr_matrix(self.W)
+
+            D = np.asarray(W.sum(axis=-1)).squeeze()
+            D_rsqrt = D ** (-0.5)
+
+            W_scaled = W.multiply(D_rsqrt @ D_rsqrt[:, np.newaxis]).toarray()
```
fehiepsi, line 1278: I think performing .multipy(...).multiply(...) is faster (than creating a dense matrix and perform sparse * dense)

numpyro/distributions/continuous.py
```diff
@@ -1188,6 +1192,150 @@ def infer_shapes(
...
+
+            lam = np.linalg.eigvalsh(W_scaled)
```
fehiepsi, line 1280: I think we need `todense` here for W_scaled and add a todo note to find eigval of sparse matrix)

numpyro/distributions/continuous.py
```diff
@@ -1188,6 +1192,150 @@ def infer_shapes(
...
+    :param numpy.ndarray or scipy.sparse.csr_matrix W: symmetric adjacency matrix where 1
+        indicates adjacency between sites and 0 otherwise
+    :param bool whether to use a sparse form of W in calculations (must be True is W is a
+        scipy.sparse.csr_matrix)
+    """
+    arg_constraints = {
+        "loc": constraints.real_vector,
+        "alpha": constraints.open_interval(-1, 1),
+        "tau": constraints.positive,
+        "W": constraints.positive_definite,
```
fehiepsi: Because W needs not to be positive definite, could you replace this with
```
constraints.dependent_property(is_discrete=False, event_dim=2)
```

numpyro/distributions/continuous.py
```diff
@@ -1188,6 +1192,150 @@ def infer_shapes(
...
+    }
+    support = constraints.real_vector
+    reparametrized_params = [
+        "loc",
+        "alpha",
+        "tau",
```
fehiepsi, line 1223: I think we can add W here too.

theorashid:
`jnp.shape` _does_ work for sparse matrices.

> Good point about the condition of D. I think you can add some checks on the sparse branch (like positive D and lam). And remove the positive definite constraint for W.

I have removed the constraint for W and I have added the sparse/W branches to `__init__`. Should I add the checks for a correct W (no islands – all sites have neighbours `W.sum(axis=-1).min() > 0`; symmetric `W = W.T`) in __init__ or somewhere else?

More basic shape issues:
```{python}
dist.CAR(loc=0,alpha=0.1,tau=0.1,
    W=np.array(
        [
            [0.0, 1.0, 1.0, 0.0],
            [1.0, 0.0, 0.0, 1.0],
            [1.0, 0.0, 0.0, 1.0],
            [0.0, 1.0, 1.0, 0.0],
        ]
    )
).sample(jax.random.PRNGKey(0))
```
gives the error `TypeError: list indices must be integers or slices, not tuple` but I'm not sure what that means

I have added some docs. Any idea why the error `docstring of numpyro.distributions.continuous.CAR:14:Field list ends without a blank line; unexpected unindent.` is showing? I think it's to do with `:param numpy.ndarray or scipy.sparse.csr_matrix W:` but I'm not sure and not sure how to check.

numpyro/distributions/continuous.py
fehiepsi: You are right, I missed that code.

numpyro/distributions/continuous.py
```diff
@@ -1188,6 +1192,139 @@ def infer_shapes(
...
+
+    :param float or ndarray loc: mean of the multivariate normal
+    :param float alpha: autoregression parameter. For most cases, the value should lie
+        between 0 (sites are independent, collapses to an iid multivariate normal) and
+        1 (perfect autocorrelation between sites), but the specification allows for negative
+        correlations.
+    :param float tau: positive precision for the multivariate normal
+    :param numpy.ndarray or scipy.sparse.csr_matrix W: symmetric adjacency matrix where 1
+        indicates adjacency between sites and 0 otherwise
+    :param bool whether to use a sparse form of W in calculations (must be True is W is a
```
fehiepsi, line 1210: Probably you want: `:param bool is_sparse: ...`

numpyro/distributions/continuous.py
```diff
@@ -1188,6 +1192,139 @@ def infer_shapes(
...
 
 
+class CAR(Distribution):
+    r"""
+    The Conditional Autoregressive (CAR) distribution is a special case of the multivariate
+    normal in which the precision matrix is structured according to the adjacency matrix of
+    sites. The amount of autocorrelation between sites is controlled by :math:`\alpha`. The
+    distribution is a popular prior for areal spatial data.
+
+    :param float or ndarray loc: mean of the multivariate normal
```
fehiepsi, line 1202: nit: `ndarray` -> `np.ndarray`

numpyro/distributions/continuous.py
```diff
@@ -1188,6 +1192,139 @@ def infer_shapes(
...
+    :param float alpha: autoregression parameter. For most cases, the value should lie
+        between 0 (sites are independent, collapses to an iid multivariate normal) and
+        1 (perfect autocorrelation between sites), but the specification allows for negative
+        correlations.
+    :param float tau: positive precision for the multivariate normal
+    :param numpy.ndarray or scipy.sparse.csr_matrix W: symmetric adjacency matrix where 1
+        indicates adjacency between sites and 0 otherwise
+    :param bool whether to use a sparse form of W in calculations (must be True is W is a
+        scipy.sparse.csr_matrix)
```
fehiepsi, line 1211:
```
:class:`scipy.sparse.spmatrix`
```

numpyro/distributions/continuous.py
```diff
@@ -1188,6 +1192,139 @@ def infer_shapes(
...
+        )
+        event_shape = jnp.shape(self.W)[-1:]
+        self.loc = promote_shapes(loc, shape=batch_shape + event_shape)
+        self.alpha, self.tau = promote_shapes(alpha, tau, shape=batch_shape)
+
+        super(CAR, self).__init__(
+            batch_shape=batch_shape,
+            event_shape=event_shape,
+            validate_args=validate_args,
+        )
```
fehiepsi, line 1254: > add the checks for a correct W

I think you can add the check after this line:
```
if self._validate_args:
    if isinstance(W, np.ndarray):
        ...
    elif sparse.issparse(W):
        ...
    
```

numpyro/distributions/continuous.py
```diff
@@ -1188,6 +1192,139 @@ def infer_shapes(
...
+            self.W = W
+
+        batch_shape = lax.broadcast_shapes(
+            jnp.shape(loc)[:-1],
+            jnp.shape(alpha),
+            jnp.shape(tau),
+            jnp.shape(self.W)[:-2],
+        )
+        event_shape = jnp.shape(self.W)[-1:]
+        self.loc = promote_shapes(loc, shape=batch_shape + event_shape)
```
fehiepsi, line 1247: I guess the error that you got is due to this line. You need
```
self.loc, = promote...
```

numpyro/distributions/continuous.py
theorashid: The LKJ distribution just says "ndarray". I can open a fresh PR to change that if you want.

The reason I put `ndarray` is because it can be `jnp.ndarray` or `np.ndarray`. Does it matter or should I just put `np.ndarray`?

numpyro/distributions/continuous.py
theorashid: What does the , do?

numpyro/distributions/continuous.py
fehiepsi: > because it can be jnp.ndarray or np.ndarray

then I guess you can use `ndarray` here and at the `W` argument

numpyro/distributions/continuous.py
fehiepsi: the `promote_shapes` function returns a list with 1 element, that trick selects the first element of the list.

numpyro/distributions/continuous.py
theorashid: specifying `numpy.ndarray` is quite nice because it produces a link. But I'll just put `ndarray` for now

numpyro/distributions/continuous.py
theorashid: ah nice I didn't realise that sort of syntax worked without `, _`. Good to know!

numpyro/distributions/continuous.py
theorashid: I think this is right. Should it be `self._validate_args` or `self.validate_args`?

theorashid:
> More basic shape issues:
> 
> ```
> import numpyro.distributions as dist
> dist.CAR(loc=0,alpha=0.1,tau=0.1,
>     W=np.array(
>         [
>             [0.0, 1.0, 1.0, 0.0],
>             [1.0, 0.0, 0.0, 1.0],
>             [1.0, 0.0, 0.0, 1.0],
>             [0.0, 1.0, 1.0, 0.0],
>         ]
>     )
> ).sample(jax.random.PRNGKey(0))
> ```
> 
> gives the error `TypeError: list indices must be integers or slices, not tuple` but I'm not sure what that means

This error still persists also

numpyro/distributions/continuous.py
fehiepsi: `self._validate_args` is correct I think. I'm not sure about `self.validate_args`.

fehiepsi:
> the multipication step is possible using jax.experimental.sparse.BCOO matrices

By multiplication, do you mean matrix-vector product? It is surprised that it does not work. If BCOO can't perform the basic matrix-vector product for now, I think it is better to remove the `sparse` logic.

> This error still persists.

Could you try to add some print for `self.mean` or `self.precision_matrix` in the sample method to see where it goes wrong? I believe this is just a simple bug somewhere.

numpyro/distributions/continuous.py
```diff
@@ -1188,6 +1192,141 @@ def infer_shapes(
...
+
+            lam = jnp.linalg.eigvalsh(W_scaled)
+
+        n = D.shape[-1]
+
+        logtau = n * jnp.log(self.tau)
+        logdet = jnp.log1p(-jnp.expand_dims(self.alpha, -1) * lam).sum(-1)
+
+        logquad = self.tau * jnp.sum(
+            phi * (D * phi - jnp.expand_dims(self.alpha, -1) * W @ phi), -1
```
fehiepsi, line 1302: not sure if it helps, but you might want to do `(W @ phi)`

numpyro/distributions/continuous.py
```diff
@@ -1188,6 +1192,141 @@ def infer_shapes(
...
+
+            W = BCOO.from_scipy_sparse(W)
+
+        else:
+            D = W.sum(axis=-1)
+            D_rsqrt = D ** (-0.5)
+
+            W_scaled = W * (D_rsqrt * D_rsqrt[:, jnp.newaxis])
+
+            lam = jnp.linalg.eigvalsh(W_scaled)
```
fehiepsi, line 1294: Do you want to save the computation as we discussed in several threads? If so, you can do
```
if self.is_sparse:
    ...  # remove lam = ...
else:
    ... # remove lam = ...
if isinstance(W_scaled, np.ndarray):
    lam = np.linalg....
else:
    lam = jnp.linalg...
```

numpyro/distributions/continuous.py
```diff
@@ -1188,6 +1192,141 @@ def infer_shapes(
...
+
+            lam = np.linalg.eigvalsh(W_scaled)
+
+            W = BCOO.from_scipy_sparse(W)
+
+        else:
+            D = W.sum(axis=-1)
+            D_rsqrt = D ** (-0.5)
+
+            W_scaled = W * (D_rsqrt * D_rsqrt[:, jnp.newaxis])
```
fehiepsi, line 1292: Maybe `jnp.newaxis` -> `None`? Please ignore this if it works for numpy arrays.

numpyro/distributions/continuous.py
theorashid: I thought we wanted to use numpy for both branches because it will only do the computation once and we only need the calculation done once. Whereas I thought jax.numpy would repeat the calculation at each mcmc step.

So in fact, I should only have `lam = np.linalg.eigvalsh(W_scaled)` once outside of the if-else logic. Do you agree?

theorashid:
> By multiplication, do you mean matrix-vector product? It is surprised that it does not work. If BCOO can't perform the basic matrix-vector product for now, I think it is better to remove the `sparse` logic.

Sorry, I think I confused matters. Let's keep the sparse logic – definitely. I was wondering whether we could do the entire thing _only_ using the `jax.experimental.sparse.BCOO` array? It's up to you, the scipy docs think the CSC/CSR formats are faster the COO for multiplication (see [notes](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.coo_matrix.html)). But you might think it desirable to keep everything in jax/numpy arrays where possible. Up to you. Let me know

fehiepsi:
I see. The current implementation looks great to me. Supporting BCOO input seems to be complicated and unnecessary. We couldn't use the sparse form to compute the eigenvalues effectively so jax ops would be expensive in the sparse branch. It is better to encourage users using scipy/numpy inputs in the sparse branch to save computations. Adding a TODO to support jax sparse input looks good to me.

numpyro/distributions/continuous.py
fehiepsi: The tricky thing is if W_scaled is a jax array/tracer, we should use a jax operator for it.

numpyro/distributions/continuous.py
```diff
@@ -1188,6 +1192,141 @@ def infer_shapes(
...
+        )
+
+        if self._validate_args:
+            assert (
+                self.W.sum(axis=-1) > 0
+            ), "all sites in adjacency matrix W must have neighbours"
+
+            if sparse.issparse(self.W):
+                assert (self.W != self.W.T).nnz == 0, "W must be symmetric"
+            else:
```
fehiepsi, line 1265: Could you change this to only check if self.W is a numpy array? I don't think the code works for jax arrays/abstract arrays/tracers.

numpyro/distributions/continuous.py
theorashid: I think the current version is mathematically correct. Would the brackets speed anything up?

fehiepsi: I thought it might resolve the issue you mentioned with multiplication for jax sparse matrix. I don't actually know the order of the operators here so the parentheses might be helpful to see that matmul is performed before taking the multiplication. Not sure about the speed though. If you want to perform multiplication first then you'll need to expand alpha dims by (-2,-1). Given a batch of scalar numbers a, the multiplication of a and a matrix M is `a[...,:.:] * M`.

numpyro/distributions/continuous.py
theorashid: The numpy version works on `W_scaled` as jax. The only difference is the output. The jnp version outputs a jax array and the numpy version a numpy array. Equally, the jnp version can take in a numpy array.

If you believe the jax numpy version repeats the calculation at each step, I think we should only use the numpy version. The whole point in the stan article is to calculate the eigenvalues _once beforehand_

numpyro/distributions/continuous.py
theorashid: `np.array_equal(jnp.asarray(W), jnp.asarray(W).T)` works fine so I will leave it as is

numpyro/distributions/continuous.py
theorashid: multiplication is fine for jax sparse matrix. I was confused because there is no `.multiply`, but `@` and `*` work fine.

I _think_ the line is fine as it is. Fewer paretheses make it more readable and scalar * W is still a sparse jax matrix

theorashid:
> Could you try to add some print for `self.mean` or `self.precision_matrix` in the sample method to see where it goes wrong? I believe this is just a simple bug somewhere.

this error was fixed by me restarting my interactive session!

theorashid:
`sample()` seems to be working now. If we're happy with `W @ phi` and using `np.linalg` so calculations are only done once, I think it's time to write some tests.

I'm fairly new to tests in python and the numpyro testing suite is quite overwhelming. Could you give a quick overview of how they work?

In terms of combinations of inputs to test, I think we need:
- W as np.array, jax.numpy.array, scipy.sparse matrix
- alpha positive, alpha negative
- loc scalar, loc vector

tau is just a positive constant. Let me know whether to add more

numpyro/distributions/continuous.py
fehiepsi: Could you try to `jax.jit` any function using CAR with validation? I think some errors will happen. You can perform numpy operators on jax arrays outside of jit context but cannot inside a jit context. In principle, we should only use jax operators on jax arrays. You can try to jit something like
```
import jax
import numpy

@jax.jit
def f():
  x = jax.numpy.ones(10)
  assert numpy.all(x)
  return x

f()
```

theorashid: How about just using `(self.W == self.W.T).all()` instead of `np.array_equal()`. That will be jnp/np agnostic

numpyro/distributions/continuous.py
theorashid: I see, so np.linalg under jit might not work. Even converting the W_scaled jax array to a numpy array before the operation probably won't work under jit e.g.
```
@jax.jit
def f():
  x = jax.numpy.ones(10)
  x = np.asarray(x)
  return x

f()
```

Is there a workaround to avoid multiple computations of `lam` for the jax case?

numpyro/distributions/continuous.py
fehiepsi: Even so, `assert jax_array` will raise errors under jitting. A way to think about this is to consider jax arrays as abstract arrays which do not store any concrete values. Under jit, we only know a few things of a jax array: shape, dtype.

numpyro/distributions/continuous.py
fehiepsi: If we want to avoid computations, we need to treat those values as constant. You just need to provide numpy arrays and use numpy operators on the top of them. The only way we can access values of abstract arrays that I know is to use jax host_callback. But with host_callback, the computation is not cached and more expensive.

test/test_distributions.py
fehiepsi: I think we just need to add a few test cases here:
+ a vector of loc, a vector of alpha, a scalar of tau, a numpy matrix with is_sparse=True
+ a vector of loc, a scalar alpha, a vector of tau, a sparse matrix with is_sparse=False
+ a vector of loc, a scalar alpha, a scalar of tau,, a batch of numpy matrices with is_sparse=True
+ a batch of vectors of loc, a vector of alpha, a vector of tau, a batch of numpy matrices with is_sparse=False

theorashid:
Given our discussion on jitted jax arrays being placeholders where the values are never constant, I've chosen to not allow `W` as jax arrays. We should only be doing the lam calculation once. I think the only way to make it work would be to use `@partial(jit, static_argnums=(...)`, but given the jitting is done outside of the class definition, it isn't possible.

Do we need to add `D = jnp.asarray(D)`, `W = jnp.asarray(W)` and `lam = jnp.asarray(lam)` at the end of the non-sparse branch or are the lines 
```
logdet = jnp.log1p(-jnp.expand_dims(self.alpha, -1) * lam).sum(-1)
logquad = self.tau * jnp.sum(
    phi * (D * phi - jnp.expand_dims(self.alpha, -1) * W @ phi), -1
)
```

fine with numpy inputs?

test/test_distributions.py
theorashid: I'm not good at thinking about batching. Would you be able to show one of the examples with batching and I'll do the rest. Thanks

fehiepsi:
> the lines ... fine with numpy inputs?

Yes, I think so. Multiplying jax arrays with constant (i.e. numpy) arrays is a common operator I believe

> We should only be doing the lam calculation once

I guess we only need this when W is constant, otherwise, we should compute `lam` again when `W` has a dynamic value (e.g. when we want to do some inference over W). I feel that we don't need to restrict the flexibility here but it is up to you.

numpyro/distributions/continuous.py
```diff
@@ -1188,6 +1192,145 @@ def infer_shapes(
...
+
+    def __init__(self, loc, alpha, tau, W, *, is_sparse=False, validate_args=None):
+        if jnp.ndim(loc) == 0:
+            (loc,) = promote_shapes(loc, shape=(1,))
+
+        self.is_sparse = is_sparse
+
+        if self.is_sparse:
+            if not sparse.issparse(W):
+                assert isinstance(W, np.ndarray)
```
fehiepsi, line 1236: maybe `assert sparse.issparse(W) or (isinstance(W, np.ndarray), W.ndim == 2)` is clearer

numpyro/distributions/continuous.py
```diff
@@ -1188,6 +1192,145 @@ def infer_shapes(
...
+
+    :param float or ndarray loc: mean of the multivariate normal
+    :param float alpha: autoregression parameter. For most cases, the value should lie
+        between 0 (sites are independent, collapses to an iid multivariate normal) and
+        1 (perfect autocorrelation between sites), but the specification allows for negative
+        correlations.
+    :param float tau: positive precision for the multivariate normal
+    :param numpy.ndarray or scipy.sparse.csr_matrix W: symmetric adjacency matrix where 1
+        indicates adjacency between sites and 0 otherwise
+    :param bool is_sparse: whether to use a sparse form of W in calculations (must be True is
```
fehiepsi, line 1210: typo: must be True if

numpyro/distributions/continuous.py
```diff
@@ -1188,6 +1192,145 @@ def infer_shapes(
...
+        W is a :class:`scipy.sparse.spmatrix`)
```
fehiepsi, line 1211: Maybe we need to mention that currently, only supports 2-dimensional W.

numpyro/distributions/continuous.py
```diff
@@ -1188,6 +1192,145 @@ def infer_shapes(
...
+
+        if self._validate_args:
+            assert (
+                self.W.sum(axis=-1) > 0
+            ), "all sites in adjacency matrix W must have neighbours"
+
+            if sparse.issparse(self.W):
+                assert (self.W != self.W.T).nnz == 0, "W must be symmetric"
+            else:
+                assert np.array_equal(self.W, self.W.T), "W must be symmetric"
```
fehiepsi, line 1271: Maybe changing `self.W.T` to `np.swapaxes(self.W, -2, -1)` to not limit ourselves to 2D W.

numpyro/distributions/continuous.py
```diff
@@ -1188,6 +1192,145 @@ def infer_shapes(
...
+
+        super(CAR, self).__init__(
+            batch_shape=batch_shape,
+            event_shape=event_shape,
+            validate_args=validate_args,
+        )
+
+        if self._validate_args:
+            assert (
+                self.W.sum(axis=-1) > 0
```
fehiepsi, line 1265: You might need `.all()` at the end here.

numpyro/distributions/continuous.py
```diff
@@ -1188,6 +1192,145 @@ def infer_shapes(
...
+        mvn = MultivariateNormal(self.mean, precision_matrix=self.precision_matrix)
+        return mvn.sample(key, sample_shape=sample_shape)
+
+    @validate_sample
+    def log_prob(self, value):
+        phi = value - self.loc
+        W = self.W
+
+        if self.is_sparse:
+            D = np.asarray(W.sum(axis=-1)).squeeze()
```
fehiepsi, line 1284: Maybe `.squeeze(-1)` or `.squeeze(-2)` for clarity?

numpyro/distributions/continuous.py
fehiepsi: I think you need to expand alpha dims to `jnp.expand_dims(alpha, (-2, -1))` if we want to multiply first, but I might be wrong. Let's see what happens in tests.

> Would the brackets speed anything up?

Just think that performing matrix-vector would be better in case we have a batch of `alpha`. Doing `alpha * W` first will create a batch of matrices, which makes the follow-up matrix-vector product much more expensive. ;)

test/test_distributions.py
```diff
@@ -366,6 +366,21 @@ def get_sp_dist(jax_dist):
...
+        np.array(
+            [
+                [0.0, 1.0, 1.0, 0.0],
+                [1.0, 0.0, 0.0, 1.0],
+                [1.0, 0.0, 0.0, 1.0],
+                [0.0, 1.0, 1.0, 0.0],
+            ]
+        ),
+        False,
+    ),
```
fehiepsi, line 383: + a batch (with shape `(2,)`) of alpha
```
    T(
        dist.CAR,
        np.array([0., 1., 3., 4.]),
        np.array([-0.2, 0.3]),
        0.1,
        np.array(
            [
                [0.0, 1.0, 1.0, 0.0],
                [1.0, 0.0, 0.0, 1.0],
                [1.0, 0.0, 0.0, 1.0],
                [0.0, 1.0, 1.0, 0.0],
            ]
        ),
    ),
```
+ a batch (with shape `(2,)`) of tau
```
    T(
        dist.CAR,
        np.array([0., 1., 3., 4.]),
        np.array(0.1),
        np.array([0.3, 0.7]),
        np.array(
            [
                [0.0, 1.0, 1.0, 0.0],
                [1.0, 0.0, 0.0, 1.0],
                [1.0, 0.0, 0.0, 1.0],
                [0.0, 1.0, 1.0, 0.0],
            ]
        ),
    ),
```
+ a batch of W
```
    T(
        dist.CAR,
        np.array([0., 1., 3., 4.]),
        np.array(0.1),
        np.array([0.3, 0.7]),
        np.broadcast_to(np.array(
            [
                [0.0, 1.0, 1.0, 0.0],
                [1.0, 0.0, 0.0, 1.0],
                [1.0, 0.0, 0.0, 1.0],
                [0.0, 1.0, 1.0, 0.0],
            ]
        ), (2, 4, 4)),
    ),
```
+ sparse test
```
        partial(dist.CAR, is_sparse=True),
        np.array([0., 1., 3., 4.]),
        0.0,
        0.1,
        np.array(
            [
                [0.0, 1.0, 1.0, 0.0],
                [1.0, 0.0, 0.0, 1.0],
                [1.0, 0.0, 0.0, 1.0],
                [0.0, 1.0, 1.0, 0.0],
            ]
        ),
```

test/test_distributions.py
fehiepsi: I just added a couple of patterns in a comment below. Please mix them as desired :)

numpyro/distributions/continuous.py
theorashid: I think it will always only support 2-dimensional W. It's an adjacency matrix of sites. In a 1D problem, the matrix would be tri-diagonal with a diagonal 0 (1 for sites to left and right, 0 otherwise).

It's similar to how the covariance matrix K for Gaussian processes is always 2D (for single outputs).

Would you still like me to mention that W must be 2D?

numpyro/distributions/continuous.py
theorashid: I can't think of any occasions when W would not be 2D. Or what a multidimensional W would mean. It is a constant, a property of the sites' positions in space.

theorashid:
I cannot think of any occasion when we would want to do inference on `W` itself. W is a static property of the problem, and normally fed in as a `constant` in ppls rather than `data`.

Is it still okay to have W in `reparametrized_params`?

numpyro/distributions/continuous.py
fehiepsi: Sure. If you want to support only 2D inputs, we just need to mention so in the docstring and raise an error to ask users to make a feature request if they want a batch of matrices. I have seen several requests for batch GPs but supporting it would require us to rewrite a bunch of computations in pyro.contrib.gp. The situation here is simpler because the current code already almost works (modulo tests pass). In case you don't want to support batched W, just ignore the corresponding tests.

theorashid: What sort of occasions would a batched GP be useful? Multi-output?

When might a batched CAR be useful?

fehiepsi:
If so, I think you can remove it and also remove the corresponding entry in arg_constraint, so the distribution is still fully reparameterized.

numpyro/distributions/continuous.py
fehiepsi: Just think about this for simplicity: we typically use a model for 1 time series, but there are situations that we need to use that model for multiple time series. We can even define some shared hyperpriors to infer some properties across time series.

Some references for batch gps: https://docs.gpytorch.ai/en/latest/examples/08_Advanced_Usage/index.html#batch-gps or https://forum.pyro.ai/t/gaussian-processes-fit-using-plates/1742?u=fehiepsi

I would recommend not supporting it if you think it is unnecessary or complicated. Personally, I don't use CAR so I don't know how people uses it in practice (but to me, batching is useful in general).

numpyro/distributions/continuous.py
fehiepsi: I see. Then please just raise ValueError if W has batch dimensions.

numpyro/distributions/continuous.py
theorashid: Ahhh okay. Yes, for one of my models I am looking into using a GP with multiple outputs. I'm more used to defining them like [this](https://docs.gpytorch.ai/en/v1.5.1/examples/03_Multitask_Exact_GPs/Multitask_GP_Regression.html).

I haven't seen people doing batch CARs before. When I have seen multiple CAR effects in models, including my own, I have seen them defined as two separate effects rather than one call.

Let's stick to no batching for now. Given that the CAR distribution in 1D can be used to specify a gaussian random walk, I can see people _might_ want to use it for multiple time series, so I'll put in a TODO: support batched W. And we can see if there is demand. But I think batched GP demand will be far more pressing than batched CAR

numpyro/distributions/continuous.py
theorashid: I've just used assert W.ndim == 2, but is there anything else to put?

numpyro/distributions/continuous.py
theorashid: Unless of course, you think batched W is hardly any more work...

theorashid:
Not sure I understand the flake8 error 
```
flake8
./numpyro/distributions/continuous.py:1240:13: F631 assertion is always true, perhaps remove parentheses?
```

theorashid:
Needed to use `and` in the assert statement because non-empty tuples are always True.

Do you know why the workflows are failing? Seems to be on tests that I haven't created myself

theorashid:
Other than the workflow failings, I think this is pretty-near ready. So I've taken it out of the draft phase

fehiepsi:
> why the workflows are failing?

Could you merge the master branch? I think @OlaRonning has fixed those issues recently.

numpyro/distributions/continuous.py
fehiepsi: Yeah, I think it is unnecessary for now. We just need to left a message that it can be supported to encourage users to submit feature requests if they need. ;)

theorashid:
I never get this right and end up with some rebasing nightmare. What are the correct git commands?

OlaRonning:
Yes, they were fixed with #1400.

numpyro/distributions/continuous.py
```diff
@@ -1188,6 +1192,143 @@ def infer_shapes(
...
+        "tau",
+    ]
+
+    def __init__(self, loc, alpha, tau, W, *, is_sparse=False, validate_args=None):
+        if jnp.ndim(loc) == 0:
+            (loc,) = promote_shapes(loc, shape=(1,))
+
+        self.is_sparse = is_sparse
+
+        # TODO: support batched W
```
fehiepsi, line 1232: Could you hide the `scipy` dependency?
```
# This code is outside of class defintion
def _is_sparse(W):
    from scipy import sparse
    return sparse.issparse(W)

def _to_sparse(W):
    from scipy import sparse
    return sparse.csr_matrix(W)
```
and raise ValueError instead:
```
if W.ndim == 2:
    raise ValueError("Currently, we only support 2-dimensional W. Please make a feature request if you need higher dimensional W.")
if not (isinstance(W, np.ndarray) or _is_sparse(W)):
    raise ValueError("W needs to be a numpy array or a scipy sparse matrix. Please make a feature request if you need to support jax ndarrays.")
if self.is_sparse:
    # TODO: look into future jax sparse csr functionality and other developments
    self.W = _to_sparse(W)
else:
    assert not _is_sparse(W), "W is a sparse matrix so please specify `is_sparse=True`."
    self.W = W
```

numpyro/distributions/continuous.py
```diff
@@ -1188,6 +1192,143 @@ def infer_shapes(
...
+
+        super(CAR, self).__init__(
+            batch_shape=batch_shape,
+            event_shape=event_shape,
+            validate_args=validate_args,
+        )
+
+        if self._validate_args:
+            assert (
+                self.W.sum(axis=-1).all() > 0
```
fehiepsi, line 1263: I think you need `(self.W.sum(axis=-1) > 0).all()`

numpyro/distributions/continuous.py
```diff
@@ -1188,6 +1192,143 @@ def infer_shapes(
...
+            ), "all sites in adjacency matrix W must have neighbours"
+
+            if sparse.issparse(self.W):
```
fehiepsi, line 1266: nit: `if self.is_sparse`

test/test_distributions.py
```diff
@@ -366,6 +366,48 @@ def get_sp_dist(jax_dist):
...
+                [0.0, 1.0, 1.0, 0.0],
+                [1.0, 0.0, 0.0, 1.0],
+                [1.0, 0.0, 0.0, 1.0],
+                [0.0, 1.0, 1.0, 0.0],
+            ]
+        ),
+    ),
+    T(
+        partial(dist.CAR, is_sparse=True),
+        np.array([0.0, 1.0, 3.0, 4.0]),
```
fehiepsi, line 399: Maybe add batch dimensions to loc here?
```
np.array([[0.0, 1.0, 3.0, 4.0], [2., -1., -3., 2.]]),
```

test/test_distributions.py
```diff
@@ -366,6 +366,48 @@ def get_sp_dist(jax_dist):
         np.broadcast_to(np.identity(3), (2, 3, 3)),
         None,
     ),
+    T(
+        dist.CAR,
+        np.array([0.0, 1.0, 3.0, 4.0]),
```
fehiepsi, line 371: Because we support scalar `loc`, maybe using `1.2` here instead of an array?

fehiepsi (commented):
I just take another pass. The current implementation and tests look great to me. I just have a few small comments left.

> Implement ICAR

I think it is better to leave it for a future PR. This PR is already long.

fehiepsi:
Assume that remote upstream points to pyro-ppl and remote origin points to theorashid, you can do
```
git checkout car  # your current branch
git fetch upstream master
git merge upstream/master
git push origin car
```

theorashid:
Hmm something is failing but I'm not sure what. The log files won't load for me

numpyro/distributions/continuous.py
fehiepsi: Seems like we need (W @ phi[..., None]).squeeze(axis=-1) here in case `phi` has batch dimension.

test/test_distributions.py
```diff
@@ -366,6 +366,48 @@ def get_sp_dist(jax_dist):
...
+            [
+                [0.0, 1.0, 1.0, 0.0],
+                [1.0, 0.0, 0.0, 1.0],
+                [1.0, 0.0, 0.0, 1.0],
+                [0.0, 1.0, 1.0, 0.0],
+            ]
+        ),
+    ),
+    T(
+        partial(dist.CAR, is_sparse=True),
```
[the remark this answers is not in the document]
theorashid: nit: L185 uses `SparsePoisson` without the `_`.  Which would you like me to use? I can change the Poisson too

numpyro/distributions/continuous.py
theorashid: Sorry I only just saw the comment above. This completely makes sense:

> In dense case, even for scalar alpha, doing (alpha * W) @ phi would be more expensive than alpha * (W @ phi) (because alpha is multiplied with a matrix in the former while it is multiplied with a vector in the later).

I think `jnp.expand_dims(self.alpha, -1)` now that the multiplication by `alpha` is done after `W @ phi`, so I don't think we need to create `jnp.expand_dims(alpha, (-2, -1))`.

theorashid:
Sorry, I don't really understand the tests. There are too many for me!

fehiepsi:
No worries, I can take a look tomorrow

numpyro/distributions/continuous.py
```diff
@@ -1188,6 +1191,169 @@ def infer_shapes(
...
+    :param numpy.ndarray or scipy.sparse.csr_matrix W: symmetric adjacency matrix where 1
+        indicates adjacency between sites and 0 otherwise
+    :param bool is_sparse: whether to use a sparse form of W in calculations (must be True if
+        W is a :class:`scipy.sparse.spmatrix`)
+    """
+
+    arg_constraints = {
+        "loc": constraints.real_vector,
+        "alpha": constraints.open_interval(-1, 1),
+        "tau": constraints.positive,
```
fehiepsi, line 1228: Could you add `"W": constraints.dependent(is_discrete=False, event_dim=2)` again? Sorry for the confusion.

numpyro/distributions/continuous.py
```diff
@@ -1188,6 +1191,169 @@ def infer_shapes(
...
+    }
+    support = constraints.real_vector
+    reparametrized_params = [
+        "loc",
+        "alpha",
+        "tau",
```
fehiepsi, line 1234: Could you add `"W"` here? We can take derivative w.r.t. `W` when `W` is a jax device array. This is easier than adding some exceptions to several tests.

numpyro/distributions/continuous.py
```diff
@@ -1188,6 +1191,169 @@ def infer_shapes(
...
+            ), "W is a sparse matrix so please specify `is_sparse=True`."
+            # TODO: look into static jax ndarray representation
+            self.W = W
+
+        batch_shape = lax.broadcast_shapes(
+            jnp.shape(loc)[:-1],
+            jnp.shape(alpha),
+            jnp.shape(tau),
+            jnp.shape(self.W)[:-2],
+        )
```
fehiepsi, line 1269: Could you move the logic after line 1243 after this?
```diff
+        batch_shape = lax.broadcast_shapes(
+            jnp.shape(loc)[:-1],
+            jnp.shape(alpha),
+            jnp.shape(tau),
+            jnp.shape(W)[:-2],
+        )
+
+        if is_sparse:
+            if W.ndim != 2:
+                raise ValueError(
+                    "Currently, we only support 2-dimensional W. Please make a feature request",
+                    " if you need higher dimensional W.",
+                )
+            if not (isinstance(W, np.ndarray) or _is_sparse(W)):
+                raise ValueError(
+                    "W needs to be a numpy array or a scipy sparse matrix. Please make a feature",
+                    " request if you need to support jax ndarrays.",
+                )

-        # TODO: support batched W
-        if W.ndim != 2:
-            raise ValueError(
-                "Currently, we only support 2-dimensional W. Please make a feature request",
-                " if you need higher dimensional W.",
-            )
-        if not (isinstance(W, np.ndarray) or _is_sparse(W)):
-            raise ValueError(
-                "W needs to be a numpy array or a scipy sparse matrix. Please make a feature",
-                " request if you need to support jax ndarrays.",
-            )
-        if self.is_sparse:
             # TODO: look into future jax sparse csr functionality and other developments
             self.W = _to_sparse(W)
         else:
             assert not _is_sparse(
                 W
             ), "W is a sparse matrix so please specify `is_sparse=True`."
-            # TODO: look into static jax ndarray representation
-            self.W = W
+            self.W, = promote_shapes(W, shape=batch_shape + W.shape[-2:])
```

numpyro/distributions/continuous.py
```diff
@@ -1188,6 +1191,169 @@ def infer_shapes(
...
+        (self.loc,) = promote_shapes(loc, shape=batch_shape + event_shape)
+        self.alpha, self.tau = promote_shapes(alpha, tau, shape=batch_shape)
+
+        super(CAR, self).__init__(
+            batch_shape=batch_shape,
+            event_shape=event_shape,
+            validate_args=validate_args,
+        )
+
+        if self._validate_args:
```
fehiepsi, line 1280: `if self._validate_args and (isinstance(W, np.ndarray) or is_sparse):`

numpyro/distributions/continuous.py
```diff
@@ -1188,6 +1191,169 @@ def infer_shapes(
...
+            assert (
+                self.W.sum(axis=-1) > 0
+            ).all() > 0, "all sites in adjacency matrix W must have neighbours"
+
+            if self.is_sparse:
+                assert (self.W != self.W.T).nnz == 0, "W must be symmetric"
+            else:
+                assert np.array_equal(self.W, self.W.T), "W must be symmetric"
```
fehiepsi, line 1288: `assert np.array_equal(self.W, np.swapaxes(self.W, -2, -1)), "W must be symmetric"`

numpyro/distributions/continuous.py
```diff
@@ -1188,6 +1191,169 @@ def infer_shapes(
...
+
+            W_scaled = W.multiply(D_rsqrt).multiply(D_rsqrt[:, np.newaxis]).toarray()
+
+            W = BCOO.from_scipy_sparse(W)
+
+        else:
+            D = W.sum(axis=-1)
+            D_rsqrt = D ** (-0.5)
+
+            W_scaled = W * (D_rsqrt * D_rsqrt[:, np.newaxis])
```
fehiepsi, line 1312: `W_scaled = W * (D_rsqrt[..., None, :] * D_rsqrt[..., None])`

numpyro/distributions/continuous.py
```diff
@@ -1188,6 +1191,169 @@ def infer_shapes(
...
+
+        # TODO: look into sparse eignvalue methods
+        lam = np.linalg.eigvalsh(W_scaled)
```
fehiepsi, line 1315:
```
-        lam = np.linalg.eigvalsh(W_scaled)
+        if isinstance(W_scaled, np.ndarray):
+            lam = np.linalg.eigvalsh(W_scaled)
+        else:
+            lam = jnp.linalg.eigvalsh(W_scaled)
```

numpyro/distributions/continuous.py
```diff
@@ -1188,6 +1191,169 @@ def infer_shapes(
...
+
+        n = D.shape[-1]
+
+        logtau = n * jnp.log(self.tau)
+        logdet = jnp.log1p(-jnp.expand_dims(self.alpha, -1) * lam).sum(-1)
```
fehiepsi, line 1320: The Stan implementation is not normalized, could you add
```
logdet = logdet + jnp.log(D).sum(-1)
```

numpyro/distributions/continuous.py
```diff
@@ -1188,6 +1191,169 @@ def infer_shapes(
...
+            phi
+            * (
+                D * phi
+                - jnp.expand_dims(self.alpha, -1)
+                * (W @ phi[..., jnp.newaxis]).squeeze(axis=-1)
+            ),
+            -1,
+        )
+
+        return -0.5 * (logtau + logdet + logquad)
```
fehiepsi, line 1332:
```
-        return -0.5 * (logtau + logdet + logquad)
+        return 0.5 * (-n * jnp.log(2 * jnp.pi) + logtau + logdet - logquad)
```

numpyro/distributions/continuous.py
```diff
@@ -1188,6 +1191,169 @@ def infer_shapes(
...
+        D = W.sum(axis=-1, keepdims=True) * jnp.eye(W.shape[-1])
+        tau = jnp.expand_dims(self.tau, (-2, -1))
+        alpha = jnp.expand_dims(self.alpha, (-2, -1))
+        return tau * (D - alpha * W)
+
+    @staticmethod
+    def infer_shapes(loc, alpha, tau, W):
+        event_shape = W[-1:]
+        batch_shape = lax.broadcast_shapes(loc[:-1], alpha, tau, W[:-2])
+        return batch_shape, event_shape
```
fehiepsi, line 1354: We also need
```
+    def tree_flatten(self):
+        if self.is_sparse:
+            return (self.loc, self.alpha, self.tau), (self.is_sparse, self.W)
+        else:
+            return (self.loc, self.alpha, self.tau, self.W), (self.is_sparse,)
+
+    @classmethod
+        is_sparse = aux_data[0]
+        if is_sparse:
+            loc, alpha, tau = params
+            W = aux_data[1]
+        else:
+            loc, alpha, tau, W = params
+        return cls(loc, alpha, tau, W, is_sparse=is_sparse)
```

test/test_distributions.py
```diff
@@ -200,6 +200,11 @@ def tree_unflatten(cls, aux_data, params):
         return dist.FoldedDistribution.tree_unflatten(aux_data, params)
 
 
+class _SparseCAR(dist.CAR):
+    def __init__(self, loc, alpha, tau, W):
+        super().__init__(loc, alpha, tau, W, is_sparse=True)
```
fehiepsi, line 205:
```diff
-    def __init__(self, loc, alpha, tau, W):
-        super().__init__(loc, alpha, tau, W, is_sparse=True)
+    reparametrized_params = ["loc", "alpha", "tau"]
+
+    def __init__(self, loc, alpha, tau, W, *, is_sparse=True, validate_args=None):
+        super().__init__(loc, alpha, tau, W, is_sparse=True, validate_args=validate_args)
```

test/test_distributions.py
```diff
@@ -366,6 +371,48 @@ def get_sp_dist(jax_dist):
...
+        0.1,
+        np.array(
+            [
+                [0.0, 1.0, 1.0, 0.0],
+                [1.0, 0.0, 0.0, 1.0],
+                [1.0, 0.0, 0.0, 1.0],
+                [0.0, 1.0, 1.0, 0.0],
+            ]
+        ),
+    ),
```
fehiepsi, line 415: We need skip some tests here:
+ variance is not implemented
+ we can't jit/grad `W` parameter of SparseCAR

```diff
 _DIST_MAP = {
@@ -1024,6 +1026,7 @@ def test_jit_log_likelihood(jax_dist, sp_dist, params):
         "_ImproperWrapper",
         "LKJ",
         "LKJCholesky",
+        "_SparseCAR",
     ):
         pytest.xfail(reason="non-jittable params")

@@ -1378,6 +1381,9 @@ def test_log_prob_gradient(jax_dist, sp_dist, params):

     eps = 1e-3
     for i in range(len(params)):
+        if jax_dist is _SparseCAR and i == 3:
+            # skip taking grad w.r.t. W
+            continue
         if isinstance(
             params[i], dist.Distribution
         ):  # skip taking grad w.r.t. base_dist
@@ -1508,6 +1514,8 @@ def test_mean_var(jax_dist, sp_dist, params):
     else:
         if jnp.all(jnp.isfinite(d_jax.mean)):
             assert_allclose(jnp.mean(samples, 0), d_jax.mean, rtol=0.05, atol=1e-2)
+        if isinstance(d_jax, dist.CAR):
+            pytest.skip("CAR distribution does not have `variance` implemented.")
         if jnp.all(jnp.isfinite(d_jax.variance)):
```

fehiepsi (commented):
It is quite tricky to get some tests passed. >___<

numpyro/distributions/continuous.py
theorashid: Are we sure `jnp.log(2*jnp.pi)` is the only constant? In the [stan article](https://mc-stan.org/users/documentation/case-studies/mbjoseph-CARStan.html), they also drop `0.5 * log(c)` from the determinant term.

Do the constants matter?

theorashid:
Wow, that's a lot of work to get it passing tests. I'll push now and see if the tests are working while I think about parameter names – I'll see if there are analogous terms from random walk distributions.

A few questions:
- Do `lam = jnp.linalg.eigvalsh(W_scaled)` and putting `W` in `reparametrized_params` mean that we are allowing W as jax array? Or is this just so we can help it through some tests rather than changing the tests? Is it worth changing the documentation to jax array W allowed but discouraged?
- Does `(self.W,) = promote_shapes(W, shape=batch_shape + W.shape[-2:])` mean we now support batched W?

numpyro/distributions/continuous.py
fehiepsi: Yes, we need `0.5 * log(c)` (see the above comment at `logdet` term).

theorashid:
> Yes. There is a test that we perform jax.lax.scan over the distribution to see if it can create a new batched distribution. The new flatten, unflatten methods handle that logic.

Okay I'll change the docs / todos so that we now allow batched W. Pretty cool if someone actually uses this. I have 2 CAR effects in my model, maybe we should have a go writing this as a batch! I'll post in the forum when I do. I'll update the docs for jax W too.

>tau -> conditional precision
alpha -> spatial correlation
W -> adjacency matrix
or their abbreviations (like adj_matrix, spatial_corr,...). (just a suggestion, I believe you will have better names for them).

You can decide between any of these based on numpyro's naming conventions:
- tau -> `precision`/`conditional_precision`/`cond_precision`
- alpha -> `correlation`/`autocorrelation`
- W -> `adjacency_matrix`/`adj_matrix`

I've just been through some old spatial epidemiology textbooks to check that `tau` is marginal and `alpha` is indeed a correlation. I removed the word `spatial` because you can use this distribution for 1D, 2D, 3D etc, the same way you can GPs, you just need a graph.

numpyro/distributions/continuous.py
```diff
@@ -1188,6 +1191,193 @@ def infer_shapes(
...
+        if jnp.ndim(loc) == 0:
+            (loc,) = promote_shapes(loc, shape=(1,))
+
+        self.is_sparse = is_sparse
+
+        batch_shape = lax.broadcast_shapes(
+            jnp.shape(loc)[:-1],
+            jnp.shape(alpha),
+            jnp.shape(tau),
+            jnp.shape(self.W)[:-2],
```
fehiepsi, line 1249: typo: W -> self.W

fehiepsi:
I don't think numpyro has corresponding naming for those parameters so please use any of them (all of them look reasonable to me).

theorashid:
Hopefully that all runs without typos. Maybe give the new docs a quick look and check the maths?

fehiepsi (approved):
Thanks so much for this great contribution! This PR is a record: almost 180 comments so far (3x more than the runner-up). 💯

theorashid:
Amazing! Hopefully my next PR will need less effort

PR #1400, opened 2022-04-30
Fix/bvm sampler

OlaRonning:
Fixed sampling in sine distribution. The accept-reject sampler can accept only one dimension of the ACG envelope distribution.

### DIFF
--- a/numpyro/contrib/einstein/kernels.py
+++ b/numpyro/contrib/einstein/kernels.py
@@ -4,14 +4,15 @@
 from abc import ABC, abstractmethod
 from typing import Callable, Dict, List, Tuple
 
+import numpy as np
+
+from jax import random
 import jax.numpy as jnp
 import jax.scipy.linalg
 import jax.scipy.stats
-import numpy as np
-from jax import random
 
-import numpyro.distributions as dist
 from numpyro.contrib.einstein.util import safe_norm, sqrth_and_inv_sqrth
+import numpyro.distributions as dist
 
 
 class PrecondMatrix(ABC):
@@ -337,10 +338,6 @@ class MixtureKernel(SteinKernel):
         for kf in self.kernel_fns:
             rng_key, krng_key = random.split(rng_key)
             kf.init(krng_key, particles_shape)
-
-
-class GaussianProbabilityProductKernel(SteinKernel):
-    pass
 
 
 class HessianPrecondMatrix(PrecondMatrix):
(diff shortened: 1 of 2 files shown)


test/test_pickle.py
```diff
@@ -4,9 +4,11 @@
 import pickle
 
 import numpy as np
+from numpy.testing import assert_allclose
 import pytest
 
-from jax import random, test_util
+from jax import random, tree_multimap
+from jax._src.tree_util import tree_all
```
fehiepsi, line 11: It is better to import from `jax.tree_util` namespace.

fehiepsi (commented):
LGTM. Do you want to add tests to prevent issues like this? The reshape logic in the implementation is a bit complicated to follow.

numpyro/distributions/discrete.py
```diff
@@ -316,7 +316,7 @@ def log_prob(self, value):
         value = jnp.broadcast_to(value, batch_shape + (1,))
         logits = self.logits
         log_pmf = jnp.broadcast_to(logits, batch_shape + jnp.shape(logits)[-1:])
-        return jnp.take_along_axis(log_pmf, value, axis=-1)[..., 0]
+        return jnp.take_along_axis(log_pmf, jnp.array(value, dtype=int), axis=-1)[..., 0]
```
OlaRonning, line 319: Maybe the dtype should be fixed upstream.

OlaRonning:
I'll simplify the logic a bit and add tests.

test/contrib/test_module.py
```diff
@@ -7,7 +7,8 @@
 from numpy.testing import assert_allclose
 import pytest
 
-from jax import random, test_util
+from jax import random, tree_map
```
fehiepsi, line 10: nit: it is better to import `tree_map` from `tree_util` (see [this comment](https://github.com/google/jax/blob/main/jax/__init__.py#L104))

numpyro/distributions/discrete.py
```diff
@@ -316,7 +316,9 @@ def log_prob(self, value):
         value = jnp.broadcast_to(value, batch_shape + (1,))
         logits = self.logits
         log_pmf = jnp.broadcast_to(logits, batch_shape + jnp.shape(logits)[-1:])
-        return jnp.take_along_axis(log_pmf, value, axis=-1)[..., 0]
+        return jnp.take_along_axis(log_pmf, jnp.array(value, dtype=int), axis=-1)[
```
fehiepsi, line 319: Why do we need this change? I thought `value` is already `int`?

fehiepsi: Oh, I see. One of the tests is failing. Could you update [this line](https://github.com/pyro-ppl/numpyro/blob/master/test/contrib/test_infer_discrete.py#L457) to `z in [0, 1]`?

PR #1404, opened 2022-05-13
Add inference guide to readme

martinjankowiak:
We have a lot more inference algorithms then we once did so it's useful to offer some kind of guide in the README


fehiepsi (approved):
This overview is super helpful!! Thanks, Martin!

martinjankowiak:
@fehiepsi we should double check which mcmc algorithms automatically do enumeration and which do not so that that information can be incorporated into the overview

fehiepsi:
I think all mcmc algorithms currently do enumeration automatically (including nested sampler). For DiscreteHMCGibbs, we will do enumeration (if possible) over discrete sites that are marked "enumerated".

martinjankowiak:
what about e.g. HCMCECS? i guess there you'd need to compute a proxy for the enumerated potential which probably isn't supported without additional work?

fehiepsi:
You are right. It is fine for the default value `proxy=None`. We need to raise an error for taylor proxy.

PR #1415, opened 2022-06-01
small typos in mortality example docs

theorashid:
`pi` should be indexed from 0 (sorry, I'm used to R) and there's no need for `*` for multiply in the docs


fehiepsi (approved):
Thanks, @theorashid !

PR #1419, opened 2022-06-03
Fix edge case of Categorical due to simplex numerical issues

### LINKED ISSUE #1402: Sampling from posterior predictive with Categorical and AutoDiagonalNormal occasionally samples outside support
We encountered a subtle problem when sampling from the posterior predictive distribution for a model containing `Categorical` and using the `AutoDiagonalNormal`: In some rare cases, the sample drawn from the `Categorical` would be one large than the maximum category index. E.g., a model like
```
def model(num_categories):
    sample("ps", Dirichlet(jnp.ones(num_categories)))
    sample("x", Categorical(probs=ps))
```
may sample a value of `num_categories` even though the support of the categorical distribution is `(0, ..., num_categories-1)`.

This seems to arise from the combination of
- `numpyro.distributions.util._categorical` (and hence `numpyro.distributions.CategoricalProbs`) [expecting the passed probability vector to sum up exactly to one](https://github.com/pyro-ppl/numpyro/blob/1dbe63f72f1cd8befb76d0b8e7b4f284e7852404/numpyro/distributions/util.py#L187)
- `AutoDiagonalNormal` (or probably rather the `StickBreakingTransform`) not guaranteeing that due to numerical imprecision

We observed samples from the `AutoDiagonalNormal` guide for which the probability vector `ps` summed up to something just shy of one. In rare cases, the uniform (0,1) sample used in `numpyro.distributions.util._categorical` falls above the total sum, resulting in the function to return the largest category index plus one as a value.

Our current quick-and-dirty workaround for this is to simply clamp the sampled values to the support but surely there are better ways of handling this, but I haven't given it much thought. Naively it would seem that explicit renormalisation of the probability vector at some point would be the way to go, but I'm not sure that will actually be robust towards the numerical issues that seem to lead to this in the first place..

fehiepsi:
Fixes #1402

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -9,6 +9,7 @@ multipledispatch
 nbsphinx==0.8.6
 numpy
 optax>=0.0.6
+pyyaml
 readthedocs-sphinx-search==0.1.0
 sphinx==4.0.3
 sphinx-gallery
--- a/numpyro/distributions/discrete.py
+++ b/numpyro/distributions/discrete.py
@@ -65,7 +65,9 @@ def _to_probs_multinom(logits):
 
 def _to_logits_multinom(probs):
     minval = jnp.finfo(jnp.result_type(probs)).min
-    return jnp.clip(jnp.log(probs) - logsumexp(probs, axis=-1, keepdims=True), a_min=minval)
+    return jnp.clip(
+        jnp.log(probs) - logsumexp(probs, axis=-1, keepdims=True), a_min=minval
+    )
 
 
 class BernoulliProbs(Distribution):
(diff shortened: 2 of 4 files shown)


ordabayevy (commented):
LGTM! Can you see if https://github.com/pyro-ppl/funsor/issues/594 is also fixed?

`FUNSOR_BACKEND=jax pytest test_distribution.py -k test_categorical_event_dim_conversion`

fehiepsi:
Thanks for taking a look, Yerdos. There are many tests are failing after this change. I'm not sure why, will take a closer look later this week.

fehiepsi:
@ordabayevy It seems that funsor test passes after this.

ordabayevy (commented):
Looks good to me. Did you figure out what was wrong with the normalization in `_to_logits_multinom`?

ordabayevy (approved):

fehiepsi:
The formula that I used is wrong. We should do `log(probs/probs.sum(-1, True))`

fehiepsi:
Thanks for reviewing, Yerdos!

Issue #1398, opened 2022-04-28
Log Uniform distribution


andrewfowlie:
How about a Log Uniform distribution, i.e., the log of a variable is uniformly distributed. Implementation:

```
class LogUniform(dist.Uniform):
    def sample(self, key, sample_shape=()):
        shape = sample_shape + self.batch_shape
        sample = jax.random.uniform(key, shape=shape, minval=jnp.log(self.low), maxval=jnp.log(self.high))
        return jnp.exp(sample)
```

A more polished version would check that low and high are both > 0. This is rather simple for users to do, but quite convenient.

fehiepsi:
Please feel free to submit a PR. I think we can also use:
```
d = dist.TransformedDistribution(dist.Uniform(low, high), dist.transforms.ExpTransform())
```

yayami3:
can be close by #1423

fehiepsi:
Thanks @yayami3!

PR #1423, opened 2022-06-07
add LogUniform

andrewfowlie:
adds loguniform dist as in #1398. note

- no docs yet. no docstrings in other distributions. how do I doc this?
- no tests yet, I'm looking into it
- the low and high bounds are those of the un-logged parameter, i.e. `LogUniform(1, 10)` means `log_10 X` is uniform between 0 and 1.

### DIFF
--- a/numpyro/distributions/continuous.py
+++ b/numpyro/distributions/continuous.py
@@ -1033,6 +1033,33 @@ class Logistic(Distribution):
         return self.loc + self.scale * logit(q)
 
 
+class LogUniform(TransformedDistribution):
+    arg_constraints = {"low": constraints.positive, "high": constraints.positive}
+    support = constraints.positive
+    reparametrized_params = ["low", "high"]
+
+    def __init__(self, low=1., high=10., *, validate_args=None):
+        base_dist = Uniform(jnp.log(low), jnp.log(high))
+        self.low, self.high = low, high
+        super(LogUniform, self).__init__(
+            base_dist, ExpTransform(), validate_args=validate_args
+        )
+
+    @property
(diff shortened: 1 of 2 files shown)


numpyro/distributions/continuous.py
```diff
@@ -1033,6 +1033,33 @@ def icdf(self, q):
         return self.loc + self.scale * logit(q)
 
 
+class LogUniform(TransformedDistribution):
```
fehiepsi, line 1036: You can see docs of some distributions [here](https://num.pyro.ai/en/stable/distributions.html#numpyro.distributions.continuous.LKJ).

fehiepsi (commented):
Hi @andrewfowlie, you can add some entries of LogUniform [here](https://github.com/pyro-ppl/numpyro/blob/master/test/test_distributions.py#L287). For docs, you can add an entry to sphinx [here](https://github.com/pyro-ppl/numpyro/blob/master/docs/source/distributions.rst).

numpyro/distributions/continuous.py
```diff
@@ -1033,6 +1033,33 @@ def icdf(self, q):
...
+    arg_constraints = {"low": constraints.positive, "high": constraints.positive}
+    support = constraints.positive
+    reparametrized_params = ["low", "high"]
+
+    def __init__(self, low=1., high=10., *, validate_args=None):
```
fehiepsi, line 1041: I think 1. and 10. are a bit arbitrary. How about removing their default values?

fehiepsi:
Close in favor of #1495

PR #1426, opened 2022-06-09
add AutoSemiDAIS

martinjankowiak:
@fehiepsi this is closer to working but there still appears to be some scaling issue.

run `pytest -s -v test_autoguide.py::test_autosemidais` to see apparent discrepancies.

### DIFF
--- a/numpyro/distributions/distribution.py
+++ b/numpyro/distributions/distribution.py
@@ -558,10 +558,9 @@ class ExpandedDistribution(Distribution):
             event_shape = jnp.shape(x)[batch_ndims:]
             return x.reshape(sample_shape + self.batch_shape + event_shape)
 
-        # TODO: See why we need intermediates here
         intermediates = tree_util.tree_map(reshape_sample, intermediates)
         samples = reshape_sample(samples)
-        return samples, []
+        return samples, intermediates
 
     def rsample(self, key, sample_shape=()):
         return self._sample(
@@ -580,7 +579,8 @@ class ExpandedDistribution(Distribution):
     def sample(self, key, sample_shape=()):
         return self.sample_with_intermediates(key, sample_shape)[0]
 
-    def log_prob(self, value):
+    def log_prob(self, value, intermediates=None):
+        # TODO: utilize `intermediates`
         shape = lax.broadcast_shapes(
             self.batch_shape,
             jnp.shape(value)[: max(jnp.ndim(value) - self.event_dim, 0)],
--- a/numpyro/distributions/distribution.py
+++ b/numpyro/distributions/distribution.py
@@ -558,9 +558,10 @@ class ExpandedDistribution(Distribution):
             event_shape = jnp.shape(x)[batch_ndims:]
             return x.reshape(sample_shape + self.batch_shape + event_shape)
 
+        # TODO: See why we need intermediates here
         intermediates = tree_util.tree_map(reshape_sample, intermediates)
         samples = reshape_sample(samples)
-        return samples, intermediates
+        return samples, []
 
     def rsample(self, key, sample_shape=()):
         return self._sample(
--- a/numpyro/handlers.py
+++ b/numpyro/handlers.py
@@ -259,11 +259,13 @@ class block(Messenger):
        >>> assert 'b' in trace_block_a
     """
 
-    def __init__(self, fn=None, hide_fn=None, hide=None):
+    def __init__(self, fn=None, hide_fn=None, hide=None, expose_types=None):
         if hide_fn is not None:
             self.hide_fn = hide_fn
         elif hide is not None:
             self.hide_fn = lambda msg: msg.get("name") in hide
+        elif expose_types is not None:
+            self.hide_fn = lambda msg: msg.get("type") not in expose_types
         else:
             self.hide_fn = lambda msg: True
         super(block, self).__init__(fn)
--- a/numpyro/handlers.py
+++ b/numpyro/handlers.py
@@ -259,13 +259,11 @@ class block(Messenger):
        >>> assert 'b' in trace_block_a
     """
 
-    def __init__(self, fn=None, hide_fn=None, hide=None, expose_types=None):
+    def __init__(self, fn=None, hide_fn=None, hide=None):
         if hide_fn is not None:
             self.hide_fn = hide_fn
         elif hide is not None:
             self.hide_fn = lambda msg: msg.get("name") in hide
-        elif expose_types is not None:
-            self.hide_fn = lambda msg: msg.get("type") not in expose_types
         else:
             self.hide_fn = lambda msg: True
         super(block, self).__init__(fn)
--- a/numpyro/infer/autoguide.py
+++ b/numpyro/infer/autoguide.py
@@ -893,11 +893,25 @@ class AutoSemiDAIS(AutoGuide):
 
     Usage::
 
-        base_guide = AutoNormal(model)
-        guide = AutoDAIS(model, base_guide)
+        def global_model():
+            return numpyro.sample("theta", dist.Normal(0, 1))
+
+        def local_model(theta):
+            with numpyro.plate("data", 8, subsample_size=2):
+                tau = numpyro.sample("tau", dist.Gamma(5.0, 5.0))
+                numpyro.sample("obs", dist.Normal(0.0, tau), obs=jnp.ones(2))
+
+        model = lambda: local_model(global_model())
+        base_guide = AutoNormal(global_model)
+        guide = AutoSemiDAIS(model, local_model, base_guide, K=4)
         svi = SVI(model, guide, ...)
 
     :param callable model: A NumPyro model.
+    :param callable local_model: The portion of `model` that includes the local latent variables only.
+        The signature of `local_model` should be the return type of the global model with global latent
+        variables only.
+    :param callable base_guide: A guide for the global latent variables, e.g. an autoguide.
+        The return type should be a dictionary of latent sample sites and corresponding samples.
     :param str prefix: A prefix that will be prefixed to all param internal sites.
     :param int K: A positive integer that controls the number of HMC steps used.
         Defaults to 4.
@@ -907,8 +921,8 @@ class AutoSemiDAIS(AutoGuide):
         Defaults to 0.1.
     :param float gamma_init: The initial value of the learnable damping factor used
         during partial momentum refreshments in HMC. Defaults to 0.9.
-    :param float init_scale: Initial scale for the standard deviation of the base
-        variational distribution for each (unconstrained transformed) latent variable. Defaults to 0.1.
+    :param float init_scale: Initial scale for the standard deviation of the variational
+        distribution for each (unconstrained transformed) local latent variable. Defaults to 0.1.
     """
 
     def __init__(
@@ -917,11 +931,11 @@ class AutoSemiDAIS(AutoGuide):
         local_model,
         base_guide,
         *,
+        prefix="auto",
         K=4,
         eta_init=0.01,
         eta_max=0.1,
         gamma_init=0.9,
-        prefix="auto",
         init_scale=0.1,
     ):
         # TODO does the value of init_loc_fn used here matter at all?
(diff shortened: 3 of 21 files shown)


numpyro/infer/autoguide.py
```diff
@@ -864,6 +865,314 @@ def _single_sample(_rng_key):
...
+            value = transform(unconstrained_value)
+            event_ndim = site["fn"].event_dim
+            if numpyro.get_mask() is False:
+                log_density = 0.0
+            else:
+                log_density = -transform.log_abs_det_jacobian(
+                    unconstrained_value, value
+                )
+                # Q: is this the right sign?
+                # Q: is (N / subsample_size) correct?
```
martinjankowiak, line 1027: is (N / subsample_size) correct?

numpyro/infer/autoguide.py
```diff
@@ -864,6 +865,314 @@ def _single_sample(_rng_key):
...
+                log_density = (N / subsample_size) * sum_rightmost(
+                    log_density, jnp.ndim(log_density) - jnp.ndim(value) + event_ndim
+                )
+            delta_dist = dist.Delta(
+                value, log_density=log_density, event_dim=event_ndim
+            )
+            result[name] = numpyro.sample(name, delta_dist)
+
+        # Q: does this need global_latents inside?
```
martinjankowiak, line 1036: does this need global_latents inside?

numpyro/infer/autoguide.py
```diff
@@ -864,6 +865,314 @@ def _single_sample(_rng_key):
...
+                eps, beta = eps_beta
+                eta = eta0 + eta_coeff * beta
+                eta = jnp.clip(eta, a_min=0.0, a_max=self.eta_max)
+                assert eps.shape == (subsample_size, D)
+                assert eta.shape == beta.shape == (subsample_size,)
+                z_prev, v_prev, log_factor = carry
+                z_half = z_prev + v_prev * eta[:, None] * inv_mass_matrix
+                q_grad = (1.0 - beta[:, None]) * grad(base_z_dist_log_prob)(z_half)
+                # Note: rescale log_density to be order 1.
+                # Q: is it true that local_log_density is scaled by (N / subsample_size) because of the plate?
```
martinjankowiak, line 1131: is it true that local_log_density is scaled by (N / subsample_size) because of the plate?

numpyro/infer/autoguide.py
```diff
@@ -864,6 +865,314 @@ def _single_sample(_rng_key):
...
+            assert eps.shape == (subsample_size, D, K)
+            assert betas.shape == (subsample_size, K)
+
+            eps_T = jnp.moveaxis(eps, -1, 0)
+            (z, _, log_factor), _ = jax.lax.scan(
+                scan_body, (z_0, v_0, jnp.zeros(subsample_size)), (eps_T, betas.T)
+            )
+            assert log_factor.shape == (subsample_size,)
+
+            # Q: will this be scaled by (N / subsamplesize) because it's in the plate?
```
martinjankowiak, line 1153: will this be scaled by (N / subsamplesize) because it's in the plate?

test/infer/test_autoguide.py
```diff
@@ -670,3 +675,146 @@ def model():
...
+    print("svi_result16.params[auto_eta_coeff].mean()", svi_result16.params['auto_eta_coeff'].mean(), " (should be positive)")
+    print("svi_result16.params[auto_eta].mean()", svi_result16.params['auto_eta0'].mean())
+
+    guide8 = AutoSemiDAIS(model8, partial(local_model, 8), base_guide, K=4, eta_max=0.25, eta_init=0.005)
+    with handlers.seed(rng_seed=0):
+        guide8()  # initialize guide since we are not training this guide
+        # instead we reuse svi_result16 parameters
+    samples8 = guide8.sample_posterior(random.PRNGKey(1), svi_result16.params)
+    assert samples8["theta"].shape == (P,) and samples8["tau"].shape == (8,)
+    dais_elbo8 = jax.vmap(lambda k: -Trace_ELBO().loss(k, svi_result16.params, model8, guide8))(random.split(random.PRNGKey(0), 5000)).mean().item()
```
martinjankowiak, line 725: is there any reason reusing `svi_result16` like this would be incorrect?

fehiepsi (commented):
@martinjankowiak I just removed the scaling factor of the grad term. It seems to work.

test/infer/test_autoguide.py
fehiepsi: I'm not sure. It seems to be valid.

martinjankowiak:
> @martinjankowiak I just removed the scaling factor of the grad term. It seems to work.

this only appears to work because it's driven into a regime where eta ~= 0 and the hmc dynamics are trivial

```
    print(
        "svi_result16.params[auto_eta_coeff].mean()",
        svi_result16.params["auto_eta_coeff"].mean(),
        " (should be positive)",
    )
```

numpyro/infer/autoguide.py
martinjankowiak: if this is true we need the multiplicative factor in 

```
p_grad = beta[:, None] * (subsample_size / N) * grad(local_log_density)(z_half)
```
because `p_grad` and `q_grad` need to have the same scale. `q_grad` comes from `base_z_dist_log_prob` so i guess it should be unscaled

test/infer/test_autoguide.py
martinjankowiak: is it possible the parameters get mixed up somehow? so that the local parameters get assigned to the wrong datapoints? e.g. `z_0_loc`

martinjankowiak:
@fehiepsi i think the mean thing to understand right now is why `dais_elbo8 != dais_elbo16`. these two code paths are supposed be computing the same elbos. the only difference should be the mini-batch size. in particular they should be equal in expectation and only differ by variance, which should be small for a large number of samples.

numpyro/infer/autoguide.py
fehiepsi: You  are right, just double check that we need such factor here.

fehiepsi:
@martinjankowiak I just fixed a bug at unconstrained_reparam. Could you take a look? The elbos seem to be consistent now.

numpyro/infer/util.py
```diff
@@ -217,8 +217,6 @@ def _unconstrain_reparam(params, site):
         log_det = sum_rightmost(
             log_det, jnp.ndim(log_det) - jnp.ndim(value) + len(site["fn"].event_shape)
         )
-        if site["scale"] is not None:
-            log_det = site["scale"] * log_det
```
martinjankowiak, line 221: @fehiepsi can you please explain how exactly this logic effects our autoguide?

martinjankowiak: i'd like to understand better where exactly this bug would be triggered. only inside a plate? both factor and sample? etc. in particular i'd like to understand which different code path in local.py avoided this issue...

martinjankowiak: in any case the test now behaves as expected : ) thanks! 💯

numpyro/distributions/distribution.py
```diff
@@ -558,9 +558,10 @@ def reshape_sample(x):
             event_shape = jnp.shape(x)[batch_ndims:]
             return x.reshape(sample_shape + self.batch_shape + event_shape)
 
+        # TODO: See why we need intermediates here
```
martinjankowiak, line 561: we still need to double check this...

numpyro/infer/autoguide.py
```diff
@@ -864,6 +865,373 @@ def _single_sample(_rng_key):
...
+                tau = numpyro.sample("tau", dist.Gamma(5.0, 5.0))
+                numpyro.sample("obs", dist.Normal(0.0, tau), obs=jnp.ones(2))
+
+        model = lambda: local_model(global_model())
+        base_guide = AutoNormal(global_model)
+        guide = AutoSemiDAIS(model, local_model, base_guide, K=4)
+        svi = SVI(model, guide, ...)
+
+    :param callable model: A NumPyro model.
+    :param callable local_model: The portion of `model` that includes the local latent variables only.
```
martinjankowiak, line 910: are these the right/complete requirements for local_model and base_guide?

numpyro/infer/autoguide.py
```diff
@@ -864,6 +865,373 @@ def _single_sample(_rng_key):
...
+        base_guide,
+        *,
+        prefix="auto",
+        K=4,
+        eta_init=0.01,
+        eta_max=0.1,
+        gamma_init=0.9,
+        init_scale=0.1,
+    ):
+        # TODO does the value of init_loc_fn used here matter at all?
```
martinjankowiak, line 941: this is only used to setup `_local_potential_fn_gen` right? i.e. the value of `init_loc_fn` doesn't matter?

numpyro/infer/autoguide.py
```diff
@@ -864,6 +865,373 @@ def _single_sample(_rng_key):
...
+        subsample_axes = {}
+        plate_dim = None
+        for name, site in self.prototype_trace.items():
+            if site["type"] == "sample" and not site["is_observed"]:
+                for frame in site["cond_indep_stack"]:
+                    if frame.name == plate_name:
+                        if plate_dim is None:
+                            plate_dim = frame.dim
+                        local_vars.append(name)
+                        # if name in self._init_locs:  # Q: is it ok i removed this in favor of is_observed?
```
martinjankowiak, line 989: is this ok?

numpyro/infer/autoguide.py
```diff
@@ -864,6 +865,373 @@ def _single_sample(_rng_key):
...
+            for name, unconstrained_value in self._unpack_local_latent(
+                local_flat
+            ).items():
+                site = self.prototype_trace[name]
+                transform = biject_to(site["fn"].support)
+                value = transform(unconstrained_value)
+                results[name] = value
+            return results
+
+        if sample_shape:
```
martinjankowiak, line 1229: still need to implement this

martinjankowiak:
@fehiepsi ok i think this should be good to go once we resolve the remaining comments/todos above.

any additional tests we should add?

numpyro/infer/util.py
fehiepsi: As in the new tests that I added, when we do unconstrained reparam, we introduce new factor terms. Those factor terms will belong to the same context as the corresponding latent variables, i.e. they will have the same `scale` field in their messages. So with those lines, we are scaling logdet terms 2 times.

martinjankowiak: ah gotcha i think that makes sense

numpyro/distributions/distribution.py
fehiepsi: I think we can add `intermediates` arguments to log_prob method to utilize this information. This is useful for Expanded(TransformDistribution).

numpyro/handlers.py
```diff
@@ -259,11 +259,13 @@ class block(Messenger):
        >>> assert 'b' in trace_block_a
     """
 
-    def __init__(self, fn=None, hide_fn=None, hide=None):
+    def __init__(self, fn=None, hide_fn=None, hide=None, expose_types=None):
```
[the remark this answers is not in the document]
fehiepsi: I think we can revert this change because we don't need this. Or keep this as-is if you think expose_types is useful for something else

numpyro/infer/autoguide.py
fehiepsi: I think so. Which information else that you have in mind?

numpyro/distributions/distribution.py
martinjankowiak: ok i'll leave this to you 👍

numpyro/infer/autoguide.py
martinjankowiak: is it enough to add `result.update(global_latents)`?

fehiepsi (approved):
LGTM!

numpyro/infer/autoguide.py
fehiepsi: that's right

martinjankowiak (commented):
@fehiepsi great thanks for all your help!

Issue #1422, opened 2022-06-07
Stein-bnn Confidence Interval plot


caxelrud:
I modified the stein_bnn example to include a CI plot.
At this moment, I am using the training data to check the model.
Not knowing the results of the prediction function, I used the several results to create the mean and distribution information and used in the plot.
Is it valid?
Do you recommend another method?  
Attached the jupyter notebook file.
[stein_bnn.zip](https://github.com/pyro-ppl/numpyro/files/8849789/stein_bnn.zip)

OlaRonning:
Hi @caxelrud,

Would you mind uploading your code as a [gist](https://gist.github.com/) and linking it instead of the zip file? Then I'd love to have a look at what you've done.

caxelrud:
I created a gist at:
https://gist.github.com/caxelrud/66288633f771da52c893736e9684d65f

OlaRonning:
@caxelrud, your final plot shows the uncertainty (variance) due to the latent variable (the variance of the Stein particles). If you are interested in the epistemic uncertainty (uncertainty in likelihood term) you should increase the number of samples (`num_samples`) in `Predictive`
``` python
pred = Predictive(
        model,
        guide=stein.guide,
        params=stein.get_params(result.state),
        num_samples=1,  # <---- increase to assess epistemic uncertainity
        batch_ndims=1
    )
```
Feel free to open a PR updating the example with your CI plot.

PR #1430, opened 2022-06-15
Removed the safe norm in Stein kernels

OlaRonning:
The `safe_norm` in the Stein kernels protects against zero division when using AD on `linalg.norm` for particles at the same position. However, the norms are subsequently squared, so by explicitly canceling the square root of the euclidean distance and the squaring in code; we avoid manifesting a division in the backward pass of AD. Therefore we can do away with the safe norm (and avoid the pesky error introduced by adding `1e-5` when particles collapse). 

(I also added the visual from #1422)


numpyro/contrib/einstein/kernels.py
```diff
@@ -7,11 +7,12 @@
 import numpy as np
 
 from jax import random
+from jax.lax import stop_gradient
 import jax.numpy as jnp
 import jax.scipy.linalg
 import jax.scipy.stats
 
-from numpyro.contrib.einstein.util import safe_norm, sqrth_and_inv_sqrth
+from numpyro.contrib.einstein.util import median_bandwidth, sqrth_and_inv_sqrth
```
OlaRonning, line 15: Gathered the median bandwidth computation in a single function as it was computed in multiple (some incorrect) ways in the code.

fehiepsi (approved):
LGTM, thanks Ola!

PR #1435, opened 2022-06-20
fixed bug in rbf kernel

OlaRonning:
Fixed a missing parenthesis in RBF kernel, and reverted all test cases to correct values.


OlaRonning:
Thanks, @fehiepsi. I should have realized myself.

PR #1436, opened 2022-06-20
set self._postprocess_fn = None in BarkerMH

martinjankowiak:
i found this is necessary to use BarkerMH with a `potential_fn` so i'm a bit confused why some of the tests in `test_mcmc.py` pass, but in any case some logic like this seems to be missing


fehiepsi:
Could you fix [this test](https://github.com/pyro-ppl/numpyro/blob/master/test/infer/test_mcmc.py#L41) to cover this change?

fehiepsi (approved):

Issue #1433, opened 2022-06-19
A bug in AutoMultivariateNormal guide


xidulu:
The scale_tril should be parsed as the scale_tril argument of MVN instead of as the second positional argument (which corresponds to the covariance_matrix) I think?
https://github.com/pyro-ppl/numpyro/blob/master/numpyro/infer/autoguide.py#L1349

Otherwise, the covariance matrix returned by the guide seems to be the lower-cholesky factor.,
![Screen Shot 2022-06-19 at 1 58 14 PM](https://user-images.githubusercontent.com/26022201/174494236-86d4afad-2ef1-4f3a-b894-1db0cf00f515.png)

xidulu:
> Oops, what a bug! Could you submit the fix, @xidulu?

Sure : )

PR #1437, opened 2022-06-21
Fix get_posterior in AutoMultivariateNormal guide

xidulu:
Fix https://github.com/pyro-ppl/numpyro/issues/1433


fehiepsi (approved):
Thanks, @xidulu!

PR #1449, opened 2022-07-11
Move contrib.render functionality to infer.inspect

mattja:
Moving the `get_model_relations()` and `render_model()` to `numpyro.infer.inspect`, matching pyro interface.
Also updating documentation and tests accordingly.


numpyro/contrib/render.py
```diff
@@ -1,385 +0,0 @@
-# Copyright Contributors to the Pyro project.
-# SPDX-License-Identifier: Apache-2.0
-
-import itertools
```
fehiepsi, line 4: Can we keep this file with a single line for backward-compatibility.
```
from numpyro.infer.inspect import render
```

numpyro/__init__.py
```diff
@@ -4,7 +4,7 @@
 import logging
 
 from numpyro import compat, diagnostics, distributions, handlers, infer, ops, optim
-from numpyro.contrib.render import render_model
+from numpyro.infer.inspect import render_model
```
fehiepsi, line 7: nit: run `isort` and swap line 7 and 8

docs/source/utilities.rst
```diff
@@ -101,15 +101,9 @@ Model inspection
...
     :member-order: bysource
 
-.. autofunction:: numpyro.contrib.render.get_model_relations
-
 Visualization Utilities
 =======================
 
-render_model
-------------
-.. autofunction:: numpyro.contrib.render.render_model
```
fehiepsi, line 111: Can we keep this entry with the new path?

fehiepsi (commented):
(sorry for the late response!)

Thanks for the refactoring to keep API similar to Pyro! I just have a few minor comments

numpyro/contrib/render.py
mattja: Done. importing the four functions here for full backwards compatibility.

numpyro/__init__.py
mattja: Fixed this and the other 2 lint errors.

docs/source/utilities.rst
mattja: Done. Turned off automodule for `infer.inspect` to avoid duplicate entry.

fehiepsi (approved):
Thanks, @mattja!

Issue #1461, opened 2022-08-03
Binomial(...).sample has an infinite loop when combined with jax.vmap and p=0


tbenthompson:
This seems to be the same bug the cropped up in the [Poisson case](https://github.com/pyro-ppl/numpyro/issues/582). And is a result of [combining `lax.cond`, `vmap` and data-dependent inputs](https://github.com/google/jax/issues/2947):
- one of the branches in `_binomial_dispatch` results in an infinite loop when p=0.
- normally that branch is only taken when `n * p > 10`
- but `vmap` causes jax to convert the conditional execution into a select operation.
- so, when `dist.Binomial(...).sample` is wrapped in a `vmap`, both code paths execute and we get an infinite loop.

The solution is to just wrap the branch in an additional condition check. I'll send in a PR doing this soon. 

Here's a reproducing example:
```
import jax
import numpyro.distributions as dist

def lets_break_this(p, key):
    # One of many ways of making n 
    n = (jax.random.uniform(key) > 0.5) * 3
    _, key = jax.random.split(key)
    return dist.Binomial(total_count=n, probs=p).sample(key)

print('this works fine')
jax.jit(lets_break_this)(0.1, jax.random.PRNGKey(0))
jax.jit(jax.vmap(lets_break_this, in_axes=(None, 0)))(
    0.1, jax.random.split(jax.random.PRNGKey(0), 1)
)

print('this works fine too!')
jax.jit(lets_break_this)(0, jax.random.PRNGKey(0))

print('jit with vmap and p=0 results in hanging forever')
jax.jit(jax.vmap(lets_break_this, in_axes=(None, 0)))(
    0, jax.random.split(jax.random.PRNGKey(0), 1)
)
```

PR #1462, opened 2022-08-03
Avoid infinite loop in vmapped Binomial with p=0

tbenthompson:
Closes https://github.com/pyro-ppl/numpyro/issues/1461

### DIFF
--- a/numpyro/distributions/util.py
+++ b/numpyro/distributions/util.py
@@ -110,7 +110,7 @@ def _binomial_btrs(key, p, n):
         # a HLO select that will execute both branches. This is a workaround
         # that avoids the resulting infinite loop when p=0. This should also
         # improve performance in less catastrophic cases.
-        cond_exclude_small_mu = p * n > _binomial_mu_thresh
+        cond_exclude_small_mu = p * n >= _binomial_mu_thresh
         cond_main = lax.cond(
             early_accept | early_reject,
             (),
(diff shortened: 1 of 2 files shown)


numpyro/distributions/util.py
```diff
@@ -103,13 +106,19 @@ def accept_fn(k, u, v):
         k, key, u, v = val
         early_accept = (jnp.abs(u) <= tr_params.u_r) & (v <= tr_params.v_r)
         early_reject = (k < 0) | (k > n)
-        return lax.cond(
+        # when vmapped _binomial_dispatch will convert the cond condition into
+        # a HLO select that will execute both branches. This is a workaround
+        # that avoids the resulting infinite loop when p=0. This should also
+        # improve performance in less catastrophic cases.
+        cond_exclude_small_mu = p * n > _binomial_mu_thresh
```
fehiepsi, line 113: nit: `>` -> `>=`

fehiepsi (approved):
Woohoo, thanks @tbenthompson! This is tricky to debug and do the right fix.

numpyro/distributions/util.py
tbenthompson: good catch!

tbenthompson:
Thanks! I think this is ready to merge any time. Looks like the CI is failing for unrelated reasons to do with FutureWarnings.

fehiepsi:
Thanks, @tbenthompson! Could you help me filter out the warning at [this line](https://github.com/pyro-ppl/numpyro/blob/master/setup.cfg#L22)?

martinjankowiak:
@tbenthompson  you probably want to merge in master after #1464

fehiepsi (approved):

PR #1475, opened 2022-09-05
Draft of switch `StudentT` `cdf` to use tfp's `betainc`

colehaus:
Jax's `betainc` doesn't have gradients defined for all parameters while tfp's does.

See the related PR here: https://github.com/pyro-ppl/numpyro/pull/1471 and the initial discussion here: https://github.com/pyro-ppl/numpyro/issues/1452.

I'm not sure exactly how you want to handle the dependency declarations since `tensorflow` and `tensorflow-probability` are sort of heavy dependencies to bring in (i.e. should they be promoted to `install_requires`?).

Also, the type casting stuff is a bit ugly but tfp checks that array types match and `self.df` sometimes had a `float64` dtype in tests while `beta_value` has a `float32` dtype in each test.

### DIFF
--- a/numpyro/distributions/continuous.py
+++ b/numpyro/distributions/continuous.py
@@ -27,7 +27,6 @@
 
 import numpy as np
 
-import jax
 from jax import lax
 from jax.experimental.sparse import BCOO
 import jax.nn as nn
--- a/numpyro/distributions/continuous.py
+++ b/numpyro/distributions/continuous.py
@@ -1910,19 +1910,16 @@ class StudentT(Distribution):
         return jnp.broadcast_to(var, self.batch_shape)
 
     def cdf(self, value):
-        from tensorflow_probability.substrates.jax.math import betainc as tfp_betainc
+        try:
+            from tensorflow_probability.substrates.jax.math import betainc as betainc_fn
+        except ImportError:
+            from jax.scipy.special import betainc as betainc_fn
 
         # Ref: https://en.wikipedia.org/wiki/Student's_t-distribution#Related_distributions
         # X^2 ~ F(1, df) -> df / (df + X^2) ~ Beta(df/2, 0.5)
         scaled = (value - self.loc) / self.scale
         scaled_squared = scaled * scaled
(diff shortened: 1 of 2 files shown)


numpyro/distributions/continuous.py
```diff
@@ -1909,18 +1910,40 @@ def variance(self):
         return jnp.broadcast_to(var, self.batch_shape)
 
     def cdf(self, value):
+        from tensorflow_probability.substrates.jax.math import betainc as tfp_betainc
```
fehiepsi, line 1913: I think you can add a try/except logic here
```
try:
    from ...
    betainc_fn = tfp_betainc
else:
    betainc_fn = betainc_fn
```

numpyro/distributions/continuous.py
```diff
@@ -1909,18 +1910,40 @@ def variance(self):
...
         # X^2 ~ F(1, df) -> df / (df + X^2) ~ Beta(df/2, 0.5)
         scaled = (value - self.loc) / self.scale
         scaled_squared = scaled * scaled
         beta_value = self.df / (self.df + scaled_squared)
+
+        float_type = (
+            jnp.promote_types(self.df.dtype, beta_value.dtype)
+            if jax.config.read("jax_enable_x64")
+            else jnp.dtype("float32")
+        )
```
fehiepsi, line 1925: This is unnecessary, you can do something like
```
betainc_fn(0.5 * jnp.asarray(self.df), 0.5, jnp.asarray(beta_value))
```

numpyro/distributions/continuous.py
```diff
@@ -27,6 +27,7 @@
 
 import numpy as np
 
+import jax
```
fehiepsi, line 30: This is causing lint error.

colehaus:
Ah, sorry. Was slightly non-trivial to run the lint checks with a .venv, but I cherry-picked my config changes from the other PR and ran `make lint` locally so it should pass this time.

fehiepsi:
This is a great addition! Thanks, @colehaus.

fehiepsi (approved):

PR #1478, opened 2022-09-12
Add icdf for Beta, Gamma, and StudentT

### LINKED ISSUE #1365: icdf of beta distribution is not implemented
```python
>>> import numpyro.distributions as dist
>>> d = dist.Beta(1, 5)
>>> d.icdf(0.975)
  File "<stdin>", line 1, in <module>
  File "/home/ayaka/venv/lib/python3.10/site-packages/numpyro/distributions/distribution.py", line 463, in icdf
    raise NotImplementedError
NotImplementedError
```

NumPyro version: 0.9.1

fehiepsi:
Fixes #1365

### DIFF
--- a/numpyro/distributions/util.py
+++ b/numpyro/distributions/util.py
@@ -414,6 +414,33 @@ def clamp_probs(probs):
     return jnp.clip(probs, a_min=finfo.tiny, a_max=1.0 - finfo.eps)
 
 
+def betainc(a, b, x):
+    try:
+        from tensorflow_probability.substrates.jax.math import betainc as betainc_fn
+    except ImportError:
+        from jax.scipy.special import betainc as betainc_fn
+
+    return betainc_fn(a, b, x)
+
+
+def betaincinv(a, b, y):
+    try:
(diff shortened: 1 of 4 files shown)


numpyro/distributions/util.py
```diff
@@ -414,6 +414,37 @@ def clamp_probs(probs):
     return jnp.clip(probs, a_min=finfo.tiny, a_max=1.0 - finfo.eps)
 
 
+def betainc(a, b, x):
+    try:
+        from tensorflow_probability.substrates.jax.math import betainc as betainc_fn
+    except ImportError:
```
martinjankowiak, line 420: do you want to include an error message here as below?

martinjankowiak (commented):
are all the cdfs and icdfs here compared against scipy in the tests?

numpyro/distributions/util.py
fehiepsi: Here we use the default jax implementation (which does not have grad over all params)

fehiepsi:
Yes, those cdf/icdf are compared against scipy in the tests

martinjankowiak (approved):

Issue #1178, opened 2021-10-07
Support MatrixNormal distribution


quattro:
Hi all, the [MatrixNormal](https://en.wikipedia.org/wiki/Matrix_normal_distribution) provides a clean way to parameterize the distribution of `n x k` matrices `X ~ MN_{nk}(M, U, V)` as a function of its `n x k` mean `M`, `n x n` row-wise covariance `U` and `k x k` column-wise covariance `V`. 

This can be equivalently modeled as a classical multi-variate normal by applying the `vec` operation of `X` and kronecker product between `U` and `V` to get `vec(X) ~ N(vec(M), V kron U)`. However directly evaluating this likelihood ignores the computational advantages offered by the `MatrixNormal` formulation (computing inverse of `V kron U` requires O((nk)^3) compared with the O(n^3 + k^3) formulation). 

It would be great to see this implemented in numpyro to exploit its variance/covariance structure for efficient likelihood calculation. There are some corner cases that devolve into batched (or transposed batched) MVNs (ie when either U or V is identity), but in its full generality it could be useful (ie certain factor analysis methods, or multi-variate outcomes models).

Implementation should be straightforward (I've got a non-batched version worked out), but could see batching complicate implementation details.

kaijennissen:
Any progress on this?

kaijennissen:
I was able to implement a basic example.

**Sampling:**
I've used the fact that if `X` is a `(n x p)` matrix with `x_(ij) ~ N(0,1)` and `Y=loc + U @ X @ V'`  then  `Y ~ MN(loc, A@A',B@B')`.
Assume we have `sample_shape=(s,)`, `batch_shape=()` . Then we can sample X form a Normal distribution with `sample_shape = (s,)+event_shape` where `event_shape=(n,p)`.
We then have to map to matrix multiplication along the (single) first axis which can be achieved using `vmap`, e.g.

```python
X = dist.Normal(0,1).sample(rng, sample_shape=(s,)+event_shape)
Y = loc + jax.vmap(lambda x: U@x@V)(X)
```

I was able to come up with an extension for `sample_shape=(s,)` and `batch_shape=(b,)`by first using `vmap` to map the matrix multiplication along the (single) batch axis and then map this along the (single) sampling axis.
To my surprise this also worked for `batch_shape=(b1,b1)` but I cannot explain why.
Any suggestions on how I could implement arbitrary batch and sample shapes without chaining `vmap` for each of the dimensions?

[...]

PR #1479, opened 2022-09-13
Matrix Normal distribution

kaijennissen:
First draft for the MatrixNormal distribution as requested in #1178. 
Currently `sample` as well as `log_prob` use the close relationship between Matrix Normal and Multivariate Normal as described [here](https://en.wikipedia.org/wiki/Matrix_normal_distribution).


fehiepsi:
@kaijennissen Sorry for the slow response. I thought this is in the draft phase. I'll try to review this week.

numpyro/distributions/continuous.py
```diff
@@ -1033,6 +1033,115 @@ def icdf(self, q):
...
+        ), "cov shape does not match"
+
+        mvn = MultivariateNormal(loc=loc, covariance_matrix=cov)
+
+        assert mvn.batch_shape == self.batch_shape, "batch shape of MVN does not match"
+        assert mvn.event_shape == (
+            np.prod(self.event_shape),
+        ), "event shape of MVN does not match"
+
+        samples_mvn = mvn.sample(key, sample_shape=sample_shape)
```
fehiepsi, line 1113: This seems expensive. Could we simply use the procedure in [this section](https://en.wikipedia.org/wiki/Matrix_normal_distribution#Drawing_values_from_the_distribution)?
```
eps = random.normal(
       key, shape=sample_shape + self.batch_shape + self.event_shape)
return self.loc + self.scale_tril_row @ eps @ jnp.swapaxes(self.scale_tril_column, -2, -1)
```

numpyro/distributions/continuous.py
```diff
@@ -1033,6 +1033,115 @@ def icdf(self, q):
...
+class MatrixNormal(Distribution):
+    arg_constraints = {
+        "loc": constraints.real_vector,
+        "scale_columns": constraints.positive_definite,
+        "scale_rows": constraints.positive_definite,
+    }
+    support = constraints.real_matrix
+    reparametrized_params = ["loc", "scale_columns", "scale_rows"]
+
+    def __init__(self, loc, scale_columns, scale_rows, validate_args=None):
```
fehiepsi, line 1052: I think it is better to use `scale_tril_row`, `scale_tril_column`, which has `constraints.lower_cholesky` constraint.

numpyro/distributions/continuous.py
```diff
@@ -1033,6 +1033,115 @@ def icdf(self, q):
...
+            p,
+        ), "scale_rows.shape does not match"
+
+        super(MatrixNormal, self).__init__(
+            batch_shape=batch_shape, event_shape=event_shape
+        )
+
+        self.loc = loc
+        self.scale_columns = scale_columns
+        self.scale_rows = scale_rows
```
fehiepsi, line 1088: I think you can use
```
batch_shape = lax.broadcast_shapes(
    jnp.shape(loc)[:-2], jnp.shape(scale_tril_row)[:-2], jnp.shape(scale_tril_column)[:-2])
self.loc, self.scale_tril_row, self.scale_tril_column = _promote_shapes(loc, scale_tril_row, scale_tril_column, shape=batch_shape)
```

numpyro/distributions/continuous.py
```diff
@@ -1033,6 +1033,115 @@ def icdf(self, q):
...
+
+    @property
+    def mean(self):
+        return self.loc
```
fehiepsi, line 1092: We need broadcasting [like this](https://github.com/pyro-ppl/numpyro/blob/6fdee9d126d9857f071318614988ccccf09bac20/numpyro/distributions/continuous.py#L1273)

numpyro/distributions/continuous.py
```diff
@@ -1033,6 +1033,115 @@ def icdf(self, q):
...
+        assert cov.shape == self.batch_shape + (
+            mvn_dim,
+            mvn_dim,
+        ), "cov shape does not match"
+
+        mvn = MultivariateNormal(loc=loc, covariance_matrix=cov)
+        values = jnp.reshape(
+            self.loc, newshape=self.batch_shape + (mvn_dim,), order="F"
+        )
+        log_prob_mvn = mvn.log_prob(values)
```
fehiepsi, line 1141: This seems to be expensive: creating a matrix (loc.shape[-1] * loc.shape[-2], loc.shape[-1] * loc.shape[-2]). I think we can use the formula in [wikipedia](https://en.wikipedia.org/wiki/Matrix_normal_distribution#Definition)
```
row_log_det = jnp.log(jnp.diagonal(self.scale_tril_row, axis1=-2, axis2=-1)).sum(-1)
col_log_det = jnp.log(jnp.diagonal(self.scale_tril_column, axis1=-2, axis2=-1)).sum(-1)
# compute the trace term
diff = vale - self.loc
diff_row_solve = solve_triangular(self.scale_tril_row, diff, lower=True)
diff_col_solve = solve_triangular(self.scale_tril_column, jnp.swapaxes(diff_row_solve, -2, -1), lower=True)
trace_term = jnp.square(diff_col_solve.reshape(diff_col_solve.shape[:-2] + (-1))).sum(-1)
```

numpyro/distributions/continuous.py
kaijennissen: Doesn't  `_promote_shapes` take care of this?

numpyro/distributions/continuous.py
kaijennissen: Of course. I didn't take care of the case that `loc` and `scale_tril_column`/`scale_tril_row`do not match.

numpyro/distributions/continuous.py
kaijennissen: It is. Your suggestion is much better. But I do not fully understand the calculation of the`trace_term`.
Using the notation from Wikipedia and `D=(X-M)` the `trace_term` is given by `trace_term=tr(inv(V)@D.T@inv(U)@D)`. Since we us a lower triangular parametrisation`ìnv(U)@D=inv(scale_row_tril@scale_row_tril.T)@D`=inv(scale_row_tril.T)@inv(scale_row_tril)@D=solve_triangular(lower_row_tril,solve_triangular(scale_row_tril,D,lower=True),lower=False)`

numpyro/distributions/continuous.py
kaijennissen: Indeed. This is much simpler.

numpyro/distributions/continuous.py
kaijennissen: That doesn't work. 
```
n=2
p=3
loc = np.random.random(n * p).reshape(n, p)
scale_tril_row=np.random.random((n, n))
scale_tril_row*=np.tri(*scale_tril_row.shape)
scale_tril_column = np.random.random((p, p))
scale_tril_column  *=np.tri(*scale_tril_column.shape)

batch_shape = jax.lax.broadcast_shapes(
    jnp.shape(loc)[:-2],
    jnp.shape(scale_tril_row)[:-2],
    jnp.shape(scale_tril_column)[:-2],
)

promote_shapes(
    loc, scale_tril_row, scale_tril_column, shape=batch_shape
)
```

numpyro/distributions/continuous.py
fehiepsi: Promoting shapes only adds singleton dimensions to the input.

numpyro/distributions/continuous.py
fehiepsi: We can rewrite `tr(inv(V)@D.T@inv(U)@D)` into
```
tr(inv(V_tril).T@inv(V_tril)@D.T@inv(U_tril).T@inv(U_tril)@D)
= tr(inv(V_tril)@D.T@inv(U_tril).T @ inv(U_tril)@D@inv(V_tril).T)
= tr(X.T @ X)
where X = inv(U_tril) @ (inv(V_tril) @ D.T).T
```

kaijennissen: I've only looked at the term inside the trace but the key is the cyclic property of the trace (`tr(A@B@C)=tr(C@A@B)`) right?

fehiepsi: That's right. That helps us factorize the trace term into the form X @ X.T, so that we can compute trace easily.

numpyro/distributions/continuous.py
fehiepsi: Sorry, I think we need to promote shapes individually. The total shapes of those matrices are not broadcastable. Something like
```
self.loc, = promote_shapes(loc, shape=batch_shape + loc.shape[-2:])
```

numpyro/distributions/continuous.py
kaijennissen: The current implementation works for single samples. With `vmap` I also was able to e broadcasted to `1-D`samples. For other cases I need a little more time.

kaijennissen: @fehiepsi I've added a draft for a `log_prob`method which should deal with arbitrary sample and batch shapes.

numpyro/distributions/continuous.py
```diff
@@ -1075,6 +1075,145 @@ def cdf(self, x):
         return self.base_dist.cdf(jnp.log(x))
 
 
+def _batch_solve_triangular(A, B, sample_shape, batch_shape, event_shape):
+    """Extende solve_triangular for the case that B.ndim > A.ndim.
+        This is achived by first flattening the leadind B.ndim - A.ndim axes of B and
+        then moving them to the end.
```
fehiepsi, line 1081: This is an excellent approach!

numpyro/distributions/continuous.py
```diff
@@ -1075,6 +1075,145 @@ def cdf(self, x):
...
+        B (_type_): _description_
+        sample_shape (_type_): _description_
+        batch_shape (_type_): _description_
+        event_shape (_type_): _description_
+
+    Returns:
+        _type_: _description_
+    """
+    n, p = event_shape
+    B_flat = jnp.moveaxis(B.reshape((-1,) + batch_shape + event_shape), 0, -2).reshape(
```
fehiepsi, line 1094: Can we broadcast `B` to `sample_shape + batch_shape + event_shape` before performing `reshape`? Most of the time it will be a no-op but might be convenient if we have a batched distribution but only wants to evaluate log_prob over a non-batched sample.

numpyro/distributions/continuous.py
```diff
@@ -1075,6 +1075,145 @@ def cdf(self, x):
...
+        batch_shape + (n,) + (-1,)
+    )
+
+    X_flat = solve_triangular(A, B_flat, lower=True)
```
fehiepsi, line 1098: Can we broadcast `A` to `batch_shape + A.shape[-2:]`?

numpyro/distributions/continuous.py
```diff
@@ -1075,6 +1075,145 @@ def cdf(self, x):
...
+    """Extende solve_triangular for the case that B.ndim > A.ndim.
+        This is achived by first flattening the leadind B.ndim - A.ndim axes of B and
+        then moving them to the end.
+
+    Args:
+        A (_type_): A triangular array.
+        B (_type_): _description_
+        sample_shape (_type_): _description_
+        batch_shape (_type_): _description_
+        event_shape (_type_): _description_
```
fehiepsi, line 1088: I think you can revise the logic into
```python
event_shape = B.shape[-2:]
batch_shape = lax.broadcast_shapes(A.shape[:-2], B.shape[-A.ndim:-2])
sample_shape = B.shape[:-A.ndim]
```

fehiepsi (commented):
Looks great to me!! Could you add this distribution to [src file](https://github.com/pyro-ppl/numpyro/blob/master/docs/source/distributions.rst)?

numpyro/distributions/continuous.py
```diff
@@ -1075,6 +1075,132 @@ def cdf(self, x):
...
 
 
+def _batch_solve_triangular(A, B):
+    """Extende solve_triangular for the case that B.ndim > A.ndim.
+        This is achived by first flattening the leadind B.ndim - A.ndim axes of B and
+        then moving them to the end.
+
+    Args:
+        A (_type_): A triangular array.
+        B (_type_): _description_
```
fehiepsi, line 1085: Could you update docstring here and the MatrixNormal distribution? You can use `jnp.ndarray` for `_type_`. It would be nice to use
```
:param: ...
:return: ...
```
for consistency (as in [here](https://github.com/pyro-ppl/numpyro/blob/7f7e71077014813e06eb214728b7d6feebcb7f67/numpyro/distributions/mixtures.py#L27-L35))

fehiepsi (approved):

numpyro/distributions/continuous.py
kaijennissen: Like this? 
```
:param jnp.ndarray (...,M,M) A: An array with lower triangular structure in the last two dimensions.
:param jnp.ndarray (...,M,) or (...,M,N) B: Right-hand side matrix in A x = B.
```

numpyro/distributions/continuous.py
```diff
@@ -1076,18 +1076,17 @@ def cdf(self, x):
...
 
-    Args:
-        A (_type_): A triangular array.
-        B (_type_): _description_
 
-    Returns:
-        _type_: _description_
-    """
+    :param jnp.ndarray (...,M,M) A: An array with lower triangular structure in the last two dimensions.
+    :param jnp.ndarray (...,M,) or (...,M,N) B: Right-hand side matrix in A x = B.
```
fehiepsi, line 1086: nit: we only support (...,M,N) shape for B.

fehiepsi (approved):
Thanks, @kaijennissen!

numpyro/distributions/continuous.py
fehiepsi: Yup

test/test_distributions.py
```diff
@@ -391,6 +399,24 @@ def get_sp_dist(jax_dist):
     T(dist.LogUniform, 1.0, 2.0),
     T(dist.LogUniform, 1.0, np.array([2.0, 3.0])),
     T(dist.LogUniform, np.array([1.0, 2.0]), np.array([[3.0], [4.0]])),
+    T(
+        dist.MatrixNormal,
+        1.0 * np.arange(6).reshape(3, 2),
```
fehiepsi, line 404: Tests are failing at the `sample` method. Probably the event shape is `(2, 3)`, not (3, 2)?

test/test_distributions.py
```diff
@@ -391,6 +399,24 @@ def get_sp_dist(jax_dist):
...
+        dist.MatrixNormal,
+        1.0 * np.arange(12).reshape((2, 3, 2)),
+        np.array([[1.0, 0], [0.4, 0.5]]) * np.ones((2, 2, 2)),
+        np.array([[1.0, 0, 0], [0.3, 0.36, 0], [0.4, 0.49, 4]]) * np.ones((2, 3, 3)),
+    ),
+    T(
+        dist.MatrixNormal,
+        1.0 * np.arange(36).reshape((2, 3, 3, 2)),
+        np.broadcast_to(np.identity(2), (2, 3, 2, 2)),
+        np.broadcast_to(np.identity(3), (2, 3, 3, 3)),
```
fehiepsi, line 417: Maybe we can remove this broadcast operator to check if our batching logic is working?

test/test_distributions.py
```diff
@@ -391,6 +399,24 @@ def get_sp_dist(jax_dist):
...
     T(dist.LogUniform, np.array([1.0, 2.0]), np.array([[3.0], [4.0]])),
+    T(
+        dist.MatrixNormal,
+        1.0 * np.arange(6).reshape(3, 2),
+        np.array([[1.0, 0], [0.4, 1]]),
+        np.array([[1.0, 0, 0], [0.3, 0.36, 0], [0.4, 0.49, 4]]),
+    ),
+    T(
+        dist.MatrixNormal,
+        1.0 * np.arange(12).reshape((2, 3, 2)),
```
fehiepsi, line 410: Maybe we use `np.arange(6.).reshape((2, 3))` to check if broadcasting logic for `loc` is working?

kaijennissen: This is on me. I changed the order of `scale_tril_row` and `scale_tril_column` in `__init__` to match Wikipedia but forgot to changes this in `test_distribution.py`.

test/test_distributions.py
kaijennissen: Kepp `loc`as is and `np.identity(2)`, `np.identity(3)`?

kaijennissen: That does work but breaks the test since I'm looping over the batch dimension which do not match.
`samples.shape=sample_shape+batch_shape+event_shape` while `jax_dist.scale_tril_row=(1,1)+event_shape`.

kaijennissen: Found a solution.

fehiepsi (approved):

PR #1480, opened 2022-09-15
Fix minor issue in example bayesian_hierarchical_stacking.ipynb

cpieringer:
Fix:
- Change the name `exp_lpd_point` into `lpd_point` in the definition and body of the function `stack`.
- Change the name of the argument in `mcmc.run( ... lpd_point=lpd_point`


fehiepsi:
Thanks for the fix, @cpieringer! Could you install the latest `black` release and run `black .` to format the notebook? CI is failing because of that.

fehiepsi (approved):

PR #1483, opened 2022-09-28
Update diagnostics.py

ayhteo:
In `numpyro.diagnostics` , the `effective_sample_size` could throw a jax immutable array error if the user uses a jax array as input. Updating this with a conversion from jax array to np array using `jax.device_get` could remove this error. Other functions in diagnostics do not have the same problem as they do not use item assignment.


fehiepsi:
Thanks @ayhteo!

fehiepsi (approved):

PR #1495, opened 2022-10-30
Feature log uniform

yayami3:
I want to take over this PR #1423 

What I did is

- make parameter low and high as non default
- add test


numpyro/distributions/continuous.py
```diff
@@ -1033,6 +1033,40 @@ def icdf(self, q):
         return self.loc + self.scale * logit(q)
 
 
+class LogUniform(TransformedDistribution):
```
fehiepsi, line 1036: I think you can rename this to [Reciprocal](https://en.wikipedia.org/wiki/Reciprocal_distribution) if the support is `[low, high]` (rather than `[exp(low), exp(high)]` as in `LogUniform`).

fehiepsi:
cc @andrewfowlie

numpyro/distributions/continuous.py
yayami3: My understanding is that Reciprocal and [log-uniform](https://docs.scipy.org/doc/scipy/tutorial/stats/continuous_loguniform.html) are identical. Isn't it ?

fehiepsi: I'm not sure. In LogNormal, loc and scale are parameters of the base distribution. So probably using LocUniform is confusing. What do you think?

yayami3: I also can find implemantation in [Matlab](https://www.mathworks.com/help/stats/loguniform-distribution.html) define support as `[low, high]` 
I think it's better to follow these.

andrewfowlie:
I'm very busy right now, so happy for someone else to take over this.  A few comments:

- personally, I'd much prefer a name like LogUniform to Reciprocal. The name Reciprocal tells us about the density for x being \propto 1 / x; the name LogUniform tells us that this distribution is uniform in log x. The latter is much more relevant
- I'd much prefer if the low and high parameters were not logged. E.g., usage was

      LogUniform(1e-3, 1e6)

and not

      LogUniform(Log(1e-3), Log(1e-6))

I think the latter is more verbose and confusing for users. E.g., what base of log? A user might try to do

      LogUniform(-3, 6)

and be caught out if we are using natural log rather than base 10. Another advantage is if low and high are not logged, it is trivial to switch between Uniform and LogUniform


      Uniform(1e-3, 1e6)
      LogUniform(1e-3, 1e6)

fehiepsi:
Thanks @yayami3 and @andrewfowlie! The explanation is really helpful!!

@yayami3 could you add an entry for this distribution in the docs https://github.com/pyro-ppl/numpyro/blob/master/docs/source/distributions.rst

yayami3:
Thanks for your help! @fehiepsi  and @andrewfowlie 

Added docs.

fehiepsi (approved):
This is great!! Thank you both for the contribution!

PR #1506, opened 2022-12-01
Add MultivariateBeta distribution.

hessammehr:
This new distribution has beta marginals on each axis but incorporates covariance through a multivariate Gaussian copula as explained [here](https://stats.stackexchange.com/a/549262). I found this distribution useful for modelling correlated point clouds within a fixed interval.

A small example:
```python
mv_normal = dist.MultivariateNormal(covariance_matrix=jnp.array([[1.0, 0.75], [0.75, 1.0]]))
conc1 = jnp.array([2.0, 3.5])
conc0 = jnp.array([4.0, 10.0])

mv_beta = MultivariateBeta(mv_normal, conc1, conc0)
s = mv_beta.sample(sample_shape=(1000,), key=key)
probs = np.exp(mv_beta.log_prob(s))
sns.pairplot(pd.DataFrame(s), plot_kws=dict(s=5, alpha=probs / probs.max()))
```

image


fehiepsi:
This is a super cool distribution! I have several high level comments on this:
+ Using mv_normal seems to be unnecessary. I guess we can use an argument `corr_cholesky` here instead (we have LKJCholesky distribution that can be used to do inference over this correlation matrix).
+ As mentioned in the reference, we have use this construction for other univariate distribution. How about we create a class
```python
class GaussianCopula(Distribution):
    # not sure what would be a better name for transform_dist
    # we allow users to provide either corr_cholesky or the corr_matrix (as in MultivariateNormal distribution)
    def __init__(self, transform_dist, correlation_cholesky=None, correlation_matrix=None):
        self.base_dist = dist.MultivariateNormal(0, scale_tril=correlation_cholesky, covariance_matrix=correlation_matrix)
        self.transform_dist = transform_dist
        ...
```
where we can provide `transform_dist = Beta(concentration0, concentation1)`

hessammehr:
Hi @fehiepsi, thanks for the comments. I agree that passing `scale_tril` and `correlation_matrix` is cleaner and avoids the possibility of passing an off-center (`loc != 0`) multivariate normal. I'll have a go at implementing the general case for any univariate distribution.

numpyro/distributions/continuous.py
```diff
@@ -1680,38 +1680,46 @@ def infer_shapes(df, loc, scale_tril):
...
+class GaussianCopula(Distribution):
     arg_constraints = {
-        "concentration1": constraints.positive,
-        "concentration0": constraints.positive,
+        "covariance_matrix": constraints.positive_definite,
+        "precision_matrix": constraints.positive_definite,
+        "scale_tril": constraints.lower_cholesky,
     }
-    reparametrized_params = ["concentration1", "concentration0"]
+    support = constraints.real_vector
```
fehiepsi, line 1689: You can use
```python
@property
def support(self):
    return constraints.independent(self.transform_dist.support, 1)
```

numpyro/distributions/continuous.py
```diff
@@ -1680,38 +1680,46 @@ def infer_shapes(df, loc, scale_tril):
...
+    reparametrized_params = [
+        "covariance_matrix",
+        "precision_matrix",
+        "scale_tril",
```
fehiepsi, line 1693: I think we only need to use correlation matrix to account for the correlation of the variables here (so only `correlation_matrix`/`correlation_cholesky` are needed). Otherwise, we might face identification issue where multiple `covariance_matrix` parameters give the same distribution.

numpyro/distributions/continuous.py
```diff
@@ -1680,38 +1680,46 @@ def infer_shapes(df, loc, scale_tril):
...
+            covariance_matrix=covariance_matrix,
+            precision_matrix=precision_matrix,
+            scale_tril=scale_tril,
         )
+        self.normal = Normal(scale=jnp.diag(self.base_dist.covariance_matrix))
 
-        self.beta = Beta(self.concentration1, self.concentration0)
-        self.normal = Normal(scale=jnp.diag(mv_normal.covariance_matrix))
-        self.mv_normal = mv_normal
+        event_shape = self.base_dist.event_shape
```
fehiepsi, line 1714: I think we can add an assertion that `len(self.base_dist.event_shape) == 0` and set `event_shape = self.base_dist.batch_shape[-1:]`.

numpyro/distributions/continuous.py
```diff
@@ -1680,38 +1680,46 @@ def infer_shapes(df, loc, scale_tril):
...
+        delta_dim = len(event_shape) - len(transform_dist.event_shape)
 
         batch_shape = lax.broadcast_shapes(
-            jnp.shape(concentration1),
-            jnp.shape(concentration0),
-            mv_normal.batch_shape,
+            self.transform_dist.batch_shape[:-delta_dim],
+            self.base_dist.batch_shape,
```
fehiepsi, line 1719: I guess batch shape is `lax.broadcast_shapes(transform_dist.batch_shape, base_dist.batch_shape[:-1])` if `len(base_dist.batch_shape) >= 1` else `transform_dist.batch_shape` (so that we can specify `Beta(1, 1), correlation_matrix=jnp.eye(3)` as arguments)

numpyro/distributions/continuous.py
```diff
@@ -1720,20 +1728,20 @@ def __init__(
     def sample(self, key, sample_shape=()):
         assert is_prng_key(key)
 
-        shape = sample_shape + self.batch_shape[:-len(self.mv_normal.batch_shape)]
-        normal_samples = self.mv_normal.sample(sample_shape=shape, key=key)
+        shape = sample_shape + self.batch_shape[: -len(self.base_dist.batch_shape)]
+        normal_samples = self.base_dist.sample(sample_shape=shape, key=key)
```
fehiepsi, line 1732: I think you can do `self.base_dist.expand(sample_shape + self.batch_shape).sample(key)` here

numpyro/distributions/continuous.py
```diff
@@ -1680,38 +1680,46 @@ def infer_shapes(df, loc, scale_tril):
...
     ):
-        self.concentration1, self.concentration0 = promote_shapes(
-            concentration1, concentration0
+        self.transform_dist = transform_dist
+        self.base_dist = MultivariateNormal(
+            covariance_matrix=covariance_matrix,
+            precision_matrix=precision_matrix,
+            scale_tril=scale_tril,
         )
+        self.normal = Normal(scale=jnp.diag(self.base_dist.covariance_matrix))
```
fehiepsi, line 1712: `jnp.diag` only works for 1,2 dimensional arrays. I think you want `diagonal(..., -1, -2)` instead.

numpyro/distributions/continuous.py
```diff
@@ -1680,6 +1680,70 @@ def infer_shapes(df, loc, scale_tril):
         return batch_shape, event_shape
 
 
+class GaussianCopula(Distribution):
```
fehiepsi, line 1683: Maybe we can move this generalized distribution into a `copula.py` file, subclass this with
```python
class GaussianCopulaBeta(GaussianCopula):
    arg_constraints = {...}
    def __init__(self, concentration0, concentration1, correlation...):
        super().__init__(Beta(concentration0, concentration1), correlation...)
        self.concentration0, self.concentration1 = promote_shapes(concentration0, concentration1, shape=self.batch_shape + self.event_shape)
```
and use `GaussianCopulaBeta` for testing?

numpyro/distributions/continuous.py
```diff
@@ -1680,6 +1680,70 @@ def infer_shapes(df, loc, scale_tril):
...
+            event_shape=event_shape,
+            validate_args=validate_args,
+        )
+
+    def sample(self, key, sample_shape=()):
+        assert is_prng_key(key)
+
+        shape = sample_shape + self.batch_shape[: -len(self.base_dist.batch_shape)]
+        normal_samples = self.base_dist.sample(sample_shape=shape, key=key)
+        cdf = self.normal.cdf(normal_samples)
```
fehiepsi, line 1733: Probably I misunderstand this - I guess we only need `cdf = dist.Normal(0, 1).cdf(normal_samples)` here?

fehiepsi: Oh, I see the reason. You are using covariance matrix, not correlation matrix. As mentioned in another comment, it's better to use correlation matrix to help inference over this parameter.

numpyro/distributions/continuous.py
```diff
@@ -1680,6 +1680,70 @@ def infer_shapes(df, loc, scale_tril):
...
+        "covariance_matrix",
+        "precision_matrix",
+        "scale_tril",
+    ]
+
+    support = constraints.unit_interval
+
+    def __init__(
+        self,
+        transform_dist,
```
fehiepsi, line 1700: Looking at your implementation, I guess we can call this `marginal_dist`?

numpyro/distributions/continuous.py
```diff
@@ -1680,6 +1680,70 @@ def infer_shapes(df, loc, scale_tril):
...
+        shape = sample_shape + self.batch_shape[: -len(self.base_dist.batch_shape)]
+        normal_samples = self.base_dist.sample(sample_shape=shape, key=key)
+        cdf = self.normal.cdf(normal_samples)
+        return self.transform_dist.icdf(cdf)
+
+    @validate_sample
+    def log_prob(self, value):
+        marginal_lps = self.transform_dist.log_prob(value)
+        quantiles = self.normal.icdf(value)
+        return self.base_dist.log_prob(quantiles) + marginal_lps.sum(axis=-1)
```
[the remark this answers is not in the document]
fehiepsi: Looking at [Sklar's theorem](https://en.wikipedia.org/wiki/Copula_(probability_theory)#Sklar's_theorem), it seems to me that we need to return `copula_lp + marginal_lps.sum(axis=-1)` where the (Gaussian) copula density can be found [here](https://en.wikipedia.org/wiki/Copula_(probability_theory)#Gaussian_copula):
```
copula_lp = self.base_dist.log_prob(quantiles) + 0.5 * (quantiles ** 2).sum(-1) + 0.5 * jnp.log(2 * jnp.pi) * quantiles.shape[-1]
```

numpyro/distributions/continuous.py
hessammehr: Good idea, I'll rename it to that.

numpyro/distributions/continuous.py
hessammehr: I see the following decorator used with `support` used elsewhere. Would you recommend using it? I suppose we don't necessarily know the event dim?

```python
@constraints.dependent_property(is_discrete=False, event_dim=0)
def support(self):
    ...
```

numpyro/distributions/continuous.py
hessammehr: Just to make sure I understand, is the suggestion to keep `scale_tril`? I'm not quite sure how to keep the correlations only (since the individual variances are not important, as you mentioned)

numpyro/distributions/continuous.py
hessammehr: Surely, `len(self.base_dist.event_shape) == 1`, no?

numpyro/distributions/continuous.py
fritzo: Wouldn't the support be static `constraints.independent(constraints.unit_interval, 1)`?  Sorry if I misunderstand, but if `Beta.support == constraints.unit_interval`, I'd expect this distribution to have static support.

> I suppose we don't necessarily know the event dim?

IIUC we know `event_dim == 1`, so `event_shape == (n,)` for some `n` we don't know, but constraints don't need to know that `n`.

Sorry if I'm misunderstanding something!

hessammehr: @fritzo Thanks for the comment. You are right `event_dim` is known even though `event_shape` is not.

hessammehr:
@fehiepsi That should be all of the fixes/suggestions in our discussion above implemented, hopefully. I have a few simple test cases implemented in a local Jupyter notebook already, which I will aim to incorporate into the test suite soon.

Thanks for the feedback and tips. I found them extremely useful. 🙏

numpyro/distributions/copula.py
```diff
@@ -0,0 +1,74 @@
...
+        shape = sample_shape + self.batch_shape
+        normal_samples = self.base_dist.expand(shape).sample(key)
+        cdf = self.normal.cdf(normal_samples)
+        return self.marginal_dist.icdf(cdf)
+
+    @validate_sample
+    def log_prob(self, value):
+        marginal_lps = self.marginal_dist.log_prob(value)
+        quantiles = self.normal.icdf(value)
+        #
```
fehiepsi, line 60: could you comment here that the gaussian copula log prob follows this reference
```
# Ref: https://en.wikipedia.org/wiki/Copula_(probability_theory)#Gaussian_copula
```

numpyro/distributions/copula.py
```diff
@@ -0,0 +1,74 @@
...
+        copula_lp = (
+            self.base_dist.log_prob(quantiles)
+            + 0.5 * (quantiles**2).sum(-1)
+            + 0.5 * jnp.log(2 * jnp.pi) * quantiles.shape[-1]
+        )
+        return copula_lp + marginal_lps.sum(axis=-1)
```
fehiepsi, line 66: It's useful to comment the ref here.
```
# Ref: https://en.wikipedia.org/wiki/Copula_(probability_theory)#Sklar's_theorem
```

numpyro/distributions/copula.py
```diff
@@ -0,0 +1,74 @@
...
+
+    @property
+    def mean(self):
+        return jnp.broadcast_to(self.marginal_dist.mean, self.shape())
+
+    @constraints.dependent_property(event_dim=1)
```
fehiepsi, line 72: we can also set `is_discrete=False` here

numpyro/distributions/copula.py
```diff
@@ -0,0 +1,74 @@
...
+        "correlation_matrix",
+        "correlation_cholesky",
+    ]
+
+    def __init__(
+        self,
+        marginal_dist,
+        correlation_matrix=None,
+        correlation_cholesky=None,
+        validate_args=None,
```
fehiepsi, line 27: nit: we need `*,` before validate_args like other distributions

numpyro/distributions/copula.py
```diff
@@ -0,0 +1,74 @@
...
+import numpyro.distributions.constraints as constraints
+from numpyro.distributions.continuous import MultivariateNormal, Normal
+from numpyro.distributions.distribution import Distribution
+from numpyro.distributions.util import is_prng_key, validate_sample
+
+
+class GaussianCopula(Distribution):
+    arg_constraints = {
+        "correlation_matrix": constraints.corr_matrix,
+        "correlation_cholesky": constraints.lower_cholesky,
```
fehiepsi, line 15: this should be `constraints.corr_cholesky`

numpyro/distributions/copula.py
```diff
@@ -0,0 +1,74 @@
...
+        correlation_matrix=None,
+        correlation_cholesky=None,
+        validate_args=None,
+    ):
+        self.marginal_dist = marginal_dist
+        self.base_dist = MultivariateNormal(
+            covariance_matrix=correlation_matrix,
+            scale_tril=correlation_cholesky,
+        )
+        self.normal = Normal()
```
fehiepsi, line 34: I guess this is not needed. We can use `Normal().cdf/icdf` in the `sample` and `log_prob` method.

numpyro/distributions/copula.py
```diff
@@ -0,0 +1,74 @@
...
+        "correlation_cholesky",
+    ]
+
+    def __init__(
+        self,
+        marginal_dist,
+        correlation_matrix=None,
+        correlation_cholesky=None,
+        validate_args=None,
+    ):
```
fehiepsi, line 28: Could we have the assertion
```
if len(marginal_dist.event_shape) > 0:
    raise ValueError("`marginal_dist` needs to be a univariate distribution.")
```

numpyro/distributions/copula.py
```diff
@@ -0,0 +1,74 @@
...
+            covariance_matrix=correlation_matrix,
+            scale_tril=correlation_cholesky,
+        )
+        self.normal = Normal()
+
+        event_shape = self.base_dist.event_shape
+        batch_shape = lax.broadcast_shapes(
+            self.marginal_dist.batch_shape[:-1],
+            self.base_dist.batch_shape,
+        )
```
fehiepsi, line 40: I think we can have
```
    if len(self.marginal_dist.batch_shape) > 0:
        batch_shape = lax.broadcast_shapes(
            self.marginal_dist.batch_shape[:-1],
            self.base_dist.batch_shape,
        )
    else:
        batch_shape = self.base_dist.batch_shape
```

numpyro/distributions/copula.py
```diff
@@ -0,0 +1,74 @@
...
+
+from jax import lax, numpy as jnp
+
+import numpyro.distributions.constraints as constraints
+from numpyro.distributions.continuous import MultivariateNormal, Normal
+from numpyro.distributions.distribution import Distribution
+from numpyro.distributions.util import is_prng_key, validate_sample
+
+
+class GaussianCopula(Distribution):
```
fehiepsi, line 12: Could you add docstring to this distribution?

numpyro/distributions/continuous.py
fehiepsi: sorry, I meant `len(marginal_dist.event_shape) == 0`.

fehiepsi (commented):
Looks great to me overall! I just have a few small comments below. For testing, you can add your original multivariate Beta
```python
class GaussianCopulaBeta(GaussianCopula):
    arg_constraints = {"concentration0": constraints.positive,
        "concentration1": constraints.positive,
        "correlation_matrix": constraints.corr_matrix,
        "correlation_cholesky": constraints.corr_cholesky}
    support = constraints.independent(constraints.unit_interval, 1)

    def __init__(self, concentration0, concentration1, correlation_matrix=None, correlation_cholesky=None, *, validate_args=False):
        super().__init__(Beta(concentration0, concentration1), correlation_matrix, correlation_cholesky, validate_args=validate_args)
        self.concentration0, self.concentration1 = promote_shapes(concentration0, concentration1, shape=self.batch_shape + self.event_shape)
```
and add several test cases for it in [this list](https://github.com/pyro-ppl/numpyro/blob/master/test/test_distributions.py#L455). Our current tests already cover goodness of fit, mean, shape,...

numpyro/distributions/copula.py
hessammehr: I'm not sure this check is necessary, in the case that `self.marginal_dist.batch_shape == ()` `self.marginal_dist.batch_shape[:-1]` still gives the correct value of `()`. My tests with `Beta(1.0, 2.0)` seem to behave correctly regardless of the event/batch shape of the base distribution.

Happy to make this change regardless if you feel that it should be included.

fehiepsi: Oh nice! I thought the slice op will give error - thanks

numpyro/distributions/copula.py
```diff
@@ -0,0 +1,108 @@
...
+        )
+        return copula_lp + marginal_lps.sum(axis=-1)
+
+    @property
+    def mean(self):
+        return jnp.broadcast_to(self.marginal_dist.mean, self.shape())
+
+    @constraints.dependent_property(is_discrete=False, event_dim=1)
+    def support(self):
+        return constraints.independent(self.marginal_dist.support, 1)
```
fehiepsi, line 92: Could you add some JAX methods to help convert this distribution into pytree
```python
    def tree_flatten(self):
        marginal_flatten, marginal_aux = self.marginal_dist.tree_flatten()
        return (marginal_flatten, self.base_dist.scale_tril), (type(self.marginal_dist), marginal_aux,)

    @classmethod
        marginal_flatten, correlation_cholesky = params
        marginal_cls, marginal_aux = aux_data
        marginal_dist = marginal_cls.tree_unflatten(marginal_aux, marginal_flatten)
        return cls(marginal_dist, correlation_cholesky=correlation_cholesky)
```

numpyro/distributions/copula.py
```diff
@@ -0,0 +1,108 @@
...
+    ):
+        super().__init__(
+            Beta(concentration1, concentration0),
+            correlation_matrix,
+            correlation_cholesky,
+            validate_args=validate_args,
+        )
+        self.concentration1, self.concentration0 = promote_shapes(
+            concentration1, concentration0, shape=self.batch_shape + self.event_shape
+        )
```
fehiepsi, line 121: Could you add a jax unflatten method
```
    @classmethod
        marginal_flatten, correlation_cholesky = params
        marginal_cls, marginal_aux = aux_data
        marginal_dist = marginal_cls.tree_unflatten(marginal_aux, marginal_flatten)
        return cls(marginal_dist.concentration1, marginal_dist.concentration0, correlation_cholesky=correlation_cholesky)
```

test/test_distributions.py
```diff
@@ -358,6 +358,14 @@ def get_sp_dist(jax_dist):
...
     T(dist.GaussianRandomWalk, 0.1, 10),
     T(dist.GaussianRandomWalk, np.array([0.1, 0.3, 0.25]), 10),
+    T(
+        dist.GaussianCopulaBeta,
+        np.array([7.0, 2.0]),
+        jnp.array([4.0, 10.0]),
+        np.array([[1.0, 0.75], [0.75, 1.0]]),
+    ),
+    T(dist.GaussianCopulaBeta, 2.0, 1.0, np.eye(3)),
+    T(dist.GaussianCopulaBeta, 2.0, 1.0, np.full((5, 3, 3), np.eye(3))),
```
fehiepsi, line 368: Could you modify [this line](https://github.com/pyro-ppl/numpyro/blob/master/test/test_distributions.py#L1685) to skip checking variance of this distribution?
```
        if isinstance(d_jax, (dist.CAR, dist.GaussianCopula)):
            pytest.skip(f"{type(d_jax)} distribution does not have `variance` implemented.")
```

hessammehr:
@fehiepsi thanks for the comments/suggestions. There is one outstanding issue with the goodness of fit test which I suspect might be linked to an issue in the calculation of `log_prob`.

fehiepsi:
I guess you need
```
quantiles = Normal().icdf(self.marginal_dist.cdf(value))
```
?

hessammehr:
Thanks, @fehiepsi, yes that makes sense! I can confirm the `log_prob`s are identical between the marginal distribution and an uncorrelated copula, which is good. Unfortunately the gof tests are still failing. I managed to make 2 out of 3 pass by increasing the number of samples to 50000 for this particular distribution but the last one falls short of the 2e-5 cutoff by a factor of ~2 and increasing the number of samples further doesn't seem to help.

test/test_distributions.py
fehiepsi: pls ignore, we have `variance` now.

test/test_distributions.py
```diff
@@ -358,6 +358,14 @@ def get_sp_dist(jax_dist):
...
     T(dist.GaussianRandomWalk, 0.1, 10),
     T(dist.GaussianRandomWalk, np.array([0.1, 0.3, 0.25]), 10),
+    T(
+        dist.GaussianCopulaBeta,
+        np.array([7.0, 2.0]),
+        np.array([4.0, 10.0]),
+        np.array([[1.0, 0.75], [0.75, 1.0]]),
+    ),
+    T(dist.GaussianCopulaBeta, 2.0, 1.0, np.eye(3)),
+    T(dist.GaussianCopulaBeta, 2.0, 1.0, np.full((5, 3, 3), np.eye(3))),
```
fehiepsi, line 368: Could you help me add dtype conversion logic to [those utilities](https://github.com/pyro-ppl/numpyro/blob/41669121410102245e61df473ecbe5ff8109065d/numpyro/distributions/util.py#L423)?
```
dtype = jnp.result_type(float)
jnp.asarray(a/b/x/y) -> jnp.asarray(a/b/x/y, dtype=dtype)
```

fehiepsi:
Maybe using Beta(2, 1) in [0, 1]^3 is a bit extreme (toward 1 in the [0, 1] domain) for the test. How about using `Beta(2, 1.5)` instead?

hessammehr:
Switching to `Beta(2, 1.5)` seems to have fixed the goodness of fit issue, thanks @fehiepsi. I also ended up adding type conversions to both `betainc` and `betaincinv` since we make use of both here.

fehiepsi:
Maybe `jnp.asarray` is not robust to custom gradient. How about using `jnp.array` rather than `jnp.asarray`? Could you also update similar change for `gammaincinv` in the util file?

hessammehr:
@fehiepsi that seems to get rid of the incompatible types, thanks.

hessammehr:
Just a quick heads-up. The change to `betainc` may have caused a large performance regression. I'll have a closer look to confirm.

fehiepsi (approved):
Looks great to me! Could you address the flatten/unflatten comments? Essentially, given a copula dist `d`, we want `jit(lambda x: x)(d)` is still a copula with valid sample/log_prob methods. Currently, I don't think that we can call log_prob/sample with the default flattening logic.

numpyro/distributions/copula.py
fehiepsi: @hessammehr we need to update the Beta version because
```
return cls(marginal_dist, correlation_cholesky=correlation_cholesky)
```
in the general class does not work here

numpyro/distributions/copula.py
```diff
@@ -0,0 +1,151 @@
...
+        normal_samples = self.base_dist.expand(shape).sample(key)
+        cdf = Normal().cdf(normal_samples)
+        return self.marginal_dist.icdf(cdf)
+
+    @validate_sample
+    def log_prob(self, value):
+        # Ref: https://en.wikipedia.org/wiki/Copula_(probability_theory)#Gaussian_copula
+        # see also https://github.com/pyro-ppl/numpyro/pull/1506#discussion_r1037525015
+        marginal_lps = self.marginal_dist.log_prob(value)
+        quantiles = Normal().icdf(self.marginal_dist.cdf(value))
```
fehiepsi, line 78: @hessammehr I guess you can clip here. cdf might return 0 or 1 so we can clip by tiny and eps. You can use `util.clamp_probs` for this

numpyro/distributions/copula.py
```diff
@@ -0,0 +1,164 @@
...
+
+        super().__init__(
+            Beta(concentration1, concentration0),
+            correlation_matrix,
+            correlation_cholesky,
+            validate_args=validate_args,
+        )
+        self.concentration1, self.concentration0 = promote_shapes(
+            concentration1, concentration0, shape=self.batch_shape + self.event_shape
+        )
```
hessammehr, line 153: @fehiepsi I cannot say with certainty whether this shape promotion is necessary and it seems strange that we now have two sets of concentration factors (one inside the Beta marginal and one stored in the class itself).

numpyro/distributions/copula.py
```diff
@@ -0,0 +1,164 @@
...
+
+    def tree_flatten(self):
+        marginal_data, _ = self.marginal_dist.tree_flatten()
+        return (marginal_data, self.base_dist.scale_tril), None
+
+    @classmethod
+    def tree_unflatten(cls, aux_data, params):
+        (concentration0, concentration1), correlation_cholesky = params
```
hessammehr, line 161: It looks like `Beta.tree_flatten` returns the concentration factors in reverse. Is this something we want to raise a separate issue for?

numpyro/distributions/copula.py
fehiepsi: You can remove them. They're not used to compute anything here

numpyro/distributions/copula.py
hessammehr: No problem, I'll create a separate PR.

numpyro/distributions/copula.py
hessammehr: Okay, thanks!

numpyro/distributions/copula.py
hessammehr: Two possible solutions:

1. In the generic flattening algorithm, do not sort the `arg_constraints` so they are produced in the same order listed in the definition of each distribution.
2. Override `tree_flatten` and `tree_unflatten` just for beta.

Which one would you be tempted to go with?

fehiepsi: I think the first solution is better. Thanks!

fehiepsi (approved):
Happy to merge this when tests pass. Thanks, @hessammehr!

PR #1507, opened 2022-12-05
Adopt singleton semantics for globally defined constraint instances

### LINKED ISSUE #1378: Constraints object are not robust to pickling
Hi, and first of all, thank you for this awesome project!

Properties of `Transforms` object such as `codomain` rely on identity checks such as this one:
https://github.com/pyro-ppl/numpyro/blob/34e0cdf4fa0ab9a0300a0d894d6758419fb46f40/numpyro/distributions/transforms.py#L173
which are not robust to serialization roundtrips (of the form `pickle.loads(pickle.dumps(obj))`, which lead to the creation of new `Constraints` objects). Because of this, the following code snippet fails with: 

```python
In [1]: import jax.numpy as jnp
   ...: import pickle
   ...: from numpyro.distributions.transforms import AffineTransform
   ...:
   ...: a = AffineTransform(loc=jnp.array([0., 0.]), scale=jnp.array([1., 1.]))
   ...: print(a.inv._inv.codomain)
   ...:
   ...: a_roundtripped = pickle.loads(pickle.dumps(a))
   ...: print(a_roundtripped.inv._inv.codomain)
```

```python
Real()
NotImplementedError                       Traceback (most recent call last)
<ipython-input-2-399a4b34ce15> in <module>
      7
      8 a_roundtripped = pickle.loads(pickle.dumps(a))
----> 9 print(a_roundtripped.inv._inv.codomain)

~/repos/numpyro/numpyro/distributions/transforms.py in codomain(self)
    195                 )
    196         else:
--> 197             raise NotImplementedError
    198
    199     def __call__(self, x):

NotImplementedError:
```

There are at least 2 ways that I can think of to fix this issue: 

- either rely on equality checks instead of identity checks, and include sensible `__eq__` methods on constraints objects
- or reduce `constraints` objects using the little known singleton reducing semantics by making `Constraints.__reduce__` return a string, and not the traditional reconstruction tuple:
```python
In [1]: import pickle
   ...: class _Real:
   ...:     def __reduce__(self):
   ...:         # ask `pickle.load` to grab the top-level `real` object from the module in which _Real is defined
   ...:         return "real"
   ...: real = _Real()
   ...: pickle.loads(pickle.dumps(real)) is real
Out[1]: True
```
The latter option would be an elegant solution, only if it makes sense to treat constraints as singleton instances.

pierreglaser:
Closes #1378.


fritzo:
Doh, thanks for fixing this issue!  In retrospect we probably should have avoided the singletons and required the slightly less clever notation `constraints.Real()` 😅 

LGTM, but I defer to @fehiepsi

fritzo:
On second thought, I think it would be cleaner to handle this as in general (rather than merely for pickling) by following the singleton pattern:
```python
class _Real(Constraint):
    def __new__(cls):
        if not hasattr(cls, "_instance"):
            cls._instance = super(_Real, cls).__new__(cls)
        return cls._instance
```
That way `_Real() is _Real()` even outside of pickling.  WDYT?

pierreglaser:
I don't have a strong opinion. `_Real` is private anyways, so it should not be accessed nor instantiated outside of the `constraints` module; only its singleton`real` should be. But as I said, no strong opinion.

fritzo:
I guess another advantage of `.__new__()` over `.__reduce__()` is that it works with `copy.copy`:
```python
constraints.real is copy.copy(constraints.real)
```
And thanks for bringing up this issue!  It's also a problem with `torch.distributions.constraints` and we've never thought it through fully.  In PyTorch we tried to make the `_Real` class private but devs ended up using it in type annotations and unit tests anyway, ignoring the intentionally public `type(real)` 🤷

fritzo:
> copy.copy actually __reduces__ 

Oh thanks for explaining, I didn't realize that!

fehiepsi:
Thanks @pierreglaser (and @fritzo for chiming in)! Could you also add this logic or the `__new__` one for other constraints?

pierreglaser:
Done.

numpyro/distributions/constraints.py
```diff
@@ -95,7 +95,22 @@ def feasible_like(self, prototype):
...
+class _SingletonConstraint(Constraint):
+    """
+    A constraint type which has only one canonical instance, like constraints._real,
+    and unlike constraints._Interval.
+    """
+
+    def __new__(cls):
+        if not hasattr(cls, "_instance"):
+            cls._instance = super(_SingletonConstraint, cls).__new__(cls)
+        elif type(cls._instance) is not cls:
```
fehiepsi, line 107: Can we join two branches:
```
if (not hasattr(cls, "_instance")) or (type(cls._instance) is not cls):
```

fehiepsi (dismissed):
LGTM! Thanks @pierreglaser!

pierreglaser:
Actually, the current logic provides no way to treat as singleton instances the constraints coming from parametrized classes and exposed as top-level variables of the `constraints` module, such as:

```python
positive = _GreaterThan(0.0)
```

Treating them as complete singleton instances would require introspecting the arguments of the constructor and using a lookup-table mechanism to return instances already created for a given set of arguments.

Preserving identity semantics during pickling (which is ultimately what this PR was for) can be done by adding a lookup table of all global variables of the `constraints` module, and pickling the global variables by reference during `__reduce__`. Note that this logic does not require any singleton logic, such as the   `__new__` method introduced in previous commits.

Ultimately, I think addressing  1 (treating constraints a singleton when applicable) is more complex, and to be fair, out of scope for this PR, which originally addressed 2 (preserving identity semantics during pickling). So I'd rather rollback any changes made to treat `Constraint` objects as singleton, and simply address 2 in this PR.

WDYT @fehiepsi @fritzo ?

pierreglaser:
(By the way, I can't parse the lint failure from the CI, any help would be appreciated)

fritzo:
> no way to treat as singleton instances the constraints coming from parametrized classes

Thanks for thinking hard about this issue!

I believe that's fine.  Here's my reasoning:
1. The biggest pain point I've seen with pickling constraints is when unpickling non-parametrized constraints like `constraints.real` causing errors in code that dispatches on `if constraint is constraints.real`.  I've only ever seen `constraint is` used with non-parametrized classes; all parametrized classes dispatch via `isinstance` or something like `singledispatch` which examine the _type_ rather than the _identity_ of a constraint object.
2. At least in Pyro (unsure about NumPyro), we try to avoid data-dependent control flow.  In this situation of pickling constraints, 'data' means the parameters of a constraint, and 'control flow' is something that depends/dispatches on the _identity_ of a constraint object rather than it's _type_.  Thus it seems fine to deduplicate/singletonize only the non-parametrized classes.
3. More generally, we've tried to avoid data-dependent memoization, which I believe is what would be required to deduplicate parametrized types.  This is more important in PyTorch where data is mutable, but it may also be a good habit in JAX even if data is immutable.

pierreglaser:
I agree that, say, for *any* instance of the parametrized `_GreaterThan` other than `positive`, it makes sense to use `isinstance` and introspection checks, for the special case of `numpyro.distributions.constraints.positive` (also an instance of `_GreaterThan`), I would have imagined that the semantics is exactly the same as for other top-level constraints like `numpyro.distributions.constraints.real`. Can you confirm your opinion on that point?

numpyro/distributions/constraints.py
```diff
@@ -95,7 +95,22 @@ def feasible_like(self, prototype):
         raise NotImplementedError
 
 
-class _Boolean(Constraint):
+class _SingletonConstraint(Constraint):
+    """
+    A constraint type which has only one canonical instance, like constraints._real,
+    and unlike constraints._Interval.
```
fehiepsi, line 117: typo: constraints.real,  constraints.interval

fehiepsi (commented):
Though it's convenient to have `constraint is constraints.positive`, I feel that it is a better practice to use the check `isinstance(constraint, constraints.greater_than)` for a couple of reasons:
+ _TOPLEVEL_CONSTRAINTS is not exposed to users, so the semantic for user-defined constraints will be different from the "toplevel" ones,
+ avoid deviating from Pyro/PyTorch (related issue https://github.com/pytorch/pytorch/issues/50616)

I'm not sure how to tell users about the differences of `real` and `greater_than`. Probably we can add an attribute named `is_singleton` which is False by default.

I did several `grep`s and found that we didn't follow the expected behavior at those places:
+ [is constraints.unit_interval](https://github.com/pyro-ppl/numpyro/blob/aefadc1b78da3b0671fd053d11817d426ef68c8b/numpyro/distributions/transforms.py#L1100)
+ [is constraints.positive](https://github.com/pyro-ppl/numpyro/blob/aefadc1b78da3b0671fd053d11817d426ef68c8b/numpyro/distributions/transforms.py#L1070)
+ [in (constraints.real, constraints.real_vector)](https://github.com/pyro-ppl/numpyro/blob/aefadc1b78da3b0671fd053d11817d426ef68c8b/numpyro/infer/util.py#L212)

All of these were for efficiency. I'll remove them in a separate PR.

fritzo:
> Can you confirm your opinion on that point [whether constraints.positive should behave like constraints.real]?

I do see the parsimony of `constraint is constraints.positive` but I agree with @fehiepsi that it's probably safer to use `isinstance` here in practice.  Practically, I think we could get your idea working for python-only data (no arrays) using a dependent-singleton pattern

<details>

something like this:
```python
class _DependentSingleton:
    def __new__(cls, *args):  # TODO convert **kwargs to *args
        # only cache simple python things
        if not all(isinstance(arg, (int, float)) for arg in args):
            return super().__new__(*args)

        # weakly memoize
        cache = cls.__dict__.setdefault("_instance_cache", WeakValueDictionary())
        instance = cache.get(args)
        if instance is None:
            instance = super().__new__(*args)
            cache[args] = instance
        return instance
```

</details>

but that's a bit complex and seems out of scope for your nice clean first PR.

fehiepsi:
Alternatively, I guess we can also subclass those constraints, like
```
class _Positive(_GreaterThan):
    def __init__(self):
        super().__init__(0.)

positive = _Positive()
```

pierreglaser:
> Though it's convenient to have constraint is constraints.positive, I feel that it is a better practice to use the check isinstance(constraint, constraints.greater_than) for a couple of reasons

As such, this `isinstance` check will capture any `greater_than` instance, and not only `positive`-like instances. A full-blown `positive` check would require parameter introspection, and becomes much more verbose than an identity check. 

> _TOPLEVEL_CONSTRAINTS is not exposed to users, so the semantic for user-defined constraints will be different from the "toplevel" ones

IMO, defining equality/identity semantics for user-defined constraints should be left to the user defining them. To me, this sounds more understandable and uncontroversial than explaining that the two built-in`constraints.real` and `constraints.positive`, objects of  seemingly similar nature, should be treated differently for implementation-details reasons.

> Alternatively, I guess we can also subclass those constraints, like
> 
> ```
> class _Positive(_GreaterThan):
>     def __init__(self):
>         super().__init__(0.)
> 
> positive = _Positive()
> ```

I like this solution a lot @fehiepsi  :) it is more elegant than a global lookup table, and preserves identity semantics for top-level singleton objects, which I find more natural and concise than instance+attribute checking. If that sounds good to you, I'd happily implement this + singleton-style `__new__`/`__reduce__` for all parametrized and non-parametrized singletons, after which the PR should be good to merge, cc @fritzo. 

Long term, my suggestion would maybe be to expose a `check_constraint_equals` utility that unifies equality/identity checks for both singleton-style constraints and non-singleton-style constraints, which would work for all builtin constraints, and default to an `__eq__` check for user-defined ones. This change could be accompanied a paragraph in the docs about it, indicating that user should define `__eq__` for user-defined constraints subclasses in order for their instances to work with `check_constraint_equals`.

pierreglaser:
Regarding the (still open) https://github.com/pytorch/pytorch/issues/50616, I don't believe that `numpyro` (and `pyro`, provided that its constraint module is structured as in `numpyro`) should expose singleton **classes**. `constraint.real` is in essence a singleton object and different from parametrized `interval` instances. It is logical than they should be treated differently in equality check - the (limited) increase in code compacity resulting from exposing `_Real` instead of `_Real()`  cannot outweight that point.

fehiepsi:
> A full-blown positive check would require parameter introspection, and becomes much more verbose than an identity check.

Agreed! I tried to add some logic to check for `constraints.positive/unit_interval` in #1519 but eventually gave up due to the complexity (e.g. we need to check if `lower_bound` is `int/float`, then compare the value to `0`).

> I'd happily implement this + singleton-style __new__/__reduce__ for all parametrized and non-parametrized singletons, after which the PR should be good to merge

LGTM. It's great to have this in 0.11.0 release. (not important: if possible, could you revert the checks for `positive`, `unit_interval` in #1519 and add a comment that the check is possible after this PR? - I don't have a strong opinion regarding this - given that there are many positive-support distributions, we can gain some speed).

pierreglaser:
@fehiepsi @fritzo I implemented the  solution agreed upon above. I think the PR is ready for another round of review.

numpyro/distributions/constraints.py
```diff
@@ -284,6 +302,15 @@ def feasible_like(self, prototype):
...
 
+class _RealVector(_IndependentConstraint, _SingletonConstraint):
+    def __init__(self):
+        super().__init__(_Real(), 1)
+
+
+class _RealMatrix(_IndependentConstraint, _SingletonConstraint):
+    def __init__(self):
+        super().__init__(_Real(), 2)
+
```
fritzo, line 313: nit: run `black` or `make format`

fritzo (approved):
Looks great, thanks for making this module much more consistent!

numpyro/infer/util.py
```diff
@@ -209,10 +209,7 @@ def _unconstrain_reparam(params, site):
             if jnp.ndim(p) > expected_unconstrained_dim:
                 p = p[i]
 
-        if support is constraints.real or (
-            isinstance(support, constraints.independent)
-            and support.base_constraint is constraints.real
-        ):
```
fehiepsi, line 215: could you revert this change? the current logic also supports `real_matrix` etc.

numpyro/distributions/transforms.py
```diff
@@ -1065,8 +1065,11 @@ def _transform_to_corr_matrix(constraint):
     )
 
 
+@biject_to.register(type(constraints.positive))
```
fehiepsi, line 1068: maybe having a separate `biject_to` for this?
```
@biject_to.register(type(constraints.positive))
def _transform_to_positive(constraint):
    return ExpTransform()
```

numpyro/distributions/transforms.py
```diff
@@ -1085,16 +1088,22 @@ def _transform_to_less_than(constraint):
...
+@biject_to.register(type(constraints.real_vector))
 @biject_to.register(constraints.independent)
 def _biject_to_independent(constraint):
     return IndependentTransform(
         biject_to(constraint.base_constraint), constraint.reinterpreted_batch_ndims
     )
 
 
+@biject_to.register(type(constraints.circular))
+@biject_to.register(type(constraints.unit_interval))
```
fehiepsi, line 1101: like above, maybe
```
@biject_to.register(type(constraints.unit_interval))
def _transform_to_unit_interval(constraint):
    return SigmoidTransform()
```

pierreglaser:
@fritzo @fehiepsi thanks for the review. all comments should have been addressed now.

test/test_pickle.py
```diff
@@ -13,6 +13,32 @@
 
 import numpyro
 import numpyro.distributions as dist
+from numpyro.distributions.constraints import (
+    boolean,
+    circular,
+    corr_cholesky,
+    corr_matrix,
+    dependent,
```
fehiepsi, line 21: Do you want to test for `dependent`? It seems that this is causing lint failing. You can detect lint issues with `make lint` (see [contributing docs](https://github.com/pyro-ppl/numpyro/blob/master/CONTRIBUTING.md#development))

fehiepsi (approved):
LGTM pending lint issue.

test/test_pickle.py
pierreglaser: my bad, missed that.

pierreglaser:
Addressed the linting issue. Some test related to `funsor` is failing though. Can't look at it now, but will investigate later.

fehiepsi:
Hi @pierreglaser, sorry, I'll take a closer look. It seems that in funsor, we use support's name for some [logics](https://github.com/pyro-ppl/funsor/blob/master/funsor/distribution.py#L340). Maybe it's easiest to let me to add a patch and create an issue there.

fehiepsi (approved):

Issue #1383, opened 2022-04-06
Allow rendering deterministic sites


fehiepsi:
Currently, we skip deterministic sites while rendering the model. It would be nice to support plotting deterministic nodes (e.g. to make the rendered model clearer when using reparam handlers). This is especially useful for hierarchical models like the one in #1295

YanniPapandreou:
Hi, is anyone working on this? If not I'd be happy to give this a try - I'm quite new to this so any guidance would be great :)

fehiepsi:
Thanks,  @YanniPapandreou! The inspect logic is in [this file](https://github.com/pyro-ppl/numpyro/blob/f81ce325de402d9a9891497136a36e3b44f982cc/numpyro/infer/inspect.py#L192). Currently, we only consider messages with `sample` type. You can modify the dictionaries in that method to also include `deterministic` sites. You can get started by making a model with deterministic site, obtain a trace of it, play around with provenance logic.

When you start looking into provenance, you will see that we use log_prob method to determine the dependency. For deterministic nodes, you can simply use `identity` function instead.

YanniPapandreou:
Thanks @fehiepsi ! I have started looking into this. 

Do you have any resources for learning about provenance logic? I have managed to get the deterministic sites but I am not sure I understand how `log_prob` is used for samples to get the dependency. I am also not sure about which `identity` function you are referring to is this **the** identity function which seems to be implemented in `numpyro.utils`?

fehiepsi:
Sorry @YanniPapandreou, I miss your response! Let me make the previous comment clearer.

#### provenance

This is used to see which variable in the output depends on which variable in the input. For example,
```python
from numpyro.ops.provenance import ProvenanceArray, eval_provenance, get_provenance
a = ProvenanceArray(1., frozenset({"a"}))
b = ProvenanceArray(2., frozenset({"b"}))
c = 3.
f = lambda a, b, c: (a + b + c, a - c)
d, e = eval_provenance(f, a, b, c)
assert get_provenance(d) == frozenset({"a", "b"})
assert get_provenance(e) == frozenset({"a"})
```

#### model dependencies

[...]

PR #1508, opened 2022-12-05
Rendering deterministic sites [WIP]

YanniPapandreou:
First draft for rendering deterministic sites as requested in #1383.

As detailed in my comment [here](https://github.com/pyro-ppl/numpyro/issues/1383#issuecomment-1336274247) I have changed the following:

-  Added `"deterministic"` to [the list](https://github.com/pyro-ppl/numpyro/blob/b22d9c1ccbfa543852bfa2c0d63d206d0b47ac92/numpyro/handlers.py#L777) you suggested.
-  Changed `get_model_relations`  (and `get_log_probs`) [here](https://github.com/pyro-ppl/numpyro/blob/b22d9c1ccbfa543852bfa2c0d63d206d0b47ac92/numpyro/infer/inspect.py#L256) instead of `get_dependencies` and `_get_log_probs` which you referred to above. Please let me know if I am wrong but it seems to me that `get_model_relations` is used in `render_model` instead of `get_dependencies`.

With these changes for the following model:

```python
def model(data):
    m = numpyro.sample("m", dist.Normal(0, 1))
    sd = numpyro.sample("sd", dist.LogNormal(m, 1))
    m_transformed = numpyro.deterministic("m_transformed", m + 1)
    with numpyro.plate("N", len(data)):
        numpyro.sample("obs", dist.Normal(m_transformed, sd), obs=data)
```
 
I was able to obtain this diagram:

image

This correctly shows that `m_transformed` depends on `m` but it misses the fact that `obs` depends on `m` via `m_transformed`. If I uncomment [line 386](https://github.com/pyro-ppl/numpyro/commit/bbc4769bb12e0e17f770929fa78028019921fa0a) I instead get this diagram:

image

I think that uncommenting the line is wrong as I am basically creating a new ProvenanceArray for `m_transformed` which only depends on itself. 

@fehiepsi  do you have any suggestions on how I can make `obs` realise it depends on `m_transformed`?


numpyro/infer/inspect.py
```diff
@@ -364,16 +366,24 @@ def get_log_probs(sample):
             data=sample
```
fehiepsi, line 366: Thanks @YanniPapandreou for the clear explanation! It was my mistake - suggested updating `substitute` handler. You can revert that change and use the follow handler instead.
```python
class substitute_deterministic(handlers.substitute):
        if msg["type"] == "deterministic":
            msg["args"] = (msg["value"],)
            msg["value"] = self.data.get(msg["name"])

with handlers.trace() as tr, handlers.seed(rng_seed=0):
    with handlers.substitute(data=sample), substitute_deterministic(data=sample):
        model(*model_args, **model_kwargs)
```
Essentially, we are moving the actual value (obtained from the computation over other values) to `msg["args"]`, then in `provenance_arrays` below, we can do
```
            elif site["type"] == "deterministic":
                provenance_arrays[name] = site["args"][0]
```
This way we can know which variables that this deterministic variable depends on.

numpyro/infer/inspect.py
```diff
@@ -364,16 +366,24 @@ def get_log_probs(sample):
...
+            elif site["type"] == "deterministic":
+                provenance_arrays[name] = site["value"]
+        return provenance_arrays
 
     samples = {
         name: ProvenanceArray(site["value"], frozenset({name}))
         for name, site in trace.items()
         if (site["type"] == "sample" and not site["is_observed"])
+        # or site["type"] == "deterministic"
     }
```
fehiepsi, line 388: After getting `sample_sample` dict, we know which variables the deterministic site depends on, I guess we can (in most cases) tell which plates the site belongs to:
```
deterministic_sample = {name: v for name, v in sample_sample.items() if sample_dist[name] == "Deterministic"}
for name, deps in sample_sample.items():
    if sample_dist[name] == "Deterministic":
        for plate, samples in plate_samples.items():
            if len(set(deps) - set(samples)) == 0:
                samples.append(name)
                
```

fehiepsi (commented):
Could we have a different symbol for Deterministic sites in `render_graph`. Something like 
```
style = "filled,dashed" if node_data[rv]["distribution"] == "Deterministic" else "filled"
```

numpyro/infer/inspect.py
YanniPapandreou: > Thanks @YanniPapandreou for the clear explanation! It was my mistake - suggested updating `substitute` handler. You can revert that change and use the follow handler instead.

I have reverted this, for reference please see [here](https://github.com/pyro-ppl/numpyro/pull/1508/commits/e08a561c43f2a1fce684845a009faeb634ee05cd).

I have also created the handler you suggested in the `get_log_probs` function [here](https://github.com/pyro-ppl/numpyro/pull/1508/commits/ce10130630e3d6a8ef32c294d3bd209fa0cb88b3). I was getting errors with the code as is so I added the following key,value pairs to the message when it is a deterministic site:

```python
class substitute_deterministic(handlers.substitute):
                if msg["type"] == "deterministic":
                    msg["args"] = (msg["value"],)
                    msg["kwargs"] = {}
                    msg["value"] = self.data.get(msg["name"])
                    msg["fn"] = lambda x: x
```
This seems to work on the example.

numpyro/infer/inspect.py
```diff
@@ -554,8 +565,10 @@ def render_graph(graph_specification, render_distributions=False):
                     "$params", ""
                 )  # incase of neural network parameters
 
+            # use different symbol for Deterministic site
+            node_style = "filled,dashed" if node_data[rv]["distribution"] == "Deterministic" else "filled"
```
YanniPapandreou, line 569: > Could we have a different symbol for Deterministic sites in `render_graph`. Something like
> 
> ```
> style = "filled,dashed" if node_data[rv]["distribution"] == "Deterministic" else "filled"
> ```

These changes here deal with this.

YanniPapandreou:
@fehiepsi thanks for all your comments! I have added responses above and in the commit [here](https://github.com/pyro-ppl/numpyro/pull/1508/commits/ce10130630e3d6a8ef32c294d3bd209fa0cb88b3). With these changes I obtain the following diagram for the example above:

image

numpyro/infer/inspect.py
```diff
@@ -359,21 +361,40 @@ def _resolve_plate_samples(plate_samples):
     }
 
     def get_log_probs(sample):
+        class substitute_deterministic(handlers.substitute):
+            def process_message(self, msg):
+                if msg["type"] == "deterministic":
+                    msg["args"] = (msg["value"],)
+                    msg["kwargs"] = {}
+                    msg["value"] = self.data.get(msg["name"])
+                    msg["fn"] = lambda x: x
```
fehiepsi, line 370: nit: this is required when self.data does not include the deterministic value (no substitute). You provided deterministic inputs in `samples` below so no need to have this logic. Keeping this as-is is fine btw.

numpyro/infer/inspect.py
fehiepsi: Sorry, the logic to decide which plates a deterministic site belongs to is more complicated. Let me think a bit more.

fehiepsi: I think we need to store plate info in the deterministic message. Let me add another comment for it.

fehiepsi (commented):
Sorry @YanniPapandreou ! I thought that I already sent the reviews but it turns out that it's pending.

fehiepsi:
After the change, could you add a section in the tutorial https://github.com/pyro-ppl/numpyro/blob/master/notebooks/source/model_rendering.ipynb to illustrate how it looks. Thanks!

numpyro/infer/inspect.py
```diff
@@ -359,21 +361,40 @@ def _resolve_plate_samples(plate_samples):
...
+                model(*model_args, **model_kwargs)
+        # with handlers.trace() as tr, handlers.seed(rng_seed=0),handlers.substitute(
+        #     data=sample
+        # ):
+        #     model(*model_args, **model_kwargs)
+        # return {
+        #     name: site["fn"].log_prob(site["value"])
+        #     for name, site in tr.items()
+        #     if site["type"] == "sample"
+        # }
```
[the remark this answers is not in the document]
YanniPapandreou: I've removed these comments :)

numpyro/infer/inspect.py
```diff
@@ -328,7 +330,7 @@ def get_trace():
     sample_dist = {
         name: site["fn_name"]
         for name, site in trace.items()
-        if site["type"] == "sample"
+        if site["type"] == "sample" or site["type"] == "deterministic"
     }
 
     sample_plates = {
```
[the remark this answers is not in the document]
YanniPapandreou: I've made these changes as well. The relevant commits are [here](https://github.com/pyro-ppl/numpyro/commit/63f1c93e849600726fbba78256968c8adb719704) and [here](https://github.com/pyro-ppl/numpyro/commit/3225b1d77b93b8de6d89db7b1da9c150b3fe1ddc). Please let me know if anything is is wrong with the second one as I wasn't sure where exactly to place

```python
        if msg["type"] not in ("param", "sample", "plate", "deterministic"):
        ...
        if msg["type"] == "deterministic":
            return
```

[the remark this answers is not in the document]
fehiepsi: Thanks! The changes are correct. I think we need to use `if site["type"] in ["sample", "deterministic"]` at 3 lines below.

fehiepsi (commented):
Thanks @YanniPapandreou! I think you need to move "Rendering deterministic sites" section after "Distribution and Constraint annotations" to not affect the latter content.

Btw, what's the graph looks like for this model?
```
def model(data):
    with numpyro.plate("N", len(data)):
        m = numpyro.sample("m", dist.Normal(0, 1))
        sd = numpyro.sample("sd", dist.LogNormal(m, 1))
        # deterministic site
        m_transformed = numpyro.deterministic("m_transformed", m + 1)
        numpyro.sample("obs", dist.Normal(m_transformed, sd), obs=data)
```

numpyro/infer/inspect.py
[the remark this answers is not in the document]
YanniPapandreou: I am not sure where exactly you are referring to but I added this in the `inspect.py` file like so:

```python
  sample_plates = {
      name: [frame.name for frame in site["cond_indep_stack"]]
      for name, site in trace.items()
      if site["type"] in ["sample", "deterministic"]
  }
```

The commit is [here](https://github.com/pyro-ppl/numpyro/pull/1508/commits/7c077e3866646211175e8c6002230859dcc08d92).

fehiepsi:
Could you double check? The notebook seems to be the same as before but there is this new one [.virtual_documents/notebooks/source/model_rendering.ipynb](https://github.com/pyro-ppl/numpyro/pull/1508/files#diff-0732b1d3cddc053e3e53b0fbc9ae362e0a55401e46fd84080d24b45c9945bbe1)

numpyro/infer/inspect.py
```diff
@@ -359,21 +361,31 @@ def _resolve_plate_samples(plate_samples):
...
     def get_log_probs(sample):
+        class substitute_deterministic(handlers.substitute):
+            def process_message(self, msg):
+                if msg["type"] == "deterministic":
+                    msg["args"] = (msg["value"],)
+                    msg["kwargs"] = {}
+                    msg["value"] = self.data.get(msg["name"])
+                    msg["fn"] = lambda x: x
+
+
```
fehiepsi, line 372: It seems that CI is failing due to two blank lines. Could you run `make lint` or `make format` in the cloned numpyro folder to fix those issues

YanniPapandreou:
> Could you double check? The notebook seems to be the same as before but there is this new one [.virtual_documents/notebooks/source/model_rendering.ipynb](https://github.com/pyro-ppl/numpyro/pull/1508/files#diff-0732b1d3cddc053e3e53b0fbc9ae362e0a55401e46fd84080d24b45c9945bbe1)

Opps my bad, I accidentally pushed the wrong file

numpyro/infer/inspect.py
YanniPapandreou: `make lint` seems to produce an error, but `make format` works. I have pushed these changes.

fehiepsi:
@YanniPapandreou CI is failing with the change. Could you help me fix them by replacing [this line](https://github.com/pyro-ppl/numpyro/blob/2b70d7d985ad5badedf571073176d987b4e8f8b4/numpyro/handlers.py#L576) by
```
if key not in ("type", "name", "value", "cond_indep_stack"):
```
and [this line](https://github.com/pyro-ppl/numpyro/blob/2b70d7d985ad5badedf571073176d987b4e8f8b4/numpyro/distributions/transforms.py#L614) by
```
Array([0.3, 1.5], dtype=float32)
```

YanniPapandreou:
> @YanniPapandreou CI is failing with the change. Could you help me fix them by replacing [this line](https://github.com/pyro-ppl/numpyro/blob/2b70d7d985ad5badedf571073176d987b4e8f8b4/numpyro/handlers.py#L576) by
> 
> ```
> if key not in ("type", "name", "value", "cond_indep_stack"):
> ```
> 
> and [this line](https://github.com/pyro-ppl/numpyro/blob/2b70d7d985ad5badedf571073176d987b4e8f8b4/numpyro/distributions/transforms.py#L614) by
> 
> ```
> Array([0.3, 1.5], dtype=float32)
> ```

@fehiepsi  done and pushed these commits. Please let me know if anything else needs to be done :)

numpyro/distributions/transforms.py
```diff
@@ -611,7 +611,7 @@ class LowerCholeskyAffine(Transform):
        >>> scale_tril = jnp.array([[0.3, 0.0], [1.0, 0.5]])
        >>> affine = LowerCholeskyAffine(loc=loc, scale_tril=scale_tril)
        >>> affine(base)
-       DeviceArray([0.3, 1.5], dtype=float32)
+       Array([0.3, 1.5], dtype=float32)
```
fehiepsi, line 614: Sorry, @YanniPapandreou! This change is needed for jax 0.4 but currently, CI is running with python 3.7 which is not compatible with jax 0.4. So CI is installing jax < 0.4. Could you revert this change?

YanniPapandreou: @fehiepsi done and pushed!

fehiepsi (approved):
Thanks for this cool feature, @YanniPapandreou!

YanniPapandreou:
@fehiepsi thanks a lot for helping me along the way and thanks for this great package :)

PR #1509, opened 2022-12-09
Distribution flatten/unflatten: Do not sort attrs.

hessammehr:
Currently, `Distribution.tree_flatten` sorts distribution attrs alphabetically, so the order is sometimes different from what is declared in the definition, e.g. the beta distribution `Beta(concentration1, concentration0).tree_flatten() = (concentration0, concentration1)`. With this PR, attrs are used in the same order as declared within `Distribution.arg_constraints`.


fehiepsi (approved):
Thanks, @hessammehr!

Issue #1511, opened 2022-12-15
SineBivariateVonMises generates NaN log probability for negative correlation parameter


19MM97:
When using [SineBivariateVonMises](https://num.pyro.ai/en/stable/distributions.html#sinebivariatevonmises) in an MCMC setting with the NUTS sampler and hard assigning the correlation parameter of the SinveBivariateVonMises to a negative value, the model can not be initialised. Sampling from a SineBivariateVonMises with a negative correlation parameter does work though.
 @fehiepsi traced, that the log probability becomes NaN.

How to recreate:

`data = SineBivariateVonMises(0, 0, 1, 1, -1).sample(PRNGKey(42), (1000, ))`

and train this model:

```
@numpyro.handlers.reparam(
    config={"phi_loc": CircularReparam(), 
            "psi_loc": CircularReparam(),
           }
)
def min_example(data_2d):
    phi_loc = sample('phi_loc', VonMises(0, 10))
    psi_loc = sample('psi_loc', VonMises(0, 10))
    
    phi_conc = sample('phi_conc', Beta(1, 1))
    psi_conc = sample('psi_conc', Beta(1, 1))
    
    conc = -sample('conc', HalfNormal(1))
    
    depInd = SineBivariateVonMises(phi_loc, psi_loc, 70 * phi_conc, 70 * psi_conc, conc)
    
    obs = sample('obs', depInd, obs=data_2d)
```

Further information can be found in this [forum post](https://forum.pyro.ai/t/negative-correlation-sinebivariatevonmises/4786/3)

fehiepsi:
Could you add a full code to replicate the issue? We have the valid parameters
```
{'phi_loc': DeviceArray(0.29627275, dtype=float32),
 'psi_loc': DeviceArray(-0.03310037, dtype=float32),
 'phi_concentration': DeviceArray(19.143257, dtype=float32),
 'psi_concentration': DeviceArray(44.272667, dtype=float32),
 'correlation': DeviceArray(-0.6486334, dtype=float32)}
```
but gets invalid log_prob with distribution
```
SineBivariateVonMises(phi_loc, psi_loc, 70 * phi_conc, 70 * psi_conc, conc)
```
and observed `data_2d`. I guess we don't need to whole training process to see the issue.

fehiepsi:
cc @OlaRonning

OlaRonning:
[...]

PR #1515, opened 2022-12-17
Fixed `log_prob` for negative correlation in `SineBivariateVonMises`

OlaRonning:
Removes bias on correlation in `SBvM` normalization and fixes `NaN` in normalization for negative correlation (`corr < 0.`).  

From #1511.

### DIFF
--- a/numpyro/distributions/directional.py
+++ b/numpyro/distributions/directional.py
@@ -399,7 +399,7 @@ class SineBivariateVonMises(Distribution):
         lbinoms = num - 2 * den
 
         fs = lbinoms.reshape(-1, 1) + m * (
-            jnp.log(jnp.clip(corr**2, a_min=jnp.finfo(float).tiny))
+            jnp.log(jnp.clip(corr**2, a_min=jnp.finfo(jnp.result_type(float)).tiny))
             - jnp.log(4 * jnp.prod(conc, axis=-1))
         )
         fs += log_I1(49, conc, terms=51).sum(-1)
--- a/numpyro/distributions/directional.py
+++ b/numpyro/distributions/directional.py
@@ -389,7 +389,7 @@ class SineBivariateVonMises(Distribution):
 
     @lazy_property
     def norm_const(self):
-        corr = jnp.reshape(self.correlation, (1, -1)) + 1e-8
+        corr = jnp.reshape(self.correlation, (1, -1))
         conc = jnp.stack(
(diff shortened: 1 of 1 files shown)


numpyro/distributions/directional.py
```diff
@@ -400,7 +400,7 @@ def norm_const(self):
 
         fs = (
             lbinoms.reshape(-1, 1)
-            + 2 * m * jnp.log(corr)
+            + m * jnp.log(corr**2)
```
fehiepsi, line 403: Nice catch! I guess you can clip by tiny here in case users set corr=0?

numpyro/distributions/directional.py
```diff
@@ -400,7 +400,7 @@ def norm_const(self):
...
             - m * jnp.log(4 * jnp.prod(conc, axis=-1))
         )
         fs += log_I1(49, conc, terms=51).sum(-1)
```
fehiepsi, line 406: I'm not sure that I follow but it seems to me that norm_const does not have batch dimensions after fs.max() operator below. Is it intended?

fehiepsi: Probably you can remove mfs term because logsumexp does it under the hood.

numpyro/distributions/directional.py
OlaRonning: Sure, that makes sense.

numpyro/distributions/directional.py
OlaRonning: I see. I'll take out the `mfs` term.

numpyro/distributions/directional.py
```diff
@@ -398,14 +398,12 @@ def norm_const(self):
         den = special.gammaln(m + 1.0)
         lbinoms = num - 2 * den
 
-        fs = (
-            lbinoms.reshape(-1, 1)
-            + 2 * m * jnp.log(corr)
-            - m * jnp.log(4 * jnp.prod(conc, axis=-1))
+        fs = lbinoms.reshape(-1, 1) + m * (
+            jnp.log(jnp.clip(corr**2, a_min=jnp.finfo(float).tiny))
```
fehiepsi, line 402: The float finfo is too small for single precision, how about `jnp.finfo(jnp.result_type(float)).tiny`?

fehiepsi (dismissed):
Thanks, Ola! Just a small comment on the floating point.

fehiepsi (approved):

PR #1519, opened 2022-12-28
Avoid comparing constraints positive and unit_interval

fehiepsi:
Following #1507, this PR revises the logic for several unexpected usages which check for positive and unit_interval constraints.


fritzo (approved):

fehiepsi:
Thanks, Fritz!

PR #1528, opened 2023-01-28
Guess max plate nesting

ordabayevy:
Implemented guess of max plate nesting using `jax.eval_shape` interpreter.


fehiepsi (approved):
This is a pretty clean solution. Thanks, Yerdos!

PR #1531, opened 2023-01-31
Optimize reduction of enumerated guide sites

ordabayevy:
Use `opt_einsum` and `shared_intermediates` to reduce vars in dice factors not present in the cost term.

### DIFF
--- a/numpyro/infer/elbo.py
+++ b/numpyro/infer/elbo.py
@@ -849,8 +849,14 @@ class TraceEnum_ELBO(ELBO):
 
     def loss(self, rng_key, param_map, model, guide, *args, **kwargs):
         def single_particle_elbo(rng_key):
+            from opt_einsum import shared_intermediates
+
             import funsor
+            from funsor.cnf import _eager_contract_tensors
             from numpyro.contrib.funsor import to_data, to_funsor
+
+            logsumexp_backend = "funsor.einsum.numpy_log"
+            cache = dict()
 
             model_seed, guide_seed = random.split(rng_key)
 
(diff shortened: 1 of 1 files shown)


numpyro/infer/elbo.py
```diff
@@ -849,9 +849,15 @@ def __init__(self, num_particles=1, max_plate_nesting=float("inf")):
...
     def loss(self, rng_key, param_map, model, guide, *args, **kwargs):
         def single_particle_elbo(rng_key):
+            from opt_einsum import shared_intermediates
+
             import funsor
+            from funsor.cnf import _eager_contract_tensors
             from numpyro.contrib.funsor import to_data, to_funsor
 
+            logsumexp_backend = "funsor.einsum.numpy_log"
+            cache = dict()
```
fehiepsi, line 859: maybe using
```
with shared_intermediates() as cache:  # create a cache
    pass
```
following einsum tutorial? https://optimized-einsum.readthedocs.io/en/stable/sharing_intermediates.html

numpyro/infer/elbo.py
```diff
@@ -994,25 +1001,35 @@ def single_particle_elbo(rng_key):
                         )
                         for key in deps
                     ]
-
-                if dice_factors:
-                    dice_factor = reduce(lambda a, b: a + b, dice_factors)
-                    cost = cost * funsor.ops.exp(dice_factor)
-                if (scale is not None) and (not is_identically_one(scale)):
-                    cost = cost * to_funsor(scale)
```
fehiepsi, line 1002: IIUC those computations are performed in eager contract tensor?

ordabayevy: Which line are you referring to? 
Just to be clear eager contract tensor is used in cases that look like this:
``` 
Compute expectation of cost(b) wrt q(a)q(b|a):
# eliminate `a` first:
logq(b) = eager_contract_tensor(Variable(a), [logq(a), logq(b|a)])
# then compute expectation wrt q(b):
cost = sum(cost(b) * logq(b).exp())
```

numpyro/infer/elbo.py
ordabayevy: Done!

fehiepsi (approved):
LGTM given tests are passing.

ordabayevy:
Thanks Du for reviewing!

PR #1532, opened 2023-02-01
Mark gaussian shells flaky on CI

fehiepsi:
I tested that the example works with the latest jax/jaxlib version locally but they are flaky on CI. So I added a mark.


martinjankowiak:
what do the errors look like?

fehiepsi:
It just freezes. Seems like jaxns is going into an infinite loop. The current master branch is failing due to this.

martinjankowiak:
i see. i wonder if we can fix this by e.g. using a large `width`. in any case probably best to punt to jaxns

PR #1533, opened 2023-02-02
Use analytic kl divergence in `TraceEnum_ELBO`

ordabayevy:
This PR adds support for use of analytic kl divergence when possible to `TraceEnum_ELBO`.

### DIFF
--- a/numpyro/contrib/funsor/infer_util.py
+++ b/numpyro/contrib/funsor/infer_util.py
@@ -83,7 +83,12 @@ def config_enumerate(fn=None, default="parallel"):
 
 def config_kl(fn=None, sites=None):
     """
-    Configures analytic kl for all sites in ``sites`` in a NumPyro model.
+    Configures the ``kl`` flag in the ``infer`` dict for all sample sites in
+    ``sites`` in a NumPyro model. If ``kl`` is ``analytic``, and ``TraceEnum_ELBO``
+    is being used, an attempt is made to analytically compute the KL divergence
+    in the ELBO at the corresponding site. If ``analytic`` is specified and
+    it is not possible to analytically compute the KL divergence then an error
+    is raised.
 
     This can be used as either a function::
 
--- a/numpyro/contrib/funsor/infer_util.py
+++ b/numpyro/contrib/funsor/infer_util.py
@@ -83,7 +83,7 @@ def config_enumerate(fn=None, default="parallel"):
 
 def config_kl(fn=None, sites=None):
     """
-    Configures analytic kl for all relevant sites in a NumPyro model.
+    Configures analytic kl for all sites in ``sites`` in a NumPyro model.
 
     This can be used as either a function::
 
--- a/numpyro/contrib/funsor/infer_util.py
+++ b/numpyro/contrib/funsor/infer_util.py
@@ -81,6 +81,39 @@ def config_enumerate(fn=None, default="parallel"):
     return infer_config(fn, config_fn)
 
 
+def config_kl(fn=None, sites=None):
+    """
+    Configures analytic kl for all relevant sites in a NumPyro model.
+
+    This can be used as either a function::
+
+        model = config_kl(model)
+
+    or as a decorator::
+
+        @config_kl
+        def model(*args, **kwargs):
+            ...
+
+    :param callable fn: Python callable with NumPyro primitives.
+    :param set sites: Sites for which to use analytic KL solution.
+        If ``None`` all sites are set to analytic.
+    """
+    if fn is None:  # support use as a decorator
+        return functools.partial(config_kl, sites=sites)
+
+    def config_fn(site):
+        if (
+            site["type"] == "sample"
+            and (not site["is_observed"])
+            and (sites is None or site["name"] in sites)
+        ):
+            return {"kl": site["infer"].get("kl", "analytic")}
+        return {}
+
+    return infer_config(fn, config_fn)
+
+
 def _get_shift(name):
     """helper function used internally in sarkka_bilmes_product"""
     return len(re.search(r"^(_PREV_)*", name).group(0)) // 6
--- a/numpyro/handlers.py
+++ b/numpyro/handlers.py
@@ -221,7 +221,7 @@ class replay(Messenger):
             if guide_msg["type"] != "sample" or guide_msg["is_observed"]:
                 raise RuntimeError(f"Site {name} must be sampled in trace.")
             msg["value"] = guide_msg["value"]
-            msg["infer"] = guide_msg["infer"]
+            msg["infer"] = guide_msg["infer"].copy()
 
 
 class block(Messenger):
--- a/numpyro/infer/elbo.py
+++ b/numpyro/infer/elbo.py
@@ -708,7 +708,7 @@ class TraceGraph_ELBO(ELBO):
 
 
 def get_importance_trace_enum(
-    model, guide, args, kwargs, params, max_plate_nesting, model_deps
+    model, guide, args, kwargs, params, max_plate_nesting, model_deps, guide_desc
 ):
     """
     (EXPERIMENTAL) Returns traces from the enumerated guide and the enumerated model that is run against it.
@@ -743,7 +743,7 @@ def get_importance_trace_enum(
                 if (
                     is_model
                     and (name in guide_trace)
-                    and model_deps[name].isdisjoint(sum_vars)
+                    and model_deps[name].isdisjoint(sum_vars | guide_desc[name])
                 ):
                     try:
                         kl_qp = kl_divergence(
@@ -923,9 +923,15 @@ class TraceEnum_ELBO(ELBO):
 
             seeded_model = seed(model, model_seed)
             seeded_guide = seed(guide, guide_seed)
+            # get dependencies on nonreparametrizable variables
             model_deps, guide_deps = get_nonreparam_deps(
                 seeded_model, seeded_guide, args, kwargs, param_map
             )
+            # get descendants of variables in the guide
+            guide_desc = defaultdict(frozenset)
+            for name, deps in guide_deps.items():
+                for d in deps:
+                    guide_desc[d] |= frozenset([name])
 
             seeded_model = seed(model, model_seed)
             seeded_guide = seed(guide, guide_seed)
@@ -937,6 +943,7 @@ class TraceEnum_ELBO(ELBO):
                 param_map,
                 self.max_plate_nesting,
                 model_deps,
+                guide_desc,
             )
 
             check_model_guide_match(model_trace, guide_trace)
--- a/numpyro/infer/elbo.py
+++ b/numpyro/infer/elbo.py
@@ -947,6 +947,7 @@ class TraceEnum_ELBO(ELBO):
                 guide_desc,
             )
 
+            # TODO: fix the check below
             # check_model_guide_match(model_trace, guide_trace)
             _validate_model(model_trace, plate_warning="strict")
 
--- a/numpyro/infer/elbo.py
+++ b/numpyro/infer/elbo.py
@@ -941,7 +941,7 @@ class TraceEnum_ELBO(ELBO):
                 guide_desc,
             )
 
-            check_model_guide_match(model_trace, guide_trace)
+            # check_model_guide_match(model_trace, guide_trace)
             _validate_model(model_trace, plate_warning="strict")
 
             model_vars = frozenset(model_deps)
--- a/numpyro/infer/elbo.py
+++ b/numpyro/infer/elbo.py
@@ -707,7 +707,9 @@ class TraceGraph_ELBO(ELBO):
             return -jnp.mean(vmap(single_particle_elbo)(rng_keys))
 
 
-def get_importance_trace_enum(model, guide, args, kwargs, params, max_plate_nesting):
+def get_importance_trace_enum(
+    model, guide, args, kwargs, params, max_plate_nesting, model_deps
+):
     """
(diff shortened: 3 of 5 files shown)


numpyro/infer/elbo.py
```diff
@@ -932,10 +932,26 @@ def single_particle_elbo(rng_key):
...
                     assert len(group_names) == 1
                     name = next(iter(group_names))
-                    cost = model_trace[name]["log_prob"]
-                    scale = model_trace[name]["scale"]
-                    deps = model_deps[name]
-                    dice_factors = [guide_trace[key]["log_measure"] for key in deps]
+                    try:
+                        cost = -kl_divergence(
+                            guide_trace[name]["fn"], model_trace[name]["fn"]
+                        )
```
ordabayevy, line 938: Maybe move `try/except` logic to `get_importance_trace_enum` to avoid wasteful computation of `log_prob`s?

fehiepsi (commented):
So this one computes KL before marginalizing out discrete variables? I guess the order does not matter.

numpyro/infer/elbo.py
martinjankowiak: are you sure this is correct? what are the conditions under which you can compute the KL? note the conditions i wrote down for the mean field elbo:

```
The precise condition is that for every
    latent variable `z` in the guide, its parents in the model must not include
    any latent variables that are descendants of `z` in the guide. Here
    'parents in the model' and 'descendants in the guide' is with respect
    to the corresponding (statistical) dependency structure. For example, this
    condition is always satisfied if the model and guide have identical
    dependency structures.
```

numpyro/infer/elbo.py
```diff
@@ -728,43 +730,72 @@ def get_importance_trace_enum(model, guide, args, kwargs, params, max_plate_nest
...
-                    intermediates = site["intermediates"]
+                value = site["value"]
+                intermediates = site["intermediates"]
+                dim_to_name = site["infer"]["dim_to_name"]
+
+                # compute log factor
+                if (
+                    is_model
+                    and (name in guide_trace)
+                    and model_deps[name].isdisjoint(sum_vars | guide_desc[name])
```
ordabayevy, line 746: @martinjankowiak can you check this? I think `model_deps[name].isdisjoint(guide_desc[name])` should take care of the condition for analytic kl now.

numpyro/infer/elbo.py
```diff
@@ -728,43 +730,72 @@ def get_importance_trace_enum(model, guide, args, kwargs, params, max_plate_nest
...
+                        if intermediates:
+                            log_prob = site["fn"].log_prob(value, intermediates)
+                        else:
+                            log_prob = site["fn"].log_prob(value)
+                        site["log_prob"] = to_funsor(
+                            log_prob, output=funsor.Real, dim_to_name=dim_to_name
+                        )
+                elif not is_model and (model_trace[name].get("kl", None) is not None):
+                    # skip logq computation if analytic kl was computed
+                    pass
```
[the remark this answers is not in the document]
ordabayevy: @fehiepsi IIUC your question it doesn't compute KL for sites that are being contracted in the model.

numpyro/infer/elbo.py
martinjankowiak: can you remind me of some of the context? does this PR support both model-side and guide-side enumeration? what do i need to know about what's happening on the discrete side?

ordabayevy: Yes, it supports both model-side and guide-side enumeration. It also uses provenance tracking to get the dependency structure of terms in the model on model-enumerated sites. Then this info is used to partition model terms into uncontracted terms and terms that are being contracted together. Then each cost term from model/guide is integrated wrt guide-enumerated sites.

martinjankowiak: do we have any tests that cover the KL branch specifically? more generally what tests do we have to cover the case with continuous latents + TraceEnum?

suppose the model is a K-component gaussian mixture. for model side-enumeration we can't compute the KL because we don't have the formula. for guide-side enumeration with a K-component gaussian mixture guide, we can compute both (i.e. one continuous and one discrete) KL divergences, right?

can we check these two simple cases somehow?

is it the case that model-side enumerations makes most KL divergence computations intractable?

ordabayevy: I will try to write some tests.

numpyro/infer/elbo.py
```diff
@@ -892,32 +923,26 @@ def single_particle_elbo(rng_key):
 
             seeded_model = seed(model, model_seed)
             seeded_guide = seed(guide, guide_seed)
+            model_deps, guide_deps = get_nonreparam_deps(
+                seeded_model, seeded_guide, args, kwargs, param_map
+            )
 
-            model_trace, guide_trace = get_importance_trace_enum(
+            seeded_model = seed(model, model_seed)
+            seeded_guide = seed(guide, guide_seed)
```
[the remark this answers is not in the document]
fehiepsi: yeah, it's hard to reason about this :(

test/infer/test_gradient.py
```diff
@@ -133,3 +133,216 @@ def actual_loss_fn(params_raw):
...
+    probs_z2 = jnp.array([0.5, 0.5])
+    probs_z1 = jnp.array([[0.4, 0.6], [0.6, 0.4]])
+    probs_z3 = jnp.array([[[0.4, 0.6], [0.1, 0.9]], [[0.7, 0.3], [0.2, 0.8]]])
+    z2 = pyro.sample("z2", dist.Categorical(probs_z2))
+    z1 = pyro.sample("z1", dist.Categorical(probs_z1[z2]))
+    pyro.sample("z3", dist.Categorical(probs_z3[z1, z2]))
+
+
+@config_enumerate
+def kl_model_3_z2z3(params):
```
ordabayevy, line 188: For ease of reading the code the suffix of model names has latent sites to which analytic KL can be applied.

ordabayevy (commented):
@martinjankowiak @fehiepsi
- Added tests that compare analytic kl to enumeration. In both cases exact solution is obtained so it is easy to compare. Feel free to add more tests.
- Added `config_kl` function that takes as an argument `sites` set to which analytic kl should be applied
- If `infer={"kl": "analytic"} but analytic kl cannot be applied then an error is raised. So it is strict in a sense that analytic kl is always applied when configured as such.

fehiepsi (approved):
Thanks, Yerdos. LGTM pending comments from @martinjankowiak.

martinjankowiak (commented):
@ordabayevy looks great left some comments!

can we add some tests which also contain continuous latent variables?
for example some of the GMMs i mentioned.

the simplest case is just to check that
```
def model():
    pyro.sample("z0", dist.Normal(0.0, 1.0))
    probs_z1 = jnp.array([0.5, 0.5])
    pyro.sample("z1", dist.Categorical(probs_z1))
```
gives the correct ELBO when computed using a similarly structure guide.

test/infer/test_gradient.py
```diff
@@ -133,3 +133,236 @@ def actual_loss_fn(params_raw):
...
+        (kl_model_3_z2z3, set(["z2", "z3"]), True),
+        (kl_model_4_z3, set(["z1", "z2", "z3"]), False),
+        (kl_model_4_z3, set(["z3"]), True),
+        (kl_model_5_z2, set(["z1", "z2", "z3"]), False),
+        (kl_model_5_z2, set(["z2"]), True),
+        (kl_model_6_, set(["z1", "z2", "z3"]), False),
+        (kl_model_6_, set(), True),
+        (kl_model_7_z3, set(["z1", "z2", "z3"]), False),
+        (kl_model_7_z3, set(["z3"]), True),
+    ][-2:],
```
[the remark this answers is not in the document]
ordabayevy: Done

test/infer/test_gradient.py
```diff
@@ -133,3 +133,236 @@ def actual_loss_fn(params_raw):
...
+    elbo = infer.TraceEnum_ELBO()
+
+    # Expected grads averaged over num_particles
+    def expected_loss_fn(params_raw):
+        params = jax.tree_util.tree_map(transform, params_raw)
+        return elbo.loss(random.PRNGKey(0), {}, model, guide, params)
+
+    expected_loss, expected_grads = jax.value_and_grad(expected_loss_fn)(params_raw)
+
+    # Grads based on exact integration
```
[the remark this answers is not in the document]
ordabayevy: In both cases integration is exact. Expected values are based on enumeration. Actual value are based on the mix of analytic kl and enumeration. Fixed the comments accordingly.

numpyro/handlers.py
```diff
@@ -221,7 +221,7 @@ def process_message(self, msg):
             if guide_msg["type"] != "sample" or guide_msg["is_observed"]:
                 raise RuntimeError(f"Site {name} must be sampled in trace.")
             msg["value"] = guide_msg["value"]
-            msg["infer"] = guide_msg["infer"]
+            msg["infer"] = guide_msg["infer"].copy()
```
ordabayevy, line 224: @martinjankowiak @fehiepsi  Is it okay to copy the `infer` dictionary? Can that possibly mess up other inference algorithms?

I had to create a copy here because computing `dim_to_name` for the site in the model overwrites it in the guide as well. But there can be cases like in new tests in this PR when the guide's `dim_to_name` is different than the one in the model.

numpyro/contrib/funsor/infer_util.py
```diff
@@ -81,6 +81,39 @@ def config_fn(site):
     return infer_config(fn, config_fn)
 
 
+def config_kl(fn=None, sites=None):
+    """
+    Configures analytic kl for all sites in ``sites`` in a NumPyro model.
```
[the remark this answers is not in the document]
ordabayevy: Done

ordabayevy:
@martinjankowiak I'll work on adding more tests that have continuous latent variables.

martinjankowiak:
> "How would we know if it can be computed in principle?" 
i just mean that the graph dependency structure doesn't disallow integration. so maybe a particular site can be integrated out (in principle) if both sites are normal or if both are studentt but we only have the `kl(normal, normal)` pattern in our list of analytic expressions

numpyro/handlers.py
fehiepsi: I think it is fine.

ordabayevy:
@martinjankowiak added a test for a model with continuous latent variable. Does it look alright to you?

> i just mean that the graph dependency structure doesn't disallow integration. so maybe a particular site can be integrated out (in principle) if both sites are normal or if both are studentt but we only have the kl(normal, normal) pattern in our list of analytic expressions

Hmm.. not sure how that would be implemented.

test/infer/test_gradient.py
```diff
@@ -366,3 +366,90 @@ def actual_loss_fn(params_raw):
 
             assert_equal(actual_loss, expected_loss, prec=1e-5)
             assert_equal(actual_grads, expected_grads, prec=1e-5)
+
+
+def kl_model_8_z1z2(params):
```
martinjankowiak, line 371: can we explicitly compare the corresponding ELBO with mean field guide directly to something like `kl_div(Cat, Cat) + kl_div(Norm, Norm)` in the test?

test/infer/test_gradient.py
```diff
@@ -366,3 +366,90 @@ def actual_loss_fn(params_raw):
...
+    probs_z2 = jnp.array([0.0, 0.0])
+    z1 = pyro.sample("z1", dist.Categorical(probs_z1))
+    pyro.sample("z2", dist.Normal(probs_z2[z1], 1.0))
+
+
+@pytest.mark.parametrize(
+    "model",
+    [kl_model_8_z1z2, kl_model_9_z1z2],
+)
+def test_analytic_kl_continuous(model):
```
martinjankowiak, line 398: i'm not entirely sure i understand what this testing. can we include a branch with score function gradients for the discrete latient variable, i.e. no enumeration?

ordabayevy:
@martinjankowiak added new tests:

- models in `test_analytic_kl_2` and `test_analytic_kl_3` have a mixture of enumerated, non-reparam, and reparam (kl) sites but with different dependency structures
- `test_analytic_kl_4` tests `handlers.scale` and `plate`.

test/infer/test_gradient.py
```diff
@@ -368,88 +369,255 @@ def actual_loss_fn(params_raw):
...
         }
-        return elbo.loss(random.PRNGKey(0), {}, model, guide, params)
+        kl_z1 = kl_divergence(
+            dist.Categorical(params["probs_z1"]),
+            dist.Categorical(jnp.array([0.5, 0.5])),
+        )
+        kl_z2 = kl_divergence(
+            dist.Normal(params["probs_z2"]), dist.Normal(jnp.array(0.0))
+        )
+        return scale * (z1_dim * kl_z1 + z2_dim * kl_z2)
```
martinjankowiak, line 610: can we add a test where the scale only hits some sites and not all?

martinjankowiak (approved):

martinjankowiak:
looks great @ordabayevy ! thanks for adding all these detailed tests. would be nice to get that additional `scale` test in here but i think this is basically ready to merge

ordabayevy:
@fehiepsi  @martinjankowiak thanks for reviewing. @martinjankowiak I will just a second scale to that test. Should be easy to do.

martinjankowiak:
@ordabayevy btw i imagine these added tests are pretty fast right?

ordabayevy:
@martinjankowiak on my MacBook Pro they finish in 16s

PR #1535, opened 2023-02-04
Remove redundant `if` clause in `replay`

ordabayevy:
Second if statement is the same as the first one.


fehiepsi (approved):
Thanks, Yerdos!

PR #1536, opened 2023-02-04
Raise an error if there is no common scale when model enumerated

ordabayevy:
It turns out that current subsample scaling logic is not handled correctly in `TraceEnum_ELBO` (sorry for that). I added some tests to demonstrate that. I think the best place to handle subsampling scaling is in `funsor.sum_product.sum_product`. For now I added a check for a common scale which raises an error if there is no common scale.

### DIFF
--- a/numpyro/infer/elbo.py
+++ b/numpyro/infer/elbo.py
@@ -970,7 +970,7 @@ class TraceEnum_ELBO(ELBO):
                         [
                             value
                             for plate, value in plate_to_scale.items()
-                            if (plate in f.inputs) or (plate is None)
+                            if plate in f.inputs
                         ]
                         for f in group_factors
                     )
@@ -986,7 +986,7 @@ class TraceEnum_ELBO(ELBO):
                         plates=group_plates,
                         eliminate=group_sum_vars | elim_plates,
                     )
-                    scale = None
+                    scale = plate_to_scale.get(None, None)
                     # combine deps
                     deps = frozenset().union(
                         *[model_deps[name] for name in group_names]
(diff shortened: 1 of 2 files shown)


numpyro/infer/elbo.py
```diff
@@ -960,24 +960,18 @@ def single_particle_elbo(rng_key):
...
-                                    )
-                            else:
-                                group_scales[plate] = value
-                    scale = (
-                        reduce(lambda a, b: a * b, group_scales.values())
-                        if group_scales
-                        else None
+                    scales_set = set(
+                        [
+                            model_trace[name]["scale"]
```
fehiepsi, line 965: I'm not sure how to check the uniqueness of jax array scales because jax arrays is not hashable. Could you replace this by something like
```
scale = None
for name in (group_names | group_sum_vars):
    site_scale = model_trace[name]["scale"]
    if isinstance(scale, (int, float)) and isinstance(site_scale, (int, float)) and (site_scale != scale):
        raise ValueError(...)
    scale = site_scale
```

ordabayevy: I copied the implementation from Pyro which first checks that array is scalar, converts it to scalar, and then compares them. Does it look alright to you?

https://github.com/pyro-ppl/pyro/blob/dev/pyro/infer/traceenum_elbo.py#L37-L41

> Btw, does this mean that we don't support enumeration for models with both global and local variables?

Yeah, if you enumerate a global variable than you cannot subsample a local variable that depends on it.
I'm working on it but that will require first changing `funsor.sum_product.sum_product` to handle plate-wise scaling there.

fehiepsi: Because we can't convert a tracer to float, `float(site_scale)` will fail when scale is a tracer. Maybe we need to raise error if it is a jnp.array?

ordabayevy: I see. Changed it so that it raises error if scale is `jnp.array`.

fehiepsi (approved):
LGTM. Thanks, Yerdos.

ordabayevy:
Thanks for reviewing @fehiepsi .

PR #1537, opened 2023-02-08
Properly handle contraction of guide plates in `TraceEnum_ELBO`

ordabayevy:
There was a bug in `TraceEnum_ELBO` demonstrated in a new `test_enum_elbo::test_guide_plate_contraction` test. When a cost term depends on a non-reparameterizable guide site with extra plate dims they need to be product-contracted. Instead these plate dims were passed to `_eager_contract_tensors` as `reduced_vars` and sum-contracted. So I replaced `_eager_contract_tensors` with `funsor.sum_product.sum_product` which can do plated sum-products and eliminate plates.

### DIFF
--- a/test/contrib/test_enum_elbo.py
+++ b/test/contrib/test_enum_elbo.py
@@ -2477,20 +2477,20 @@ def test_model_enum_subsample_3(scale):
 def test_guide_plate_contraction():
     def model(params):
         with pyro.plate("a_axis", size=2):
-            a = pyro.sample("a", dist.Categorical(jnp.array([0.2, 0.8])))
+            a = pyro.sample("a", dist.Poisson(jnp.array(3.0)))
         pyro.sample("b", dist.Normal(jnp.sum(a), 1.0), obs=1)
 
     def guide(params):
         probs_a = pyro.param(
-            "probs_a", params["probs_a"], constraint=constraints.simplex
+            "probs_a", params["probs_a"], constraint=constraints.positive
         )
         with pyro.plate("a_axis", size=2):
-            pyro.sample("a", dist.Categorical(probs_a))
+            pyro.sample("a", dist.Poisson(probs_a))
 
     params = {
-        "probs_a": jnp.array([[0.4, 0.6], [0.7, 0.3]]),
+        "probs_a": jnp.array([3.0, 2.5]),
     }
-    transform = dist.biject_to(dist.constraints.simplex)
+    transform = dist.biject_to(dist.constraints.positive)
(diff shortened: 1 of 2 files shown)


test/contrib/test_enum_elbo.py
```diff
@@ -2472,3 +2472,44 @@ def actual_loss_fn(params_raw):
 
         assert_equal(actual_loss, expected_loss, prec=1e-3)
         assert_equal(actual_grads, expected_grads, prec=1e-5)
+
+
+def test_guide_plate_contraction():
+    def model(params):
+        with pyro.plate("a_axis", size=2):
+            a = pyro.sample("a", dist.Categorical(jnp.array([0.2, 0.8])))
+        pyro.sample("b", dist.Normal(jnp.sum(a), 1.0), obs=1)
```
fehiepsi, line 2481: It seems to me that this invalidates this restriction: https://pyro.ai/examples/enumeration.html#Restriction-2:-no-downstream-coupling? Could you elaborate why we can enumerate here?

ordabayevy: It's not being enumerated here but being used as a non-reparameterizable site. I should have used other distribution like Poisson to make it less confusing :)

But the point is that `b` depends on a non-reparametrizable site `a` which has `a_axis` plate. The `dice_factor` needs to be product-contracted to eliminate the extra `a_axis` plate before multiplying the cost term for site `b`. Instead `a_axis` is being passed to `_eager_contract_tensors` as `reduced_vars` and sum-contracted with `logsumexp`.

fehiepsi: I see. So we are using TraceEnum_ELBO but enumeration is disabled for those cases. Could you add a warning for this?

ordabayevy: Do you mean in general or for this test?

fehiepsi: I mean in general. I think using TraceEnum_ELBO without enumeration is confusing. Maybe raise error if we can't enumerate sites with `infer={"enumerate": "parallel"}`?

ordabayevy: I changed the distribution to Poisson in the test since it can be any non-reparameterizable distributions. I can open another issue/PR for enumeration configuration since it is a separate issue.

fehiepsi: sounds reasonable to me, thanks!

fehiepsi (approved):

ordabayevy:
Thanks for reviewing @fehiepsi !

PR #1543, opened 2023-02-28
Fix provenance for jax 0.4.4

### LINKED ISSUE #1542: `NameError: unbound axis name: _provenance` with the latest jax version
I'm getting the error message below when running `pytest test/contrib/test_enum_elbo.py` with the latest version of jax (0.4.4):

```
self = Traced<ShapedArray(int32[1;_provenance:frozenset({'w'})])>with<DynamicJaxprTrace(level=4/0)>, other = 0

    def deferring_binary_op(self, other):
      if hasattr(other, '__jax_array__'):
        other = other.__jax_array__()
      args = (other, self) if swap else (self, other)
      if isinstance(other, _accepted_binop_types):
>       return binary_op(*args)
E       NameError: unbound axis name: _provenance. The following axis names (e.g. defined by pmap) are available to collective operations: []

/home/yordabay/anaconda3/envs/numpyro/lib/python3.8/site-packages/jax/_src/numpy/lax_numpy.py:5071: NameError
```

fehiepsi:
Fixes #1542

After the merge of jit & pjit in jax 0.4.4, our provenance trace is no longer composed with jitted primitives (like jnp.sin), so we need to play with jaxpr, which does not involve jitting.

### DIFF
--- a/numpyro/infer/elbo.py
+++ b/numpyro/infer/elbo.py
@@ -560,12 +560,15 @@ def _substitute_nonreparam(data, msg):
         return value
 
 
-def _get_latents(guide, args, kwargs, params):
+def _get_latents(model, guide, args, kwargs, params):
+    model = seed(substitute(model, data=params), rng_seed=0)
     guide = seed(substitute(guide, data=params), rng_seed=0)
     guide_tr = trace(guide).get_trace(*args, **kwargs)
+    model_tr = trace(replay(model, guide_tr)).get_trace(*args, **kwargs)
+    model_tr.update(guide_tr)
     return {
         name: site["value"]
-        for name, site in guide_tr.items()
+        for name, site in model_tr.items()
         if site["type"] == "sample" and not site.get("is_observed", False)
     }
 
@@ -573,7 +576,9 @@ def _get_latents(guide, args, kwargs, params):
 def get_nonreparam_deps(model, guide, args, kwargs, param_map, latents=None):
     """Find dependencies on non-reparameterizable sample sites for each cost term in the model and the guide."""
     if latents is None:
-        latents = eval_shape(partial(_get_latents, guide, args, kwargs, param_map))
+        latents = eval_shape(
+            partial(_get_latents, model, guide, args, kwargs, param_map)
+        )
 
     def fn(**latents):
         subs_fn = partial(_substitute_nonreparam, latents)
--- a/numpyro/infer/elbo.py
+++ b/numpyro/infer/elbo.py
@@ -561,13 +561,14 @@ def _substitute_nonreparam(data, msg):
 
 
 def _get_latents(model, guide, args, kwargs, params):
-    model = seed(model, rng_seed=0)
-    guide = seed(guide, rng_seed=0)
-    model_tr, guide_tr = get_importance_trace(model, guide, args, kwargs, params)
-    guide_tr.update(model_tr)
+    model = seed(substitute(model, data=params), rng_seed=0)
+    guide = seed(substitute(guide, data=params), rng_seed=0)
+    guide_tr = trace(guide).get_trace(*args, **kwargs)
+    model_tr = trace(replay(model, guide_tr)).get_trace(*args, **kwargs)
+    model_tr.update(guide_tr)
     return {
         name: site["value"]
-        for name, site in guide_tr.items()
+        for name, site in model_tr.items()
         if site["type"] == "sample" and not site.get("is_observed", False)
     }
 
--- a/numpyro/infer/elbo.py
+++ b/numpyro/infer/elbo.py
@@ -560,15 +560,12 @@ def _substitute_nonreparam(data, msg):
         return value
 
 
-def _get_latents(model, guide, args, kwargs, params):
-    model = seed(substitute(model, data=params), rng_seed=0)
+def _get_latents(guide, args, kwargs, params):
     guide = seed(substitute(guide, data=params), rng_seed=0)
     guide_tr = trace(guide).get_trace(*args, **kwargs)
-    model_tr = trace(replay(model, guide_tr)).get_trace(*args, **kwargs)
-    model_tr.update(guide_tr)
     return {
         name: site["value"]
-        for name, site in model_tr.items()
+        for name, site in guide_tr.items()
         if site["type"] == "sample" and not site.get("is_observed", False)
     }
 
@@ -576,9 +573,7 @@ def _get_latents(model, guide, args, kwargs, params):
 def get_nonreparam_deps(model, guide, args, kwargs, param_map, latents=None):
     """Find dependencies on non-reparameterizable sample sites for each cost term in the model and the guide."""
     if latents is None:
-        latents = eval_shape(
-            partial(_get_latents, model, guide, args, kwargs, param_map)
-        )
+        latents = eval_shape(partial(_get_latents, guide, args, kwargs, param_map))
 
     def fn(**latents):
         subs_fn = partial(_substitute_nonreparam, latents)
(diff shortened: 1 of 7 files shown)


fehiepsi:
@ordabayevy I couldn't figure out how to make `track_nonreparam` work for the latest jax version. Do you think that we can refactor it so that it does not involve `named_shape` manipulation?

ordabayevy:
@fehiepsi not sure how that would be done at the moment but I'll have a closer look at it later today.

fehiepsi:
No worries, I think we just need to introduce a new handler with proper logic. Let me take a stab at that.

ordabayevy:
@fehiepsi Let me know if you want to pair code on this. I had a stab at it yesterday but got very strange behavior with provenance. The approach I tried was to introduce a new jax primitive to attach provenance instead of hacking `named_shape`.

```py
from jax import core
track_p = core.Primitive("track")
def track_provenance(x, provenance):
   return track_p.bind(x, provenance=provenance)
...

class track_nonreparam(Messenger):
    ...
    msg["value"] = track_provenance(msg["value"], frozenset([msg["name"]))
```

fehiepsi:
Hi @ordabayevy, thanks, I'm happy to pair code. I talked with some jax devs and it is recommended to use custom interpreter https://jax.readthedocs.io/en/latest/notebooks/Writing_custom_interpreters_in_Jax.html I'll send you an invitation

ordabayevy:
`eval_provenance` interpreter pair coded with @fehiepsi 

```py
from jax.linear_util import wrap_init

def eval_provenance(f, **kwargs):
  # flatten the function and its arguments
  args_flat, in_tree = jax.tree_util.tree_flatten(((), kwargs))
  wrapped_fun, out_tree = jax.api_util.flatten_fun(wrap_init(f), in_tree)
  
  closed_jaxpr = jax.make_jaxpr(wrapped_fun.call_wrapped)(*args_flat)
  jaxpr = closed_jaxpr.jaxpr
  print(jaxpr)
  # Mapping from variable -> value
  env = {}
  provenance = {v: frozenset({n}) for v, n in zip(jaxpr.invars, kwargs.keys())}
  
  def read(var):
    # Literals are values baked into the Jaxpr
    if type(var) is core.Literal:
      return var.val
    return env[var]

  def read_provenance(var):
    # Literals are values baked into the Jaxpr
    if type(var) is core.Literal:
      return frozenset()
    return provenance[var]

  def write(var, val):
    env[var] = val

  # Bind args and consts to environment
  safe_map(write, jaxpr.invars, kwargs.values())
  # safe_map(write, jaxpr.constvars, consts)

  # Loop through equations and evaluate primitives using `bind`
  for eqn in jaxpr.eqns:
    # compute provenance
    inprovenance = safe_map(read_provenance, eqn.invars)
    outprovenance = frozenset().union(*inprovenance)
    for var in eqn.outvars:
      provenance[var] = outprovenance
    # Read inputs to equation from environment
    invals = safe_map(read, eqn.invars)
    # `bind` is how a primitive is called
    outvals = eqn.primitive.bind(*invals, **eqn.params)
    # Primitives may return multiple outputs or not
    if not eqn.primitive.multiple_results: 
      outvals = [outvals]
    # Write the results of the primitive into the environment
    safe_map(write, eqn.outvars, outvals) 
  # Read the final result of the Jaxpr from the environment
  out = [jax.ShapeDtypeStruct(env[v].shape, env[v].dtype, {"provenance": provenance[v]}) for v in jaxpr.outvars]
  output = jax.tree_util.tree_unflatten(out_tree(), out)
  return jax.tree_util.tree_map(lambda x: x.named_shape["provenance"], output)
```

ordabayevy:
Hey @fehiepsi . How about adding a new jax primitive for attaching provenance:

```py
from jax import core
track_p = core.Primitive("track")
track_p.def_impl(lambda x, provenance: jnp.array(x))
track_p.def_abstract_eval(lambda x, provenance: x)
def track_provenance(x, provenance):
   return track_p.bind(x, provenance=provenance)
```

Above we only need to change to this:

```py
  for eqn in jaxpr.eqns:
    # compute provenance
    inprovenance = safe_map(read_provenance, eqn.invars)
    outprovenance = frozenset().union(*inprovenance)
    for var in eqn.outvars:
      # track provenance here
      if eqn.primitive is track_p:
        outprovenance |= eqn.params["provenance"]
      provenance[var] = outprovenance
```

Example:

```py
def f(a, b, c):
  a = track_provenance(a, frozenset("z"))
  x = jax.lax.fori_loop(0, 10, lambda i, x: x+c, a)
  return jax.scipy.special.xlogy(a, b), jax.jit(jnp.sin)(c), {"x": a*c}, x, jax.value_and_grad(lambda x: x ** 2)(b), jax.vmap(jnp.tanh)(a + b + jnp.ones(3))

eval_provenance(f, a=3, b=5.0, c=2)
```

Outputs:

```
(frozenset({'a', 'b', 'z'}),
 frozenset({'c'}),
 {'x': frozenset({'a', 'c', 'z'})},
 frozenset({'a', 'c', 'z'}),
 (frozenset({'b'}), frozenset({'b'})),
 frozenset({'a', 'b', 'z'}))
```


Then I think we can just change our `track_nonreparam` and keep everything else in `elbo.py` as it was before:

```py
class track_nonreparam(Messenger):
    def postprocess_message(self, msg):
        if (
            msg["type"] == "sample"
            and (not msg["is_observed"])
            and (not msg["fn"].has_rsample)
        ):
            msg["value"] = track_provenance(msg["value"], frozenset([msg["name"]))
```

What do you think?

fehiepsi:
I'm a bit worried if we need to add extra rules for the primitive? Assuming that our provenance logic happens inside a vmapped function, I guess we need to define batching rule etc.

ordabayevy:
Ah okay, that is probably true. Let's stick to the current implementation.

fehiepsi:
@ordabayevy I split out the process of abstract evaluation and provenance tracking because it's not clear to me know to do abstract evaluation for high order primitives like call_p, xla_call_p, xla_pmap_p, closed_call_p, pjit_p without touching too many internal jax functionalities. We didn't face this issue if we perform actual calculation though `eqn.bind` (our tests passed during the pair code session)

test/ops/test_provenance.py
```diff
@@ -1,21 +1,86 @@
...
+        return x + y
+
+    jaxpr = jax.make_jaxpr(f)(jnp.zeros((3, 4), jnp.float32))
+    assert len(jaxpr.consts) == 1
+    assert eval_provenance(f, x=3) == {"x"}
+
+
+def test_provenance_fori():
+    def f(x, y, z):
+        del z
```
ordabayevy, line 48: Why do you have to delete `z` here?

ordabayevy (approved):
This looks awesome @fehiepsi , thanks for fixing it! I just left a few comments.

test/ops/test_provenance.py
fehiepsi: we don't use z in this function. it is just a good practice i think.

numpyro/infer/elbo.py
```diff
@@ -610,6 +553,38 @@ def get_importance_log_probs(model, guide, args, kwargs, params):
...
+def _substitute_nonreparam(data, msg):
+    if msg["name"] in data and not msg["fn"].has_rsample:
+        value = msg["fn"](*msg["args"], **msg["kwargs"])
+        value = 0 * value + data[msg["name"]]
+        return value
+
+
+def _get_latents(guide, args, kwargs, params):
+    guide = seed(substitute(guide, data=params), rng_seed=0)
+    guide_tr = trace(guide).get_trace(*args, **kwargs)
```
ordabayevy, line 565: I think unlike `TraceGraph_ELBO` in `TracEnum_ELBO` we need latents from the model trace as well because it can have model enumerated variables.

ordabayevy (approved):
LGTM! Tests are passing :tada::tada:

fehiepsi:
Thanks a lot for looking at this issue, Yerdos!

PR #1550, opened 2023-03-05
Add init to mean strategy

fehiepsi:
Based on @vitkl work at #1031.


numpyro/infer/initialization.py
```diff
@@ -45,6 +45,36 @@ def init_to_median(site=None, num_samples=15):
             return init_to_uniform(site)
 
 
+def init_to_mean(site=None):
+    """
+    Initialize to the prior mean. For priors with no `.sample` method implemented,
```
martinjankowiak, line 50: do you mean `.mean`? perhaps it'd be better to do the backoff sequence
mean -> sample(num_samples).mean() -> init_to_uniform

fehiepsi: good idea! I just made the change.

martinjankowiak (approved):

PR #1551, opened 2023-03-05
Add Gompertz distribution

fehiepsi:
Based on the great work of @jackpot-nfer at #1127.


numpyro/distributions/continuous.py
```diff
@@ -668,6 +669,68 @@ def cdf(self, x):
         return 1 - self.base_dist.cdf(1 / x)
 
 
+class Gompertz(Distribution):
+    """Gompertz Distribution sampling utilities.
```
OlaRonning, line 673: Use raw string to avoid `\\` in math mode.

numpyro/distributions/continuous.py
```diff
@@ -668,6 +669,68 @@ def cdf(self, x):
...
 
+class Gompertz(Distribution):
+    r"""Gompertz Distribution sampling utilities.
+
+    This implementation is based off the derivations from Wikipedia page for the
+    Gompertz distribution. See https://en.wikipedia.org/wiki/Gompertz_distribution.
+
+    However, we call the parameter "eta" a concentration parameter.
+    We also let the user specify a rate parameter instead of the scale parameter "b".
+    The two are related by rate = 1 / scale. By default, `rate` = 1.
```
martinjankowiak, line 680: this seems to be consistent with wikipedia but wikipedia doesn't make a whole lot of sense here since usually the `scale` parameter should have units of `x`. compare to e.g. usage of 'scale' in https://en.wikipedia.org/wiki/Exponential_distribution

martinjankowiak: sounds good

martinjankowiak (dismissed):
lgtm just left suggested changes for docstring

numpyro/distributions/continuous.py
```diff
@@ -668,6 +669,62 @@ def cdf(self, x):
         return 1 - self.base_dist.cdf(1 / x)
 
 
+class Gompertz(Distribution):
+    r"""Gompertz Distribution sampling utilities.
```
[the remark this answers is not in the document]
fehiepsi: done

numpyro/distributions/continuous.py
```diff
@@ -668,6 +669,62 @@ def cdf(self, x):
...
+
+    This implementation is based off the derivations from Wikipedia page for the
```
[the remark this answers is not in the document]
fehiepsi: done

numpyro/distributions/continuous.py
```diff
@@ -668,6 +669,62 @@ def cdf(self, x):
...
+        )
+
+    def cdf(self, value):
+        return -jnp.expm1(-self.concentration * jnp.expm1(value * self.rate))
+
+    def icdf(self, q):
+        return jnp.log1p(-jnp.log1p(-q) / self.concentration) / self.rate
+
+    @property
+    def mean(self):
```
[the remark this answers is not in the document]
fehiepsi: That's right. We only skip the variance test.

numpyro/distributions/continuous.py
fehiepsi: done

fehiepsi:
Thanks, Martin!

PR #1553, opened 2023-03-08
SVI shared parameter initialization

### LINKED ISSUE #1552: [FR] SVI: avoid initializing model params when already specified in guide
(Context: https://forum.pyro.ai/t/sharing-params-between-model-guide-i-e-tied-weights/5033/)

Sharing parameter values between a model and its guide is helpful in some applications, e.g. symmetric autoencoders. This can *almost* be accomplished by defining a param as usual in the guide and calling `numypro.param("shared_site_name")` in the model, but [since `SVI` only replays sample values](https://github.com/pyro-ppl/numpyro/blob/737d7d9ec0dc5d6649f85bbedddf28ca07908e20/numpyro/infer/svi.py#L186-L187) and not parameter values before tracing the model, the return value of this `param()` call defaults to `None` which breaks any downstream code that uses it.

The [desired behavior in SVI is to avoid initializing any model parameters already specified in the guide](https://forum.pyro.ai/t/sharing-params-between-model-guide-i-e-tied-weights/5033/6). I can see a couple of ways of accomplishing this:

1. Explicitly substitute initialized parameter values from the guide before tracing the model. Here's a strawman implementation which uses `substitute()` to avoid changing the semantics of `replay()`: https://github.com/pyro-ppl/numpyro/compare/master...msegado:numpyro:feat-shared-param-init.
2. Change the `replay()` handler to support replaying param values. This should probably be opt-in via an `include_params` kwarg or similar to avoid breaking existing user code, and would also require a change to the docs.

The first seems like the more natural approach to me, but I figured it would be good to ask before opening a PR 🙂 Thoughts?

msegado:
Closes #1552

### DIFF
--- a/numpyro/infer/svi.py
+++ b/numpyro/infer/svi.py
@@ -21,7 +21,7 @@ from jax.tree_util import tree_map
 
 from numpyro.distributions import constraints
 from numpyro.distributions.transforms import biject_to
-from numpyro.handlers import replay, seed, trace
+from numpyro.handlers import replay, seed, substitute, trace
 from numpyro.infer.util import helpful_support_errors, transform_fn
 from numpyro.optim import _NumPyroOptim, optax_to_numpyro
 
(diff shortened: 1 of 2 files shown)


msegado:
Very welcome - thanks for the great PPL 🙂

PR #1558, opened 2023-03-19
Allow pickled mcmc object to run post warmup phase

fehiepsi:
The issue is reported in Pyro forum https://forum.pyro.ai/t/error-loading-and-running-a-pickled-mcmc-object/4956


OlaRonning (approved):
LGTM

PR #1561, opened 2023-03-24
Add `init_params` argument to `svi.init()` and `svi.run()`

### LINKED ISSUE #1560: [FR] Run SVI with partial `init_state`
I'm trying to port the [GMM example](https://pyro.ai/examples/gmm.html) to NumPyro. In that tutorial first the `global_guide` is trained which is then used as part of the `full_guide`:

<details>

```py
@config_enumerate
def full_guide(data):
    # Global variables.
    with poutine.block(hide_types=["param"]):  # Keep our learned values of global parameters.
        global_guide(data)

    # Local variables.
    with pyro.plate('data', len(data)):
        assignment_probs = pyro.param('assignment_probs', torch.ones(len(data), K) / K,
                                      constraint=constraints.unit_interval)
        pyro.sample('assignment', dist.Categorical(assignment_probs))
```

</details>

I want to be able to reuse pre-trained parameters from a smaller model (`global_guide`) in a larger model (`full_guide`), something like this:

```py
# train global_guide
global_svi = SVI(model, global_guide, optim, loss=elbo)
global_svi_result = global_svi.run(random.PRNGKey(0), 200, data)

# train full_guide using pre-trained global_guide
# however global_svi_result.state is only a partial state, i.e. it doesn't have all of full_guide parameters
full_svi = SVI(model, full_guide, optim, loss=elbo)
full_svi_result = full_svi.run(random.PRNGKey(0), 200, init_state=global_svi_result.state)
```

However, that would require initializing SVI with partial `init_state`. There might be other use cases for this feature.

ordabayevy:
Closes #1560.

### DIFF
--- a/numpyro/infer/svi.py
+++ b/numpyro/infer/svi.py
@@ -169,7 +169,7 @@ class SVI(object):
 
             self.optim = optax_to_numpyro(optim)
 
-    def init(self, rng_key, *args, **kwargs):
+    def init(self, rng_key, init_params=None, *args, **kwargs):
         """
         Gets the initial SVI state.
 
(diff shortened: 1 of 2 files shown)


fehiepsi (approved):
Thanks Yerdos!

PR #1564, opened 2023-03-27
Add inference utilities to transform between unconstrained and constrained space

### LINKED ISSUE #1554: Inverse bjiector transformation (from constrained to unconstrained space)
Hi,
Is it possible, with the current interface, to design a function that takes as input the pytree of _constrained_ parameter values and return _unconstrained_ values, given a model with Pyro primitives? For instance, I am looking for a function that does the inverse transform of `constrain_fn()` from `numpyro.infer.util`.

Sorry to ask here but I could not find the answer in the code. So maybe there is no easy implementation of it yet.

aymgal:
This is a proposal to fix #1554

### DIFF
--- a/numpyro/infer/util.py
+++ b/numpyro/infer/util.py
@@ -181,6 +181,10 @@ def constrain_fn(model, model_args, model_kwargs, params, return_deterministic=F
             if site["type"] == "sample":
                 with helpful_support_errors(site):
                     return biject_to(site["fn"].support)(params[site["name"]])
+            elif site["type"] == "param":
+                constraint = site["kwargs"].pop("constraint", constraints.real)
+                with helpful_support_errors(site):
+                    return biject_to(constraint)(params[site["name"]])
             else:
                 return params[site["name"]]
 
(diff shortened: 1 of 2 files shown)


numpyro/infer/util.py
```diff
@@ -193,6 +193,70 @@ def substitute_fn(site):
...
+    from unconstrained space to constrained space.
+
+    :param model: a callable containing NumPyro primitives.
+    :param tuple model_args: args provided to the model.
+    :param dict model_kwargs: kwargs provided to the model.
+    :param dict params: dictionary of unconstrained values keyed by site
+        names.
+    :return: `dict` of transformation keyed by site names.
+    """
+    transforms = get_transforms(model, model_args, model_kwargs, params)
```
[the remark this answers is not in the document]
aymgal: Indeed you're right. Let me know if your points have been addressed by the latest commit :)

numpyro/infer/util.py
```diff
@@ -157,7 +157,8 @@ def transform_fn(transforms, params, invert=False):
     return {k: transforms[k](v) if k in transforms else v for k, v in params.items()}
 
 
-def constrain_fn(model, model_args, model_kwargs, params, return_deterministic=False):
+def constrain_fn(model, model_args, model_kwargs, params, 
+                 include_param_sites=False, return_deterministic=False):
```
fehiepsi, line 161: I think you can remove those `include_param_sites` flags and keep the True behavior by default.

aymgal: This was for ensuring backward compatibility. I removed these in the last comit

numpyro/infer/util.py
```diff
@@ -193,6 +197,39 @@ def substitute_fn(site):
...
+    :param model: a callable containing NumPyro primitives.
+    :param tuple model_args: args provided to the model.
+    :param dict model_kwargs: kwargs provided to the model.
+    :param dict params: dictionary of values keyed by site names.
+    :return: `dict` of transformation keyed by site names.
+    """
+    substituted_model = substitute(model, data=params)
+    transforms, _, _, _ = _get_model_transforms(substituted_model, model_args, model_kwargs)
+    return transforms
+
```
fehiepsi, line 215: Could you run `make format` to fix lint issue? I guess you need to a new line here

numpyro/infer/util.py
fehiepsi: I see. I guess it won't affect the old behavior. If users provide param sites, your solution will return expected results, while the current master branch will raise an error - so this is an improvement.

aymgal:
@fehiepsi fyi I rebased the commits of this PR as I had to do lots of syntax fixing recently... should be ok now

fehiepsi:
Thanks @aymgal! It's great to have this utility available.

fehiepsi (approved):

Issue #1569, opened 2023-03-28
Rendering of mathcal in docs


theorashid:
Weird issue: \mathcal{N} is not rendering in the docs on my browser. I don't know if this is a numpyro issue. For [the webpage from the docs](https://num.pyro.ai/en/stable/examples/mortality.html)

From chrome:
<img width="944" alt="Screenshot 2023-03-29 at 00 03 19" src="https://user-images.githubusercontent.com/30835680/228386261-433ee997-86d7-4256-91dd-828f9d00747a.png">

From safari (it works):
<img width="944" alt="Screenshot 2023-03-29 at 00 03 52" src="https://user-images.githubusercontent.com/30835680/228386336-78c872c1-4d1d-4bf2-b669-480de9a41167.png">

If this problem is specific to my machine and there's no quick fix, feel free to close the issue immediately.

fehiepsi:
thanks for pointing out. I think we need to replace those matcal symbols by something else.

fehiepsi:
Hi @theorashid, thanks for offering. Here are the affected tutorials:
+ https://num.pyro.ai/en/stable/tutorials/bayesian_hierarchical_linear_regression.html
+ https://num.pyro.ai/en/stable/tutorials/tbip.html
+ https://num.pyro.ai/en/stable/examples/mortality.html

I think it is better to use explicit distribution names at those places. :)

theorashid:
So you'd just like me to replace e.g. $\mathcal{N}(0,1)$ with $\text{Normal}(0, 1)$?

fehiepsi:
Yeah, I think it would work

theorashid:
Closed by https://github.com/pyro-ppl/numpyro/pull/1571

PR #1571, opened 2023-03-30
Fix the three tutorials affected by \mathcal rendering

theorashid:
This is for #1569

Replaced by \text{Normal} and \text{Half-Normal}. Have a quick skim of the diff to check, but should be straightforward


fehiepsi (approved):
Thanks for the quick fix, @theorashid!

PR #1574, opened 2023-04-09
Do not mutate shapes of ExpandedDistribution for map-free ops

### LINKED ISSUE #1573: Unflattening `ExpandedDistribution` can unnecessarily mutate it's base distribution's batch shape
While working on fixing the failing tests in #1529, I realized that the operations in the flattening/unflattening routines of `ExpandedDistribution` add additional items to the `batch_shape` of the `base_dist` attribute of `ExpandedDistribution` instances. This mutation allows to return mapped  `ExpandedDistribution`, as discussed in #684.

However, this mutation is currently performed even if the flattened `ExpandedDistribution` did not go through any mapping operation. Consequently, the following snippet - which compares the output of a function returning an `ExpandedDistribution` with the one returned by its jitted equivalent - fails:

```python
import jax
import jax.numpy as jnp

def bs(arg):
    return dist.Normal(arg, arg).expand([10, 3, *arg.shape])


d = bs(jnp.array(1.0))
dj = jax.jit(bs)(jnp.array(1.0))

assert isinstance(d, dist.ExpandedDistribution)
assert isinstance(dj, dist.ExpandedDistribution)
assert d.base_dist.batch_shape == dj.base_dist.batch_shape
```

This issue can be fixed by inspecting, at unflattening-time, whether any mapping operation was performed on the base distribution.

pierreglaser:
Fixes #1573.

### DIFF
--- a/numpyro/distributions/distribution.py
+++ b/numpyro/distributions/distribution.py
@@ -613,15 +613,27 @@ class ExpandedDistribution(Distribution):
             self.base_dist,
         )
         base_flatten, base_aux = base_dist.tree_flatten()
-        return base_flatten, (type(self.base_dist), base_aux, self.batch_shape)
+        return base_flatten, (
+            type(self.base_dist),
+            base_aux,
+            self.batch_shape,
+            prepend_ndim,
+        )
 
     @classmethod
     def tree_unflatten(cls, aux_data, params):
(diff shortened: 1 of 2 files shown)


fehiepsi (approved):
Looks great to me! I'm sorry for the very slow response!

PR #1575, opened 2023-04-11
Jittable transforms

pierreglaser:
Standalone PR that will vastly ease the flattening/unflattening logic of `TransformedDistribution`s that will eventually get implemented in #1529 or one of its follow-ups.

cc @fehiepsi.


numpyro/distributions/constraints.py
```diff
@@ -202,8 +224,17 @@ def __call__(self, x=None, *, is_discrete=NotImplemented, event_dim=NotImplement
             event_dim = self._event_dim
         return _Dependent(is_discrete=is_discrete, event_dim=event_dim)
 
+    def tree_flatten(self):
+        return (), ((), dict())
+
+    @classmethod
+    def tree_unflatten(cls, aux_data, params):
+        return cls()
```
fehiepsi, line 232: I guess this is unnecessary given that you subclass ParameterFreeConstraint.

numpyro/distributions/transforms.py
```diff
@@ -1021,6 +1120,13 @@ def forward_shape(self, shape):
     def inverse_shape(self, shape):
         raise NotImplementedError
 
+    def tree_flatten(self):
+        # XXX: what if unpack_fn is a parametrized callable pytree?
```
fehiepsi, line 1124: I guess this is fine. I don't think we have usage case where unpack_fn depends on extra jax arrays.

fehiepsi (dismissed):
Thanks so much @pierreglaser! I just have a small nit. Super nice PR!!!

numpyro/distributions/constraints.py
pierreglaser: the `_Dependent` currently does not subclass `ParameterFreeConstraint` for now. However, taking a look back at this, I why I didn't pass `event_dim` and `_is_discrete` in the auxiliary data tuple when flattening. I could add a small test for it.

numpyro/distributions/transforms.py
pierreglaser: Agreed. Would you like me to downgrade this `#XXX`  to a simple `# note`, or to remove it entirely?

pierreglaser:
Thanks for the review @fehiepsi! I'm currently in the process of extending #1529 to all distributions, I'm around 20% there :-)

numpyro/distributions/constraints.py
fehiepsi: Oops, sorry. I thought those methods belong to _SingletonConstraint class. Yeah, you're right, we can add those attributes to the auxiliary tuple.

fehiepsi:
Supporting all distributions is complicated I would say. But with jittable support/transform, things might be easier, hopefully. Thanks for working on it!!

fehiepsi:
Hi @pierreglaser, I would like to make a release this week. Do you want to have this PR in?

numpyro/distributions/transforms.py
fehiepsi: I think just leave this as is.

pierreglaser:
Yes, let me finish the PR today and we can merge this.

pierreglaser:
Ok @fehiepsi, I added a few more test to make sure that one could vmap over constraint and transforms, and as a side effect, I modified a bit the `__eq__` methods of these objects to make them robust to inputs containing tracers. I think this PR is good to go!

numpyro/distributions/constraints.py
pierreglaser: Done!

pierreglaser:
I think the linting failure is not coming from this PR....

numpyro/distributions/constraints.py
```diff
@@ -243,8 +280,16 @@ def __repr__(self):
...
-class _Positive(_GreaterThan, _SingletonConstraint):
+    def __eq__(self, other):
+        return isinstance(other, _GreaterThan) & jnp.array_equal(
+            self.lower_bound, other.lower_bound
+        )
+
+
+class _Positive(_SingletonConstraint, _GreaterThan):
     def __init__(self):
         super().__init__(0.0)
```
fehiepsi, line 294: I'm not sure. I feel that this call will not set lower_bound to 0 after you switch the order of parent class. What's the motivation for that? I guess it causes the lint issue.

pierreglaser: I think it would, since `_SingletonConstraints` just grabs the `positive` object from the global namespace during `__new__` (and my [tests](https://github.com/pierreglaser/numpyro/blob/8ba77674f7a07d479c8c98290e46cb1f4f5334ab/test/test_constraints.py#L23) should test for this). The tests were passing even with that change in, what caused the doc failing was the last 2 commits. That being said, it is true that the error mention `lower_bound`...

numpyro/distributions/constraints.py
```diff
@@ -243,8 +280,16 @@ def __repr__(self):
     def feasible_like(self, prototype):
         return jax.numpy.broadcast_to(self.lower_bound + 1, jax.numpy.shape(prototype))
 
+    def tree_flatten(self):
+        return (self.lower_bound,), (("lower_bound",), dict())
 
-class _Positive(_GreaterThan, _SingletonConstraint):
+    def __eq__(self, other):
+        return isinstance(other, _GreaterThan) & jnp.array_equal(
```
pierreglaser: Mmh, I think the use of the "bitwise and" operator as a jax-compatible `and` (that does not try to coerce an abstract boolean to a concrete value) is causing `__eq__` to attempt accessing attribute like `lower_bound` even when `other` is not of type `_GreaterThan`. For some reason this happens during the tests, but it could have happened otherwise, so good the tests caught it!

pierreglaser:
OK @fehiepsi, I pushed a fix to the doc errors, that actually revealed some lack of robustness in the updated equality checks for constraints and transform in this PR. I added tests so hopefully that won't be a problem in the future :-)

fehiepsi (approved):
Woohoo, thanks for the fix!!

Issue #1576, opened 2023-04-13
Support pickling MCMC objects that use enumeration for discrete latent variables


PatNicholson:
When using enumeration for automatic marginalization of discrete latent variables, we find that we cannot pickle the MCMC object after mcmc.run(...). The following:

```python
import pickle

import numpyro
from numpyro.infer import MCMC, NUTS
import numpyro.distributions as dist
import jax.numpy as jnp
from jax import random

def model(y, n_components=2):
  with numpyro.plate('components', n_components):
    mu = numpyro.sample('mu', dist.Normal(0.0,1.0))
    tau = numpyro.sample('tau', dist.HalfCauchy(1.0))

  theta = numpyro.sample('theta', dist.Dirichlet(jnp.ones(n_components)))

  with numpyro.plate('obs', len(y)):
    component = numpyro.sample('component', dist.Categorical(theta),
                               infer={'enumerate': 'parallel'})
    numpyro.sample('y', dist.Normal(mu[component], tau[component]), obs=y)

nuts_kernel = NUTS(model)
mcmc = MCMC(sampler=nuts_kernel, num_warmup=500, num_samples=500, num_chains=2, progress_bar=False)
mcmc.run(random.PRNGKey(0), jnp.asarray([0.0, 1.0]))

with open("model.p", "wb") as f:
  pickle.dump(mcmc, f)
```

yields the error

```python
AttributeError                            Traceback (most recent call last)
<ipython-input-12-9a8a1284f203> in <cell line: 1>()
      1 with open("model.p", "wb") as f:
----> 2   pickle.dump(mcmc, f)

AttributeError: Can't pickle local object 'config_enumerate.<locals>.config_fn'
```

It would seem that config_fn needs to be cleaned up in the same way as other functions were for https://github.com/pyro-ppl/numpyro/issues/954.

PR #1577, opened 2023-04-14
Support pickling MCMC objects with enumeration

tare:
This PR aims to solve the issue reported in https://github.com/pyro-ppl/numpyro/issues/1576.

Moved the inner functions
https://github.com/pyro-ppl/numpyro/blob/0589859404433e3cfe08e90a37ab0ca272392d69/numpyro/contrib/funsor/infer_util.py#L72-L79 and  https://github.com/pyro-ppl/numpyro/blob/0589859404433e3cfe08e90a37ab0ca272392d69/numpyro/contrib/funsor/infer_util.py#L110-L119 outside of the functions `config_enumerate` and `config_kl`, respectively. I'm not sure in which use cases the current version of `config_kl` would produce issues but since it was using the same nested function pattern I updated it as well.

Implemented a couple of tests based on some existing tests to verify the appropriateness of the changes.

### DIFF
--- a/numpyro/contrib/funsor/infer_util.py
+++ b/numpyro/contrib/funsor/infer_util.py
@@ -42,6 +42,17 @@ def plate_to_enum_plate():
         numpyro.plate.__new__ = lambda *args, **kwargs: object.__new__(numpyro.plate)
 
 
+def _config_fn_enumerate(site, default):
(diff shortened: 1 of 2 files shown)


fehiepsi (approved):
Nice solution! Thanks for supporting this feature, @tare!

PR #1579, opened 2023-04-19
Fix broken links in documentation

tare:
https://github.com/pyro-ppl/numpyro/blob/master/README.md, https://num.pyro.ai/en/stable/getting_started.html, https://num.pyro.ai/en/stable/autoguide.html, and https://num.pyro.ai/en/stable/mcmc.html have outdated anchor links. For instance, the preamble on the https://num.pyro.ai/en/stable/autoguide.html page uses https://num.pyro.ai/en/latest/autoguide.html#autodelta instead of https://num.pyro.ai/en/stable/autoguide.html#numpyro.infer.autoguide.AutoDelta.

I fixed the broken anchor links in `README.md`, `docs/source/mcmc.rst`, and `docs/source/autoguide.rst`. It looks like https://num.pyro.ai/en/stable/getting_started.html is generated based on `README.md`.

Additionally, I replaced `http` with `https` when appropriate.

In some rare cases, the `readthedocs` link is used instead of the direct link, e.g. https://numpyro.readthedocs.io/en/latest/distributions.html vs. https://num.pyro.ai/en/latest/distributions.html. If desired, I can replace the readthedocs links with the direct links in order to make the linking style more consistent.


fehiepsi:
Could you remove `scipy` restriction [here](https://github.com/pyro-ppl/numpyro/blob/ffca0b8f0898b5332d5c07203b949467e5e82efe/setup.py#L59)? Because of that, numpyro is trying to install an old jax version.

fehiepsi (commented):
wow, it's tricky to find the reasons of those issues. Thanks for the fixes, @tare!!

fehiepsi (approved):

PR #1584, opened 2023-05-08
Include deterministic variables in AutoDelta's sample_posterior

### LINKED ISSUE #951: Guide samples for AutoDelta don't contain deterministic variables
I noticed that samples from the `AutoDelta` guide don't contain deterministic transforms of random variables in the model. Here's a minimal example which reproduces the issue:

```Python
import numpyro
from numpyro import optim
from numpyro.infer import SVI
from numpyro.infer.autoguide import AutoDelta
from numpyro.infer.elbo import Trace_ELBO
from jax import random

def model():
    x = numpyro.sample('x', dist.Normal(0, 1).expand([10]))
    numpyro.deterministic('x2', x**2)
    
guide = AutoDelta(model)
svi = SVI(model, guide, optim.Adam(0.01), Trace_ELBO())
svi_state = svi.init(random.PRNGKey(0))
svi_result = svi.run(random.PRNGKey(1), 1000)
guide_samples = guide.sample_posterior(random.PRNGKey(2), svi_result.params)
print(guide_samples['x2'])
```
Is this related to #946?

nikmich1:
This attempts to fix #951, by including deterministic variables in the output of AutoDelta's sample_posterior

Changes made to AutoDelta's `sample_posterior` method:
- After generating latent samples, generate a list of deterministic variables
- If no deterministic variables are present in the model, return the latent samples as before
- If there are deterministic variables, create a `Predictive` instance using the latent samples and generate samples of only the deterministic variables with the `return_sites` keyword. Since the Predictive instance in the new `sample_posterior` method requires the model's `*args` and `**kwargs`, these are now passed to `sample_posterior` as well
- Add the deterministic samples to the return value of the `sample_posterior` method

**Important:** Since the `Predictive` instance is called in the new `sample_posterior` method, now the model needs to be callable without arguments, if no arguments are passed to `sample_posterior`. 
Therefore, the following existing tests have been slightly modified, such that the model can also be called without passing data
- test_autoguide_deterministic
- test_logistic_regression

New tests added:
- Test, if new `sample_posterior` method contains deterministic variables (test_autodelta_capture_deterministic_variables)
- Test shapes, if `sample_shape` argument is used in the new `sample_posterior` method (test_autodelta_sample_posterior_with_sample_shape)

### DIFF
--- a/numpyro/infer/autoguide.py
+++ b/numpyro/infer/autoguide.py
@@ -44,6 +44,7 @@ from numpyro.distributions.util import (
     periodic_repeat,
     sum_rightmost,
 )
+from numpyro.infer import Predictive
 from numpyro.infer.elbo import Trace_ELBO
 from numpyro.infer.initialization import init_to_median, init_to_uniform
 from numpyro.infer.util import helpful_support_errors, initialize_model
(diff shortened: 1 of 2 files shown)


numpyro/infer/autoguide.py
```diff
@@ -455,12 +456,25 @@ def __call__(self, *args, **kwargs):
 
         return result
 
-    def sample_posterior(self, rng_key, params, sample_shape=()):
+    def sample_posterior(self, rng_key, params, sample_shape=(), *args, **kwargs):
```
fehiepsi, line 459: nit: `*args, sample_shape=(), **kwargs`

fehiepsi (approved):
Looks great!! Thanks for the clear implementation and adding tests, @nikmich1!

numpyro/infer/autoguide.py
nikmich1: Thank you for the review! Just corrected the order arguments

fehiepsi (approved):

PR #1590, opened 2023-05-19
Stein mixtures v2

OlaRonning:
In merging the original Stein mixture paper with ours, we found an inconsistency in the original article. We correct that here and can (probably) connect the current form with optimizing the ELBO (hence a lower bound on the marginal likelihood). Feel free to PM me for details if you are interested.

The current version uses a [quick and dirty version to separate the `model_density` from the `guide_density`](https://github.com/aleatory-science/numpyro/blob/feature/sm/numpyro/contrib/einstein/stein_loss.py#L12). I'm sure this can be done more elegantly.

## TODO
- [x] Update documentation
- [x] Update tests.
- [x] Refactor/rework [stein loss](https://github.com/aleatory-science/numpyro/blob/feature/sm/numpyro/contrib/einstein/stein_loss.py#L12)
- [x] Add test for `SteinLoss`
- [ ] Refactor/rework [Predictive](https://github.com/aleatory-science/numpyro/blob/feature/sm/numpyro/infer/util.py#L1012) broadcasting solution models for multiple parameters.


test/contrib/einstein/test_steinvi.py
```diff
@@ -234,35 +232,6 @@ def model(obs):
                 assert_array_approx_equal(init_value, np.full(expected_shape, 0.0))
 
 
-def test_svgd_loss_and_grads():
```
OlaRonning, line 237: Stein reports the 2-norm of Stein force instead of  ELBO loss--the system converges when the norm is zero. Note that the norm may not reach zero due to stochasticity in the gradient.

fehiepsi (dismissed):
LGTM. Do you want me to look at some particular parts? @OlaRonning

numpyro/infer/util.py
```diff
@@ -1008,11 +1008,26 @@ def __call__(self, rng_key, *args, **kwargs):
         elif self.batch_ndims == 1:  # batch over parameters
             batch_size = jnp.shape(tree_flatten(self.params)[0][0])[0]
             rng_keys = random.split(rng_key, batch_size)
+            # TODO: better way to broadcast the model across particles?
```
fehiepsi, line 1011: this seems dangerous to me, we define batchsize as the first dimension of one latent variable, if we allow different first dimensions, it is ambiguity which one needs to be batch size.

OlaRonning: Another solution is to sample the particle assignments first and then treat the predictive distribution as ‘nbatch_dim=0’. I could put the (mixture component) assignment logic in a predictive wrapper under Einstein. The way it doesn't interfere with the other inference methods.

OlaRonning:
It looks like `files changed` is against an old version of Master.

Issue #1598, opened 2023-06-01
Clarify how to sample the centered parameter in LocScaleReparam


Madhav-Kanda:
The documentation for [LocScaleReparam](https://num.pyro.ai/en/stable/reparam.html?highlight=locscalereparam#numpyro.infer.reparam.LocScaleReparam) in case of centeredness mentions that: " If None (default) learn a per-site per-element centering parameter in [0,1]", but upon looking through the implementation of the algorithm, I did not find anything relevant to finding the correct centeredness (as far as I understand the implementation by default it takes the centeredness to be 0.5 and does not optimize it).

Example:
In the case of the eight schools' examples, the ideal centeredness should be close to 0. Still, upon using LocScaleReparam with None as its parameter, we get the centeredness is 0.5, as seen below (which I believe is hard coded in the implementation of LocScaleReparam itself for the case of None).

```
J = 8
y = np.array([28.0, 8.0, -3.0, 7.0, -1.0, 1.0, 18.0, 12.0])
sigma = np.array([15.0, 10.0, 16.0, 11.0, 9.0, 11.0, 10.0, 18.0])
```

```
def eight_schools_noncentered(J, sigma, lambd, y=None):
    mu = numpyro.sample('mu', dist.Normal(2, 5))
    tau = numpyro.sample('tau', dist.HalfCauchy(5))
    with numpyro.plate('J', J):
        with numpyro.handlers.reparam(config={'theta': LocScaleReparam(centered=lambd)}):
            theta = numpyro.sample('theta', dist.Normal(mu, tau))
        numpyro.sample('obs', dist.Normal(theta, sigma), obs=y)
```

```
nuts_kernel = NUTS(eight_schools_noncentered)
mcmc = MCMC(nuts_kernel, num_warmup=1000, num_samples=1000)
rng_key = random.PRNGKey(0)
mcmc.run(rng_key, J, sigma, lambd = None,y=y, extra_fields=('potential_energy',))
mcmc.print_summary()
```

```
mcmc = MCMC(nuts_kernel, num_warmup=1000, num_samples=1000)
mcmc.run(rng_key, J, sigma, lambd = 0.5,y=y, extra_fields=('potential_energy',))
mcmc.print_summary()
```

In the above case both of them provide the same print_summary. Thus, I believe that either the code for the LocScaleReparam should be updated to include SVI for finding the best centeredness or the documentation should be updated accordingly.

fehiepsi:
> does not optimize it

[...]

PR #1599, opened 2023-06-01
Updated LocScaleReparam's documentation

Madhav-Kanda:
Refer #1598

### DIFF
--- a/numpyro/infer/reparam.py
+++ b/numpyro/infer/reparam.py
@@ -72,7 +72,10 @@ class LocScaleReparam(Reparam):
     1. *Automatic Reparameterisation of Probabilistic Programs*,
        Maria I. Gorinova, Dave Moore, Matthew D. Hoffman (2019)
 
-    :param float centered: optional centered parameter. If None (default) considers centered parameter to be 0.5. If 0, fully
+    :param float centered: optional centered parameter. If None (default) learn
+        a per-site per-element centering parameter in ``[0,1]`` initialized at value 0.5.
+        To sample the parameter, consider using :class:`~numpyro.handlers.lift` handler with a
+        prior like ``Uniform(0, 1)`` to cast the parameter to a latent variable. If 0, fully
         decenter the distribution; if 1, preserve the centered distribution
         unchanged.
     :param shape_params: list of additional parameter names to copy unchanged from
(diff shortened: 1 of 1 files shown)


numpyro/infer/reparam.py
```diff
@@ -72,8 +72,7 @@ class LocScaleReparam(Reparam):
     1. *Automatic Reparameterisation of Probabilistic Programs*,
        Maria I. Gorinova, Dave Moore, Matthew D. Hoffman (2019)
 
-    :param float centered: optional centered parameter. If None (default) learn
-        a per-site per-element centering parameter in ``[0,1]``. If 0, fully
+    :param float centered: optional centered parameter. If None (default) considers centered parameter to be 0.5. If 0, fully
```
fehiepsi, line 75: Could you change this to something like:
```
    :param float centered: optional centered parameter. If None (default) learn
        a per-site per-element centering parameter in ``[0,1]`` initialized at value 0.5.
        To sample the parameter, consider using :class:`~numpyro.handlers.lift` handler with a
        prior like ``Uniform(0, 1)`` to cast the parameter to a latent variable. If 0, fully
```

Madhav-Kanda: Hi, I would change this, but based on my understanding of the current implementation, it would not learn the center param even when we keep it None. It would simply keep the centered param as 0.5. To learn the correct parameter based on Maria Gorinova's paper, we will have to use Variational Inference. Please correct me if I need to correct something. Thanks for the quick response.

fehiepsi: That's why we say this is a parameter, not a latent variable. When using MCMC, you are drawing samples of latent variables. When you say "optimizing", it typically means that you are using inference methods with optimizers like SGD, Adam and gradient descent to optimize the corresponding parameters. When you say "sampling", it means that you are using MCMC to draw samples of latent "random variable".

fehiepsi: note that in numpyro, MCMC does not sample "parameters", it only samples latent variables.

Madhav-Kanda: Oh, okay. Thank you!

fehiepsi (approved):

PR #1608, opened 2023-06-16
Add support for local variables in RenyiELBO

fehiepsi:


numpyro/contrib/nested_sampling.py
```diff
@@ -260,7 +260,6 @@ def prior_model() -> PriorModelGen:
             num_live_points=model.U_ndims * 25,
             num_parallel_samplers=1,
             max_samples=1e4,
-            uncert_improvement_patience=2,
```
martinjankowiak, line 263: intentional?

numpyro/infer/elbo.py
```diff
@@ -311,53 +312,115 @@ class RenyiELBO(ELBO):
         Here :math:`\alpha \neq 1`. Default is 0.
     :param num_particles: The number of particles/samples
         used to form the objective (gradient) estimator. Default is 2.
+    :param callable create_plates: An optional function inputing the same
```
martinjankowiak, line 315: can we add a usage example to the doc string?

numpyro/infer/elbo.py
```diff
@@ -311,53 +312,115 @@ class RenyiELBO(ELBO):
...
+            }
+            if site_plates:
+                common_plates = set.intersection(*site_plates.values())
+            else:
+                common_plates = set()
+            common_plate_scale = 1.0
+            for frame in common_plates:
+                subsample_size = frame.size
+                size = model_trace[frame.name]["args"][0]
+                if size > subsample_size and frame.name not in indep_subsample_plates:
```
martinjankowiak, line 375: what is the difference between `frame`s that are in and not in `indep_subsample_plates`?

numpyro/infer/elbo.py
```diff
@@ -311,53 +312,115 @@ class RenyiELBO(ELBO):
...
+                        log_prob = jnp.sum(log_prob, axis=dim, keepdims=True)
+                        squeeze_axes = squeeze_axes + (dim,)
+                    log_prob = jnp.squeeze(log_prob, squeeze_axes)
+                    log_densities[trace_type] = log_densities[trace_type] + log_prob
+
             # log p(z) - log q(z)
-            elbo = model_log_density - guide_log_density
-            return elbo
+            elbo = log_densities["model"] - log_densities["guide"]
+            # Do not scale elbo at common plate dimensions.
```
martinjankowiak, line 400: doesn't this sort of break the pyro dimension? what scale is this exactly?

numpyro/infer/elbo.py
```diff
@@ -311,53 +312,115 @@ class RenyiELBO(ELBO):
...
         scaled_elbos = (1.0 - self.alpha) * elbos
-        avg_log_exp = logsumexp(scaled_elbos) - jnp.log(self.num_particles)
+        avg_log_exp = logsumexp(scaled_elbos, axis=0) - jnp.log(self.num_particles)
         weights = jnp.exp(scaled_elbos - avg_log_exp)
         renyi_elbo = avg_log_exp / (1.0 - self.alpha)
-        weighted_elbo = jnp.dot(stop_gradient(weights), elbos) / self.num_particles
-        return -(stop_gradient(renyi_elbo - weighted_elbo) + weighted_elbo)
+        weighted_elbo = (stop_gradient(weights) * elbos).mean(0)
+        loss = -(stop_gradient(renyi_elbo - weighted_elbo) + weighted_elbo)
+        return loss.sum() * jnp.mean(common_plate_scale)
```
martinjankowiak, line 423: is this a real mean or could it be `common_plate_scale[0]`?

numpyro/infer/elbo.py
```diff
@@ -311,53 +312,115 @@ class RenyiELBO(ELBO):
...
+            )
+        else:
+            compute_elbo_fn = single_particle_elbo
 
         rng_keys = random.split(rng_key, self.num_particles)
-        elbos = vmap(single_particle_elbo)(rng_keys)
+        elbos, common_plate_scale = vmap(compute_elbo_fn)(rng_keys)
         scaled_elbos = (1.0 - self.alpha) * elbos
-        avg_log_exp = logsumexp(scaled_elbos) - jnp.log(self.num_particles)
+        avg_log_exp = logsumexp(scaled_elbos, axis=0) - jnp.log(self.num_particles)
```
martinjankowiak, line 418: can we add a few shape asserts in these few lines? (assembling the loss)

test/infer/test_svi.py
```diff
@@ -62,6 +62,54 @@ def renyi_loss_fn(x):
...
+
+    # Test that the scales are applied correctly.
+    # Here for each particle, log_p - log_q = log(0.6)
+    full_loss = renyi_loss_fn()
+    subsample_loss = renyi_loss_fn(subsample_size=2)
+    assert_allclose(full_loss, -jnp.log(0.6) * 100, rtol=1e-6)
+    assert_allclose(subsample_loss, full_loss, rtol=1e-6)
+
+
+def test_renyi_create_plates():
```
martinjankowiak, line 88: don't we need to test multiple nested plates and multiple non-nested plates?

numpyro/contrib/nested_sampling.py
fehiepsi: yup, this is causing CI failing in master branch.

numpyro/infer/elbo.py
fehiepsi: done

numpyro/infer/elbo.py
fehiepsi: done

numpyro/infer/elbo.py
fehiepsi: it could be common_plate_scale[0].

numpyro/infer/elbo.py
fehiepsi: I just added comments in the code. Basically this decides whether or not we want to keep those subsample plate dimensions in the important weight.

numpyro/infer/elbo.py
fehiepsi: I added a better test to illustrate this. Assume "N" is a plate with size N and subsample size n, the scale is N / n. We store the scaled log probabilities in the trace. Here we remove such scaling operator because we don't want to perform Monte Carlo estimation at this dimension before computing renyi_elbo. We will perform MC estimation after computing renyi_elbo for each item of this plate.

test/infer/test_svi.py
fehiepsi: just added tests for both cases. non-nested plates are less interesting but agreed that it's nice to have a test.

fehiepsi:
Thanks for reviewing, @martinjankowiak! Let me know if you have further questions.

numpyro/infer/elbo.py
```diff
@@ -311,53 +312,166 @@ class RenyiELBO(ELBO):
...
+        def guide(data):
+            w_loc = numpyro.param("w_loc", 1.)
+            w_scale = numpyro.param("w_scale", 1.)
+            with numpyro.plate("batch", 10000, subsample_size=100):
+                batch = numpyro.subsample(data, event_dim=0)
+                loc = w_loc * batch
+                scale = jnp.exp(w_scale * batch)
+                numpyro.sample("latent", dist.Normal(loc, scale))
+
+        def create_plates(data):
```
martinjankowiak, line 338: i guess the user will wonder why this information can't be extracted from the model/guide...

numpyro/infer/elbo.py
```diff
@@ -311,53 +312,166 @@ class RenyiELBO(ELBO):
...
+        if site_plates:
+            indep_plates = set.intersection(*site_plates.values())
+        else:
+            indep_plates = set()
+        indep_plate_scale = 1.0
+        for frame in indep_plates:
+            subsample_size = frame.size
+            size = model_trace[frame.name]["args"][0]
+            if size > subsample_size:
+                # If a subsample plate is not obtained via create_plates, subsample
```
martinjankowiak, line 405: why is this exactly?

fehiepsi: For each particle, the model/guide is seeded using a particular seed, hence subsample indices will be different across particles.

martinjankowiak: i see it seems like this is something we could simplify with special logic for seeding plates...

martinjankowiak: what if `seed` returned extra plate seed info that could be passed around? something like
```
 seeded_model, plate_seeds = seed(model, model_seed, plate_seeds=None)
 seeded_guide, _ = seed(guide, guide_seed, plate_seeds=plate_seeds)
```

numpyro/infer/elbo.py
fehiepsi: I think sometimes, users will want to compute MC approximation before computing importance weights. If that usage case is rare, we can assume that subsample indices are always the same across particles. In this case, there are seperal ways to assure getting constant seeds for subsample plates across particles. E.g. we can introduce seed_plate handler and apply it before vmapping over particles. WDYT? should we always perform MC after computing the weights?

numpyro/infer/elbo.py
fehiepsi: that's right. there are several ways to assure constant seeds. the question is whether we want to maintain two behaviors. see my other comment

numpyro/infer/elbo.py
martinjankowiak: are you referring to a usage case that results in biased elbo estimates?

fehiepsi: yup. so I just removed it.

numpyro/infer/elbo.py
```diff
@@ -327,37 +348,88 @@ def __init__(self, alpha=0, num_particles=2):
...
+        for frame in indep_plates:
+            subsample_size = frame.size
+            size = model_trace[frame.name]["args"][0]
+            if size > subsample_size:
+                indep_plate_scale = indep_plate_scale * size / subsample_size
+        indep_plate_dims = {frame.dim for frame in indep_plates}
+
+        log_densities = {}
+        for trace_type, tr in {"guide": guide_trace, "model": model_trace}.items():
+            log_densities[trace_type] = 1.0
```
martinjankowiak, line 383: i guess the value doesn't matter but `0.0` would be clearer?

martinjankowiak (dismissed):
lgtm but it's hard to reason about how these changes might affect the dice bits (i.e. for non-reparameterizable guide sites). i think it'd be good to add a simple test, e.g. check that if you optimize renyi elbo the result is approximately 0.0 for `model = guide` with `model = Categorical(...)`. or do we have other tests that cover the non-reparameterizable case?

edit: maybe `test_svi.py::test_svi_discrete_latent` is sufficient? wdyt?

test/infer/test_svi.py
```diff
@@ -85,11 +193,12 @@ def body_fn(i, val):
...
         return svi_state
 
-    svi_state = fori_loop(0, 2000, body_fn, svi_state)
+    svi_state = fori_loop(0, 4000, body_fn, svi_state)
     params = svi.get_params(svi_state)
+    actual_posterior_mean = 0.75  # (8 + 1) / (8 + 1 + 2 + 1)
     assert_allclose(
         params["alpha_q"] / (params["alpha_q"] + params["beta_q"]),
-        0.8,
+        actual_posterior_mean,
```
martinjankowiak, line 201: does this pass with tighter tolerance?

numpyro/infer/elbo.py
```diff
@@ -327,37 +348,88 @@ def __init__(self, alpha=0, num_particles=2):
...
+        }
+        # We will compute Renyi elbos separately across dimensions
+        # defined in indep_plates. Then the final elbo is the sum
+        # of those independent elbos.
+        if site_plates:
+            indep_plates = set.intersection(*site_plates.values())
+        else:
+            indep_plates = set()
+        indep_plate_scale = 1.0
+        for frame in indep_plates:
```
martinjankowiak, line 374: also i guess we should raise an error if there are frames not in `indep_plates` that do subsampling

test/infer/test_svi.py
fehiepsi: it passes with a bit smaller tolerance and tuning

numpyro/infer/elbo.py
fehiepsi: just added it but I'm not sure it's necessary

test/infer/test_svi.py
fehiepsi: I think it's tricky to use small tolerance for this test (both elbo and renyi-elbo): the losses converge and fluctuate around the optimal value.

numpyro/infer/elbo.py
```diff
@@ -370,6 +370,16 @@ def _single_particle_elbo(self, model, guide, param_map, args, kwargs, rng_key):
             indep_plates = set.intersection(*site_plates.values())
         else:
             indep_plates = set()
+        for frame in set.union(*site_plates.values()):
+            if frame not in indep_plates:
+                subsample_size = frame.size
+                size = model_trace[frame.name]["args"][0]
+                if size > subsample_size:
+                    raise ValueError(
+                        f"Subsample plate `{frame.name}` should cover"
```
martinjankowiak, line 379: how about

`RenyiELBO` only supports subsampling in plates that are common to all `sample` sites, e.g. a data plate that encloses the entire model.

martinjankowiak (dismissed):

numpyro/infer/elbo.py
fehiepsi: done

martinjankowiak (approved):

PR #1612, opened 2023-06-21
Uses `init_loc_fn` to initialize mixture particles

OlaRonning:
Currently, `SteinVI` adds Gaussian noise to a location parameter of an `autoguide` (in unconstraint space) to initialize the particles. While it's a reasonably simple solution, it's hidden from the user.

A better solution (as in less opaque) is to use the `autoguide.init_loc_fn` to find the initial configuration of the particles. 

I've sketched the change as a PR to make it easier to discuss. If we agree this is a better solution, I'll add some unit tests.


fehiepsi:
yes, it makes sense to use init_loc_fn to initialize the location parameter.

fehiepsi (approved):
Thanks, Ola!

PR #1617, opened 2023-06-29
Mixture assert message string should be an f-string

adrn:
This is a very minor PR, but I just hit this error message while working on a model and noticed that it should be an f-string.


fehiepsi (approved):
Thanks, @adrn!

PR #1619, opened 2023-06-29
Support for VAE in AutoSemiDAIS

fehiepsi:
More precisely, this PR allows custom local guide in AutoSemiDAIS.

### DIFF
--- a/numpyro/infer/autoguide.py
+++ b/numpyro/infer/autoguide.py
@@ -1409,7 +1409,7 @@ class AutoSemiDAIS(AutoGuide):
                     with handlers.block():
                         with warnings.catch_warnings():
                             warnings.simplefilter("ignore")
-                            scale = N / idx.shape[0]
+                            scale = N / subsample_size
                             return (
                                 -potential_energy(
                                     subsample_guide,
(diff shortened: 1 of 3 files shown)


martinjankowiak:
@fehiepsi i'm confused what you're trying to support here. the whole point of `AutoSemiDAIS` is that DAIS takes care of local latent variables. do you want to override that? or do you want DAIS to take care of latent variables not specified in the local guide? or something else?

numpyro/infer/autoguide.py
```diff
@@ -1134,6 +1138,7 @@ def local_model(theta):
     :param callable global_guide: A guide for the global latent variables, e.g. an autoguide.
         The return type should be a dictionary of latent sample sites names and corresponding samples.
         If there is no global variable in the model, we can set this to None.
+    :param callable local_guide: An optional guide for the local latent variables.
```
martinjankowiak, line 1141: `An optional guide for specifying the DAIS base distribution for local latent variables` ?

test/infer/test_autoguide.py
```diff
@@ -837,7 +837,11 @@ def model():
             batch = numpyro.subsample(data, event_dim=0)
             numpyro.sample("x", dist.Normal(batch, 1))
 
-    guide = AutoSemiDAIS(model, model, None)
+    def create_plates():
+        return numpyro.plate("N", 10, subsample_size=5, dim=-1)
+
+    local_guide = AutoNormal(model, create_plates=create_plates)
```
martinjankowiak, line 843: is it worth checking explicitly that this gives the same result as the `local_guide=None` case?

martinjankowiak (approved):

numpyro/infer/autoguide.py
fehiepsi: done

martinjankowiak (approved):

PR #1622, opened 2023-07-29
Fix quantile of mvn autoguides

fehiepsi:


martinjankowiak (commented):
can we introduce a test that would fail before the PR but not after?

martinjankowiak (approved):

PR #1624, opened 2023-08-15
Add vectorized_particles to ELBO

fehiepsi:
Allow to use lax.map instead of vmap in ELBO to reduce memory requirement

### DIFF
--- a/numpyro/infer/elbo.py
+++ b/numpyro/infer/elbo.py
@@ -34,8 +34,9 @@ class ELBO:
 
     :param num_particles: The number of particles/samples used to form the ELBO
         (gradient) estimators.
-    :param vectorize_particles: Whether to use `jax.vmap` to obtain elbos over the
-        particles. If False, we will use `jax.lax.map`. Defaults to True.
+    :param vectorize_particles: Whether to use `jax.vmap` to compute ELBOs over the
+        num_particles-many particles in parallel. If False use `jax.lax.map`.
+        Defaults to True.
     """
 
     """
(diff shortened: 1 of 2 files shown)


numpyro/infer/elbo.py
```diff
@@ -33,6 +34,8 @@ class ELBO:
 
     :param num_particles: The number of particles/samples used to form the ELBO
         (gradient) estimators.
+    :param vectorize_particles: Whether to use `jax.vmap` to obtain elbos over the
```
martinjankowiak, line 37: is this only for eval or also for training?

fehiepsi: this is for both.

martinjankowiak: nit:
> :param vectorize_particles: Whether to use `jax.vmap` to compute ELBOs over the `num_particles`-many particles in parallel. If False use `jax.lax.map`. Defaults to True.

numpyro/infer/elbo.py
```diff
@@ -108,11 +112,10 @@ class Trace_ELBO(ELBO):
...
```
martinjankowiak, line 115: ditto

numpyro/infer/elbo.py
```diff
@@ -311,6 +320,8 @@ class RenyiELBO(ELBO):
         Here :math:`\alpha \neq 1`. Default is 0.
     :param num_particles: The number of particles/samples
         used to form the objective (gradient) estimator. Default is 2.
+    :param vectorize_particles: Whether to use `jax.vmap` to obtain elbos over the
+        particles. If False, we will use `jax.lax.map`. Defaults to True.
```
martinjankowiak, line 324: ditto

martinjankowiak (approved):
lgtm

PR #1625, opened 2023-08-15
Respect log_density in kl of delta

fehiepsi:


OlaRonning (approved):

fehiepsi:
Thanks Ola, we typically use Delta in auto guides. I noticed this bug while inspecting Trace mean field elbo.

PR #1628, opened 2023-08-17
Fix bug in SineBivariateVonMises sampler

deoxyribose:
See https://cran.r-project.org/web/packages/simdd/index.html for R implementation.
Before fix, highly correlated distributions would fail to sample correctly, and max_sample_iter would exceed the default 1000, silently.
After fix, the same distribution sample correctly distributed values.

### DIFF
--- a/numpyro/distributions/directional.py
+++ b/numpyro/distributions/directional.py
@@ -477,7 +477,7 @@ class SineBivariateVonMises(Distribution):
             phi_key, key = random.split(key)
             accept_key, acg_key, phi_key = random.split(phi_key, 3)
 
-            x = jnp.sqrt(1 + 2 * eig / b0) * random.normal(acg_key, shape)
+            x = 1./jnp.sqrt(1 + 2 * eig / b0) * random.normal(acg_key, shape)
             x /= jnp.linalg.norm(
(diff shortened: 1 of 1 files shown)


numpyro/distributions/directional.py
```diff
@@ -477,7 +477,7 @@ def update_fn(curr):
             phi_key, key = random.split(key)
             accept_key, acg_key, phi_key = random.split(phi_key, 3)
 
-            x = jnp.sqrt(1 + 2 * eig / b0) * random.normal(acg_key, shape)
+            x = 1./jnp.sqrt(1 + 2 * eig / b0) * random.normal(acg_key, shape)
```
fehiepsi, line 480: nit: you can use `jnp.rsqrt` for this.

deoxyribose: Cool, I made the edit.

OlaRonning (approved):
Thanks @deoxyribose. This is indeed correct and consistent with the R implementation. Could you make a PR with the exact change [here](https://github.com/pyro-ppl/pyro/blob/dev/pyro/distributions/sine_bivariate_von_mises.py#L207).

PR #1630, opened 2023-08-21
dispatch _promote_batch_shape_expanded to Independent

deoxyribose:
Independent distributions aren't dispatched to in batch_utils - the fix reuses the batch promoting for ExpandedDistribution and runs without errors, but hasn't been tested thoroughly, perhaps @pierreglaser can review?

### DIFF
--- a/numpyro/distributions/batch_util.py
+++ b/numpyro/distributions/batch_util.py
@@ -524,8 +524,8 @@ def _default_promote_batch_shape(d: Distribution):
     return new_self
 
 
-@promote_batch_shape.register
-def _promote_batch_shape_expanded(d: Union[ExpandedDistribution, Independent]):
+@promote_batch_shape.register(Independent)
+def _promote_batch_shape_expanded(d: ExpandedDistribution):
     orig_delta_batch_shape = d.batch_shape[
         : len(d.batch_shape) - len(d.base_dist.batch_shape)
     ]
@@ -562,7 +562,6 @@ def _promote_batch_shape_expanded(d: Union[ExpandedDistribution, Independent]):
 
     new_self.base_dist = new_base_dist
     return new_self
-
 
 @promote_batch_shape.register
 def _promote_batch_shape_masked(d: MaskedDistribution):
--- a/numpyro/distributions/batch_util.py
+++ b/numpyro/distributions/batch_util.py
@@ -564,14 +564,20 @@ def _promote_batch_shape_expanded(d: ExpandedDistribution):
     return new_self
 
 
-promote_batch_shape.register(Independent, _promote_batch_shape_expanded)
-
-
 @promote_batch_shape.register
 def _promote_batch_shape_masked(d: MaskedDistribution):
     new_self = copy.copy(d)
     new_base_dist = promote_batch_shape(d.base_dist)
     new_self._batch_shape = new_base_dist.batch_shape
+    new_self.base_dist = new_base_dist
+    return new_self
+
+
+@promote_batch_shape.register
+def _promote_batch_shape_independent(d: Independent):
+    new_self = copy.copy(d)
+    new_base_dist = promote_batch_shape(d.base_dist)
+    new_self._batch_shape = new_base_dist.batch_shape[: d.event_dim]
     new_self.base_dist = new_base_dist
     return new_self
 
--- a/numpyro/distributions/batch_util.py
+++ b/numpyro/distributions/batch_util.py
@@ -53,6 +53,7 @@ from numpyro.distributions.distribution import (
     ExpandedDistribution,
     MaskedDistribution,
     Unit,
+    Independent,
 )
 from numpyro.distributions.transforms import (
     AffineTransform,
@@ -524,7 +525,7 @@ def _default_promote_batch_shape(d: Distribution):
 
 
 @promote_batch_shape.register
-def _promote_batch_shape_expanded(d: ExpandedDistribution):
+def _promote_batch_shape_expanded(d: Union[ExpandedDistribution, Independent]):
     orig_delta_batch_shape = d.batch_shape[
         : len(d.batch_shape) - len(d.base_dist.batch_shape)
     ]
--- a/numpyro/distributions/batch_util.py
+++ b/numpyro/distributions/batch_util.py
@@ -51,9 +51,9 @@ from numpyro.distributions.discrete import (
 from numpyro.distributions.distribution import (
     Distribution,
     ExpandedDistribution,
+    Independent,
     MaskedDistribution,
     Unit,
-    Independent,
 )
 from numpyro.distributions.transforms import (
     AffineTransform,
--- a/numpyro/distributions/batch_util.py
+++ b/numpyro/distributions/batch_util.py
@@ -524,7 +524,7 @@ def _default_promote_batch_shape(d: Distribution):
     return new_self
 
 
-@promote_batch_shape.register(Independent)
+@promote_batch_shape.register
 def _promote_batch_shape_expanded(d: ExpandedDistribution):
     orig_delta_batch_shape = d.batch_shape[
         : len(d.batch_shape) - len(d.base_dist.batch_shape)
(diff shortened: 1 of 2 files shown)


deoxyribose:
[functools.singledispatch does not support Union types](https://bugs.python.org/issue46014)
A workaround would be to add
```python
    def register_for_union_type(target_fn):
        def decorator(fn):
            arg_type = list(get_type_hints(fn).values())[0]
            assert arg_type.__origin__ is Union
            for type_ in arg_type.__args__:
                fn = target_fn.register(type_)(fn)
            return fn

        return decorator
```
and do
```python
@register_for_union_type(promote_batch_shape)
def _promote_batch_shape_expanded(d: Union[ExpandedDistribution, Independent]):
```

fehiepsi:
I think you can do
```
promote_batch_shape.register(dist.Independent)(_promote_batch_shape_expanded)

File: numpyro/distributions/batch_util.py [hunk 1]
```diff
@@ -523,7 +524,7 @@ def _default_promote_batch_shape(d: Distribution):
     return new_self
 
 
-@promote_batch_shape.register
+@promote_batch_shape.register(Independent)
```
fehiepsi, line 527: This will disable behavior for ExpandedDistribution. You might need to have a separate register for Inpedendent (reuse _promote_batch_shape_expanded as in my comment). Anyway, could you add description for why this is needed. The default behavior does not work for Independent?

deoxyribose: Here's an example that reproduces the error I got:

```python
import numpyro
import jax.numpy as jnp
import jax.random as random
import numpy as np
from numpyro.handlers import mask
from optax import adamw
from numpyro.contrib.control_flow import scan

def mask_inside_plate_inside_scan(y=None, T=10):
    def transition(carry, y_curr):
        x_prev, t = carry
        with numpyro.plate("N", 10, dim=-1):
            with mask(mask=(t < T)):
                x_curr = numpyro.sample('x', dist.Normal(jnp.zeros((10,8)), jnp.ones((10,8))).to_event(1))
                y_curr = numpyro.sample('y', dist.Normal(x_curr, jnp.ones((10,8))).to_event(1), obs=y_curr)
                return (x_curr, t+1), None
    x0 = numpyro.sample('x_0', dist.Normal(jnp.zeros((10,8)), jnp.ones((10,8))).to_event(1))
    print
    x, t = scan(transition, (x0, 0), y, length=T)
    return (x, y)

    x, y = mask_inside_plate_inside_scan()
```

```
  File "/home/frans/Desktop/test_batch_util_independent.py", line 25, in <module>
    x, y = mask_inside_plate_inside_scan()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/frans/Desktop/test_batch_util_independent.py", line 21, in mask_inside_plate_inside_scan
    x, t = scan(transition, (x0, 0), y, length=T)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/frans/protein_dmm/numpyro/numpyro/contrib/control_flow/scan.py", line 441, in scan
          ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/frans/protein_dmm/numpyro/numpyro/primitives.py", line 53, in apply_stack
    default_process_message(msg)
  File "/home/frans/protein_dmm/numpyro/numpyro/primitives.py", line 28, in default_process_message
    msg["value"] = msg["fn"](*msg["args"], **msg["kwargs"])
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/frans/protein_dmm/numpyro/numpyro/contrib/control_flow/scan.py", line 313, in scan_wrapper
    site["fn"] = promote_batch_shape(site["fn"])
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/functools.py", line 909, in wrapper
    return dispatch(args[0].__class__)(*args, **kw)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/frans/protein_dmm/numpyro/numpyro/distributions/batch_util.py", line 569, in _promote_batch_shape_masked
    new_base_dist = promote_batch_shape(d.base_dist)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/functools.py", line 909, in wrapper
    return dispatch(args[0].__class__)(*args, **kw)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/frans/protein_dmm/numpyro/numpyro/distributions/batch_util.py", line 515, in _default_promote_batch_shape
    attr_name = list(d.arg_constraints.keys())[0]
                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^
IndexError: list index out of range
``

The problem is that x is a `MaskedDistribution`, whose base_dist is an `Independent` (whose base_dist is `Normal`). Since `Independent` isn't being dispatched to, it gets dispatched to `_default_promote_batch_shape`, which tries to access `arg_constraints`, which aren't there, but under the base_dist.
The error seems to only appear when inside a scan.

I misunderstood your comment before, but I've applied it now.

Comment (fehiepsi) [reply numpyro/distributions/batch_util.py]: The example is great!! Could you add the above test case to [test_control_flow](https://github.com/pyro-ppl/numpyro/blob/master/test/contrib/test_control_flow.py)? Thanks!

Comment (deoxyribose) [reply numpyro/distributions/batch_util.py]: I've added it now.

**Review by fehiepsi — APPROVED** [review]
Thanks!

File: numpyro/distributions/batch_util.py [hunk 2]
```diff
@@ -563,6 +564,9 @@ def _promote_batch_shape_expanded(d: ExpandedDistribution):
     return new_self
 
 
+promote_batch_shape.register(Independent, _promote_batch_shape_expanded)
```
pierreglaser, line 567: Mmh, I don't see any reason why the logic for batch shape promotion of `Independent` and `ExpandedDistribution` should be the same. Promoting `ExpandedDistribution` requires some care in maintaining broadcastability between itself and its `base_dist` when left-most inserting batch dimensions - I don't think `Independent` would require this kind of care.

Could you instead create  `_promote_batch_shape_independent` which simply promotes an independent's base distribution?

pierreglaser (changes requested):
Thanks @deoxyribose. Comments above :-)

test/contrib/test_control_flow.py
```diff
@@ -210,3 +210,31 @@ def transition_fn(c, val):
...
+
+        x0 = numpyro.sample(
+            "x_0", dist.Normal(jnp.zeros((10, 3)), jnp.ones((10, 3))).to_event(1)
+        )
+
+        x, t = scan(transition, (x0, 0), y, length=T)
+        return (x, y)
+
+    with numpyro.handlers.seed(rng_seed=0):
+        x, y = model()
```
pierreglaser, line 240: Could we be a bit more stringent, and test that some `numpyro` functionalities like computing log probs work well on the model, as done in other `scan` tests? That way we will spot more potential bugs in our implementation ahead of merging :-)

numpyro/distributions/batch_util.py
pierreglaser: Mmmh, unclear - `MaskedDistribution` enforces `dist.batch_shape == dist.base_dist.batch_shape` while an `Independent` distribution will have a different batch shape that its base.

deoxyribose: I've committed an ` _promote_batch_shape_independent` now, it passes the test, and outputs the correct batch_shape, but beyond that I'm not sure whether the logic is correct.

test/contrib/test_control_flow.py
deoxyribose: I've added a `log_density` call to the test.

fehiepsi (approved):
Thanks @deoxyribose!

PR #1652, opened 2023-09-27
Add ICDF and CDF to BernoulliProbs

12kleingordon34:
I was using `numpyro` for a particular application which required an ICDF and CDF for a Bernoulli, however this was not yet implemented.

I use Heaviside functions to calculate the CDF and ICDF. Tests are a little more complicated, as for a Bernoulli one doesn't recover the original quantiles when passing uniform samples through an ICDF, and then back through a CDF. I added a few specific lines for the Bernoulli such that the mean of ICDF(quantiles) is roughly equal to the probability parameter for the bernoulli. To test the CDF, I check that the proportion of ones yielded through CDF(ICDF(quantiles)) is roughly equal to p, and that the proportion of (1 - p) values through the same function is roughly equal to (1 - p).

### DIFF
--- a/numpyro/distributions/discrete.py
+++ b/numpyro/distributions/discrete.py
@@ -102,6 +102,13 @@ class BernoulliProbs(Distribution):
     def variance(self):
         return self.probs * (1 - self.probs)
 
+    def cdf(self, value):
+        return ((1 - self.probs) * jnp.heaviside(value, 1)
+                + self.probs * jnp.heaviside(value - 1, 1))
+
+    def icdf(self, q):
+        return jnp.heaviside(q - (1 - self.probs), 1)
+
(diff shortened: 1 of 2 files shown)


fehiepsi (approved):
Thanks so much for your first contribution, @12kleingordon34! Could you also add the same methods to BernoulliLogits? The implementation would be the same.

fehiepsi:
Hi @12kleingordon34, we will make a new release in the next few days. Do you want to incorporate this PR in the release?

12kleingordon34:
Hi @fehiepsi , apologies for leaving this for so long. I'll add the bernoulli logits by the end of next week with the corresponding test, but if it's too late i'll leave it for the next one.

fehiepsi:
Friendly ping :)

PR #1654, opened 2023-09-29
Added kl_divergence for multivariate normals

lumip:
KL implementation for `MultivariateNormal`

EDIT: just saw that there is already a pull request for this from a year ago (https://github.com/pyro-ppl/numpyro/pull/1487). What's the hold up with that?

### DIFF
--- a/numpyro/distributions/kl.py
+++ b/numpyro/distributions/kl.py
@@ -27,19 +27,20 @@
 
 from multipledispatch import dispatch
 
-from jax import lax, vmap
+from jax import lax
 import jax.numpy as jnp
 from jax.scipy.special import betaln, digamma, gammaln
-from jax.scipy.linalg import solve_triangular
 
 from numpyro.distributions.continuous import (
     Beta,
     Dirichlet,
     Gamma,
     Kumaraswamy,
-    Normal,
     MultivariateNormal,
+    Normal,
     Weibull,
+    _batch_solve_triangular,
+    _batch_trace_from_cholesky,
 )
 from numpyro.distributions.discrete import CategoricalProbs
 from numpyro.distributions.distribution import (
@@ -142,58 +143,45 @@ def kl_divergence(p: MultivariateNormal, q: MultivariateNormal):
 
     if p.event_shape != q.event_shape:
         raise ValueError(
-            "Distributions must be have the same event shape, but are"
+            "Distributions must have the same event shape, but are"
             f" {p.event_shape} and {q.event_shape} for p and q, respectively."
         )
 
-    if p.batch_shape != q.batch_shape:
+    min_batch_ndim = min(len(p.batch_shape), len(q.batch_shape))
+    if p.batch_shape[-min_batch_ndim:] != q.batch_shape[-min_batch_ndim:]:
         raise ValueError(
-            "Distributions must be have the same batch shape, but are"
-            f" {p.batch_shape} and {q.batch_shape} for p and q, respectively."
+            "Distributions must have the same batch shape in common batch dimensions, "
+            f"but are {p.batch_shape} and {q.batch_shape} for p and q,"
+            "respectively."
         )
+    result_batch_shape = (
+        p.batch_shape if len(p.batch_shape) >= len(q.batch_shape) else q.batch_shape
+    )
 
     assert len(p.event_shape) == 1, "event_shape must be one-dimensional"
     D = p.event_shape[0]
 
     assert p.mean.shape == p.batch_shape + p.event_shape
-    assert q.mean.shape == p.mean.shape
+    assert q.mean.shape == q.batch_shape + q.event_shape
 
-    def _single_mvn_kl(p_mean, p_scale_tril, q_mean, q_scale_tril):
-        assert jnp.ndim(p_mean) == 1
-        assert jnp.ndim(q_mean) == 1
-        assert jnp.ndim(p_scale_tril) == 2
-        assert jnp.ndim(q_scale_tril) == 2
+    p_half_log_det = jnp.log(jnp.diagonal(p.scale_tril, axis1=-2, axis2=-1)).sum(-1)
+    assert p_half_log_det.shape == p.batch_shape
 
-        p_half_log_det = jnp.log(
-            jnp.diagonal(p_scale_tril)
-        ).sum(-1)
-        q_half_log_det = jnp.log(
-            jnp.diagonal(q_scale_tril)
-        ).sum(-1)
-        log_det_ratio = 2 * (p_half_log_det - q_half_log_det)
+    q_half_log_det = jnp.log(jnp.diagonal(q.scale_tril, axis1=-2, axis2=-1)).sum(-1)
+    assert q_half_log_det.shape == q.batch_shape
 
-        Lq_inv = solve_triangular(q_scale_tril, jnp.eye(D), lower=True)
+    log_det_ratio = 2 * (p_half_log_det - q_half_log_det)
+    assert log_det_ratio.shape == result_batch_shape
 
-        tr = jnp.sum(jnp.diagonal(
-            Lq_inv.T @ (Lq_inv @ p_scale_tril) @ p_scale_tril.T
-        ))
+    Lq_inv = _batch_solve_triangular(q.scale_tril, jnp.eye(D))
 
-        z = jnp.matmul(Lq_inv, (p_mean - q_mean))
-        t1 = jnp.dot(z, z)
+    tr = _batch_trace_from_cholesky(Lq_inv @ p.scale_tril)
+    assert tr.shape == result_batch_shape
 
-        return .5 * (tr + t1 - D - log_det_ratio)
+    t1 = jnp.square(Lq_inv @ (p.loc - q.loc)[..., jnp.newaxis]).sum((-2, -1))
+    assert t1.shape == result_batch_shape
 
-    p_mean_flat = jnp.reshape(p.mean, (-1, D))
-    p_scale_tril_flat = jnp.reshape(p.scale_tril, (-1, D, D))
-
-    q_mean_flat = jnp.reshape(q.mean, (-1, D))
-    q_scale_tril_flat = jnp.reshape(q.scale_tril, (-1, D, D))
-
-    kl_flat = vmap(_single_mvn_kl)(p_mean_flat, p_scale_tril_flat, q_mean_flat, q_scale_tril_flat)
-    assert jnp.ndim(kl_flat) == 1
-
-    kl = jnp.reshape(kl_flat, p.batch_shape)
-    return kl
+    return 0.5 * (tr + t1 - D - log_det_ratio)
 
 
 @dispatch(Beta, Beta)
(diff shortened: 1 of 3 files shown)


numpyro/distributions/kl.py
```diff
@@ -134,6 +136,66 @@ def kl_divergence(p, q):
...
+def kl_divergence(p: MultivariateNormal, q: MultivariateNormal):
+    # cf https://statproofbook.github.io/P/mvn-kl.html
+
+    if p.event_shape != q.event_shape:
+        raise ValueError(
+            "Distributions must be have the same event shape, but are"
+            f" {p.event_shape} and {q.event_shape} for p and q, respectively."
+        )
+
+    if p.batch_shape != q.batch_shape:
```
fehiepsi, line 149: This check is unnecessary. It looks to me that the only operator which does not support batching is `solve_triangular`. I'll comment below.

numpyro/distributions/kl.py
```diff
@@ -134,6 +136,66 @@ def kl_divergence(p, q):
...
+        log_det_ratio = 2 * (p_half_log_det - q_half_log_det)
+
+        Lq_inv = solve_triangular(q_scale_tril, jnp.eye(D), lower=True)
+
+        tr = jnp.sum(jnp.diagonal(
+            Lq_inv.T @ (Lq_inv @ p_scale_tril) @ p_scale_tril.T
+        ))
+
+        z = jnp.matmul(Lq_inv, (p_mean - q_mean))
+        t1 = jnp.dot(z, z)
```
fehiepsi, line 182: You can use [_batch_mahalanobis](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/distributions/continuous.py#L1370) for this:
```
t1 = _batch_mahalanobis(q.scale_tril, q.loc - p.loc)
```

numpyro/distributions/kl.py
```diff
@@ -134,6 +136,66 @@ def kl_divergence(p, q):
...
+        q_half_log_det = jnp.log(
+            jnp.diagonal(q_scale_tril)
+        ).sum(-1)
+        log_det_ratio = 2 * (p_half_log_det - q_half_log_det)
+
+        Lq_inv = solve_triangular(q_scale_tril, jnp.eye(D), lower=True)
+
+        tr = jnp.sum(jnp.diagonal(
+            Lq_inv.T @ (Lq_inv @ p_scale_tril) @ p_scale_tril.T
+        ))
```
fehiepsi, line 179: You can use [_batch_solve_triangular](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/distributions/continuous.py#L1240) for this:
```
tmp = _batch_solve_triangular(q.scale_tril, p.scale_tril)
tr = jnp.square(tmp).sum((-1, -2))
```

fehiepsi (commented):
Thank you @lumip! The author of #1487 did not response so I think we can use this PR instead. I have a few comments below to address the batching issues.

numpyro/distributions/kl.py
fehiepsi: Actually, I like your approach better. We can
```Lq_inv = vmap(lambda A, x: solve_triangular(A, x, lower=True))(q.scale_tril.reshape((-1,) + q.scale_tril.shape[-2:]), jnp.eye(D))```.

Then compute
```
tr = jnp.square(Lq_inv @ p.scale_tril).sum((-1, -2))
t1 = jnp.square(Lq_inv @ (p.loc - q.loc)[..., None]).sum((-1, -2))
```

**Comment by lumip** [comment]
Thanks for the feedback @fehiepsi . I have modified the code according to your suggestions and also added a test case where the batch dimensions of the two distributions are not identical (and fixed the linter complaints). If everything is good now, I'd still rebase onto the recent master and merge the commits.

File: numpyro/distributions/kl.py [hunk 4]
```diff
@@ -142,58 +143,45 @@ def kl_divergence(p: MultivariateNormal, q: MultivariateNormal):
...
-        q_half_log_det = jnp.log(
-            jnp.diagonal(q_scale_tril)
-        ).sum(-1)
-        log_det_ratio = 2 * (p_half_log_det - q_half_log_det)
+    p_half_log_det = jnp.log(jnp.diagonal(p.scale_tril, axis1=-2, axis2=-1)).sum(-1)
+    assert p_half_log_det.shape == p.batch_shape
 
-        Lq_inv = solve_triangular(q_scale_tril, jnp.eye(D), lower=True)
+    q_half_log_det = jnp.log(jnp.diagonal(q.scale_tril, axis1=-2, axis2=-1)).sum(-1)
+    assert q_half_log_det.shape == q.batch_shape
```
fehiepsi, line 171: nit: this assertion might not be true. In MultivariateNormal implementation, we avoid unnecessary broadcasting (e.g. we can have a batch of means with a single scale_tril).

numpyro/distributions/kl.py
```diff
@@ -142,58 +143,45 @@ def kl_divergence(p: MultivariateNormal, q: MultivariateNormal):
...
     if p.event_shape != q.event_shape:
         raise ValueError(
-            "Distributions must be have the same event shape, but are"
+            "Distributions must have the same event shape, but are"
             f" {p.event_shape} and {q.event_shape} for p and q, respectively."
         )
 
-    if p.batch_shape != q.batch_shape:
+    min_batch_ndim = min(len(p.batch_shape), len(q.batch_shape))
+    if p.batch_shape[-min_batch_ndim:] != q.batch_shape[-min_batch_ndim:]:
```
fehiepsi, line 151: how about only assert that p.batch_shape and q.batch_shape can be broadcasted.
```
try:
    result_batch_shape = jnp.broadcast_shapes(p.batch_shape, q.batch_shape)
except ValueError:
    raise ...
```

numpyro/distributions/kl.py
```diff
@@ -142,58 +143,45 @@ def kl_divergence(p: MultivariateNormal, q: MultivariateNormal):
...
 
     assert p.mean.shape == p.batch_shape + p.event_shape
-    assert q.mean.shape == p.mean.shape
-
-    def _single_mvn_kl(p_mean, p_scale_tril, q_mean, q_scale_tril):
-        assert jnp.ndim(p_mean) == 1
-        assert jnp.ndim(q_mean) == 1
-        assert jnp.ndim(p_scale_tril) == 2
-        assert jnp.ndim(q_scale_tril) == 2
+    assert q.mean.shape == q.batch_shape + q.event_shape
```
fehiepsi, line 165: those assertions are unnecessary.

numpyro/distributions/kl.py
```diff
@@ -142,58 +143,45 @@ def kl_divergence(p: MultivariateNormal, q: MultivariateNormal):
...
+    log_det_ratio = 2 * (p_half_log_det - q_half_log_det)
+    assert log_det_ratio.shape == result_batch_shape
 
-        z = jnp.matmul(Lq_inv, (p_mean - q_mean))
-        t1 = jnp.dot(z, z)
+    Lq_inv = _batch_solve_triangular(q.scale_tril, jnp.eye(D))
 
-        return .5 * (tr + t1 - D - log_det_ratio)
+    tr = _batch_trace_from_cholesky(Lq_inv @ p.scale_tril)
+    assert tr.shape == result_batch_shape
```
fehiepsi, line 179: this assertion might not be true.

numpyro/distributions/kl.py
```diff
@@ -142,58 +143,45 @@ def kl_divergence(p: MultivariateNormal, q: MultivariateNormal):
...
 
-    p_mean_flat = jnp.reshape(p.mean, (-1, D))
-    p_scale_tril_flat = jnp.reshape(p.scale_tril, (-1, D, D))
+    t1 = jnp.square(Lq_inv @ (p.loc - q.loc)[..., jnp.newaxis]).sum((-2, -1))
+    assert t1.shape == result_batch_shape
```
fehiepsi, line 182: this assertion might not be true

numpyro/distributions/kl.py
```diff
@@ -142,58 +143,45 @@ def kl_divergence(p: MultivariateNormal, q: MultivariateNormal):
...
 
-        p_half_log_det = jnp.log(
-            jnp.diagonal(p_scale_tril)
-        ).sum(-1)
-        q_half_log_det = jnp.log(
-            jnp.diagonal(q_scale_tril)
-        ).sum(-1)
-        log_det_ratio = 2 * (p_half_log_det - q_half_log_det)
+    p_half_log_det = jnp.log(jnp.diagonal(p.scale_tril, axis1=-2, axis2=-1)).sum(-1)
+    assert p_half_log_det.shape == p.batch_shape
```
fehiepsi, line 168: this assertion might not be true

test/test_distributions.py
```diff
@@ -2849,12 +2849,24 @@ def test_kl_expanded_normal(batch_shape, event_shape):
...
-def test_kl_multivariate_normal_consistency_with_independent_normals(batch_shape):
-    event_shape = (5, )
-    shape = batch_shape + event_shape
+@pytest.mark.parametrize(
+    "batch_shape_p, batch_shape_q",
+    [
+        ((), ()),
+        ((1,), (1,)),
+        ((2, 3), (2, 3)),
+        ((5, 2, 3), (2, 3)),
```
fehiepsi, line 2858: could you change this to `(5, 1, 3)` and `(2, 3)`?

test/test_distributions.py
```diff
@@ -2849,12 +2849,24 @@ def test_kl_expanded_normal(batch_shape, event_shape):
...
+        ((2, 3), (5, 2, 3)),
```
fehiepsi, line 2859: nit: maybe `((1, 3), (5, 2, 3))`?

fehiepsi (commented):
LGTM, thanks @lumip! I have a few small comments on the assertions.

numpyro/distributions/kl.py
lumip: done

numpyro/distributions/kl.py
lumip: Right, I wasn't thinking of that. Removed the assertion and added tests for those cases.

numpyro/distributions/kl.py
lumip: removed

numpyro/distributions/kl.py
lumip: fixed

numpyro/distributions/kl.py
lumip: fixed

numpyro/distributions/kl.py
lumip: removed

test/test_distributions.py
lumip: done

test/test_distributions.py
lumip: done

fehiepsi (approved):
Thanks, @lumip!

PR #1661, opened 2023-10-11
correct event_dim use

deoxyribose:
Corrects use of event_dim in https://github.com/pyro-ppl/numpyro/pull/1630, which didn't work for tensor_dims larger than 2.


fehiepsi (approved):
Nice catch! Thanks, @deoxyribose!

Issue #1635, opened 2023-09-03
A tutorial for Automatic Partial centering of model using LocScaleReparam


Madhav-Kanda:
Hi,
I wonder if I could contribute an example/tutorial for doing automatic partial centering of the model using LocScaleReparam based on [Automatic Reparameterisation of Probabilistic Programs](https://arxiv.org/abs/1906.03028). I could not find any example/tutorial that guides how to use LocScalReparam along with SVI to get centeredness between 0 & 1 and consequently reparameterizing the model accordingly. Let me know what you think.

fehiepsi:
Yes, please feel free to contribute. Thanks, @Madhav-Kanda!

PR #1663, opened 2023-10-17
Tutorial on Automatic Partial centering of model using LocScaleReparam

Madhav-Kanda:
Hi, based on the issue #1635,  I have created a tutorial on Automatic Partial centering of model by using Variationally Inferred Parameterisation.  I have tried to create the tutorial following the template similar to the recently added tutorials. Let me know if there are any mistakes or error in the tutorial.


fehiepsi:
Awesome work! Thanks for the contribution, @Madhav-Kanda! Could you add the tutorial into the Introductory tutorial section of [this index file](https://github.com/pyro-ppl/numpyro/blob/master/docs/source/index.rst). Optionally, it would also be great to have a thumbnail for the tutorial in [this folder](https://github.com/pyro-ppl/numpyro/tree/master/docs/source/_static/img/tutorials). Thanks!

fehiepsi:
Could you move the file to the notebooks folders? https://github.com/pyro-ppl/numpyro/tree/master/notebooks/source

martinjankowiak:
thanks @Madhav-Kanda !

some additional comments:
- can you consistently use the z spelling for parameterization instead of mixing both spellings? including file names?
- nit: "the Neal's" => "Neal's"
- typo: hieratchical
- this phrase is confusing and should probably be removed: "under which the posterior is most like an independent normal distribution"

Madhav-Kanda:
Thanks for the suggestions @martinjankowiak, I have made changes based on the last three points but, I couldn't understand the first point:
> * can you consistently use the z spelling for parameterization instead of mixing both spellings? including file names?

Could you please provide an example to explain what do you mean by this?

martinjankowiak:
well as it stands file names and text switch back and forth between "(re)parameteri**Z**ation" and "(re)parameteri**S**ation"

Madhav-Kanda:
@martinjankowiak, @fehiepsi, I have made the suggested changes. Let me know if I missed out on something. Thanks!

fehiepsi (approved):

fehiepsi:
Thanks, @Madhav-Kanda!

PR #1665, opened 2023-10-20
Allow AutoSemiDAIS to work without global variable

fehiepsi:


numpyro/infer/autoguide.py
```diff
@@ -1250,13 +1250,16 @@ def local_model(theta):
         during partial momentum refreshments in HMC. Defaults to 0.9.
     :param float init_scale: Initial scale for the standard deviation of the variational
         distribution for each (unconstrained transformed) local latent variable. Defaults to 0.1.
+    :param str subsample_plate: Optional name of the subsample plate site. This is required
+        when the model does not have subsample plate (like in VAE settings).
```
martinjankowiak, line 1254: does not?

numpyro/infer/autoguide.py
```diff
@@ -1250,13 +1250,16 @@ def local_model(theta):
...
+    :param bool use_global_dais_params: Whether to use global parameters for DAIS dynamics.
```
martinjankowiak, line 1255: nit: Whether parameters controlling DAIS dynamics (HMC step size, HMC mass matrix, etc) should be global (i.e. common to all data points in the subsample plate) or local (i.e. each data point in the subsample plate has individual parameters).

numpyro/infer/autoguide.py
```diff
@@ -1301,6 +1308,10 @@ def _setup_prototype(self, *args, **kwargs):
             and isinstance(site["args"][1], int)
             and site["args"][0] > site["args"][1]
         }
+        if self.subsample_plate is not None:
```
martinjankowiak, line 1311: i'm a bit confused by these args/checks. afaik we should support the following cases:
- there is a single plate but there is no subsampling
- there is a single plate and it is subsampled
any other scenario (e.g. 0 plates or > 1 plates) is not supported. is that right?

numpyro/infer/autoguide.py
```diff
@@ -1527,17 +1571,32 @@ def base_z_dist_log_prob(z):
...
-                )
-                z_0_scale_init = jnp.ones((N, D)) * self.init_scale
-                z_0_scale = numpyro.param(
-                    "{}_z_0_scale".format(self.prefix),
-                    z_0_scale_init,
-                    constraint=constraints.positive,
-                    event_dim=1,
-                )
+                if self.use_global_dais_params:
+                    z_0_loc_init = jnp.zeros(D)
```
martinjankowiak, line 1575: i don't think this part makes sense. the `z` params should always be local.

martinjankowiak (commented):
does this PR support amortized q(z|x)?

numpyro/infer/autoguide.py
fehiepsi: hmm, I think amortized guide should have `local_guide` specified.

numpyro/infer/autoguide.py
fehiepsi: IIRC we allow multiple plates but only one subsample plate. I just added an `elif not subsample_plates: ...` branch to cover the case you mentioned above:
+ there is a single plate but there is no subsampling

numpyro/infer/autoguide.py
fehiepsi: done

numpyro/infer/autoguide.py
fehiepsi: just revised it to clarify that this is required when the model has a subsample plate without `subsample_size` specified.

fehiepsi:
yes, this supports amortized `local_guide`. Users need to specify it in the construction.

numpyro/infer/autoguide.py
martinjankowiak: but what if i just want non-amortized mean-field variational distributions for the locals? i would need to specify `local_guide` as opposed to relying on a convenient default behavior?

martinjankowiak (approved):

numpyro/infer/autoguide.py
fehiepsi: Yes, in that case, you might want to set `use_global_dais_params=False`. We might also change the semantics of this flag to `global_dais_params=None/"dynamic"/"full"`. wdyt?

martinjankowiak: what exactly are the three options?

fehiepsi: None is the current master behavior, or False in this PR. dynamic is your request. full is this PR.

martinjankowiak: sorry for being slow here (traveling). i think `use_global_dais_params` which controls e.g. `betas` should be entirely separate from whatever controls the distribution over `z_0`. what about the following behavior:
- if `local_guide` is provided use that.
- otherwise if `local_guide=None` and there exist local variables instantiate a auto mean-field guide?

fehiepsi: Hmm, that will imply that if users want to have global params for the base dist, they will need to use local_guide? I don't have a preference here. To me, base params play a similar role as betas; the dynamic will depend on model density etc. In the DAIS paper, the author even uses a fixed base dist.

martinjankowiak: i don't think there's any reason in practice why you'd really want global/shared params for the base dist q(z_0). having global params for e.g. beta makes sense because it's basically a "higher order quantity" and so harder to estimate and probably varies less from data point to data point. just in the same way that we'd probably generally be more comfortable in sharing scales/variances across data points than locs/means.

fehiepsi: makes sense to me, thanks!

PR #1666, opened 2023-10-20
Support multi sample guides in Trace_ELBO

fehiepsi:
This PR enables running SVI with guides which propose multiple samples.

### DIFF
--- a/numpyro/infer/elbo.py
+++ b/numpyro/infer/elbo.py
@@ -167,7 +167,10 @@ class Trace_ELBO(ELBO):
 
                 def get_model_density(key, latent):
                     with seed(rng_seed=key), substitute(data={**latent, **plates}):
-                        model_log_density, _ = log_density(model, args, kwargs, params)
+                        model_log_density, model_trace = log_density(
+                            model, args, kwargs, params
+                        )
+                    _validate_model(model_trace, plate_warning="loose")
                     return model_log_density
 
                 num_guide_samples = None
--- a/numpyro/infer/elbo.py
+++ b/numpyro/infer/elbo.py
@@ -50,7 +50,16 @@ class ELBO:
         self.num_particles = num_particles
         self.vectorize_particles = vectorize_particles
 
-    def loss(self, rng_key, param_map, model, guide, *args, **kwargs):
+    def loss(
+        self,
+        rng_key,
+        param_map,
+        model,
+        guide,
+        *args,
+        multi_sample_guide=False,
+        **kwargs,
+    ):
         """
         Evaluates the ELBO with an estimator that uses num_particles many samples/particles.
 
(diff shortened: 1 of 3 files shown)


numpyro/infer/elbo.py
```diff
@@ -116,15 +124,30 @@ class Trace_ELBO(ELBO):
...
         num_particles-many particles in parallel. If False use `jax.lax.map`.
         Defaults to True.
+    :param multi_sample_guide: Whether to make an assumption that the guide proposes
+        multiple samples.
     """
 
+    def __init__(
+        self, num_particles=1, vectorize_particles=True, multi_sample_guide=False
+    ):
+        self.multi_sample_guide = multi_sample_guide
```
martinjankowiak, line 134: what happens if `multi_sample_guide==True` and `num_particles > 1`?

numpyro/infer/elbo.py
```diff
@@ -116,15 +124,30 @@ class Trace_ELBO(ELBO):
     :param vectorize_particles: Whether to use `jax.vmap` to compute ELBOs over the
         num_particles-many particles in parallel. If False use `jax.lax.map`.
         Defaults to True.
+    :param multi_sample_guide: Whether to make an assumption that the guide proposes
```
martinjankowiak, line 127: why not infer this from `isinstance(guide, ...)`? do we expect guides with this attribute apart from a small number of `AutoGuide`s we construct?

numpyro/infer/elbo.py
```diff
@@ -135,29 +158,65 @@ def single_particle_elbo(rng_key):
...
+                model_log_density = model_log_density.sum(0)
+                # log p(z) - log q(z)
+                elbo_particle = (model_log_density - guide_log_density) / seeds.shape[0]
+            else:
+                seeded_model = seed(model, model_seed)
+                replay_model = replay(seeded_model, guide_trace)
+                model_log_density, model_trace = log_density(
+                    replay_model, args, kwargs, params
+                )
+                check_model_guide_match(model_trace, guide_trace)
```
martinjankowiak, line 198: so we lose these two checks if `multi_sample_guide==True`?

fehiepsi: we can check the later one. The first one is tricky.

numpyro/infer/elbo.py
fehiepsi: please feel free to remove this when such auto guides are available.

numpyro/infer/elbo.py
fehiepsi: num_particles behaves independently. i.e. we might have num_particles x num_samples_from_guide samples.

martinjankowiak (approved):

PR #1644, opened 2023-09-22
Add AutoGuideList

tare:
Hello! This PR eventually aims to address #1638.

Using the Pyro implementation (please see [guides.py](https://github.com/pyro-ppl/pyro/blob/dev/pyro/infer/autoguide/guides.py#L183-L275) and [block_messenger.py](https://github.com/pyro-ppl/pyro/blob/dev/pyro/poutine/block_messenger.py#L9-L142)) as an example I implemented `numpyro.infer.autoguide.AutoGuideList` and modified `numpyro.handlers.block` with the following main differences:
- discarded the deprecated `add` alias from `numpyro.infer.autoguide.AutoGuideList`
- the order of parameters in the constructor of the `block` handler differs from Pyro as I thought changing the positions of the current parameters could break things

Here is a simple example using`AutoGuideList`
```python
from jax import random
import jax.numpy as jnp

import numpyro
from numpyro import handlers, optim
import numpyro.distributions as dist
from numpyro.infer import SVI, Trace_ELBO, Predictive
from numpyro.infer.autoguide import AutoNormal, AutoDelta, AutoGuideList

N, dim = 3000, 3
data = random.normal(random.PRNGKey(0), (N, dim))
true_coefs = jnp.arange(1.0, dim + 1.0)
logits = jnp.sum(true_coefs * data, axis=-1)
labels = dist.Bernoulli(logits=logits).sample(random.PRNGKey(1))


def model(data, labels):
    coefs = numpyro.sample("coefs", dist.Normal(jnp.zeros(dim), jnp.ones(dim)))
    offset = numpyro.sample("offset", dist.Uniform(-1, 1))
    logits = offset + jnp.sum(coefs * data, axis=-1)
    return numpyro.sample("obs", dist.Bernoulli(logits=logits), obs=labels)


adam = optim.Adam(0.01)
rng_key_init = random.PRNGKey(0)

guide = AutoGuideList(model)
guide.append(
    AutoNormal(
        numpyro.handlers.block(numpyro.handlers.seed(model, rng_seed=0), hide=["coefs"])
    )
)
guide.append(
    AutoDelta(
        numpyro.handlers.block(
            numpyro.handlers.seed(model, rng_seed=1), expose=["coefs"]
        )
    )
)

svi = SVI(model, guide, adam, Trace_ELBO())
svi_result = svi.run(rng_key_init, 2000, data, labels)

params = svi_result.params

predictive = Predictive(guide, params=params, num_samples=1000)
[...]

PR #1669, opened 2023-11-06
Reference AutoGuideList in docs

tare:
I didn't remember to update the documentation in #1644. This PR updates the documentation so that `AutoGuideList` is mentioned.


fehiepsi (approved):
Nice catch! Thanks, @tare.

Issue #1672, opened 2023-11-09
[Bug] DiscreteHMCGibbs does not properly initialize discrete latents with init_params argument


amifalk:
Minimal working example:

```python
import jax.numpy as jnp
from jax.random import PRNGKey

import numpyro
import numpyro.distributions as dist
from numpyro.infer import MCMC, NUTS, DiscreteHMCGibbs

def model(data):
    with numpyro.plate('n_obs', 10):
        mu = numpyro.sample('mu', dist.Categorical(jnp.array([.25, .25, .25, .25])))
        numpyro.sample('dat', dist.Normal(mu, 1), obs=data)
        
mcmc = MCMC(DiscreteHMCGibbs(NUTS(model), modified=True), 
            num_warmup=1000,
            num_samples=100,
            num_chains=1)

data = dist.Normal(0, 1).sample(PRNGKey(0), (10,))
unconstrained_params = {'mu': jnp.array([0, 0, 0, 0, 0, 1, 1, 1, 1, 1])}

mcmc.run(PRNGKey(1), data, init_params=unconstrained_params)
```

yields the following error: 

```TypeError: grad requires real- or complex-valued inputs (input dtype that is a sub-dtype of np.inexact), but got int32. If you want to use Boolean- or integer-valued inputs, use vjp or set allow_int to True.```

I think the issue stems from the HMCGibbs init method, which DiscreteHMCGibbs inherits from. Below, you can see that the init_params arg doesn't consider that init_params can also be gibbs_sites.

```python
    def init(self, rng_key, num_warmup, init_params, model_args, model_kwargs):
        model_kwargs = {} if model_kwargs is None else model_kwargs.copy()
        if self._prototype_trace is None:
            rng_key, key_u = random.split(rng_key)
            # We use init strategy to get around ImproperUniform which does not have
            # sample method.
            self._prototype_trace = trace(
                substitute(seed(self.model, key_u), substitute_fn=init_to_sample)
            ).get_trace(*model_args, **model_kwargs)

        rng_key, key_z = random.split(rng_key)
        gibbs_sites = {
            name: site["value"]
            for name, site in self._prototype_trace.items()
            if name in self._gibbs_sites
        }
        model_kwargs["_gibbs_sites"] = gibbs_sites
        hmc_state = self.inner_kernel.init(
            key_z, num_warmup, init_params, model_args, model_kwargs
        )

        z = {**gibbs_sites, **hmc_state.z}

[...]

PR #1673, opened 2023-11-09
fix init_params bug in hmcgibbs

amifalk:
Fixes bug described in #1672 - params updated via gibbs were not initialized when init_params was set

### DIFF
--- a/numpyro/infer/hmc_gibbs.py
+++ b/numpyro/infer/hmc_gibbs.py
@@ -136,7 +136,7 @@ class HMCGibbs(MCMCKernel):
         gibbs_sites = {}
 
         for name, site in self._prototype_trace.items():
-            if init_params and name in init_params:
+            if init_params and (name in init_params) and (name in self._gibbs_sites):
                 gibbs_sites[name] = init_params.pop(name)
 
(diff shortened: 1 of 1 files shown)


numpyro/infer/hmc_gibbs.py
```diff
@@ -132,11 +132,17 @@ def init(self, rng_key, num_warmup, init_params, model_args, model_kwargs):
...
-            for name, site in self._prototype_trace.items()
-            if name in self._gibbs_sites
-        }
+
+        gibbs_sites = {}
+
+        for name, site in self._prototype_trace.items():
+            if init_params and name in init_params:
+                gibbs_sites[name] = init_params[name]
+                init_params.pop(name)
```
fehiepsi, line 141: I think you can combine those lines into `... = init_params.pop(name)`.

fehiepsi (approved):

amifalk:
While using the fix this morning, I realized the previous code did not account for init_params that are not gibbs sites, so I added an additional check.

fehiepsi:
Thanks for the fix, @amifalk!

PR #1679, opened 2023-11-14
Update continuous.py

### LINKED ISSUE #1634: Typo in docs
In the LKJ distribution, there is a code snippet with example [here](https://num.pyro.ai/en/stable/distributions.html#numpyro.distributions.continuous.LKJ)

In this example, the following line:
`# we can also use a faster formula `cov_mat = jnp.outer(theta, theta) * corr_mat``

Should be changed to
`# we can also use a faster formula `cov_mat = jnp.outer(sigma, sigma) * corr_mat``

etymology:
closes #1634. Changes a typo theta -> sigma.


fehiepsi (approved):
Good catch, thanks!

PR #1686, opened 2023-11-24
Fix faulty interaction between `jax.vmap` and `validate_args=True`

### LINKED ISSUE #1684: How to use vmap in the context of a model?
I was excited to see that numpyro now supports vmap. But I'm a little confused about what exactly can be done with it. When I do this:

```
import numpyro
import jax
from jax import numpy as jnp
dist = numpyro.distributions

loc = jnp.array([0,1])
scale = jnp.array([0,1])
d = jax.vmap(dist.Normal, in_axes=(0,0))(loc,scale)
key = jax.random.PRNGKey(0)
d.sample(key)
```

Then it works fine—I get a sample from a standard normal in two dimensions. However, when I try to do this:

```
import numpyro
import jax
from jax import numpy as jnp
dist = numpyro.distributions

loc = jnp.array([0,1])
scale = jnp.array([0,1])
def model():
    numpyro.sample("x",jax.vmap(dist.Normal, in_axes=(0,0))(loc,scale))

kernel = numpyro.infer.NUTS(model)
mcmc = numpyro.infer.MCMC(kernel, num_warmup=250, num_samples=750)
key = jax.random.PRNGKey(0)
mcmc.run(key)
```

Then I get `TypeError: '>' not supported between instances of 'object' and 'float'` which seem to be an error in `_GreaterThan.__call__` in `numpyro/distributions/constraints.py`.

Am I using `vmap` incorrectly here? Is there any other way that I can `vmap` distributions and then build numpyro models that I can do MCMC on, etc.?

pierreglaser:
Closes #1684 

@fehiepsi I changed the reproducer to a pattern that felt more prone to happen in practice than https://github.com/pyro-ppl/numpyro/issues/1684#issuecomment-1823149622:

```python
    v_dist = jax.vmap(
        lambda loc, scale: dist.Normal(loc=loc, scale=scale, validate_args=True),
        in_axes=(0, 0),
    )(jnp.zeros((2,)), jnp.zeros((2,)))
```

### DIFF
--- a/numpyro/distributions/distribution.py
+++ b/numpyro/distributions/distribution.py
@@ -158,10 +158,10 @@ class Distribution(metaclass=DistributionMeta):
         return all_pytree_data_fields
 
     @classmethod
-    def gather_pytree_aux_fields(cls):
+    def gather_pytree_aux_fields(cls) -> tuple:
         bases = inspect.getmro(cls)
 
-        all_pytree_aux_fields = ()
+        all_pytree_aux_fields = ("_validate_args",)
         for base in bases:
             if issubclass(base, Distribution):
                 all_pytree_aux_fields += base.__dict__.get("pytree_aux_fields", ())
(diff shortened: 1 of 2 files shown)


fehiepsi:
Nice fix! Thanks, @pierreglaser!

fehiepsi (approved):

PR #1687, opened 2023-11-24
Enforce supported python version in setup.py

### LINKED ISSUE #1656: Specify python version for pip
Can the python version be specified for pip in the setup.py?  Specifically using the method here:
- https://stackoverflow.com/questions/45778790/specifying-the-minimum-python-version-required-for-a-module-in-pip

i.e. `python_requires` so that it's picked up by pip

Currently, the update of this to drop 3.9 support broke a build of mine on a python 3.8 image.  I can get around this by pegging it to 0.12.1, but if python_requires is added, I believe pip will do this dependency resolution automatically.

pierreglaser:
Fixes #1656: `numpyro` has dropped `python 3.8` support since #1629. This PR ensures that `PyPI` will provide the correct version of `numpyro`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -86,6 +86,7 @@ setup(
         "tpu": f"jax[tpu]{_jax_version_constraints}",
         "cuda": f"jax[cuda]{_jax_version_constraints}",
     },
+    python_requires=">=3.9",
     long_description=long_description,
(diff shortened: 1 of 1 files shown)


fehiepsi (approved):
Thanks, @pierreglaser!

PR #1688, opened 2023-11-25
Use Cholesky decomp instead of inverting kernel

DanWaxman:
For the GP example, the training covariance matrix is currently inverted; it's faster and more numerically stable to use the Cholesky decomposition instead. It's a rather small change, but people tend to copy-paste examples, so I think a proper implementation is important here.

In the example, I've checked that the results are numerically similar (slightly different, I think in part to sampling): 
```python
vmap_args = (
    random.split(rng_key_predict, samples["kernel_var"].shape[0]),
    samples["kernel_var"],
    samples["kernel_length"],
    samples["kernel_noise"],
)
means, predictions, K = vmap(
    lambda rng_key, var, length, noise: predict(
        rng_key, X, Y, X_test, var, length, noise
    )
)(*vmap_args)

means_2, predictions_2, K_2 = vmap(
    lambda rng_key, var, length, noise: predict_2(
        rng_key, X, Y, X_test, var, length, noise
    )
)(*vmap_args)

mean_prediction = np.mean(means, axis=0)
mean_prediction_2 = np.mean(means_2, axis=0)
percentiles = np.percentile(predictions, [5.0, 95.0], axis=0)
percentiles_2 = np.percentile(predictions_2, [5.0, 95.0], axis=0)
print(
    "means np.allclose?",
    np.allclose(mean_prediction, mean_prediction_2),
    " | Max difference:",
    np.max(np.abs(mean_prediction - mean_prediction_2)),
)
print("Max difference in percentiles:", np.max(np.abs(percentiles - percentiles_2)))
```
gives
```
>>> means close? False  | Max difference: 4.529953e-06
>>> Max difference in percentiles: 0.02630336433649072
```

### DIFF
--- a/examples/gp.py
+++ b/examples/gp.py
@@ -98,18 +98,28 @@ def run_inference(model, args, rng_key, X, Y):
 
 
 # do GP prediction for a given set of hyperparameters. this makes use of the well-known
-# formula for gaussian process predictions
-def predict(rng_key, X, Y, X_test, var, length, noise):
+# formula for Gaussian process predictions
+def predict(rng_key, X, Y, X_test, var, length, noise, use_cholesky=True):
     # compute kernels between train and test data, etc.
     k_pp = kernel(X_test, X_test, var, length, noise, include_noise=True)
     k_pX = kernel(X_test, X, var, length, noise, include_noise=False)
     k_XX = kernel(X, X, var, length, noise, include_noise=True)
-    K_xx_cho = jax.scipy.linalg.cho_factor(k_XX)
-    K = k_pp - jnp.matmul(k_pX, jax.scipy.linalg.cho_solve(K_xx_cho, k_pX.T))
+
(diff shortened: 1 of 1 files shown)


DanWaxman:
Done, thanks @fehiepsi! I made `use_cholesky=True` the default, and used command line option `--no_cholesky` due to the well-known issue of parsing `bool`s using `argparse`.

examples/gp.py
```diff
@@ -184,6 +194,7 @@ def main(args):
         type=str,
         choices=["median", "feasible", "value", "uniform", "sample"],
     )
+    parser.add_argument("--no_cholesky", dest="use_cholesky", action="store_false")
```
fehiepsi, line 197: maybe `no-cholesky` for consistency?

DanWaxman: Fair enough, done!

fehiepsi (approved):
Thanks for the improvement, @DanWaxman!!

Issue #1691, opened 2023-11-30
[FR] add affine-invariant ensemble sampling (including DE-MCMC) and ensemble slice sampling


amifalk:
as implemented in [emcee](https://emcee.readthedocs.io/en/stable/) and [zeus](https://zeus-mcmc.readthedocs.io/en/latest/index.html) - both methods are gradient free. For likelihood free models, differential evolution monte carlo (DE-MCMC) tends to be [robust](https://www.sciencedirect.com/science/article/abs/pii/S0022249616301663).

I have minimal working drafts of the two classes [here](https://github.com/amifalk/numpyro/blob/demc/numpyro/infer/ensemble.py) with utility functions [here](https://github.com/amifalk/numpyro/blob/demc/numpyro/infer/ensemble_util.py). I tried to make the api align with emcee and zeus where possible. 

Also, ```batch_ravel_pytree()``` in ensemble_util.py should make it easier to implement future mcmc methods that share information between chains.

Let me know what you think!

fehiepsi:
Thanks, @amifalk! It's great to have a likelihood-free inference algorithm in NumPyro. Welcome to contribute!

Issue #1192, opened 2021-10-14
Allow mixing vectorized and parallel sampling


hessammehr:
Currently the `"vectorized"` and `"parallel"` values for the `chain_methods` parameter of `MCMC(...)` have mutually exclusive outcomes. The former can run multiple chains in parallel but on a single GPU/TPU while the latter can utilize multiple GPU/TPU but only runs one parallel chain on each.

It would be useful to have the option of doing both at the same time, e.g. 2 GPUs each running 4 vectorized chains in parallel. Perhaps this could be made explicit by passing `num_chains=(2, 4)`, i.e. 2 parallel devices; 4 vectorized chains on each. If this does sound like a desirable feature I should be able to implement it with a little bit of help/advice.

PR #1692, opened 2023-11-30
Add two ensemble sampling methods

amifalk:
As described in #1691 

Currently, I've only implemented a subset of emcee and zeus moves, but it should be trivial to extend in the future. I also don't have support for potential_fn.

Should there be separate tests for these modules or should I try to work them in to existing ones? The pattern isn't terribly clean with existing tests because AIES and ESS can only be run with multiple chains.


numpyro/infer/ensemble.py
```diff
@@ -0,0 +1,724 @@
...
+        approximation of the walkers of the complementary ensemble.
+        """
+        def gaussian_move(rng_key, inactive, mu):
+            n_active_chains, n_params = inactive.shape
+            cov = jnp.cov(inactive, rowvar=False)
+
+            return (
+                2.0
+                * mu
+                * dist.MultivariateNormal(0, cov).sample(
```
martinjankowiak, line 542: haven't looked through your code in detail but i believe you can get the MVN sample much more cheaply by "coloring" (i.e. multiplying) a vector of `N(0,1)` normal noise by `inactive`. if you use `MultivariateNormal` like this it will compute a cholesky decomposition at cubic cost

amifalk: These don't seem to be sampling from the same distribution - am I misunderstanding what you meant?

```python
import numpy as np

n_obs = 1_000
n_dim = 2

particles = np.random.exponential(10, (n_obs, n_dim))
cov = np.cov(particles, rowvar=False)

cholesky_samp = np.random.multivariate_normal(np.mean(particles, axis=0),
                                              cov=cov,
                                              size=(n_obs,))


cheaper_samp = particles * np.random.standard_normal((n_obs, n_dim))

plt.scatter(cholesky_samp[:, 0], cholesky_samp[:, 1], alpha=.3)
plt.scatter(cheaper_samp[:, 0], cheaper_samp[:, 1], alpha=.3)
```
8e676227-5ac4-4d94-adfe-d782e19fea93

amifalk: It seems like avoiding the cholesky decomposition is only faster in high dimensions (for low enough n_obs), but gradient-free methods tend to fail there anyway. Do you think it's worth special casing it?

85573bf1-2da3-4d8e-b9e8-9abde8e1d913
816ed6d7-f5dc-4ff8-a2ec-2fd6a6e3f6cf

```python
#%%
from functools import partial
import jax
import jax.numpy as jnp
import jax.random as jrandom

@partial(jax.jit, static_argnums=2)
def no_chol_mvn(rng_key, data, n_samps):
    n_obs, n_dim = data.shape
    eps = jrandom.normal(rng_key, (n_samps, n_obs))

    return eps @ (data - jnp.mean(data, axis=0)) / jnp.sqrt(n_obs)

@partial(jax.jit, static_argnums=2)
def chol_mvn(rng_key, data, n_samps):
    cov = jnp.cov(data, rowvar=False, bias=True)

    return dist.MultivariateNormal(0, cov).sample(rng_key, (n_samps,))

n_obs = n_samps = jnp.geomspace(10, 10_000, num=20, dtype=jnp.int32) 
n_dim = 20

no_chol_timings = []
chol_timings = []

for obs, samps in zip(n_obs, n_samps):
    data = jrandom.exponential(jrandom.PRNGKey(0), (obs, n_dim))

    no_chol_res = %timeit -o no_chol_mvn(jrandom.PRNGKey(1), data, samps.item()).block_until_ready()
    chol_res = %timeit -o chol_mvn(jrandom.PRNGKey(1), data, samps.item()).block_until_ready()

    no_chol_timings.append(no_chol_res)
    chol_timings.append(chol_res)    

no_chol_timings = [sum(res.timings) / len(res.timings) for res in no_chol_timings]
chol_timings = [sum(res.timings) / len(res.timings) for res in chol_timings]

#%%
plt.scatter(n_obs[:15], no_chol_timings[:15], label='no_chol')
plt.scatter(n_obs[:15], chol_timings[:15], label='chol')
plt.title(f'time vs. n_obs/n_samps, n_dim={n_dim}')
plt.legend();
```

martinjankowiak: yeah if you only anticipate running this algorithm in up to ~50-100 dimensions it's probably not worth it but you can at least leave a comment in the code

martinjankowiak:
@amifalk hello what is the status of this PR? are you waiting for feedback from us? i think we sort of lost track of this over the holiday break.

amifalk:
@martinjankowiak Would you mind looking through it and making comments about what you feel is missing/needs to change? I believe all of the code lints with the exception of the batch_ravel_pytree function in `ensemble_utils.py`. I modified that from [JAX](https://jax.readthedocs.io/en/latest/_autosummary/jax.flatten_util.ravel_pytree.html) so I was reluctant to change the formatting, but I'm happy to do what's needed to make it lint - or if you have ideas of a better solution. 

At the core, these are gradient free methods that update the state of each chain by looking at the current state of the other chains, which I've implemented by storing an (n_chains, n_params) array (thus the need to use write a version of `ravel_pytree` that anticipates a batch dimension).

I'm also happy to reduce the scope of the PR if that will make it easier to review (the affine invariant ensemble sampler, AIES, is much less complex than the ensemble slice sampler, ESS).

martinjankowiak (commented):
so i'm not familiar with these algorithms in detail so it's a bit hard for me to follow all the details but the code is generally very clean and readable : )

i think the main thing that is missing is tests:
- you can put a `test_ensemble_util.py` in `tests/infer`
- you should add at least one test that does inference on a conjugate model where you can e.g. compute analytic posterior means/variances; for example you can add your methods to a few of the following tests: `test_mcmc.py::test_unnormalized_normal_x64`, `test_logistic_regression_x64`, `test_beta_bernoulli_x64`
- you have a fair number of different ways to initialize your kernels so you probably want to add some simple smoke tests or the like to `test_ensemble_mcmc.py` or similar. e.g. make sure that various combinations of init args do not error out. make simple checks about expected shapes of outputs and the like. you can also use the `with pytest.raises(ValueError, match="..expected message...")` context manager to check that some invalid initializations and the like are being caught as expected.

numpyro/infer/ensemble_util.py
```diff
@@ -0,0 +1,110 @@
...
+import jax.numpy as jnp
+
+from jax._src import dtypes
+from jax._src.tree_util import tree_flatten, tree_unflatten
+from jax._src.util import safe_zip, unzip2, HashablePartial
+
+zip = safe_zip
+
+
+def _get_nondiagonal_pairs(n):
```
martinjankowiak, line 18: @fehiepsi do you have any suggestions on the utility methods in this file? can any of this functionality be covered by existing utils?

numpyro/infer/ensemble.py
```diff
@@ -0,0 +1,735 @@
...
+
+
+class EnsembleSampler(MCMCKernel, ABC):
+    """
+    Abstract class for ensemble samplers.
+    
+    :param model: Python callable containing Pyro :mod:`~numpyro.primitives`.
+        If model is provided, `potential_fn` will be inferred using the model.
+    :param potential_fn: XXX currently unsupported.
+    :param bool randomize_split: whether or not to permute the chain order at each iteration.
```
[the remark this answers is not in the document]
amifalk: Whether or not permuting the chain order is a good idea depends on the subclass. I'll opt to keep this abstract here and specify defaults below.

amifalk:
Thanks for the review @martinjankowiak! Documentation has been updated and tests have been added (a distribution test seems to be failing for reasons unrelated to this PR). If there are any more comments and/or if @fehiepsi has a better solution for `batch_ravel_pytree`, I'm happy to make corresponding fixes.

numpyro/infer/ensemble.py
```diff
@@ -0,0 +1,735 @@
...
+        if not moves:
+            self._moves = [AIES.DEMove()]
+            self._weights = jnp.array([1.0])
+        else:
+            self._moves = list(moves.keys())
+            self._weights = jnp.array([weight for weight in moves.values()]) / len(moves)
+
+        super().__init__(model, potential_fn, randomize_split, init_strategy)
+
+    # XXX: this doesn't show because state_method='vectorized' shuts off diagnostics_str
```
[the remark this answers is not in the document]
fehiepsi: Is this `chain_method` @amifalk? I don't think progress_bar works with the vectorized method there.

numpyro/infer/ensemble_util.py
```diff
@@ -0,0 +1,113 @@
...
+    rows, cols = np.tril_indices(n, -1)  # -1 to exclude diagonal
+
+    # Combine rows-cols and cols-rows pairs
+    pairs = np.column_stack([np.concatenate([rows, cols]),
+                             np.concatenate([cols, rows])])
+
+    return jnp.asarray(pairs)
+
+
+def batch_ravel_pytree(pytree):
```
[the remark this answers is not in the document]
amifalk: Thanks, this is much cleaner!

numpyro/infer/ensemble.py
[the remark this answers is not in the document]
amifalk: Yes it is, sorry for the typo. Can we adjust [this](https://github.com/pyro-ppl/numpyro/blob/3ef0423ae96d3b21bc0bc09d497fc772e7db8dc0/numpyro/infer/mcmc.py#L427 ) line to allow the diagnostics_str for these ensemble methods? If I remove the prng_key check, it displays correctly.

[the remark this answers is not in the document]
fehiepsi: That line checks whether `vectorized` method is used. If rng_key.ndim == 1, parallel is used. Otherwise, vectorized is used. Looking at your code, it seems that rng_key.ndim == 2 in both cases, which is strange to me. Could you double check the logic for both methods?

I think you can add an attribute to MCMCKernel to indicate whether it is an ensemble kernel and skip `is_prng_key` check at that line if the kernel is an ensemble one.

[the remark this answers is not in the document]
amifalk: Good catch. Parallel sampling does not work with pmap since the chains need to talk to each other. To make that work, we would probably need to add an internal utility that does parallel sharding on the chains, kind of like what is requested in #1192. In the interest of not overcomplicating the API, I'll opt not to support it for now and make a note in the code, but if there was sufficient interest I think we could add a `shard_chains` argument to `EnsembleSampler`.

fehiepsi (approved):
Looks great to me! Could you add the classes to docs/source/mcmc file?

test/infer/test_mcmc.py
```diff
@@ -34,16 +34,27 @@ def test_unnormalized_normal_x64(kernel_cls, dense_mass):
...
+
+        init_params = random.normal(random.PRNGKey(1), (num_chains,))
+        mcmc = MCMC(
+            kernel, num_warmup=num_warmup, num_samples=num_samples, progress_bar=False,
+            num_chains=num_chains, chain_method='vectorized'
+        )
     else:
-        kernel = kernel_cls(
-            potential_fn=potential_fn, trajectory_length=8, dense_mass=dense_mass
+        if kernel_cls in [SA, BarkerMH]:
```
martinjankowiak, line 47: nit: could you unnest this so we have something like:
```
if kernel_cls in [AIES, ESS]:
    ....
elif kernel_cls in [SA, BarkerMH]:
    ....
```

martinjankowiak (approved):
looks great thanks!

can you check how much additional time your tests are adding?
i don't know how slow the additions to `test_beta_bernoulli_x64` and `test_logistic_regression_x64` might be. if they're adding too much time we may want to add something like `pytest.mark.skipif("CI" in os.environ, reason="reduce time for CI")`

amifalk:
`test_beta_bernoulli_x64` and `test_logistic_regression_x64` were quite slow (around 20-30 seconds on my machine for each sampler) so I opted to skip them. Everything should be building correctly now!

martinjankowiak:
@amifalk thanks! it might be good to keep one of the tests if it only adds ~50 seconds in total.... what do you think @fehiepsi ?

docs/source/mcmc.rst
```diff
@@ -9,7 +9,9 @@ We provide a high-level overview of the MCMC algorithms in NumPyro:
...
-* `SA <https://num.pyro.ai/en/latest/mcmc.html#numpyro.infer.sa.SA>`_ is the only MCMC method in NumPyro that does not leverage gradients. It is only applicable to models with continuous latent variables. It is expected to perform best for models whose latent dimension is low to moderate. It may be a good choice for models with non-differentiable log densities. Note that SA generally requires a *very* large number of samples, as mixing tends to be slow. On the plus side individual steps can be fast.
+* `SA <https://num.pyro.ai/en/latest/mcmc.html#numpyro.infer.sa.SA>`_ does not leverage gradients. It is only applicable to models with continuous latent variables. It is expected to perform best for models whose latent dimension is low to moderate. It may be a good choice for models with non-differentiable log densities. Note that SA generally requires a *very* large number of samples, as mixing tends to be slow. On the plus side individual steps can be fast.
```
martinjankowiak, line 12: can we change this to:

> is a gradient-free MCMC method. . . . .

fehiepsi:
@amifalk Could you help fix the failing test by adding a catch at this [line](https://github.com/pyro-ppl/numpyro/blob/master/test/test_distributions.py#L1747)?
```
except AttributeError, ValueError:
```

martinjankowiak:
thanks @amifalk ! btw what's your interest in these algorithms? do you have a non-differentiable log density? because unless the problem is very multi-modal or otherwise difficult HMC should work pretty well if a gradient is available. also have you tried comparing these algorithms to `SA`?

amifalk:
> thanks @amifalk ! btw what's your interest in these algorithms? do you have a non-differentiable log density? because unless the problem is very multi-modal or otherwise difficult HMC should work pretty well if a gradient is available. also have you tried comparing these algorithms to `SA`?

@martinjankowiak We work with some likelihood-free models like [LCA](https://www.sciencedirect.com/science/article/abs/pii/S0022249616301663) that we perform approximate inference on through simulation. We actually did try SA first, but it didn't seem robust to the noisy likelihood.

xiesl97:
hello, does `AIES` supports `chain_method='parallel'`?

Issue #1685, opened 2023-11-24
Scan doesn't work with replay


deoxyribose:
Given a model adapted from the scan doctest:

```python
def gaussian_hmm(y=None, T=T, N=N):
    def transition(x_prev, y_curr):
        with numpyro.plate('data', N):
            x_curr = numpyro.sample('x', dist.Normal(x_prev, 1.5))
            y_curr = numpyro.sample('y', dist.Normal(x_curr, 0.3), obs=y_curr)
        return x_curr, (x_curr, y_curr)
    with numpyro.plate('data', N):
        x0 = numpyro.sample('x_0', dist.Normal(jnp.zeros(N), 5.))
    _, (x, y) = scan(transition, x0, y, length=T)
    return (x, y)
    
with numpyro.handlers.seed(rng_seed=0):
    x, y = gaussian_hmm()

svi = SVI(gaussian_hmm, custom_guide, Adam(0.1), Trace_ELBO(num_particles=1), y=y, T=T, N=N)

n_steps = 5000
results = svi.run(random.PRNGKey(0), n_steps)
```

HMC gives the expected results. SVI runs without errors, but gives bad results, because the ELBO is not computed correctly. The scan operator blocks all handlers except substitute and condition, including the replay effect handler used in ELBO.

A preliminary implementation can be seen here https://github.com/pyro-ppl/numpyro/compare/master...aleatory-science:numpyro:master, which uses a similar logic to the way substitute is implemented for scan, and reuses the `substitute_stack` to store the replay trace.

deoxyribose:
I could make a PR as it is, but note that `scan_enum` hasn't been correspondingly edited (if it needs to be), and I'm not sure if and how the implementation otherwise needs to changed, or tested. @fehiepsi do you want to review what's there now, or is there something else I can add first?

fehiepsi:
Thanks @deoxyribose !

PR #1693, opened 2023-12-04
Support scan for Trace_ELBO

deoxyribose:
Issue #1685

Reuse the `substitute_stack` to store the replay trace, and replay the sites in the scan'ed function one iteration at a time.

### DIFF
--- a/numpyro/contrib/control_flow/scan.py
+++ b/numpyro/contrib/control_flow/scan.py
@@ -26,10 +26,11 @@ def _replay_wrapper(replay_trace, trace, i, length):
         ):
             return site
 
+        site = site.copy()
         site["value"] = site["value"][i]
         return site
 
-    return {k: get_ith_value(v.copy()) for k, v in replay_trace.items()}
+    return {k: get_ith_value(v) for k, v in replay_trace.items()}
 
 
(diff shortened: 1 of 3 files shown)


fehiepsi:
Nice support, @deoxyribose! Could you add tests for this change?

deoxyribose:
I've added a test that does SVI with an AutoNormal guide, and checks that results are relatively accurate. I've kept model and data size to a minimum, the test takes around 3-4 seconds on my machine. There already was a test combining scan, SVI and AutoNormal (test_subsample_guide in test/infer/test_autoguide.py), but it doesn't run inference.

numpyro/contrib/control_flow/scan.py
```diff
@@ -15,6 +15,23 @@
...
+            site["name"] not in trace
+            or site_len != length
+            or site["type"] not in ("sample", "deterministic")
+        ):
+            return site
+
+        site["value"] = site["value"][i]
+        return site
+
+    return {k: get_ith_value(v.copy()) for k, v in replay_trace.items()}
```
fehiepsi, line 32: nit: it is better to do the copy inside `git_ith_value`

fehiepsi (approved):
Looks great! Thanks, @deoxyribose !

PR #1700, opened 2023-12-19
Add ruff support

juanitorduz:
This PR proposes adding ruff as a linter and formater. See https://github.com/pyro-ppl/pyro/pull/3231

- I made the configuration very explicit, see https://github.com/astral-sh/ruff?tab=readme-ov-file#configuration

- Just very minor changes to the code style were needed.

- Locally,  `make lint` and `make format` work as expected.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,8 +34,8 @@ exclude = [
 line-length = 88
 indent-width = 4
 
-# Assume Python 3.8
-target-version = "py38"
+# Assume Python 3.9 (minimum supported version).
+target-version = "py39"
 
 [tool.ruff.lint]
 # Enable Pyflakes (`F`) and a subset of the pycodestyle (`E`)  codes by default.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,7 @@
+[project]
+name = "numpyro"
+requires-python = ">=3.9"
+
 [tool.ruff]
 # Exclude a variety of commonly ignored directories.
 exclude = [
@@ -33,9 +37,6 @@ exclude = [
 # Same as Black.
 line-length = 88
 indent-width = 4
-
-# Assume Python 3.9 (minimum supported version).
-target-version = "py39"
 
 [tool.ruff.lint]
 # Enable Pyflakes (`F`) and a subset of the pycodestyle (`E`)  codes by default.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,67 @@
+[tool.ruff]
+# Exclude a variety of commonly ignored directories.
+exclude = [
+    ".bzr",
+    ".direnv",
+    ".eggs",
+    ".git",
+    ".git-rewrite",
+    ".hg",
+    ".ipynb_checkpoints",
+    ".mypy_cache",
+    ".nox",
+    ".pants.d",
+    ".pyenv",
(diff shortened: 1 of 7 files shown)


pyproject.toml
```diff
@@ -0,0 +1,67 @@
...
+    "build",
+    "dist",
+    "docs/src",
+    "node_modules",
+    "site-packages",
+    "venv",
+]
+
+# Same as Black.
+line-length = 120
```
fehiepsi, line 34: I think this is 88.

pyproject.toml
```diff
@@ -0,0 +1,67 @@
...
+indent-width = 4
+
+# Assume Python 3.8
+target-version = "py38"
```
fehiepsi, line 38: Would this cause any trouble? I think the version depends on user's system.

pyproject.toml
juanitorduz: ok! fixed in https://github.com/pyro-ppl/numpyro/pull/1700/commits/6b11e07577ab96c0569a2b9463c78d17fa2eb06e

pyproject.toml
juanitorduz: According to https://github.com/astral-sh/ruff/issues/2519#issuecomment-1467101887, is ok to indicate the minimum version, which according to `setup.py` is `3.9`, so I added it in https://github.com/pyro-ppl/numpyro/pull/1700/commits/b2292eef29fd517219f3d15a78aa69524455a68c

fehiepsi: In https://github.com/astral-sh/ruff/blob/main/CHANGELOG.md#015 it's recommended to use `project.requires-python`

setup.cfg
```diff
@@ -1,23 +1,3 @@
...
-known_first_party = funsor, numpyro, test
-known_third_party = opt_einsum
-known_jax = flax, haiku, jax, optax, tensorflow_probability
-sections = FUTURE, STDLIB, THIRDPARTY, JAX, FIRSTPARTY, LOCALFOLDER
-force_sort_within_sections = true
-combine_as_imports = true
-multi_line_output = 3
-skip=docs
-
 [tool:pytest]
```
fehiepsi, line 1: Do you think that we can also move pytest config to the pyproject file. I saw it is supported [here](https://docs.pytest.org/en/7.3.x/reference/customize.html#pyproject-toml). (not necessary in this PR though)

juanitorduz: Indeed! I want to do this in another PR. In addition I wanna add pre-commit hooks :) (also in a different PR)

pyproject.toml
juanitorduz: Changed in https://github.com/pyro-ppl/numpyro/pull/1700/commits/442353e2c26659c0d062728385f26721a8210508

fehiepsi (approved):
This is super helpful! Thanks @juanitorduz!

juanitorduz:
I will continue with other PRs to remove setup.cfg and add pre-commit hooks :)

fehiepsi:
The change seems to cause the master branch to [fail](https://github.com/pyro-ppl/numpyro/actions/runs/7299291844/job/19891830200). Could you take a look?

juanitorduz:
@fehiepsi This should fix it https://github.com/pyro-ppl/numpyro/pull/1703 .

PR #1703, opened 2023-12-22
Hot fix: remove version requirement

juanitorduz:
Follow up of https://github.com/pyro-ppl/numpyro/pull/1700. The `[project]` config needs a version, but we are not using it for the package itself. We can safely simply remove it.


fehiepsi (approved):

fehiepsi:
Thanks, @juanitorduz!

PR #1704, opened 2023-12-22
Remove `setup.cfg`

juanitorduz:
As discussed in https://github.com/pyro-ppl/numpyro/pull/1700, we would like to adopt `pyproject.toml` so we can remove `setup.cfg` after moving the `pytest` options. 

I also do some minor improvements to the CI.


fehiepsi (approved):
Thanks, @juanitorduz!

PR #1705, opened 2023-12-22
Support optional pre-commit hooks

juanitorduz:
For local development, [pre-commit](https://pre-commit.com/) hooks are very useful so that the contributor can check the code formatting before every-commit. 

For now, this is optional and does not intervene with the CI. In the future, you might want to use the pre-commit configuration as the linter so that everything is in sync and the versions of ruff are not open. See, for example, https://github.com/pymc-labs/pymc-marketing/blob/main/.github/workflows/ci.yml#L25-L26. If you ever decide to go this path I will be happy to support :)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,6 +4,7 @@ repos:
       - id: ruff
         name: ruff
         entry: ruff
+        args: ["--fix", "--show-source"]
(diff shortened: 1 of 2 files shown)


fehiepsi (approved):
woohoo, this would save time dealing with lint issues. thanks!

Issue #1710, opened 2023-12-27
numpyro.render_model render unexpected graph with handler.condition


yayami3:
The reproduced code is as follows.

```
import numpyro
import numpyro.distributions as dist

from numpyro.handlers import condition

def model(y=None):
    alpha = numpyro.sample("alpha", dist.Normal(0., 1.))
    beta = numpyro.sample("beta", dist.Normal(alpha, 1.))
    gamma = numpyro.sample("gamma", dist.HalfNormal(beta))

    return gamma
    
cond = {
    "beta": 0.,
}
model_cond = condition(model, cond)

numpyro.render_model(model_cond, filename="model_cond_asis.png")

```

This code render the graph below. In this case, the edge between A and B would be cut off.

![model_cond_asis](https://github.com/pyro-ppl/numpyro/assets/116920988/0a9539bf-aa1b-4ad7-a1a4-702e25d678c9)

When removing `and not site["is_observed"]` from https://github.com/pyro-ppl/numpyro/blob/09551d99bf990e5e4f3afdb9c96aadf87780c2b5/numpyro/infer/inspect.py#L388
, it works for my case. Output graph is below.
![model_cond](https://github.com/pyro-ppl/numpyro/assets/116920988/629e869e-d69a-4bd4-a373-4aa2c94035ee)

Is there any case it shouldn't work without `and not site["is_observed"]`  ?

fehiepsi:
I think your suggestion is right. A typical model has no variable depending on "observed" ones, so the observed variables are eliminated from the dependencies. Please feel free to make a PR for the fix.

PR #1711, opened 2023-12-29
Change case to render graph for conditioned nodes.

yayami3:
Solves #1710 .
Conditioned nodes were to be isolated so far.
This change makes the graph what it is supposed to be.


fehiepsi (approved):
Thanks @yayami3! glad that all tests pass.

yayami3:
Thanks @fehiepsi
Happy new week!

PR #1714, opened 2024-01-05
Constrain values for AutoDelta.median()

### LINKED ISSUE #1712: [FR] constrain values for AutoDelta.median()
In a similar vein to #951. All Normal guides and guides that inherit from AutoContinuous automatically constrain the parameters before returning them when median() is called. AutoDelta is the only exception.

Internally, this is handled by calling `self._postprocess_fn` on the guide params, which removes the auto_loc boilerplate attached to every variable and returns deterministic sites as a nice side effect. This would be a convenient feature to have in AutoDelta (and I'm happy to add it).

amifalk:
Fixes #1712. Note that in order to apply _postprocess_fn correctly, each Delta param is constrained using `TransformedDistribution` instead of  `param.constraints`.


fehiepsi:
Oops, sorry, I thought you needed the `AutoDelta.median` method. I'll need to rethink your proposal because this changes behavior.

PR #1718, opened 2024-01-09
Add isort (ruff)

juanitorduz:
I just realized `isort` was not in the default ruff config 🤦 . Sorry!

Related to https://github.com/pyro-ppl/numpyro/pull/1700


fehiepsi (approved):
Thanks, @juanitorduz !

PR #1720, opened 2024-01-15
Improve current ruff rules

juanitorduz:
I keep improving the changes introduced in https://github.com/pyro-ppl/numpyro/pull/1700/. Note that in the `setup.cfg` file we had `max-line-length = 120`. This PR updates this condition and adds all the remaining `pyflake` rules.  Note this is very similar to `pyro`'s config https://github.com/pyro-ppl/pyro/blob/dev/pyproject.toml

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,8 +35,6 @@ line-length = 120
 indent-width = 4
 
 [tool.ruff.lint]
-# Enable Pyflakes (`F`) and a subset of the pycodestyle (`E`)  codes by default.
-# We also add isort.
 select = ["E", "F", "I", "W"]
 ignore = ["E203"]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,7 @@ exclude = [
 ]
 
 # Same as Black.
-line-length = 120
+line-length = 88 
 indent-width = 4
 
 [tool.ruff.lint]
@@ -44,6 +44,9 @@ unfixable = []
 
 # Allow unused variables when underscore-prefixed.
 dummy-variable-rgx = "^(_+|(_+[a-zA-Z0-9_]*[a-zA-Z0-9]+?))$"
+
+[tool.ruff.lint.pycodestyle]
+max-line-length = 120
 
 [tool.ruff.format]
 # Like Black, use double quotes for strings.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,7 @@ exclude = [
 ]
 
 # Same as Black.
-line-length = 88 
+line-length = 88
 indent-width = 4
 
 [tool.ruff.lint]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,13 +31,13 @@ exclude = [
 ]
 
 # Same as Black.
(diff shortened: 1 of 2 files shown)


juanitorduz:
The tests are not failing because of `ruff` but because of the docs:

```bash
make -C docs html
make[1]: Entering directory '/home/runner/work/numpyro/numpyro/docs'
Running Sphinx v4.5.0
/opt/hostedtoolcache/Python/3.9.1[8](https://github.com/pyro-ppl/numpyro/actions/runs/7531465780/job/20500091407?pr=1720#step:6:9)/x64/lib/python3.[9](https://github.com/pyro-ppl/numpyro/actions/runs/7531465780/job/20500091407?pr=1720#step:6:10)/site-packages/nbsphinx.py:1450: RuntimeWarning: You are using an unsupported version of pandoc (2.9.2.1).
Your version must be at least (2.14.2) but less than (4.0.0).
Refer to https://pandoc.org/installing.html.
Continuing with doubts...
  nbconvert.utils.pandoc.check_pandoc_version()
loading translations [en]... done

Sphinx version error:
The sphinxcontrib.applehelp extension used by this project needs at least Sphinx v5.0; it therefore cannot be built with this version.
make[1]: *** [Makefile:20: html] Error 2
make[1]: Leaving directory '/home/runner/work/numpyro/numpyro/docs'
make: *** [Makefile:26: docs] Error 2
Error: Process completed with exit code 2.
```

juanitorduz:
A suggested fix: https://github.com/pyro-ppl/numpyro/pull/1721

pyproject.toml
```diff
@@ -31,13 +31,11 @@ exclude = [
 ]
 
 # Same as Black.
-line-length = 88
+line-length = 120
```
fehiepsi, line 34: I guess we prefer 88 here

juanitorduz: I think before we started the migration to ruff we had 120 (see original PR). 
I can of course change it to 88. I might need to format some code a bit if I remember correctly. Is that on ?

juanitorduz: If I set `88` I get `Found 821 errors` (this comes from the 120 rule I mentioned). I could either keep `120` or set it to `88` and ignore the `E501` rule as it is very hard to fix 821 errors :)

fehiepsi: Previously, we allowed doc length 120 and code length 88. Do you think that we can maintain that behavior? If we set 120 here, will long code be formatted?

juanitorduz: If we set 120 there is not refactor needed at all. I guess we can simply add 88 and ignore the rule for docstrings

fehiepsi: Seems like
```
[tool.ruff.lint.pycodestyle]
max-line-length = 120
```
does the job

juanitorduz: Indeed! Thanks 💪 ! See https://github.com/pyro-ppl/numpyro/pull/1720/commits/493da8eff5fdee5ee33c6d4f64883e456b1dcff6

fehiepsi (approved):
Thanks, @juanitorduz!

PR #1727, opened 2024-01-30
Fix import for jaxns compatibility

bmorris3:
I ran the [nested sampling for gaussian shells](https://num.pyro.ai/en/stable/examples/gaussian_shells.html) tutorial which depends on `jaxns`, and I found that importing anything from `numpyro/contrib/nested_sampling.py` produces an error. The error occurs because the `NestedSamplerResults` class is not in the top-level `jaxns` namespace. 

This PR simply imports from the correct submodule.

### DIFF
--- a/numpyro/contrib/nested_sampling.py
+++ b/numpyro/contrib/nested_sampling.py
@@ -10,7 +10,6 @@ try:
     from jaxns import (
         ExactNestedSampler as OrigNestedSampler,
         Model,
-        NestedSamplerResults,
         Prior,
         TerminationCondition,
         plot_cornerplot,
@@ -18,6 +17,7 @@ try:
         resample,
         summary,
     )
+    from jaxns.internals.types import NestedSamplerResults
 except ImportError as e:
     raise ImportError(
         "To use this module, please install `jaxns` package. It can be"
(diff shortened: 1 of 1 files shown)


bmorris3:
I found some other lines that needed to be updated for jaxns (2.4.7). After [212baee](https://github.com/pyro-ppl/numpyro/pull/1727/commits/212baee9941c8db3b5b1595d2ea85729ea363391), you can run the [tutorial python script](https://num.pyro.ai/en/stable/examples/gaussian_shells.html) on the  numpyro docs without modification with numpyro dev.

fehiepsi:
Thanks, @bmorris3! We currently only support [jaxns=2.2.6](https://github.com/pyro-ppl/numpyro/blob/master/setup.py#L63). Please pin the version to a new one if you want to use the latest jaxns.

fehiepsi:
@bmorris3 We plan to make a new release this week. Do you want to pump jaxns versions in [setup](https://github.com/pyro-ppl/numpyro/blob/master/setup.py#L63) and [docs requirements](https://github.com/pyro-ppl/numpyro/blob/master/docs/requirements.txt#L7)?

fehiepsi:
Hi @bmorris3, this is fixed in https://github.com/pyro-ppl/numpyro/pull/1738. numpyro will be compatible with jaxns 2.4.8 in the next release.

Issue #1724, opened 2024-01-26
Incorrect elbo normalisation when multi_sample_guide is True


danielward27:
Hello,

Was having a read of the ELBO implementation and it looks like there is a (minor) bug.

If ``self.multi_sample_guide`` is True, then ``elbo_particle``, is defined at https://github.com/pyro-ppl/numpyro/blob/01089cfad4078e21e062c02147bcc636331548f1/numpyro/infer/elbo.py#L194 and overwritten without normalisation by the seeds at https://github.com/pyro-ppl/numpyro/blob/01089cfad4078e21e062c02147bcc636331548f1/numpyro/infer/elbo.py#L214

If ``self.multi_sample_guide`` is False, then it is again overwritten at the same place, but this time with the same value. I can submit a pull request with line 214 deleted if that's all that is needed.

fehiepsi:
Good catch! Thanks, @danielward27!

PR #1728, opened 2024-02-04
fix elbo normalization with multi_sample_guide=True

danielward27:
Fixes https://github.com/pyro-ppl/numpyro/issues/1724


fehiepsi (approved):
Thanks, @danielward27!

Issue #1726, opened 2024-01-30
Support forward mode differentiation for SVI


AndreaSalati:
Hello everybody. I am encountering a problem with the VonMises distribution, and in particular with its concentration parameter. I am trying to perform a very simple MLE of a hierarchical model.

```
def model(X):
    
    plate =  numpyro.plate("data", Nc)
    kappa = numpyro.param("kappa", 1., constraint = numpyro.distributions.constraints.positive)
    mu = numpyro.param("mu", 0., constraint = numpyro.distributions.constraints.interval(-np.pi, np.pi))
    
    with plate:
        phi = numpyro.sample("phi", numpyro.distributions.VonMises(mu, kappa))

    with plate:
        numpyro.sample("X", numpyro.distributions.Normal(phi, 1.), obs=X)

def guide(X):
    pass
```
When I run SVI, I get:

`ValueError: Reverse-mode differentiation does not work for lax.while_loop or lax.fori_loop with dynamic start/stop values. Try using lax.scan, or using fori_loop with static start/stop.`

Playing around with the model I noticed that if I fix the kappa (concentration parameter) to constant, and optimize only the mean of the VonMises, it works.
Also following the [docs](https://num.pyro.ai/en/stable/reparam.html#numpyro.infer.reparam.CircularReparam) I added on top of my function:

`@handlers.reparam(config={"phi": CircularReparam()})`
which changes the error message to:

`NotImplementedError: `
I finally tried (based on this) changing my sample to:

```
    with plate:
        with handlers.reparam(config={'phi': CircularReparam()}):
            phi = numpyro.sample("phi", numpyro.distributions.VonMises(mu, 2.0))
```
Which also ends up with the `NotImplementedError.`

Is there a trick, or the VonMises distribution is just not well implemented yet?
Have a good day!

PS. HMC works with for sampling the conc parameter
A

fehiepsi:
A solution is to add a flag into the optimizer to replace [value_and_grad](https://github.com/pyro-ppl/numpyro/blob/01089cfad4078e21e062c02147bcc636331548f1/numpyro/optim.py#L80) by jacfwd like in [hmc implementation](https://github.com/pyro-ppl/numpyro/blob/01089cfad4078e21e062c02147bcc636331548f1/numpyro/infer/hmc_util.py#L242-L246).

PR #1731, opened 2024-02-07
Support forward mode differentiation for SVI

juanitorduz:
Closes https://github.com/pyro-ppl/numpyro/issues/1726

Trying here a "good first issue".

### DIFF
--- a/numpyro/optim.py
+++ b/numpyro/optim.py
@@ -278,7 +278,9 @@ class Minimize(_NumPyroOptim):
         self._method = method
         self._kwargs = kwargs
 
-    def eval_and_update(self, fn: Callable[[Any], tuple], state: _IterOptState):
+    def eval_and_update(
+            self, fn: Callable[[Any], tuple], state: _IterOptState, forward_mode_differentiation=False
+        ):
         i, (flat_params, unravel_fn) = state
 
         def loss_fn(x):
--- a/numpyro/optim.py
+++ b/numpyro/optim.py
@@ -11,12 +11,14 @@ from collections import namedtuple
 from collections.abc import Callable
 from typing import Any, TypeVar
 
-from jax import lax, value_and_grad
+from jax import lax
 from jax.example_libraries import optimizers
 from jax.flatten_util import ravel_pytree
 import jax.numpy as jnp
 from jax.scipy.optimize import minimize
 from jax.tree_util import register_pytree_node, tree_map
+
+from numpyro.infer.hmc_util import _value_and_grad
 
 __all__ = [
     "Adam",
@@ -61,7 +63,9 @@ class _NumPyroOptim(object):
         opt_state = self.update_fn(i, g, opt_state)
         return i + 1, opt_state
 
-    def eval_and_update(self, fn: Callable[[Any], tuple], state: _IterOptState):
+    def eval_and_update(
+            self, fn: Callable[[Any], tuple], state: _IterOptState, forward_mode_differentiation: bool = False
+    ):
         """
         Performs an optimization step for the objective function `fn`.
         For most optimizers, the update is performed based on the gradient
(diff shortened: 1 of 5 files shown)


fehiepsi:
Looks great to me. Could you add a simple test rwith while loop in the model?

juanitorduz:
I added a simple test `test/infer/test_svi.py::test_forward_mode_differentiation` in https://github.com/pyro-ppl/numpyro/pull/1731/commits/532fb5b186d445fda626c3eb682a3d8b666dfb21 but it is failing with

```python
    @functools.wraps(update)
    def tree_update(i, grad_tree, opt_state):
      states_flat, tree, subtrees = opt_state
      grad_flat, tree2 = tree_flatten(grad_tree)
      if tree2 != tree:
        msg = ("optimizer update function was passed a gradient tree that did "
               "not match the parameter tree structure with which it was "
               "initialized: parameter tree {} and grad tree {}.")
>       raise TypeError(msg.format(tree, tree2))
E       TypeError: optimizer update function was passed a gradient tree that did not match the parameter tree structure with which it was initialized: parameter tree PyTreeDef({'loc': *, 'scale': *}) and grad tree PyTreeDef(({'loc': *, 'scale': *}, None))
```

And I am not sure if its because of the implementation of the test is wrong 😑 . Any tips? Thanks

numpyro/optim.py
```diff
@@ -34,6 +34,11 @@
 _OptState = TypeVar("_OptState")
 _IterOptState = tuple[int, _OptState]
 
+def _value_and_grad(f, x, forward_mode_differentiation=False):
+    if forward_mode_differentiation:
+        return f(x), jacfwd(f, has_aux=True)(x)
```
fehiepsi, line 39: I think you can set has_aux to False.

A faster way, which just does 1 forward pass, is to redefine f:
```
def wrapper(x):
  out, aux = f(x)
  return out, (out, aux)

grad, (out, aux) = jacfwd(f, has_aux=True)(x)
return (out, aux), grad
```

We can also apply this trick in hmc forward mode implementation.

juanitorduz: ok! So I added  your suggestion in https://github.com/pyro-ppl/numpyro/pull/1731/commits/ef3a3f76ba53fe984b50f4adfb410c71913d3b4b and now I getting and error downstream 🤔 . Do I need to further unpack somewhere?

```python
        if progress_bar:
            losses = []
            with tqdm.trange(1, num_steps + 1) as t:
                batch = max(num_steps // 20, 1)
                for i in t:
                    svi_state, loss = jit(body_fn)(svi_state, None)
                    losses.append(jax.device_get(loss))
                    if i % batch == 0:
                        if stable_update:
                            valid_losses = [x for x in losses[i - batch :] if x == x]
                            num_valid = len(valid_losses)
                            if num_valid == 0:
                                avg_loss = float("nan")
                            else:
                                avg_loss = sum(valid_losses) / num_valid
                        else:
>                           avg_loss = sum(losses[i - batch :]) / batch
E                           TypeError: unsupported operand type(s) for +: 'int' and 'dict'

numpyro/infer/svi.py:409: TypeError
```

juanitorduz: It worked in https://github.com/pyro-ppl/numpyro/pull/1731/commits/a64c092f2e630b99fc4833927e05292775519251 !

juanitorduz: Same trick in https://github.com/pyro-ppl/numpyro/pull/1731/commits/1259db7b79ac1aecce97fa87a984200906c50e1e for hmc

fehiepsi (approved):
Woohoo, thanks for supporting this feature!

juanitorduz:
Thank you for your guidance @fehiepsi 🙏🙂

PR #1732, opened 2024-02-07
Raise `NotImplementedError` rather than return it.

tillahoffmann:
The forward transform of the `Transform` base class and summary statistics of truncated distributions *return* rather than *raise* exceptions. This PR ensures exceptions are raised.


fehiepsi (approved):
wow what a bug. thanks for the fix!

PR #1738, opened 2024-02-19
fixing nested_sampling.py

renecotyfanboy:
I did a small fix for the nested sampling contributed code so that it works with jaxns 2.4.8
It tested it by running the gaussian shell example, which leads to similar results as what we would have expected before 

Capture d’écran 2024-02-19 à 13 53 40

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -4,7 +4,7 @@ funsor
 ipython
 jax
 jaxlib
-jaxns==2.2.6
+jaxns==2.4.8
 Jinja2
 matplotlib
 multipledispatch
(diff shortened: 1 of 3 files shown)


renecotyfanboy:
Sorry, I didn't check that there was already an open PR for this task #1727. I added a commit for bumping the jaxns version to the latest, and will close my PR if required

fehiepsi:
Thanks, @renecotyfanboy! Could you run make format and make lint to fix the lint issue?

renecotyfanboy:
I manually linted the file with ruff, the `make lint` and `make format` tools fails locally due to the header script : 

```
python scripts/update_headers.py
  File "/Users/sdupourque/PycharmProjects/numpyro/scripts/update_headers.py", line 45, in <module>
    while lines[lineno].startswith(comment.format("Copyright")):
IndexError: list index out of range
```

fehiepsi:
Seems like there is a python file which has conflict format in your folder. CI works btw. :)

PR #1743, opened 2024-02-23
Address google/jax#19885 for numpyro.

tillahoffmann:
This should address performance issues for `LowRankMultivariateNormal` distributions with batch dimensions. Only a single change was required to fix the issue. There are other `[matmul] + [identity]` expressions in the codebase, but, for some reason, they don't cause any issues. The tests verify that no warning is emitted.

### DIFF
--- a/numpyro/distributions/continuous.py
+++ b/numpyro/distributions/continuous.py
@@ -71,7 +71,7 @@ from numpyro.distributions.util import (
     validate_sample,
     vec_to_tril_matrix,
 )
-from numpyro.util import is_prng_key
+from numpyro.util import add_diag, is_prng_key
 
 
 class AsymmetricLaplace(Distribution):
@@ -1081,10 +1081,7 @@ class LKJCholesky(Distribution):
         # correct the diagonal
         # NB: beta_sample = sum(w ** 2) because norm 2 of u is 1.
         diag = jnp.ones(cholesky.shape[:-1]).at[..., 1:].set(jnp.sqrt(1 - beta_sample))
-        cholesky = cholesky + jnp.expand_dims(diag, axis=-1) * jnp.identity(
-            self.dimension
-        )
-        return cholesky
+        return add_diag(cholesky, diag)
 
     def sample(self, key, sample_shape=()):
         assert is_prng_key(key)
@@ -1860,7 +1857,7 @@ def _batch_capacitance_tril(W, D):
     Wt_Dinv = jnp.swapaxes(W, -1, -2) / jnp.expand_dims(D, -2)
     K = jnp.matmul(Wt_Dinv, W)
     # could be inefficient
-    return jnp.linalg.cholesky(jnp.subtract(K, -jnp.identity(K.shape[-1])))
+    return jnp.linalg.cholesky(add_diag(K, 1))
 
 
 def _batch_lowrank_logdet(W, D, capacitance_tril):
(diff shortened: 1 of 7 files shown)


numpyro/distributions/continuous.py
```diff
@@ -1860,7 +1860,7 @@ def _batch_capacitance_tril(W, D):
     Wt_Dinv = jnp.swapaxes(W, -1, -2) / jnp.expand_dims(D, -2)
     K = jnp.matmul(Wt_Dinv, W)
     # could be inefficient
-    return jnp.linalg.cholesky(jnp.add(K, jnp.identity(K.shape[-1])))
+    return jnp.linalg.cholesky(jnp.subtract(K, -jnp.identity(K.shape[-1])))
```
fehiepsi, line 1863: This is interesting. I couldn't come up with a good solution to compute K + eye. Switching to subtract is a bit unfortunate to me.

tillahoffmann: Yeah, it's not great. I'll have a think if `K.at[..., <diag indices>].add(1)` could do the job.

fehiepsi: For some reason, I couldn't reproduce the issue locally. :(

tillahoffmann: It seems to only appear on some systems. For example, I can't reproduce the issue on my MacBook Pro with M1 chip, but it is reproducible in GitHub Actions (see [here](https://github.com/tillahoffmann/google-jax-19885/actions/runs/8013883192/job/21891600980) for an example run).

tillahoffmann: [Here](https://nbviewer.org/gist/tillahoffmann/a30d6cb2a4ff6250773a72ea41912221) is a comparison of different implementations. Using the `at` indexing seems to do a reasonable job. I've just pushed an update.

numpyro/distributions/continuous.py
```diff
@@ -71,7 +71,7 @@
     validate_sample,
     vec_to_tril_matrix,
 )
-from numpyro.util import is_prng_key
+from numpyro.util import add_diag, is_prng_key
```
fehiepsi, line 74: Super cool!!

Could you move this helper to `numpyro.distributions.util` instead?

numpyro/distributions/continuous.py
fehiepsi: You would need to add `.block_until_ready` for a fair comparision: see https://jax.readthedocs.io/en/latest/tutorials/quickstart.html#using-jit-to-speed-up-functions. I guess `add/substract` is more performant than slice update.

tillahoffmann: Yes, good point. Forgot about that. Using the `block_until_ready` call, I get

```
_original_batch_capacitance_tril
677 ms ± 26.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
_sub_batch_capacitance_tril
5.97 ms ± 713 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
_add_batch_capacitance_tril
5.44 ms ± 900 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

on the machine that's raising the warning. On my local M1 chip, I get

```
_original_batch_capacitance_tril
11.5 ms ± 3.52 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
_sub_batch_capacitance_tril
12 ms ± 716 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
_add_batch_capacitance_tril
8.41 ms ± 1.9 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

fehiepsi (approved):

numpyro/distributions/continuous.py
fehiepsi: Thanks, Till!

Issue #1742, opened 2024-02-22
Can't pickle MCMC object (MixedHMC kernel) when `chain_method="parallel"`


msaintja:
Hello, I'm encountering an issue when trying to pickle (whether with `pickle` or `dill` which I've seen being used here) a `numpyro.infer.mcmc.MCMC` object after sampling with MixedHMC. I've included some example code for a model, a bit of a mix of what is in the MixedHMC documentation and the Gaussian Mixture Model tutorial:
```python
import dill
import jax
from jax import random
import jax.numpy as jnp
import numpyro
import numpyro.distributions as dist
from numpyro.infer import HMC, MCMC, MixedHMC

# @title Parameters
PARALLEL_SAMPLING = True # @param {type:"boolean"}
NCHAINS = 2 # @param {type:"integer"}
NSAMPLES = 100 # @param {type:"integer"}

numpyro.set_platform('cpu')
if PARALLEL_SAMPLING:
    numpyro.set_host_device_count(NCHAINS)

key = random.PRNGKey(1)

N = 20
K = 2 
w = jnp.array([0.7, 0.3])
μ = jnp.array([-2, 3])

def model(data):
    # Global variables.
    weights = numpyro.sample("weights", dist.Dirichlet(jnp.ones(K)/K))
    with numpyro.plate("K", K):
        locs = numpyro.sample("locs", dist.Normal(0.0, 10.0))

    with numpyro.plate("data", len(data)):
        # Local variables.
        k = numpyro.sample("k", dist.Categorical(weights))
        numpyro.sample("obs", dist.Normal(locs[k], 1.0), obs=data)

y = random.choice(key, jnp.arange(K), shape=(N,), p=w)
data = random.normal(key, shape=(N,)) + jnp.take(μ, y)

kernel = MixedHMC(HMC(model))
mcmc = MCMC(kernel, num_warmup=NSAMPLES//2, num_samples=NSAMPLES, num_chains=NCHAINS, progress_bar=True, chain_method=("parallel" if PARALLEL_SAMPLING else "sequential"))
mcmc.run(key, data)

with open("chains_test.pkl", 'wb') as f:
    dill.dump(mcmc, f)

```
The error thrown is a `ConcretizationTypeError` (from JAX), and seems to point towards the line with the Dirichlet distribution.

Specifically, the error that I get is the following:
```
/usr/local/lib/python3.10/dist-packages/jax/_src/core.py in __reduce__(self)
[...]

PR #1746, opened 2024-02-24
Fix for pickling an MCMC object with HMCGibbs (and MixedHMC) samplers and parallel chains

msaintja:
This PR addresses issue #1742.

When using `MixedHMC`, and sampling with option `chain_method="parallel"` with more than one chain, using `dill`/`pickle` to save the `MCMC` object resulted in a `ConcretizationTypeError`, as some attributes may contain JAX tracers.
From limited testing, these would be the `_support_sizes_flat` attribute of the MCMC object, as well as some attributes of `_prototype_trace[...]["fn"]`.

Setting those to `None` in the state copy passed to `dill` would allow to circumvent the JAX error.
I'm not sure how critical it is to maintain those attributes in a pickled save, however a simple dump-load-run seems to work fine with the example code presented in #1742:

```python
with open("chains_test.pkl", 'wb') as f:
    dill.dump(mcmc, f)
with open("chains_test.pkl", 'rb') as f:
    mcmc = dill.load(f)
mcmc.post_warmup_state = mcmc.last_state
mcmc.run(key, data)
```

### DIFF
--- a/numpyro/infer/hmc_gibbs.py
+++ b/numpyro/infer/hmc_gibbs.py
@@ -186,6 +186,11 @@ class HMCGibbs(MCMCKernel):
 
         return HMCGibbsState(z, hmc_state, rng_key)
 
+    def __getstate__(self):
+        state = self.__dict__.copy()
+        state["_prototype_trace"] = None
+        return state
+
 
 def _discrete_gibbs_proposal_body_fn(
     z_init_flat, unravel_fn, pe_init, potential_fn, idx, i, val


numpyro/infer/mixed_hmc.py
```diff
@@ -307,4 +307,6 @@ def body_fn(i, vals):
     def __getstate__(self):
         state = self.__dict__.copy()
         state["_wa_update"] = None
+        state["_prototype_trace"] = None
```
fehiepsi, line 310: could you also do this for the super class: DiscreteHMCGibbs?

msaintja: `_prototype_trace` is actually from `HMCGibbs` (parent of `DiscreteHMCGibbs`), so I've made the change there. I'm leaving the override of `__getstate__` in `MixedHMC`, as it requires additional attributes to be set to None compared to `HMCGibbs`.

I've tested that the issue previously occurred with `HMCGibbs`, `DiscreteHMCGibbs` and `HMCECS`, but should be fixed with the latest commit.

fehiepsi (approved):
Thanks for the quick fix and contributing to numpyro, @msaintja!!

Issue #1744, opened 2024-02-23
Adding HMCECS proxy functions


OlaRonning:
Hi,

I'm working on a neural proxy function for HMCECS and have a Taylor expansion proxy with an approximate Hessian. However, the file is becoming somewhat unruly as the proxies are currently in `hmc_gibbs.py`. If I move (only) the proxy functions to a separate file under `contrib` and keep using the static method interface for HMCECS (i.e., `HMCECS.taylor_proxy`), there is no change to the user interface, and I think it would be easier to work with. 

Let me know what you think.

edit: change would look like [this](https://github.com/aleatory-science/numpyro/pull/3) (moved PR to aleatory)

fehiepsi:
yeah, putting it in contrib sounds good to me.

OlaRonning:
Thanks for merging, Du.

PR #1748, opened 2024-02-27
Factored out HMCECS proxies to contrib

OlaRonning:
Moved HMCECS proxies to contrib. 

related to #1744


fehiepsi (approved):
Thanks, Ola!

PR #1751, opened 2024-02-29
Add ZeroSumNormal distribution

kylejcaron:
This ports the ZeroSumNormal distribution (by @aseyboldt and @alexandorra) over from PyMC. Note that it is a special case of a normal distribution that enforces a zerosum constraint over specified dimensions. 

$$
\begin{align*}
    ZSN(\sigma) = N \Big( 0, \sigma^2 (I - \tfrac{1}{n}J) \Big) \\
    \text{where} \ ~ J_{ij} = 1 \ ~ \text{and} \\
    n = \text{nbr of zero-sum axes}
\end{align*}
$$

This was a bit ambitious as a first contribution so I added the WIP tag on this to be safe.


kylejcaron:
Calling out that the logic behind having an event_shape=1 by default helps tests pass (and is similar to the conventions for the multivariatenormal), but makes model fitting very slow. Currently trying to figure this out

kylejcaron:
This seems to be working. Need to test some more edge cases and hopefully simplify the logic in the init function

numpyro/distributions/continuous.py
```diff
@@ -2444,3 +2444,110 @@ def cdf(self, value):
 
     def icdf(self, value):
         return self._ald.icdf(value)
+
+
+class ZeroSumNormal(Distribution):
```
ricardoV94, line 2449: Perhaps add a mention that this is adapted from the PyMC implementation?

kylejcaron: @ricardoV94 do y'all have an official reference on the ZeroSumNormal I can link? Otherwise I figure I could reference the docs? https://www.pymc.io/projects/docs/en/stable/api/distributions/generated/pymc.ZeroSumNormal.html

ricardoV94: A permalink to the github code may be better

kylejcaron: I added both for now. Let me know if there's anything else I can do, and I'll be sure to tag you again when its a little closer to its finalized stage

kylejcaron:
I think this is almost there - a question I still have is if I'm handling the support_shape argument correctly in conjunction with the event_shape and batch_shape.

I was under the impression that the support_shape argument is most similar to an event shape, where if I have support_shape=(20,), and sample from it I end up with an array of shape=(20,), and if I then pass those samples through the log prob function I would get 1 value (shape=() ).

But I think this gets a little weird in conjunction with the `plate` or `.expand` method, which adjusts the batch_shape of the distribution, not the event shape. So if typical usage might be the following

```python
with numpyro.plate("n_categories", 20):
     a_category = numpyro.sample("a_category", dist.ZeroSumNormal(1, n_zerosum_axes=1))
```
and theres a zero sum constraint placed over the 20 different category parameters, I feel like this is more of a batch_shape thing?

numpyro/distributions/continuous.py
AlexAndorra: Nice @kylejcaron !! Great to see ZSN is spreading 🤩 
You can also link to an episode I did with Adrian, where we go through the origins and utility of ZSN: https://learnbayesstats.com/episode/74-optimizing-nuts-developing-zerosumnormal-distribution-adrian-seyboldt/

fehiepsi (commented):
Thanks @kylejcaron! First time I know about this distribution. Could you include some applications of the distribution in the description? A couple of comments:
+ It seems to me that `n_zerosum_axes` is the number of event dimensions of this distribution. I think it is better to use only two parameters `scale` and `event_dim`. You can add the interpretation for them in the description.
```
batch_shape = scale.shape[:-event_dim]
event_shape = scale.shape[-event_dim:]
n_zerosum_axes := event_dim
support_shape := event_shape
```
+ you will need to define a Constraint for this; constraints.real is not the domain of this distribution
+ if you plan to use this distribution as a latent variable, you will need to define a transform for the domain
+ Looking at the PyMC doc, it is not clear to me how we can sample from this distribution. Could you add a better reference or describe it in the description?
+ Do you know how to check for the correctness of log_prob? We have gof_test for other distributions but it seems that you disable it for this distribution.

fehiepsi:
> But I think this gets a little weird in conjunction with the plate or .expand method

The issue happens because you use `constraints.real` as the support of this distribution. You need to define something like [LowerCholesky](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/distributions/constraints.py#L525) constraint.

fehiepsi:
Hmm, the formula in the description looks strange to me. If ZSN(scale) = Normal(...), then this is just a reparameterized version of Normal distribution. It is not clear to me how to enforce zero-sum there.

tillahoffmann:
Great to see a zero-sum distribution being implemented! 

Implementing it using a `ParameterFreeTransform` could do the trick as described in [Ogle and Baker (2020, p. 12)](10.1002/eap.2159). Since each sample of size `n` is constrained such that it sums to zero, posterior sampling is difficult unless the constrained vector is first transformed to an unconstrained vector of size `n - 1` akin to the `StickBreakingTransform` for the `Dirichlet` distribution. This does however come at the cost of having to place a full-rank multivariate normal prior on the unconstrained space.

We're just writing up a manuscript for a sum-to-zero transformation/distribution whose `log_prob` evaluation scales linearly with `n`. Not quite there yet, but hope to be able to share soon.

numpyro/distributions/continuous.py
kylejcaron: absolutely, will do! I found that episode very helpful

kylejcaron:
> Hmm, the formula in the description looks strange to me. If ZSN(scale) = Normal(...), then this is just a reparameterized version of Normal distribution. It is not clear to me how to enforce zero-sum there.

You're right, thats a likelihood of a normal distribution thats scaled to accommodate that zerosum axes are transformed to be shape n-1. I realized this is missing the zero sum transform like @tillahoffmann is mentioning, where the original input is transformed to a vector of size n-1 along each zerosum axis - I'm working on implementing that now.  Here's the equation from the [Olga and Barber paper (eq. 9, page 12)](https://esajournals.onlinelibrary.wiley.com/doi/abs/10.1002/eap.2159)

$$
\begin{gather}
     \epsilon_{-J} \sim Normal_{J-1}(0,\Sigma) \\
     \Sigma_{j,k} = - \frac{\sigma^2}{J} \,\ j \neq k \ \text{and} \ \Sigma_{j,j} = 
     \sigma^2_{\epsilon} \\
     \epsilon_{J} = - \Sigma_{j=1}^{J-1} \epsilon_j
\end{gather}
$$



Question for both of you, are there any resources that discuss numpyro transforms a little further? I'm having trouble understanding how I would implement this - creating a ZeroSumTransform class seems simple, but how to call that is where I'm struggling. Most of the transforms I'm seeing right now take a base distribution and apply a transform to the base distribution - I cant tell if thats necessarily the right approach here. 

An example of what I'm referring to is the lognormal distribution, which is a transform on top of a normal distribution

```python
class LogNormal(TransformedDistribution):
    arg_constraints = {"loc": constraints.real, "scale": constraints.positive}
    support = constraints.positive
    reparametrized_params = ["loc", "scale"]

    def __init__(self, loc=0.0, scale=1.0, *, validate_args=None):
        base_dist = Normal(loc, scale)
        self.loc, self.scale = base_dist.loc, base_dist.scale
        super(LogNormal, self).__init__(
            base_dist, ExpTransform(), validate_args=validate_args
        )
```

I think this pattern wouldn't work well with the current sampling method, and potentially with the log prob as well?

kylejcaron:
> * It seems to me that `n_zerosum_axes` is the number of event dimensions of this distribution. I think it is better to use only two parameters `scale` and `event_dim`. You can add the interpretation for them in the description.

Yes I can add some applications in the description! In the meantime, the typical use cases are to use this as an alternative for dummy-encoding, where there can be a parameter for every single category of interest while still having an intercept parameter without running into identifiability issues

Technically I think n_zerosum_axes can be different from the event_dim. For example, lets say we have a 2d categorical variable of shape=(2, 50), we may only want to enforce a zerosum constraint over the last dim, i.e.

```python
with numpyro.plate("n_conditions", 2):
     with numpyro.plate("n_categories", 20, dim=-1):
          a_category = numpyro.sample("a_category", dist.ZeroSumNormal(1, n_zerosum_axes=1))
```

fehiepsi:
I think you can use the LKJCholesky as an example. It has a non-trivial support which requires sort of [lower cholesky transform](https://github.com/pyro-ppl/numpyro/blob/7769a32df065fcd3bdf8b59964ced9578379e021/numpyro/distributions/transforms.py#L742). Re event dim: this denotes correlated dimensions, which is the last dimension of your ZeroSum example. Plate denote independent dimension. So you need to remove the category plate if you think that those categories are correlated (via zero sum enforcement).

kylejcaron:
> I think you can use the LKJCholesky as an example. It has a non-trivial support which requires sort of [lower cholesky transform](https://github.com/pyro-ppl/numpyro/blob/7769a32df065fcd3bdf8b59964ced9578379e021/numpyro/distributions/transforms.py#L742). Re event dim: this denotes correlated dimensions, which is the last dimension of your ZeroSum example. Plate denote independent dimension. So you need to remove the category plate if you think that those categories are correlated (via zero sum enforcement).

To make sure I have this right - lets say we're modeling the effects of the 50 states in the USA and some continuous regressor X. The traditional way to do this simply may be to have a separate intercept parameter for every state and no global intercept

__example 1:__

```python
def model_example1(X, state, y=None):
    with numpyro.plate("states", 50):
        alpha_state = numpyro.sample("alpha_state", dist.Normal(0, 1))

    beta = numpyro.sample("beta", dist.Normal( 0, 1) )
    sigma = numpyro.sample("sigma", dist.Exponential(1) )

    with numpyro.plate("data", len(X)):
        numpyro.sample("obs", dist.Normal(alpha_state[state] + beta*X, sigma), obs=y)
```

If we want to have a global intercept parameter for interpretability, the ZeroSumNormal could come in as follows 

__example 2a:__

```python
def model_example2a(X, state, y=None):
    with numpyro.plate("states", 50):
        alpha_state = numpyro.sample("alpha_state", dist.ZeroSumNormal(scale=1, n_zerosum_axes=1))
   
    global_intercept = numpyro.sample("global_intercept", dist.Normal( 0, 1) )
    beta = numpyro.sample("beta", dist.Normal( 0, 1) )
    sigma = numpyro.sample("sigma", dist.Exponential(1) )

    with numpyro.plate("data", len(X)):
        numpyro.sample("obs", dist.Normal(global_intercept + alpha_state[state] + beta*X, sigma), obs=y)
```

I didnt quite think about how the zerosum constraint creates correlated dimensions. Would that mean the actual better pattern to use instead of example 2a) is the following?

__example 2b:__

```python
def model_example2b(X, state, y=None):
   
    alpha_state = numpyro.sample("alpha_state", dist.ZeroSumNormal(scale=1, n_zerosum_axes=1, support_shape=(50,))
   
    global_intercept = numpyro.sample("global_intercept", dist.Normal( 0, 1) )
    beta = numpyro.sample("beta", dist.Normal( 0, 1) )
    sigma = numpyro.sample("sigma", dist.Exponential(1) )

    with numpyro.plate("data", len(X)):
        numpyro.sample("obs", dist.Normal(global_intercept + alpha_state[state] + beta*X, sigma), obs=y)
```

Or are you saying that we can keep the structure of example 2a (and just rename `n_zerosum_axes` to `event_dim`) and then have internal logic that separates out the `event_shape` so that the dims (the 50 states in this case) arent independent of eachother

aseyboldt:
Nice to see this implemented in numpyro as well :-)

> Implementing it using a ParameterFreeTransform could do the trick as described in [Ogle and Baker (2020, p. 12)](https://github.com/pyro-ppl/numpyro/pull/10.1002/eap.2159). Since each sample of size n is constrained such that it sums to zero, posterior sampling is difficult unless the constrained vector is first transformed to an unconstrained vector of size n - 1 akin to the StickBreakingTransform for the Dirichlet distribution. This does however come at the cost of having to place a full-rank multivariate normal prior on the unconstrained space.

> We're just writing up a manuscript for a sum-to-zero transformation/distribution whose log_prob evaluation scales linearly with n. Not quite there yet, but hope to be able to share soon.

I didn't follow the discussion fully, but the current implementation in PyMC already scales linearly in n. We use the fact that the subspace defined by the zero-sum property is a linear subspace, so we can just map the vector space that is one dim smaller using a householder transformation.

fehiepsi:
> example 2b:

yes, that's my interpretation from the formula. The sum of `alpha_state` is zero hence they are correlated.

> subspace defined by the zero-sum property is a linear subspace

yeah, I think we can just follow the way samples are generated: pad the unconstrained vector by zero across all correlated dimensions, then subtract the mean, etc. Probably householder transform performs a better job. Could you provide some pointer, @aseyboldt ?

aseyboldt:
You can find the code of the transformation here: https://github.com/pymc-devs/pymc/blob/main/pymc/distributions/transforms.py#L266

The covariance of the standard ZSN has one zero eigenvalue with eigenvector $w = (1, 1, \dots)^T$, while all other eigenvalues are equal to one. We would like to find an orthogonal map that maps the span of $w$ to the span of $e_1 = (1, 0, 0, \dots)^T$. We can write Householder transformations as $H_v = I - 2vv^T$, where $||v|| = 1$, so we would like to find a $v$ such that $H_v w = \alpha e_1$.  For instance $v = \frac{w + ||w|| e_1}{||w + ||w|| e_1||}$ does the job (see for instance https://de.wikipedia.org/wiki/Householdertransformation#Konstruktion_einer_spezifischen_Spiegelung The English wikipedia article doesn't contain that, but I'm sure there are tons of references out there somewhere...). $H_v$ is its own inverse, so we can map between the transformed and untransformed spaces by just applying $H_v$, and we can do that without computing the whole matrix $H_v x = (I - 2vv^T)x = x - 2v(v^Tx)$. And since $H_v$ is orthogonal, the jacobian det is one.

In this transformed space the ZSN is just a diagonal normal with the first variance zero, so we can get rid of the first entry of that array entirely, and get a map to $n-1$ dimensional space.

If your design for distributions allows you to define the density on the transformed space directly, that makes things really easy here, since on the transformed space we have an iid normal distribution. If not (as is unfortunately the case in pymc) you have to trick a bit to still write down a density in the untransformed space that works out (See https://github.com/pymc-devs/pymc/blob/main/pymc/distributions/multivariate.py#L2841).

Hope that helps, if not feel free to ask.

To elaborate on the use case of the ZSN a bit, that would I think be mostly those two:

- In regressions with catogorical outcomes we can use ZSN dists instead of using one outcome as the reference. In a frequentist setting the choice of reference doesn't matter, but if we have priors it often does. But if no outcome is special in a way where it would make sense to use it as a reference we can use the ZSN to avoid that arbitrary choice.
- Hierarchical models with categorical predictors are often over parameterized if we don't take some care. Say for instance we have a model like this (using pymc code, by I hope you still get the idea):

```
with pm.Model():
    # Fixed for simplicity, nothing really changes if we estimate standard deviations
    intercept_sd = 5
    group_sd = 2

    intercept = pm.Normal("intercept", sigma=intercept_sd)
    group_effect = pm.Normal("group_effect", sigma=group_sd, dims="group")

    mu = intercept + group_effect[group_idx]

    sigma = 1
    pm.Normal("data", mu=mu, sigma=sigma, observed=data)
```

Here `mu` has `n_groups` degrees of freedom, but we have `n_groups + 1` degrees of freedom in the model. So if the data size increases (while keeping the same number of groups) there must be a 1-dimensional subspace where the posterior variance stays constant, while all other posterior variances go to zero. But this means that the condition number of the posterior variance goes to infinity. We don't want that...

We can fix this by splitting the `group_effect` into two parts. For any $x ~ N(0, \sigma^2 I)$ we can always write it as $$x_{\text{deflect} ~ ZSN(\sigma^2)\\ x_\text{mean} ~ N(0, \frac{\sigma^2}{\text{len}(x)})$$, so

```
with pm.Model():
    # Fixed for simplicity, nothing really changes if we estimate standard deviations
    intercept_sd = 5
    group_sd = 2

    intercept = pm.Normal("intercept", sigma=intercept_sd)
    group_mean = pm.Normal("group_mean", sigma=pt.sqrt(group_sd**2 / n_groups))
    group_deflect = pm.ZeroSumNormal("group_deflect", sigma=group_sd, dims="group")
    group_effect = group_mean + group_deflect

    mu = intercept + group_mean + group_deflect[group_idx]

    sigma = 1
    pm.Normal("data", mu=mu, sigma=sigma, observed=data)
```

But now we can merge `intercept + group_mean` into one normally distributed random variable, and that way get rid of the extra degree of freedom:

```
with pm.Model():
    # Fixed for simplicity, nothing really changes if we estimate standard deviations
    intercept_sd = 5
    group_sd = 2

    intercept_plus_group_mean = pm.Normal("intercept_plus_group_mean", sigma=pt.sqrt(intercept_sd ** 2 + group_sd**2 / n_groups))
    group_deflect = pm.ZeroSumNormal("group_deflect", sigma=group_sd, dims="group")
    group_effect = group_mean + group_deflect

    mu = intercept_plus_group_mean + group_deflect[group_idx]

    sigma = 1
    pm.Normal("data", mu=mu, sigma=sigma, observed=data)
```

This also works for more complicated hierarchical models, in some cases you need several zero-sum axes then.

If necessary, draws from the original intercept and group_mean can then be reconstructed after sampling.
A nice bonus of this is that I noticed in many applications I really want to know `group_deflect`, and not `group_effect`, especially if there really is only a fixed number of groups, and it doesn't really make sense to think about the infinite population of groups. So for instance if each group is a US state, do we really want to figure out how California differs from the mean of a (completely imaginary) infinite population of US states, or do we want to know how it differs from the mean of the actual US states?

kylejcaron:
@aseyboldt thank you for the great response thats really helpful. 

> you have to trick a bit to still write down a density in the untransformed space that works out

I was under the impression that im pymc the ZeroSumTransform is applied and then the logprob is contributed, is that not actually the case? Where exactly do the forward and backward transforms get applied? (I'm admittedly having trouble following the `_default_transform` logic)

kylejcaron:
@fehiepsi I added an potential `ZeroSumTransform` class but I'm struggling to figure out if its feasible. The idea is to use a pattern like the one below but Im struggling to figure out if the shape handling can work.

```python
class ZeroSumNormal(TransformedDistribution):
    arg_constraints = {"scale": constraints.positive}
    support = constraints.real # FIXME
    reparametrized_params = ["scale"]
    pytree_aux_fields = ("n_zerosum_axes","support_shape",)

    def __init__(self, scale=1.0, n_zerosum_axes=None, support_shape=None, *, validate_args=None):
        if not all(tuple(i == 1 for i in jnp.shape( scale ))):
            raise ValueError("scale must have length one across the zero-sum axes")

        if jnp.ndim(scale) == 0:
            (scale,) = promote_shapes(scale, shape=(1,))

        self.n_zerosum_axes = self.check_zerosum_axes(n_zerosum_axes)
        sample_shape = scale.shape[:1]
        batch_shape = scale.shape[1:]
        event_shape = support_shape 
        
        self.scale = jnp.broadcast_to(scale, batch_shape + event_shape)
        base_dist = dist.Normal(0, self.scale) 

        super(ZeroSumNormal, self).__init__(
            base_dist,
            # inverse the transform so that when calculating the logprob, a degree of freedom is dropped first
            ZeroSumTransform(jnp.arange(-self.n_zerosum_axes,0)).inv,
            validate_args=validate_args
        )
```

aseyboldt:
@kylejcaron 

> I was under the impression that in pymc the ZeroSumTransform is applied and then the logprob is contributed, is that not actually the case? Where exactly do the forward and backward transforms get applied? (I'm admittedly having trouble following the _default_transform logic)

Yes, that's how it works. In the computation graph of the logp we start with a point in the transformed space. We map that to the untransformed space, compute the logp there (that's what's in the logp function I linked to earlier), and add the log_jac_det of the transformation (which is zero in this case, as the transformation is orthogonal).

What I meant was that is the case of the zerosumnormal it would actually be a little bit more convenient to take the point in the transformed space, compute the logp there and then just map the point to the untransformed space for further use in the model, because in the transformed space the logp is very simple.

But it isn't that much more difficult in the untransformed space, so no big deal either way. Wikipedia for instance has a short section about densities of multivariate normal dists with singular covariances. (https://en.wikipedia.org/wiki/Multivariate_normal_distribution#Degenerate_case).
We can use that to derive a simple expression for the logp on the untransformed space as well, one that also doesn't require building a big matrix (Or equivalently we could just do the transformation algebraically to figure out an expression, this turns out to be the same thing):
In that expression we have the pseudo-determinant and the pseudoinverse. The pseudo-determinant is just $\sigma^2 (n-1)$, because we have one eigenvalue that's zero. Since we know the eigenvalue decomposition of the covariance (one eigenvalue zero, all other $\sigma^2$), we also know the eigenvalue decomposition of the pseudoinverse: one eigenvalue zero, all other $\sigma^{-2}$, which we can write as $\sigma^{-2}(I - ww^T)$, where $w$ is the normalized vector of all ones. So if `x` is a point in the zero-sum space (ie $w^Tx = 0$), we have $x^T\Sigma^+ x = \sigma^{-2} x^Tx$.

If there are several zero-sum axes you have to take some extra care, as then several eigenvalues will be zero, so you have to keep track of how many of those there are. (which happens [here](https://github.com/pymc-devs/pymc/blob/main/pymc/distributions/multivariate.py#L2832) in the code).

tillahoffmann:
Thank you for the additional context regarding the Householder transformation, @aseyboldt. Very cool!

> If your design for distributions allows you to define the density on the transformed space directly, that makes things really easy here, since on the transformed space we have an iid normal distribution.

If I understand correctly, the `ZeroSumNormal` distribution could be implemented as

```python
class ZeroSumNormal(TransformedDistribution):
    def __init__(self, scale, event_shape):
        event_ndim = len(event_shape)
        transformed_shape = tuple(size - 1 for size in event_shape)
        zero_sum_axes = tuple(-(i + 1) for i in range(event_ndim))
        super().__init__(
            Normal(0, scale).expand(transformed_shape).to_event(event_ndim),
            ZeroSumTransform(zero_sum_axes),
        )
```

It declares a diagonal normal with shape that is one smaller than the target event shape. Then transforms the noise to the zero-sum space along the last `len(event_shape)` dimensions.

aseyboldt:
That does look correct to me.

fehiepsi:
Thanks so much for your detailed explanations, @aseyboldt! They're very helpful for me.

@kylejcaron @tillahoffmann Yes, I think we can implement ZeroSumNormal as a TransformedDistribution with ZeroSumTransform. The support would be `constraints.zero_sum(event_dim)` which asserts that the sum over each event dimension is zero. I think providing `scale` and `event_shape` makes sense. The batch shape will be `jnp.shape(scale)[:-len(event_shape)]`. The most important class would be ZeroSumTransform(event_dim). Its domain is `constraints.independent(constraints.real, event_dim)` and the codomain is `constraints.zero_sum(event_dim)`.

kylejcaron:
> Thanks so much for your detailed explanations, @aseyboldt! They're very helpful for me.
> 
> @kylejcaron @tillahoffmann Yes, I think we can implement ZeroSumNormal as a TransformedDistribution with ZeroSumTransform. The support would be `constraints.zero_sum(event_dim)` which asserts that the sum over each event dimension is zero. I think providing `scale` and `event_shape` makes sense. The batch shape will be `jnp.shape(scale)[:-len(event_shape)]`. The most important class would be ZeroSumTransform(event_dim). Its domain is `constraints.independent(constraints.real, event_dim)` and the codomain is `constraints.zero_sum(event_dim)`.

I went ahead and made some changes based on all of your advice - when I attempt to fit a model, the guide and model shape end up not lining up (guide shape = 20, model_shape = 19), causing a failure [here](https://github.com/pyro-ppl/numpyro/blob/84973a98aafd0ebd9ae936a1b4173587c5cdec1b/numpyro/infer/util.py#L78)

Am I missing a transform somewhere or is a hack for this needed so that they dont need to broadcast for this distribution?

tillahoffmann:
> I went ahead and made some changes based on all of your advice - when I attempt to fit a model, the guide and model shape end up not lining up (guide shape = 20, model_shape = 19), causing a failure [here](https://github.com/pyro-ppl/numpyro/blob/84973a98aafd0ebd9ae936a1b4173587c5cdec1b/numpyro/infer/util.py#L78)

Sounds great! I think you may have to implement `forward_shape` and `inverse_shape` on the transform so the autoguide can determine how many variational parameters to use.

test/test_distributions.py
```diff
@@ -1454,6 +1467,9 @@ def test_gof(jax_dist, sp_dist, params):
     if jax_dist is dist.ProjectedNormal:
         dim = samples.shape[-1] - 1
 
+    if jax_dist is dist.ZeroSumNormal:
+        pytest.skip("skip gof test for ZeroSumNormal")
```
kylejcaron, line 1471: Any thoughts for an alternative here to avoid skipping the test? Currently, the `gof` result is 0, so it fails and I added this as a temporary workaround

numpyro/distributions/constraints.py
```diff
@@ -697,6 +698,31 @@ def feasible_like(self, prototype):
...
 
 
+class _ZeroSum(Constraint):
+    def __init__(self, event_dim=1):
+        self.event_dim = event_dim
+        super().__init__()
+
+    def __call__(self, x):
+        jnp = np if isinstance(x, (np.ndarray, np.generic)) else jax.numpy
+        zerosum_true = []
```
tillahoffmann, line 708: Could do something like `zerosum_true = True` and then in the loop `zerosum_true = zerosum_true & jnp.allclose(x.sum(dim), 0, ...)`. That means we don't have to keep track of the list.

numpyro/distributions/continuous.py
```diff
@@ -2438,3 +2439,72 @@ def cdf(self, value):
...
+
+
+class ZeroSumNormal(TransformedDistribution):
+    r"""
+    Zero Sum Normal distribution adapted from PyMC [1] as described in [2,3]. This is a Normal distribution where one or
+    more axes are constrained to sum to zero (the last axis by default).
+
+    Sample code for using ZeroSumNormal in the context of a single axis to zero-contrain::
+
+        def model(category_ind, y): # category_ind is an indexed categorical variable with 20 categories
```
tillahoffmann, line 2451: This might need `>>>` to be picked up by the doctests. Do you want to call the model so it's execution is validated by doctests?

numpyro/distributions/transforms.py
```diff
@@ -1380,6 +1382,70 @@ def __eq__(self, other):
...
+class ZeroSumTransform(Transform):
+    """A transform that constrains an array to sum to zero, adapted from PyMC [1] as described in [2,3]
+
+    **References**
+    [1] https://github.com/pymc-devs/pymc/blob/244fb97b01ad0f3dadf5c3837b65839e2a59a0e8/pymc/distributions/transforms.py#L266
+    [2] https://www.pymc.io/projects/docs/en/stable/api/distributions/generated/pymc.ZeroSumNormal.html
+    [3] https://learnbayesstats.com/episode/74-optimizing-nuts-developing-zerosumnormal-distribution-adrian-seyboldt/
+    """
+
+    def __init__(self, zerosum_axes):
```
tillahoffmann, line 1394: It might be useful to just have `transform_ndims` as in #1762 because the event dimensions are the last dimensions. Say we have an input tensor with shape `(3, 4, 5, 6)` and `zerosum_axes = (-1, -3)`. Then the event dimension implied by `codomain` is two, but in practice the last three axes are dependent.

tillahoffmann (commented):
Nice! Could also add the transform to the test here:

https://github.com/pyro-ppl/numpyro/blob/84973a98aafd0ebd9ae936a1b4173587c5cdec1b/test/test_transforms.py#L301

numpyro/distributions/continuous.py
kylejcaron: I wasnt planning on it but this looks like a good idea to me I can adjust this

numpyro/distributions/transforms.py
kylejcaron: To make sure I'm interpreting correctly: would the idea be to replace something like `ZeroSumTransform(zerosum_axes=(-2, -1))` with `ZeroSumTransform(transform_ndims=2)` as a way to avoid edge cases since we know the event_shape will always be at the end?

tillahoffmann: Yes, that would align with these two:

https://github.com/pyro-ppl/numpyro/blob/84973a98aafd0ebd9ae936a1b4173587c5cdec1b/numpyro/distributions/transforms.py#L569

https://github.com/pyro-ppl/numpyro/blob/84973a98aafd0ebd9ae936a1b4173587c5cdec1b/numpyro/distributions/transforms.py#L1226-L1230

But I'll defer to @fehiepsi.

kylejcaron: This also solves that `test_parametrized_transform_eq` was failing (I think due to non-jittable parameters when I was supplying a tuple)

numpyro/distributions/continuous.py
kylejcaron: added!

numpyro/distributions/continuous.py
```diff
@@ -2438,3 +2439,99 @@ def cdf(self, value):
...
+            validate_args=validate_args,
+        )
+
+    @constraints.dependent_property(is_discrete=False)
+    def support(self):
+        return constraints.zero_sum(len(self.event_shape))
+
+    @property
+    def mean(self):
+        return jnp.broadcast_to(0, self.batch_shape)
```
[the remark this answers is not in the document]
kylejcaron: Still thinking about this - lets say `batch_shape = (100,)` and `event_shape=(10,)`. The idea of the ZeroSumNormal is that the mean across the event_shape is 0. If we including the event shape then the means of each of those 10 individual parameters in the event shape wont necessarily need to be zero - but again, the mean across them will be zero

I'm thinking this means the mean function should stay as it is above? Is there a  numpyro specific reason that the event_shape should be included? Or am I thinking about this wrong?

docs/source/distributions.rst
```diff
@@ -380,6 +380,13 @@ Weibull
     :show-inheritance:
     :member-order: bysource
 
+ZeroSumNormal
+^^^^^^^
```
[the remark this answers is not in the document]
kylejcaron: good call thank you!

docs/source/distributions.rst
```diff
@@ -1014,6 +1021,15 @@ StickBreakingTransform
     :show-inheritance:
     :member-order: bysource
 
+ZeroSumTransform
+^^^^^^^^^^^^^^^^^^^^^^^^
```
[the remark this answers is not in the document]
kylejcaron: good call thank you!

numpyro/distributions/continuous.py
```diff
@@ -2438,3 +2439,99 @@ def cdf(self, value):
...
+
+    def __init__(self, scale, event_shape, *, validate_args=None):
+        event_ndim = len(event_shape)
+        if jnp.ndim(scale) == 0:
+            (scale,) = promote_shapes(scale, shape=(1,))
+        transformed_shape = tuple(size - 1 for size in event_shape)
+        self.scale = scale
+        super().__init__(
+            Normal(0, scale).expand(transformed_shape).to_event(event_ndim),
+            ZeroSumTransform(event_ndim).inv,
```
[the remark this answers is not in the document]
kylejcaron: that makes sense, makes alot of the references more simple as well - changed!

numpyro/distributions/continuous.py
```diff
@@ -2438,3 +2439,99 @@ def cdf(self, value):
 
     def icdf(self, value):
         return self._ald.icdf(value)
+
+
+class ZeroSumNormal(TransformedDistribution):
+    r"""
+    Zero Sum Normal distribution adapted from PyMC [1] as described in [2,3]. This is a Normal distribution where one or
```
kylejcaron, line 2446: @AlexAndorra @aseyboldt @ricardoV94 hey folks, this PR is nearing ready to go. Thank you for all of your help and please let me know if there's more I can add to properly credit all of you and pymc. I've found the ZeroSumNormal to be such a useful tool, cant thank you all enough!

numpyro/distributions/transforms.py
```diff
@@ -1380,6 +1382,92 @@ def __eq__(self, other):
         return jnp.array_equal(self.transition_matrix, other.transition_matrix)
 
 
+class ZeroSumTransform(Transform):
+    """A transform that constrains an array to sum to zero, adapted from PyMC [1] as described in [2,3]
```
kylejcaron, line 1386: @AlexAndorra @aseyboldt @ricardoV94 same as I said above, this PR is nearing ready to go - let me know if there's more I can add to properly credit all of you and pym

numpyro/distributions/continuous.py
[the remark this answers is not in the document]
fehiepsi: in numpyro, `mean` is the mean value across sample_shape of all samples (which have shape sample_shape+batch_shape+event_shape) drawn from this distribution.

numpyro/distributions/continuous.py
```diff
@@ -2438,3 +2439,99 @@ def cdf(self, value):
...
+
+    @property
+    def variance(self):
+        event_ndim = len(self.event_shape)
+        zero_sum_axes = tuple(range(-event_ndim, 0))
+        theoretical_var = self.scale.astype(float) ** 2
+        for axis in zero_sum_axes:
+            theoretical_var *= 1 - 1 / self.event_shape[axis]
+
+        return theoretical_var
```
fehiepsi, line 2537: Similar to the mean, we need to broadcast this to batch_shape + event_shape

kylejcaron: got it, updated!

numpyro/distributions/continuous.py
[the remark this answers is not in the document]
kylejcaron: I see, thank you for the explanation! just updated

numpyro/distributions/transforms.py
kylejcaron: calling this resolved,  switched to `transform_ndims` and tests are passing

test/test_distributions.py
kylejcaron: sounds good, moved it up earlier

numpyro/distributions/continuous.py
aseyboldt: Looks good to me. Glad to hear you found it useful :-)

numpyro/distributions/continuous.py
```diff
@@ -2438,3 +2439,99 @@ def cdf(self, value):
...
+            validate_args=validate_args,
+        )
+
+    @constraints.dependent_property(is_discrete=False)
+    def support(self):
+        return constraints.zero_sum(len(self.event_shape))
+
+    @property
+    def mean(self):
+        return jnp.broadcast_to(0, self.batch_shape + self.event_shape)
```
fehiepsi, line 2527: nit: it's better to use `jnp.zeros(self.batch_shape + self.event_shape)`

numpyro/distributions/continuous.py
```diff
@@ -2438,3 +2439,99 @@ def cdf(self, value):
...
+
+    @property
+    def variance(self):
+        event_ndim = len(self.event_shape)
+        zero_sum_axes = tuple(range(-event_ndim, 0))
+        theoretical_var = self.scale.astype(float) ** 2
```
fehiepsi, line 2533: in case scale is a python float, we can't access its `astype` method. I guess it is better to use `jnp.square(self.scale)` here

fehiepsi (commented):
LGTM pending feedback from @tillahoffmann! Thanks for the awesome work, @kylejcaron!

numpyro/distributions/continuous.py
kylejcaron: good call, fixed!

numpyro/distributions/continuous.py
kylejcaron: ah I didnt consider that, that makes sense thank you - its fixed!

numpyro/distributions/constraints.py
```diff
@@ -697,6 +698,30 @@ def feasible_like(self, prototype):
...
+class _ZeroSum(Constraint):
+    def __init__(self, event_dim=1):
+        self.event_dim = event_dim
+        super().__init__()
+
+    def __call__(self, x):
+        jnp = np if isinstance(x, (np.ndarray, np.generic)) else jax.numpy
+        zerosum_true = True
+        for dim in range(-self.event_dim, 0):
+            zerosum_true = zerosum_true & jnp.allclose(
```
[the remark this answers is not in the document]
kylejcaron: Good callout, the tolerance seems like it can be much more strict than it was. I actually decided to follow the first example and use `tol = jnp.finfo(x.dtype).eps * x.shape[-1] * 10` as the `atol` and removed the rtol - do you think thats sufficient?

numpyro/distributions/continuous.py
```diff
@@ -2438,3 +2439,99 @@ def cdf(self, value):
...
+    [2] https://www.pymc.io/projects/docs/en/stable/api/distributions/generated/pymc.ZeroSumNormal.html
+    [3] https://learnbayesstats.com/episode/74-optimizing-nuts-developing-zerosumnormal-distribution-adrian-seyboldt/
+    """
+
+    arg_constraints = {"scale": constraints.positive}
+    reparametrized_params = ["scale"]
+
+    def __init__(self, scale, event_shape, *, validate_args=None):
+        event_ndim = len(event_shape)
+        if jnp.ndim(scale) == 0:
```
tillahoffmann, line 2511: Why do we need to promote the scalar here?

numpyro/distributions/transforms.py
```diff
@@ -1530,3 +1618,8 @@ def _transform_to_softplus_lower_cholesky(constraint):
 @biject_to.register(constraints.simplex)
 def _transform_to_simplex(constraint):
     return StickBreakingTransform()
+
+
+@biject_to.register(constraints.zero_sum)
+def _transform_to_zero_sum(constraint):
+    return ZeroSumTransform(constraint.event_dim).inv
```
tillahoffmann, line 1625: I think this shouldn't have the `.inv` here.

```python
>>> from numpyro.distributions.transforms import biject_to
>>> from numpyro.distributions.constraints import zero_sum
>>> from jax import random
>>> 
>>> transform = biject_to(zero_sum(1))
>>> transform(random.normal(random.key(7), [3]))
Array([-0.7128072 , -0.14349887], dtype=float32)

```

tillahoffmann (commented):
Looks great! A few more little comments.

numpyro/distributions/transforms.py
fehiepsi: good catch!

numpyro/distributions/continuous.py
kylejcaron: You're right we dont here good callout

numpyro/distributions/constraints.py
```diff
@@ -697,6 +698,29 @@ def feasible_like(self, prototype):
...
+        zerosum_true = True
+        for dim in range(-self.event_dim, 0):
+            zerosum_true = zerosum_true & jnp.allclose(x.sum(dim), 0, atol=tol)
+        return zerosum_true
+
+    def __eq__(self, other):
+        return type(self) is type(other) and self.event_dim == other.event_dim
+
+    def feasible_like(self, prototype):
+        return jax.numpy.broadcast_to(0, prototype.shape)
```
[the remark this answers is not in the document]
kylejcaron: yes good idea, changed!

numpyro/distributions/transforms.py
kylejcaron: youre right it definitely shouldnt be there - I just fixed this. question, should one of the tests have been catching that? I didn't see anything pop up with the test cases I added

numpyro/distributions/constraints.py
```diff
@@ -697,6 +698,29 @@ def feasible_like(self, prototype):
...
+        zerosum_true = True
+        for dim in range(-self.event_dim, 0):
+            zerosum_true = zerosum_true & jnp.allclose(x.sum(dim), 0, atol=tol)
+        return zerosum_true
+
+    def __eq__(self, other):
+        return type(self) is type(other) and self.event_dim == other.event_dim
+
+    def feasible_like(self, prototype):
+        return jax.numpy.zeros_like(0, prototype.shape)
```
tillahoffmann, line 718:
```suggestion
        return jax.numpy.zeros_like(prototype)
```

numpyro/distributions/continuous.py
```diff
@@ -2438,3 +2439,97 @@ def cdf(self, value):
...
+    [3] https://learnbayesstats.com/episode/74-optimizing-nuts-developing-zerosumnormal-distribution-adrian-seyboldt/
+    """
+
+    arg_constraints = {"scale": constraints.positive}
+    reparametrized_params = ["scale"]
+
+    def __init__(self, scale, event_shape, *, validate_args=None):
+        event_ndim = len(event_shape)
+        transformed_shape = tuple(size - 1 for size in event_shape)
+        self.scale = scale
```
tillahoffmann, line 2512: It may be useful to implement `self.scale` as a property that gets the scale from the base distribution because the scale attribute doesn't make the round-trip for jitting. Something like this.

```python
@property
def scale(self):
    return self.base_dist.scale
```

Here's what I mean.

```python
>>> import jax
>>> from numpyro import distributions
>>>
>>> class MyDist(distributions.TransformedDistribution):
...     def __init__(self, scale):
...         self.scale = scale
...         super().__init__(
...             distributions.Normal(0, self.scale), 
...             distributions.transforms.ExpTransform(),
...         )
>>>
>>> dist = MyDist(3)
>>> other = jax.jit(lambda x: x)(dist)
>>> dist.scale
3
>>> other.scale
AttributeError: 'MyDist' object has no attribute 'scale'
```

numpyro/distributions/transforms.py
tillahoffmann: It might be worth adding another test that checks that `biject_to` actually bijects to the desired constraint. Maybe for another PR?

tillahoffmann (commented):
Few more little ones, sorry this is such a piecemeal review.

numpyro/distributions/transforms.py
kylejcaron: I think thats a good call, I'll call this resolved for now then. Thank you!

kylejcaron:
> Few more little ones, sorry this is such a piecemeal review.

dont worry about it I really appreciate all of the help, I still have a lot to learn - thank you! unless you can find any other issues, that discussion on the scale param might be the last unresolved piece

tillahoffmann (approved):
This looks great! Thank you, @kylejcaron!

kylejcaron:
> Also thanks for a detailed review, @tillahoffmann!

Thank you everyone for all of the help its been much appreciated! @fehiepsi I had to re-request review, the doctest for distributions/continuous.py was failing and didnt show up when I ran `make doctest` (maybe because the nested_sampling module was failing first?). 

I was able to test it manually with `python -m doctest -v numpyro/distributions/continuous.py` and can confirm its fixed

PR #1754, opened 2024-03-05
Add explicit reparametrizer.

tillahoffmann:
This PR adds an `ExplicitReparam` of a latent variable `x` to a transformed space `y = transform(x)` with more amenable geometry. This reparametrizer is similar to `TransformReparam` but allows reparametrizations to be decoupled from the model declaration.

### DIFF
--- a/numpyro/infer/reparam.py
+++ b/numpyro/infer/reparam.py
@@ -3,6 +3,7 @@
 
 from abc import ABC, abstractmethod
 import math
+from typing import Iterable
 
 import numpy as np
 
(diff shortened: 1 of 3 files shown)


numpyro/infer/reparam.py
```diff
@@ -346,3 +347,48 @@ def __call__(self, name, fn, obs):
...
+            isinstance(t, dist.transforms.Transform) for t in transform
+        ):
+            transform = dist.transforms.ComposeTransform(transform)
+        self.transform = transform
+
+    def __call__(self, name, fn, obs):
+        assert obs is None, "ExplicitReparam does not support observe statements"
+        transformed = dist.TransformedDistribution(fn, self.transform)
+        x = numpyro.sample(f"{name}_base", transformed)
+        return None, self.transform.inv(x)
```
fehiepsi, line 394: Thanks, @tillahoffmann! I think a simple solution is to subclass the distribution and redefine the support (e.g. using `constraints.softplus_positive`). Using the explicit transform is more expensive because we are transforming the base value to the original domain two times (here and while computing the log probability of the base value). It is more suitable for PyTorch distribution, where caching is supported.

fehiepsi (approved):

numpyro/infer/reparam.py
fehiepsi: Oops, I already merged. sorry

PR #1758, opened 2024-03-14
Add Initial SDVI Implementation

treigerm:
This is follow-up work from https://github.com/pyro-ppl/numpyro/pull/1715 and now adding Support Decomposition VI (SDVI) which is a variational method for programs with stochastic support as discussed in  #1697.

Once this is merged I will add a simple tutorial on how to use these algorithms properly. At the moment, only the most basic version of DCC and SDVI are implemented, so the idea is that over time we can gradually add more bells and whistles (most prominently run inference in different program paths in parallel). For now, I wanted to keep the implementation simple to keep the PRs at a reasonable size.

### DIFF
--- a/docs/source/contrib.rst
+++ b/docs/source/contrib.rst
@@ -77,7 +77,19 @@ SteinVI Kernels
 Stochastic Support
 ~~~~~~~~~~~~~~~~~~
 
+.. autoclass:: numpyro.contrib.stochastic_support.dcc.StochasticSupportInference
+    :members:
+    :undoc-members:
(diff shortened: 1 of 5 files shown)


treigerm:
Thanks @fehiepsi! There wasn't anything in particular that I think needed attention. So if there's nothing which you think sticks out as problematic, I am happy for you to merge!

fehiepsi (approved):

fehiepsi:
Yeah, your plan sounds good to me.

PR #1762, opened 2024-03-15
Add complex constraint and real Fourier transform.

tillahoffmann:
This PR

- adds a `_Complex` constraint,
- modifies the `_Real` constraint to inherit from `_Complex` and check that values are indeed real,
- adds `RealFastFourierTransform`.

The `RealFastFourierTransform` interface differs slightly from `jax.numpy.fft.rfftn` to respect that batch dimensions precede  event dimensions. I.e., rather than specifying the axes along which to transform, one specifies the number of dimensions to transform akin to `reinterpreted_batch_ndims` in the `IndependentTransform`. The interface does not support specifying the `norm` parameter because `jit`-ing isn't happy with string arguments (cf. https://github.com/google/jax/issues/3045).

The motivation for this PR is to implement fast Gaussian processes for stationary kernels (cf. section 4 in https://arxiv.org/pdf/2301.08836v4.pdf).


numpyro/distributions/constraints.py
```diff
@@ -638,6 +639,11 @@ def feasible_like(self, prototype):
         return jax.numpy.zeros_like(prototype)
 
 
+class _Real(_Complex):
+    def __call__(self, x):
+        return super().__call__(x) & (jax.numpy.isreal(x))
```
tillahoffmann, line 644: Surprisingly, adding the `&` leads to some warnings *not* being emitted in `test/test_distributions.py::test_distribution_constraints`, but only when `jit`-ed.

fehiepsi: I think `isreal` is a jax function, which will generate a tracer under `jit`.

One option is to switch to np like in PositiveDefinite. It is better to have separate instances for real and complex though. I don't think that current constraints work for complex numbers.

numpyro/distributions/transforms.py
```diff
@@ -1200,6 +1201,80 @@ def __eq__(self, other):
...
+    :param ndims: Number of trailing dimensions to transform.
+    """
+
+    domain = constraints.dependent
+    codomain = constraints.dependent
+
+    def __init__(
+        self,
+        shape=None,
+        ndims=1,
```
fehiepsi, line 1225: maybe using `event_dim` to match patterns in constraints. Could you elaborate on why we need this attribute?

numpyro/distributions/transforms.py
```diff
@@ -1200,6 +1201,80 @@ def __eq__(self, other):
...
+    def inverse_shape(self, shape: tuple) -> tuple:
+        if self.shape:
+            return _normalize_rfft_shape(shape, self.shape)
+        size = 2 * (shape[-1] - 1)
+        return shape[:-1] + (size,)
+
+    def log_abs_det_jacobian(
+        self, x: jnp.ndarray, y: jnp.ndarray, intermediates: None = None
+    ) -> jnp.ndarray:
+        return 0.0
```
fehiepsi, line 1260: Could you return the whole batch shape for this?

numpyro/distributions/transforms.py
```diff
@@ -1200,6 +1201,80 @@ def __eq__(self, other):
...
+    """
+    N-dimensional discrete fast Fourier transform for real input.
+
+    :param shape: Length of each transformed axis to use from the input, defaults to the
+        input size.
+    :param ndims: Number of trailing dimensions to transform.
+    """
+
+    domain = constraints.dependent
+    codomain = constraints.dependent
```
fehiepsi, line 1220: maybe remove those attributes? You are redefining them below.

numpyro/distributions/transforms.py
```diff
@@ -1200,6 +1201,80 @@ def __eq__(self, other):
...
+
+    def __init__(
+        self,
+        shape=None,
```
fehiepsi, line 1224: The name `shape` is confusing. Maybe transform_length or something?

numpyro/distributions/transforms.py
tillahoffmann: The motivation is to specify the number of dimensions along which to transform, e.g., `ndim=1` gives the one-dimensional rfft, `ndim=2` gives the two-dimensional rfft, etc. Maybe `event_ndims` or `transform_ndims` for consistency with `reinterpreted_batch_ndims` in `IndependentTransform`?

fehiepsi: Oh, I didn't notice that you are supporting n-dimensional fft. Thanks for explaning!
Re event_ndims: looks better to me. We used `event_dim` just to be consistent with initial distribution designs.

fehiepsi: Actually, I like `transform_shape` and `transform_ndims` better. It is up to you. :)

tillahoffmann: I've kept it with `transform_ndims` and `transform_shape` for now. I was thinking `shape` could be good because that's the terminology numpy uses (although they just call the parameter `s`).

numpyro/distributions/transforms.py
```diff
@@ -1207,6 +1208,87 @@ def __eq__(self, other):
...
+            "transform_shape": self.transform_shape,
+            "transform_ndims": self.transform_ndims,
+        }
+        return (), ((), aux_data)
+
+    @property
+    def domain(self) -> constraints.Constraint:
+        return constraints._IndependentConstraint(
+            constraints._Real(), self.transform_ndims
+        )
```
fehiepsi, line 1276: maybe `constraints.independent(constraints.real, ...)`

numpyro/distributions/transforms.py
```diff
@@ -1207,6 +1208,87 @@ def __eq__(self, other):
...
+
+    @property
+    def codomain(self) -> constraints.Constraint:
+        return constraints._IndependentConstraint(
+            constraints._Complex(), self.transform_ndims
```
fehiepsi, line 1281: similar to above

numpyro/distributions/transforms.py
```diff
@@ -1207,6 +1208,87 @@ def __eq__(self, other):
...
+    def inverse_shape(self, shape: tuple) -> tuple:
+        if self.transform_shape:
+            return _normalize_rfft_shape(shape, self.transform_shape)
+        size = 2 * (shape[-1] - 1)
+        return shape[:-1] + (size,)
+
+    def log_abs_det_jacobian(
+        self, x: jnp.ndarray, y: jnp.ndarray, intermediates: None = None
+    ) -> jnp.ndarray:
+        return jnp.ones_like(x, shape=x.shape[: -self.transform_ndims])
```
fehiepsi, line 1263: typo: zeros_like

I think you need to broadcast `x.shape[: -self.transform_ndims]` and `y.shape[: -self.transform_ndims]`

numpyro/distributions/constraints.py
```diff
@@ -638,6 +639,15 @@ def feasible_like(self, prototype):
         return jax.numpy.zeros_like(prototype)
 
 
+class _Real(_SingletonConstraint):
+    def __call__(self, x):
+        # XXX: consider to relax this condition to [-inf, inf] interval
+        return (x == x) & (x != float("inf")) & (x != float("-inf")) & np.isreal(x)
```
fehiepsi, line 645: Could you remove `np.isreal` here? We don't check for real input across all constraints. If x is a tracer, np.isreal might not work.

fehiepsi:
Awesome work!! thanks, Till!

fehiepsi (approved):

fehiepsi (approved):

numpyro/distributions/transforms.py
```diff
@@ -1207,6 +1208,86 @@ def __eq__(self, other):
...
+        size = 2 * (shape[-1] - 1)
+        return shape[:-1] + (size,)
+
+    def log_abs_det_jacobian(
+        self, x: jnp.ndarray, y: jnp.ndarray, intermediates: None = None
+    ) -> jnp.ndarray:
+        shape = jnp.broadcast_shapes(
+            x.shape[: -self.transform_ndims], y.shape[: -self.transform_ndims]
+        )
+        return jnp.ones_like(x, shape=shape)
```
fehiepsi, line 1266: oops, wouldn't this zeros_like?

tillahoffmann: Yes, sorry. Fixed.

tillahoffmann:
Thank you for the fast review and merging! 

I just realized we probably want to add the option to unpack the complex coefficients with shape `(..., n // 2 + 1)` for a signal of shape `(..., n)` to a real tensor with shape `(..., n)` because the rest of the library is designed for real-valued tensors. This would be equivalent to what we're doing in Stan [here](https://github.com/onnela-lab/gptools/blob/567a6a4761e374266ff932d92a7b9f2d154ba850/stan/gptools/stan/gptools/fft1.stan#L33-L51).

One option is to add an argument `unpack` (or some more informative name) to the transform to indicate if real or complex values should be returned. Another is to add a separate unpacking transform. What do you think?

fehiepsi:
Interesting, I feel that it's better to have a separate transform for the unpack version.

tillahoffmann:
That could look like

```python
class RealFastFourierUnpackTransform(Transform):
    """
    :param size: Size of the last dimension of the transform, required because the size
        cannot be inferred from the shape of the coefficients.
    """
    def __init__(self, size):
        self.size = size

    ...
```

If we added it with an `unpack` parameter, we could infer the size from the argument `x`. I agree that having a separate transform seems like a good idea, but it would also be nice if one doesn't have to specify the shape ahead of time, e.g., if one wants to use the same transform object on different size input. Not sure which one is the better compromise.

PR #1766, opened 2024-03-20
Add `RecursiveLinearTransform` for linear state space models.

tillahoffmann:
This PR adds a `RecursiveLinearTransform` which is a linear transformation applied recursively such that $y_t = A y_{t - 1} + x_t$ for $t > 0$, where $x_t$ and $y_t$ are $p$-vectors and $A$ is a $p\times p$ transition matrix. The series is initialized by $y_0 = 0$.

This transform can be used to easily declare linear state space models, e.g., a Cauchy random walk is

```python
>>> from jax import random
>>> import numpyro
>>> from numpyro import distributions as dist
>>>
>>> def cauchy_random_walk():
...     return numpyro.sample(
...         "x",
...         dist.TransformedDistribution(
...             dist.Cauchy(0, 1).expand([10, 1]).to_event(1),
...             dist.transforms.RecursiveLinearTransform(jnp.eye(1)),
...         ),
...     )
```

A Kalman-style model for a rocket with state `y = (position, velocity)` is

```python
>>> def rocket_trajectory():
...     scale = numpyro.sample(
...         "scale",
...         dist.HalfCauchy(1).expand([2]).to_event(1),
...     )
...     transition_matrix = jnp.array([[1, 1], [0, 1]])
...     return numpyro.sample(
...         "x",
...         dist.TransformedDistribution(
...             dist.Normal(0, scale).expand([10, 2]).to_event(1),
...             dist.transforms.RecursiveLinearTransform(transition_matrix),
...         ),
...     )
```

This PR also makes a few minor changes (happy to factor out if you prefer):

- Reformat the `ExplicitReparam` from #1754 to comply with the stricter linting.
- Add the `RealFastFourierTransform` from #1762 to the documentation.
- Ignore `venv` directory in the `update_headers.py` script.
- Add autogenerated documentation sources `docs/source/{examples,tutorials,getting_started.rst}` to `.gitignore`.
- Verify `log_abs_det_jacobian` implementation using autodiff.

### DIFF
--- a/numpyro/distributions/transforms.py
+++ b/numpyro/distributions/transforms.py
@@ -1288,6 +1288,98 @@ class RealFastFourierTransform(Transform):
         )
 
 
+class RecursiveLinearTransform(Transform):
+    """
+    Apply a linear transformation recursively such that
+    :math:`y_t = A y_{t - 1} + x_t` for :math:`t > 0`, where :math:`x_t` and :math:`y_t`
+    are vectors and :math:`A` is a transition matrix. The series is initialized by
+    :math:`y_0 = 0`.
+
+    :param transition_matrix: Transition matrix :math:`A` for successive states.
+
+    **Example:**
+
+    .. doctest::
+
+        >>> from jax import random
+        >>> from jax import numpy as jnp
+        >>> import numpyro
+        >>> from numpyro import distributions as dist
+        >>>
+        >>> def cauchy_random_walk():
+        ...     return numpyro.sample(
+        ...         "x",
+        ...         dist.TransformedDistribution(
+        ...             dist.Cauchy(0, 1).expand([10, 1]).to_event(1),
+        ...             dist.transforms.RecursiveLinearTransform(jnp.eye(1)),
+        ...         ),
+        ...     )
+        >>>
+        >>> numpyro.handlers.seed(cauchy_random_walk, 0)().shape
+        (10, 1)
+        >>>
+        >>> def rocket_trajectory():
+        ...     scale = numpyro.sample(
+        ...         "scale",
+        ...         dist.HalfCauchy(1).expand([2]).to_event(1),
+        ...     )
+        ...     transition_matrix = jnp.array([[1, 1], [0, 1]])
+        ...     return numpyro.sample(
+        ...         "x",
+        ...         dist.TransformedDistribution(
+        ...             dist.Normal(0, scale).expand([10, 2]).to_event(1),
+        ...             dist.transforms.RecursiveLinearTransform(transition_matrix),
+        ...         ),
+        ...     )
+        >>>
+        >>> numpyro.handlers.seed(rocket_trajectory, 0)().shape
+        (10, 2)
+    """
+
+    domain = constraints.real_matrix
+    codomain = constraints.real_matrix
+
+    def __init__(self, transition_matrix: jnp.ndarray) -> None:
+        self.transition_matrix = transition_matrix
+
+    def __call__(self, x: jnp.ndarray) -> jnp.ndarray:
+        # Move the time axis to the first position so we can scan over it.
+        x = jnp.moveaxis(x, -2, 0)
+
+        def f(y, x):
+            y = (self.transition_matrix * y[..., None, :]).sum(axis=-1) + x
+            return y, y
+
+        _, y = lax.scan(f, jnp.zeros_like(x, shape=x.shape[1:]), x)
+        return jnp.moveaxis(y, 0, -2)
+
+    def _inverse(self, y: jnp.ndarray) -> jnp.ndarray:
+        # Move the time axis to the first position so we can scan over it.
+        y = jnp.moveaxis(y, -2, 0)
+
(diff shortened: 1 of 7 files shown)


numpyro/distributions/transforms.py
```diff
@@ -1347,26 +1347,25 @@ def __call__(self, x: jnp.ndarray) -> jnp.ndarray:
         x = jnp.moveaxis(x, -2, 0)
 
         def f(y, x):
-            y = (self.transition_matrix * y[..., None, :]).sum(axis=-1) + x
+            y = y @ self.transition_matrix.T + x
```
fehiepsi, line 1350: I think you might want `jnp.swapaxes(self.transition_matrix, -1, -2)`

numpyro/distributions/transforms.py
```diff
@@ -1347,26 +1347,25 @@ def __call__(self, x: jnp.ndarray) -> jnp.ndarray:
...
     def _inverse(self, y: jnp.ndarray) -> jnp.ndarray:
-        # Move the time axis to the first position so we can scan over it.
+        # Move the time axis to the first position so we can scan over it in reverse.
         y = jnp.moveaxis(y, -2, 0)
 
-        def f(y_t, y_tm1):
-            x = y_t - (self.transition_matrix * y_tm1[..., None, :]).sum(axis=-1)
-            return y_tm1, x
+        def f(y, prev):
+            x = y - prev @ self.transition_matrix.T
```
fehiepsi, line 1361: similarly, `jnp.swapaxes(self.transition_matrix, -1, -2)`

numpyro/distributions/transforms.py
```diff
@@ -1347,26 +1347,25 @@ def __call__(self, x: jnp.ndarray) -> jnp.ndarray:
...
+            return prev, x
 
         _, x = lax.scan(f, y[-1], jnp.roll(y, 1, axis=0).at[0].set(0), reverse=True)
         return jnp.moveaxis(x, 0, -2)
 
     def log_abs_det_jacobian(self, x: jnp.ndarray, y: jnp.ndarray, intermediates=None):
-        slogdet = jnp.linalg.slogdet(self.transition_matrix)
-        return jnp.broadcast_to(slogdet.logabsdet, x.shape[:-2]) * x.shape[-2]
+        return jnp.zeros_like(x, shape=x.shape[:-2])
```
fehiepsi, line 1368: sounds reasonable to me, this is sort of a shear transformation, so the Jacobian determinant is 1.

tillahoffmann: Yes, because of the temporal nature of the transform, the Jacobian is triangular. Because the `x` only appears additively, the diagonal is one, leading to the unit Jacobian.

numpyro/distributions/transforms.py
```diff
@@ -1288,6 +1288,98 @@ def __eq__(self, other):
...
 
 
+class RecursiveLinearTransform(Transform):
+    """
+    Apply a linear transformation recursively such that
+    :math:`y_t = A y_{t - 1} + x_t` for :math:`t > 0`, where :math:`x_t` and :math:`y_t`
+    are vectors and :math:`A` is a transition matrix. The series is initialized by
+    :math:`y_0 = 0`.
+
+    :param transition_matrix: Transition matrix :math:`A` for successive states.
```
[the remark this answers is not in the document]
tillahoffmann: > Maybe `'matrix'` -> `squared matrix` for clarity.

👍 

> Currently, the bias `x` is time-dependent, but the matrix is constant. Do you plan to make the class name more verbose to reflect that?

Sorry, I explained poorly. The `x` here is the argument of the transform rather than the bias in the `AffineTransform`, for example. Keeping the transition matrix constant means that the transform can be applied to sequences of arbitrary length.

numpyro/distributions/transforms.py
tillahoffmann: I've updated it to use `einsum` because we have shapes `(..., p, p)` for the transition matrices `A` and `(..., n, p)` for the states `x`. Inside the `scan` function, we are dealing with state of shape `(..., p)` because we're scanning along the `n` dimension. There may very well be a better way to do this.

numpyro/distributions/transforms.py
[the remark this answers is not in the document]
fehiepsi: x is the (time-dependent) bias (or noise if we place a normal distribution over it) of a linear dynamical model. Typically, people use other notations, something like x_t = Ax{t-1} + b_t.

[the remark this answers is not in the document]
tillahoffmann: Fair. I was using the x/y notation here to stick with the typical arguments of the `Transform` classes. Do you have an idea for a more explanatory name?

fehiepsi (approved):

fehiepsi:
Thanks, @tillahoffmann!

numpyro/distributions/transforms.py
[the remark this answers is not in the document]
fehiepsi: The name looks 👌 to me. :)

Issue #1538, opened 2023-02-13
Wishart distributions?


saraelshawa:
Hi, 

Is it possible to sample from a Wishart distribution using numpyro? 

Thank in advance!

fehiepsi:
I think so. You can use TFP distributions with numpyro: e.g. [WishartTriL](https://www.tensorflow.org/probability/api_docs/python/tfp/distributions/WishartTriL).

saraelshawa:
Thank you. I was having issues with `WishartTril` using TFP so I thought numpyro might have it's own implementation :)

fehiepsi:
I guess you can use [LKJCholesky](https://num.pyro.ai/en/stable/distributions.html#lkjcholesky) instead of `WishartTril`. We don't have Wishart distribution implemented.

PR #1770, opened 2024-03-29
Verify result of `biject_to` satisfies the constraint.

tillahoffmann:
Following on from #1751, this PR adds a test that the result of calling `biject_to` satisfies the desired constraint. It also updates the correlation matrix constraint to reduce sensitivity to numerical errors. cc @kylejcaron

### DIFF
--- a/test/test_transforms.py
+++ b/test/test_transforms.py
@@ -10,6 +10,7 @@ import pytest
 from jax import jacfwd, jit, random, tree_map, vmap
 import jax.numpy as jnp
 
+from numpyro.distributions import constraints
 from numpyro.distributions.flows import (
     BlockNeuralAutoregressiveTransform,
     InverseAutoregressiveTransform,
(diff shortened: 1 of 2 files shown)


fehiepsi (approved):
Thanks, @tillahoffmann! while you are on this transform test, could you also change the global jax arrays in this file to numpy arrays? We don't want to execute jax code before setting global information like platform or precision.

test/test_transforms.py
```diff
@@ -351,3 +348,41 @@ def test_batched_recursive_linear_transform():
...
     assert jnp.allclose(x, transform.inv(y), atol=1e-6)
+
+
+@pytest.mark.parametrize(
+    "constraint, shape",
+    [
+        (constraints.circular, (3,)),
+        (constraints.complex, (3,)),
+        (constraints.corr_cholesky, (10, 10)),
+        (constraints.corr_matrix, (21,)),
```
fehiepsi, line 359: @tillahoffmann This test is failing in my system with jax dev. Could we relax this to 15 (which passes)?

tillahoffmann: Sounds good. I can send a mini PR or integrate it into #1538?

fehiepsi: Either of them sounds good to me. Thanks, Till!

PR #1775, opened 2024-04-04
Remove the deprecated import jax.tree_map

fehiepsi:
Also fixing the `enable_x64` logic which is not compatible with the latest version of jax. (we need to use a boolean instead of an integer for the setting)


OlaRonning (approved):
lgtm

PR #1776, opened 2024-04-09
Avoid unnecessary reshape for trivial expand

fehiepsi:
This PR saves unnecessary transpose, reshape operators of ExpandedDistribution's sample method. The current tests should cover the change.

This requires us to revert the order of `expanded_sizes` such that the dimensions are ordered from left to right (rather than the current right -> left).


OlaRonning (approved):
It took a while to trigger the interstitial_sizes case, but everything seems to work as expected.

fehiepsi:
Thanks, Ola!

PR #1777, opened 2024-04-10
Allow to use Delta on numpy arrays without moving them to jax devices

fehiepsi:
Currently, the `broadcast_to` operator in `sample` converts numpy arrays into jax arrays. Keeping the original type would be nice in case we only want Delta to hold some value in the trace.

### DIFF
--- a/numpyro/distributions/distribution.py
+++ b/numpyro/distributions/distribution.py
@@ -1138,6 +1138,8 @@ class Delta(Distribution):
         return constraints.independent(constraints.real, self.event_dim)
 
     def sample(self, key, sample_shape=()):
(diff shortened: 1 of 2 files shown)


OlaRonning:
@fehiepsi I checked a few small examples, and they work fine, but I don't understand the use case.

fehiepsi:
Hi Ola, in one of my applications, I want to store a dictionary in the trace. Having this functionality would be convenient. I can create a custom Delta distribution for that, but using the default one would be nicer.

OlaRonning (approved):
Thanks for the explanation. LGTM.

PR #1779, opened 2024-04-11
Add Wishart distribution.

tillahoffmann:
This PR adds the Wishart distribution for covariance matrices. It should probably be rebased and reviewed after #1778 is merged.


numpyro/distributions/continuous.py
```diff
@@ -2533,3 +2533,133 @@ def variance(self):
...
+        concentration=None,
+        scale_matrix=None,
+        rate_matrix=None,
+        scale_tril=None,
+        validate_args=None,
+    ):
+        # Determine the shapes.
+        batch_shape = None
+        event_shape = None
+        for x in [scale_matrix, rate_matrix, scale_tril]:
```
fehiepsi, line 2576: Can we assert that only one of those arguments needs to be specified?

numpyro/distributions/continuous.py
```diff
@@ -2533,3 +2533,133 @@ def variance(self):
...
+                "One of `scale_matrix`, `rate_matrix`, or `scale_tril` must be "
+                "specified."
+            )
+
+        # Coerce to scale_tril parameter.
+        if scale_matrix is not None:
+            self.scale_matrix = jnp.broadcast_to(
+                scale_matrix, batch_shape + event_shape
+            )
+            self.scale_tril = jnp.linalg.cholesky(scale_matrix)
```
fehiepsi, line 2594: We need to promote shapes like in MultivariateNormal distribution. No need to broadcast. This saves computation when `concentration` is batched.

numpyro/distributions/continuous.py
```diff
@@ -2533,3 +2533,133 @@ def variance(self):
...
+            batch_shape=batch_shape,
+            event_shape=event_shape,
+            validate_args=validate_args,
+        )
+
+    @validate_sample
+    def log_prob(self, value):
+        p = value.shape[-1]
+        rate_matrix, value = jnp.broadcast_arrays(self.rate_matrix, value)
+        trace = jnp.trace(lax.batch_matmul(rate_matrix, value), axis1=-1, axis2=-2)
```
fehiepsi, line 2613: Just curious, why we need to use batch_matmul?

numpyro/distributions/continuous.py
```diff
@@ -2533,3 +2533,133 @@ def variance(self):
...
+        super().__init__(
+            batch_shape=batch_shape,
+            event_shape=event_shape,
+            validate_args=validate_args,
+        )
+
+    @validate_sample
+    def log_prob(self, value):
+        p = value.shape[-1]
+        rate_matrix, value = jnp.broadcast_arrays(self.rate_matrix, value)
```
fehiepsi, line 2612: I think it is more efficient if we use `scale_tril` for log_prob computation, like in the MVN distribution.

numpyro/distributions/continuous.py
```diff
@@ -2533,3 +2533,133 @@ def variance(self):
...
+        trace = jnp.trace(lax.batch_matmul(rate_matrix, value), axis1=-1, axis2=-2)
+        return (
+            (self.concentration - p - 1) / 2 * jnp.linalg.slogdet(value).logabsdet
+            - trace / 2
+            - self.concentration * p / 2 * jnp.log(2)
+            - multigammaln(self.concentration / 2, p)
+            + self.concentration * jnp.linalg.slogdet(rate_matrix).logabsdet / 2
```
fehiepsi, line 2619: `slogdet` can be computed efficiently with `scale_tril`.

numpyro/distributions/continuous.py
```diff
@@ -2533,3 +2533,133 @@ def variance(self):
...
+            validate_args=validate_args,
+        )
+
+    @validate_sample
+    def log_prob(self, value):
+        p = value.shape[-1]
+        rate_matrix, value = jnp.broadcast_arrays(self.rate_matrix, value)
+        trace = jnp.trace(lax.batch_matmul(rate_matrix, value), axis1=-1, axis2=-2)
+        return (
+            (self.concentration - p - 1) / 2 * jnp.linalg.slogdet(value).logabsdet
```
fehiepsi, line 2615: Do you think it is better to have a lower cholesky version of Wishart? I feel that working directly on the positive definite space might lead to numerical issues. (similar to LKJ and LKJCholesky)

fehiepsi (commented):
Thanks @tillahoffmann! It is great to have this distribution. I left initial comments below.

numpyro/distributions/continuous.py
tillahoffmann: Sounds good, I'll have a look at using the batch Mahalanobis functions.

numpyro/distributions/continuous.py
tillahoffmann: I always get a bit flustered by the semantics of `matmul` depending on the dimensionality of the arguments. But, here, regular `matmul` does the trick because both arguments are at least 2d. I'll change that.

numpyro/distributions/continuous.py
tillahoffmann: Yes, that sounds like a good idea. I'll have a look at the LKJ implementations.

numpyro/distributions/continuous.py
tillahoffmann: Yup, I'll also update `MultivariateNormal` to assert exactly one of them is not `None`.

numpyro/distributions/util.py
```diff
@@ -623,6 +623,17 @@ def is_prng_key(key):
         return False
 
 
+def assert_one_of(**kwargs):
```
tillahoffmann, line 626: Added this `assert_one_of` function because the pattern appears commonly across the code base.

test/test_distributions.py
```diff
@@ -1146,8 +1233,15 @@ def test_dist_shape(jax_dist, sp_dist, params, prepend_shape):
     "jax_dist, sp_dist, params", CONTINUOUS + DISCRETE + DIRECTIONAL
 )
 def test_infer_shapes(jax_dist, sp_dist, params):
-    shapes = tuple(getattr(p, "shape", ()) for p in params)
-    shapes = tuple(x() if callable(x) else x for x in shapes)
+    shapes = []
```
tillahoffmann, line 1237: The current shape construction provides an empty tuple if the parameter is `None`. This meant that `infer_shape` on `MultivariateNormal` worked because it iterated over all shape parameters even though it should probably stop when it encounters the first specified shape (updated in this PR). This change leaves the shape as `None` if the parameter is `None`.

test/test_transforms.py
```diff
@@ -360,12 +360,12 @@ def test_batched_recursive_linear_transform():
         (constraints.circular, (3,)),
         (constraints.complex, (3,)),
         (constraints.corr_cholesky, (10, 10)),
-        (constraints.corr_matrix, (21,)),
```
tillahoffmann, line 363: As discussed in #1770.

numpyro/distributions/continuous.py
```diff
@@ -2533,3 +2540,250 @@ def variance(self):
...
+        )
+        i, j = jnp.tril_indices(p, -1)
+        assert i.size == p * (p - 1) // 2
+        latent = latent.at[..., i, j].set(
+            random.normal(rng_offdiag, latent.shape[:-2] + (i.size,))
+        )
+        return jnp.matmul(*jnp.broadcast_arrays(self.scale_tril, latent))
+
+    @lazy_property
+    def mean(self):
```
tillahoffmann, line 2754: The `mean`, `variance`, and `log_prob` implementations seem to be working fine but were derived on the back of an envelope. A second pair of eyes on whether these make sense would be great!

numpyro/distributions/continuous.py
```diff
@@ -2615,3 +2623,260 @@ def variance(self):
...
+    support = constraints.positive_definite
+    reparametrized_params = [
+        "scale_matrix",
+        "rate_matrix",
+        "scale_tril",
+    ]
+
+    def __init__(
+        self,
+        concentration=None,
```
fehiepsi, line 2657: nit: remove `=None`

fehiepsi (commented):
Sorry for the late review, @tillahoffmann!

numpyro/distributions/continuous.py
```diff
@@ -2615,3 +2623,260 @@ def variance(self):
...
+        concentration=(), scale_matrix=None, rate_matrix=None, scale_tril=None
+    ):
+        return WishartCholesky.infer_shapes(
+            concentration, scale_matrix, rate_matrix, scale_tril
+        )
+
+    def entropy(self):
+        p = self.event_shape[-1]
+        return (
+            (p + 1) * jnp.linalg.slogdet(self.scale_tril).logabsdet
```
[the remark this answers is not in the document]
tillahoffmann: Yes, nice catch! Updated.

numpyro/distributions/continuous.py
```diff
@@ -2615,3 +2623,260 @@ def variance(self):
...
+            + p * (p + 1) / 2 * jnp.log(2)
+            + multigammaln(self.concentration / 2, p)
```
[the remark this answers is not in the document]
tillahoffmann: I think that's the implementation being used here. The import is on line 45. I added a custom implementation of `multidigamma` which I could not locate in the `jax.scipy.special` module.

numpyro/distributions/continuous.py
fehiepsi: I couldn't follow the math but I trust the test. Really appreciate your work!

numpyro/distributions/continuous.py
```diff
@@ -2615,3 +2623,261 @@ def variance(self):
...
+        # scale_tril. Thus,
+        # t = trace(value.T @ inv(scale_tril).T @ inv(scale_tril) @ value), and we can
+        # rewrite as t = trace(x.T @ x) for x = inv(scale_tril) @ value which we can
+        # obtain easily by solving a triangular system. x is again triangular such that
+        # trace(x @ x.T) is equal to the sum of squares of elements.
+        x = solve_triangular(*jnp.broadcast_arrays(self.scale_tril, value), lower=True)
+        trace = jnp.square(x).sum(axis=(-1, -2))
+        p = value.shape[-1]
+        return (
+            (self.concentration - p - 1) * jnp.linalg.slogdet(value).logabsdet
```
fehiepsi, line 2802: slogdet can be computed using the diagonal part of value

numpyro/distributions/continuous.py
```diff
@@ -2615,3 +2623,261 @@ def variance(self):
...
+            - trace / 2
+            + p * (1 - self.concentration / 2) * jnp.log(2)
+            - multigammaln(self.concentration / 2, p)
+            - self.concentration * jnp.linalg.slogdet(self.scale_tril).logabsdet
```
fehiepsi, line 2806: similar to above, slogdet can be computed efficiently

fehiepsi (commented):
Thanks, @tillahoffmann! Just have a comment at the computation of slogdet.

numpyro/distributions/continuous.py
```diff
@@ -1562,12 +1565,17 @@ def variance(self):
...
+            precision_matrix=precision_matrix,
+            scale_tril=scale_tril,
+        )
         batch_shape, event_shape = loc[:-1], loc[-1:]
         for matrix in [covariance_matrix, precision_matrix, scale_tril]:
             if matrix is not None:
                 batch_shape = lax.broadcast_shapes(batch_shape, matrix[:-2])
                 event_shape = lax.broadcast_shapes(event_shape, matrix[-1:])
-        return batch_shape, event_shape
+                return batch_shape, event_shape
```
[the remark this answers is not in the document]
tillahoffmann: Yes, `assert_one_of` ensures that exactly one of the matrices is available, and returning early avoids looping over `precision_matrix` and `scale_tril` if `covariance_matrix` is given, for example. Either with/without the indent should work though.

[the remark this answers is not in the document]
fehiepsi: got it, thanks!

fehiepsi (approved):

PR #1783, opened 2024-04-17
Fix image sizes censoring notebook

juanitorduz:
When looking into https://num.pyro.ai/en/latest/tutorials/censoring.html it seems the figures are a bit too wide. This PR makes them more even so that they look better in the docs. No other change has been made to the notebook :)


fehiepsi (approved):

fehiepsi:
Thanks, @juanitorduz!

PR #1784, opened 2024-04-18
Relax check of positive definite constraint.

tillahoffmann:
This PR brings the check for positive definite matrices in line with the check for correlation matrices which allows small variations due to numerical issues.

### Before

```python
>>> from jax import random
>>> from numpyro.distributions.constraints import positive_definite

>>> x = jnp.eye(3)
>>> positive_definite(x)
Array(True, dtype=bool)

>>> y = x + 1e-15 * random.normal(random.key(7), x.shape)
>>> positive_definite(y)
Array(False, dtype=bool)
```

### After

```python
>>> positive_definite(y)
Array(True, dtype=bool)
```


fehiepsi (approved):

fehiepsi:
Thanks Till!

PR #1787, opened 2024-04-25
Add entropy implementations.

tillahoffmann:
This PR adds entropy implementations, using `scipy.stats` distributions as the reference implementation for tests where possible. There is some overlap with #1706.

Currently, only one `entropy` implementation is available (for the `LowRankMultivariateNormal`). It might be useful to switch `entropy` to use `lazy_property` as for `mean` and `variance`. That would make the interface consistent (entropy is a property of the distribution like mean or variance), and it would allow caching if the evaluation is more involved. What do you think, @fehiepsi?


numpyro/distributions/continuous.py
```diff
@@ -198,6 +199,17 @@ def cdf(self, value):
     def icdf(self, q):
         return betaincinv(self.concentration1, self.concentration0, q)
 
+    def entropy(self):
+        total = self.concentration0 + self.concentration1
+        return (
+            gammaln(self.concentration0)
+            + gammaln(self.concentration1)
+            - gammaln(total)
```
fehiepsi, line 207: you can use `betaln` for this. It is more stable I think.

fehiepsi (commented):
Thanks, @tillahoffmann! I agree that making it a lazy property would be better. But we'll want to keep the api be consistent with pyro/pytorch distributions. If you worry about the performance, I guess we can have the lazy property `_entropy`. What do you think?

fehiepsi (approved):

PR #1789, opened 2024-04-30
Predictive fix when deterministic sites are present

### LINKED ISSUE #1772: numpyro.deterministic static on infer.Predictive
For some reason after fitting the model the numpyro.deterministic shape remains static, after trying to predict with a different shape it throws a shape error.

Example in lightweight-mmm:

```
# extra_features.shape = (10,3) / trying to predict 10 new time periods

extra_features_effect = numpyro.deterministic(
    name="extra_features_effect",
    value=jnp.einsum(
        extra_features_einsum, extra_features, coef_extra_features
    ),
)

# extra_features_effect.shape = (30,3)  / output is resized to the size of the model as when fit; 30 periods
```

This throws a size error, see:
https://github.com/google/lightweight_mmm/issues/309
and
https://github.com/google/lightweight_mmm/issues/308

kylejcaron:
This PR attempts to fix #1772 - when deterministic sites are included in `Predictive` with params, `Predictive` won't generate new samples for those deterministic sites. 

This PR solves that by ignoring deterministic sites in the Predictive `substitute` call. Added a new test to cover this scenario as well.

### DIFF
--- a/numpyro/infer/util.py
+++ b/numpyro/infer/util.py
@@ -781,7 +781,7 @@ def _predictive(
             posterior_samples,
         )
         prototype_trace = trace(
-            seed(substitute(masked_model, prototype_sample), subkey)
+            seed(predictive_substitute(masked_model, prototype_sample), subkey)
         ).get_trace(*model_args, **model_kwargs)
         first_available_dim = -_guess_max_plate_nesting(prototype_trace) - 1
 
@@ -803,7 +803,7 @@ def _predictive(
             )
         else:
             model_trace = trace(
-                seed(substitute(masked_model, samples), rng_key)
+                seed(predictive_substitute(masked_model, samples), rng_key)
             ).get_trace(*model_args, **model_kwargs)
             pred_samples = {name: site["value"] for name, site in model_trace.items()}
 
--- a/numpyro/infer/util.py
+++ b/numpyro/infer/util.py
@@ -20,7 +20,14 @@ import numpyro
 from numpyro.distributions import constraints
 from numpyro.distributions.transforms import biject_to
 from numpyro.distributions.util import is_identically_one, sum_rightmost
-from numpyro.handlers import condition, replay, seed, substitute, trace
+from numpyro.handlers import (
+    condition,
+    predictive_substitute,
+    replay,
+    seed,
+    substitute,
+    trace,
+)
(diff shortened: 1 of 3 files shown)


numpyro/infer/util.py
```diff
@@ -795,8 +795,17 @@ def single_prediction(val):
...
+
+            def _samples_wo_deterministic(msg):
+                return (
+                    samples.get(msg["name"]) if msg["type"] != "deterministic" else None
+                )
+
             model_trace = trace(
-                seed(substitute(masked_model, samples), rng_key)
+                seed(
+                    substitute(masked_model, substitute_fn=_samples_wo_deterministic),
```
fehiepsi, line 806: Could you help me change line `777` to `condition` so that? It would be nice to add an argument like `exclude_deterministic_from_posterior` to `Predictive` to maintain two behaviors. We will pass such argument to this `_predictive` function to control the behavior.

kylejcaron: I added the conditional logic, I tried the name `exclude_deterministic_params` because it was shorter but let me know if that's insufficient

For this could you elaborate?
> Could you help me change line 777 to condition so that? 

Would I be changing L777 from `substitute` to `condition`?

And should I add the deterministic fix there as well?

numpyro/infer/util.py
```diff
@@ -870,6 +882,9 @@ class Predictive(object):
 
         + set `batch_ndims=1` to get predictions from a one dimensional batch of the guide and parameters
           with shapes `(num_samples x batch_size x ...)`
+    :param exclude_deterministic_params: indicates whether deterministic sites should get sampled
```
fehiepsi, line 885: maybe just `exclude_deterministic`? `params` are parameters of the model. The description could be: indicates whether to ignore deterministic sites from the posterior samples.

numpyro/infer/util.py
fehiepsi: Yup, using condition there is fine because we don't substitute deterministic sites under the`condition` handler.

kylejcaron: ok just switched it to `condition`

fehiepsi (approved):
Thanks for fixing the breakage, @kylejcaron! We'll make a patch release soon.

PR #1791, opened 2024-05-02
Different supports in component distributions for mixture models

### LINKED ISSUE #1780: [FR] Support for different supports in component distributions for mixture models
# Description
I am submitting a feature request to enhance the capabilities of Numpyro in handling mixture models with component distributions having different supports. Currently, when attempting to construct a mixture model using `numpyro.distributions.MixtureGeneral`, all component distributions are required to have the same support.

In my specific use case, I am implementing a truncated power law mixture model where one of the component distributions utilizes a truncated power law model with a high-mass cutoff represented by the Heaviside step function. The other component distribution involves a Gaussian distribution.

The formulation of the model is as follows:

$$
p(m_1\mid\theta) \propto (1-\lambda)A(\theta)m_1^{-\alpha}\Theta(m_{\text{max}}-m_1) + \lambda B(\theta)\exp\left(-\frac{(m_1-\mu_m)^2}{2\sigma_m^2}\right)
$$

Where $\theta$ represents the parameter vector $\left(\lambda,\alpha,m_{\text{min}},\delta_m,m_{\text{max}},\mu_m,\sigma_m\right)$, $A(\cdot)$ and $B(\cdot)$ are normalization factors.

To implement this model efficiently, it is essential to have the flexibility to use component distributions with varying supports within the same mixture model.

Therefore, I kindly request the implementation of this feature to enable users to construct mixture models with component distributions having different supports. If there is any solution which does not requires new feature, will be great.

Qazalbash:
Fixes #1780.

### DIFF
--- a/numpyro/distributions/mixtures.py
+++ b/numpyro/distributions/mixtures.py
@@ -7,7 +7,6 @@ import jax.numpy as jnp
 
 from numpyro.distributions import Distribution, constraints
 from numpyro.distributions.discrete import CategoricalLogits, CategoricalProbs
-from numpyro.distributions.util import validate_sample
 from numpyro.util import is_prng_key
 
 
@@ -145,7 +144,7 @@ class _MixtureBase(Distribution):
     def sample(self, key, sample_shape=()):
         return self.sample_with_intermediates(key=key, sample_shape=sample_shape)[0]
 
-    @validate_sample
+    # @validate_sample
     def log_prob(self, value, intermediates=None):
         del intermediates
         sum_log_probs = self.component_log_probs(value)
@@ -308,7 +307,7 @@ class MixtureGeneral(_MixtureBase):
         for d in component_distributions:
             if not isinstance(d, Distribution):
                 raise ValueError(
-                    "All elements of 'component_distributions' must be instaces of "
+                    "All elements of 'component_distributions' must be instances of "
                     "numpyro.distributions.Distribution subclasses"
                 )
         if len(component_distributions) != self.mixture_size:
@@ -320,11 +319,9 @@ class MixtureGeneral(_MixtureBase):
         # TODO: It would be good to check that the support of all the component
         # distributions match, but for now we just check the type, since __eq__
         # isn't consistently implemented for all support types.
-        support_type = type(component_distributions[0].support)
-        if any(
-            type(d.support) is not support_type for d in component_distributions[1:]
-        ):
-            raise ValueError("All component distributions must have the same support.")
+        # support_type = type(component_distributions[0].support)
+        # if any(type(d.support) is not support_type for d in component_distributions[1:]):
+        #     raise ValueError("All component distributions must have the same support.")
 
         self._mixing_distribution = mixing_distribution
         self._component_distributions = component_distributions
@@ -389,9 +386,21 @@ class MixtureGeneral(_MixtureBase):
         return jnp.stack(samples, axis=self.mixture_dim)
 
     def component_log_probs(self, value):
-        component_log_probs = jnp.stack(
-            [d.log_prob(value) for d in self.component_distributions], axis=-1
-        )
+        if self._validate_args:
+            mask = jnp.stack(
+                [d.support(value) for d in self.component_distributions], axis=-1
+            )
+            component_log_probs = jnp.where(
+                mask,
+                jnp.stack(
+                    [d.log_prob(value) for d in self.component_distributions], axis=-1
+                ),
+                -jnp.inf,
+            )
+        else:
+            component_log_probs = jnp.stack(
+                [d.log_prob(value) for d in self.component_distributions], axis=-1
+            )
         return jax.nn.log_softmax(self.mixing_distribution.logits) + component_log_probs
 
 
--- a/numpyro/distributions/mixtures.py
+++ b/numpyro/distributions/mixtures.py
@@ -7,6 +7,7 @@ import jax.numpy as jnp
 
 from numpyro.distributions import Distribution, constraints
 from numpyro.distributions.discrete import CategoricalLogits, CategoricalProbs
+from numpyro.distributions.util import validate_sample
 from numpyro.util import is_prng_key
 
 
@@ -144,7 +145,7 @@ class _MixtureBase(Distribution):
     def sample(self, key, sample_shape=()):
         return self.sample_with_intermediates(key=key, sample_shape=sample_shape)[0]
 
-    # @validate_sample
+    @validate_sample
     def log_prob(self, value, intermediates=None):
         del intermediates
         sum_log_probs = self.component_log_probs(value)
@@ -270,6 +271,9 @@ class MixtureGeneral(_MixtureBase):
         ``mixture_size``.
     :param component_distributions: A list of ``mixture_size``
         :class:`~numpyro.distributions.Distribution` objects.
+    :param support: A :class:`~numpyro.distributions.constraints.Constraint`
+        object specifying the support of the mixture distribution. If not
+        provided, the support will be inferred from the component distributions.
 
     **Example**
 
@@ -285,6 +289,20 @@ class MixtureGeneral(_MixtureBase):
        ...     dist.Normal(loc=0.6, scale=1.2),
        ... ]
        >>> mixture = dist.MixtureGeneral(mixing_dist, component_dists)
+       >>> mixture.sample(jax.random.PRNGKey(42)).shape
+       ()
+
+    .. doctest::
+
+       >>> import jax
+       >>> import jax.numpy as jnp
+       >>> import numpyro.distributions as dist
+       >>> mixing_dist = dist.Categorical(probs=jnp.ones(2) / 2.)
+       >>> component_dists = [
+       ...     dist.Normal(loc=0.0, scale=1.0),
+       ...     dist.HalfNormal(scale=0.3),
+       ... ]
+       >>> mixture = dist.MixtureGeneral(mixing_dist, component_dists, support=dist.constraints.real)
        >>> mixture.sample(jax.random.PRNGKey(42)).shape
        ()
     """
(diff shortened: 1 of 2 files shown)


numpyro/distributions/mixtures.py
```diff
@@ -145,7 +144,7 @@ def sample_with_intermediates(self, key, sample_shape=()):
     def sample(self, key, sample_shape=()):
         return self.sample_with_intermediates(key=key, sample_shape=sample_shape)[0]
 
-    @validate_sample
+    # @validate_sample
```
fehiepsi, line 147: I think we need to enable `validate_sample` for all components. One way to achieve this is to enable it in the constructor
```
for d in component_distributions:
  d._validate_args = True
```

numpyro/distributions/mixtures.py
```diff
@@ -320,11 +319,9 @@ def __init__(
         # TODO: It would be good to check that the support of all the component
         # distributions match, but for now we just check the type, since __eq__
         # isn't consistently implemented for all support types.
-        support_type = type(component_distributions[0].support)
-        if any(
-            type(d.support) is not support_type for d in component_distributions[1:]
-        ):
-            raise ValueError("All component distributions must have the same support.")
+        # support_type = type(component_distributions[0].support)
```
fehiepsi, line 322: You can introduce a `support` argument:
```
def __init__(
        self, mixing_distribution, component_distributions, *, support=None, validate_args=None
    ):
```
and we will execute this check when `support=None`.

numpyro/distributions/mixtures.py
fehiepsi: Oops, sorry. You already addressed this in `component_log_probs`.

numpyro/distributions/mixtures.py
```diff
@@ -389,9 +386,21 @@ def component_sample(self, key, sample_shape=()):
...
-            [d.log_prob(value) for d in self.component_distributions], axis=-1
-        )
+        if self._validate_args:
+            mask = jnp.stack(
+                [d.support(value) for d in self.component_distributions], axis=-1
+            )
+            component_log_probs = jnp.where(
+                mask,
+                jnp.stack(
+                    [d.log_prob(value) for d in self.component_distributions], axis=-1
```
fehiepsi, line 396: It is a little bit cheaper to avoid the duplication of the masking logic
```
component_log_probs = []
for d in self.component_distributions:
  log_prob = d.log_prob(value)
  if (self._support is not None) and (not d._validate_args):
    mask = d.support(value)
    log_prob = jnp.where(mask, log_prob, -jnp.inf)
  component_log_probs.append(log_prob)
...
```

fehiepsi (commented):
Thanks for your contribution, @Qazalbash! Could you add a simple test to check whether a mixture of Normal and HalfNormal is computed correctly?

numpyro/distributions/mixtures.py
Qazalbash: I might be missing something, but I think this check is unnecessary. Can you help me understand why we need to check the types again, even after we've already resolved the issue for different support scenarios?

numpyro/distributions/mixtures.py
Qazalbash: This makes more sense.

Qazalbash: > ```python
> if (self._support is not None) and (not d._validate_args):
> ```
I think `not` is not required with `d._validate_args`.

numpyro/distributions/mixtures.py
fehiepsi: If supports are different, we don't have a mechanism to obtain support of the mixture. We currently use the support of the first component, which only works when all supports are the same.

Qazalbash: Seems reasonable, I am adding this change too.

Qazalbash:
@fehiepsi I have made the changes and added the unit tests.

numpyro/distributions/mixtures.py
```diff
@@ -389,9 +397,19 @@ def component_sample(self, key, sample_shape=()):
         return jnp.stack(samples, axis=self.mixture_dim)
 
     def component_log_probs(self, value):
-        component_log_probs = jnp.stack(
-            [d.log_prob(value) for d in self.component_distributions], axis=-1
-        )
+        if self._validate_args:
```
fehiepsi, line 400: This `if/else` is no longer necessary. The if branch is enough.

numpyro/distributions/mixtures.py
```diff
@@ -389,9 +397,19 @@ def component_sample(self, key, sample_shape=()):
...
+            component_log_probs = []
+            for d in self.component_distributions:
+                log_prob = d.log_prob(value)
+                if (self._support is not None) and d._validate_args:
```
fehiepsi, line 404: if `d._validate_args` is True, `log_prob` is already masked so we can skip the masking logic here.

numpyro/distributions/mixtures.py
fehiepsi, line 147: This is required.

numpyro/distributions/mixtures.py
```diff
@@ -320,11 +324,15 @@ def __init__(
         # TODO: It would be good to check that the support of all the component
         # distributions match, but for now we just check the type, since __eq__
         # isn't consistently implemented for all support types.
-        support_type = type(component_distributions[0].support)
-        if any(
-            type(d.support) is not support_type for d in component_distributions[1:]
-        ):
-            raise ValueError("All component distributions must have the same support.")
+        self._support = support
```
fehiepsi, line 327: In the property definition `def support(self):`, could you return `self._support` if it is not None? That is the main purpose of this `support` argument (to make sure that MCMC transform is defined correctly)

test/test_distributions_mixture.py
```diff
@@ -78,6 +85,46 @@ def test_mixture_broadcast_batch_shape(
...
+    mixing_probabilities = jnp.ones(2) / 2
+    mixing_distribution = dist.Categorical(probs=mixing_probabilities)
+    component_distribution = [
+        get_normal(batch_shape, validate_args_for_normal),
+        get_half_normal(batch_shape, validate_args_for_half_normal),
+    ]
+    mixture = dist.MixtureGeneral(
+        mixing_distribution=mixing_distribution,
+        component_distributions=component_distribution,
+        support=True,
```
fehiepsi, line 107: `support` should be `constraints.real`. Maybe adding a check `is_instance(support, constraints.Constraint` in the mixture constructor to make sure that the argument has correct type.

numpyro/distributions/mixtures.py
```diff
@@ -294,7 +293,12 @@ class MixtureGeneral(_MixtureBase):
     pytree_aux_fields = ("_mixture_size",)
 
     def __init__(
-        self, mixing_distribution, component_distributions, *, validate_args=None
+        self,
+        mixing_distribution,
+        component_distributions,
+        *,
+        support=None,
```
fehiepsi, line 300: Could you also add docstring to this argument to illustrate the usage?

test/test_distributions_mixture.py
```diff
@@ -78,6 +85,46 @@ def test_mixture_broadcast_batch_shape(
     _test_mixture(mixing_distribution, component_distribution)
 
 
+@pytest.mark.parametrize("batch_shape", [(), (1,), (7,), (2, 5)])
+@pytest.mark.parametrize("validate_args_for_normal", [False, True])
+@pytest.mark.parametrize("validate_args_for_half_normal", [False, True])
```
fehiepsi, line 90: Setting `validate_args=True` in the components seems unnecessary to me (assume that `support` is provided like below).

numpyro/distributions/mixtures.py
Qazalbash: Make sense.

test/test_distributions_mixture.py
Qazalbash: I have a question which is not relevant to this PR. I've noticed NumPyro doesn't use type annotations. Is this due to historical reasons, Python's dynamic typing, performance considerations, or building on other frameworks? I'm curious to learn more about the team's approach to code quality and maintainability.

numpyro/distributions/mixtures.py
```diff
@@ -288,13 +291,32 @@ class MixtureGeneral(_MixtureBase):
...
+       >>> component_dists = [
+       ...     dist.Normal(loc=0.0, scale=1.0),
+       ...     dist.HalfNormal(scale=0.3),
+       ... ]
+       >>> mixture = dist.MixtureGeneral(mixing_dist, component_dists, support=dist.constraints.real)
+       >>> mixture.sample(jax.random.PRNGKey(42)).shape
+       ()
     """
 
     pytree_data_fields = ("_mixing_distribution", "_component_distributions")
```
fehiepsi, line 310: Could you add `_support` to this tuple?  CI is currently failing.

Qazalbash: Done

Qazalbash:
I used python3.11, let me test everything on python3.9!

Qazalbash:
Apparently, all these tests were passing on python3.11. They are still failing on different `atol` values.

test/test_distributions_mixture.py
```diff
@@ -78,6 +85,41 @@ def test_mixture_broadcast_batch_shape(
...
+    mixture = dist.MixtureGeneral(
+        mixing_distribution=mixing_distribution,
+        component_distributions=component_distribution,
+        support=dist.constraints.real,
+    )
+    assert mixture.batch_shape == batch_shape
+    sample_shape = (11,)
+    xx = component_distribution[0].sample(rng_key, sample_shape)
+    log_prob_0 = component_distribution[0].log_prob(xx)
+    log_prob_1 = component_distribution[1].log_prob(xx)
```
fehiepsi, line 109: This log prob could be not masked. Maybe this causes tests failing?

Qazalbash: This solved the problem.

PR #1793, opened 2024-05-04
add constraints.greater_than_eq, constraints.positive_semidefinite, constraints.nonnegative 

### LINKED ISSUE #1729: [FR] Add constraints.greater_than_eq
Currently, `constraints.greater_than_eq` is supported in [torch](https://pytorch.org/docs/stable/distributions.html#torch.distributions.constraints.greater_than_eq), but we only have `constraints.greater_than` in numpyro. Can we add an alias for `greater_than_eq` with something like `constraints.interval(x, jnp.inf)` to the constraints namespace (and maybe something for positive semidefinite too)?

amifalk:
Fixes #1729

### DIFF
--- a/numpyro/contrib/funsor/infer_util.py
+++ b/numpyro/contrib/funsor/infer_util.py
@@ -1,7 +1,7 @@
 # Copyright Contributors to the Pyro project.
 # SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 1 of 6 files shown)


fehiepsi (approved):
Thanks, @amifalk!

PR #1800, opened 2024-05-14
Add further entropy implementations.

tillahoffmann:
Add entropy for `Independent`, `Expanded`, `Delta`, and `InverseGamma`. Also fixes the `__repr__` method of `SineSkewed`.


fehiepsi (approved):
Thanks, Till!

PR #1801, opened 2024-05-17
example: multidimensional HSGP

brendancooley:
Adds an example to extend the existing HSGP tutorials ([1](https://num.pyro.ai/en/latest/examples/hsgp.html) and [2](https://num.pyro.ai/en/latest/tutorials/hsgp_example.html)) to multidimensional inputs. 

While I was working on this I noticed that @juanitorduz contributed a really nice [module](https://num.pyro.ai/en/latest/contrib.html#hilbert-space-gaussian-processes-approximation) with a bunch of utility functions for HSGPs. Very happy to see these making their way into `numpyro`. I think a natural next step would be to take on Juan's call to adapt his eigenfunctions/values implementations to the multidimensional case. I believe I have a good implementation in the notebook that I have tested for D=1 and D=2 but certainly welcome feedback from him and others in the community (@omarfsosa ?).

Perhaps before merging this I could extend the HSGP contrib module to support multidimensional inputs, and then refactor this notebook and the other HSGP tutorials to import the supporting functionality from the contrib module. Suggestions very welcome.


juanitorduz:
@brendancooley This extension is fantastic! Regarding your suggestion: 

> Perhaps before merging this I could extend the HSGP contrib module to support multidimensional inputs, and then refactor this notebook and the other HSGP tutorials to import the supporting functionality from the contrib module. 

In my opinion, this would be the way to go! Extend the current HSGP module functions to support the multidimensional case. As we have already uni-tests in place, we could add these new cases and make sure we still support the one-dimensional case. After that we could showcase in your example notebook how to use the API.

Really exciting 🙌

brendancooley:
> @brendancooley This extension is fantastic! Regarding your suggestion:
> 
> > Perhaps before merging this I could extend the HSGP contrib module to support multidimensional inputs, and then refactor this notebook and the other HSGP tutorials to import the supporting functionality from the contrib module.
> 
> In my opinion, this would be the way to go! Extend the current HSGP module functions to support the multidimensional case. As we have already uni-tests in place, we could add these new cases and make sure we still support the one-dimensional case. After that we could showcase in your example notebook how to use the API.
> 
> Really exciting 🙌

Awesome, will get moving on this

brendancooley:
Converted to draft, will refactor to reflect changes in #1803 and re-open when ready

brendancooley:
@fehiepsi @juanitorduz ready for review here.

juanitorduz:
@brendancooley This example looks great! I left some small comments and suggestions.

I would suggest a bit more content on the storyline between cells to guide a non-expert reader through the content. Not really about the HSGP approximation but more about the connection between the sections. Just a suggestion :)

brendancooley:
> @brendancooley This example looks great! I left some small comments and suggestions.
> 
> I would suggest a bit more content on the storyline between cells to guide a non-expert reader through the content. Not really about the HSGP approximation but more about the connection between the sections. Just a suggestion :)

Thanks for the prompt feedback! Agree that it ended up a little terse...I'll try and fill in some words between those monstrous matplotlib functions and add some extra comparisons between the models. Will reach back when comments have been addressed.

brendancooley:
<p>Ah to clarify -- this is the HSGP versus the exact (not HSGP versus True). So this is just showing that their posterior predictive means line up for all data points. I will make sure I clarify, easy to see why you would think that.</p>

---

 <a href='https://app.reviewnb.com/pyro-ppl/numpyro/pull/1801/discussion/#comment-2144645092'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

brendancooley:
Added trace plots from arviz for both the exact model and the approximation, and a density plot to compare the approximation to the exact model. Took the time to write better expositions for each of the code blocks. Thanks to @juanitorduz for the prodding and helpful feedback.

@fehiepsi believe this is ready to go.

juanitorduz:
Hey @brendancooley it looks great! I left a comment to see if we can get rid of these divergences.  Besides that, it looks great!

brendancooley:
@juanitorduz bumped the `target_accept_prob` and switched the initialization strategy. 

```
Number of divergences: 0
```
and trace plots look better. 

It runs so fast that experimentation is a breeze!

juanitorduz (approved):
Looks great! Thanks !

brendancooley:
<p>done!</p>

---

 <a href='https://app.reviewnb.com/pyro-ppl/numpyro/pull/1801/discussion/#comment-2156638155'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

fehiepsi (approved):
Great tutorial, thanks @brendancooley!

Also thanks for reviewing, @juanitorduz!

juanitorduz:
(are we waiting for additional reviews 😄 ?)

PR #1807, opened 2024-05-30
gh-1806: Implementation of Doubly Truncated Power Law and Lower Truncated Power Law

### LINKED ISSUE #1806: [FR] Truncated Power Law distribution
# Description
I miss truncated power law distribution in NumPyro alot. I have implemented them on side for my own use. I was thinking if we could add them to NumPyro, it would be great.

Qazalbash:
This PR contains the implementation of Power Law distributions, that fixes #1806. There are two different types of power law implementations based on their truncation.

1. `DoublyTruncatedPowerLaw`
2. `LowerTruncatedPowerLaw`

They are named in such way to avoid confusion with `TwoSidedTruncated*` and `LeftTruncated*` distributions.


Qazalbash:
@fehiepsi Power Law is not in SciPy. Can you guide me for the Unit Tests?

fehiepsi:
Hi @Qazalbash, you could add if/else statements to skip some tests (like some other distributions). Do you have reference for this distribution? I couldn't find it on wikipedia.

Qazalbash:
@fehiepsi unfortunately I couldn't find much on internet too. I have seen many papers in astrophysics using these distributions. Therefore, I did the derivations. If you like to see, I can share them. This way we can cross check it too.

Qazalbash:
`test_sample_gradient` and `test_log_prob_gradient` are failing only in case of `alpha=-1.0`. The gradient is coming out to be `jnp.nan`. I can not understand why, so I have pushed the changes for review.

fehiepsi:
I think you need double where. See [this note](https://github.com/tensorflow/probability/blob/main/discussion/where-nan.pdf)

Qazalbash:
Your suggestion worked! Values are still diverging. I have rechecked a couple of time, my derivations and implementation as well. I think you can double-check it!

InfinityMod:
@Qazalbash Please see my pull request in your repository. Probably that could solve the problems.

Qazalbash:
@InfinityMod, I will look into it.

Qazalbash:
@fehiepsi test failure related to gradients of `DoublyTruncatedPowerLaw` and `LowerTruncatedPowerLaw` are fixed. Please feel free to review it.

fehiepsi:
Sorry, I forgot to change the label. Will review shortly.

test/test_distributions.py
```diff
@@ -51,6 +52,10 @@
 )
 from numpyro.nn import AutoregressiveNN
 
+# Enable 64bit support for higher accuracy
+if sys.maxsize > 2**32:
+    jax.config.update("jax_enable_x64", True)
```
fehiepsi, line 57: Could you revert this change? You can cover the failing tests by adding this line at the top of the test
```python
if jax_dist in [LowerTruncatedPowerLaw, DoublyTruncatedPowerLaw]:
    numpyro.enable_x64()
```

test/test_distributions.py
```diff
@@ -913,6 +918,11 @@ def get_sp_dist(jax_dist):
             ]
         ),  # Covariance
     ),
+    T(dist.LowerTruncatedPowerLaw, -jnp.pi, jnp.array([2.0, 5.0, 10.0, 50.0])),
+    T(dist.DoublyTruncatedPowerLaw, -1.0, 1.0, 2.0),
+    T(dist.DoublyTruncatedPowerLaw, jnp.pi, 5.0, 50.0),
+    T(dist.DoublyTruncatedPowerLaw, -1.0, 5.0, 50.0),
+    T(dist.DoublyTruncatedPowerLaw, jnp.pi, 1.0, 2.0),
```
fehiepsi, line 925: could you use np instead of jnp for those global arrays?

fehiepsi (commented):
Thanks @Qazalbash! I left a few minor comments. I haven't checked the math - but I think the tests will cover it.

test/test_distributions.py
Qazalbash: Resolved in [aba947b](https://github.com/pyro-ppl/numpyro/pull/1807/commits/aba947b30ed46c9d978b05b2ab1f1981e2b0e040).

test/test_distributions.py
Qazalbash: Resolved in 96d5f6c

fehiepsi (approved):

Qazalbash:
@fehiepsi They are still failing!

fehiepsi:
Could you merge with the master branch?

docs/source/distributions.rst
```diff
@@ -404,6 +404,22 @@ ZeroSumNormal
...
+    :show-inheritance:
+    :member-order: bysource
+
+LowerTruncatedPowerLaw
+^^^^^^^^^^^^^^^^^^^^^^
+.. autoclass:: numpyro.distributions.continuous.LowerTruncatedPowerLaw
+    :members:
+    :undoc-members:
+    :show-inheritance:
+    :member-order: bysource
```
fehiepsi, line 421: Could you add those doc entries to the right alphabet order?

Qazalbash: Do you want me to reorder the distribution position by their alphabetical order or do you want `:member-order:` to be alphabetical?

fehiepsi: Actually, I think we should move those distributions to the Truncated section https://num.pyro.ai/en/stable/distributions.html#truncated-distributions and move the implementation to the `distributions/truncated.py` script

fehiepsi: > reorder the distribution position by their alphabetical order

I meant this one

Qazalbash: > Actually, I think we should move those distributions to the Truncated section https://num.pyro.ai/en/stable/distributions.html#truncated-distributions and move the implementation to the `distributions/truncated.py` script

I was doing this in the first place. The reason for not doing so is that I observed all the distributions in that file are truncated versions of different "untruncated" distributions. `DoublyTruncatedPowerLaw` and `LowerTruncatedPowerLaw` are by definition a truncated distribution just like `HalfNormal`. If still like to see them in that file I can move it.

Also at the time of writing three CI actions are running please stop the older two they are wasting energy and resources.

InfinityMod:
Thanks also from my side for pushing the topic further forward!
I was under heavy work load, so just found yet time to look into the topic now again.

docs/source/distributions.rst
fehiepsi: It is better to move to `truncated.py`. The api there is more aligned with the distributions in this PR.

> at the time of writing three CI actions are running please stop the older two

Thanks! For some reasons, the tests for those distributions are very slow. I guess there are some infinite loops are going on? (maybe due to the definition of custom_jvp)

Qazalbash: > I guess there are some infinite loops are going on? (maybe due to the custom jvp rules)

custom jvp rules are implemented by @InfinityMod he can share his insights.

> It is better to move to `truncated.py`. The api there is more aligned with the distributions in this PR.

I will shift them!

Qazalbash:
@InfinityMod The only problem left is the failing of test cases, can you look into it?

InfinityMod:
Hi, please see #1860 to solve the long testing times and the problems significantly when increasing the numerical accuracy to 64-bit. @Qazalbash, @fehiepsi, looking forward to the review and the validation.

Qazalbash:
Could you change the base of #1860 to #1807?

InfinityMod:
I'm not sure if it can still be merged into the master branch if I change the base right now.
Another solution would be to wait until the merge request went through for #1860 and then change for #1807 the code of _utils.py. It should be much effort, as it's only one line to change. Then, changes align with each other.

test/test_distributions.py
```diff
@@ -3370,7 +3428,7 @@ def _assert_not_jax_issue_19885(
 ) -> None:
     # jit-ing identity plus matrix multiplication leads to performance degradation as
     # discussed in https://github.com/google/jax/issues/19885. This assertion verifies
-    # that the issue does not affect perforance in numpyro.
+    # that the issue does not affect perforamce in numpyro.
```
fehiepsi, line 3431: the change also has typo :)

Qazalbash: Fixed in 0589bea.

fehiepsi:
i just merged the other PR. Thanks both!

Qazalbash:
@fehiepsi I have merged the `master` into this branch, please kill the action on 0589bea.

InfinityMod:
Ah, testing went through now. No stopping anymore.
However, some tests failed. But that's really due to the computation accuracy of the implemented (other) functions.
Any suggestions on what to do? I think we can't improve the accuracy for all these tests just because they failed on x64 computational accuracy..

fehiepsi:
for some reasons, float64 is enabled for other distributions, not just power law ones. I will think a bit more on why it is the case.

Qazalbash:
@fehiepsi @InfinityMod Finally all test cases are passing. I fixed them by disabling the higher precession at the start of every test because previously we were enabling it and not disabling it after the test, which let other tests and distributions be tested on x64 precession.

## My thought process!

I tried to fix them in the following fashion,

### Switch between enable and disable (f1da2d54e1e05eab1daf5e5793ef15dac7427c03)

```python
if jax_dist in [dist.LowerTruncatedPowerLaw, dist.DoublyTruncatedPowerLaw]:
    numpyro.enable_x64()
else:
    numpyro.enable_x64(False)
```

### Disable before x64 in tests for power laws (42ed59df20ebd6a0807aa298d2382a4be09cdc33)

```python
numpyro.enable_x64(False)
if jax_dist in [dist.LowerTruncatedPowerLaw, dist.DoublyTruncatedPowerLaw]:
    numpyro.enable_x64()
```

### Why no sandwich?

The next obvious thing was to sandwich tests between enable and disable, like,

```python
if jax_dist in [dist.LowerTruncatedPowerLaw, dist.DoublyTruncatedPowerLaw]:
    numpyro.enable_x64()

# tests

if jax_dist in [dist.LowerTruncatedPowerLaw, dist.DoublyTruncatedPowerLaw]:
    numpyro.enable_x64(False)
```

This is okay if every test comes to the last line. Many tests were using `pytest.skip` which stops it from coming to an end and we have to add an x64 disable statement at those points with a concern that we might have missed any point where we could place it.

### Disable every time (2671edb8b19b86b7ca6396ff92a4223caaeac5c9)

The best way I could figure out was to disable it at the start of every test and it worked!

test/test_distributions.py
```diff
@@ -3415,6 +3689,9 @@ def test_lowrank_mvn_19885(capfd: pytest.CaptureFixture) -> None:
 
 
 def test_gaussian_random_walk_linear_recursive_equivalence():
+    # The reason to add the disable x64 is explained in the github comment, please
+    # follow https://github.com/pyro-ppl/numpyro/pull/1807#issuecomment-2344443900
+    numpyro.enable_x64(False)
```
fehiepsi, line 3694: Sorry for being slow to respond. A better way to handle this issue is to have a separate entry for TruncatedPowerLaw to "test with x64" in the `test-modeling` workflow, like https://github.com/pyro-ppl/numpyro/blob/f5aca91fa19189ffb5e58ac0ff6513afdb6ba216/.github/workflows/ci.yml#L106-L108
```
    - name: Test x64
      run: |
        JAX_ENABLE_X64=1 pytest -vs test/test_distributions.py -k PowerLaw
```
and skip the tests if x32 is detected
```
if jax_dist in [dist.LowerTruncatedPowerLaw, dist.DoublyTruncatedPowerLaw] and jnp.result_type(float) == jnp.float32:
        pytest.skip("Tests for power law distributions need to run under x64 mode.")
```

Qazalbash: Resolved in 34d9f82

fehiepsi:
Glad to merge this awesome work! Thanks for adding `codespell`. 💯

InfinityMod:
Awesome, glad it finally worked out!

Issue #1805, opened 2024-05-22
contrib.hsgp: support vector-valued kernel hyperparameters


brendancooley:
#1803 implements support for multidimensional Hilbert Space Gaussian Process approximations. However, it only supports estimation of a single set of kernel hyperparameters (e.g. squared exponential `lengthscale`). In principal, the lengthscale can vary across dimensions of the input space (see [Riutort-Mayol 2022](https://link.springer.com/article/10.1007/s11222-022-10167-2#auth-Gabriel-Riutort_Mayol-Aff1) eq 1, 2, and 3 for the associated spectral density functions).

Implementation requires updating the `spectral_density_matern` and `spectral_density_squared_exponential` (`numpyro.contrib.hsgp.spectral_densities`) to accept and process array-valued inputs for the `length` parameter (this may also require upstream changes to the `vmap` in `diag_spectral_density_squared_exponential` and `diag_spectral_density_matern`). The test models `test_squared_exponential_gp_model` and `test_matern_gp_model` (`test.contrib.hsgp.test_approximation`) should be updated to optionally sample vector-valued lengthscales and test cases demonstrating the functionality should be created.

brendancooley:
Notes to self as I start working on this...

We might consider first adding some tests that ensure that our kernel approximations come close to matching the exact versions for `m` large enough. Something like the following:
```python
from sklearn.gaussian_process.kernels import RBF

import jax.numpy as jnp

from numpyro.contrib.hsgp.laplacian import eigenfunctions, sqrt_eigenvalues
from numpyro.contrib.hsgp.spectral_densities import (
    diag_spectral_density_squared_exponential,
)

x1 = jnp.array([1.0, 1.0])[None, ...]
x2 = jnp.array([0.0, 0.0])[None, ...]
m = 10
ell = 3
sqrt_eig_v = sqrt_eigenvalues(ell=ell, m=m, dim=2)
eig_f1 = eigenfunctions(x1, ell=ell, m=m)
eig_f2 = eigenfunctions(x2, ell=ell, m=m)
spd = diag_spectral_density_squared_exponential(1.0, 1.0, ell, m, 2)[None, ...]
approx = (eig_f1 * eig_f2 * spd).sum(axis=1)
exact = RBF(1.0)(x1, x2)
assert jnp.isclose(approx, exact)
```

samanklesaria:
[...]

PR #1803, opened 2024-05-21
Support multidimensional HSGP (numpyro.contrib.hsgp)

brendancooley:
- Implements `eigenindices` function, following Eq 10 in Riutort-Mayol et al
- Update upstream functionality to accept a $D \times m^\star$ array of eigenvalues
- Update `eigenfunctions` to handle multidimensional `x`. Vector-valued inputs are treated as unidimensional problems. Otherwise the trailing dimension of `x` is inferred as the dimension of the input space. All preceding dims are treated as batch dims.
- Update `laplacian` and `approximation` functions to accept list-valued inputs for `m` and `ell`. If an int (`m`) or float (`ell`) is passed when the problem is multidimensional, the same value is used for all dimensions.
- Miscellaneous cleanup and extensions of docstrings and methodology overview in "Contributed Code"

The results in @juanitorduz notebook replicate with one change to function arguments. 

TODO (on this PR or future)
* [ ] support vector-valued kernel hyperparameters (alpha, length, etc)
* [ ] Update notebook in #1801 to import utility functions from `numpyro.contrib.hsgp`
* [ ] Update @omarfsosa [birthday problem example](https://num.pyro.ai/en/stable/examples/hsgp.html) to import utility functions from `numpyro.contrib.hsgp`

PR #1808, opened 2024-06-01
fix(contrib.hsgp): fix incorrect dim arg to partial spectral density fns

brendancooley:
As I was working on #1801 I noticed a small (but impactful) bug in my multidim hsgp implementation (#1803). 

When I move over to working on #1805 I'll try and write some tests that will surface this kind of issue in the multidimensional case. But for now this at least gives correct results in my wip example. 

@juanitorduz @fehiepsi


juanitorduz (approved):

fehiepsi (approved):

fehiepsi:
Thanks, @brendancooley!

PR #1811, opened 2024-06-09
fix(contrib.hsgp): convert matern spectral density from frequency domain

brendancooley:
- add tests to compare approximate covariance functions to exact versions from sklearn

As I was implementing these tests (to support development in #1805) I noticed that the Matérn spectral density implementation was not giving results consistent with scikit-learn. After staring at Rasmussen and Williams Chapter 4 for a bit I finally noticed that their spectral densities are in the frequency domain, with the inputs scaled by $2 \pi$. Riutort-Mayol 2023 adopt Rasmussen and Williams' parameterization initially but note that the input vectors are frequencies and convert appropriately for the specific $\nu = \infty$, $\nu =3/2$ and $\nu =5/2$ examples (see screenshots below).

After dropping the $4 \pi^2$ term from the Matern spectral density function I am able to replicate the exact covariance functions from scikit-learn. 

image

image

@juanitorduz let me know if I'm missing anything or misunderstanding here -- still trying to wrap my head around some of this stuff and get the terminology right. Additionally, let me know if you have any suggestions on how to make the test suite stronger. The goal is to add vector-valued alpha and length examples to the tests and use those to guide development on #1805.

### DIFF
--- a/test/contrib/hsgp/test_approximation.py
+++ b/test/contrib/hsgp/test_approximation.py
@@ -7,6 +7,7 @@ from functools import reduce
 from operator import mul
 from typing import Literal
 
+import numpy as np
 import pytest
 from sklearn.gaussian_process.kernels import RBF, ExpSineSquared, Matern
 
@@ -75,11 +76,11 @@ def synthetic_two_dim_data() -> tuple[ArrayImpl, ArrayImpl]:
 @pytest.mark.parametrize(
     argnames="x1, x2, length, ell",
     argvalues=[
-        (jnp.array([[1.0]]), jnp.array([[0.0]]), jnp.array([1.0]), 5.0),
+        (np.array([[1.0]]), np.array([[0.0]]), np.array([1.0]), 5.0),
         (
-            jnp.array([[1.5, 1.25]]),
-            jnp.array([[0.0, 0.0]]),
-            jnp.array([1.0]),
+            np.array([[1.5, 1.25]]),
+            np.array([[0.0, 0.0]]),
+            np.array([1.0]),
             5.0,
         ),
     ],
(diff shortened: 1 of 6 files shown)


juanitorduz (approved):
Wow! Thanks for checking this! It is a great idea! It is easy to get lost in the pi terms 😄

juanitorduz:
@brendancooley, do we also need to fix the formula in the LaTeX docstring?

brendancooley:
> @brendancooley, do we also need to fix the formula in the LaTeX docstring?

Good call! Just pushed fix

juanitorduz (approved):

test/contrib/hsgp/test_approximation.py
```diff
@@ -65,6 +72,130 @@ def synthetic_two_dim_data() -> tuple[ArrayImpl, ArrayImpl]:
     return generate_synthetic_two_dim_data(**kwargs)
 
 
+@pytest.mark.parametrize(
+    argnames="x1, x2, length, ell",
+    argvalues=[
+        (jnp.array([[1.0]]), jnp.array([[0.0]]), jnp.array([1.0]), 5.0),
```
fehiepsi, line 78: Could you change those global arrays into numpy arrays `np.array(...)`? We don't want to trigger any global jax code because users might need to change floating precision or the default device platform.

brendancooley: Got it. We had a bunch of examples of this pattern in the hsgp test suite. Switched them all over and added some handling where needed for numpy arrays in the source code https://github.com/pyro-ppl/numpyro/pull/1811/commits/c9c1457cad9c2e230ab7f149ffa2925d119b3667

juanitorduz: Tests are failing because a type OR operation. I'm not sure about the type hint to use... Maybe we need 

```python
from jaxlib.xla_extension import ArrayImpl
``` 
?

brendancooley: python 3.9...

let me know how this looks to you https://github.com/pyro-ppl/numpyro/pull/1811/commits/56e35f5a521ac9d40810b699b27d1f977c703b91

juanitorduz: It is green ✅🙌

juanitorduz: I'm not a JAX typing expert but this looks very readable 😄

fehiepsi (approved):

PR #1817, opened 2024-06-19
filter out tests waiting for next tfp release

juanitorduz:
Partially addresses https://github.com/pyro-ppl/numpyro/issues/1814 . We must keep in mind removing these `skip` statements once we see a new release.

PR #1819, opened 2024-06-21
Support vector lengthscales for RBF and Matern kernels

samanklesaria:
Resolves pyro-ppl/numpyro#1805

This allows vector lengthscales in the HSGP approximations to RBF and Matern kernels. Extends brendancooley@ef4a24bdd0e741c8a1c1b6228dfa1d890531c2a1

### DIFF
--- a/numpyro/contrib/hsgp/spectral_densities.py
+++ b/numpyro/contrib/hsgp/spectral_densities.py
@@ -17,7 +17,7 @@ from numpyro.contrib.hsgp.laplacian import sqrt_eigenvalues
 
 
 def align_param(dim, param):
-    return jnp.broadcast_arrays(param, jnp.zeros(dim))[0]
+    return jnp.broadcast_to(param, jnp.broadcast_shapes(jnp.shape(param), (dim,)))
 
 
 def spectral_density_squared_exponential(
--- a/numpyro/contrib/hsgp/spectral_densities.py
+++ b/numpyro/contrib/hsgp/spectral_densities.py
@@ -16,8 +16,12 @@ from jax.scipy import special
 from numpyro.contrib.hsgp.laplacian import sqrt_eigenvalues
 
 
+def align_param(dim, param):
+    return jnp.broadcast_arrays(param, jnp.zeros(dim))[0]
+
+
 def spectral_density_squared_exponential(
-    dim: int, w: ArrayImpl, alpha: float, length: float
+    dim: int, w: ArrayImpl, alpha: float, length: float | ArrayImpl
 ) -> float:
     """
     Spectral density of the squared exponential kernel.
@@ -44,13 +48,14 @@ def spectral_density_squared_exponential(
     :return: spectral density value
     :rtype: float
     """
-    c = alpha * (jnp.sqrt(2 * jnp.pi) * length) ** dim
-    e = jnp.exp(-0.5 * (length**2) * jnp.dot(w, w))
+    length = align_param(dim, length)
+    c = alpha * jnp.prod(jnp.sqrt(2 * jnp.pi) * length, axis=-1)
+    e = jnp.exp(-0.5 * jnp.sum(w**2 * length**2, axis=-1))
     return c * e
 
 
 def spectral_density_matern(
-    dim: int, nu: float, w: ArrayImpl, alpha: float, length: float
+    dim: int, nu: float, w: ArrayImpl, alpha: float, length: float | ArrayImpl
 ) -> float:
     """
     Spectral density of the Matérn kernel.
@@ -79,6 +84,7 @@ def spectral_density_matern(
     :return: spectral density value
     :rtype: float
     """  # noqa: E501
+    length = align_param(dim, length)
     c1 = (
         alpha
         * (2 ** (dim))
@@ -86,15 +92,15 @@ def spectral_density_matern(
         * ((2 * nu) ** nu)
         * special.gamma(nu + dim / 2)
     )
-    c2 = (2 * nu / (length**2) + jnp.dot(w, w)) ** (-nu - dim / 2)
-    c3 = special.gamma(nu) * length ** (2 * nu)
+    s = jnp.sum(length**2 * w**2, axis=-1)
+    c2 = jnp.prod(length, axis=-1) * (2 * nu + s) ** (-nu - dim / 2)
+    c3 = special.gamma(nu)
     return c1 * c2 / c3
 
 
-# TODO support length-D kernel hyperparameters
 def diag_spectral_density_squared_exponential(
     alpha: float,
-    length: float,
+    length: float | list[float],
     ell: float | int | list[float | int],
     m: int | list[int],
     dim: int,
--- a/numpyro/contrib/hsgp/util.py
+++ b/numpyro/contrib/hsgp/util.py
@@ -1,0 +1,10 @@
+# Copyright Contributors to the Pyro project.
+# SPDX-License-Identifier: Apache-2.0
+
+from typing import Union
+
+import numpy as np
+
+import jax
+
+ARRAY_TYPE = Union[jax.Array, np.ndarray]  # jax.Array covers tracers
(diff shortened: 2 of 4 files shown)


test/contrib/hsgp/test_approximation.py
```diff
@@ -154,8 +214,17 @@ def test_kernel_approx_squared_matern(
...
-    exact = Matern(length_scale=length, nu=nu)(x1, x2)
-    assert jnp.isclose(approx, exact, rtol=1e-3)
+
+    def _exact_matern(length):
+        return Matern(length_scale=length, nu=nu)(x1, x2).squeeze(axis=-1)
+
+    if isinstance(length, float | int):
+        exact = _exact_matern(length)
+    elif length.ndim == 1:
+        exact = _exact_matern(length)
```
juanitorduz, line 224: why do you need two conditions for `exact = _exact_matern(length)`?

numpyro/contrib/hsgp/util.py
```diff
@@ -0,0 +1,10 @@
+# Copyright Contributors to the Pyro project.
+# SPDX-License-Identifier: Apache-2.0
+
+from typing import Union
+
+import numpy as np
+
+import jax
+
+ARRAY_TYPE = Union[jax.Array, np.ndarray]  # jax.Array covers tracers
```
juanitorduz, line 10: I wonder if this type can be used in other NumPyro modules.

juanitorduz (commented):
Thanks! At first glance, this looks good! But let's see that all tests pass :) @brendancooley do you want to take a look :) ?

numpyro/contrib/hsgp/util.py
brendancooley: Using `ArrayImpl` is only an issue when a model gets compiled and the arrays turn into tracers. `isinstance(X, jax.Array)` will work for both jax arrays and tracers.

numpyro/contrib/hsgp/spectral_densities.py
```diff
@@ -16,8 +16,12 @@
 from numpyro.contrib.hsgp.laplacian import sqrt_eigenvalues
 
 
+def align_param(dim, param):
+    return jnp.broadcast_arrays(param, jnp.zeros(dim))[0]
```
brendancooley, line 20: cleaner solution than mine, looks good

numpyro/contrib/hsgp/util.py
brendancooley: Some details here https://jax.readthedocs.io/en/latest/jax_array_migration.html

I believe this is best practice for typing jax arrays, but I am not sure

brendancooley (commented):
Looks good and works on my offline example. We can see what CI says. Thanks @samanklesaria! Great that we have a group working on this and checking one another's work.

I think next step (maybe another PR) would be to support vector-valued `alpha` for the batch dims. I think it may "just work" with the code as is but it would be helpful to update type annotations, and maybe generalize @juanitorduz's code in `contrib.hsgp.approximation` to sample matrix/tensor-valued `beta`s in `linear_approximation` when batch dimensions are detected.

juanitorduz:
@samanklesaria can you please rebase or sync with the master branch? Today we merged some fixes on the CI, see https://github.com/pyro-ppl/numpyro/pull/1817

test/contrib/hsgp/test_approximation.py
juanitorduz: (this is just a question, no need to use the OR statement in view of readability)

numpyro/contrib/hsgp/util.py
juanitorduz: I am definitively not an expert in type hints, so following the recommendation from the docs seems the safest path :)

juanitorduz:
Yes! The alpha vectorization + docs we can do in another PR :)

test/contrib/hsgp/test_approximation.py
samanklesaria: I was copying the code from `test_kernel_approx_squared_exponential`. Yes, I believe this is in service to readability. If we use an 'or' statement here, we should probably use it there as well.

samanklesaria:
> @samanklesaria can you please rebase or sync with the master branch? Today we merged some fixes on the CI, see #1817

Done!

numpyro/contrib/hsgp/spectral_densities.py
samanklesaria: Good idea! Fixed now.

samanklesaria:
Should sampling matrix/tensor-valued betas in `linear_approximation` be done in a separate PR?

test/contrib/hsgp/test_approximation.py
```diff
@@ -100,13 +127,26 @@ def test_kernel_approx_squared_exponential(
...
     eig_f1 = eigenfunctions(x1, ell=ell, m=m)
     eig_f2 = eigenfunctions(x2, ell=ell, m=m)
     approx = (eig_f1 * eig_f2) @ spd
-    exact = RBF(length)(x1, x2)
-    assert jnp.isclose(approx, exact, rtol=1e-3)
+
+    def _exact_rbf(length):
+        return RBF(length)(x1, x2).squeeze(axis=-1)
+
+    if isinstance(length, float | int):
```
juanitorduz, line 143: The tests are failing becuase of this line

``python
>       if isinstance(length, float | int):
E       TypeError: unsupported operand type(s) for |: 'type' and 'type'
```

Maybe you can use

```python
 isinstance(length, float) |  isinstance(length, int)
```
I am not sure if there is a more elegant way

juanitorduz:
> Should sampling matrix/tensor-valued betas in `linear_approximation` be done in a separate PR?

Personally, I think the scope of this PR is fine. I like working on small iterations so any additional feature can be done in a different PR (at least from my side)

test/contrib/hsgp/test_approximation.py
samanklesaria: Looks like the union shorthand is only for Python 3.10 and higher. We could just use the `Union` constructor explicitly, which should support anything above 3.5.

numpyro/contrib/hsgp/util.py
brendancooley: works for me!

test/contrib/hsgp/test_approximation.py
brendancooley: either way fine by me

juanitorduz:
@samanklesaria it seems there are other more places where you need to change the syntax (similarly as last commit) 😄

samanklesaria:
The current version might have fixed things, but I should probably install a copy of python3.9 locally to test it for sure.

juanitorduz:
There is one test failing because the last change

```python
if isinstance(length, Union[float, int])
```

Union is a type hint so this won't work. I suggest you make the change as I suggested above 😄

test/contrib/hsgp/test_approximation.py
```diff
@@ -100,13 +131,26 @@ def test_kernel_approx_squared_exponential(
...
     eig_f1 = eigenfunctions(x1, ell=ell, m=m)
     eig_f2 = eigenfunctions(x2, ell=ell, m=m)
     approx = (eig_f1 * eig_f2) @ spd
-    exact = RBF(length)(x1, x2)
-    assert jnp.isclose(approx, exact, rtol=1e-3)
+
+    def _exact_rbf(length):
+        return RBF(length)(x1, x2).squeeze(axis=-1)
+
+    if isinstance(length, Union[float, int]):
```
juanitorduz, line 147:
```suggestion
    if isinstance(length,  int) | isinstance(length, float):
```

juanitorduz (approved):
Looks 🟢 ! Great! 🙌

fehiepsi (approved):

fehiepsi:
Thanks for contributing @samanklesaria ! Looking hsgp is having a great momentum.

PR #1821, opened 2024-06-24
Update jax.tree_util.tree_map to jax.tree.map

fehiepsi:
Since jax 0.4.25 (released Feb 24), `jax.tree_util.tree_map` is deprecated, in favor of `jax.tree.map`. This PR updates numpyro to use the new pattern.

In addition, `jnp.clip(..., a_min=..., a_max=...)` is deprecated. I change the pattern to `jnp.clip(..., ..., ...)` to remove the deprecation warning in the tests.


numpyro/distributions/transforms.py
```diff
@@ -1417,7 +1415,7 @@ def _inverse(self, y: jnp.ndarray) -> jnp.ndarray:
         return y
 
     def extend_axis_rev(self, array: jnp.ndarray, axis: int) -> jnp.ndarray:
-        normalized_axis = normalize_axis_tuple(axis, array.ndim)[0]
+        normalized_axis = axis if axis >= 0 else jnp.ndim(array) + axis
```
martinjankowiak, line 1418: what about this change?

martinjankowiak (approved):
lgtm although i might have chosen to keep explicit `min` and `max`

fehiepsi:
The current numpy versions still use a_min a_max. They have a PR to deprecate those arguments but it has not been merged yet.

numpyro/distributions/transforms.py
fehiepsi: This raises a warning that this function is hidden and accessing it through numpy.core.numeric is deprecated.

PR #1822, opened 2024-06-24
Less than eq constaint

Qazalbash:
I noticed in #1729 and #1793, contributor has contributed greater than equals to constraint. I think less than equals to should also be there.


fehiepsi (approved):

fehiepsi:
Thanks, @Qazalbash!

Issue #1446, opened 2022-07-04
Example code for using random_flax_module with flax.linen.BatchNorm (mutable)


nirmalya-broad:
Hi, 

I am trying to use random_flax_module on a class that uses flax.linen.BatchNorm that uses mutable parameters. Is there any example on how to use that? Here is my code:

The model:

```
class NSBlock(nn.Module):

    train: bool
    dim: int
    ks_1: int = 3
    ks_2: int = 3
    dl_1: int = 1
    dl_2: int = 1
    mp_ks: int = 3
    mp_st: int = 1

    @nn.compact
    def __call__(self, x):
        lpad1 = (self.dl_1 * (self.ks_1 - 1)) // 2
        conv_r1 = nn.Conv(features = self.dim,
                kernel_size = (self.ks_1, self.ks_1),
                kernel_dilation = self.dl_1,
                padding= lpad1)
        bn_r1 = nn.BatchNorm(momentum = 0.9, use_running_average = not self.train)

        lpad2 = (self.dl_2 * (self.ks_2 - 1)) // 2
        conv_r2 = nn.Conv(features = self.dim,
                kernel_size = (self.ks_2, self.ks_2),
                kernel_dilation = self.dl_2,
                padding= lpad2)
        bn_r2 = nn.BatchNorm(momentum = 0.9, use_running_average = not self.train)

        x_0 = conv_r1(x)
        x_1 = bn_r1(x_0)
        x_2 = nn.relu(x_1)

        x_3 = conv_r2(x_2)
        x_4 = bn_r2(x_3)

        y = x + x_4
        rpad = (self.mp_ks - 1) // 2
        z = nn.max_pool(y, window_shape = (1, self.mp_ks),
                padding = ((0, 0), (rpad, rpad)), strides = (1, self.mp_st))
        return z


class NeuSomaticNet(nn.Module):

    train : bool
    dim: int = 64
    nsblocks = [
        [3, 5, 1, 1, 3, 1],
        [3, 5, 1, 1, 3, 2],
        [3, 5, 2, 1, 3, 2],
        [3, 5, 4, 2, 3, 2],
    ]

    @nn.compact
    def __call__(self, x_0):

        conv1 = nn.Conv(features =  self.dim, kernel_size = (1, 3),
            padding = (0, 1), strides = 1)
        bn1 = nn.BatchNorm(momentum = 0.9, use_running_average = not self.train)

        x_1 = conv1(x_0)
[...]

PR #1823, opened 2024-06-30
Fixes `random_flax_module` with `flax.linen.BatchNorm`

juanitorduz:
Fixes https://github.com/pyro-ppl/numpyro/issues/1446


test/contrib/test_module.py
```diff
@@ -300,3 +306,8 @@ def model():
         assert tr["nn$state"]["type"] == "mutable"
     else:
         assert set(tr.keys()) == {"nn$params", "x", "y"}
+
+    # test svi
+    guide = AutoDelta(model)
+    svi = SVI(model, guide, numpyro.optim.Adam(0.01), Trace_ELBO())
+    svi.run(random.PRNGKey(100), 10)
```
juanitorduz, line 313: Without these SVI lines, the tests were passing

juanitorduz:
The first two commits are from an old branch 🤦 . We can squash and merge instead.

numpyro/infer/util.py
```diff
@@ -660,7 +660,7 @@ def initialize_model(
         data={
             k: site["value"]
             for k, site in model_trace.items()
-            if site["type"] in ["param"]
+            if site["type"] in ["param", "mutable"]
```
fehiepsi, line 663: maybe also prng_key, so that random keys can be used for dropout

juanitorduz: As in `if site["type"] in ["param", "mutable", "prng_key"]` If I do this and ever the dropout changes `numpyro.prng_key()` - > `random.PRNGKey(0)` I get the error

```python
def apply_without_state(params, *args, **kwargs):
>       return nn_module.apply({"params": params}, *args, **kwargs)
E       ValueError: The ``rngs`` argument passed to an apply function should be a ``jax.PRNGKey`` or a dictionary mapping strings to ``jax.PRNGKey``.
E       --------------------
```

Or am I missing something 😅 ?

fehiepsi: maybe we need to allow substitute rng_key: https://github.com/pyro-ppl/numpyro/blob/5af9ebda72bd7aeb08c61e4248ecd0d982473224/numpyro/handlers.py#L808

juanitorduz: If I add it as 

```python
msg["type"] not in ("sample", "param", "mutable", "plate", "deterministic", "prng_key")
```

The tests fail with 

```python
        if (
            msg["type"] not in ("sample", "param", "mutable", "plate", "deterministic", "prng_key")
        ) or msg.get("_control_flow_done", False):
            if msg["type"] == "control_flow":
                if self.data is not None:
                    msg["kwargs"]["substitute_stack"].append(("substitute", self.data))
                if self.substitute_fn is not None:
                    msg["kwargs"]["substitute_stack"].append(
                        ("substitute", self.substitute_fn)
                    )
            return
    
        if self.data is not None:
>           value = self.data.get(msg["name"])
E           KeyError: 'name'
```

juanitorduz: Or I I remove "mutable" I get the same error

```python
    def apply_without_state(params, *args, **kwargs):
>       return nn_module.apply({"params": params}, *args, **kwargs)
E       ValueError: The ``rngs`` argument passed to an apply function should be a ``jax.PRNGKey`` or a dictionary mapping strings to ``jax.PRNGKey``.
E       --------------------
E       For simplicity, JAX has removed its internal frames from the traceback of the following exception. Set JAX_TRACEBACK_FILTERING=off to include these.
```

fehiepsi: oh, prng_key has no name :( let me think a bit more how to define a static model with PRNGKey in the initialize_model utility.

juanitorduz: Sure! No rush! Is now late in Berlin anyway 😴😄

fehiepsi: Maybe using `substitute_fn` instead of `data` in initialize_model, where we return random.PRNGKey(0) if msg["type"] == "prng_key"

juanitorduz: I tried trying what (I think) you suggested in https://github.com/pyro-ppl/numpyro/pull/1823/commits/a37f7cea9d252ea2cc362a3a9346ab7f90de597a But I keep getting the `KeyError: 'name'`. Please let me know if I missed something or have a tip on how to fix this one :)

numpyro/infer/util.py
```diff
@@ -653,16 +653,17 @@ def initialize_model(
...
-        },
-    )
+    def substitute_fn(site):
+        if site["type"] in ["param", "mutable"]:
+            return site["value"]
+        elif site["type"] == "prng_key":
+            return random.PRNGKey(0)
+
+    model = substitute(model, substitute_fn=substitute_fn)
+
```
fehiepsi, line 666: let's keep the original model as-is and define a new substitute_fn
```
def substitute_key(msg):
    if msg["type"] == "prng_key":
        return random.PRNGKey(0)

if any(site["type"] == "prng_key" for site in model_trace.values()):
    model = substitute(model, substitute_fn=substitute_key)
```
```
```

numpyro/handlers.py
```diff
@@ -805,7 +805,8 @@ def __init__(self, fn=None, data=None, substitute_fn=None):
 
     def process_message(self, msg):
         if (
-            msg["type"] not in ("sample", "param", "mutable", "plate", "deterministic")
+            msg["type"]
+            not in ("sample", "param", "mutable", "plate", "deterministic", "prng_key")
```
fehiepsi, line 809: I guess we need to cover the case no name available 10 lines below
```
value = self.data.get(msg.get("name"))
```

numpyro/infer/util.py
juanitorduz: As in https://github.com/pyro-ppl/numpyro/pull/1823/commits/232c89c08954a19d5578198137f567f3e62081a4? 

When I run the test locally I keep getting ` KeyError: 'name'` coming from 

```python
    def _unconstrain_reparam(params, site):
>       name = site["name"]
E       KeyError: 'name'
E       --------------------
```

I am afraid this will happen everytime we have a `site["name"]` statement.

fehiepsi: I see. It seems that we can't substitute sites without "name" attribute. How about we define a new handler
```
class _substitute_default_key(Messenger):
    def process_msg(self, msg):
        if msg["type"] == "prng_key" and msg["value"] is None:
            msg["value"] = random.PRNGKey(0)

model = _substitute_default_key(model)

juanitorduz: @fehiepsi I think this does the trick indeed! :D

juanitorduz:
https://preview.redd.it/sometimes-progress-looks-like-failure-v0-x4yghusfc1o81.jpg?width=640&crop=smart&auto=webp&s=df5a49d386fca48eaeffaea8945f2060782fcfbc

juanitorduz:
image

juanitorduz:
@fehiepsi should I leave `_substitute_default_key` in `utils.py` or is `handlers.py` a better place?

fehiepsi:
Leaving it in `utils` sounds reasonable to me. It is just a workaround for the edge case.

Thanks for fixing the issue!

fehiepsi (approved):

PR #1829, opened 2024-07-06
Constraints for multiple intervals

Qazalbash:
I have implemented three different constraints that involve multiple closed intervals.

- [x] Union of intervals

$$
x \in \bigcup_{i=1}^{n} [a_i, b_i] \implies \bigvee_{i=1}^{n} (x \in [a_i, b_i])
$$

- [x] Intersection of intervals

$$
x \in \bigcap_{i=1}^{n} [a_i, b_i] \implies \bigwedge_{i=1}^{n} (x \in [a_i, b_i])
$$

- [x] Unique intervals. The idea is we have $n$ different parameters and each has its unique constraining intervals.

$$ 
\bigwedge_{i=1}^{n} (x_i \in [a_i, b_i])
$$

### DIFF
--- a/numpyro/distributions/transforms.py
+++ b/numpyro/distributions/transforms.py
@@ -1639,6 +1639,7 @@ def _transform_to_zero_sum(constraint):
 
 
 @biject_to.register(constraints.union_of_closed_intervals)
+@biject_to.register(constraints.intersection_of_closed_intervals)
 def _transform_to_union_of_intervals(constraint):
     return ComposeTransform(
         [
--- a/numpyro/distributions/transforms.py
+++ b/numpyro/distributions/transforms.py
@@ -1636,3 +1636,16 @@ def _transform_to_simplex(constraint):
 @biject_to.register(constraints.zero_sum)
 def _transform_to_zero_sum(constraint):
     return ZeroSumTransform(constraint.event_dim)
+
+
+@biject_to.register(constraints.union_of_closed_intervals)
+def _transform_to_union_of_intervals(constraint):
+    return ComposeTransform(
+        [
+            ReshapeTransform(
+                forward_shape=constraint.lower_bounds.shape,
+                inverse_shape=(constraint.lower_bounds.size,),
+            ),
+            ExpTransform(),
+        ]
+    )
--- a/numpyro/distributions/transforms.py
+++ b/numpyro/distributions/transforms.py
@@ -1650,3 +1650,15 @@ def _transform_to_union_of_intervals(constraint):
             ExpTransform(),
         ]
     )
+
+
+@biject_to.register(constraints.unique_intervals)
+def _transform_to_unique_intervals(constraint):
+    return ComposeTransform(
+        [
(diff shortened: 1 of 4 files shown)


fehiepsi:
Hi @Qazalbash, do you have any concrete example of those constraints?

numpyro/distributions/transforms.py
```diff
@@ -1636,3 +1636,29 @@ def _transform_to_simplex(constraint):
...
+@biject_to.register(constraints.union_of_closed_intervals)
+@biject_to.register(constraints.intersection_of_closed_intervals)
+def _transform_to_union_of_intervals(constraint):
+    return ComposeTransform(
+        [
+            ReshapeTransform(
+                forward_shape=constraint.lower_bounds.shape,
+                inverse_shape=(constraint.lower_bounds.size,),
+            ),
+            ExpTransform(),
```
fehiepsi, line 1650: I don't understand why this transform is bijective. Could you clarify?

Qazalbash:
For the union of intervals, an [example](https://math.stackexchange.com/questions/1924568/pdf-of-a-piecewise-transformation). I now realize we can avoid the intersection of intervals to a single interval. For the unique interval case, I was writing transformation defined as,

$$
f:(x,y)\to (x',y')
$$

where, $0 < y \leq x$ and,

$$
x'=\frac{(xy)^{3/5}}{(x+y)^{1/5}},\qquad y'=\frac{xy}{(x+y)^{2}}
$$

and turns out the bounds for each are,

$$
0 < x'<\infty \qquad 0 < y' \leq \frac{1}{4}
$$

I had to deal with a similar situation where each element in the `event_dim=1` represents a quantity and has its bounds and I had to write a separate constraint for them.

numpyro/distributions/transforms.py
Qazalbash: I wrote this code snippet to test the constraints of the system. This was the best example I could come up with based on a quick review of the test cases. Please note that this code is not based on any specific logic, so feel free to correct me if I'm wrong or suggest improvements!

fehiepsi: I see. Probably the tests pass because the domain is quite loose. Exp transform has positive codomain, not union of intervals.

fehiepsi:
For the unique interval case, you can use `independent(interval(lower_bound, upper_bound))`. You are right that the intersection case is just another interval. The trickiest one is the union of intervals, which I think you can assume they do not overlap and maps interval i to (i/n,i/n+1/n) where n is the number of intervals. I guess it is better to avoid dealing with such complicated constraint.

Qazalbash:
I got your point. I still don't get `independent(interval(lower_bound, upper_bound))`. Can you explain it using my mentioned use case? After that, I will close the PR!

fehiepsi:
In that example, I guess you can use the constraint
```
constraints.independent(constraints.interval(np.zeros(2), np.array([1e6, 0.25])))
```

Qazalbash:
Thank you!

Issue #1834, opened 2024-07-22
Samples are outside the support for DiscreteUniform distribution


Deathn0t:
Hello, 

I noticed that samples have value outside the support for `DiscreteUniform` distribution. Here is a simple reproducible example:

```python
import jax.random
import numpyro

import numpyro.distributions as dist

from numpyro.infer import HMC, MCMC, MixedHMC


def model():
    x = numpyro.sample("x", dist.DiscreteUniform(1, 2))


num_samples = 10
kernel = HMC(model, trajectory_length=1.2)
kernel = MixedHMC(kernel, num_discrete_updates=20)
mcmc = MCMC(kernel, num_warmup=1000, num_samples=num_samples, progress_bar=False)
key = jax.random.PRNGKey(0)
mcmc.run(key)
samples = mcmc.get_samples()

print(samples)
```

Which outputs:

```
{'x': Array([1, 1, 0, 0, 0, 0, 0, 0, 0, 1], dtype=int32)}
```

Am I using it wrongly or is it a real bug?

Thank you very much for your help.

fehiepsi:
Thanks @Deathn0t! It is a bug at this line https://github.com/pyro-ppl/numpyro/blob/f6eb6ce152bd8e903dd56eeb5909ae0b59e24abe/numpyro/infer/hmc_gibbs.py#L283

We should pass in the enumerate support there. Something like
```
support_size = enumerate_support.shape[0]
proposal_idx = random.randint(rng_proposal, (), minval=0, maxval=support_size)
proposal = enumerate_support[proposal_idx]
```

Deathn0t:
Hi @fehiepsi , thank you for the hint! I will try to look at it today and keep you updated if I am blocked. If I see things working I will open PR.

Deathn0t:
Hi @fehiepsi , I had a look at it today. It seems the Gibbs proposal is used instead of the RW:

https://github.com/pyro-ppl/numpyro/blob/f6eb6ce152bd8e903dd56eeb5909ae0b59e24abe/numpyro/infer/hmc_gibbs.py#L219

PR #1835, opened 2024-07-24
Fixing issue Samples are outside the support for DiscreteUniform dist…

Deathn0t:
This fixes issue #1834 for `MixedHMC` sampling with `DiscreteUniform` distribution sampling outside the support without using the `enumerate_support`.

### DIFF
--- a/numpyro/infer/hmc_gibbs.py
+++ b/numpyro/infer/hmc_gibbs.py
@@ -434,6 +434,13 @@ class DiscreteHMCGibbs(HMCGibbs):
             and site["fn"].has_enumerate_support
             and not site["is_observed"]
         }
+        self._support_enumerates = {
+            name: site["fn"].enumerate_support(False)
+            for name, site in self._prototype_trace.items()
+            if site["type"] == "sample"
+            and site["fn"].has_enumerate_support
+            and not site["is_observed"]
+        }
         self._gibbs_sites = [
             name
             for name, site in self._prototype_trace.items()
--- a/numpyro/infer/mixed_hmc.py
+++ b/numpyro/infer/mixed_hmc.py
@@ -6,6 +6,7 @@ from functools import partial
 
 from jax import grad, jacfwd, lax, random
 from jax.flatten_util import ravel_pytree
+import jax
 import jax.numpy as jnp
 
 from numpyro.infer.hmc import momentum_generator
@@ -301,6 +302,11 @@ class MixedHMC(DiscreteHMCGibbs):
             adapt_state=adapt_state,
         )
 
+        z_discrete = jax.tree.map(
+            lambda idx, support: support[idx],
+            z_discrete,
+            self._support_enumerates,
+        )
         z = {**z_discrete, **hmc_state.z}
         return MixedHMCState(z, hmc_state, rng_key, accept_prob)
 


numpyro/infer/mixed_hmc.py
```diff
@@ -301,6 +302,11 @@ def body_fn(i, vals):
             adapt_state=adapt_state,
         )
 
+        z_discrete = jax.tree.map(
+            lambda idx, support: support[idx],
+            z_discrete,
+            self._support_enumerates,
+        )
```
fehiepsi, line 309: Doing this might return in-support values but I worry that the algorithms are wrong. To compute potential energy correctly in the algorithm, we need to work with in-support values. I think you can pass `support_enumerates` into `self._discrete_proposal_fn` and change the proposal logic there.
```
    proposal = random.randint(rng_proposal, (), minval=0, maxval=support_size)
    # z_new_flat = z_discrete_flat.at[idx].set(proposal)
    z_new_flat = z_discrete_flat.at[idx].set(support_enumerate[proposal])
```
or for modified rw proposal
```
    i = random.randint(rng_proposal, (), minval=0, maxval=support_size - 1)
    # proposal = jnp.where(i >= z_discrete_flat[idx], i + 1, i)
    # proposal = jnp.where(random.bernoulli(rng_stay, stay_prob), idx, proposal)
    proposal_index = jnp.where(support_size[i] == z_discrete_flat[idx], support_size - 1, i)
    proposal = jnp.where(random.bernoulli(rng_stay, stay_prob), idx, support_size[proposal_index])
    z_new_flat = z_discrete_flat.at[idx].set(proposal)
```
or at discrete gibbs proposal
```
    proposal_index = jnp.where(support_enumerate[i] == z_init_flat[idx], support_size - 1, i)
    z_new_flat = z_init_flat.at[idx].set(support_enumerate[proposal_index])
```

Deathn0t: Ok, thank you for the feedback. I will try this.

fehiepsi: I use `print` most of the time. When actual values are needed, I sometimes use `jax.disable_jit()`

Deathn0t: @fehiepsi I have issues with passing enumerate supports and traced values as the support arrays can have different sizes. I was thinking maybe to just pass the "lower bound of the support" as offset and combined with `support_sizes` it should make the trick. Are there discrete variables where the support is not a simple discrete range with step 1 between values?

Deathn0t: for `modified_rw_proposal` I think you used `support_size` in place of `support_enumerate`, shouldn't it be:

```python
    i = random.randint(rng_proposal, (), minval=0, maxval=support_size - 1)
    # proposal = jnp.where(i >= z_discrete_flat[idx], i + 1, i)
    # proposal = jnp.where(random.bernoulli(rng_stay, stay_prob), idx, proposal)
    proposal_index = jnp.where(support_enumerate[i] == z_discrete_flat[idx], support_size - 1, i)
    proposal = jnp.where(random.bernoulli(rng_stay, stay_prob), idx, support_enumerate[proposal_index])
    z_new_flat = z_discrete_flat.at[idx].set(proposal)
```

numpyro/infer/hmc_gibbs.py
```diff
@@ -434,6 +470,32 @@ def init(self, rng_key, num_warmup, init_params, model_args, model_kwargs):
...
+                and site["fn"].has_enumerate_support
+                and not site["is_observed"]
+            )
+        )
+        # All support_enumerates should have the same length to be used in the loop
+        # Each support is padded with zeros to have the same length
+        self._support_enumerates = np.zeros(
+            (len(self._support_sizes), max_length_support_enumerates), dtype=int
+        )
+        for i, (name, site) in enumerate(self._prototype_trace.items()):
```
fehiepsi, line 487: great solution! I just have a couple of comments:
+ it might be better to loop over names in support_sizes and get site via `site = self._prototype_trace[name]`
+ we use `ravel_pytree` to flatten support_sizes. so we might want to keep the same behavior here. I don't have a great solution for this, maybe
```python
support_enumerates = {}
for name, support_size in self._support_sizes:
    site = self._prototype_trace[name]
    enumerate_support = site["fn"].enumerate_support(False)
    padded_enumerate_support = np.pad(enumerate_support, (0, max_length_support_enumerates - enumerate_support.shape[0]))
    padded_enumerate_support = np.broadcast_to(padded_enumerate_support, support_size.shape + (max_length_support_enumerates,))
    support_enumerates[name] = padded_enumerate_support

self._support_enumerates = jax.vmap(lambda x: ravel_pytree(x)[0], in_axes=1, out_axes=1)(support_enumerates)
```

numpyro/infer/mixed_hmc.py
fehiepsi: thanks! your solutions are super cool! I haven't thought of different support sizes previously.

Deathn0t:
@fehiepsi it worked fine with `ravel_pytree` as well, I just had to adapt the `in_axes=0`.

fehiepsi:
I think we need to ravel along the first axis. The second axis (corresponds to `max_length_support_enumerates`) is the batch dimension. The current code might run but I guess things are mixed up.

numpyro/infer/hmc_gibbs.py
```diff
@@ -434,6 +471,39 @@ def init(self, rng_key, num_warmup, init_params, model_args, model_kwargs):
...
+        # Each support is padded with zeros to have the same length
+        # ravel is used to maintain a consistant behaviour with `support_sizes`
+
+        max_length_support_enumerates = max(
+            (
+                site["fn"].enumerate_support(False).shape[0]
+                for site in self._prototype_trace.values()
+                if site["type"] == "sample"
+                and site["fn"].has_enumerate_support
+                and not site["is_observed"]
```
fehiepsi, line 485: nit: it is better to loop over `support_sizes`: `for name, site in self._prototype_trace.items() if name in support_sizes`

Deathn0t:
> I think we need to ravel along the first axis. The second axis (corresponds to `max_length_support_enumerates`) is the batch dimension. The current code might run but I guess things are mixed up.

the first axis is `in_axes=0`?

fehiepsi:
we vmap over the batch axis, which is the second axis.

fehiepsi:
Could you also add a simple test (as in the issue) for this? you can run `make lint` and `make format` to fix lint issues.

Deathn0t:
I applied the lint/format and I added a test.

> we vmap over the batch axis, which is the second axis, i.e. in_axes=1

ok, but the `support_size` values have shape `()`.

So the following line:

```python
support_size.shape + (max_length_support_enumerates,),
```

is just equivalent to `(max_length_support_enumerates,)`, and therefore `in_axis=1` fails. What format should the `enumerate_size` and `enumerate_support` have?

fehiepsi:
That is a good point. I thought support sizes contain flatten arrays. Sorry for the confusion. I guess we need to move the enumerate dimension to the first axis before vmapping like you did
```
support_enumerates[name] = np.move_axes(padded_enumerate_support, -1, 0)
```

Deathn0t:
I tried the following direction:

```python
        max_length_support_enumerates = np.max(
            [size for size in self._support_sizes.values()]
        )

        support_enumerates = {}
        for name, support_size in self._support_sizes.items():
            site = self._prototype_trace[name]
            enumerate_support = site["fn"].enumerate_support(True).T
            # Only the last dimension that corresponds to support size is padded
            pad_width = [(0, 0) for _ in range(len(enumerate_support.shape) - 1)] + [
                (0, max_length_support_enumerates - enumerate_support.shape[-1])
            ]
            padded_enumerate_support = np.pad(enumerate_support, pad_width)

            support_enumerates[name] = padded_enumerate_support

        self._support_enumerates = jax.vmap(
            lambda x: ravel_pytree(x)[0], in_axes=len(support_size.shape), out_axes=1
        )(support_enumerates)
```

which work with the following cases:

```python
def model_1():
    numpyro.sample("x0", dist.DiscreteUniform(10, 12))
    numpyro.sample("x1", dist.Categorical(np.asarray([0.25, 0.25, 0.25, 0.25])))

def model_2():
    numpyro.sample("x0", dist.Categorical(0.25 * jnp.ones((4,))))
    numpyro.sample("x1", dist.Categorical(0.1 * jnp.ones((10,))))

def model_3():
    numpyro.sample("x0", dist.Categorical(0.25 * jnp.ones((3, 4))))
    numpyro.sample("x1", dist.Categorical(0.1 * jnp.ones((3, 10))))
```

But fails when I try to batch `DiscreteUniform`:


```python
def model_4():
    numpyro.sample("x1", dist.DiscreteUniform(10 * jnp.ones((3,)), 19 * jnp.ones((3,))))
```

fehiepsi:
The `in_axes=len(support_size.shape)` might not be the same across different latent variables. I think you can move the batch dimension to the front like in my last comment.

By the way, maybe we need to use `size.reshape(-1)[0]` instead of size
```
max_length_support_enumerates = np.max(
            [size for size in self._support_sizes.values()]
        )
```

fehiepsi:
Hmm, there seems to have a bug at DiscreteUniform.enumerate_support. `self.low` should be `jnp.reshape(self.low, -1)[0]`

fehiepsi:
@Deathn0t The fix is in #1859. Could you test whether the change works now?

Deathn0t:
@fehiepsi sorry for the delay... other things happened I couldn't follow up. Yes, let me test this now!

Deathn0t:
@fehiepsi the 4 cases I put in the test are now passing, assuming changes from #1859 are used!

PR #1837, opened 2024-07-25
Simplify the provenance logic to prepare for the removal of jax named shape

fehiepsi:
See https://github.com/google/jax/pull/21069 as a context. The current tests fail with jax's dev branch.

Current tests at `test/test_model_rendering` and `test/ops/test_provenance` pass locally.


ordabayevy (approved):
LGTM!

Just to clarify for myself: These code changes don't use any new functionality, right? I just wonder why did we use `named_shape` in the first place.

fehiepsi:
Yes, this just simplifies the logic. Previously we used named shape because we need some functionality to flatten unflatten the provenance. Turns out that those jax tree utilities also work for frozenset leaves.

PR #1841, opened 2024-08-03
[Draft] Enable pmap progress bar with cpu backend / remove deprecated host_callback

andrewdipper:
Switched from `host_callback.id_tap` to `io_callback`.  This allows the progressbar to work when the jax backend is `cpu`.  In addition `host_callback.id_tap` is deprecated.   However `io_callback` doesn't allow identification of the device in the callback.  

As a result the progressbars are not tied to the physical devices.  The resulting effect is the progressbars behave the same but they are always shown in sorted order (by most to least complete) with no possible identification of the underlying device.

Left as draft due to the behavioral changes / needs more visual testing.

The original error with progressbar + pmap on cpu:
`NotImplementedError: host_callback functionality isn't supported with PJRT C API. See https://jax.readthedocs.io/en/latest/debugging/index.html and https://jax.readthedocs.io/en/latest/notebooks/external_callbacks.html for alternatives. Please file a feature request at https://github.com/google/jax/issues if none of the alternatives are sufficient.`


andrewdipper:
There might be some race issues here - I'm not exactly sure what the `io_callback` guarantees are

numpyro/util.py
```diff
@@ -201,24 +202,40 @@ def progress_bar_factory(num_samples, num_chains):
 
     remainder = num_samples % print_rate
 
+    idx_map = {}
```
fehiepsi, line 205: could you add some comments here for how this works?

numpyro/util.py
```diff
@@ -201,24 +202,40 @@ def progress_bar_factory(num_samples, num_chains):
...
     tqdm_bars = {}
     finished_chains = []
+    lock = Lock()
```
fehiepsi, line 208: could you add a comment for why this is needed?

numpyro/util.py
```diff
@@ -231,26 +248,20 @@ def _update_progress_bar(iter_num):
 
         _ = lax.cond(
             iter_num == 1,
-            lambda _: host_callback.id_tap(
-                _update_tqdm, 0, result=iter_num, tap_with_device=True
-            ),
-            lambda _: iter_num,
+            lambda _: io_callback(_update_tqdm, None, -1, 0),
```
fehiepsi, line 251: i'm not sure why -1, 0 is used here

numpyro/util.py
andrewdipper: added by _`calc_chain_idx`

numpyro/util.py
andrewdipper: done

numpyro/util.py
andrewdipper: The -1 was a mistake, changed to 0,0.   The second 0 is consistent with the current implementation.  I believe this is 0 instead on 1 so that subsequent calls using print_rate and remainder add to the total.  But it does seem to be off by 1

andrewdipper:
Maybe hold on reviewing this - there's probably a better way that keeps a fixed mapping between progress bars and chains.  In discussion here https://github.com/blackjax-devs/blackjax/pull/712

andrewdipper:
This is updated - a resource counter assigns each chain a progress bar.  This id is then saved in the fori loop carry.  

The issue is the function returned by `progress_bar_fori_loop` expects the loop carry to have a spot to save the chain id.  So if there's any use of `progress_bar_factory` externally this would break it.  I'm not sure what'd be the best way to transition it smoothly

numpyro/util.py
```diff
@@ -201,58 +202,64 @@ def progress_bar_factory(num_samples, num_chains):
...
+        with lock:
+            idx = idx_counter
+            idx_counter += 1
+        return idx
+
+    def _update_tqdm(iter_num, increment, chain):
+        iter_num = int(iter_num)
+        increment = int(increment)
+        chain = int(chain)
+        if iter_num == 0:
```
fehiepsi, line 225: because iter_num is not used, it is clearer to use the assertion `chain == -1`

numpyro/util.py
```diff
@@ -378,10 +386,15 @@ def loop_fn(collection):
...
             return fori_loop(
-                0, upper, _body_fn_pbar, (init_val, collection, start_idx, thinning)
+                0,
+                upper,
+                _body_fn_pbar,
+                ((init_val, collection, start_idx, thinning), -1),  # -1 for chain id
             )
 
-        last_val, collection, _, _ = maybe_jit(loop_fn, donate_argnums=0)(collection)
+        (last_val, collection, _, _), _ = maybe_jit(loop_fn, donate_argnums=0)(
```
fehiepsi, line 395: no need to change this if we dont return chain in loop_fn

numpyro/util.py
```diff
@@ -201,58 +202,64 @@ def progress_bar_factory(num_samples, num_chains):
...
+            lambda _: chain,
             operand=None,
         )
-        _ = lax.cond(
+        chain = lax.cond(
             iter_num % print_rate == 0,
-            lambda _: host_callback.id_tap(
-                _update_tqdm, print_rate, result=iter_num, tap_with_device=True
+            lambda _: io_callback(
+                _update_tqdm, jnp.array(0), iter_num, print_rate, chain
```
fehiepsi, line 251: seems like no need for `iter_num`

fehiepsi:
> So if there's any use of progress_bar_factory externally this would break it

Do you know when this happens?

numpyro/util.py
fehiepsi: Could you clarify the purpose of Lock in more details? I'm not familiar with its usage. What happens if we dont use it?

andrewdipper: Updated.  The locking is only around `idx_counter` since after a chain has a chain id there isn't access to resources that are shared across threads.

numpyro/util.py
andrewdipper: agreed - done

numpyro/util.py
```diff
@@ -201,58 +202,64 @@ def progress_bar_factory(num_samples, num_chains):
...
+            idx = idx_counter
+            idx_counter += 1
+        return idx
+
+    def _update_tqdm(iter_num, increment, chain):
+        iter_num = int(iter_num)
+        increment = int(increment)
+        chain = int(chain)
+        if iter_num == 0:
+            chain = _calc_chain_idx(iter_num)
```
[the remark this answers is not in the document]
andrewdipper: done

numpyro/util.py
andrewdipper: done

numpyro/util.py
andrewdipper: ah good point

andrewdipper:
> Do you know when this happens?

I'm not sure - I'd suspect that the vast majority of times `progress_bar_factory` is simply used through calls to `fori_collect`.  If it is used externally I'd guess it'd only be in specialized cases - but I don't have an example in mind.

fehiepsi (approved):

fehiepsi:
LGTM. CI seems to suggest that there is no degration in performance. Thanks for the great solution, @andrewdipper!

Issue #1836, opened 2024-07-24
Crash when using `TruncatedNormal` in `parallel` MCMC, but not in `sequential` MCMC


damonbayer:
I have encountered a very strange issue while using `TruncatedNormal`. With `numpyro.set_host_device_count(2)` commented out, this script runs fine. Uncommenting it causes this script to crash.

Everything works fine with `my_distribution = dist.Normal()`, so it appears to be related to the `TruncatedNormal`.

```python
import jax.random as random
import numpyro
import numpyro.distributions as dist

from numpyro.infer import MCMC, NUTS

numpyro.set_host_device_count(2)

my_distribution = dist.TruncatedNormal(low=1, high=2)
n_data_samples = 1
n_mcmc_samples = 1
n_mcmc_warmup = 0
n_mcmc_chains = 2
rng_key = random.PRNGKey(0)


data_samples = my_distribution.sample(rng_key, (n_data_samples,))


def my_model():
    numpyro.sample("obs", my_distribution, obs=data_samples)


nuts_kernel = NUTS(my_model)
mcmc = MCMC(
    nuts_kernel,
    num_warmup=n_mcmc_warmup,
    num_samples=n_mcmc_samples,
    num_chains=n_mcmc_chains,
)
mcmc.run(rng_key)
```

fehiepsi:
It is better to move the definition to the model. Some distributions cache info during running, which might cause tracer leakage.
```
def my_model():
    my_distribution = dist.TruncatedNormal(low=1, high=2)
    numpyro.sample("obs", my_distribution, obs=data_samples)
```

damonbayer:
Moving the definition to the model does work around the issue.

Is this considered a bug? It seems to only affect `TruncatedNormal` when both `low` and `high` are specified. If only one of them is specified, the model runs fine. This observation, combined with the fact that it only affects parallel sampling, makes me think that this could be fixed.

fehiepsi:
We use `lazy_property` here and there https://github.com/pyro-ppl/numpyro/blob/master/numpyro/distributions/truncated.py#L215

Maybe a general fix is to not turn those properties into attributes at https://github.com/pyro-ppl/numpyro/blob/master/numpyro/distributions/util.py#L695 if it is a tracer:`numpyro.util.not_jax_tracer(value)`. Do you want to submit that enhancement?

damonbayer:
[...]

Issue #1651, opened 2023-09-26
Leaked trace with truncated distribution stored in dictionary outside model scope


alexlyttle:
**Update:** This appears to not be specific to rendering a model. It also occurs when doing MCMC (see Example 2).

I got an `UnexpectedTracerError` when using a `TruncatedNormal` distribution and `numpyro.render_model`. This appears to only happen when the `model` function samples from a truncated distribution stored in a `dict` defined outside of the function scope. After the error occurs, it repeats whenever the distribution is sampled. A temporary fix is to either not use `numpyro.render_model` at all, or to not store the distribution in a `dict` outside of the model function. I appreciate this is a bit of an unusual bug that may not affect many people, but I thought I had better raise the issue. Also, I can do without `numpyro.render_model` for now.

## Example 1

Rendering a model.

```python
import numpyro
from numpyro.distributions import TruncatedNormal
from numpyro.handlers import trace, seed
from jax import random, checking_leaks

prior = {"x": TruncatedNormal(low=0)}  # <-- error only occurs when low or high are defined

def model():
    numpyro.sample("x", prior["x"])

with checking_leaks():
    numpyro.render_model(model)
```

<details>
<summary>Show exception and traceback</summary>

```python
---------------------------------------------------------------------------
Exception                                 Traceback (most recent call last)
Cell In[3], line 7
      4     numpyro.sample("x", prior["x"])
      6 with checking_leaks():
----> 7     numpyro.render_model(model)

File ~/miniconda3/envs/numpyro/lib/python3.10/site-packages/numpyro/infer/inspect.py:622, in render_model(model, model_args, model_kwargs, filename, render_distributions, render_params)
    599 def render_model(
    600     model,
    601     model_args=None,
   (...)
    605     render_params=False,
    606 ):
    607     """
    608     Wrap all functions needed to automatically render a model.
    609 
   (...)
    620     :param bool render_params: Whether to show params in the plot.
    621     """
--> 622     relations = get_model_relations(
    623         model,
    624         model_args=model_args,
    625         model_kwargs=model_kwargs,
    626     )
    627     graph_spec = generate_graph_specification(relations, render_params=render_params)
    628     graph = render_graph(graph_spec, render_distributions=render_distributions)

File ~/miniconda3/envs/numpyro/lib/python3.10/site-packages/numpyro/infer/inspect.py:322, in get_model_relations(model, model_args, model_kwargs)
    319     return PytreeTrace(trace)
[...]

PR #1843, opened 2024-08-06
avoids setting jax tracer as lazy property attribute 

sbidari:
Add a conditional if `numpyro.util.not_jax_tracer(value)` before setting `lazy_property` as an attribute of `TruncatedDistribution`

This resolves errors described in #1836 and https://github.com/CDCgov/multisignal-epi-inference/issues/282

tests added:
check valid sampling of `TruncatedDistribution` in parallel (failed previously as described in #1836 )
check predictive methods (prior predictive and inference) can be run multiple times on the same model built using `TruncatedDistribution` (error encountered here https://github.com/CDCgov/multisignal-epi-inference/issues/282)

### DIFF
--- a/test/test_distributions.py
+++ b/test/test_distributions.py
@@ -3423,4 +3423,3 @@ def test_gaussian_random_walk_linear_recursive_equivalence():
     x2 = dist2.sample(random.PRNGKey(7))
     assert jnp.allclose(x1, x2.squeeze())
     assert jnp.allclose(dist1.log_prob(x1), dist2.log_prob(x2))
-
--- a/test/test_distributions.py
+++ b/test/test_distributions.py
@@ -23,6 +23,7 @@ import jax.random as random
 from jax.scipy.special import expit, logsumexp
 from jax.scipy.stats import norm as jax_norm, truncnorm as jax_truncnorm
 
+import numpyro
 import numpyro.distributions as dist
 from numpyro.distributions import (
     SineBivariateVonMises,
@@ -49,6 +50,7 @@ from numpyro.distributions.util import (
     sum_rightmost,
     vec_to_tril_matrix,
 )
+from numpyro.infer import MCMC, NUTS, Predictive
 from numpyro.nn import AutoregressiveNN
 
 TEST_FAILURE_RATE = 2e-5  # For all goodness-of-fit tests.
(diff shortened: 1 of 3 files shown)


test/test_distributions.py
```diff
@@ -3423,3 +3425,65 @@ def test_gaussian_random_walk_linear_recursive_equivalence():
...
+        dist.TruncatedNormal(low=-1, high=1),
+        dist.TruncatedCauchy(low=-1, high=1),
+    ],
+)
+def test_repeated_predictive_method_in_truncated_distribution(my_dist):
+    n_data_samples = 10
+    n_prior_samples = 10
+    n_mcmc_samples = 10
+    n_mcmc_warmup = 10
+    n_mcmc_chains = 2
```
dylanhmorris, line 3471:
```suggestion
    n_prior_samples = 10
```

test/test_distributions.py
```diff
@@ -3423,3 +3425,65 @@ def test_gaussian_random_walk_linear_recursive_equivalence():
...
+    rng_key = jax.random.PRNGKey(0)
+
+    data_samples = my_dist.sample(rng_key, (n_data_samples,))
```
dylanhmorris, line 3474:
```suggestion
```

test/test_distributions.py
```diff
@@ -3423,3 +3425,65 @@ def test_gaussian_random_walk_linear_recursive_equivalence():
...
+    prior_predictive(rng_key)
+
+    nuts_kernel = NUTS(my_model)
+    mcmc = MCMC(
+        nuts_kernel,
+        num_warmup=n_mcmc_warmup,
+        num_samples=n_mcmc_samples,
+        num_chains=n_mcmc_chains,
+    )
+    mcmc.run(rng_key, obs=data_samples)
```
dylanhmorris, line 3489:
```suggestion
```

test/test_distributions.py
```diff
@@ -3423,3 +3425,65 @@ def test_gaussian_random_walk_linear_recursive_equivalence():
...
+    n_mcmc_chains = 2
+    rng_key = jax.random.PRNGKey(0)
+
+    data_samples = my_dist.sample(rng_key, (n_data_samples,))
+
+    def my_model(obs=None):
+        numpyro.sample("obs", my_dist, obs=obs)
+
+    prior_predictive = Predictive(my_model, num_samples=n_prior_samples)
+    prior_predictive(rng_key)
```
dylanhmorris, line 3480: Is one call to `prior_predictive()` sufficient to reproduce the error without the patch, or are two calls needed, as the reproducible example from https://github.com/CDCgov/multisignal-epi-inference/issues/282?

dylanhmorris (commented):
Thanks for this, @sbidari. I wonder whether it might be worth annotating the test functions (via their docstrings) to explain why they should run without error if things are working and why/how they fail prior without the patch in this PR

test/test_distributions.py
sbidari: Thanks @dylanhmorris. The second call is initiated when running the inference. 

But your comments made me realize I don't need two separate tests; the second test can check if the chains run in parallel and two calls to predictive method (prior predictive and inference run here)

fehiepsi (approved):
LGTM pending reviews from @dylanhmorris. Thanks @sbidari!

test/test_distributions.py
sbidari: closing this as resolved in https://github.com/pyro-ppl/numpyro/pull/1843/commits/82dbc935323bb352e1fc8c1d2e58cabaef523108

sbidari:
Thanks @fehiepsi!

@dylanhmorris I added docstring with links to where issues are described and combined the two test functions.

test/test_distributions.py
```diff
@@ -3423,3 +3426,42 @@ def test_gaussian_random_walk_linear_recursive_equivalence():
...
+
+
+@pytest.mark.parametrize(
+    "my_dist",
+    [
+        dist.TruncatedNormal(low=-1, high=1),
+        dist.TruncatedCauchy(low=-1, high=1),
+    ],
+)
+def test_tracer_leakage_in_truncated_distribution(my_dist):
```
dylanhmorris, line 3438:
```suggestion
def test_no_tracer_leakage_in_truncated_distribution(my_dist):
```

dylanhmorris (approved):
Thanks, @sbidari. One suggestion you can take or leave on naming the test (since it should pass if there _isn't_ tracer leakage). @fehiepsi: this LGTM.

sbidari:
I am not sure why the test is failing in CI with `UserWarning: There are not enough devices to run parallel chains: expected 2 but got 1.` despite using `numpyro.set_host_device_count(2)` at the beginning. Any thoughts @fehiepsi?

test/test_distributions.py
```diff
@@ -49,8 +50,10 @@
     sum_rightmost,
     vec_to_tril_matrix,
 )
+from numpyro.infer import MCMC, NUTS, Predictive
```
fehiepsi, line 53: It is better to avoid adding inference tests in this file. Maybe simple tests like are enough?
```
def test_jitted_lazy_property():
    d = dist.TruncatedNormal(...)
    jax.vmap(d.log_prob)(jnp.ones(3))
```

dylanhmorris: Somewhat related: since the key change is to the `distributions.util` module, I might suggest putting this test in 
`test_distributions_util.py` rather than `test_distributions.py`. Do you agree, @fehiepsi?

dylanhmorris: I like this idea, but at least for me the specific suggested test passes even with the existing `lazy_property` implementation (I have confirmed that the MRE from #1836 does indeed error).

dylanhmorris:
Noting here that when this lands it will also provide a more general fix for #1651

test/test_distributions.py
fehiepsi: maybe calling vmap 2 times? it is surprised to me that jax does not catch this

fehiepsi: re distributiomz_util: that works :)

dylanhmorris: Have proposed a pair of tests in https://github.com/sbidari/numpyro/pull/1 and confirmed they fail with the latest `numpyro` on PyPI but pass with @sbidari's patch.

sbidari: Thanks @dylanhmorris for rewriting the tests! Ready for review @fehiepsi

fehiepsi (approved):
Thanks @sbidari!

damonbayer:
@fehiepsi Is there a release planned for this fix? We would appreciate having it for [cdcgov/multisignal-epi-inference](https://github.com/cdcgov/multisignal-epi-inference/) as soon as possible.

fehiepsi:
I can make a patch release in upcoming days. In the mean time, you can patch the utility.

damonbayer:
@fehiepsi Do you have a date in mind?

PR #1849, opened 2024-08-12
enable progress bar for multi-gpu

andrewdipper:
The switch from using `host_callback` to `io_callback` for progress bars means we can enable multi-gpu progress bars.  Missed this change in https://github.com/pyro-ppl/numpyro/pull/1841.  Verified on kaggle multi-gpu platform.


fehiepsi (approved):
Thanks, @andrewdipper!

Issue #1845, opened 2024-08-07
Predictive: better docs for setting obs to `None`


tomwallis:
Hello,

It wasn't clear to me until I dug into some strange behaviour that the `Predictive` method requires setting `obs` to `None`. I see now that the examples do it, but it's not made explicit that this is required for the class to work as intended. Could this be made more explicit in the [docs](https://num.pyro.ai/en/latest/utilities.html#predictive)? Thanks in advance!

fehiepsi:
You are right. It is not trivial to realize that api. Do you want to enhance the docs?

tomwallis:
I can try. To do this, should I build the docs locally, or are the changes to the source files sufficient?

tillahoffmann:
Expanding the documentation in the source file linked below should do the trick. You don't need to build the documentation locally, but it can be helpful for checking that the output looks like what you intended.

https://github.com/pyro-ppl/numpyro/blob/d61f15c0d05a7a86ca464248e5c6f8814b56e457/numpyro/infer/util.py#L851-L918

tomwallis:
Thanks @tillahoffmann. I have created a PR that hopefully does the trick. I was not able to build the docs locally without more time investment, so I have not checked the intended output.

PR #1850, opened 2024-08-12
Update docs to make Predictive behavior more clear

tomwallis:
This PR addresses [this issue](https://github.com/pyro-ppl/numpyro/issues/1845#issuecomment-2284569263), improving the docs for how the `Predictive` class is expected to function. In particular, I try to make it more clear that the observations should be set to `None` to explicitly make prior or posterior predictions. 

I updated the docstring for the class itself, and also two of the example `ipynb`s.

In the docstring files there are also some `black` import re-orderings.

### DIFF
--- a/numpyro/infer/util.py
+++ b/numpyro/infer/util.py
@@ -1,18 +1,19 @@
 # Copyright Contributors to the Pyro project.
 # SPDX-License-Identifier: Apache-2.0
 
-import warnings
 from collections import namedtuple
 from contextlib import contextmanager
 from functools import partial
 from typing import Callable, Optional
+import warnings
+
+import numpy as np
 
 import jax
-import numpy as np
-import jax.numpy as jnp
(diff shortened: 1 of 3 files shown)


fehiepsi (approved):
Thanks, @tomwallis!

numpyro/infer/util.py
```diff
@@ -1,19 +1,18 @@
...
 from contextlib import contextmanager
 from functools import partial
 from typing import Callable, Optional
-import warnings
-
-import numpy as np
 
 import jax
+import jax.numpy as jnp
+import numpy as np
```
fehiepsi, line 12: Could you revert the order of those imports? This seems to make CI failing.

fehiepsi:
@tomwallis We are going to make a new patch release soon. Do you want to incorporate this change in the release?

tomwallis:
Sorry @fehiepsi I was on holiday then too busy. Hopefully this fixes the issue.

PR #1853, opened 2024-08-21
Generic `cdf` and `icdf` implementation for scalar `TransformedDistribution`s.

tillahoffmann:
This PR

- Adds a `sign` attribute to `Transform` which is the sign of the derivative for bijective scalar transforms. This matches the `sign` attribute of transforms in Pyro.
- Implements `cdf` for `TransformedDistribution` by transforming from the domain of the transformed distribution to the domain of the base distribution and evaluating the cdf.
- Implements `icdf` for `TransformedDistribution` by calling `icdf` of the base distribution and applying the transforms.

This means cdfs for `LogNormal`, `InverseGamma`, `LogUniform`, and `Pareto` do not require a custom implementation. As a side effect, this PR adds `icdf` support for `LogNormal`, `InverseGamma`, and `LogUniform`; it also adds `cdf` and `icdf` support for `RelaxedBernoulliLogits`.

### DIFF
--- a/numpyro/distributions/continuous.py
+++ b/numpyro/distributions/continuous.py
@@ -698,9 +698,6 @@ class InverseGamma(TransformedDistribution):
         a = (self.rate / (self.concentration - 1)) ** 2 / (self.concentration - 2)
         return jnp.where(self.concentration <= 2, jnp.inf, a)
 
-    def cdf(self, x):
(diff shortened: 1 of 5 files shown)


fehiepsi (approved):
Beautiful! Thanks, Till!

PR #1860, opened 2024-09-09
Fix binominal distribution

InfinityMod:
Hi, Pull Request #1807 ends in an infinite loop while running the test setup for the distributions, causing the tests to end until the time limit is reached. 

The misbehavior can be tested just by running the test_distributions setup or only especially the 
**test_vmapped_binominal_p0** function solely with the following code included directly after the imports in **test_distributions.py**:
```
# Enable 64bit support for higher accuracy
if sys.maxsize > 2**32:
    jax.config.update("jax_enable_x64", True)
```

While I can't 100% say why this endless loop only occurs when increasing the numeric accuracy to 64 bits, I'm sure to have a suitable solution for the behavior (which also seems faulty at 32 bits).

The problem stems from the **_binomial_dispatch** function in **util.py** (see the description commit). In short, zero values for the probability p are also passed to the dispatch function, even if filtered out afterward via _lax.cond_.  Therefore, the underlying while_loop in the **_binomial_inversion** function runs infinitely.

This merge request solves this issue by ensuring that no p values equal to zero are passed to the underlying functions. As _lax.cond_ is still filtering out the results of the zero-corrected values, there's no change for instances using the mentioned methods.

### DIFF
--- a/numpyro/distributions/util.py
+++ b/numpyro/distributions/util.py
@@ -136,6 +136,7 @@ def _binomial_inversion(key, p, n):
         i, key, geom_acc = val
         key, key_u = random.split(key)
         u = random.uniform(key_u)
+        #FIXME: we run here against -inf as log1_p equals zero log(0+1) for p = 0
         geom = jnp.floor(jnp.log1p(-u) / log1_p) + 1
         geom_acc = geom_acc + geom
         return i + 1, key, geom_acc
@@ -146,6 +147,7 @@ def _binomial_inversion(key, p, n):
         # this cond_exclude_large_mu is unnecessary for correctness but will
         # still improve performance.
         cond_exclude_large_mu = p * n < _binomial_mu_thresh
+        #FIXME: for p equals 0 the while loop will never end, as -inf is not catched
         return cond_exclude_large_mu & (geom_acc <= n)
 
     log1_p = jnp.log1p(-p)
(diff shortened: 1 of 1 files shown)


numpyro/distributions/util.py
```diff
@@ -156,6 +156,8 @@ def _binom_inv_cond_fn(val):
 def _binomial_dispatch(key, p, n):
     def dispatch(key, p, n):
         is_le_mid = p <= 0.5
+        #Make sure p=0 is never taken into account as a fix for possible zeros in p.
```
fehiepsi, line 159: could we just simply clip p to tiny? `jnp.clip(p, minval=jnp.finfo(p.dtype).tiny)`

InfinityMod: This works for me: ```jnp.clip(p, jnp.finfo(jnp.float32).tiny)```

fehiepsi: this might affect samples with small `p` in x64. How about updating `_binomial_inversion` to
```
geom = jnp.floor(jnp.log1p(-u) / log1_p) + 1
...
log1_p = jnp.log1p(-p)
log1_p = jnp.where(log1_p == 0, -jnp.finfo(log1_p.dtype).tiny, log1_p)
```
The issue seems to come from `log1_p=0` and `jnp.log1p(-u) < 0`, which leads to a negative `geom`

fehiepsi: Could you also fix the lint issue? I think you need a space `# comment` for the comment.

fehiepsi (approved):
Thanks @InfinityMod! This issue is subtle.

PR #1866, opened 2024-09-24
Refactor `validate_args` method on distributions (fixes #1865).

### LINKED ISSUE #1865: Refactor argument validation?
What are your thoughts on refactoring the argument validation code below into a separate method `validate_args()` (or another name) so it can be invoked on an existing instance? The motivation is that I often have jitted functions that return distribution instances which I would like to validate.

https://github.com/pyro-ppl/numpyro/blob/94f4b99710d855bea456210cf91e6e55eeac3926/numpyro/distributions/distribution.py#L231-L246

tillahoffmann:
This PR moves argument validation into a separate method. I've added an argument `strict` to the method such that an error is raised if validation is requested inside jitted code. The constructor uses `strict = False` to maintain backwards compatibility.

### DIFF
--- a/numpyro/distributions/distribution.py
+++ b/numpyro/distributions/distribution.py
@@ -229,22 +229,33 @@ class Distribution(metaclass=DistributionMeta):
         if validate_args is not None:
             self._validate_args = validate_args
         if self._validate_args:
-            for param, constraint in self.arg_constraints.items():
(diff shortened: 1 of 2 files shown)


fehiepsi (approved):
Thanks, @tillahoffmann!

PR #1871, opened 2024-09-30
Fixing nested sampling

renecotyfanboy:
There are some import discrepancies between `jaxns<2.6` and `jaxns 2.6.*`. This small fix seems to do the trick!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -118,6 +118,9 @@ jobs:
     - name: Test custom prng
       run: |
         JAX_ENABLE_CUSTOM_PRNG=1 pytest -vs test/infer/test_mcmc.py
+    - name: Test nested sampling
+      run: |
+        JAX_ENABLE_X64=1 pytest -vs test/contrib/test_nested_sampling.py
 
 
   examples:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -104,8 +104,8 @@ jobs:
     - name: Test with pytest
       run: |
         pytest -vs --durations=20 test/infer/test_mcmc.py
-        pytest -vs --durations=20 test/infer --ignore=test/infer/test_mcmc.py
-        pytest -vs --durations=20 test/contrib --ignore=test/contrib/stochastic_support/test_dcc.py
+        pytest -vs --durations=20 test/infer --ignore=test/infer/test_mcmc.py --ignore=test/contrib/test_nested_sampling.py
+        pytest -vs --durations=20 test/contrib --ignore=test/contrib/stochastic_support/test_dcc.py --ignore=test/contrib/test_nested_sampling.py
     - name: Test x64
       run: |
         JAX_ENABLE_X64=1 pytest -vs test/infer/test_mcmc.py -k x64
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -105,7 +105,7 @@ jobs:
       run: |
         pytest -vs --durations=20 test/infer/test_mcmc.py
         pytest -vs --durations=20 test/infer --ignore=test/infer/test_mcmc.py --ignore=test/contrib/test_nested_sampling.py
-        pytest -vs --durations=20 test/contrib --ignore=test/contrib/stochastic_support/test_dcc.py --ignore=test/contrib/test_nested_sampling.py
+        pytest -vs --durations=20 test/contrib --ignore=test/contrib/stochastic_support/test_dcc.py
     - name: Test x64
       run: |
         JAX_ENABLE_X64=1 pytest -vs test/infer/test_mcmc.py -k x64
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -4,7 +4,7 @@ funsor
 ipython
 jax
 jaxlib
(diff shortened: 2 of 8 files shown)


fehiepsi:
could you update jaxns version in setup and docs/requirements https://github.com/pyro-ppl/numpyro/blob/8e9313fd64a34162bc1c08b20ed310373e82e347/setup.py#L64

docs/requirements.txt
```diff
@@ -4,7 +4,7 @@ funsor
 ipython
 jax
 jaxlib
-jaxns==2.4.8
+jaxns>=2.6.2
```
fehiepsi, line 7: could you pin the version instead? i dont think the version number reflects the stability of the library. We have many breakages in the past.

renecotyfanboy: I squashed everything in a new commit

renecotyfanboy:
@fehiepsi any idea on how to enforce double precision in the test base for this specific test ? I don't know how JAX and pytest work together and if it is possible to enable x64 for a single test

fehiepsi:
You can add a flag like this one https://github.com/pyro-ppl/numpyro/blob/8e9313fd64a34162bc1c08b20ed310373e82e347/.github/workflows/ci.yml#L111

renecotyfanboy:
Okay I hope it works now, I had a lot of trouble since latest versions of `jaxns` enforces double precision to summarize, I had to : 
- Makes a special case for pytest if double precision is not enabled to skip NS related tests 
- Add an extra invoke for dedicated `jaxns` tests
- Change the expected result of a doctest in affine transform in the documentation, as the output is jnp.float64 instead of float32. All the other tests ran smoothly on my computer 
- Updated a bit the `jaxns` contributed code. To be fair, at some point I should take some time to update it properly with the new interfaces proposed by the package

.github/workflows/ci.yml
```diff
@@ -118,6 +118,9 @@ jobs:
     - name: Test custom prng
       run: |
         JAX_ENABLE_CUSTOM_PRNG=1 pytest -vs test/infer/test_mcmc.py
+    - name: Test nested sampling
+      run: |
+        JAX_ENABLE_X64=1 pytest -vs test/contrib/test_nested_sampling.py
```
fehiepsi, line 123: you can ignore nested sampling at line 1078

renecotyfanboy: I am not sure which line you are talking about here. I made sure that the `jaxns` related tests are skipped if double precision is not enabled.

renecotyfanboy:
Okay @fehiepsi this should work now. For you to know, most of the struggles came from the fact that importing `jaxns` automatically enable double precision. So top-level imports in pytests triggered the double precision even if not required. I don't import it if the double precision is disabled, and the related test suite is automatically skipped in this situation.

examples/gaussian_shells.py
```diff
@@ -18,6 +18,8 @@
        https://github.com/joshspeagle/dynesty/blob/master/demos/Examples%20--%20Gaussian%20Shells.ipynb
 """
 
+# ruff: noqa: E402
```
fehiepsi, line 21: nit: maybe remove?

numpyro/distributions/transforms.py
```diff
@@ -714,9 +714,9 @@ class LowerCholeskyAffine(Transform):
 
        >>> import jax.numpy as jnp
        >>> from numpyro.distributions.transforms import LowerCholeskyAffine
-       >>> base = jnp.ones(2)
-       >>> loc = jnp.zeros(2)
-       >>> scale_tril = jnp.array([[0.3, 0.0], [1.0, 0.5]])
+       >>> base = jnp.ones(2, dtype=jnp.float32)
+       >>> loc = jnp.zeros(2, dtype=jnp.float32)
+       >>> scale_tril = jnp.array([[0.3, 0.0], [1.0, 0.5]], dtype=jnp.float32)
```
fehiepsi, line 719: I think it is better to skip doctest nestedsampling as follows
```diff
diff --git a/docs/source/conf.py b/docs/source/conf.py
index 4025960..74dea96 100644
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -34,6 +34,8 @@ from numpyro.infer.hmc import hmc  # noqa: E402

 hmc(None, None)

+autodoc_mock_imports = ["jaxns"]
+
 # -- Project information -----------------------------------------------------

 project = "NumPyro"
diff --git a/numpyro/contrib/nested_sampling.py b/numpyro/contrib/nested_sampling.py
index 324fbdd..be42f75 100644
--- a/numpyro/contrib/nested_sampling.py
+++ b/numpyro/contrib/nested_sampling.py
@@ -145,9 +145,7 @@ class NestedSampler:
     :param dict termination_kwargs: keyword arguments to terminate the sampler. Please
         refer to the upstream :meth:`jaxns.NestedSampler.__call__` method.

-    **Example**
-
-    .. doctest::
+    Example::

         >>> from jax import random
         >>> import jax.numpy as jnp
```

renecotyfanboy:
I think it's ready for review

fehiepsi (approved):
Woohoo, thanks @renecotyfanboy!

PR #1874, opened 2024-10-02
gh-1870: Refactor `log_prob` method in `_MixtureBase` class to handle `-jnp.inf`

### LINKED ISSUE #1870: Grads w.r.t. weights of `MixtureGeneral` Distribution are giving `nan`s
Hi,

We have created some models where we estimate the weights of the `MixtureGeneral` distribution. However, when computing the gradient of this argument, we are encountering `nan` values. We enabled `jax.config.update("debug_nan", True)` to diagnose the issue, and it pointed to the following line:

https://github.com/pyro-ppl/numpyro/blob/8e9313fd64a34162bc1c08b20ed310373e82e347/numpyro/distributions/mixtures.py#L152

I suspect that after the implementation of https://github.com/pyro-ppl/numpyro/pull/1791, extra care is needed to handle `inf` and `nan` values, possibly by using a double `where` for a safe `logsumexp`.

> [!IMPORTANT]
> This is an urgent issue, so a prompt response would be greatly appreciated.

Qazalbash:
closes #1870

### DIFF
--- a/numpyro/distributions/mixtures.py
+++ b/numpyro/distributions/mixtures.py
@@ -149,7 +149,10 @@ class _MixtureBase(Distribution):
     def log_prob(self, value, intermediates=None):
         del intermediates
         sum_log_probs = self.component_log_probs(value)
-        return jax.nn.logsumexp(sum_log_probs, axis=-1)
+        safe_sum_log_probs = jnp.where(
+            jnp.isneginf(sum_log_probs), -jnp.inf, sum_log_probs
+        )
+        return jax.nn.logsumexp(safe_sum_log_probs, axis=-1)
 
 
 class MixtureSameFamily(_MixtureBase):


numpyro/distributions/mixtures.py
```diff
@@ -149,7 +149,10 @@ def sample(self, key, sample_shape=()):
     def log_prob(self, value, intermediates=None):
         del intermediates
         sum_log_probs = self.component_log_probs(value)
-        return jax.nn.logsumexp(sum_log_probs, axis=-1)
+        safe_sum_log_probs = jnp.where(
+            jnp.isneginf(sum_log_probs), -jnp.inf, sum_log_probs
+        )
```
fehiepsi, line 154: Sounds like a grad bug of logsumexp to me.

fehiepsi (approved):

fehiepsi:
Thanks @Qazalbash!

PR #1877, opened 2024-10-05
use biased autocorrelation estimate by default (resolves #1785)

### LINKED ISSUE #1785: Use biased autocorrelation estimate by default
I was noticing some very erratic and unexpected behaviour from the `effective_sample_size` diagnostic, which was due to some extreme values in the far right tail of the `autocorrelation` function.  This can be reproduced pretty easily by plotting the autocorrelation of a sequence of 1000 IID Gaussians.

This behaviour seems to be due to the following line https://github.com/pyro-ppl/numpyro/blob/2f1bccdba2fc7b0a6ec235ca1bd5ce2417a0635c/numpyro/diagnostics.py#L130 which I believe is there to make the estimate unbiased.  Stan however uses the biased estimate of Geyer (1992) (see https://github.com/stan-dev/stan/blob/634034deb3abd6314d980c1aab083f64269f4019/src/stan/analyze/mcmc/autocovariance.hpp#L60), presumably to stop the issue of the variance in the right tail exploding.

I have local changes which (optionally) use the biased estimate. I think this should be the default, partly to be consistent with Stan, but also because it seems too erratic to be useful with the current behaviour.  It seems to be an intentional departure from the Stan implementation however, so I thought best to open an issue here to discuss.

jonny-so:
This changes the autocorrelation function to use the same (biased) estimate as Stan by default. The unbiased estimate is known to have "wild" tails, with variance that is O(1) at longer lags regardless of the length of the chain. See Priestley (1981, Section 5.3) for a discussion of the properties of these estimators. Note that the old (unbiased) estimator is only strictly unbiased when the mean is known, but I have followed Priestley (1981) in referring to it as unbiased.

I have added two checks to ``test_autocorrelation`` that highlight the difference in the tail behaviour.

### DIFF
--- a/numpyro/diagnostics.py
+++ b/numpyro/diagnostics.py
@@ -96,12 +96,13 @@ def _fft_next_fast_len(target):
         target += 1
 
 
-def autocorrelation(x, axis=0):
+def autocorrelation(x, axis=0, unbiased=False):
     """
     Computes the autocorrelation of samples at dimension ``axis``.
 
     :param numpy.ndarray x: the input array.
     :param int axis: the dimension to calculate autocorrelation.
+    :param unbiased: use an unbiased estimator of the autocorrelation.
     :return: autocorrelation of ``x``.
     :rtype: numpy.ndarray
     """
@@ -127,7 +128,13 @@ def autocorrelation(x, axis=0):
 
     # truncate and normalize the result, then transpose back to original shape
     autocorr = autocorr[..., :N]
-    autocorr = autocorr / np.arange(N, 0.0, -1)
+
+    # the unbiased estimator is known to have "wild" tails, due to few samples at longer lags.
+    # see Geyer (1992) and Priestley (1981) for a discussion. also note that it is only strictly
+    # unbiased when the mean is known, whereas we it estimate from samples here.
+    if unbiased:
+        autocorr = autocorr / np.arange(N, 0.0, -1)
+
     with np.errstate(invalid="ignore", divide="ignore"):
         autocorr = autocorr / autocorr[..., :1]
     return np.swapaxes(autocorr, axis, -1)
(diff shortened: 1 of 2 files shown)


numpyro/diagnostics.py
```diff
@@ -96,12 +96,13 @@ def _fft_next_fast_len(target):
         target += 1
 
 
-def autocorrelation(x, axis=0):
+def autocorrelation(x, axis=0, unbiased=False):
```
fehiepsi, line 99: maybe change unbiased to bias and set bias=True by default.

numpyro/diagnostics.py
```diff
@@ -96,12 +96,13 @@ def _fft_next_fast_len(target):
...
     """
     Computes the autocorrelation of samples at dimension ``axis``.
 
     :param numpy.ndarray x: the input array.
     :param int axis: the dimension to calculate autocorrelation.
+    :param unbiased: use an unbiased estimator of the autocorrelation.
```
fehiepsi, line 105: whether to use a biased estimator

fehiepsi (commented):
Do you know how this change will affect ESS of the current users' experiments?

jonny-so:
> Do you know how this change will affect ESS of the current users' experiments?

It will change it for sure, my experience with the current ESS was that it was too erratic to even be useful, which is why I started looking into this in the first place. If it is a concern we could always make `bias=False` the default, perhaps with a warning in the comments. My view is that given it is diagnostic code it is relatively low risk to change the default, but I'm happy to go with either.

jonny-so:
Actually, the biased/unbiasedness refers to the autocovariance, not the autocorrelation. Maybe the parameter names need a rethink.

fehiepsi:
I think we should go with the biased version by default if the unbiased one is not practical.

fehiepsi (approved):
Thanks, @jonny-so!

PR #1880, opened 2024-10-06
feat: add api compat arg to NS.get_samples

nstarman:
The standard API, which tools like `arviz` rely on, requires the argument `group_by_chain`.
See https://github.com/pyro-ppl/numpyro/blob/e4622925d4ceb2f1cf7436f57ecb8dcf2a6ddd96/numpyro/infer/mcmc.py#L702.
This PR adds the kwarg as a no-op.


fehiepsi:
if group_by_chain is true, I guess we need to add singleton dimension in front. Otherwise we can raise error if it is true.

nstarman:
I like the grouping by chain suggestion. Arviz wants that for the shape.

fehiepsi (approved):

nstarman:
The CI error is unrelated and blocking the actual unit tests. Happy to rebase when a sphinx-fixing PR is merged.

juanitorduz:
> The CI error is unrelated and blocking the actual unit tests. Happy to rebase when a sphinx-fixing PR is merged.

Pushing a fix in https://github.com/pyro-ppl/numpyro/pull/1882

fehiepsi:
The fix is merged thanks to @juanitorduz!

nstarman:
👍. Rebased.

nstarman:
@fehiepsi, is this good to go in?

PR #1882, opened 2024-10-07
remove depreciated line from docs

juanitorduz:
Trying to solve the error in https://github.com/pyro-ppl/numpyro/actions/runs/11218235719/job/31185189614?pr=1880

```python
phinx.errors.SphinxWarning: Calling get_html_theme_path is deprecated. If you are calling it to define html_theme_path, you are safe to remove that code.
```

So that is what I did (works locally)

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -7,7 +7,6 @@ import shutil
 import sys
 
 import nbsphinx
-import sphinx_rtd_theme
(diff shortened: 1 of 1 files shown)


fehiepsi (approved):

fehiepsi:
Thanks @juanitorduz !

PR #1883, opened 2024-10-08
Refactoring SteinVI

OlaRonning:
SVGD has been extended to a few varieties ([ASVGD](https://arxiv.org/abs/2101.09815), [GSVGD](https://arxiv.org/abs/2202.03297), [HSVGD](https://arxiv.org/abs/2106.10760)) which I want to include in the einstein module. This PR consists of the changes to `SteinVI` to allow for the extensions and introduction of SVGD and ASVGD. *GSVGD and HSVGD will be in subsequent PRs*.

To this end, I've:
 1. Added a specialized constructor for SVGD. 
    - The SVGD constructor does not require a guide because it's always a delta 
    - It correctly sets the scaling on the attractive force to `1/m` for `m` particles.
    - It doesn't allow users to change `num_elbo_particles` because it's always 1 for SVGD.
 2. Added a `setup_run` method to SteinVI.
    - The method encapsulates a step of SteinVI, which inheriting constructors can manipulate.
 3. Added a constructor for ASVGD.
    -  ASVGD introduces an annealing schedule on the attractive force.
    - This inherits from SVGD and overwrites `setup_run` to change the `loss_temperature`.
 3. Removed [Jacobian projection from the Stein force](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/contrib/einstein/steinvi.py#L337-L339)
    - The projection is unnecessary as we do not allow `AutoIAFNormal`, `AutoBNAFNormal`, `AutoDAIS`, `AutoSemiDAIS` and `AutoSurrogateLikelihoodDAIS`.
    - This simplifies the force computation to attractive+repulsive.

### Misc changes
1. The normalization factor in the `ProbabilityProductKernel` has been removed. The kernel is still a proper kernel; however, this version avoids vanishing/exploding when the guide variances deviate from 1 for "high" dimensional models.
2. Added a rng_key to the kernel. This is convenient when experimenting with kernels.
3. Removed the `enum` parameter as it is currently unsupported.
4. Kernel smoothing of the attractive force and the repulsion is now computed on particles in unconstraint space, consistent with particles moving in unconstraint space. **NB: Constraint particles are removed from the test**.
5. Added the manual computation for kernel tests as comments.
    - I inadvertently changed the value of one of the particles when doing the manual computations, which is why the evaluation has changed.

### TODO
- [x] Add docstrings to ASVGD and SVGD.
- [x] Add manual computation to the IMQKernel and GraphicalKernel tests.
- [x] ~~Fix SteinVI documentation~~ I'll do this in a separate PR.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -7,7 +7,6 @@ import shutil
 import sys
 
 import nbsphinx
-import sphinx_rtd_theme
 
 # import pkg_resources
 
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -226,8 +226,8 @@ html_favicon = "_static/img/favicon/favicon.ico"
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
 #
-html_theme = "sphinx_rtd_theme"
-html_theme_path = [sphinx_rtd_theme.get_html_theme_path()]
+# html_theme = "sphinx_rtd_theme"  # <--- FIXME: gives deprecated warning!
+# html_theme_path = [sphinx_rtd_theme.get_html_theme_path()]
 
 # Theme options are theme-specific and customize the look and feel of a theme
 # further.  For a list of options available for each theme, see the
--- a/numpyro/contrib/einstein/mixture_guide_predictive.py
+++ b/numpyro/contrib/einstein/mixture_guide_predictive.py
@@ -6,8 +6,7 @@ from functools import partial
 from typing import Optional
 import warnings
 
-from jax import numpy as jnp, random, vmap
-from jax.tree_util import tree_flatten, tree_map
+from jax import numpy as jnp, random, tree, vmap
 
 from numpyro.handlers import substitute
 from numpyro.infer import Predictive
@@ -77,7 +76,7 @@ class MixtureGuidePredictive:
 
         self.guide = guide
         self.return_sites = return_sites
-        self.num_mixture_components = jnp.shape(tree_flatten(params)[0][0])[0]
+        self.num_mixture_components = jnp.shape(tree.flatten(params)[0][0])[0]
         self.mixture_assignment_sitename = mixture_assignment_sitename
 
     def _call_with_params(self, rng_key, params, args, kwargs):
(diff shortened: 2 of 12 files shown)


docs/source/conf.py
```diff
@@ -226,8 +226,8 @@
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
 #
-html_theme = "sphinx_rtd_theme"
-html_theme_path = [sphinx_rtd_theme.get_html_theme_path()]
+# html_theme = "sphinx_rtd_theme"  # <--- FIXME: gives deprecated warning!
```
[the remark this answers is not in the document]
fehiepsi: it is fixed upstream. Could you rebase?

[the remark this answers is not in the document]
OlaRonning: Excellent, thanks :+1:

OlaRonning:
@fehiepsi Let me know if you want me to split it into multiple PRs.

numpyro/contrib/einstein/mixture_guide_predictive.py
```diff
@@ -4,21 +4,27 @@
 from collections.abc import Callable, Sequence
 from functools import partial
 from typing import Optional
+import warnings
 
-import jax
 from jax import numpy as jnp, random, vmap
+from jax.tree_util import tree_flatten, tree_map
```
fehiepsi, line 10: This is deprecated. You can use `jax.tree.flatten` and `jax.tree.map` now.

numpyro/contrib/einstein/steinvi.py
```diff
@@ -2,30 +2,25 @@
...
 import functools
 from functools import partial
 from itertools import chain
 import operator
 
-import jax
-from jax import grad, jacfwd, numpy as jnp, random, vmap
+from jax import grad, numpy as jnp, random, vmap
 from jax.flatten_util import ravel_pytree
+from jax.tree_util import tree_map
```
fehiepsi, line 13: same as above

test/contrib/einstein/test_steinvi_util.py
```diff
@@ -8,8 +8,8 @@
 import pytest
 import scipy
 
-import jax
 from jax import numpy as jnp
+from jax.tree_util import tree_flatten, tree_map
```
fehiepsi, line 12: same as above

fehiepsi (approved):
LGTM pending minor comments.

PR #1888, opened 2024-10-12
build: refactor build infrastructure

nstarman:
This PR refactors the build infrastructure to
1. remove the setup.py in favor of pyproject.toml
2. dynamically generate the version from the pyproject (this is not a VCS dynamic versioning)
3. be compatible with `uv` but still all other tools, like `conda`, 
4. use `src` folder layout.


fehiepsi:
Hi @nstarman, thanks for the refactor. We would prefer to keep the current setup. I'm not sure why we need to be compatible with uv (shouldn't uv thing be compatible with the python ecosystem?) and why we need to use the src layout.

nstarman:
`uv` is compatible with the python ecosystem. It just adds config to `pyproject.toml` for development and adds a lockfile, which is useful.
Is there a reason for keeping the outdated `setup.py`?
When I've been doing dev in my last PR my editable build is producing a `numpyro.egg-info` in the top-level env. The `src` layout is better for isolating dev environments. Because of `git-mv` current PRs can rebase without consequence.

fehiepsi:
re setup.py: https://packaging.python.org/en/latest/discussions/setup-py-deprecated/#setup-py-deprecated

nstarman:
> re setup.py: https://packaging.python.org/en/latest/discussions/setup-py-deprecated/#setup-py-deprecated

Yes, it's not deprecated. I was careful to use "outdated" as new standards have superseded `setup.py`, offering standardization, better integrations with developer tooling, safety from arbitrary code execution, improved reproducibility in builds, support for PEP 517/518, and longevity (version 21.3 of pip has started moving away from supporting editable installs via setup.py)

fehiepsi:
Thanks. Could you make a new PR just for setup.py changes?

fehiepsi:
Close, in favor of https://github.com/pyro-ppl/numpyro/pull/1891

PR #1891, opened 2024-10-16
build: switch to pyproject.toml

nstarman:


nstarman:
The version is now specified 2x, in `numpyro.version.py::__version__` and in `pyproject.toml`.
I could fix that in this PR or in a followup. Don't know exactly what it will take, possibly `importlib.metadata` will work. Or maybe `setuptools_scm` dynamic version writing.

nstarman:
Needed to incorporate changes from recent Jax NS fixes PR.

fehiepsi (commented):
Maybe it's good to use setup.py for version handling?

nstarman:
There are many pyproject.toml based solutions, it's just about choosing one. Do you want to manually bump the version or automatically using version control? Or use tooling like commitizen or GH actions?

fehiepsi:
I would prefer having a minimal setup.py file for it. Not sure if versioning is the reason jax team still keeps it.

Qazalbash:
I need `pyproject.toml` to complete #2129. If this PR is no longer active, would it be okay for me to use the existing `pyproject.toml`, or is @nstarman planning to continue working on it in the near future?

PR #1894, opened 2024-10-23
Fix handling of event dimensions in `ComposeTransform` (fixes #1893).

### LINKED ISSUE #1893: Incorrect event dimension for `ComposeTransform` if transform reduces event dimensions
An example is probably the clearest illustration. If the transform increases the number of event dimensions, everything works as expected.

```python
>>> from numpyro.distributions.transforms import CorrCholeskyTransform, ComposeTransform
>>> part = CorrCholeskyTransform()
>>> part.domain.event_dim, part.codomain.event_dim
(1, 2)
>>> composed = ComposeTransform([part])
>>> composed.domain.event_dim, composed.codomain.event_dim
(1, 2)
```

If the transform reduces the number of event dimensions, wrapping the transform in a `ComposeTransform` leads to unexpected results.

```python
>>> part = CorrCholeskyTransform().inv
>>> part.domain.event_dim, part.codomain.event_dim
(2, 1)
>>> composed = ComposeTransform([part])
>>> composed.domain.event_dim, composed.codomain.event_dim
(3, 1)
```

I had a brief look at the code below, but I couldn't quite get my head around it.

https://github.com/pyro-ppl/numpyro/blob/8ace34f1aca094897328730853b021d53b230f65/numpyro/distributions/transforms.py#L292-L298

Here's a minimal test that can be added to `test/test_transforms.py` to reproduce.

```python
@pytest.mark.parametrize("transform", [
    CorrCholeskyTransform(),  # passes
    CorrCholeskyTransform().inv,  # fails
])
def test_compose_domain_codomain(transform):
    composed = ComposeTransform([transform])
    assert transform.domain.event_dim == composed.domain.event_dim
    assert transform.codomain.event_dim == composed.codomain.event_dim
```

tillahoffmann:
This PR fixes #1893, I think. It looks like the implicit end in the slice caused some trouble if there's only one transform.

### DIFF
--- a/numpyro/distributions/transforms.py
+++ b/numpyro/distributions/transforms.py
@@ -282,7 +282,7 @@ class AffineTransform(Transform):
 
 def _get_compose_transform_input_event_dim(parts):
     input_event_dim = parts[-1].domain.event_dim
-    for part in parts[len(parts) - 1 :: -1]:
+    for part in parts[:-1][::-1]:
         input_event_dim = part.domain.event_dim + max(
             input_event_dim - part.codomain.event_dim, 0
         )
--- a/test/test_transforms.py
+++ b/test/test_transforms.py
@@ -397,3 +397,16 @@ def test_biject_to(constraint, shape):
     expected_shape = constrained.shape[: constrained.ndim - constraint.event_dim]
     assert passed.shape == expected_shape
     assert jnp.all(passed)
+
+
+@pytest.mark.parametrize(
+    "transform",
+    [
+        CorrCholeskyTransform(),
+        CorrCholeskyTransform().inv,
+    ],
+)
+def test_compose_domain_codomain(transform):
+    composed = ComposeTransform([transform])
+    assert transform.domain.event_dim == composed.domain.event_dim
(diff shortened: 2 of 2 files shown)


numpyro/distributions/transforms.py
```diff
@@ -282,7 +282,7 @@ def __eq__(self, other):
 
 def _get_compose_transform_input_event_dim(parts):
     input_event_dim = parts[-1].domain.event_dim
-    for part in parts[len(parts) - 1 :: -1]:
+    for part in parts[len(parts) - 1 : len(parts) : -1]:
```
fehiepsi, line 285: Thanks for the catch! I think we need to ignore the -1 part here. Maybe using `parts[:-1][::-1]` to be less confusing?

fehiepsi: It was surprised to me that
```
x = [10]
print(x[-2::-1])  # Output: [10] 
```

tillahoffmann: Sounds great!

fehiepsi (approved):
Thanks, Till!

PR #1898, opened 2024-11-04
Updated SteinVI docs

OlaRonning:
Added ASVGD and SVGD to the SteinVI documentation.

### DIFF
--- a/docs/source/contrib.rst
+++ b/docs/source/contrib.rst
@@ -16,7 +16,7 @@ Nested Sampling is a non-MCMC approach that works for arbitrary probability mode
 
 Stein Variational Inference
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~
-Stein Variational Inference (SteinVI) is a family of VI techniques for approximate Bayesian inference based on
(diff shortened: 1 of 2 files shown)


juanitorduz:
Hi! I have thought about giving it a try to SteinVI, but I am unsure about the intuition of the kernel selection and maybe a rule of thumb for the number of particles and other parameters. I know there are some papers referenced, but it would be great to have some comments or tips around these practical details (of course not as part of this PR, but just a kind request if possible and easy to add sometime) :)

OlaRonning:
@juanitorduz  I've added some recommendations regarding the number of particles, number of ELBO particles, optimizer and kernel in the class description for [SteinVI](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/contrib/einstein/steinvi.py#L78-L96). I figured that is what shows up in users IDE.

Can you write what you'd like recommendations for in an issue or on the forum? That way, I can look at it, and we don't pollute this PR :)

juanitorduz:
Thanks! Absolutely! I will do 🙇

fehiepsi (approved):

Issue #1861, opened 2024-09-21
Exclude_deterministic argument in Predictive does not apply for models with discrete latents


fehiepsi:
See the forum issue https://forum.pyro.ai/t/enumerate-support-for-batch-dimensions-of-custom-distribution/7656/4

We need to move the logic of exclude_deterministic to the `infer_discrete` branch https://github.com/pyro-ppl/numpyro/blob/3cde93d0f25490b9b90c1c423816c6cfd9ea23ed/numpyro/infer/util.py#L793-L818

PR #1901, opened 2024-11-10
[BUG] Latent exclude deterministic

juanitorduz:
Closes https://github.com/pyro-ppl/numpyro/issues/1861

The test fails in the master branch but passes with the change in the `_predictive` function.

Is this enough or do we want to test something else?


fehiepsi (approved):
Thanks @juanitorduz!

Issue #1645, opened 2023-09-22
pmap over `num_particles` in SVI


snehjp2:
Hi, 

In `Trace_ELBO`, the `num_particles` argument allows one to effectively introduce a batch size in estimating the ELBO gradient if `num_particles > 1`. By default, it's vectorized over the `num_particles`. Is it possible to also distribute the batch dimension over devices (e.g. when running on multiple GPUs). My particular application is prone to jax OOM errors and would benefit from distribution over `jax.pmap`.

fehiepsi:
If you got OOM, you can set vectorize particles to False. You can also use PositionalSharding like in [MCMC](https://num.pyro.ai/en/stable/mcmc.html?highlight=sharding#numpyro.infer.mcmc.MCMC) I guess.

If you want to pmap over particles, could you make a PR for it? I think we can just simply allow a **callable** `vectorize_particles` and call it [here](https://github.com/pyro-ppl/numpyro/blob/6e3f00774d7f79b53008c332bc9d2e68c8702f24/numpyro/infer/elbo.py#L171).

PR #1902, opened 2024-11-10
Allow callable for vectorization over particles on SVI

juanitorduz:
Closes https://github.com/pyro-ppl/numpyro/issues/1645

### DIFF
--- a/numpyro/infer/elbo.py
+++ b/numpyro/infer/elbo.py
@@ -34,9 +34,9 @@ class ELBO:
 
     :param num_particles: The number of particles/samples used to form the ELBO
         (gradient) estimators.
-    :param vectorize_particles: Whether to use `jax.vmap` to compute ELBOs over the
-        num_particles-many particles in parallel. If False use `jax.lax.map`.
-        Defaults to True.
+    :param vectorize_particles: Callable to vectorize the computation of ELBOs over
+        the num_particles-many particles. Defaults to `jax.lax.map`. Other options are
+        `jax.vmap` and `jax.pmap`.
     """
 
     """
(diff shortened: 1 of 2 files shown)


numpyro/infer/elbo.py
```diff
@@ -34,9 +34,9 @@ class ELBO:
 
     :param num_particles: The number of particles/samples used to form the ELBO
         (gradient) estimators.
-    :param vectorize_particles: Whether to use `jax.vmap` to compute ELBOs over the
-        num_particles-many particles in parallel. If False use `jax.lax.map`.
-        Defaults to True.
+    :param vectorize_particles: Callable to vectorize the computation of ELBOs over
```
fehiepsi, line 37: this change breaks current users. You can allow callable and check callable first. Defaults still to True (vmap).

numpyro/infer/elbo.py
```diff
@@ -46,7 +46,7 @@ class ELBO:
     """
     can_infer_discrete = False
 
-    def __init__(self, num_particles=1, vectorize_particles=True):
+    def __init__(self, num_particles=1, vectorize_particles=jax.lax.map):
         self.num_particles = num_particles
         self.vectorize_particles = vectorize_particles
```
fehiepsi, line 51: maybe
```python

def _apply_vmap(fn, keys):
    return vmap(fn)(keys)
...

    self.vectorize_particles = vectorize_particles
    if callable(vectorize_particles):
        self.vectorize_particles_fn = vectorize_particles
    elif vectorize_particles is True:
        self.vectorize_particles_fn = apply_vmap
    else:
        self.vectorize_particles_fn = jax.lax.map
```

juanitorduz: cool! I added this suggestion!

juanitorduz:
Thanks for the suggestion @fehiepsi ! I added it to the PR and added some tests.

fehiepsi (approved):

fehiepsi:
Thanks, @juanitorduz!

PR #1905, opened 2024-11-12
Allow to collect nested dict keys in mcmc

### LINKED ISSUE #1903: 'z' values for HMC and NUTS seem to be in the constrained space which is inconsistent with the current docs
Hi,

Thank you for this amazing package! 

The docs for [HMCState](https://num.pyro.ai/en/stable/mcmc.html#numpyro.infer.hmc.HMCState) currently mention:
z - Python collection representing values **(unconstrained samples from the posterior)** at latent sites.

However, I think it's actually returning the parameter values in the constrained space.
Here's a toy model which I think shows this issue.

```python
import jax
import jax.numpy as jnp
import numpy as np
import numpyro
import numpyro.distributions as dist, numpyro.infer as infer
from functools import partial

jax.config.update('jax_enable_x64', True)

rng = np.random.default_rng(seed=12)

x = np.linspace(0, 10, 100)
m_true, b_true = 1.0, 5.0
y_true = m_true*x + b_true
sigma = 1.5
yerr = sigma * rng.standard_normal(x.size)

y = y_true + yerr
fig, ax = plt.subplots(dpi=200)
ax.plot(x, y_true, color='k')
ax.errorbar(x, y, yerr=1, ls='none', marker='.')
ax.set_xlabel('x')
ax.set_ylabel('y')
```
image

```python
# define numpyro model
def model(x, yerr, y=None):
    m = numpyro.sample('m', dist.TruncatedNormal(low=0.0))
    b = numpyro.sample('b', dist.Normal())

    y_model = m * x + b

    numpyro.sample('obs', numpyro.distributions.Normal(y_model, yerr), obs=y)


kernel = numpyro.infer.NUTS(model)

sampler = numpyro.infer.MCMC(
    kernel,
    num_warmup=500,
    num_samples=1000,
    progress_bar=True,
)
sampler.run(jax.random.PRNGKey(12), x, yerr=sigma, y=y)
sampler.print_summary()
samples = sampler.get_samples()  # constrained_space
z = sampler._states['z']  # docs mention this is in unconstrained space
print(jnp.allclose(z['m'], samples['m']), jnp.allclose(z['b'], samples['b']))  # (True, True)

_unconstrain_fn = partial(infer.util.unconstrain_fn,
                          model,
                          (x, sigma),
                          {'y': y},
                          )

unconstrained_parameters = jax.vmap(_unconstrain_fn, in_axes=({'m': 0, 'b': 0},))(samples)

fig, axs = plt.subplots(nrows=2, dpi=200)
for i, param in enumerate(['m', 'b']):
    axs[i].plot(jnp.squeeze(z[param]), label='z')
    axs[i].plot(unconstrained_parameters[param], ls='--', label='unconstrained')
axs[0].axhline(m_true, lw=1, color='k', label='true value')
axs[1].axhline(b_true, lw=1, color='k', label='true value')

axs[0].set_ylabel('m')
axs[1].set_ylabel('b')

for ax in axs:
    ax.legend(fontsize=8)
```
image

For the **m** parameter, the values returned from calling `sampler._states['z']` are exactly the same as the samples returned from `sampler.get_samples()`, which are in the constrained space. I've plotted the unconstrained values for **m** which seem to be correct given that they take on negative values (which would be outside the support for the `TruncatedNormal` I defined). And `b` is the same since it's unconstrained anyways.

My guess is that this behavior is coming from lines 494-501 from the `infer.mcmc.py` file,
https://github.com/pyro-ppl/numpyro/blob/c8a09908552ec90192e89b3d495b376e2038284b/numpyro/infer/mcmc.py#L494C9-L501C34
```python
        if len(site_values) > 0 and jnp.shape(site_values[0])[0] > 0:
            if self._jit_model_args:
                states[self._sample_field] = postprocess_fn(
                    states[self._sample_field], args, kwargs
                )
            else:
                states[self._sample_field] = postprocess_fn(states[self._sample_field])
        return states, last_state
```
since it's calling `postprocess_fn` on `states` (`last state` is in the unconstrained space though), and the `postprocess_fn` is no longer the identity function at this point.

If this is the intended behavior (and I'm not totally misunderstanding something), would it be possible to update the docs to highlight that `z` is actually in the **constrained space**?

Thanks again for all the work you put into this package!

fehiepsi:
Fixes #1903

@soichiro-hattori I think this should address your usage case. Please see the test as an example.

### DIFF
--- a/numpyro/infer/mcmc.py
+++ b/numpyro/infer/mcmc.py
@@ -586,7 +586,10 @@ class MCMC(object):
         :param extra_fields: Extra fields (aside from :meth:`~numpyro.infer.mcmc.MCMCKernel.default_fields`)
             from the state object (e.g. :data:`numpyro.infer.hmc.HMCState` for HMC) to collect during
             the MCMC run. Exclude sample sites from collection with "~`sampler.sample_field`.`sample_site`".
-            e.g. "~z.a" will prevent site "a" from being collected if you're using the NUTS sampler.
+            e.g. "~z.a" will prevent site "a" from being collected if you're using the NUTS sampler. To
+            collect samples of a site "a" in the unconstrained space, we can specify the variable here, e.g.
+            `extra_fields=("z.a",)`.
+
         :type extra_fields: tuple or list
         :param bool collect_warmup: Whether to collect samples from the warmup phase. Defaults
             to `False`.
@@ -623,7 +626,8 @@ class MCMC(object):
             during the MCMC run. Note that subfields can be accessed using dots, e.g.
             `"adapt_state.step_size"` can be used to collect step sizes at each step. Exclude sample sites from
             collection with "~`sampler.sample_field`.`sample_site`". e.g. "~z.a" will prevent site "a" from
-            being collected if you're using the NUTS sampler.
+            being collected if you're using the NUTS sampler. To collect samples of a site "a" in the
+            unconstrained space, we can specify the variable here, e.g. `extra_fields=("z.a",)`.
         :type extra_fields: tuple or list of str
         :param init_params: Initial parameters to begin sampling. The type must be consistent
             with the input type to `potential_fn` provided to the kernel. If the kernel is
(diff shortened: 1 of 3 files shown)


OlaRonning (approved):
LGTM

Issue #299, opened 2019-08-27
Use type hints whenever possible


neerajprad:
Since we are using python 3, we can start taking advantage of type hints wherever possible.

akihironitta:
@neerajprad @jpchen Could I work on this issue?

neerajprad:
@akihironitta - The codebase and number of contributors to numpyro core is small enough that this is low priority for us at the moment. When I created the issue, I was thinking that we could do this in a piecemeal fashion as we were refactoring the codebase. If you are interested, my suggestion would be to start with one of the other tasks and refactor to use type hints on the internal code that your PR touches.

juanitorduz:
@fehiepsi is this something you would like to add? I have done it for other libraries and we could tackle this on PR per module at a time (as in https://github.com/pyro-ppl/pyro/issues/2550) . We could also use https://github.com/patrick-kidger/jaxtyping (which I use and like quite a lot ;) )

fehiepsi:
I think it would be nice to have. Because this would require a significant amount of work, please make the PRs modular and incremental.

juanitorduz:
> I think it would be nice to have. Because this would require a significant amount of work, please make the PRs modular and incremental.

Ok! We can test a simple first PR for a module and see if we like it  👍

PR #1906, opened 2024-11-12
Proof of Concept: Types and MyPy

juanitorduz:
This is the first iteration in the journey to add type hints to the project https://github.com/pyro-ppl/numpyro/issues/299 .

We can add types on small and modular iterations, and the community can help (so these can be labeled as good first issues). I found this nice blog https://breadcrumbscollector.tech/mypy-how-to-use-it-in-my-project/ that provides tips on how to add types gradually. 

Let me you what you think about this proof-of-concept PR

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@ lint: FORCE
 	ruff check .
 	ruff format . --check
 	python scripts/update_headers.py --check
+	mypy --install-types --non-interactive numpyro
 
 license: FORCE
 	python scripts/update_headers.py
--- a/numpyro/contrib/control_flow/scan.py
+++ b/numpyro/contrib/control_flow/scan.py
@@ -3,6 +3,7 @@
 
 from collections import OrderedDict
 from functools import partial
+from typing import Callable
 
 import jax
 from jax import device_put, lax, random
@@ -278,7 +279,7 @@ def scan_wrapper(
     length,
     reverse,
     rng_key=None,
-    substitute_stack=[],
+    substitute_stack=None,
     enum=False,
     history=1,
     first_available_dim=None,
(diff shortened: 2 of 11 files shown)


numpyro/contrib/control_flow/scan.py
```diff
@@ -278,7 +279,7 @@ def scan_wrapper(
     length,
     reverse,
     rng_key=None,
-    substitute_stack=[],
+    substitute_stack=None,
```
juanitorduz, line 282: This is something I spotted while adding hints. We would want to avoid having mutables as defaults.

Makefile
```diff
@@ -4,6 +4,7 @@ lint: FORCE
 	ruff check .
 	ruff format . --check
 	python scripts/update_headers.py --check
+	mypy --install-types --non-interactive numpyro
```
juanitorduz, line 7: See https://github.com/pyro-ppl/pyro/blob/455f7b3b8b21f8e93a96235fc6bd58cb60f8a3fa/Makefile#L24

pyproject.toml
```diff
@@ -102,3 +102,15 @@ doctest_optionflags = [
     "NORMALIZE_WHITESPACE",
     "IGNORE_EXCEPTION_DETAIL",
 ]
+
+[tool.mypy]
+ignore_errors = true
+ignore_missing_imports = true
+
+[[tool.mypy.overrides]]
+module = [
```
juanitorduz, line 111: Here, we can keep adding the modules we want to check. Eventually, we would like to simply remove this and check everything

fehiepsi:
LGTM! The changes are pretty clean to me. Do you want to submit this?

fehiepsi (approved):

juanitorduz:
> LGTM! The changes are pretty clean to me. Do you want to submit this?

Yes, for me, we can merge, and I continue (bit by bit) adding hints (you can also advertise this task and maybe others can help as well, maybe even some of the ones that did it for Pyro ;) )

fehiepsi:
Thanks @juanitorduz!

cc @ordabayevy

PR #1907, opened 2024-11-13
Add types for contrib.stochastic_support and improve the HSGP module

juanitorduz:
Related to https://github.com/pyro-ppl/numpyro/issues/299


numpyro/contrib/stochastic_support/dcc.py
```diff
@@ -3,6 +3,9 @@
 
 from abc import ABC, abstractmethod
 from collections import OrderedDict, namedtuple
+from typing import Any, Callable, OrderedDict as OrderedDictType
+
+from jaxlib.xla_extension import ArrayImpl
```
tillahoffmann, line 8: Regarding the default typing, what are the implications of using `ArrayImpl` vs `jax.numpy.ndarray`, for example?

juanitorduz: This is a great question! Thanks for the feedback! So we started trying with type-hints with `ArrayImpl` in https://github.com/pyro-ppl/numpyro/pull/1819#discussion_r1649747153 

On the other hand in https://jax.readthedocs.io/en/latest/jax.typing.html its suggested to use `jax.typing.ArrayLike`. However when I use this for `x: ArrayLike` and have inside the function `x.ndim`, `mypy` complains with 

```python
error: Item "builtins.bool" of "Array | ndarray[Any, Any] | numpy.bool | number[Any] | builtins.bool | int | float | complex" has no attribute "ndim"  [union-attr]
```

Hence, this is why I am continuing with `ArrayImpl` and waiting for feedback :)

fehiepsi: I would like to avoid importing stuff from jaxlib. I guess it is better to use `jax.Array`

fehiepsi: Re ndim: you can use jnp.ndim(x), which will accept ArrayLike instances, including python scalars.

juanitorduz: Great input! I'll try that then 🙂!

juanitorduz:
ok! After changing the types (and also updating https://github.com/pyro-ppl/numpyro/pull/1906). I think this is the way to go as suggested ( I even caught minor issues) :) 

As always, feedback is welcome!

fehiepsi (approved):
Looks great! I like ArrayLike -> Array, looks consistent with jax api. If not much work, let's also replace jnp.ndarray by jax.Array (optional).

PR #1912, opened 2024-11-17
[Types] Diagnostics and Utils

juanitorduz:
Add more type-hints. Related to https://github.com/pyro-ppl/numpyro/issues/299 .

### DIFF
--- a/numpyro/diagnostics.py
+++ b/numpyro/diagnostics.py
@@ -24,7 +24,7 @@ __all__ = [
 ]
 
 
-def _compute_chain_variance_stats(x):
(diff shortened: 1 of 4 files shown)


numpyro/diagnostics.py
```diff
@@ -24,7 +24,7 @@
 ]
 
 
-def _compute_chain_variance_stats(x):
+def _compute_chain_variance_stats(x: np.ndarray) -> tuple[np.ndarray, np.ndarray]:
```
juanitorduz, line 27: All of these functions assume true numpy arrays, so this was the only way to make it work without adding many changes.

fehiepsi (approved):

PR #1913, opened 2024-11-18
Add method to retrieve distribution arguments.

tillahoffmann:
This PR separates getting distribution arguments and validating them (cf. #1866). The utility function allows for retrieval of arguments without having to check which arguments were set when the distribution was constructed (e.g., `scale_tril`, `precision_matrix`, or `covariance_matrix` for `MultivariateNormal` distributions).


fehiepsi (approved):
Thanks, Till!

Issue #1922, opened 2024-12-02
Python 3.9 broken


tingiskhan:
I noticed that version 0.16.0 breaks Python 3.9 support with [this change](https://github.com/pyro-ppl/numpyro/commit/c110053eb95f5398f11b93ce3efa21de4559012c) with the following error message:
```
  File "/home/..../numpyro/diagnostics.py", line 233, in <module>
    samples: dict | np.ndarray, prob: float = 0.90, group_by_chain: bool = True
TypeError: unsupported operand type(s) for |: 'type' and 'type'
```
The setup.py file still indicates support for Python 3.9, so I assume this is supposed to work?

juanitorduz:
@fehiepsi do you wanna support python 3.9? If we do we can amend these changes but need to make sure we set python 3.9 in https://github.com/pyro-ppl/numpyro/blob/master/.github/workflows/ci.yml#L17 because otherwise we can not catch these type of issues in the test suite. However, I think JAX just supports 3.10 and higher https://github.com/jax-ml/jax/blob/main/setup.py#L119?

victor-4p:
> @fehiepsi do you wanna support python 3.9? If we do we can amend these changes but need to make sure we set python 3.9 in https://github.com/pyro-ppl/numpyro/blob/master/.github/workflows/ci.yml#L17 because otherwise we can not catch these type of issues in the test suite. However, I think JAX just supports 3.10 and higher https://github.com/jax-ml/jax/blob/main/setup.py#L119?

From a user perspective I would only want having python_version>=3.10 in setup.py s.t. I don't install incompatible versions. It seems to be as you say that JAX > 0.4.30 only supports Python 3.10+.

PR #1923, opened 2024-12-02
Support Python 3.9

juanitorduz:
Closes https://github.com/pyro-ppl/numpyro/issues/1922

The lint commit https://github.com/pyro-ppl/numpyro/pull/1923/commits/d32b54a8dc0be62e80999f595aff5653b059b5da failed as expected. The fix in https://github.com/pyro-ppl/numpyro/actions/runs/12120151093/job/33788225675?pr=1923 makes the Python 3.9 linter happy.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.9","3.10","3.13"]
+        python-version: ["3.9","3.10","3.12"]
 
     steps:
     - uses: actions/checkout@v2


fehiepsi:
Could you also add 3.12 and 3.13 to the setup file and add 3.13 to the lint check? (we can fix bug on those python versions in later release if users report) Thanks!

juanitorduz:
@fehiepsi Do we want to extend the run the CI tests with Python 3.9 as well?

fehiepsi:
yes, I think we can add tests for modeling and inference.

juanitorduz:
ok! https://github.com/pyro-ppl/numpyro/pull/1923/commits/8346b55c92afd9745c5c6523d0e0ce08d1fa1185 tests work fine (see https://github.com/pyro-ppl/numpyro/pull/1923/commits/8346b55c92afd9745c5c6523d0e0ce08d1fa1185) However, there seems to be an issue installing dependencies with Python 3.13 👀

fehiepsi (approved):
Awesome! thanks @juanitorduz!

Some dependencies in dev might not be compatible with python 3.13. That's fine for now I think. We can revisit in the future. Let me know if this is ready to merge.

juanitorduz:
Ok! In this case this first iteration is complete :)

Issue #1928, opened 2024-12-04
Pytree structure of distributions with `lazy_property` depends on previous attribute access.


tillahoffmann:
The pytree structure of distributions with `lazy_property` depends on which attributes have previously been accessed, which can cause trouble with `jax.lax.scan`. This is probably most easily explained with an example.

```python
>>> import jax
>>> from jax import numpy as jnp
>>> from numpyro import distributions as dists
>>> 
>>> 
>>> def make_dist():
>>>     return dists.MultivariateNormal(precision_matrix=jnp.eye(2))
>>> 
>>> 
>>> init = make_dist()
>>> init.covariance_matrix  # No error if this line is commented.
>>> jax.lax.scan(lambda carry, _: (make_dist(), None), init, jnp.arange(7))
---------------------------------------------------------------------------
TypeError: Scanned function carry input and carry output must have the same 
pytree structure, but they differ:
  * the input carry component carry[<flat index 1>] is a 
    <class 'jaxlib.xla_extension.ArrayImpl'> but the corresponding component 
    of the carry output is a <class 'NoneType'>, so their Python types differ

Revise the scanned function so that its output is a pair where the first 
element has the same pytree structure as the first argument.
```

Before accessing `covariance_matrix`, the field does not appear in `__dict__` such that the flattening code replaces the field by `None` here.

https://github.com/pyro-ppl/numpyro/blob/244ff38c727633f8ab5a67ee24444461fc6d3954/numpyro/distributions/distribution.py#L173-L179

Upon accessing `covariance_matrix`, the `lazy_property` is replaced by a value which in turn replaces the `None` in the pytree representation.

```python
>>> d = make_dist()
>>> "covariance_matrix" in d.__dict__
False
>>> d.covariance_matrix
Array([[1., 0.],
       [0., 1.]], dtype=float32)
>>> "covariance_matrix" in d.__dict__
True
```

For each distribution, it seems there is a set of canonical parameters which are not `lazy_property`s (`scale_tril` in the case of the `MultivariateNormal` distribution). Would it make sense to only include names in `arg_constraints` if the corresponding element on the distribution class is not a `lazy_property`? I'm thinking of these lines:

https://github.com/pyro-ppl/numpyro/blob/244ff38c727633f8ab5a67ee24444461fc6d3954/numpyro/distributions/distribution.py#L146-L159

Happy to make the change if that seems like a reasonable approach. A downside might be that caching of `lazy_property`s does not persist across jitted functions.

[...]

PR #1929, opened 2024-12-05
Only include `arg_constraints` in `pytree_data_fields` if they are not `lazy_property`s.

tillahoffmann:
This PR tries to address #1928 to obtain consistent pytree structures.

On a related note, similar behavior is observed with the `get_args` method. I'm not sure what the "correct" behavior should be in this case.

```python
>>> import jax
>>> from numpyro import distributions as dists
>>> 
>>> 
>>> def make_dist():
>>>     return dists.MultivariateNormal(precision_matrix=jnp.eye(2))
>>> 
>>> 
>>> init = make_dist()
>>> init.get_args()
{'loc': array([0.]),
 'precision_matrix': Array([[1., 0.],
        [0., 1.]], dtype=float32),
 'scale_tril': Array([[1., 0.],
        [0., 1.]], dtype=float32)}
>>> init.covariance_matrix
>>> init.get_args()
{'loc': array([0.]),
 'covariance_matrix': Array([[1., 0.],
        [0., 1.]], dtype=float32),
 'precision_matrix': Array([[1., 0.],
        [0., 1.]], dtype=float32),
 'scale_tril': Array([[1., 0.],
        [0., 1.]], dtype=float32)}
```

### DIFF
--- a/numpyro/distributions/distribution.py
+++ b/numpyro/distributions/distribution.py
@@ -147,16 +147,20 @@ class Distribution(metaclass=DistributionMeta):
     def gather_pytree_data_fields(cls):
         bases = inspect.getmro(cls)
 
-        all_pytree_data_fields = ()
+        all_pytree_data_fields = set()
         for base in bases:
             if issubclass(base, Distribution):
-                all_pytree_data_fields += base.__dict__.get(
-                    "pytree_data_fields",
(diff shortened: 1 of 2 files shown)


fehiepsi (approved):
LGTM! Re get_args: it depends on the definition of the method. It might be worth to add a clear description there. Any method with lazy properties might give inconsistent behaviors, not just with get_args.

Issue #1937, opened 2024-12-17
icdf not implemented in Truncated Distribution


TheSkyentist:
**Context:** 
I am attempted to perform Nested Sampling using the contributed `jaxns` wrapper. Part of what the wrapper attempts to do (I believe) is re-parametrize my priors to Uniform distributions (which makes sense). However this fails as the `icdf` method is not implemented for Truncated Distributions which best reflect (some of) my priors. 

**Minimal Work Example:**
```python
#! /usr/bin/env python

# JAX packages
from jax import random
from jax import numpy as jnp

# Bayesian Inference
from numpyro import distributions as dist, sample
from numpyro.contrib.nested_sampling import NestedSampler

# Random key
rng_key = random.PRNGKey(0)

# Define data
x, err = 0.0, 1 / jnp.sqrt(2 * jnp.pi)

def model(x, err, π):
    center = sample('center', dist.TruncatedNormal(0, π, low = -10, high = 10))

    # Delta function likelihood
    sample('obs', dist.Normal(center, err), obs=x)


# Model Arguements
model_args = (x, err, 5)

# Nested Sampling
constructor_kwargs = {'num_live_points': 1000, 'max_samples': 50000}
termination_kwargs = {'dlogZ': 0.01}
NS = NestedSampler(
    model=model,
    constructor_kwargs=constructor_kwargs,
    termination_kwargs=termination_kwargs,
)
NS.run(rng_key, *model_args)
NS.print_summary()
```
This fails with a `NotImplementedError` as `icdf`s are not implemented for Truncated distributions. 

**Steps Forward?**
Would it be easier to implement the `icdf` method for truncated distributions or perhaps is there a better way to transform my priors such that they are not exposed to this shortcoming? If the former is preferred, I'm happy to try and contribute to the effort, I wonder if there are special considerations that would need to be taken into account to maintain the correct statistics.

TheSkyentist:
The documentation suggests that the efficient way for building Truncated Distributions in the first place is to calculate their ICDF:
https://num.pyro.ai/en/stable/tutorials/truncated_distributions.html#3.-Sampling-from-truncated-and-folded-distributions

Whether this is how it is implemented currently remains to be seen, but perhaps allowing this implementation is just a matter of exposing existing functionality.

martinjankowiak:
[...]

PR #1938, opened 2024-12-17
Implement ICDF Methods for Truncated Distributions

TheSkyentist:
Implementation of generic Inverse Cumulative Distribution Function (ICDF) methods for Truncated Distributions. 

Sampling method for Truncated Distribution already calculates the ICDF, so the initial commit is refactoring the code into its own method. Still needs to be tested so starting this as a draft. 

Motivated by #1937 for use with Nested Sampling with `jaxns`.

### DIFF
--- a/numpyro/distributions/truncated.py
+++ b/numpyro/distributions/truncated.py
@@ -67,10 +67,13 @@ class LeftTruncatedDistribution(Distribution):
         finfo = jnp.finfo(dtype)
         minval = finfo.tiny
         u = random.uniform(key, shape=sample_shape + self.batch_shape, minval=minval)
+        return self.icdf(u)
+
+    def icdf(self, q):
         loc = self.base_dist.loc
         sign = jnp.where(loc >= self.low, 1.0, -1.0)
         return (1 - sign) * loc + sign * self.base_dist.icdf(
-            (1 - u) * self._tail_prob_at_low + u * self._tail_prob_at_high
+            (1 - q) * self._tail_prob_at_low + q * self._tail_prob_at_high
         )
 
     @validate_sample
@@ -138,7 +141,10 @@ class RightTruncatedDistribution(Distribution):
         finfo = jnp.finfo(dtype)
         minval = finfo.tiny
         u = random.uniform(key, shape=sample_shape + self.batch_shape, minval=minval)
-        return self.base_dist.icdf(u * self._cdf_at_high)
+        return self.icdf(u)
+
+    def icdf(self, q):
+        return self.base_dist.icdf(q * self._cdf_at_high)
 
     @validate_sample
     def log_prob(self, value):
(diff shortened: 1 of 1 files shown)


TheSkyentist:
Fixed an issue where the icdf functions would return non-NaN values for inputs outside of the range [0,1], although the values in the range are correct. Now the icdf function returns a NaN for inputs outside the range [0,1], which makes it consistent with other distributions as well.

numpyro/distributions/truncated.py
```diff
@@ -138,7 +142,11 @@ def sample(self, key, sample_shape=()):
         finfo = jnp.finfo(dtype)
         minval = finfo.tiny
         u = random.uniform(key, shape=sample_shape + self.batch_shape, minval=minval)
-        return self.base_dist.icdf(u * self._cdf_at_high)
+        return self.icdf(u)
+
+    def icdf(self, q):
+        ppf = self.base_dist.icdf(q * self._cdf_at_high)
+        return jnp.where(q > 1, jnp.nan, ppf)
```
fehiepsi, line 149: why do we need to cover q > 1?

numpyro/distributions/truncated.py
```diff
@@ -67,11 +67,15 @@ def sample(self, key, sample_shape=()):
...
+
+    def icdf(self, q):
         loc = self.base_dist.loc
         sign = jnp.where(loc >= self.low, 1.0, -1.0)
-        return (1 - sign) * loc + sign * self.base_dist.icdf(
-            (1 - u) * self._tail_prob_at_low + u * self._tail_prob_at_high
+        ppf = (1 - sign) * loc + sign * self.base_dist.icdf(
+            (1 - q) * self._tail_prob_at_low + q * self._tail_prob_at_high
         )
+        return jnp.where(q < 0, jnp.nan, ppf)
```
fehiepsi, line 78: why this where?

numpyro/distributions/truncated.py
```diff
@@ -235,19 +243,22 @@ def sample(self, key, sample_shape=()):
...
-        #   A = 2 * loc - icdf[(1 - u) * cdf(2*loc-low)) + u * cdf(2*loc - high)]
+        #   A = 2 * loc - icdf[(1 - q) * cdf(2*loc-low)) + q * cdf(2*loc - high)]
         loc = self.base_dist.loc
         sign = jnp.where(loc >= self.low, 1.0, -1.0)
-        return (1 - sign) * loc + sign * self.base_dist.icdf(
-            clamp_probs((1 - u) * self._tail_prob_at_low + u * self._tail_prob_at_high)
+        ppf = (1 - sign) * loc + sign * self.base_dist.icdf(
+            clamp_probs((1 - q) * self._tail_prob_at_low + q * self._tail_prob_at_high)
         )
+        return jnp.where(jnp.logical_or(q < 0, q > 1), jnp.nan, ppf)
```
fehiepsi, line 261: like above, why this where is needed?

fehiepsi (commented):
Thanks @TheSkyentist Could you double check if our tests already cover the change?

numpyro/distributions/truncated.py
TheSkyentist: Please see my above comment. If this is not implemented, the icdf value will return a numeric, non NaN value outside the input range of [0,1]. The behavior of other NumPyro distributions is to return NaN outside this range. A non-NaN value outside of the input range can potentially cause bugs and/or confusion. Of course in principle one shouldn’t call the icdf with a value outside this input range, but I figured it would be best practice to ensure expected behavior.

numpyro/distributions/truncated.py
TheSkyentist: See above comment :-)

numpyro/distributions/truncated.py
TheSkyentist: I could also imagine doing the following: If we're worried about the performance hit to `sample` because of the added `jnp.where` call, then we could wrap the un-masked `icdf` code in a separate helper function and only mask out-of-range values for the `icdf` method. If it's preferred to keep the behaviour where numeric values are returned when out of bounds, then I'm happy to revert the previous commit :-)

TheSkyentist:
I think the tests cover the change at least as far as I can see, but perhaps I'm not sure what you mean, could you clarify slightly? Thanks :-)

numpyro/distributions/truncated.py
fehiepsi: I think having the check is fine. Just curious, why not checking both q < 0 and q > 1?

TheSkyentist: If it’s a left/right truncated distribution, we only need to check on the relevant side. Otherwise the base distribution icdf handles if it’s out of range.

fehiepsi: Thanks for explaining!!

fehiepsi (approved):

PR #1940, opened 2024-12-20
Add type hints to the primitives module

juanitorduz:
Add type hints to the primitives module and refine other type hints accordingly.

### DIFF
--- a/numpyro/contrib/hsgp/approximation.py
+++ b/numpyro/contrib/hsgp/approximation.py
@@ -25,14 +25,14 @@ def _non_centered_approximation(phi: Array, spd: Array, m: int) -> Array:
     with numpyro.plate("basis", m):
         beta = numpyro.sample("beta", dist.Normal(loc=0.0, scale=1.0))
 
-    return jnp.asarray(phi @ (spd * beta))
+    return phi @ (spd * beta)
 
 
 def _centered_approximation(phi: Array, spd: Array, m: int) -> Array:
     with numpyro.plate("basis", m):
         beta = numpyro.sample("beta", dist.Normal(loc=0.0, scale=spd))
 
-    return jnp.asarray(phi @ beta)
+    return phi @ beta
 
 
 def linear_approximation(
--- a/numpyro/contrib/hsgp/approximation.py
+++ b/numpyro/contrib/hsgp/approximation.py
@@ -7,6 +7,8 @@ This module contains the low-rank approximation functions of the Hilbert space G
 
 from __future__ import annotations
 
+from typing import cast
+
 from jax import Array
 import jax.numpy as jnp
 from jax.typing import ArrayLike
@@ -21,26 +23,22 @@ from numpyro.contrib.hsgp.spectral_densities import (
 import numpyro.distributions as dist
 
 
-def _non_centered_approximation(
-    phi: ArrayLike, spd: ArrayLike, m: int | list[int]
-) -> Array:
+def _non_centered_approximation(phi: ArrayLike, spd: ArrayLike, m: int) -> Array:
     with numpyro.plate("basis", m):
         beta = numpyro.sample("beta", dist.Normal(loc=0.0, scale=1.0))
 
-    return phi @ (spd * beta)
+    return cast(Array, phi @ (spd * beta))
 
 
-def _centered_approximation(
-    phi: ArrayLike, spd: ArrayLike, m: int | list[int]
-) -> Array:
+def _centered_approximation(phi: ArrayLike, spd: ArrayLike, m: int) -> Array:
     with numpyro.plate("basis", m):
         beta = numpyro.sample("beta", dist.Normal(loc=0.0, scale=spd))
 
-    return phi @ beta
+    return cast(Array, phi @ beta)
 
 
 def linear_approximation(
-    phi: ArrayLike, spd: ArrayLike, m: int | list[int], non_centered: bool = True
+    phi: ArrayLike, spd: ArrayLike, m: int, non_centered: bool = True
 ) -> Array:
     """
     Linear approximation formula of the Hilbert space Gaussian process.
@@ -55,7 +53,7 @@ def linear_approximation(
     :param ArrayLike phi: laplacian eigenfunctions
     :param ArrayLike spd: square root of the diagonal of the spectral density evaluated at square
         root of the first `m` eigenvalues.
-    :param int | list[int] m: number of eigenfunctions in the approximation
+    :param int m: number of eigenfunctions in the approximation
     :param bool non_centered: whether to use a non-centered parameterization
     :return: The low-rank approximation linear model
     :rtype: Array
--- a/numpyro/contrib/hsgp/approximation.py
+++ b/numpyro/contrib/hsgp/approximation.py
@@ -7,8 +7,6 @@ This module contains the low-rank approximation functions of the Hilbert space G
 
 from __future__ import annotations
 
-from typing import cast
-
 from jax import Array
 import jax.numpy as jnp
 from jax.typing import ArrayLike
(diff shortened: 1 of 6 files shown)


numpyro/contrib/hsgp/approximation.py
```diff
@@ -7,6 +7,8 @@
 
 from __future__ import annotations
 
+from typing import cast
```
fehiepsi, line 10: maybe using `jnp.asarray(...)` instead of cast

juanitorduz: changed in https://github.com/pyro-ppl/numpyro/pull/1940/commits/ebc510c2aac444abe70246cb869fa835a7837a64

juanitorduz:
In https://github.com/pyro-ppl/numpyro/pull/1940/commits/ba9a07ae6e26e64201b87caeebd0b94b88a28c61 I suggest adding a ` DistributionLike` type as a `Protocol` (https://mypy.readthedocs.io/en/stable/protocols.html) which just list the minimal required attributes and methods expected from a **stochastic function** (from the docstrings)

numpyro/contrib/hsgp/approximation.py
```diff
@@ -21,26 +21,22 @@
 import numpyro.distributions as dist
 
 
-def _non_centered_approximation(
-    phi: ArrayLike, spd: ArrayLike, m: int | list[int]
```
fehiepsi, line 25: why Array is used instead of ArrayLike?

numpyro/diagnostics.py
```diff
@@ -25,7 +26,7 @@
 ]
 
 
-def _compute_chain_variance_stats(x: np.ndarray) -> tuple[np.ndarray, np.ndarray]:
+def _compute_chain_variance_stats(x: npt.NDArray) -> tuple[npt.NDArray, npt.NDArray]:
```
fehiepsi, line 29: why NDArray is used instead of ndarray?

numpyro/distributions/distribution.py
```diff
@@ -563,6 +564,38 @@ def is_discrete(self):
         return self.support.is_discrete
 
 
+@runtime_checkable
+class DistributionLike(Protocol):
```
fehiepsi, line 568: why Protocol is needed?

numpyro/primitives.py
```diff
@@ -674,7 +698,7 @@ def model(data):
     :rtype: ~jnp.ndarray
     """
     if not _PYRO_STACK:
-        return data
+        return jnp.asarray(data)
```
fehiepsi, line 701: why this is needed? can we just set the return type is ArrayLike?

juanitorduz:
I need to fix some stuff because of the last released version of mypy (https://pypi.org/project/mypy/#history) 😅

numpyro/contrib/hsgp/approximation.py
juanitorduz: because `ArrayLike` can be a scalar (see https://jax.readthedocs.io/en/latest/jax.typing.html) but we want them to be true arrays

numpyro/diagnostics.py
juanitorduz: MyPy started complaining while I was adding hints and decided to follow the typing guidelines from numpy https://numpy.org/doc/stable/reference/typing.html :) Then all errors were gone

numpyro/primitives.py
juanitorduz: Good point, changed in https://github.com/pyro-ppl/numpyro/pull/1940/commits/b4dacc09b3934c91efc906358c8be6123968d908

numpyro/distributions/distribution.py
juanitorduz: This protocol is needed to provide static type checking for functions that accept
    distribution objects. From what I see, NumPyro supports multiple distribution implementations:
    
    - numpyro.distributions.Distribution
    - funsor.Funsor 
    - tensorflow_probability.distributions.Distribution
    
    Rather than checking for each specific type, this protocol defines the common interface
    that all distribution objects must implement. This allows static type checkers to verify
    that any distribution object passed to a function has the required methods and properties,
    regardless of which implementation it uses.
    
    If we remove it then we will get many MyPy errors 😅 .
    
    I have seen this approach in other projects so this is why I suggest it :)

juanitorduz: (And tensorflow probability and fusor are optional dependencies so we can not use Union to create a common type 🙃)

numpyro/diagnostics.py
fehiepsi: interesting. thanks! maybe import `NDArray` instead of `npt`? It makes the code easier to read.

numpyro/distributions/distribution.py
fehiepsi: SGTM! I was a bit worried of introducing a new abstraction but I trust the tests :)

numpyro/contrib/hsgp/approximation.py
```diff
@@ -21,26 +21,22 @@
...
 
-def _centered_approximation(
-    phi: ArrayLike, spd: ArrayLike, m: int | list[int]
-) -> Array:
+def _centered_approximation(phi: Array, spd: Array, m: int) -> Array:
     with numpyro.plate("basis", m):
         beta = numpyro.sample("beta", dist.Normal(loc=0.0, scale=spd))
 
-    return phi @ beta
+    return jnp.asarray(phi @ beta)
```
fehiepsi, line 35: is this needed when inputs are Array now.

numpyro/contrib/hsgp/approximation.py
```diff
@@ -21,26 +21,22 @@
...
 
-def _non_centered_approximation(
-    phi: ArrayLike, spd: ArrayLike, m: int | list[int]
-) -> Array:
+def _non_centered_approximation(phi: Array, spd: Array, m: int) -> Array:
     with numpyro.plate("basis", m):
         beta = numpyro.sample("beta", dist.Normal(loc=0.0, scale=1.0))
 
-    return phi @ (spd * beta)
+    return jnp.asarray(phi @ (spd * beta))
```
fehiepsi, line 28: is this needed when inputs are Array now.

numpyro/primitives.py
```diff
@@ -4,21 +4,28 @@
...
 import jax.numpy as jnp
+from jax.typing import ArrayLike
 
 import numpyro
+from numpyro.distributions.distribution import DistributionLike
 from numpyro.util import find_stack_level, identity
 
-_PYRO_STACK = []
+# Type aliases
+MessageType = dict[str, Any]
```
fehiepsi, line 20: I guess Message is enough. We don't have Message class.

numpyro/contrib/hsgp/approximation.py
juanitorduz: True, changed in https://github.com/pyro-ppl/numpyro/pull/1940/commits/a7984ce697c5135086246c45692424b7603802a3

numpyro/contrib/hsgp/approximation.py
juanitorduz: True, changed in https://github.com/pyro-ppl/numpyro/pull/1940/commits/a7984ce697c5135086246c45692424b7603802a3

numpyro/diagnostics.py
juanitorduz: Sure! changed in https://github.com/pyro-ppl/numpyro/pull/1940/commits/1bd90ea119bb3fcb0faa8068fa2ee48707d0334f

numpyro/primitives.py
```diff
@@ -292,9 +316,9 @@ def deterministic(name, value):
     :param jnp.ndarray value: deterministic value to record in the trace.
     """
     if not _PYRO_STACK:
-        return value
+        return jnp.asarray(value)
```
fehiepsi, line 319: Could we return ArrayLike and avoid using asarray here and others? I use numpyro with string types so jnp.asarray does not work.

numpyro/primitives.py
```diff
@@ -253,15 +277,15 @@ def param(name, init_value=None, **kwargs):
         assert not callable(
             init_value
         ), "A callable init_value needs to be put inside a numpyro.handlers.seed handler."
-        return init_value
+        return jnp.asarray(init_value)
```
fehiepsi, line 280: like below, keep init_value

juanitorduz: Done! https://github.com/pyro-ppl/numpyro/pull/1940/commits/19e7cfc01fd37cb171d06a4990cddc9865c98fda

numpyro/primitives.py
juanitorduz: Done! https://github.com/pyro-ppl/numpyro/pull/1940/commits/19e7cfc01fd37cb171d06a4990cddc9865c98fda

fehiepsi (approved):
Really cool! Thanks, @juanitorduz!

juanitorduz:
I will continue with the handlers, and once we crack them, the rest should be easier (I hope). Thankyou  for your valuable feedback @fehiepsi 🙇

PR #1942, opened 2024-12-27
Add types optim module

juanitorduz:
Add types optim module

### DIFF
--- a/numpyro/optim.py
+++ b/numpyro/optim.py
@@ -9,7 +9,7 @@ suited for working with NumPyro inference algorithms.
 
 from collections import namedtuple
 from collections.abc import Callable
-from typing import Any, TypeVar, Union
+from typing import Any, TypeVar
 
 import jax
 from jax import jacfwd, lax, value_and_grad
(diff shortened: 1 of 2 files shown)


numpyro/optim.py
```diff
@@ -31,12 +32,12 @@
     "SM3",
 ]
 
-_Params = TypeVar("_Params")
+_Params = Any
```
fehiepsi, line 35: why use Any for _Params but not for OptState

fehiepsi (approved):
LGTM pending the above comment. Thanks, @juanitorduz!

juanitorduz:
Ups! I had replied the comment but forgotten to hit the "reply button" 🙈

numpyro/optim.py
fehiepsi: How about just using `params = self.get_params(state)?

juanitorduz: Unfortunately this does not work (it was my first try) :(

```python
numpyro/optim.py:103: error: Need type annotation for "params"  [var-annotated]
numpyro/optim.py:125: error: Type variable "numpyro.optim._Params" is unbound  [valid-type]
```

The reason is because of the logic above. We can not use TypeVar internally :/

juanitorduz: I have made another attempt which I think serves our purpose in https://github.com/pyro-ppl/numpyro/pull/1942/commits/de835c61e5f186d41ed992faedb780c03ea5f382 I had to skip one type check in 

```
    def get_params(opt_state: _MinimizeState) -> _Params:  # type: ignore[type-var]. # <- HERE
        flat_params, unravel_fn = opt_state
        return unravel_fn(flat_params)
```

Better?

fehiepsi: I guess we can use `Any` for both to avoid confusion.

juanitorduz: Sure! Changed in https://github.com/pyro-ppl/numpyro/pull/1942/commits/e24a7a24b9c89ebf895ea2486db14bb3754de954 :)

fehiepsi (approved):

PR #1943, opened 2025-01-01
feat: Levy distribution

Qazalbash:
Implementation of Levy distribution along with test cases


fehiepsi (approved):
Thanks, @Qazalbash! Could you add docstring for Levy to https://github.com/pyro-ppl/numpyro/blob/master/docs/source/distributions.rst?plain=1

Qazalbash:
Done.

fehiepsi (approved):

PR #1945, opened 2025-01-03
Add `support` argument to `Delta` distribution.

tillahoffmann:
This PR adds a `support` argument to the constructor of the `Delta` distribution which allows enforcement of constraints, such as having positive-definite `v`. The `distribution.support` evaluates to `constraints.independent([specified support, defaulting to real scalar], [specified event dimensions])`.


fehiepsi:
I'm curious how useful it is to have specific support for Delta?

PR #1946, opened 2025-01-06
Add batch and event shape to `repr`.

tillahoffmann:
This PR adds the batch and event shape to the representation of distributions to help with development and debugging.

### DIFF
--- a/numpyro/distributions/distribution.py
+++ b/numpyro/distributions/distribution.py
@@ -563,6 +563,13 @@ class Distribution(metaclass=DistributionMeta):
     def is_discrete(self):
         return self.support.is_discrete
 
+    def __repr__(self) -> str:
+        cls = self.__class__
+        return (
+            f"<{cls.__module__}.{cls.__name__} object at {id(self):#x} with batch "
+            f"shape {self.batch_shape} and event shape {self.event_shape}>"
+        )
+
 
 @runtime_checkable
 class DistributionLike(Protocol):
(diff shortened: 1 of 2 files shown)


numpyro/distributions/distribution.py
```diff
@@ -563,6 +563,13 @@ def icdf(self, q: ArrayLike) -> ArrayLike:
     def is_discrete(self):
         return self.support.is_discrete
 
+    def __repr__(self) -> str:
+        cls = self.__class__
+        return (
+            f"<{cls.__module__}.{cls.__name__} object at {id(self):#x} with batch "
```
OlaRonning, line 569: What is the behavior of `#x` in the f-string?

tillahoffmann: The `:#x` formats the id of the object as a `0x[hex]` to keep the `repr` method consistent with the default other than adding the shapes.

```python
>>> from numpyro import distributions as dists
>>> 
>>> d = dists.Normal().expand((5, 10, 7)).to_event(2)
>>> d
<numpyro.distributions.distribution.Independent object at 0x10723f7c0 with batch shape (5,) and event shape (10, 7)>
>>> object.__repr__(d)
'<numpyro.distributions.distribution.Independent object at 0x10723f7c0>'
```

OlaRonning (approved):
Looks good! Thanks @tillahoffmann.
I checked `print(Normal)` and `print(Normal())` and test case. All behave as expected.

PR #1959, opened 2025-01-26
fix(tests): update tolerance levels and `PRNGKey` usage for improved test stability

Qazalbash:
With the release of `jax==0.5.0`, the `jax_threefry_partitionable` mode is now enabled by default. This change has caused some test cases to fail because the random numbers generated with the same seed differ depending on the status of `jax_threefry_partitionable`. For example,

<details><summary>Example</summary>
<p>

```bash
>>> python -c "import jax; jax.print_environment_info()"
jax:    0.5.0
jaxlib: 0.5.0
numpy:  1.26.4
python: 3.10.16 | packaged by conda-forge | (main, Dec  5 2024, 14:16:10) [GCC 13.3.0]
device info: cpu-1, 1 local devices"
process_count: 1
platform: uname_result(system='Linux', node='beepboop', release='6.1.0-22-amd64', version='#1 SMP PREEMPT_DYNAMIC Debian 6.1.94-1 (2024-06-21)', machine='x86_64')

>>> JAX_THREEFRY_PARTITIONABLE=False python -c "from jax import random as jrd; key=jrd.PRNGKey(0); print(jrd.uniform(key))"
0.41845703

>>> python -c "from jax import random as jrd; key=jrd.PRNGKey(0); print(jrd.uniform(key))"
0.947667
```

</p>
</details> 

Tolerance levels are updated for failing test cases.


martinjankowiak:
@Qazalbash thanks for taking a close look at this !!

> After modifying the test cases to use appropriate seed values, most of the tests passed. However, in certain cases where the appropriate seed could not be identified, using higher numerical precision resolved the issue.

can you please explain why it's not possible (or why it's difficult) to get rid of more "magic numbers" like `19470715`? wherever possible it's probably preferably to increase the tolerance, provided the tolerance isn't being increased by too much...

Qazalbash:
While addressing a similar issue in a personal project, I found that the most practical solution was to modify the random seed. Since I am not an expert in probability or statistics, I assumed that the tolerances, as well as the relative and absolute error thresholds, are inherently tied to the algorithm's precision. To avoid altering these parameters, I opted to adjust the input data by changing the random keys.

Initially, I experimented with various integers, including prime numbers and other two-digit values. However, after some time, I decided to explore more meaningful options, such as historically significant dates or personal milestones, including my own birthday. Surprisingly, these choices proved effective in resolving the issue :).

fehiepsi:
I think we should avoid magic numbers and address the numerical issues directly. It is better to have tests which are rarely failing (rather than sometimes passing).

fehiepsi:
I can take a stab at it.

Qazalbash:
> I think we should avoid magic numbers and address the numerical issues directly. It is better to have tests which are rarely failing (rather than sometimes passing).

I second that! If you like I can test it.

fehiepsi:
Sure, thanks!!

Qazalbash:
@fehiepsi I have fixed all the tests and merged the master branch, too, except for `test/infer/test_mcmc.py::test_change_point_x64`, which is passing in 85d49828354a017b270cb0555ce87b878091ece4 and also on my machine. Its failure does not make sense to me, so I left this one to you.

test/infer/test_mcmc.py
```diff
@@ -161,7 +161,7 @@ def model(labels):
     assert samples["logits"].shape == (num_samples, N)
     # those coefficients are found by doing MAP inference using AutoDelta
     expected_coefs = jnp.array([0.97, 2.05, 3.18])
-    assert_allclose(jnp.mean(samples["coefs"], 0), expected_coefs, atol=0.1)
+    assert_allclose(jnp.mean(samples["coefs"], 0), expected_coefs, atol=0.29)
```
fehiepsi, line 164: Because the data is changed, the expected coefs are no longer valid. Could you remove the commend
```
# those coefficients are found by doing MAP inference using AutoDelta
expected_coefs = jnp.array([0.97, 2.05, 3.18])
```
and change this to
```
assert_allclose(jnp.mean(samples["coefs"], 0), true_coefs, atol=0.2)
```
If you want to keep `atol`, you can increase N so that the expected coefs are nearer to the true coefs.

test/infer/test_mcmc.py
```diff
@@ -899,7 +899,7 @@ def test_get_proposal_loc_and_scale(dense_mass):
     expected_loc = jnp.stack(expected_loc)
     expected_scale = jnp.stack(expected_scale)
     assert_allclose(actual_loc, expected_loc, rtol=1e-4)
-    assert_allclose(actual_scale, expected_scale, atol=1e-6, rtol=0.05)
+    assert_allclose(actual_scale, expected_scale, atol=1e-6, rtol=0.234)
```
fehiepsi, line 902: maybe rtol=0.3 is better?

test/test_distributions.py
```diff
@@ -53,7 +53,7 @@
 )
 from numpyro.nn import AutoregressiveNN
 
-TEST_FAILURE_RATE = 2e-5  # For all goodness-of-fit tests.
+TEST_FAILURE_RATE = 2.6e-06  # For all goodness-of-fit tests.
```
fehiepsi, line 56: I fixed the gof tests in #1965 by increasing num_samples for failing ones. You can port the logic here.

```
if any(
        name in jax_dist.__name__
        for name in ["BetaProportion", "MatrixNormal", "SineBivariateVonMises"]
    ):
        num_samples = 20000
```

test/test_distributions.py
```diff
@@ -1870,6 +1870,15 @@ def fn(*args):
...
 
     eps = 1e-3
+    atol = 0.01
+    rtol = 0.01
+    if jax_dist is dist.EulerMaruyama:
+        atol = 0.064
+        rtol = 0.042
+    elif jax_dist is dist.NegativeBinomialLogits:
+        atol = 0.013
+        rtol = 0.044
```
fehiepsi, line 1880: Sorry to let you put too much work on finding those numbers. Appreciate it!

Maybe change `rtol to 0.05` or higher for all distributions? No need to have a specific number like 0.044 I think.

test/test_distributions.py
```diff
@@ -1968,8 +1977,12 @@ def test_mean_var(jax_dist, sp_dist, params):
         if jnp.all(jnp.isfinite(sp_mean)):
             assert_allclose(jnp.mean(samples, 0), d_jax.mean, rtol=0.05, atol=1e-2)
         if jnp.all(jnp.isfinite(sp_var)):
+            rtol = 0.05
+            atol = 1e-2
+            if jax_dist is dist.InverseGamma:
+                rtol = 0.054
             assert_allclose(
-                jnp.std(samples, 0), jnp.sqrt(d_jax.variance), rtol=0.05, atol=1e-2
+                jnp.std(samples, 0), jnp.sqrt(d_jax.variance), rtol=rtol, atol=atol
```
fehiepsi, line 1985: you can change rtol to 0.06 or even 0.1 for all distributions

test/test_distributions.py
```diff
@@ -1998,8 +2011,8 @@ def test_mean_var(jax_dist, sp_dist, params):
         )
         expected_std = expected_std * (1 - jnp.identity(dimension))
 
-        assert_allclose(jnp.mean(corr_samples, axis=0), expected_mean, atol=0.01)
-        assert_allclose(jnp.std(corr_samples, axis=0), expected_std, atol=0.01)
+        assert_allclose(jnp.mean(corr_samples, axis=0), expected_mean, atol=0.011)
+        assert_allclose(jnp.std(corr_samples, axis=0), expected_std, atol=0.011)
```
fehiepsi, line 2015: using `0.02` is better than `0.011`

test/test_distributions.py
```diff
@@ -2453,7 +2466,11 @@ def test_biject_to(constraint, shape):
 
     # test inv
     z = transform.inv(y)
-    assert_allclose(x, z, atol=1e-5, rtol=1e-5)
+    atol = 1e-5
+    rtol = 1e-5
+    if constraint in [constraints.l1_ball]:
+        atol = 5e-5
```
fehiepsi, line 2472: how about using atol=1e-4 for all constraints

test/test_distributions.py
```diff
@@ -2590,9 +2607,11 @@ def test_bijective_transforms(transform, event_shape, batch_shape):
         else:
             expected = jnp.log(jnp.abs(grad(transform)(x)))
             inv_expected = jnp.log(jnp.abs(grad(transform.inv)(y)))
-
-        assert_allclose(actual, expected, atol=1e-6)
-        assert_allclose(actual, -inv_expected, atol=1e-6)
+        atol = 1e-6
+        if isinstance(transform, transforms.ComposeTransform):
+            atol = 2.2e-6
```
fehiepsi, line 2612: you can use atol=1e-5 for all transforms

test/test_distributions_util.py
```diff
@@ -286,13 +286,14 @@ def test_vec_to_tril_matrix(shape, diagonal):
...
+    key1, key2 = random.split(random.PRNGKey(0))
+    A = random.normal(key1, chol_batch_shape + (dim, dim))
     A = A @ jnp.swapaxes(A, -2, -1) + jnp.eye(dim)
-    x = random.normal(random.PRNGKey(0), vec_batch_shape + (dim,)) * 0.1
+    x = random.normal(key2, vec_batch_shape + (dim,)) * 0.1
     xxt = x[..., None] @ x[..., None, :]
     expected = jnp.linalg.cholesky(A + coef * xxt)
     actual = cholesky_update(jnp.linalg.cholesky(A), x, coef)
-    assert_allclose(actual, expected, atol=1e-4, rtol=1e-4)
+    assert_allclose(actual, expected, atol=3.8e-4, rtol=8e-4)
```
fehiepsi, line 296: using atol=1e-3 and rtol=1e-3

test/test_transforms.py
```diff
@@ -315,13 +315,11 @@ def test_bijective_transforms(transform, shape):
...
-    less_stable_transforms = (
-        CorrCholeskyTransform,
-        L1BallTransform,
-        StickBreakingTransform,
-    )
+    less_stable_transforms = (CorrCholeskyTransform, StickBreakingTransform)
     if isinstance(transform, less_stable_transforms):
         atol = 1e-2
+    elif isinstance(transform, (L1BallTransform, RecursiveLinearTransform)):
+        atol = 0.099
```
fehiepsi, line 322: maybe atol=0.1 here

fehiepsi:
Maybe the init values of the change_point test make MCMC stuck at local minimal. How about setting
```
from numpyro.infer import init_to_value
...

kernel = NUTS(model=model, init_strategy=init_to_value(values={"lambda1": 1, "lambda2": 72}))
```

test/test_distributions.py
fehiepsi: actually, my PR failed at the MatrixNormal distribution

fehiepsi: I ended up skip that test.

test/test_distributions.py
Qazalbash: What about `atol`? Should I make it same for all?

test/infer/test_mcmc.py
```diff
@@ -364,7 +362,9 @@ def model(data):
         31, 30, 13, 27, 0, 39, 37, 5, 14, 13, 22])
     # fmt: on
 
-    kernel = NUTS(model=model)
+    kernel = NUTS(
+        model=model, init_strategy=init_to_value(values={"lambda1": 1, "lambda2": 72})
+    )
```
fehiepsi, line 367: I dont understand why this test fails here but not for the master branch. Both are using python 3.9 with an old jax version. Maybe do pytest.skipif python 3.9 here.

test/test_distributions.py
Qazalbash: Nope, that's why I conditioned them out!

test/infer/test_mcmc.py
Qazalbash: Sure.

test/test_distributions.py
fehiepsi: I see. It is understandable that the sde distribution suffer froms numerical issues. Maybe use atol to 0.1 for it and use 0.02 for the rest. You can revert rtol change if it is unnecessary.

fehiepsi:
There are a couple of failing tests:
+ test/infer/test_gradient.py: you can simply increase atol I think
+ test/infer/test_autoguide.py::test_dais_vae: increase `svi.run` steps to e.g. 5000
+ test_discrete_gibbs_multiple_sites_chain: you can increase atol to 0.1

fehiepsi (approved):
Huge thanks for fixing this annoying issue, @Qazalbash!

Qazalbash:
It is pretty suspicious these test cases were passing in 85d49828354a017b270cb0555ce87b878091ece4 even though py3.10 had `jax==0.5.0`.

fehiepsi:
Could you increase atol there? I feel the test is incorrect

Qazalbash:
Some more test cases have failed!

Can you check `test/contrib/einstein/test_stein_loss.py::test_stein_particle_loss`? It has a relatively large error.
```bash
>           assert_allclose(act_loss, exp_loss)
E           AssertionError: 
E           Not equal to tolerance rtol=1e-07, atol=0
E           
E           Mismatched elements: 1 / 1 (100%)
E           Max absolute difference among violations: 14.192255
E           Max relative difference among violations: 5.6824265
E            ACTUAL: array(-16.689825, dtype=float32)
E            DESIRED: array(-2.49757, dtype=float32)

test/contrib/einstein/test_stein_loss.py:95: AssertionError
```

Qazalbash:
@fehiepsi, I enabled continue on error in 295136f12ffd185f110c58c669940ed0686224d8, to get all the failed test cases at once, oblivious to the fact the test suite passed on that commit. Do you like to revert it?

fehiepsi:
Sure. I dont think it is needed for other PRs

PR #1962, opened 2025-01-26
Make `pip install` calls compatible with `zsh`

### LINKED ISSUE #1961: Enclose `.[<optional deps>]` in quotation marks for `zsh` compatibility
Calls like this one in the `Makefile` fail in `zsh` (the default shell for macOS), because of how `zsh` [interprets square brackets](https://stackoverflow.com/a/30539963):

https://github.com/pyro-ppl/numpyro/blob/f529fb0ecee621a1cb80439db4483c48c2cea135/Makefile#L17

Enclosing the argument in quotation marks should fix the issue without creating problems in other shells

```zsh
	pip install -e '.[dev,doc,test,examples]'
```

See this discussion for more details/examples: https://github.com/mu-editor/mu/issues/852

dylanhmorris:
Closes #1961

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,7 +31,7 @@ jobs:
         python -m pip install --upgrade pip
         pip install jaxlib
         pip install jax
-        pip install .[doc,test]
+        pip install '.[doc,test]'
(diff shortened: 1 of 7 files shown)


fehiepsi (approved):
Thanks, @dylanhmorris!

PR #1964, opened 2025-01-29
Add `ComplexTransform` and update `biject_to` registry.

tillahoffmann:
I added a `complex` constraint in #1762 with `biject_to` using the identity function as shown below. This makes inference challenging in practice because we end up with a mix of real and complex parameters.

https://github.com/pyro-ppl/numpyro/blob/93e11c2e43a99a450b2d857ff1f0a84ac1057f9e/numpyro/distributions/transforms.py#L1650-L1654

This PR adds a `ComplexTransform` which turns a real tensor of shape `(..., 2)` into a complex tensor of shape `(...)`. It also updates the `biject_to` registry such that complex variables can be integrated into models (e.g., for FFTs) while ensuring that the unconstrained space is real.

### DIFF
--- a/docs/source/distributions.rst
+++ b/docs/source/distributions.rst
@@ -910,6 +910,14 @@ CholeskyTransform
     :show-inheritance:
     :member-order: bysource
 
+ComplexTransform
+^^^^^^^^^^^^^^^^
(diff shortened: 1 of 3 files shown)


fehiepsi (approved):
LGTM, thank you!

PR #1968, opened 2025-02-01
chore(ci): disable continue-on-error for all test jobs in CI workflow

Qazalbash:
[`continue-on-error`](https://stackoverflow.com/a/73357322/26167473) was temporarily introduced in #1959 and accidentally got merged into the default branch. This PR reverts the commit (295136f12ffd185f110c58c669940ed0686224d8) introducing the temporary changes.

### DIFF
--- a/test/contrib/einstein/test_stein_loss.py
+++ b/test/contrib/einstein/test_stein_loss.py
@@ -4,13 +4,14 @@
 from numpy.testing import assert_allclose
 from pytest import fail
 
-from jax import numpy as jnp, random, value_and_grad
+from jax import numpy as jnp, random, value_and_grad, vmap
 from jax.scipy.special import logsumexp
 
 import numpyro
 from numpyro.contrib.einstein.stein_loss import SteinLoss
 from numpyro.contrib.einstein.stein_util import batch_ravel_pytree
 import numpyro.distributions as dist
+from numpyro.handlers import seed, substitute, trace
 from numpyro.infer import Trace_ELBO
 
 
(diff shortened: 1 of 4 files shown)


Qazalbash:
@fehiepsi Some more test cases have failed!

Can you check `test/contrib/einstein/test_stein_loss.py::test_stein_particle_loss`? It has a relatively large error.

fehiepsi:
@OlaRonning could you take a look? i think the inspected zs is different now.

OlaRonning:
I think you're right @fehiepsi. I'll work it over in detail  in the morning.

test/contrib/einstein/test_stein_loss.py
```diff
@@ -80,7 +80,7 @@ def stein_loss_fn(chosen_particle, obs, particles, assign):
     xs = jnp.array([-1, 0.5, 3.0])
     num_particles = xs.shape[0]
     particles = {"x": xs}
-    zs = jnp.array([-0.1241799, -0.65357316, -0.96147573])  # from inspect
+    zs = jnp.array([-3.3022664, -1.06049, 0.64527285])  # from inspect
```
fehiepsi, line 83: could we just replicate the logic to generate zs here? @OlaRonning

OlaRonning: Yeah, that should be possible. I'll have a look at it now.

OlaRonning:
The stein test is passing but there is a problem with stochastic support.

Qazalbash:
I fixed the very same problem for py3.10 in 78c51a702c4e0006dfb1bbc07cda7517277d4629, by changing the random seed, because the error was too high for the previous seed on which it was passing on py3.9. Now it is passing for py3.10 and not for py3.9. I presume we will encounter such cases in the future too! We have to find something more robust that will work on both Python versions.

OlaRonning:
Looking at the failing assert in CI it looks like the dimensions are swapped. Could be spurious or could be that the inference in the test case is nonidentifiable. *Note: I haven't check the test or method in detail.*

fehiepsi:
Yeah, we just need to assert whether the actual is close to expected or expected[::-1] (using np.is_close)

OlaRonning:
This seems like the simpler solution than making the toy problem identifiable :+1:  I'll add it.

fehiepsi (approved):
Thanks @Qazalbash and @OlaRonning!

Issue #1739, opened 2024-02-19
`AutoNormal`, `AutoDelta`, and `AutoGuideList` do not support subsamples of variable size.


tillahoffmann:
`AutoNormal`, `AutoDelta`, and `AutoGuideList` raise an exception in SVI when the subsample size varies across different `log_density` evaluation. Here is an example reproducing the issue (run on `master`).

```python
import numpyro
from jax import numpy as jnp


def model(n, x=None, subsample_size=None):
    mu = numpyro.sample("mu", numpyro.distributions.Normal())
    with numpyro.plate("n", n, subsample_size=subsample_size):
        numpyro.sample("x", numpyro.distributions.Normal(mu, 1), obs=x)


def demo(guide_cls):
    n = 10
    x_obs = jnp.zeros(n)
    guide = guide_cls(model)

    with numpyro.handlers.seed(rng_seed=0):
        # Initialize the guide with the full dataset, get a trace, and replay against
        # the model.
        guide(n, x_obs)
        guide_trace = numpyro.handlers.trace(guide).get_trace()
        replayed = numpyro.handlers.replay(model, guide_trace)

        print("evaluate log density for full data")
        numpyro.infer.util.log_density(replayed, (n, x_obs), {}, {})

        print("evaluate log density for subsampled data")
        numpyro.infer.util.log_density(replayed, (n, x_obs[:3], 3), {}, {})

        print("done")

# This works just fine.
demo(numpyro.infer.autoguide.AutoDiagonalNormal)
# This raises an error (see traceback below).
demo(numpyro.infer.autoguide.AutoNormal)
```

The traceback for the failed call is as follows.

```python
evaluate log density for full data
evaluate log density for subsampled data
.../numpyro/playground/test.py:7: UserWarning: subsample_size does not match len(subsample), 3 vs 10. Did you accidentally use different subsample_size in the model and guide?
  with numpyro.plate("n", n, subsample_size=subsample_size):
Traceback (most recent call last):
  File ".../numpyro/venv/lib/python3.10/site-packages/jax/_src/lax/lax.py", line 151, in broadcast_shapes
    return _broadcast_shapes_cached(*shapes)
  File ".../numpyro/venv/lib/python3.10/site-packages/jax/_src/util.py", line 287, in wrapper
    return cached(config.config._trace_context(), *args, **kwargs)
  File ".../numpyro/venv/lib/python3.10/site-packages/jax/_src/util.py", line 280, in cached
[...]

PR #1972, opened 2025-02-05
Do not create plates for observed sites in `AutoGuide`.

tillahoffmann:
This PR updates `_setup_prototype` to only record plates for unobserved sites. I *think* this leaves the intended behavior unchanged because we only sample unobserved sites in the guide. Because plates for observed sites are not created, the subsample size can vary across different invocations of `svi.update` or between `svi.init` and `svi.update` (cf. #1739).


fehiepsi (approved):
Thanks, @tillahoffmann!

PR #1976, opened 2025-02-14
Remove cached `plates` from `AutoGuide`.

tillahoffmann:
The `plate` primitive may leak tracers through its `_indices` attribute if created outside a pure function. The `AutoGuide` implementation caches plates in a `plates` attribute which results in a tracer leak when the guide is called (see example below).

This PR removes the caching such that calls to the guide are pure.

```python
>>> import jax
>>> import numpyro


>>> def model():
...     with numpyro.plate("n", 2):
...         numpyro.sample("x", numpyro.distributions.Normal())

>>> guide = numpyro.infer.autoguide.AutoDelta(model)

>>> numpyro.handlers.seed(guide, 9)()
>>> guide.plates["n"]._indices
Array([0, 1], dtype=int32)

>>> jax.jit(lambda: numpyro.handlers.seed(guide, 9)())()
>>> guide.plates["n"]._indices
Traced<ShapedArray(int32[2])>with<DynamicJaxprTrace>
```

### DIFF
--- a/numpyro/infer/autoguide.py
+++ b/numpyro/infer/autoguide.py
@@ -97,9 +97,9 @@ class AutoGuide(ABC):
         self._prototype_frames = {}
         self._prototype_frame_full_sizes = {}
 
-    def _create_plates(self, *args, **kwargs):
+    def _create_plates(self, *args, **kwargs) -> dict:
         if self.create_plates is None:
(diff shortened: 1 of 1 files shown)


fehiepsi (approved):
Thanks, Till!

PR #1978, opened 2025-02-16
Navigate to our forum for discussions

fehiepsi:
This change makes the forum more visible when users create a new issue.


OlaRonning (approved):
Good idea, @fehiepsi c:

PR #1979, opened 2025-02-17
Unify `sample_posterior()` signatures

tare:
This PR addresses [this](https://github.com/pyro-ppl/numpyro/pull/1644) comment. That is, we unify the signatures of `sample_posterior()` methods of the current automatic guides so that all the automatic guides have the more generic version now:
```python
def sample_posterior(self, rng_key, params, *args, sample_shape=(), **kwargs):
```

Sorry for dropping the ball on this.


fehiepsi (approved):

fehiepsi:
Thanks, @tare!

PR #1988, opened 2025-02-26
Add `CirculantNormal` distribution.

tillahoffmann:
This PR adds a `CirculantNormal` distribution which is a multivariate normal distribution where the covariance matrix has circular boundary conditions. It uses the Fourier transform introduced in #1762 to evaluate the log probability, scaling with `n * log(n)` instead of the usual `n ** 3`. The PR

- Implements the `CirculantNormal` distribution.
- Adds a `PackRealFastFourierCoefficientsTransform` transform. This is required to transform a real vector of size `n` to a complex vector of size `n // 2 + 1` which can be passed to the `rfft` method. See https://github.com/pyro-ppl/numpyro/pull/1762#issuecomment-2002020501 for a brief initial discussion.
- Adds a `positive_definite_circulant_vector` constraint. This is really just required for the `test/test_distributions.py::test_distribution_constraints` tests to pass.
- Adds an example notebook, illustrating how the `CirculantNormal` can be used to accelerate Gaussian process inference.
- Fix the Jacobian of the Fourier transform (which turns out not to be the identity but depends on the `norm` parameter).


notebooks/source/circulant_gp.ipynb
```diff
@@ -0,0 +1,293 @@
+{
```
juanitorduz, line 1: <p>It would be nice to put the legends outside the plots as in <a href="https://stackoverflow.com/questions/4700614/how-to-put-the-legend-outside-the-plot" target="_blank">https://stackoverflow.com/questions/4700614/how-to-put-the-legend-outside-the-plot</a> :)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pyro-ppl/numpyro/pull/1988/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

numpyro/distributions/continuous.py
```diff
@@ -3068,3 +3070,106 @@ def entropy(self) -> ArrayLike:
...
+        assert jnp.ndim(loc) > 0, "Location parameter must have at least one dimension."
+        n = jnp.shape(loc)[-1]
+        assert_one_of(covariance_row=covariance_row, covariance_rfft=covariance_rfft)
+
+        # Evaluate covariance_rfft if not provided and validate.
+        if covariance_rfft is None:
+            assert covariance_row.shape[-1] == n
+            covariance_rfft = jnp.fft.rfft(covariance_row).real
+            shape = jnp.broadcast_shapes(loc.shape, covariance_row.shape)
+            self.covariance_row = jnp.broadcast_to(covariance_row, shape)
```
tillahoffmann, line 3111: I'm broadcasting to the full shape here instead of just promoting the shapes. That's to guarantee that `covariance_row` always has shape `batch_shape + event_shape`. If we only promoted shapes, we might end up with a shape `(1, 1, n)` if the input `covariance_row` has shape `(n,)` but the input `loc` has shape `(a, b, n)`.

As an aside, this may also be relevant to `MultivariateNormal` distribution where `covariance_matrix` may not have the right batch dimensions.

```python
>>> from numpyro.distributions import MultivariateNormal
>>> 
>>> d = MultivariateNormal(jnp.zeros((3, 4, 5)), jnp.eye(5))
>>> d
<numpyro.distributions.continuous.MultivariateNormal object at 0x145cc7d00 with 
batch shape (3, 4) and event shape (5,)>
>>> d.covariance_matrix.shape
(1, 1, 5, 5)  # Expected (3, 4, 5, 5)
```

tillahoffmann:
I think the failing tests are due to issues with the data download and not related to the changes in this PR.

numpyro/distributions/continuous.py
```diff
@@ -3068,3 +3070,106 @@ def entropy(self) -> ArrayLike:
         return jnp.broadcast_to(
             0.5 + 1.5 * jnp.euler_gamma + 0.5 * jnp.log(16 * jnp.pi), self.batch_shape
         ) + jnp.log(self.scale)
+
+
+class CirculantNormal(TransformedDistribution):
+    """
+    Multivariate normal distribution with circulant covariance matrix.
```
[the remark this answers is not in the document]
tillahoffmann: I've added a bit more background and a reference. I unfortunately couldn't find a non-paywalled link.

numpyro/distributions/continuous.py
tillahoffmann: Makes sense. I've updated the shape promotion but needed to slightly modify the `promote_shapes` function to get shapes (rather than arrays with promoted shapes) because `loc` and `covariance_rfft` are not broadcastable along the last dimension.

tillahoffmann:
There are still a few failing numerical tests. Is it possible the `gof` tests are a bit strict? E.g., increasing the sample size to 20k for all distributions results in the following tests failing locally.

```
FAILED test/test_gof.py::test_gof[GaussianRandomWalk-None-params32] - assert 0.0 > 2e-05
FAILED test/test_gof.py::test_gof[GaussianRandomWalk-None-params33] - assert 0.0 > 2e-05
FAILED test/test_gof.py::test_gof[GaussianStateSpace-None-params35] - assert 0.0 > 2e-05
FAILED test/test_gof.py::test_gof[GaussianCopulaBeta-None-params37] - assert 1.623809683759525e-05 > 2e-05
FAILED test/test_gof.py::test_gof[MultivariateStudentT-_multivariate_t_to_scipy-params97] - assert 1.0300558915599112e-05 > 2e-05
FAILED test/test_gof.py::test_gof[VonMises-<lambda>-params161] - assert 0.0 > 2e-05
```

numpyro/distributions/continuous.py
```diff
@@ -3068,3 +3071,128 @@ def entropy(self) -> ArrayLike:
         return jnp.broadcast_to(
             0.5 + 1.5 * jnp.euler_gamma + 0.5 * jnp.log(16 * jnp.pi), self.batch_shape
         ) + jnp.log(self.scale)
+
+
+class CirculantNormal(TransformedDistribution):
+    """
+    Multivariate normal distribution with covariance matrix that is positive-definite
+    and circulant [1], i.e., has periodic boundary conditions. Circulant matrices can be
```
fehiepsi, line 3079: I think I understand what is a circulant matrix (from the wikipedia page) but I don't understand why this is a MVN distribution (it seems to be a transformed normal distribution based on the implementation) and why this matrix is called covariance matrix, and what "positive-definite" means here. Could you clarify?

It would be helpful to add formula in the docstrings like https://num.pyro.ai/en/stable/distributions.html#numpyro.distributions.continuous.Levy

fehiepsi (approved):
The implementation looks great to me! Same for the tutorial. Apart from adding more description to the definition of the distribution, could you also add a thumbnail for it https://github.com/pyro-ppl/numpyro/tree/master/docs/source/_static/img/tutorials

fehiepsi:
> Is it possible the gof tests are a bit strict?

I'm not sure. It seems not robust to multivariate distributions.

tillahoffmann:
> The implementation looks great to me! Same for the tutorial. Apart from adding more description to the definition of the distribution, could you also add a thumbnail for it https://github.com/pyro-ppl/numpyro/tree/master/docs/source/_static/img/tutorials

Ah, I forgot the thumbnail. Will add that in a future PR.

Issue #1987, opened 2025-02-24
[FR] Update the flax module to adapt flax new nnx api


sejabs:
### Feature Summary
The fully featured net library Flax has launch a new simplified NNX api to "make it easier to create, inspect, debug, and analyze neural networks in JAX". However, numpyro still use old flax Linen api in the falx module.  It is necessary to upgrade this module to keep up with the development of Flax (and may deprecate Haiku module).

### Why is this needed?
It seems the mainstream net libraries based on JAX have converged to Flax NNX (merged with Haiku), and in the long term the maintenance of old flax Linen api has uncertainty. So, this feature has 2 advantages:
1. keep up with the development of JAX net ecosystem;
2. simplify numpyro code base, i.e.  consider merging flax and haiku module.

juanitorduz:
Indeed, this is related to https://github.com/pyro-ppl/numpyro/issues/1709, but I have not had the bandwidth to push this forward. In that other issue there are some hints on how to potentially get started 💪

PR #1990, opened 2025-02-28
Flax NNX integration

juanitorduz:
Closes https://github.com/pyro-ppl/numpyro/issues/1987

I gave it a go to this issue with the help of Claude Sonnet 3.7 to get started and work on iterations to make it "work". So all feedback is welcome.


juanitorduz:
There is a test failing that is unrelate with this change 

```python
FAILED test/infer/test_autoguide.py::test_laplace_approximation_warning - Failed: DID NOT WARN. No warnings of type (<class 'UserWarning'>,) were emitted.
 Emitted warnings: [].
```

numpyro/contrib/module.py
```diff
@@ -429,3 +431,381 @@ def random_haiku_module(
...
+        neural network.
+    :param list apply_rng: A list to indicate which extra rng _kinds_ are needed for
+        ``nn_module``. Defaults to None, which means no extra rng key is needed.
+    :param list mutable: A list to indicate mutable states of ``nn_module``. For example,
+        if your module has BatchNorm layer, we will need to define ``mutable=["batch_stats"]``.
+    :param kwargs: optional keyword arguments to initialize NNX neural network
+        as an alternative to `input_shape`
+    :return: a callable that takes an array as an input and returns
+        the neural network transformed output array.
+    """
```
fehiepsi, line 702: I think you can reuse `_update_params` together with https://flax.readthedocs.io/en/latest/guides/surgery.html as in `random_haiku_module`.

numpyro/contrib/module.py
```diff
@@ -429,3 +431,381 @@ def random_haiku_module(
...
+    *args,
+    input_shape=None,
+    apply_rng=None,
+    mutable=None,
+    **kwargs,
+):
+    """
+    A primitive to create a random :mod:`~flax.nnx` style neural network
+    which can be used in MCMC samplers. The parameters of the neural network
+    will be sampled from ``prior``.
```
fehiepsi, line 681: Could you include an example as in `random_haiku_module`?

numpyro/contrib/module.py
```diff
@@ -429,3 +431,381 @@ def random_haiku_module(
...
+                # Convert to our safe version
+                rngs_dict = {key: getattr(user_rngs, key) for key in user_rngs._fields}
+                rngs = SafeRngs(**rngs_dict)
+            else:
+                rngs = SafeRngs(params=rng_key)
+        else:
+            rngs = SafeRngs(params=rng_key)
+
+        # Create a new module instance
+        model = nn_module(*args, rngs=rngs, **kwargs)
```
fehiepsi, line 596: I think for `nnx`, the module is initialized before providing to numpyro.module. Hence, to update params, you can do something like
```
graph_def, state = nnx.split(nn_module)
nnx.replace_by_pure_dict(state, process_raw_dict(module_params))
nn_module = nnx.merge(graph_def, state)
```
like in https://flax.readthedocs.io/en/latest/guides/surgery.html

fehiepsi (commented):
Thanks @juanitorduz! I left a few comments. I think you can use https://flax.readthedocs.io/en/latest/guides/filters_guide.html#the-filter-dsl to extract (params, rngs, mutable) stuffs. And https://flax.readthedocs.io/en/latest/guides/surgery.html to replace params/state.

(I also feel that rng is a mutable state under nnx https://flax.readthedocs.io/en/latest/guides/randomness.html)

juanitorduz:
Thanks for the comments @fehiepsi I will look into them 🙏

numpyro/contrib/module.py
```diff
@@ -429,3 +431,381 @@ def random_haiku_module(
...
+        from flax import nnx
+    except ImportError as e:
+        raise ImportError(
+            "Looking like you want to use flax.nnx to declare "
+            "nn modules. This is an experimental feature. "
+            "You need to install the latest version of `flax` to use this feature. "
+            "It can be installed with `pip install git+https://github.com/google/flax.git`."
+        ) from e
+
+    # Add a patch for the nnx.dropout function used in tests
```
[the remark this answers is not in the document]
juanitorduz: Removed in https://github.com/pyro-ppl/numpyro/pull/1990/commits/29b9da0cef5ed60b182698edf3772fef7690e779

fehiepsi (commented):
Hi @juanitorduz, I marked this as milestone 0.18 but there is no need to rush. Looks at the implementation, it seems to me that it's painful to deal with nnx (surprising? not sure why flax team is switching to use it given the complexity). If possible, try to come up with a solution similar to flax/haiku. If complicated, then it's not worth to support.

I would suggest to reach out to nnx devs if there's something unclear. Two main questions are
+ how to get the initial parameters and the initial mutable states of nn module
+ given parameters and states, how to get output and the updated states

We need params stored as a nested dict. For the state, if it is not a dict, you can just store as a dict via `{"state": state}`.

juanitorduz:
Thanks @fehiepsi ! Indeed it has been harder than expected, but it also reflects the fact my little experience with the NNX internals 😄.

My aim is to simplify indeed! I'll  use this branch to push my progress (experiments), ani I'll let you know whenever I feel really blocked or whenever I need a review. I will also try to reach the nnx team for some tips :) 

It would be super nice to have this integration to make sure the mid-long term integration with the neural network JAX community :)

juanitorduz:
@fehiepsi I simplified the code a lot (It was a mess, I know 🙈 ), I think it is now more modular (thank to your advice) and looks similar as the flax case. In https://github.com/pyro-ppl/numpyro/pull/1990/commits/2b8956695f2051ddf83b8ac5b7d2712f4661cef4 you can see all tests pass (I had to write some custom code to make this work for Python 3.9). I think it is in a much better shape if you wanna take a look.

Moreover, I am working on a notebook example (wip) adapting a simple example (https://juanitorduz.github.io/flax_numpyro/) and showing how to use NNX + NumPyro. I hope to clean the notebook and add the relevant content this weekend :)

test/contrib/test_module.py
```diff
@@ -311,3 +316,902 @@ def model():
...
+            self.weight = nnx.Param(jnp.zeros((din, dout)))
+            self.bias = nnx.Param(jnp.zeros(dout))
+
+        def __call__(self, x):
+            return x @ self.weight + self.bias
+
+    # Define a model that uses the module
+    def model(x, y=None):
+        # Create the module
+        nn = nnx_module("nn", SimpleModule, din=3, dout=1)
```
fehiepsi, line 1149: I think `nnx` encourages "eager initialization" (its main selling point). So it is better to initialize the module outside of numpyro `model`. Something like
```
module = SimpleModule(...)

def model(x, y=None):
    nn = nnx_module("nn", module)
```
With nnx, you can inspect the module directly so lots of code can be simplified. For example, you can do
```
_, params, state = nnx.split(module, nnx.Param, nnx.filterlib.to_predicate(nnx.Not(nnx.Param)))
```
to extract params and mutable state (which stores rngs, so we don't need numpyro to pass rng to it, yay!)

numpyro/contrib/module.py
```diff
@@ -433,3 +435,242 @@ def random_haiku_module(
...
+            ".".join(str(p) for p in path): var.value
+            for path, var in flat_params.items()
+        }
+
+        if params:
+            module_params = numpyro.param(module_key, params)
+
+        # Extract mutable state if needed
+        if mutable:
+            state_dict = {}
```
fehiepsi, line 523: why you need to get state_dict here? I think simply store in a dict {"state": state} is enough because numpyro.mutable support pytree values.

numpyro/contrib/module.py
```diff
@@ -433,3 +435,242 @@ def random_haiku_module(
...
+            flat_other_state = dict(other_state.flat_state())
+
+            for m in mutable:
+                if m == "batch_stats":
+                    # Extract batch stats
+                    batch_stats = {}
```
fehiepsi, line 529: why you need to deal with batch_stats? I think you can remove the `mutable` argument and all custom logic here.

numpyro/contrib/module.py
```diff
@@ -433,3 +435,242 @@ def random_haiku_module(
         _update_params(params, new_params, prior)
     nn_new = partial(nn.func, new_params, *nn.args[1:], **nn.keywords)
     return nn_new
+
+
+def _update_state_with_params(state, params_dict):
```
fehiepsi, line 440: You can use `nnx.replace_by_pure_dict(...)` to update the params and the mutable states.

numpyro/contrib/module.py
```diff
@@ -433,3 +435,242 @@ def random_haiku_module(
...
+        graph_def, params_state, other_state = nnx.split(
+            nn_module, nnx.Param, nnx.filterlib.to_predicate(nnx.Not(nnx.Param))
+        )
+
+        # Convert params to a dictionary for numpyro.param
+        flat_params = dict(params_state.flat_state())
+        params = {
+            ".".join(str(p) for p in path): var.value
+            for path, var in flat_params.items()
+        }
```
fehiepsi, line 516: it might be better to use `nnx.to_pure_dict(...)` so that params has the same structure as in flax and haiku. Then you dont need to have separate logic for the random_nnx_module.

fehiepsi:
> Moreover, I am working on a notebook example (wip) adapting a simple example (https://juanitorduz.github.io/flax_numpyro/) and showing how to use NNX + NumPyro. I hope to clean the notebook and add the relevant content this weekend :)

Nice tutorial, looking forward to seeing the nnx version! :)

I think the implementation could be even simpler. Let's see. ;)

juanitorduz:
Thanks for the feedback @fehiepsi ! I will try to simplify even more based on your input 🙏

numpyro/contrib/module.py
juanitorduz: Yes! I did not know about this functions ! XD I was doing everything by hand 🤦 ... added in https://github.com/pyro-ppl/numpyro/pull/1990/commits/716916861122c745afdfea8f2607efa6b778e183

numpyro/contrib/module.py
juanitorduz: Yes! Done https://github.com/pyro-ppl/numpyro/pull/1990/commits/716916861122c745afdfea8f2607efa6b778e183

numpyro/contrib/module.py
juanitorduz: Removed!

numpyro/contrib/module.py
juanitorduz: You are right! I removed the whole mutable logic

test/contrib/test_module.py
juanitorduz: yes! added this!

juanitorduz:
@fehiepsi Thank you for all your tips! The code is now super short and clear. I was doing everything by hand and initially thinking I had to initialize the NNX object inside the model 🤦 . All tests are passing locally 🟢 , but on the CI it fails because a Python 3.9 error :(

fehiepsi:
I would suggest not supporting python 3.9 for new features. jax dropped supports for 3.9. Same for flax. We maintain python 3.9 compatible just to help not breaking users' current systems.

numpyro/contrib/module.py
```diff
@@ -433,3 +435,143 @@ def random_haiku_module(
...
+            "Looking like you want to use flax.nnx to declare "
+            "nn modules. This is an experimental feature. "
+            "You need to install the latest version of `flax` to use this feature. "
+            "It can be installed with `pip install git+https://github.com/google/flax.git`."
+        ) from e
+
+    module_key = name + "$params"
+    module_params = numpyro.param(module_key)
+
+    if module_params is None:
```
fehiepsi, line 474: I think you can remove this `if` and `module_params = numpyro.param(module_key)` above

numpyro/contrib/module.py
```diff
@@ -433,3 +435,143 @@ def random_haiku_module(
...
+        )
+
+        params = nnx.to_pure_dict(params_state)
+
+        if params:
+            module_params = numpyro.param(module_key, params)
+
+        state_dict = {"state": params_state}
+
+        if any(state_dict.values()):
```
fehiepsi, line 486: Use similar logic as eager_params above, but for eager_other_state

numpyro/contrib/module.py
```diff
@@ -433,3 +435,143 @@ def random_haiku_module(
...
+            "nn modules. This is an experimental feature. "
+            "You need to install the latest version of `flax` to use this feature. "
+            "It can be installed with `pip install git+https://github.com/google/flax.git`."
+        ) from e
+
+    module_key = name + "$params"
+    module_params = numpyro.param(module_key)
+
+    if module_params is None:
+        graph_def, params_state, other_state = nnx.split(
```
fehiepsi, line 475: nit: better to use `eager_params_state, eager_other_state` here.

numpyro/contrib/module.py
```diff
@@ -433,3 +435,143 @@ def random_haiku_module(
...
+            nn_module, nnx.Param, nnx.filterlib.to_predicate(nnx.Not(nnx.Param))
+        )
+
+        params = nnx.to_pure_dict(params_state)
+
+        if params:
```
fehiepsi, line 481:
```
module_params = None
if eager_params:
    module_params = numpyro.param(name + "$params")
    if module_params is None:
        module_params = numpyro.param(name + "$params", eager_params)
```

numpyro/contrib/module.py
```diff
@@ -433,3 +435,143 @@ def random_haiku_module(
...
+        if any(state_dict.values()):
+            _ = numpyro_mutable(name + "$state", state_dict)
+
+    def apply_fn(*call_args, **call_kwargs):
+        user_rngs = call_kwargs.pop("rngs", None)
+
+        model = nn_module
+
+        if module_params:
+            graph_def, state = nnx.split(model)
```
fehiepsi, line 495: you need to merge the latest params and mutable state here. Something like
```
if module_params:
    nnx.replace_by_pure_dict(state, module_params)
if state_dict:
    nnx.replace_by_pure_dict(state, state_dict)
```

numpyro/contrib/module.py
```diff
@@ -433,3 +435,143 @@ def random_haiku_module(
...
+    # Extract params and state directly from the module in one go
+    graph_def, params_state, _ = nnx.split(
+        nn_module, nnx.Param, nnx.filterlib.to_predicate(nnx.Not(nnx.Param))
+    )
+
+    param_dict = nnx.to_pure_dict(params_state)
+
+    numpyro.deterministic(f"{name}$params", param_dict)
+
+    def apply_fn(x, *fn_args, **fn_kwargs):
```
fehiepsi, line 570: why we need `x` here?

numpyro/contrib/module.py
```diff
@@ -433,3 +435,143 @@ def random_haiku_module(
...
+
+        if params:
+            module_params = numpyro.param(module_key, params)
+
+        state_dict = {"state": params_state}
+
+        if any(state_dict.values()):
+            _ = numpyro_mutable(name + "$state", state_dict)
+
+    def apply_fn(*call_args, **call_kwargs):
```
fehiepsi, line 489: maybe introduce `params` as an argument, and return `partial(apply_fn, module_params)` here?

numpyro/contrib/module.py
```diff
@@ -433,3 +435,143 @@ def random_haiku_module(
...
+    numpyro.deterministic(f"{name}$params", param_dict)
+
+    def apply_fn(x, *fn_args, **fn_kwargs):
+        # Create a new model with the sampled parameters
+        graph_def, state = nnx.split(nn_module)
+        nnx.replace_by_pure_dict(state, param_dict)
+        model = nnx.merge(graph_def, state)
+        return model(x, *fn_args, **fn_kwargs)
+
+    return apply_fn
```
fehiepsi, line 577: I don't see how you set priors here. It might be better to use an implementation similar to flax/haiku (so that we can have a single logic to update from priors). No need to inspect the model again I guess. Just get the params structure from the output of `nnx_module` call, call _update_params to update it with the samples from priors. Then return the partial function.

fehiepsi (commented):
Looks very clean! I leave a few minor comments. Thanks a lot, Juan! 💯

test/contrib/test_module.py
```diff
@@ -311,3 +314,535 @@ def model():
...
+            bias_val = self.bias.value if hasattr(self.bias, "value") else self.bias
+            return x @ w_val + bias_val
+
+    # Directly initialize with dimensions
+    input_dim = x.shape[0]
+    # Don't pass x directly to nnx_module's constructor
+    nn = nnx_module("nn", Linear, din=input_dim, dout=100)
+    mean = nn(x)
+    numpyro.sample("y", numpyro.distributions.Normal(mean, 0.1), obs=y)
+
```
fehiepsi, line 359: I guess this is for the old implementation.

test/contrib/test_module.py
```diff
@@ -311,3 +314,535 @@ def model():
...
+    def model():
+        # Use the pre-initialized module
+        nn = nnx_module("nn", net_module)
+
+        x = numpyro.sample("x", dist.Normal(0, 1).expand([4, 3]).to_event(2))
+
+        if dropout:
+            y = nn(x, rngs={"dropout": numpyro.prng_key()})
+        else:
+            y = nn(x)
```
fehiepsi, line 451: dropout key is a part of the mutable state, so using `nn(x)` should work for both cases

test/contrib/test_module.py
```diff
@@ -311,3 +314,535 @@ def model():
...
+
+        numpyro.deterministic("y", y)
+
+    with handlers.trace(model) as tr, handlers.seed(rng_seed=0):
+        model()
+
+    if batchnorm:
```
fehiepsi, line 458: `if batchnorm or dropout`. You can also check if the nn$state has the expected fields

test/contrib/test_module.py
```diff
@@ -311,3 +314,535 @@ def model():
...
+def test_random_nnx_module_mcmc(callable_prior):
+    from flax import nnx
+
+    class Linear(nnx.Module):
+        def __init__(self, din, dout, *, rngs):
+            self.w = nnx.Param(jax.random.uniform(rngs.params(), (din, dout)))
+            self.b = nnx.Param(jnp.zeros((dout,)))
+
+        def __call__(self, x):
+            # Handle different Python versions by accessing the value attribute if needed
```
fehiepsi, line 480: Do you mean different flax releases? You can target the latest release.

test/contrib/test_module.py
```diff
@@ -311,3 +314,535 @@ def model():
...
+    nuts_kernel = NUTS(model)
+    mcmc = MCMC(nuts_kernel, num_warmup=2, num_samples=2, progress_bar=False)
+    mcmc.run(random.PRNGKey(0), data, labels)
+    samples = mcmc.get_samples()
+    assert "nn$params" in samples
+    assert "w" in samples["nn$params"]
+    assert "b" in samples["nn$params"]
+
+
+def test_nnx_transformer_module():
```
fehiepsi, line 530: maybe an unnecessary test

test/contrib/test_module.py
```diff
@@ -311,3 +314,535 @@ def model():
...
+    params = samples["transformer$params"]
+    assert "kernel" in params["query"]
+    assert "kernel" in params["key"]
+    assert "kernel" in params["value"]
+    assert params["query"]["kernel"].shape[0] == 2  # num_samples
+    assert params["key"]["kernel"].shape[0] == 2
+    assert params["value"]["kernel"].shape[0] == 2
+
+
+def test_nnx_cnn_module():
```
fehiepsi, line 702: this test is also unnecessary.

juanitorduz:
Thanks for all the comments @fehiepsi ! I will tackle them in the next days 💪 ! Have a great weekend!

numpyro/contrib/module.py
juanitorduz: done in https://github.com/pyro-ppl/numpyro/pull/1990/commits/e27c36864cbc9233c60d074bc4930ddac36dc442

numpyro/contrib/module.py
```diff
@@ -433,3 +435,143 @@ def random_haiku_module(
...
+        if params:
+            module_params = numpyro.param(module_key, params)
+
+        state_dict = {"state": params_state}
+
+        if any(state_dict.values()):
+            _ = numpyro_mutable(name + "$state", state_dict)
+
+    def apply_fn(*call_args, **call_kwargs):
+        user_rngs = call_kwargs.pop("rngs", None)
```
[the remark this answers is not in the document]
juanitorduz: indeed, removed in https://github.com/pyro-ppl/numpyro/pull/1990/commits/e27c36864cbc9233c60d074bc4930ddac36dc442

numpyro/contrib/module.py
juanitorduz: removed in https://github.com/pyro-ppl/numpyro/pull/1990/commits/e27c36864cbc9233c60d074bc4930ddac36dc442

numpyro/contrib/module.py
juanitorduz: added in https://github.com/pyro-ppl/numpyro/pull/1990/commits/6cd66c3a607ac10b321dbd1c823255822eef8624

numpyro/contrib/module.py
juanitorduz: Added in https://github.com/pyro-ppl/numpyro/pull/1990/commits/6cd66c3a607ac10b321dbd1c823255822eef8624

numpyro/contrib/module.py
juanitorduz: Do you mean like 

```python
    eager_other_params = nnx.to_pure_dict(eager_other_state)
    if eager_other_params:
        module_other_params = numpyro.param(name + "$params")
    if module_other_params is None:
        module_other_params = numpyro.param(name + "$params", eager_other_params)
``` 

?

numpyro/contrib/module.py
juanitorduz: done!

numpyro/contrib/module.py
```diff
@@ -433,3 +435,137 @@ def random_haiku_module(
...
+
+    eager_params = nnx.to_pure_dict(eager_params_state)
+
+    module_params = None
+    if eager_params:
+        module_params = numpyro.param(name + "$params")
+    if module_params is None:
+        module_params = numpyro.param(name + "$params", eager_params)
+
+    state_dict = {"state": eager_params_state}
```
fehiepsi, line 483: I dont have opinion on the name of "mutable state". You can call it "other state" if you prefer. This state dict should hold values of the mutable state, and it will be stored / updated during running svi, specifically after executing the module in the apply function below.

First, we need to check if the mutable state is non-empty by checking its eager value (eager_other_state). Then retrieve its "current" value numpyro.mutable(name). If its current value is not set, then we will set it to the eager value. This logic also applies for params.

numpyro/contrib/module.py
```diff
@@ -433,3 +435,137 @@ def random_haiku_module(
...
+
+        if module_params:
+            graph_def, state = nnx.split(model)
+
+            if module_params:
+                nnx.replace_by_pure_dict(state, module_params)
+
+            model = nnx.merge(graph_def, state)
+
+        return model(*call_args, **call_kwargs)
```
fehiepsi, line 499: After the model call, the mutable state will be updated. We should update the state_dict to its updated value. Something like (yes, we are splitting the model again after the call).
```
_, new_mutable_state = nnx.split(model, ...)
state_dict[name] = new_mutable_state
```
This will work as long as state_dict is the value retrieved from numpyro.mutable calls above

numpyro/contrib/module.py
fehiepsi: See my other comment on the role of mutable state during svi run.

numpyro/contrib/module.py
fehiepsi: that's right

numpyro/contrib/module.py
juanitorduz: Hi! Does something like https://github.com/pyro-ppl/numpyro/pull/1990/commits/048c253f7af4b5c52ea912ea55d8cb5ea3f26d14 is whats expected?

numpyro/contrib/module.py
juanitorduz: ok! I tried it out here https://github.com/pyro-ppl/numpyro/pull/1990/commits/048c253f7af4b5c52ea912ea55d8cb5ea3f26d14,  am I going in the right direction?

test/contrib/test_module.py
juanitorduz: yes, I will probably remove it at the and, I will keep it for dev and debugging

test/contrib/test_module.py
juanitorduz: yes, I will probably remove it at the and, I will keep it for dev and debugging

test/contrib/test_module.py
juanitorduz: no, it was the python 3.9 issue, cleaned it in https://github.com/pyro-ppl/numpyro/pull/1990/commits/1716f2d66f6af01b3ca99c9dbcf82032ef729e3b

test/contrib/test_module.py
juanitorduz: yes, done in https://github.com/pyro-ppl/numpyro/pull/1990/commits/1716f2d66f6af01b3ca99c9dbcf82032ef729e3b

test/contrib/test_module.py
juanitorduz: yes! 🙏  removed it! https://github.com/pyro-ppl/numpyro/pull/1990/commits/1716f2d66f6af01b3ca99c9dbcf82032ef729e3b

numpyro/contrib/module.py
juanitorduz: of course 🤦 ! Here is my first attempt https://github.com/pyro-ppl/numpyro/pull/1990/commits/4fe9f1133e06acd40017ec0403deba3b4a575aa5 wdyt?

numpyro/contrib/module.py
juanitorduz: I am unsure where this goes after https://github.com/pyro-ppl/numpyro/pull/1990/commits/048c253f7af4b5c52ea912ea55d8cb5ea3f26d14 🙏

numpyro/contrib/module.py
```diff
@@ -433,3 +435,134 @@ def random_haiku_module(
...
+    if module_params is None:
+        module_params = numpyro.param(name + "$params", eager_params_state_dict)
+
+    eager_other_state_dict = nnx.to_pure_dict(eager_other_state)
+
+    module_other_params = None
+    if eager_other_state:
+        module_other_params = numpyro.param(name + "$state")
+    if module_other_params is None:
+        module_other_params = numpyro.param(name + "$state", eager_other_state_dict)
```
fehiepsi, line 489: Maybe we are using terminologies differently. To me, "other params" is mutable params (which includes batch stats, dropout rngs, etc.), so we should not "optimize" them via `numpyro.param`.

What we need to do is to use `numpyro.mutable` for those "other params" (again, i would call it "mutable params", rather than "other params"). So what I suggested was
to replace `numpyro.param` by `numpyro.mutable` at those lines. Because we need to update the mutable state, an easy approach is to put it in a dictionary like `state_dict` (not `module_other_params` here). So the implementation is (I used the variable mutable_holder to avoid the confusion)
```
mutable_holder = None
if eager_other_state_dict:
    mutable_holder = numpyro.mutable(name + "$state")
if mutable_holder is None:
    numpyro.mutable(name + "$state",  {"state": eager_other_state_dict})
```

numpyro/contrib/module.py
fehiepsi: I think you can do
```
if params:
    nnx.replace_by_pure_dict(eager_params_state, params)
if mutable_holder:
    nnx.replace_by_pure_dict(eager_other_state, mutable_holder["state"])
model = nnx.merge(graph_def, eager_params_state, eager_other_state)
model_call = model(*call_args, **call_kwargs)
_, new_mutable_state = nnx.split(model, nnx.filterlib.to_predicate(nnx.Not(nnx.Param)))
if mutable_holder:
    mutable_holder["state"] = new_mutable_state
```

numpyro/contrib/module.py
```diff
@@ -433,3 +435,134 @@ def random_haiku_module(
...
+    except ImportError as e:
+        raise ImportError(
+            "Looking like you want to use flax.nnx to declare "
+            "nn modules. This is an experimental feature. "
+            "You need to install the latest version of `flax` to use this feature. "
+            "It can be installed with `pip install git+https://github.com/google/flax.git`."
+        ) from e
+
+    graph_def, eager_params_state, eager_other_state = nnx.split(
+        nn_module, nnx.Param, nnx.filterlib.to_predicate(nnx.Not(nnx.Param))
```
fehiepsi, line 472: It seems that `nnx.split(nn_module, nnx.Param, nnx.Not(nnx.Param))` is enough.

juanitorduz: simplified it https://github.com/pyro-ppl/numpyro/pull/1990/commits/1e0206b71008cec7cb56ef9f2ef26da2f4138167

numpyro/contrib/module.py
juanitorduz: ahg yes! this was my confusion indeed 🙏 , I added the `numpyro_mutable` in https://github.com/pyro-ppl/numpyro/pull/1990/commits/1e0206b71008cec7cb56ef9f2ef26da2f4138167

juanitorduz:
**Minor comment:** When re-running the notebook again I see

```python
0%|          | 0/20000 [00:00<?, ?it/s]/Users/juanitorduz/Documents/numpyro/numpyro/infer/elbo.py:253: UserWarning: mutable state is currently ignored when num_particles > 1.
```

Is this expected? (the results do not change tho)

numpyro/contrib/module.py
```diff
@@ -433,3 +435,132 @@ def random_haiku_module(
...
+            graph_def, state = nnx.split(model)
+
+            if params:
+                nnx.replace_by_pure_dict(state, params)
+
+            model = nnx.merge(graph_def, state)
+
+        model_call = model(*call_args, **call_kwargs)
+
+        _, _, new_mutable_state = nnx.split(model, nnx.Param, nnx.Not(nnx.Param))
```
fehiepsi, line 506: nit: no need to split with nnx.Param, Not(Param) is enough.

numpyro/contrib/module.py
```diff
@@ -433,3 +435,132 @@ def random_haiku_module(
...
+
+    state_dict = {"state": eager_other_state}
+
+    def apply_fn(params, *call_args, **call_kwargs):
+        model = nn_module
+
+        if params:
+            graph_def, state = nnx.split(model)
+
+            if params:
```
fehiepsi, line 499: See the comment above - you don't need to split the model again
```
params_state = eager_params_state
if params:
    nnx.replace_by_pure_dict(params_state, params)
other_state = eager_other_state
if mutable_holder:
    nnx.replace_by_pure_dict(other_state, mutable_holder["state"])
model = nnx.merge(graph_def, params_state, other_state)
```

numpyro/contrib/module.py
```diff
@@ -433,3 +435,132 @@ def random_haiku_module(
...
+    if module_params is None:
+        module_params = numpyro.param(name + "$params", eager_params_state_dict)
+
+    eager_other_state_dict = nnx.to_pure_dict(eager_other_state)
+
+    mutable_holder = None
+    if eager_other_state_dict:
+        mutable_holder = numpyro_mutable(name + "$state")
+    if mutable_holder is None:
+        numpyro_mutable(name + "$state", {"state": eager_other_state_dict})
```
fehiepsi, line 489: mutable_holder = numpyro_mutable(...)

numpyro/contrib/module.py
```diff
@@ -433,3 +435,132 @@ def random_haiku_module(
...
+            if params:
+                nnx.replace_by_pure_dict(state, params)
+
+            model = nnx.merge(graph_def, state)
+
+        model_call = model(*call_args, **call_kwargs)
+
+        _, _, new_mutable_state = nnx.split(model, nnx.Param, nnx.Not(nnx.Param))
+
+        state_dict[name] = new_mutable_state
```
fehiepsi, line 508: nnx.to_pure_dict(new_mutable_state)

fehiepsi (commented):
Yes, it is expected. It's not clear how mutable states are updated with num_particles > 1. For dropout rng, we need a special logic to vectorize split. For batch_stats, we need sequentially updates then aggregates. I don't know - it is tricky to think about a reasonable solution for the issue.

I would suggest to use `num_particles=1` if batch norm plays a key role in your network.

numpyro/contrib/module.py
```diff
@@ -433,3 +435,132 @@ def random_haiku_module(
...
+
+    eager_other_state_dict = nnx.to_pure_dict(eager_other_state)
+
+    mutable_holder = None
+    if eager_other_state_dict:
+        mutable_holder = numpyro_mutable(name + "$state")
+    if mutable_holder is None:
+        numpyro_mutable(name + "$state", {"state": eager_other_state_dict})
+
+    state_dict = {"state": eager_other_state}
```
fehiepsi, line 491: I don't think we want to use `state_dict`. We just want to update the mutable state via `mutable_holder` above and below.

numpyro/contrib/module.py
juanitorduz: fixed in https://github.com/pyro-ppl/numpyro/pull/1990/commits/18fe2a280a14b51556b4c054d7a53a6354a89d02

numpyro/contrib/module.py
juanitorduz: yeah, makes sense, fixed in https://github.com/pyro-ppl/numpyro/pull/1990/commits/18fe2a280a14b51556b4c054d7a53a6354a89d02

numpyro/contrib/module.py
juanitorduz: Fixed in https://github.com/pyro-ppl/numpyro/pull/1990/commits/18fe2a280a14b51556b4c054d7a53a6354a89d02

numpyro/contrib/module.py
juanitorduz: I tried this and got an error in the tests

```python
FAILED test/contrib/test_module.py::test_nnx_transformer_module - ValueError: Non-exhaustive filters, got a non-empty remainder
```

¯\\_(ツ)_/¯

fehiepsi: thanks, sorry for making confusion

numpyro/contrib/module.py
fehiepsi: You will need to use the latest mutable state, not just the latest params.

juanitorduz: mmm I am a bit confused: I tried this in https://github.com/pyro-ppl/numpyro/pull/1990/commits/17db8c14c20e393e98dd57edd700ce86d8217f36 and I get from the tests

```python
FAILED test/contrib/test_module.py::test_nnx_state_dropout_smoke[True-True] - TypeError: unsupported operand type(s) for *: 'float' and 'VariableState'
FAILED test/contrib/test_module.py::test_nnx_state_dropout_smoke[True-False] - TypeError: unsupported operand type(s) for *: 'float' and 'VariableState'
```

Is if ok that the `apply` function has access to parameters outside of it like `eager_params_state` and `mutable_holder`?

fehiepsi: I guess it is due to the line `    mutable_holder["state"] = eager_other_state` above. We can remove that line because we already have the logic to get the latest mutable state from the numpyro.mutable call.

numpyro/contrib/module.py
fehiepsi: Sorry, by above, I meant the logic to get the latest mutable state via numpyro.mutable call. By below, I meant the logic after model call. We should remove this line, rather than rename.

numpyro/contrib/module.py
juanitorduz: mmm I removed this line (https://github.com/pyro-ppl/numpyro/pull/1990/commits/5ffcfe30a35ffa6bc768f9aef7222e3bc5b55225) and keep seeing the tests failing. I wonder if there is something missing in https://github.com/pyro-ppl/numpyro/commit/17db8c14c20e393e98dd57edd700ce86d8217f36 🤔

juanitorduz: I wonder if we are updating the parameters more than needed in the apply. In https://github.com/pyro-ppl/numpyro/pull/1990/commits/12933f64a5680300577496967ea2b6252fae5993 I removed the last split and update just thinking we did the update just before calling the model. It seems the test with batch-normalization passes now. What do you think, or could the error be coming from somewhere else 🙏 ?

juanitorduz: I brought it back in https://github.com/pyro-ppl/numpyro/pull/1990/commits/d25de60929d3659aa1449706b3dc8eba1a0e925d because you have been pointing out this in many times... I can't figure out the error in the tests ...

juanitorduz:
@fehiepsi Thank you very much for all the fantastic feedback these last days! I have learned a lot! Here is a summary on where we are:

- All comments are addressed except for https://github.com/pyro-ppl/numpyro/pull/1990#discussion_r1986406275 which is about a failing test and maybe a hidden bug. I would love some feedback (see comments in the thread there).
- I have updated the example notebook to use both the `nnx_module` and the `random_nnx_module`. Both are working as expected 😄 ! (Unless I missed something).
- I also build the docs locally and the nnx function and the examples are displayed as expected 🙌 .

I think we are very close to the finish line (I hope XD)

numpyro/contrib/module.py
```diff
@@ -433,3 +435,134 @@ def random_haiku_module(
...
+            nnx.replace_by_pure_dict(mutable_state, mutable_holder["state"])
+
+        model = nnx.merge(graph_def, params_state, mutable_state)
+
+        model_call = model(*call_args, **call_kwargs)
+
+        _, _, new_mutable_state = nnx.split(model, nnx.Param, nnx.Not(nnx.Param))
+
+        if mutable_holder:
+            mutable_holder["state"] = new_mutable_state
```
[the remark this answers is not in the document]
juanitorduz: It seems that the `to_mutable_dict` does not exist (depreciation or new name?), I will look for alternatives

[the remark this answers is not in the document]
fehiepsi: Sorry, it is to_pure_dict that you used above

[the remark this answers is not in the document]
fehiepsi: Oh, you might need stop_gradient as in flax/haiku_module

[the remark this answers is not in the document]
juanitorduz: Bingo! https://github.com/pyro-ppl/numpyro/pull/1990/commits/0f6763afb06cfd9fd7f4fb563f29558d12ebee0c Now tests are 🟢 !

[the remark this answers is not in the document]
fehiepsi: Nice, things would be retrieved and got updated correctly now, I think. Thanks!

numpyro/contrib/module.py
```diff
@@ -433,3 +435,135 @@ def random_haiku_module(
...
+
+        mutable_state = eager_other_state
+        if mutable_holder:
+            nnx.replace_by_pure_dict(mutable_state, mutable_holder["state"])
+
+        model = nnx.merge(graph_def, params_state, mutable_state)
+
+        model_call = model(*call_args, **call_kwargs)
+
+        _, _, new_mutable_state = nnx.split(model, nnx.Param, nnx.Not(nnx.Param))
```
fehiepsi, line 507: nit: we can move this to the `if mutable_holder` block

test/contrib/test_module.py
```diff
@@ -311,3 +315,360 @@ def model():
     guide = AutoDelta(model)
     svi = SVI(model, guide, numpyro.optim.Adam(0.01), Trace_ELBO())
     svi.run(random.PRNGKey(100), 10)
+
+
+@pytest.mark.skipif(sys.version_info[:2] == (3, 9), reason="Skipping on Python 3.9")
+def nnx_model_by_shape(x, y):
```
fehiepsi, line 321: you can remove this test now I guess. Not sure why we can still construct the module lazily.

test/contrib/test_module.py
```diff
@@ -311,3 +315,360 @@ def model():
...
+    from flax import nnx
+
+    class Linear(nnx.Module):
+        def __init__(self, din, dout, *, rngs):
+            self.w = nnx.Param(jax.random.uniform(rngs.params(), (din, dout)))
+            self.b = nnx.Param(jnp.zeros((dout,)))
+
+        def __call__(self, x):
+            w_val = self.w.value
+            b_val = self.b.value
```
fehiepsi, line 456: just wonder why `.value` is needed? In the example, https://flax.readthedocs.io/en/latest/guides/transforms.html#basic-example it seems that we can use `self.w` directly.

test/contrib/test_module.py
fehiepsi: I think we can remove this now.

test/contrib/test_module.py
```diff
@@ -311,3 +315,360 @@ def model():
...
+    # Create a pre-initialized module for eager initialization
+    rng_key = random.PRNGKey(0)
+    linear_module = Linear(din=dim, dout=1, rngs=nnx.Rngs(params=rng_key))
+
+    # Extract parameters and state for inspection
+    _, params_state = nnx.split(linear_module, nnx.Param)
+    params_dict = {
+        ".".join(str(p) for p in path): param.value
+        for path, param in dict(params_state.flat_state()).items()
+    }
```
fehiepsi, line 481: nit: maybe using `nnx.to_pure_dict` here?

fehiepsi (approved):
The tutorials look awesome! I just catch some small nits. Thanks, Juan!

test/contrib/test_module.py
juanitorduz: Done https://github.com/pyro-ppl/numpyro/pull/1990/commits/e9d6b83cac59c2333c727c14bd0be2a5c8f277d8 !

test/contrib/test_module.py
juanitorduz: Done https://github.com/pyro-ppl/numpyro/pull/1990/commits/e9d6b83cac59c2333c727c14bd0be2a5c8f277d8 !

test/contrib/test_module.py
juanitorduz: Done https://github.com/pyro-ppl/numpyro/pull/1990/commits/e9d6b83cac59c2333c727c14bd0be2a5c8f277d8 !

numpyro/contrib/module.py
juanitorduz: Done https://github.com/pyro-ppl/numpyro/pull/1990/commits/e9d6b83cac59c2333c727c14bd0be2a5c8f277d8 !

juanitorduz:
Thanks @fehiepsi ! All of the remaining comments were addressed in Done https://github.com/pyro-ppl/numpyro/pull/1990/commits/e9d6b83cac59c2333c727c14bd0be2a5c8f277d8 ! 🙏

juanitorduz:
fyi @cgarciae

juanitorduz:
Thank you very much for your guidance (and patience) @fehiepsi !

numpyro/contrib/module.py
```diff
@@ -433,3 +435,134 @@ def random_haiku_module(
...
+    except ImportError as e:
+        raise ImportError(
+            "Looking like you want to use flax.nnx to declare "
+            "nn modules. This is an experimental feature. "
+            "You need to install the latest version of `flax` to use this feature. "
+            "It can be installed with `pip install git+https://github.com/google/flax.git`."
+        ) from e
+
+    graph_def, eager_params_state, eager_other_state = nnx.split(
+        nn_module, nnx.Param, nnx.Not(nnx.Param)
```
cgarciae, line 472: I know this is merged but if you want to slightly simplify this in the future you can use `...` (Ellipsis) to match anything else e.g:

```python
graphdef, params, rest = nnx.split(model, nnx.Param, ...)
```

Issue #1991, opened 2025-02-28
log_likelihood fails with random_flax_module (and flax_module)


kylejcaron:
### Bug Description
A clear and concise description of the bug.

Versions
- jax: '0.5.1', 
- numpyro: '0.17.0'
- flax: '0.10.4'

### Steps to Reproduce

```python
import numpy as np
import numpyro
import numpyro.distributions as dist
from numpyro.infer import MCMC, NUTS
from jax import random

import flax.linen as nn
from numpyro.contrib.module import random_flax_module

# Simulate
rng = np.random.default_rng(99)
N = 1000

X = rng.normal(0, 1, size=(N,1))
mu = 1 + X @ np.array([0.5])
y = rng.normal(mu, 0.5)

# Simple linear layer
class Linear(nn.Module):
    @nn.compact
    def __call__(self, x):
        return nn.Dense(1, use_bias=True, name='Dense')(x)

def model(X, y=None):
    sigma = numpyro.sample("sigma", dist.HalfNormal(0.1))
    priors = {"Dense.bias": dist.Normal(0,2.5), "Dense.kernel": dist.Normal(0,1)}
    mlp = random_flax_module("mlp", Linear(), prior=priors, input_shape=(X.shape[1],))
    with numpyro.plate("data", X.shape[0]):
        mu = numpyro.deterministic("mu", mlp(X).squeeze(-1))
        y = numpyro.sample("y", dist.Normal(mu, sigma), obs=y)

# Fit model
kernel = NUTS(model, target_accept_prob=0.95)
mcmc = MCMC(kernel, num_warmup=1000, num_samples=1000, num_chains=2)
mcmc.run(random.PRNGKey(0), X=X, y=y)

# run log likelihood
numpyro.infer.util.log_likelihood(model, mcmc.get_samples(), X=X, y=y)
```

Here's the traceback

<details>
  <summary>Click to expand Traceback</summary>

```python
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[431], line 46
     42 mcmc = MCMC(kernel, num_warmup=1000, num_samples=1000, num_chains=2)
     43 mcmc.run(random.PRNGKey(0), X=X, y=y)
---> 46 numpyro.infer.util.log_likelihood(model, mcmc.get_samples(), X=X, y=y)

[...]

PR #1992, opened 2025-03-01
Improve rng default in the `contrib/model.py` module

juanitorduz:
Closes https://github.com/pyro-ppl/numpyro/issues/1991

### DIFF
--- a/numpyro/contrib/module.py
+++ b/numpyro/contrib/module.py
@@ -85,6 +85,8 @@ def flax_module(
         # feed in dummy data to init params
         args = (jnp.ones(input_shape),) if input_shape is not None else args
         rng_key = numpyro.prng_key()
+        if rng_key is None:
+            rng_key = random.key(0)
         # split rng_key into a dict of rng_kind: rng_key
         rngs = {}
         if apply_rng:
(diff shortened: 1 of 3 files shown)


fehiepsi:
I think it is the issue of contrib.module. We can use
```
rng_key = numpyro.prng_key() or random.key(0)
```
to optionally create dummy rng_key. I think it does not affect the functionality of the module.

juanitorduz:
> 5e40979

Do you mean like https://github.com/pyro-ppl/numpyro/pull/1992/commits/5e409799f8f42aaed4520d8b2a14d0d4c697d2d9 ? or something has to be changed in `contrib/module.py`?

juanitorduz:
> I think it is the issue of contrib.module. We can use
> 
> ```
> rng_key = numpyro.prng_key() or random.key(0)
> ```
> 
> to optionally create dummy rng_key. I think it does not affect the functionality of the module.

Do you mean like https://github.com/pyro-ppl/numpyro/pull/1992/commits/5e409799f8f42aaed4520d8b2a14d0d4c697d2d9 ? or something has to be changed in `contrib/module.py`?

fehiepsi:
I think changing in contrib is better because we know that key(0) does not change the logic there. If we change it here, users might forget to provide some posterior samples.

numpyro/contrib/module.py
```diff
@@ -85,6 +85,8 @@ def flax_module(
         # feed in dummy data to init params
         args = (jnp.ones(input_shape),) if input_shape is not None else args
         rng_key = numpyro.prng_key()
+        if rng_key is None:
+            rng_key = random.key(0)
```
juanitorduz, line 89: I had to do it this way since `rng_key = numpyro.prng_key() or random.key(0)` we giving an error

```python
E       ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

fehiepsi (approved):
Thanks, Juan!

fehiepsi:
Could you merge with the master branch?

Issue #1986, opened 2025-02-20
Unnecessary computation in `soft_vmap()`


tare:
### Bug Description

I was confused why the map branch was executed
https://github.com/pyro-ppl/numpyro/blob/627d19a8e01e3d1b6c7a594ef7f49face4921f92/numpyro/util.py#L453
when I passed `parallel=True` to `Predictive()`.

It looks like in that case `num_chunks == 2` (should be `1`) because `pad ==  chunk_size` (should be `0`) even though `batch_size % chunk_size == 0`. This is the line that causes extra padding:
https://github.com/pyro-ppl/numpyro/blob/627d19a8e01e3d1b6c7a594ef7f49face4921f92/numpyro/util.py#L441

Please let me know whether my reasoning makes sense.

[This](https://github.com/tare/numpyro/commit/6e819864ee77bfe0ca38b03191e364551c425311) commit fixes the issue.

### Steps to Reproduce
N/A

### Expected Behavior
No padding when `batch_size % chunk_size == 0`.

fehiepsi:
Nice catch! Could you send the fix @tare? Thanks :)

PR #1993, opened 2025-03-01
Fix padding calculation

tare:
This PR fixes the padding calculation issue in `soft_vmap()` as described in https://github.com/pyro-ppl/numpyro/issues/1986.


fehiepsi (approved):
Thanks, @tare!

fehiepsi:
Could you rebase on the master branch @tare ?

PR #1994, opened 2025-03-03
Apply `init_to_median` in unconstrained space.

tillahoffmann:
The elementwise median of a random variable does not necessarily belong to the support of the generating distribution. I recently ran into this issue with `ZeroSumNormal` distributions where `init_to_median` gave samples that did not sum to zero.

This PR updates `init_to_median` to work in the unconstrained space, guaranteeing valid samples. It also fixes a small bug where `init_to_mean` does not return a value.

This change should not change the initialization given fixed random key for simple bijections to the support, but it will affect more complex transformations when `median(transform(x)) != transform(median(x))`.

### DIFF
--- a/numpyro/infer/initialization.py
+++ b/numpyro/infer/initialization.py
@@ -40,7 +40,13 @@ def init_to_median(site=None, num_samples=15):
             samples = site["fn"](
                 sample_shape=(num_samples,) + sample_shape, rng_key=rng_key
             )
-            return jnp.median(samples, axis=0)
+            from numpyro.infer.util import helpful_support_errors
+
+            with helpful_support_errors(site):
+                transform = biject_to(site["fn"].support)
+            unconstrained = transform.inv(samples)
+            median_unconstrained = jnp.median(unconstrained, axis=0)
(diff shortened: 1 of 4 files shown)


numpyro/infer/initialization.py
```diff
@@ -40,7 +40,13 @@ def init_to_median(site=None, num_samples=15):
             samples = site["fn"](
                 sample_shape=(num_samples,) + sample_shape, rng_key=rng_key
             )
-            return jnp.median(samples, axis=0)
+            from numpyro.infer.util import helpful_support_errors
```
fehiepsi, line 43: Just curious, you put the import here because of circular import?

fehiepsi (approved):

numpyro/infer/initialization.py
tillahoffmann: Yes, I borrowed that line from `init_to_uniform` here.

https://github.com/pyro-ppl/numpyro/blob/627d19a8e01e3d1b6c7a594ef7f49face4921f92/numpyro/infer/initialization.py#L108-L109

Issue #1981, opened 2025-02-19
Tracking leaked tracers.


tillahoffmann:
Following #1976, I ran the test suite with tracer leak checks enabled which raised just under two hundred exceptions (results below). Many of these are likely leaking in the test code itself as opposed to in the package code. I thought I'd share as it might point to bugs. Some of the exceptions show up as `IndexError` because of jax-ml/jax#26560.

<details><summary>Details</summary>
<p>

```
FAILED test/contrib/einstein/test_steinvi.py::test_run_smoke[ASVGD-uniform_normal-kernel0] - Exception: Leaked trace MainTrace(1,DynamicJaxprTrace). Leaked tracer(s):
FAILED test/contrib/einstein/test_steinvi.py::test_run_smoke[ASVGD-uniform_normal-kernel1] - Exception: Leaked trace MainTrace(1,DynamicJaxprTrace). Leaked tracer(s):
FAILED test/contrib/einstein/test_steinvi.py::test_run_smoke[ASVGD-uniform_normal-kernel2] - Exception: Leaked trace MainTrace(1,DynamicJaxprTrace). Leaked tracer(s):
FAILED test/contrib/einstein/test_steinvi.py::test_run_smoke[ASVGD-uniform_normal-kernel3] - Exception: Leaked trace MainTrace(1,DynamicJaxprTrace). Leaked tracer(s):
FAILED test/contrib/einstein/test_steinvi.py::test_run_smoke[ASVGD-uniform_normal-kernel4] - Exception: Leaked trace MainTrace(1,DynamicJaxprTrace). Leaked tracer(s):
FAILED test/contrib/einstein/test_steinvi.py::test_run_smoke[ASVGD-regression-kernel0] - Exception: Leaked trace MainTrace(1,DynamicJaxprTrace). Leaked tracer(s):
FAILED test/contrib/einstein/test_steinvi.py::test_run_smoke[ASVGD-regression-kernel1] - Exception: Leaked trace MainTrace(1,DynamicJaxprTrace). Leaked tracer(s):
FAILED test/contrib/einstein/test_steinvi.py::test_run_smoke[ASVGD-regression-kernel2] - Exception: Leaked trace MainTrace(1,DynamicJaxprTrace). Leaked tracer(s):
FAILED test/contrib/einstein/test_steinvi.py::test_run_smoke[ASVGD-regression-kernel3] - Exception: Leaked trace MainTrace(1,DynamicJaxprTrace). Leaked tracer(s):
[...]

Issue #1982, opened 2025-02-19
Tracking leaked in ASVGD


fehiepsi:
This is a sub-issue of #1981. We shouldn't set `self.foo` for a tracer foo: https://github.com/pyro-ppl/numpyro/blob/f7746d5afa3af1d4f8d0a3ed071cf13d357080f4/numpyro/contrib/einstein/steinvi.py#L756

Repro the issue:
```
JAX_CHECK_TRACER_LEAKS=1 pytest -vs test/contrib/einstein/test_steinvi.py
```

cc @OlaRonning

fehiepsi:
I looked at the code and don't understand how setting `self.loss_temperature = ...` works there. I feel that the current implementation is incorrect.

OlaRonning:
Sorry for not getting back on this. I started on a fix, but I have been traveling this last week. My approach is to add to the _svgs_loss_and_grad. I guess adding it to the stein state works as well.

I'll wrap it up in the afternoon.

PR #1997, opened 2025-03-06
Mark xfail for leaked tracer tests

fehiepsi:
This PR adds an entry in github action to track leaked tracer failing tests reported in https://github.com/pyro-ppl/numpyro/issues/1981. I found the behaviors under JAX_CHECK_TRACER_LEAKS is a bit different from the normal behaviors so just want to perform a subset of tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,6 +78,28 @@ jobs:
     - name: Test x64
       run: |
         JAX_ENABLE_X64=1 pytest -vs test/test_distributions.py -k powerLaw
+    - name: Test tracer leak
+      if: matrix.python-version == '3.10'
+      env:
+        JAX_CHECK_TRACER_LEAKS: 1
+      run: |
+        pytest -vs test/contrib/einstein/test_steinvi.py::test_run_smoke -k ASVGD
+        pytest -vs test/contrib/stochastic_support/test_sdvi.py::test_elbos -k TraceEnum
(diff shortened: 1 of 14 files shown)


fehiepsi:
It seems that some issues were fixed so I'll just include the non-passing tests in the PR.

tillahoffmann:
Looks great! Worth considering a custom marker for this or too much?

```python
# conftest.py
def pytest_runtest_setup(item):
    marker = item.get_closest_marker("xfail_tracer_leak")
    if marker is not None and os.environ.get("JAX_CHECK_TRACER_LEAKS") == "1":
        item.add_marker(pytest.mark.xfail(reason="Expected tracer leak."))


# test files
@pytest.mark.xfail_tracer_leak
def test_...():
    ...
```

fehiepsi:
Thanks for the suggestion! I prefer using pytest.mark it directly.

There are only 4 issues left (yay!), which I have created separate github issues for each of them.
- #1982 It is clear where is the leak here
- #2000 I'll need to think more how to address this issue. The caching logic is useful, but maybe it's not reliable under jit.
- #1999 Maybe not complicated, haven't thought about it yet.
- #1998 I spent a bit of time on this before but couldn't track down the issue :(

While debugging, I found that there are many live arrays over the tests so I take this chance to remove all of them (and add a small check to conftest to assure that no live arrays will be created globally).

tillahoffmann (approved):
Looks great!

PR #2002, opened 2025-03-07
Use functional interface funsor.adjoint.adjoint to avoid tracer leak

### LINKED ISSUE #1998: scan over discrete latent variables causes tracer leak
### Bug Description

This is a part of the issues reported in https://github.com/pyro-ppl/numpyro/issues/1981. Running the following test will raise an error/xfail.

### Steps to Reproduce

```
JAX_CHECK_TRACER_LEAKS=1 pytest -vs test/contrib/test_infer_discrete.py::test_scan_hmm_smoke
```

### Expected Behavior

The test should pass.

fehiepsi:
Fixes #1998

The current `infer_discrete` tests pass so I believe the change is equivalent.


fehiepsi:
It seems that the annotation example is failing with this.

juanitorduz (approved):
Moral approval 🤗  (shall we make this part of the 0.16.0 milestone?)

PR #2001, opened 2025-03-07
Fix tracer leak in svi with mutable state

fehiepsi:
Fixes #1999

PR #2003, opened 2025-03-09
Fix Memleak in ASVGD

OlaRonning:
`ASVGD` had a memory leak caused by assigning a JAX Tracer to an attribute. This PR resolves the issue by introducing a state tuple that is passed along with the optimization process instead.

### Changes:
1. Introduced ASVGDState to track the annealing state in ASVGD.
2. Extended SteinVIState to include temperature tracking.

### TODO
 - [x] Verify on internal examples
 - [x] Update docstrings
 - [x] ~~Add leak check to unit tests~~ Ignored following: #2001

Related to #1982

### DIFF
--- a/test/contrib/einstein/test_steinvi.py
+++ b/test/contrib/einstein/test_steinvi.py
@@ -3,7 +3,6 @@
 
 from collections import namedtuple
 from functools import partial
-import os
 import string
 
 import numpy as np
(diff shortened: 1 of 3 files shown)


test/contrib/einstein/test_steinvi.py
```diff
@@ -126,7 +126,14 @@ def model(features, labels):
 def test_run_smoke(kernel, problem, method):
```
OlaRonning, line 126: @fehiepsi do I remove the `mark.xfail` decorator?

``` python
@pytest.mark.xfail(
    os.getenv("JAX_CHECK_TRACER_LEAKS") == "1", reason="Expected tracer leak"
)
```

OlaRonning: I'm going to follow [this](https://github.com/pyro-ppl/numpyro/pull/2001/files#diff-d14f559f375bf09ec589feb7b9f6393838d4d5d767c5501c50c8db5c47da5dfcL523) and remove it.
Running 
```
JAX_CHECK_TRACER_LEAKS=1 pytest -vs test/contrib/einstein/test_steinvi.py
```
passes now.

fehiepsi: Could you also remove the CI test: https://github.com/pyro-ppl/numpyro/blob/master/.github/workflows/ci.yml#L88 ?

fehiepsi (approved):
Thanks for the fix, Ola!

PR #2007, opened 2025-03-12
Fix behavior of `logdiffexp(jnp.inf, jnp.inf)`, add test that would have caught it.

dylanhmorris:
Desired output for `logdiffexp(jnp.inf, jnp.inf)` is `jnp.nan`. We were getting `-jnp.inf` because `logdiffexp(a, a) = -jnp.inf` for any `a != jnp.inf`, including `-jnp.inf`. Added:
1. conditional to fix the issue 
2. a test of `logdiffexp` behavior at `jnp.inf, jnp.inf` that fails without these changes and passes with them.

Also fixes typo in documentation caught be codespell and clarifies an explanatory comment.


fehiepsi (approved):
Thanks @dylanhmorris!

PR #2014, opened 2025-03-31
feat: Dagum distribution

Qazalbash:
## Summary

This PR contains the implementation of the [Dagum distribution](https://en.wikipedia.org/wiki/Dagum_distribution), along with the unit tests and documentation.


numpyro/distributions/continuous.py
```diff
@@ -3212,3 +3215,104 @@ def entropy(self):
...
         return self.base_dist.entropy() + log_abs_det_jacobian / 2
+
+
+class Dagum(Distribution):
+    arg_constraints = {
+        "p": constraints.positive,
+        "a": constraints.positive,
+        "b": constraints.positive,
+    }
+    support = constraints.positive
```
martinjankowiak, line 3226: think you want to add `reparametrized_params=...`?

Qazalbash: Hi, I think I should!

Qazalbash:
The stuck test case in modeling passes locally on both py3.9 and py3.10.

fehiepsi:
Thanks @Qazalbash! Sorry for the late response. I think you can revise some parameters in `T(dist.Dagum, ...)` to have lower variance.

fehiepsi:
could you merge with the master branch? the failing tests have been fixed there

fehiepsi:
Just curious, why did you remove mean/var? is it unstable or something?

numpyro/distributions/continuous.py
```diff
@@ -3279,3 +3279,41 @@ def icdf(self, q: ArrayLike) -> ArrayLike:
...
+    @property
+    def mean(self) -> ArrayLike:
+        safe_a = jnp.where(self.a > 1.0, self.a, 2.0)
+        return jnp.where(
+            self.a > 1.0,
+            self.b
+            * jnp.exp(
+                gammaln(1.0 - 1.0 / safe_a)
+                + gammaln(self.p + 1.0 / safe_a)
+                - gammaln(self.p)
```
fehiepsi: how about using https://docs.jax.dev/en/latest/_autosummary/jax.scipy.special.betaln.html for both mean and var?

numpyro/distributions/continuous.py
```diff
@@ -3212,3 +3215,105 @@ def entropy(self):
...
+            -self.p * nn.softplus(xlogy(self.a, self.b) - xlogy(self.a, value))
+        )
+
+    def icdf(self, q: ArrayLike) -> ArrayLike:
+        q_root_p = jnp.power(q, -jnp.reciprocal(self.p))
+        return self.b * jnp.power(q_root_p - 1.0, -jnp.reciprocal(self.a))
+
+    def sample(self, key, sample_shape: tuple[int, ...] = ()) -> jnp.ndarray:
+        assert is_prng_key(key)
+        return self.icdf(random.uniform(key, shape=self.shape(sample_shape)))
```
fehiepsi, line 3559: maybe it is more stable to transform Beta samples https://en.m.wikipedia.org/wiki/Beta_prime_distribution#Related_distributions

numpyro/distributions/continuous.py
Qazalbash: I have used it, it is passing the test cases.

numpyro/distributions/continuous.py
Qazalbash: I have used it, it is passing the test cases.

numpyro/distributions/continuous.py
```diff
@@ -3212,3 +3215,88 @@ def entropy(self):
...
+        "b": constraints.positive,
+    }
+    support = constraints.positive
+    reparametrized_params = ["p", "a", "b"]
+
+    def __init__(
+        self,
+        p: ArrayLike,
+        a: ArrayLike,
+        b: ArrayLike,
```
fehiepsi: maybe use `scale` for `b`? I think `p` plays the role of concentration as in Beta or Gamma distribution. Maybe `sharpness` for `a`, represents how sharp the mode is? What do you think?

Qazalbash: I followed the convention from Wikipedia. I think we can change it.

fehiepsi (approved):

PR #2015, opened 2025-04-02
Unify constraint handling for `AutoContinuous`, `AutoDelta`, `AutoNormal`.

tillahoffmann:
This PR unifies constraint handling for several auto guides. In short, distributions for all variables in the guide are constructed in unconstrained space. If the variable has non-real support, the distribution is transformed.

The original motivation was to ensure that the support of random variables in the guide and model match. `AutoDelta` and `AutoContinuous` did not meet that requirement because they use `Delta` distributions in constrained space.

### DIFF
--- a/numpyro/infer/autoguide.py
+++ b/numpyro/infer/autoguide.py
@@ -339,6 +339,28 @@ class AutoGuideList(AutoGuide):
         return result
 
 
+def _maybe_constrain_dist_for_site(
+    site: dict, base_distribution: dist.Distribution
+) -> dist.Distribution:
+    support = site["fn"].support
+
+    # Short-circuit if the support is real and return the base distribution with the
+    # correct number of event dimensions.
+    base_support = support
+    while isinstance(base_support, constraints.independent):
+        base_support = base_support.base_constraint
+    if base_support is constraints.real:
+        if support.event_dim:
+            return base_distribution.to_event(support.event_dim)
+        else:
+            return base_distribution
+
+    # Transform the distribution to the support of the site.
+    with helpful_support_errors(site):
+        transform = biject_to(support)
+    return dist.TransformedDistribution(base_distribution, transform)
+
+
 class AutoNormal(AutoGuide):
     """
     This implementation of :class:`AutoGuide` uses Normal distributions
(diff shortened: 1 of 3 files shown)


tillahoffmann:
Interestingly, the `neutra.py` example fails with a `SIGSEGV` signal. I think this is not related to the changes here, as it also crops up in #3979. I don't get the error locally. Any idea what might be the cause?

fehiepsi:
I think it is memory issue. Could you try to set a smaller number for this flag: https://github.com/pyro-ppl/numpyro/blob/master/examples/neutra.py#L205 in the test?

tillahoffmann:
The example didn't pass with a smaller number of hidden features. I had a look at the [last passing](https://github.com/pyro-ppl/numpyro/actions/runs/14149798960/job/39641617646) and [first failing](https://github.com/pyro-ppl/numpyro/actions/runs/14175514599/job/39709628191) runs. The dependencies are almost the same and don't look like they would cause this issue. Maybe something changed about the runner?

On a related note, should we consider locking the requirements using a version-locked `requirements.txt` or `uv.lock`? There were a few instances of hard-to-debug CI failures because of versions changes if I remember correctly.

```diff
20c20
< coverage==7.7.1
---
> coverage==7.8.0
32c32
< flax==0.10.4
---
> flax==0.10.5
35c35
< fsspec==2025.3.0
---
> fsspec==2025.3.2
98c98
< -e git+https://github.com/pyro-ppl/numpyro@19fbd57d96973d7d33ec594ad99110ff544e9ea7#egg=numpyro
---
> -e git+https://github.com/pyro-ppl/numpyro@4027928e31e859cfc7eacc41d8b53baba36137de#egg=numpyro
133c133
< rich==13.9.4
---
> rich==14.0.0
170c170
< xarray==2025.3.0
---
> xarray==2025.3.1
```

fehiepsi:
Could you mark this test as xfail in CI instead? like in https://github.com/pyro-ppl/numpyro/blob/master/test/test_examples.py#L52-L59

Re version lock: how does it work? what if users want to just install numpyro without upgrading/degrading other dependencies?

numpyro/infer/autoguide.py
```diff
@@ -561,26 +574,25 @@ def __call__(self, *args, **kwargs):
...
 
                 site_loc = numpyro.param(
-                    "{}_{}_loc".format(name, self.prefix),
-                    init_loc,
-                    constraint=site["fn"].support,
-                    event_dim=event_dim,
+                    f"{name}_{self.prefix}_loc", init_loc, event_dim=event_dim
                 )
-
-                site_fn = dist.Delta(site_loc).to_event(event_dim)
```
fehiepsi, line 577: Here you are finding the MAP point in unconstrained space. This class gets MAP point in constrained space.

numpyro/infer/autoguide.py
```diff
@@ -708,26 +724,11 @@ def __call__(self, *args, **kwargs):
...
         # unpack continuous latent samples
         result = {}
-
         for name, unconstrained_value in self._unpack_latent(latent).items():
             site = self.prototype_trace[name]
-            with helpful_support_errors(site):
-                transform = biject_to(site["fn"].support)
-            value = transform(unconstrained_value)
-            event_ndim = site["fn"].event_dim
-            if numpyro.get_mask() is False:
```
fehiepsi, line 718: we need this logic to save computation for prediction.

PR #1557, opened 2023-03-19
Allow users to specify total_count_max in Multinomial

fehiepsi:
Fixes #1556

PR #2016, opened 2025-04-10
feat(distributions.conjugate): support total_count_max in DirichletMultinomial distribution

brendancooley:
Follow on to #1557 for the `DirichletMultinomial` case. 

Prior to this change, users may encounter the error `ValueError("Please specify total_count_max in Multinomial distribution.")` (see [`multinomial`](https://github.com/pyro-ppl/numpyro/blob/ab1f0dc6e954ef7d54724386667e33010b2cfc8b/numpyro/distributions/util.py#L273) in `distributions.util` when using the distribution inside of traced environments.

### DIFF
--- a/numpyro/distributions/conjugate.py
+++ b/numpyro/distributions/conjugate.py
@@ -107,6 +107,7 @@ class DirichletMultinomial(Distribution):
     :param numpy.ndarray concentration: concentration parameter (alpha) for the
         Dirichlet distribution.
     :param numpy.ndarray total_count: number of Categorical trials.
+    :param int total_count_max: the maximum number of trials, i.e. `max(total_count)`
     """
 
     arg_constraints = {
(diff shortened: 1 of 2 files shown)


brendancooley:
Looks like we're encountering the same CI failure as in #2015 https://github.com/pyro-ppl/numpyro/pull/2015#issuecomment-2784447908

fehiepsi:
Could you mark this test as xfail in CI instead? like in https://github.com/pyro-ppl/numpyro/blob/master/test/test_examples.py#L52-L59

brendancooley:
> Could you mark this test as xfail in CI instead? like in https://github.com/pyro-ppl/numpyro/blob/master/test/test_examples.py#L52-L59

Absolutely. Done.

fehiepsi (approved):
Thanks @brendancooley! Could you remove committed png and graph files?

brendancooley:
> Thanks @brendancooley! Could you remove committed png and graph files?

yeesh yes of course. added `*.png` to the gitignore for future.

.gitignore
```diff
@@ -19,6 +19,7 @@ build
 
 # data files
 *.pdf
+*.png
```
juanitorduz, line 22: We might want to add png as part of the docs right https://github.com/pyro-ppl/numpyro/tree/master/docs/source/_static/img/examples ?

fehiepsi: yes, thanks!

brendancooley: Happy to put up a PR to revert the change...or maybe try and manage artifact generation during the tests a bit better so that these files don't end up polluting the root directory. Did I miss something in the contributor notes on existing ways to manage this?

Issue #2026, opened 2025-05-08
Is the documentation of MatrixNormal wrong that it expects lower cholesky of correlation matrix?


mochar:
The docstring says

```
    :param array_like scale_tril_row: Lower cholesky of rows correlation matrix.
    :param array_like scale_tril_column: Lower cholesky of columns correlation matrix.
```

but looking at the sampling code

```python
    def sample(self, key, sample_shape=()):
        eps = random.normal(
            key, shape=sample_shape + self.batch_shape + self.event_shape
        )
        samples = self.loc + self.scale_tril_row @ eps @ jnp.swapaxes(
            self.scale_tril_column, -2, -1
        )

        return samples
```

This seems to assume `scale_tril_{row,column}` are lower cholesky of the covariance matrix according to [wikipedia](https://en.wikipedia.org/wiki/Matrix_normal_distribution#Drawing_values_from_the_distribution).

fehiepsi:
Yes, I think we should call them covariance matrices. Could you submit the fix?

mochar:
https://github.com/pyro-ppl/numpyro/pull/2027

First time doing a pull request, hope its correct!

PR #2027, opened 2025-05-09
matrix normal doc: correlation -> covariance

mochar:
MatrixNormal expect lower cholesky of covariance matrices, not correlation matrices. Fix the doc string. #2026


fehiepsi (approved):
Thanks @mochar!

PR #2028, opened 2025-05-10
refactor(infer.elbo): add type hints to elbo module

brendancooley:
Picking up on the work of @juanitorduz, this MR adds type hints to the `numpyro.infer.elbo` module. Along the way, I've made a few housekeeping changes to try and make it easier to maintain and extend type hints going forward. 

The most important change warranting discussion is that I've introduced a `ParamSpec` for the parameters of a model function.  `ELBO`s are now generics over the signature of the model/guide functions they operate on. I think that this is a nice way to (softly) enforce consistency in the `*args` and `**kwargs` passed to `ELBO.loss`, but there may be some subtleties that I'm missing here.

## Summary of Changes

- Added comprehensive type annotations to elbo module
- Remove dormant `_compute_downstream_costs` code and associated tests (replaced in https://github.com/pyro-ppl/numpyro/commit/28e38d85618621cca3796e02da680fabffb8e2a8)
- Use Ruff ANN rule to require annotations in typed modules
	- exempt nested functions in `numpyro.primitives` with `noqa`
- Added `numpyro._typing` module for cross-module shared type aliases
- Converted some rng_key annotations to `jax.Array` to address typing errors and following guidance in [JEP 9263](https://docs.jax.dev/en/latest/jep/9263-typed-keys.html#type-annotations-for-prng-keys)
- Use shared type aliases and python 3.10+ union annotations in `numpyro.handlers`, with requisite import `from __future__ import annotations` to support python 3.9.

## Previous Typing MRs (for Reference)
1. [primitives](https://github.com/pyro-ppl/numpyro/pull/1942)
2. [handlers](https://github.com/pyro-ppl/numpyro/pull/1942)
3. [optim](https://github.com/pyro-ppl/numpyro/pull/1942)


juanitorduz (approved):

juanitorduz:
This looks great! (I approved, but we need a true core dev to approve it). I think the error 

```python
FAILED test/test_example_utils.py::test_mnist_data_load - urllib.error.HTTPError: HTTP Error 429: Too Many Requests
```

Is not related with these changes.

fehiepsi:
This is awesome! I'll need to look into details and learn from your designs. :)

Please ignore the failing tests. I guess it's a github issue and will be resolved by github devs soon.

numpyro/_typing.py
```diff
@@ -0,0 +1,14 @@
...
+from collections.abc import Callable
+from typing import Any
+
+from typing_extensions import ParamSpec
+
+P = ParamSpec("P")
+ModelT = Callable[P, Any]
+
+Message = dict[str, Any]
+TraceT = OrderedDict[str, Message]
```
fehiepsi, line 14: just curious, why not Trace?

numpyro/_typing.py
```diff
@@ -0,0 +1,14 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+from collections import OrderedDict
+from collections.abc import Callable
+from typing import Any
+
+from typing_extensions import ParamSpec
+
+P = ParamSpec("P")
+ModelT = Callable[P, Any]
```
fehiepsi, line 11: Does this require users to preserve signatures in downstream calls? Sometimes, I want to use keywords for positional arguments.

numpyro/infer/elbo.py
```diff
@@ -25,12 +35,23 @@
 from numpyro.ops.provenance import eval_provenance
 from numpyro.util import _validate_model, check_model_guide_match, find_stack_level
 
+if TYPE_CHECKING:
+    T = TypeVar("T")
+    mapT = Callable[[Callable, T], T]
+    MutableStateT = dict[str, Any]
+    ParticleT = jax.Array | dict[str, jax.Array]
```
fehiepsi, line 42: why dict is required?

numpyro/infer/elbo.py
```diff
@@ -25,12 +35,23 @@
 from numpyro.ops.provenance import eval_provenance
 from numpyro.util import _validate_model, check_model_guide_match, find_stack_level
 
+if TYPE_CHECKING:
+    T = TypeVar("T")
+    mapT = Callable[[Callable, T], T]
```
fehiepsi, line 40: why uses T instead of jax.Array?

numpyro/infer/elbo.py
```diff
@@ -25,12 +35,23 @@
...
+    mutable_state: MutableStateT | None
 
-def _apply_vmap(fn, keys):
+
+def _apply_vmap(fn: Callable, keys: T) -> T:
     return vmap(fn)(keys)
 
 
-class ELBO:
+class ELBO(Generic[P]):
```
fehiepsi, line 54: why this is required?

numpyro/_typing.py
brendancooley: The idea was to convey that these are simply type aliases and not numpyro-specific classes. I read them as "model type", "trace type", etc. I added the `TypeAlias` hint for clarity here. Happy to remove the `T` part of all of these if you think that's better. I added one to `Message` (`Message` -> `MessageT`) for consistency.

https://typing.python.org/en/latest/spec/aliases.html

numpyro/infer/elbo.py
brendancooley: I believe if `sum_sites=False` for `Trace_ELBO` then a per-site dict will be returned? 

```
:param sum_sites: Whether to sum the ELBO contributions from all sites or return the
        contributions as a dictionary keyed by site.
```

brendancooley: Updated the return annotations for `Trace_ELBO` losses and renamed `ParticleT` -> `LossT` for better clarity here https://github.com/pyro-ppl/numpyro/pull/2028/commits/b9f7b4b078122a22a09380089881f752334521f4

numpyro/infer/elbo.py
brendancooley: `T` should probably refer to a pytree? To cover the case in which we want to map over a dict? But my understanding is that jax still doesn't really have conventions around how to type these: https://github.com/jax-ml/jax/issues/3340

It looks like @juanitorduz encountered similar issues on the optim PR: https://github.com/pyro-ppl/numpyro/pull/1942

numpyro/infer/elbo.py
brendancooley: I don't think this was quite right. The idea I had in mind is something like "the type of an ELBO is parameterized by the arguments of it's associated model." But ELBOs aren't associated to a particular model on instantiation, so it's better to bind the ParamSpecs to particular methods (`loss` and `loss_with_mutable_state`). Updated the signatures
```python
    def loss(
        self,
        rng_key: jax.Array,
        param_map: dict[str, jax.Array],
        model: ModelT[P],
        guide: ModelT[P],
        *args: P.args,
        **kwargs: P.kwargs,
    ) -> LossT:
```

Now `P.args` and `P.kwargs` are bound to the parameters of `ModelT` which ensures that users can only pass args/kwargs which exist in the model signature, as desired.

numpyro/_typing.py
brendancooley: You can use any valid positional/keyword combination with functions using `ParamSpec`s. I believe this is why `P.args` and `P.kwargs` must be used together in a function signature.

https://peps.python.org/pep-0612/#id1

brendancooley:
Thanks @fehiepsi and @juanitorduz for the reviews here. Still getting the hang of these typing concepts myself as you can see! Hopefully it will get easier with time and practice.

numpyro/infer/elbo.py
brendancooley: example https://github.com/pyro-ppl/numpyro/blob/17cf5a727f1339f1d9fbf8058b53d2e00e264c4e/test/infer/test_svi.py#L485

numpyro/_typing.py
fehiepsi: gotcha, thanks. Using TraceT looks fine with me. I dont have strong opinion on the names.

numpyro/primitives.py
```diff
@@ -14,19 +14,17 @@
 from jax.typing import ArrayLike
 
 import numpyro
+from numpyro._typing import MessageT
 from numpyro.distributions.distribution import DistributionLike
 from numpyro.util import find_stack_level, identity
 
-# Type aliases
-Message = dict[str, Any]
```
fehiepsi, line 21: could you keep Message = MessageT in case users already use Message in their libraries?

pyproject.toml
```diff
@@ -104,6 +112,7 @@ doctest_optionflags = [
 ]
 
 [tool.mypy]
+python_version = 3.12
```
fehiepsi, line 115: does this mean that we only allow mypy to work with python 3.12?

fehiepsi (approved):
Looks awesome, thanks @brendancooley!

numpyro/primitives.py
brendancooley: oops yes, forgot this was pre-existing. reverted.

pyproject.toml
brendancooley: for python 3.9, mypy complains about modern (3.10+) type annotations on our type aliases

```python
LossT: TypeAlias = jax.Array | dict[str, jax.Array]
```

```
numpyro/infer/elbo.py:44: error: Invalid type alias: expression is not a valid type  [valid-type]
numpyro/infer/elbo.py:44: error: Unsupported left operand type for | ("type[Array]")  [operator]
Found 2 errors in 1 file (checked 88 source files)
```

Since python 3.9 is approaching [end of life](https://devguide.python.org/versions/), I went ahead and excluded the lint workflow from CI for the 3.9 case, which will let us use `X | Y` in lieu of `typing.Union[X, Y]` across the codebase without having to add inline ignores for the typechecker.

Alternatively, I could add an inline ignore or adopt use the legacy `Union[...]` type hint. Whatever you prefer!

Issue #2030, opened 2025-05-20
Request for Beta Negative Binomial distribution


ShozenD:
### Feature Summary
Adding an implementation of the Beta Negative Binomial distribution to the conjugate.py module.
https://en.wikipedia.org/wiki/Beta_negative_binomial_distribution

### Why is this needed?
Provides a parametric heavy tailed count distribution.
The distribution is provided in the latest version of Stan.
https://mc-stan.org/docs/functions-reference/unbounded_discrete_distributions.html#beta-neg-binomial

PR #2031, opened 2025-05-20
[add] Implementation of the Beta Negative Binomial Distribution

ShozenD:
Implementation proposal for [Issue #2030].
So far all unit tests are passing (including documentation tests).

I have not implemented the CDF (also not available in the most recent version of scipy).
Stan does seem to provide a CDF implementation.
Does anyone know what is the best way to do this?

### DIFF
--- a/numpyro/distributions/__init__.py
+++ b/numpyro/distributions/__init__.py
@@ -8,6 +8,7 @@ from numpyro.distributions.conjugate import (
     NegativeBinomial2,
     NegativeBinomialLogits,
     NegativeBinomialProbs,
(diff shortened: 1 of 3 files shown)


fehiepsi (approved):
Thanks @ShozenD! I think we can address cdf icdf later if needed.

Could you add an entry for this distribution in https://github.com/pyro-ppl/numpyro/blob/master/docs/source/distributions.rst ?

fehiepsi:
Hi @ShozenD, could you merge with master and add the distribution to the doc index?

Issue #2037, opened 2025-06-15
Add support for passing initial value in `RecursiveLinearTransform`


tingiskhan:
### Feature Summary
Add a parameter called `initial_value` to `RecursiveLinearTransform`.

### Why is this needed?
When using RecursiveLinearTransform for modeling structural time series, I sometimes run into limitations related to initializing the transform. My main use case involves modeling latent structural time series, and I've encountered two main issues:
1. **Initialization during inference**: for example, when modeling auto-regressive series, it's often necessary to explicitly set the initial value of the latent series.
2. **Forecasting after inference**: since there's currently no mechanism to provide an initial value for the latent state, the first forecasted step tends to be inaccurate or "off".

EDIT: there is of course the possibility to modify the mean of the distribution to be transformed, but I find this a bit cumbersome and would appreciate a simpler way.

fehiepsi:
fixed in #2038 thanks @tingiskhan !

PR #2038, opened 2025-06-15
Adds support for passing initial value to transform

tingiskhan:
### Changes made

Added support for passing a parameter called `initial_value` to `RecursiveLinearTransform` to facilitate forecasting when using the transform for latent variables.

### Links to related PRs

#2037 

### Tests

I added one test case to the "general" part, and parameterized `test_batched_recursive_linear_transform` for testing arbitrary initial values.

### Dependencies

No new dependencies introduced.

### DIFF
--- a/numpyro/distributions/transforms.py
+++ b/numpyro/distributions/transforms.py
@@ -1540,29 +1540,61 @@ class RecursiveLinearTransform(Transform):
     domain = constraints.real_matrix
     codomain = constraints.real_matrix
 
-    def __init__(self, transition_matrix: Array) -> None:
+    def __init__(self, transition_matrix: Array, initial_value: Array = None) -> None:
         self.transition_matrix = transition_matrix
+
+        batch_shape, event_shape = (
+            transition_matrix.shape[:-2],
+            transition_matrix.shape[-1:],
+        )
+
+        if initial_value is None:
+            initial_value = jnp.zeros(event_shape)
+
+        self.initial_value = jnp.broadcast_to(initial_value, batch_shape + event_shape)
 
     def __call__(self, x: Array) -> Array:
         # Move the time axis to the first position so we can scan over it.
+        sample_shape = x.shape[:-2]
         x = jnp.moveaxis(x, -2, 0)
 
         def f(y, x):
             y = jnp.einsum("...ij,...j->...i", self.transition_matrix, y) + x
             return y, y
 
-        _, y = lax.scan(f, jnp.zeros_like(x, shape=x.shape[1:]), x)
+        shape = jnp.broadcast_shapes(
+            sample_shape,
+            self.transition_matrix.shape[:-2],
+            self.initial_value.shape[:-1],
+        )
+        initial_value = jnp.broadcast_to(
+            self.initial_value, shape + self.initial_value.shape[-1:]
+        )
+
+        _, y = lax.scan(f, initial_value, x)
         return jnp.moveaxis(y, 0, -2)
 
     def _inverse(self, y: Array) -> Array:
         # Move the time axis to the first position so we can scan over it in reverse.
+        sample_shape = y.shape[:-2]
         y = jnp.moveaxis(y, -2, 0)
 
         def f(y, prev):
             x = y - jnp.einsum("...ij,...j->...i", self.transition_matrix, prev)
             return prev, x
 
-        _, x = lax.scan(f, y[-1], jnp.roll(y, 1, axis=0).at[0].set(0), reverse=True)
+        shape = jnp.broadcast_shapes(
+            sample_shape,
+            self.transition_matrix.shape[:-2],
+            self.initial_value.shape[:-1],
+        )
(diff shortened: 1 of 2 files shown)


tingiskhan:
> Thank you for the contribution, @tingiskhan! Do you think this could also be achieved using a composition of an `AffineTransform` and `RecursiveLinearTransform`?

That’s a very good point, I’ll take a look at it next week (if you don’t beat me to it)!

tingiskhan:
@tillahoffmann So, I've given it some thought and see three ways forward (there might be more):
1. Keep as is (i.e. pass initial value to `RecursiveLinearTransform`).
2. Use a composition of `AffineTransform` and original `RecursiveLinearTransform` (as you suggested). The "issue" I see with this method is that we'll need to pass the length of the "series" on `__init__` in order to construct the "loc" variable of `AffineTransform` correctly: `loc * jnp.eye((length, 1))` (where `loc` corresponds to initial value). This would break the (implicit) design of transforms a bit I think.
3. Construct a derived class from `AffineTransform` that only takes the initial value as input (corresponding to `loc` and pass 1 to `scale`) and then in `__call__`/`_inverse` handle the building of the offset given the sample (same as in 2.). 

Of these three I think either 1. or 3. is the most appropriate. I'm not sure how re-usable the new component of 3. would be, but I'm open to either suggestions.

fehiepsi:
If initial value is 0, we have y_1=x_1, so we can just prepend the initial value to x and y?

tingiskhan:
@fehiepsi Yes, I suppose that would work as well! What solution would be preferred (if at all)?

numpyro/distributions/transforms.py
```diff
@@ -1540,36 +1540,71 @@ class RecursiveLinearTransform(Transform):
...
 
+        batch_shape, event_shape = (
+            transition_matrix.shape[:-2],
+            transition_matrix.shape[-1:],
+        )
+
+        if initial_value is None:
+            initial_value = jnp.zeros(event_shape)
+
+        self.initial_value = jnp.broadcast_to(initial_value, batch_shape + event_shape)
```
fehiepsi: if not None, maybe we can broadcast batch_shape to combine with the batch shape of initial value first?

fehiepsi:
Hi @tingiskhan, I think supporting initial values makes sense.

numpyro/distributions/transforms.py
tingiskhan: Absolutely, I’ll take a look at fixing your comments this weekend

numpyro/distributions/transforms.py
```diff
@@ -1540,36 +1540,80 @@ class RecursiveLinearTransform(Transform):
...
+            f" {initial_value.shape[-1:]}."
+        )
+
+        iv_batch_shape = initial_value.shape[:-1]
+        batch_shape = jnp.broadcast_shapes(transition_batch_shape, iv_batch_shape)
+
+        self.initial_value = jnp.broadcast_to(initial_value, batch_shape + event_shape)
+        self.transition_matrix = jnp.broadcast_to(
+            transition_matrix, batch_shape + 2 * event_shape
+        )
```
fehiepsi, line 1566: I think we don't need to broadcast the transition matrix.

numpyro/distributions/transforms.py
```diff
@@ -1540,36 +1540,80 @@ class RecursiveLinearTransform(Transform):
...
         x = jnp.moveaxis(x, -2, 0)
 
         def f(y, x):
             y = jnp.einsum("...ij,...j->...i", self.transition_matrix, y) + x
             return y, y
 
-        _, y = lax.scan(f, jnp.zeros_like(x, shape=x.shape[1:]), x)
+        shape = jnp.broadcast_shapes(
+            sample_shape,
+            self.transition_matrix.shape[:-2],
```
fehiepsi: no need to include transition matrix shape

numpyro/distributions/transforms.py
```diff
@@ -1540,36 +1540,80 @@ class RecursiveLinearTransform(Transform):
...
         y = jnp.moveaxis(y, -2, 0)
 
         def f(y, prev):
             x = y - jnp.einsum("...ij,...j->...i", self.transition_matrix, prev)
             return prev, x
 
-        _, x = lax.scan(f, y[-1], jnp.roll(y, 1, axis=0).at[0].set(0), reverse=True)
+        shape = jnp.broadcast_shapes(
+            sample_shape,
+            self.transition_matrix.shape[:-2],
```
fehiepsi: like above

fehiepsi (approved):
LGTM pending some nits. Thanks @tingiskhan!

tingiskhan:
@fehiepsi, the tests were failing because of the `jax.live_arrays` check since I was creating jax arrays in `__init__` of `RecursiveLinearTransform`, and sampling an initial value in one of the parametrized tests. I had to rethink the logic of how and when we broadcast the initial value, see latest commit.

Issue #2035, opened 2025-06-03
Adding new samplers to NumPyro


reubenharry:
### Feature Summary

Recently, a couple of samplers have been proposed by the group I work with (https://arxiv.org/abs/2503.01707, https://arxiv.org/abs/2212.08549) as alternatives to NUTS HMC (see issue #1662
). Since they appear to be quite a bit faster than NUTS (at least on benchmark problems I've tried), and relatively simple, I'm interested in adding them to NumPyro, but wanted to get some advice.

Currently, implementations exist in Blackjax. In an ideal world, I'd make a new class like `class AdjustedMicrocanonical(numpyro.infer.mcmc.MCMCKernel)` which basically just wraps Blackjax.

In addition, my eventual goal would be to add not just the kernel, but also the tuning scheme (which is key to good performance). I'm curious if there's a straightforward way to do that.

## Motivation

While it's easy to write a model in NumPyro and extract the density, then use Blackjax for inference, we want to give users more direct access (basically for the purpose of increasing discoverability).

reubenharry:
Noted. I maintain the blackjax samplers, and would plan to add a stable one (we haven't changed the unadjusted MCLMC sampler for a year or so), so that might offset the problem somewhat.

What I'm more unsure of is what it would look like to add a hyperparameter adaptation algorithm to numpyro - is this feasible?

fehiepsi:
Just curious, is it easier to use blackjax for inference with those samplers? If so, I think you can add a section to https://num.pyro.ai/en/stable/tutorials/other_samplers.html to illustrate the usage.

reubenharry:
Yes, it's easier, but my motivation for adding samplers directly to Numpyro is to make them more accessible (and in particular, discoverable) for users of Numpyro.

PR #2039, opened 2025-06-23
Microcanonical Langevin Monte Carlo

reubenharry:
This PR adds the [Microcanonical Langevin Monte Carlo](https://microcanonical-monte-carlo.netlify.app/) to Blackjax, following the discussion of issue #2035. 

A couple things to resolve:

- The code is in the main directory, because if I try anything in the infer directory, such as hmc.py, I get an error `ImportError: cannot import name 'device_get' from partially initialized module 'jax' (most likely due to a circular import) (/opt/homebrew/Caskroom/miniconda/base/envs/mclmc/lib/python3.11/site-packages/jax/__init__.py)`. Does this look familiar? 
- Before merging, I would probably like to set some better defaults for the amount of tuning etc, for robust performance.

### DIFF
--- a/mclmc.py
+++ b/mclmc.py
@@ -1,0 +1,207 @@
+
+
+import argparse
+from collections import namedtuple
+import os
+
+import matplotlib.pyplot as plt
+
+import jax
+import jax.numpy as jnp
+from jax import random
+
+
+import numpyro
+import numpyro.distributions as dist
+from numpyro.infer import MCMC
+from numpyro.infer.mcmc import MCMCKernel
+import blackjax
+from numpyro.infer.util import initialize_model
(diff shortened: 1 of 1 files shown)


juanitorduz:
**Suggestion:** What about adding BlackJax as an optional dependency and adding this component intro de contrib submodule? 

And it would be great to have this simple example visible in the example gallery 🙂

juanitorduz:
> Hi, sorry for the slow response. Yes, if you're interested in doing that, it would be extremely helpful - I wasn't sure how to proceed.

FYI https://github.com/pyro-ppl/numpyro/pull/2124#issuecomment-3775256307 :)

PR #2046, opened 2025-07-16
Remove uses of OrderedDict

sid-kap:
Since python 3.6 dict has the same behavior as OrderedDict. Numpyro requires python version >=3.9, so it should be fine to use dict instead of OrderedDict.

### DIFF
--- a/numpyro/_typing.py
+++ b/numpyro/_typing.py
@@ -2,7 +2,6 @@
 # SPDX-License-Identifier: Apache-2.0
 
 
-from collections import OrderedDict
 from collections.abc import Callable
 from typing import Any, Protocol, runtime_checkable
 
@@ -15,7 +14,7 @@ P = ParamSpec("P")
 ModelT: TypeAlias = Callable[P, Any]
 
 Message: TypeAlias = dict[str, Any]
-TraceT: TypeAlias = OrderedDict[str, Message]
+TraceT: TypeAlias = dict[str, Message]
 
 
 @runtime_checkable
--- a/numpyro/contrib/stochastic_support/dcc.py
+++ b/numpyro/contrib/stochastic_support/dcc.py
@@ -2,8 +2,8 @@
 # SPDX-License-Identifier: Apache-2.0
 
 from abc import ABC, abstractmethod
-from collections import OrderedDict, namedtuple
-from typing import Any, Callable, OrderedDict as OrderedDictType, Union
+from collections import defaultdict
+from typing import Any, Callable, Union
 
 import jax
 from jax import random
@@ -61,7 +61,7 @@ class StochasticSupportInference(ABC):
 
     def _find_slps(
         self, rng_key: jax.Array, *args: Any, **kwargs: Any
-    ) -> dict[str, OrderedDictType]:
+    ) -> dict[str, dict]:
         """
         Discover the straight-line programs (SLPs) in the model by sampling from the prior.
         This implementation assumes that all branching is done via discrete sampling sites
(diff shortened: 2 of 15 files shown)


fehiepsi:
Thanks @sid-kap! Do you think https://github.com/jax-ml/jax/issues/24398 is still relevant for the changes?

sid-kap:
I think it should be fine because `dict` now keeps the keys ordered, like OrderedDict

sid-kap:
Ah you're right, my bad! Thanks for the clarification! It looks like jax handles OrderedDict correctly:
```python
import jax
x = OrderedDict([("b", 1), ("a", 2)])
jax.jit(lambda x: x)(x)
```
returns
```
OrderedDict([('b', Array(1, dtype=int32, weak_type=True)), ('a', Array(2, dtype=int32, weak_type=True))])
```

So any function that may be jitted should be kept as OrderedDict.

I'll read through the code again and see if there are any functions that are guaranteed to not be jitted, but if not it probably makes sense to close this.

fehiepsi:
Looking at the changes, I think most are valid. We just need to keep TraceT as-is because order in trace is important for inference algorithms.

numpyro/_typing.py
```diff
@@ -15,7 +14,7 @@
 ModelT: TypeAlias = Callable[P, Any]
 
 Message: TypeAlias = dict[str, Any]
-TraceT: TypeAlias = OrderedDict[str, Message]
+TraceT: TypeAlias = dict[str, Message]
```
fehiepsi, line 17: Let's keep TraceT as OrderedDict

numpyro/contrib/stochastic_support/dcc.py
```diff
@@ -61,7 +61,7 @@ def __init__(self, model: Callable, num_slp_samples: int, max_slps: int) -> None
 
     def _find_slps(
         self, rng_key: jax.Array, *args: Any, **kwargs: Any
-    ) -> dict[str, OrderedDictType]:
+    ) -> dict[str, dict]:
```
fehiepsi, line 64: I'm not sure but it seems that we need to use OrderedDict in this module.

numpyro/contrib/stochastic_support/sdvi.py
```diff
@@ -98,7 +98,7 @@ def __init__(
     def _run_inference(
         self,
         rng_key: jax.Array,
-        branching_trace: OrderedDictType,
+        branching_trace: dict,
```
fehiepsi, line 101: like above, maybe we need to use OrderedDict in this module.

numpyro/handlers.py
```diff
@@ -155,19 +154,19 @@ class trace(Messenger):
 
        >>> exec_trace = trace(seed(model, random.PRNGKey(0))).get_trace()
        >>> pp.pprint(exec_trace)  # doctest: +SKIP
-       OrderedDict([('a',
```
fehiepsi, line 158: revert this

numpyro/util.py
```diff
@@ -174,7 +173,7 @@ def identity(x, *args, **kwargs):
 def cached_by(outer_fn, *keys):
     # Restrict cache size to prevent ref cycles.
     max_size = 8
-    outer_fn._cache = getattr(outer_fn, "_cache", OrderedDict())
+    outer_fn._cache = getattr(outer_fn, "_cache", {})
```
fehiepsi, line 176: Let's use OrderedDict here I guess.

sid-kap:
Ok sorry, it sounds like most of this will need to be undone. I will close for now

Issue #2045, opened 2025-07-15
Add support for a divider parameter to `random_nnx_module()`


markean:
### Feature Summary
Add a divider as a parameter in `random_nnx_module()`.

### Why is this needed?
Currently, the default divider `"/"` is used here:
https://github.com/pyro-ppl/numpyro/blob/6fd6045a3df94e7dbb61620615920f6c7c505c89/numpyro/contrib/module.py#L565

This leads to site names like `'nn/linear1.bias_dim_0'` and `'nn/linear1.bias'`.
The presence of `"/"` in these names causes an error when wrapping them into an xarray `DataTree`:
https://github.com/pydata/xarray/blob/699d8957ec174f118108005aeb6ba99c1920167a/xarray/core/datatree.py#L209

Allowing users to set a custom divider (e.g., `"."` or `"_"`) would make it much easier to integrate NumPyro models into downstream analyses. For context, see also the related migration notice from the ArviZ side:
https://github.com/pydata/xarray/blob/699d8957ec174f118108005aeb6ba99c1920167a/xarray/core/datatree.py#L209

PR #2048, opened 2025-07-16
Allow scope divider in NNX and Equinox Random Modules

juanitorduz:
Closes https://github.com/pyro-ppl/numpyro/issues/2045


fehiepsi (approved):
Cool! Thanks @juanitorduz.

Issue #2036, opened 2025-06-07
MyPy Errors in Distributions Module


juanitorduz:
### Bug Description
In the PR, adding type hints to the `distributions` module https://github.com/pyro-ppl/numpyro/pull/2032, the CI did not check for consistency, and we have many lint errors. 

### Steps to Reproduce
Steps to reproduce the behavior.

We must always add the module into the mypy config  in the `pyproject.toml` file as 

```python
[[tool.mypy.overrides]]
module = [
    "numpyro.contrib.control_flow.*",       # types missing
    "numpyro.contrib.funsor.*",             # types missing
    "numpyro.contrib.hsgp.*",
    "numpyro.contrib.stochastic_support.*",
    "numpyro.diagnostics.*",
    "numpyro.distributions.*",    # <-- This line was missing!
    "numpyro.handlers.*",
    "numpyro.infer.elbo.*",
    "numpyro.optim.*",
    "numpyro.primitives.*",
    "numpyro.patch.*",
    "numpyro.util.*",
]
ignore_errors = false
```

When I add this line and do the mypy checks, we have many errors 

```python
...
Found 1043 errors in 14 files (checked 88 source files)
```

### Expected Behavior
We should add the line ` "numpyro.distributions.*"` to the `pyproject.toml` file as above and fix all the errors. 
As there are many, we can do it in small PRs.

FYI: @Qazalbash

Qazalbash:
Hi, thanks for the report. I was oblivious of the settings in `pyproject.toml` and blindly followed mypy's pre-commit checks. I am looking into this issue.

juanitorduz:
All good! No problem at all. We can breakdown this into small PRs so that we can all help if needed:)

PR #2032, opened 2025-05-24
feat(gh-299): Type hints in distributions modules

Qazalbash:
Hi,

I have added type hints using [`jaxtyping`](https://github.com/patrick-kidger/jaxtyping) in `numpyro.distributions.*.py` modules. I have accordingly updated the `setup.py` too.

All types of protocols have been transferred to `numpyro._typing`. I have modified the `DistributionLike` type along with two new types, `TransformLike` and `ConstraintLike`.

---

This PR is related to #299.

PR #2050, opened 2025-07-17
fix(gh-2036): MyPy Errors in Distributions Module

Qazalbash:
This PR contains the partial resolution of mypy errors passed by #2032.


numpyro/_typing.py
```diff
@@ -4,29 +4,39 @@
...
 from jax.typing import ArrayLike
 
+from numpyro.distributions import MaskedDistribution
+
 P = ParamSpec("P")
 ModelT: TypeAlias = Callable[P, Any]
 
 Message: TypeAlias = dict[str, Any]
 TraceT: TypeAlias = OrderedDict[str, Message]
+PRNGKeyT: TypeAlias = Union[jax.dtypes.prng_key, ArrayLike]
```
fehiepsi, line 22: I think we can just use jax.Array for PRNGKey

numpyro/_typing.py
```diff
@@ -4,29 +4,39 @@
...
-from typing import Any, Protocol, runtime_checkable
+from typing import Any, Optional, Protocol, Union, runtime_checkable
 
 from typing_extensions import ParamSpec, TypeAlias
 
 import jax
+from jax import Array
 from jax.typing import ArrayLike
 
+from numpyro.distributions import MaskedDistribution
```
fehiepsi, line 15: would prefer not importing MaskedDistribution here

pyproject.toml
```diff
@@ -123,6 +125,7 @@ module = [
     "numpyro.contrib.hsgp.*",
     "numpyro.contrib.stochastic_support.*",
     "numpyro.diagnostics.*",
+    "numpyro.distributions.*",
```
fehiepsi, line 128: you can comment out this and add TODO link to https://github.com/pyro-ppl/numpyro/issues/2036

numpyro/handlers.py
```diff
@@ -677,7 +676,7 @@ class scale(Messenger):
     def __init__(
         self,
         fn: Optional[Callable] = None,
-        scale: ArrayLike = 1.0,
+        scale: Array = 1.0,
```
fehiepsi, line 679: note: ArrayLike works for float

numpyro/handlers.py
```diff
@@ -583,7 +582,7 @@ class mask(Messenger):
     def __init__(
         self,
         fn: Optional[Callable] = None,
-        mask: Optional[ArrayLike] = True,
+        mask: Optional[Array] = True,
```
fehiepsi, line 585: note: ArrayLike works for boolean

numpyro/distributions/batch_util.py
```diff
@@ -69,7 +67,7 @@
 
 
 @singledispatch
-def vmap_over(d: Union[Distribution, Transform, Constraint], **kwargs):
+def vmap_over(d: DistributionT, **kwargs):
```
fehiepsi, line 70: why removing other objects?

fehiepsi (commented):
Thanks @Qazalbash! I thought we can merge your changes and iterate on later PRs. Turns out that there are many non-trivial changes so it would be nice to split this into several PRs, each for one script like `continuous.py` etc. (would prefer simple ones first).

Let's resolve the typing issues when you have bandwidth. Thanks again for this important contribution!

Qazalbash:
> Thanks @Qazalbash! I thought we can merge your changes and iterate on later PRs. Turns out that there are many non-trivial changes so it would be nice to split this into several PRs, each for one script like `continuous.py` etc. (would prefer simple ones first).
> 
> Let's resolve the typing issues when you have bandwidth. Thanks again for this important contribution!

Thanks, I will make sure these issues get resolved as early as possible.

Qazalbash:
Yes, please, it serves no purpose. Also, let me know which module we should tackle next.

Issue #2051, opened 2025-07-22
Incompatibility with jax 0.7.0


damonbayer:
`import numpyro` fails with jax 0.7.0 with the following error

```pycon
>>> import numpyro
ImportError: cannot import name 'pjit_p' from 'jax.experimental.pjit' (/Users/damon/Documents/GitHub/stf-pyrenew-added-value/.venv/lib/python3.13/site-packages/jax/experimental/pjit.py)
```

This is because of the following note from the jax 0.7.0 [release](https://github.com/jax-ml/jax/releases/tag/jax-v0.7.0):

> The `jax.extend.core.primitives.pjit_p` primitive has been renamed to `jit_p`, and its `name` attribute has changed from `"pjit"` to `"jit"`. This affects the string representations of jaxprs. The same primitive is no longer exported from the `jax.experimental.pjit` module.

PR #2052, opened 2025-07-23
Handle renaming of `pjit_p` to `jit_p`

tare:
This PR aims to address https://github.com/pyro-ppl/numpyro/issues/2051.

### DIFF
--- a/numpyro/ops/provenance.py
+++ b/numpyro/ops/provenance.py
@@ -3,8 +3,11 @@
 
 import jax
 from jax.api_util import flatten_fun, shaped_abstractify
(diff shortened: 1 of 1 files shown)


tare:
Added change from https://github.com/pyro-ppl/numpyro/pull/2056. It looks like we end up installing tensorflow_probability and tfp-nightly because  jaxns requires tensorflow_probability ([here](https://github.com/Joshuaalbert/jaxns/blob/38946971f8b53c8a2fefd38a6304b8326394bbcf/requirements.txt#L6)), which still causes issues.

PR #2058, opened 2025-08-04
Test against Python 3.13

juanitorduz:
As Python 3.13 is already quite stable, I suggest testing against it.

See https://peps.python.org/pep-0719/


fehiepsi (approved):
Thanks, @juanitorduz!

PR #2061, opened 2025-08-06
numpyro.contrib.module: only collect mutables in nnx_module() when th…

esennesh:
…ey exist

Testing Done: avoided registration of empty mutables in nnx ConvNet

### DIFF
--- a/numpyro/contrib/module.py
+++ b/numpyro/contrib/module.py
@@ -484,8 +484,6 @@ def nnx_module(name, nn_module):
 
     mutable_holder = None
     if eager_other_state_dict:
-        mutable_holder = numpyro_mutable(name + "$state")
-    if mutable_holder is None:
         mutable_holder = numpyro_mutable(
             name + "$state", {"state": eager_other_state_dict}
         )


numpyro/contrib/module.py
```diff
@@ -484,8 +484,6 @@ def nnx_module(name, nn_module):
 
     mutable_holder = None
     if eager_other_state_dict:
-        mutable_holder = numpyro_mutable(name + "$state")
-    if mutable_holder is None:
```
fehiepsi, line 488: I guess it is better to move this `if` inside the above `if`?

esennesh: My patch deletes the lower `if` since by construction, `mutable_holder is None` is always true here.  We don't want to create an empty mutable-state no matter what, since there are inference algorithms and other things that depend upon the nonexistence of mutables (which is how I found this in the first place).

fehiepsi: That makes sense. Thanks for clarifying! Could you adjust the failing test to cover the change? I think without batchnorm, we no longer have mutable state in that test.

esennesh:
Yeah, of course.  I’ve had some fires to put out the past few days, but over the weekend I’m gonna look at what’s up with the test.  Wouldn’t BatchNorm actually maintain mutable state throughout execution?  So it should have a non-None mutable.

numpyro/contrib/module.py
esennesh: All right!  Changed the test and in the no-batchnorm, no-dropout case, we no longer expect to see any `'nn$state'` in the keys found.

PR #2066, opened 2025-08-24
fix(gh-2036): MyPy Errors in `numpyro.distributions.transforms` Module

Qazalbash:
This PR contains the resolution of mypy errors passed by https://github.com/pyro-ppl/numpyro/pull/2032, in `numpyro.distributions.transforms` module.

There are two cases in particular which I am unable to resolve. You can see them by running the mypy.

`log_abs_det_jacobian` of many transforms have unused parameters. I have typed them as union of `UnusedParam` and some appropriate numpy/jax type.

Many cases were unresolvable, like, `__eq__` method expects `bool` as return type, but `&` operation between arrays return array of type bool, which conflicts with the return type, therefore I have added the tag to ignore them. You will find similar tags in the file.


juanitorduz:
This looks great, thanks! There are some minor errors 

```python
numpyro/distributions/transforms.py:82: error: Missing positional argument "x" in call to "__call__" of "TransformT"  [call-arg]
Installing missing stub packages:
numpyro/distributions/transforms.py:84: error: Name "inv" already defined on line 80  [no-redef]
numpyro/distributions/transforms.py:85: error: Incompatible types in assignment (expression has type "ReferenceType[None]", variable has type "TransformT | None")  [assignment]
numpyro/distributions/transforms.py:86: error: Incompatible return value type (got "Array | Any | None", expected "TransformT")  [return-value]
numpyro/distributions/transforms.py:1550: error: Item "ndarray[tuple[Any, ...], dtype[Any]]" of "ndarray[tuple[Any, ...], dtype[Any]] | Array" has no attribute "at"  [union-attr]
```

Do you need some help with these :) ?

Qazalbash:
```
numpyro/distributions/transforms.py:82: error: Missing positional argument "x" in call to "__call__" of "TransformT"  [call-arg]
Installing missing stub packages:
numpyro/distributions/transforms.py:84: error: Name "inv" already defined on line 80  [no-redef]
numpyro/distributions/transforms.py:85: error: Incompatible types in assignment (expression has type "ReferenceType[None]", variable has type "TransformT | None")  [assignment]
numpyro/distributions/transforms.py:86: error: Incompatible return value type (got "Array | Any | None", expected "TransformT")  [return-value]
```

These errors are from [numpyro/distributions/transforms.py#L77-L86](https://github.com/Qazalbash/numpyro/blob/3b018369e9d3e2f18275edb14989157d0ec603df/numpyro/distributions/transforms.py#L77-L86), and I am not able to understand the significance of different conditions and the weak reference. I think you can look into this matter.

I will take this one,

```
numpyro/distributions/transforms.py:1550: error: Item "ndarray[tuple[Any, ...], dtype[Any]]" of "ndarray[tuple[Any, ...], dtype[Any]] | Array" has no attribute "at"  [union-attr]
```

Qazalbash:
@juanitorduz Thanks for the changes, `mypy` is happy now.

Do I need to remove the plugin from `pyproject.toml`?

juanitorduz:
It's depreciated so it's safe to remove

numpyro/distributions/transforms.py
```diff
@@ -67,36 +68,39 @@ def _clipped_expit(x: ArrayLike) -> ArrayLike:
...
         super().__init_subclass__(**kwargs)
         register_pytree_node(cls, cls.tree_flatten, cls.tree_unflatten)
 
     @property
-    def inv(self) -> TransformT:
+    def inv(self: TransformT) -> TransformT:
         inv = None
         if self._inv is not None:
-            inv = self._inv()
+            inv = self._inv
```
juanitorduz: I think this might be causing the tests to fail ... the problem is that inv is a Transform and the associated protocol TransformT has a `__call__(x)` function that receives a parameter and this is why MyPy complains :/

juanitorduz: @fehiepsi why will this `@property` would wanna do `inv = self._inv()` when returning `inv`? I think we need to understand the logic to properly add the hints :)

fehiepsi: Here `self._inv` is a `weakref.ReferenceType` instance. This is useful when `inv` is no longer used - it can be safely garbage collected. You can add a assertion here to make sure that `self._inv` is a weak reference before calling.

juanitorduz: Thank you! @Qazalbash I added the suggested type-hint change in https://github.com/pyro-ppl/numpyro/commit/12908709ee2b0f2fbb41169d61c66f688c679bb4 and it seems to work

Qazalbash: Updated

numpyro/_typing.py
```diff
@@ -87,20 +92,30 @@ def is_discrete(self) -> bool: ...
...
+    def _inverse(self, y: Union[jax.Array, Any]) -> Union[jax.Array, Any]: ...
     def log_abs_det_jacobian(
-        self, x: ArrayLike, y: ArrayLike, intermediates=None
-    ) -> ArrayLike: ...
-    def call_with_intermediates(self, x: ArrayLike) -> tuple[ArrayLike, None]: ...
+        self,
+        x: Union[jax.Array, Any],
+        y: Union[jax.Array, Any],
+        intermediates: Optional[Any] = None,
+    ) -> Union[jax.Array, Any]: ...
```
fehiepsi: why we need those Union, shouldn't Any be enough?

numpyro/distributions/transforms.py
```diff
@@ -173,12 +177,15 @@ def sign(self) -> ArrayLike:
...
+    def __call__(self, x: StrictArrayT) -> StrictArrayT:
         return self._inv._inverse(x)
 
     def log_abs_det_jacobian(
-        self, x: ArrayLike, y: ArrayLike, intermediates=None
-    ) -> ArrayLike:
+        self,
+        x: StrictArrayT,
+        y: StrictArrayT,
+        intermediates: Union[Any, None, UnusedParam] = None,
```
fehiepsi: nit: Any is enough

numpyro/_typing.py
```diff
@@ -20,6 +22,9 @@
 Message: TypeAlias = dict[str, Any]
 TraceT: TypeAlias = OrderedDict[str, Message]
 
+# ArrayLike type has StaticScalar, StrictArrayT has everything except StaticScalars
+StrictArrayT = Union[np.ndarray, jax.Array]
```
fehiepsi: I'm not sure why StrictArrayT is needed, could you elaborate? I think we can use ArrayLike instead

Qazalbash: Some functions do not operate on `boolean` or `complex`; `ArrayLike` contains `boolean` and `complex` types, and mypy was throwing an error over it. That's why I created `StrictArrayT`, it is used where only arrays are required.

numpyro/distributions/transforms.py
Qazalbash: You also want to remove `UnusedParam`?

numpyro/_typing.py
Qazalbash: If we only use `Any`, then IDEs show types like this,
image
otherwise,
image

fehiepsi: I see. You want to see some Array there. How about using ArrayLike?

numpyro/distributions/transforms.py
fehiepsi: I guess we can use type Optional[list[PyTree]]? where `PyTree = Any` in `numpyro._typing`.

numpyro/_typing.py
fehiepsi: But we also want to support int, float at those places. Could you let me know which function has the issue?

numpyro/_typing.py
Qazalbash: If we change it to `ArrayLike`, then we get these errors.

```
numpyro/distributions/transforms.py:181: error: Incompatible return value type (got "Array | ndarray[tuple[Any, ...], dtype[Any]] | numpy.bool[builtins.bool] | number[Any, int | float | complex] | int | float | complex | Any", expected "ndarray[tuple[Any, ...], dtype[Any]] | Array")  [return-value]
numpyro/distributions/transforms.py:190: error: Unsupported operand type for unary - ("Array | ndarray[tuple[Any, ...], dtype[Any]] | numpy.bool[builtins.bool] | number[Any, int | float | complex] | int | float | complex | Any")  [operator]
numpyro/distributions/transforms.py:190: error: Incompatible return value type (got "Array | ndarray[tuple[Any, ...], dtype[Any]] | Any | number[Any, int | float | complex] | int | float | complex", expected "ndarray[tuple[Any, ...], dtype[Any]] | Array")  [return-value]
numpyro/distributions/transforms.py:366: error: Incompatible types in assignment (expression has type "Array | ndarray[tuple[Any, ...], dtype[Any]] | numpy.bool[builtins.bool] | number[Any, int | float | complex] | int | float | complex | Any", variable has type "ndarray[tuple[Any, ...], dtype[Any]] | Array")  [assignment]
numpyro/distributions/transforms.py:371: error: Incompatible types in assignment (expression has type "Array | ndarray[tuple[Any, ...], dtype[Any]] | numpy.bool[builtins.bool] | number[Any, int | float | complex] | int | float | complex | Any", variable has type "ndarray[tuple[Any, ...], dtype[Any]] | Array")  [assignment]
numpyro/distributions/transforms.py:397: error: Incompatible types in assignment (expression has type "Array | ndarray[tuple[Any, ...], dtype[Any]] | numpy.bool[builtins.bool] | number[Any, int | float | complex] | int | float | complex | Any", variable has type "ndarray[tuple[Any, ...], dtype[Any]] | Array")  [assignment]
```

numpyro/distributions/transforms.py
Qazalbash: Done!

numpyro/_typing.py
fehiepsi: How about using ArrayLike in most places and in specific cases, use NumLike? StrictArray can be used at those places which expect non-scalar shapes.
```
StrictArray = Union[np.ndarray, jax.Array]
NumLike = Union[
    jax.Array,
    np.ndarray, np.number,
    bool, int, float, complex,
]
```
(we don't need to have T trailing in those cases I think)

numpyro/_typing.py
```diff
@@ -87,20 +96,25 @@ def is_discrete(self) -> bool: ...
...
-    codomain = ConstraintT
-    _inv: "TransformT" = None
+    domain: ConstraintT = ...
+    codomain: ConstraintT = ...
+    _inv: Optional["TransformT"] = ...
 
-    def __call__(self, x: ArrayLike) -> ArrayLike: ...
-    def _inverse(self, y: ArrayLike) -> ArrayLike: ...
+    def __call__(self, x: Union[jax.Array, Any]) -> Union[jax.Array, Any]: ...
+    def _inverse(self, y: Union[jax.Array, Any]) -> Union[jax.Array, Any]: ...
```
fehiepsi: prefer: ArrayLike and not use Any in general

numpyro/_typing.py
fehiepsi: Gotcha, thanks! How about using NumLike like in my comment above?

numpyro/distributions/transforms.py
```diff
@@ -67,36 +68,39 @@ def _clipped_expit(x: ArrayLike) -> ArrayLike:
...
-    def inv(self) -> TransformT:
+    def inv(self: TransformT) -> TransformT:
         inv = None
-        if self._inv is not None:
+        if (self._inv is not None) and isinstance(self._inv, weakref.ReferenceType):
             inv = self._inv()
         if inv is None:
-            inv = _InverseTransform(self)
-            self._inv = weakref.ref(inv)
+            inv = cast(TransformT, _InverseTransform(self))
```
fehiepsi: this seems unnecessary to me

numpyro/distributions/transforms.py
```diff
@@ -67,36 +68,39 @@ def _clipped_expit(x: ArrayLike) -> ArrayLike:
...
+            self._inv = cast(TransformT, weakref.ref(inv))
```
fehiepsi: why do we need this cast, it seems incorrect to me?

numpyro/distributions/transforms.py
```diff
@@ -173,12 +177,15 @@ def sign(self) -> ArrayLike:
...
         return self._inv._inverse(x)
 
     def log_abs_det_jacobian(
-        self, x: ArrayLike, y: ArrayLike, intermediates=None
-    ) -> ArrayLike:
+        self,
+        x: StrictArrayT,
+        y: StrictArrayT,
+        intermediates: Optional[PyTree] = None,
+    ) -> StrictArrayT:
```
fehiepsi: I believe you can use ArrayLike at those places as long as the Protocol uses the type hint: `log_abs_det_jacobian(...) -> NumLike:`

numpyro/distributions/transforms.py
```diff
@@ -191,7 +198,7 @@ def inverse_shape(self, shape: tuple[int, ...]) -> tuple[int, ...]:
     def tree_flatten(self):
         return (self._inv,), (("_inv",), dict())
 
-    def __eq__(self, other: TransformT) -> bool:
+    def __eq__(self, other: object) -> bool:
```
fehiepsi, line 202: just curious, why we allow to compare with non TransformT?

numpyro/distributions/transforms.py
```diff
@@ -268,10 +275,13 @@ def __call__(self, x: ArrayLike) -> ArrayLike:
         return self.loc + self.scale * x
 
     def _inverse(self, y: ArrayLike) -> ArrayLike:
-        return (y - self.loc) / self.scale
+        return (y - self.loc) / self.scale  # type: ignore[call-overload,operator]
```
fehiepsi: as long as `loc` and `scale` are `NumLike`, this should be fine I think.

numpyro/distributions/transforms.py
```diff
@@ -372,7 +385,7 @@ def log_abs_det_jacobian(
                     )
                 )
 
-        result = 0.0
+        result = jnp.zeros(())
```
fehiepsi: if output is ArrayLike, I guess we can use 0.0 here.

numpyro/distributions/transforms.py
```diff
@@ -389,9 +402,7 @@ def log_abs_det_jacobian(
         result = result + sum_rightmost(logdet, input_event_dim - part.domain.event_dim)
         return result
 
-    def call_with_intermediates(
-        self, x: ArrayLike
-    ) -> Tuple[ArrayLike, Optional[ArrayLike]]:
+    def call_with_intermediates(self, x: ArrayLike) -> Tuple[ArrayLike, Sequence]:
```
fehiepsi: Sequence -> PyTree

Qazalbash:
@juanitorduz I thought it would be easy to fix, but to my surprise, these changes are generating more errors than before. Would you like to take over this issue?

juanitorduz:
> @juanitorduz I thought it would be easy to fix, but to my surprise, these changes are generating more errors than before. Would you like to take over this issue?

hey, sure! What about if you incrementally push the easiest suggestions (that still work) until you face a problem and then we take it from there? 🙏

Qazalbash:
@juanitorduz I think I have fixed all! See 1e2e670

numpyro/distributions/transforms.py
fehiepsi: this cast seems wrong and the above cast seems unnecessary

numpyro/distributions/transforms.py
```diff
@@ -59,44 +60,47 @@
...
             inv = self._inv()
         if inv is None:
-            inv = _InverseTransform(self)
-            self._inv = weakref.ref(inv)
+            inv = cast(TransformT, _InverseTransform(self))
+            self._inv = cast(TransformT, weakref.ref(inv))
         return inv
 
-    def __call__(self, x: ArrayLike) -> ArrayLike:
+    def __call__(self, x: Union[NonScalarArray, Any]) -> Union[NonScalarArray, Any]:
```
fehiepsi: NonScalarArray -> NumLike

numpyro/distributions/transforms.py
```diff
@@ -59,44 +60,47 @@
...
         raise NotImplementedError
 
-    def _inverse(self, y: ArrayLike) -> ArrayLike:
+    def _inverse(self, y: Union[NonScalarArray, Any]) -> Union[NonScalarArray, Any]:
```
fehiepsi: NonScalarArray -> NumLike: we need to support scalar here

numpyro/distributions/transforms.py
```diff
@@ -59,44 +60,47 @@
...
+        y: Union[NonScalarArray, Any],
+        intermediates: Optional[PyTree] = None,
+    ) -> Union[NonScalarArray, Any]:
         raise NotImplementedError
 
     def call_with_intermediates(
-        self, x: ArrayLike
-    ) -> Tuple[ArrayLike, Optional[ArrayLike]]:
+        self, x: Union[NonScalarArray, Any]
+    ) -> Tuple[Union[NonScalarArray, Any], Optional[PyTree]]:
```
fehiepsi: NonScalarArray -> NumLike

numpyro/distributions/transforms.py
```diff
@@ -59,44 +60,47 @@
...
         raise NotImplementedError
 
     def log_abs_det_jacobian(
-        self, x: ArrayLike, y: ArrayLike, intermediates=None
-    ) -> ArrayLike:
+        self,
+        x: Union[NonScalarArray, Any],
+        y: Union[NonScalarArray, Any],
+        intermediates: Optional[PyTree] = None,
+    ) -> Union[NonScalarArray, Any]:
```
fehiepsi: NonScalarArray -> NumLike

numpyro/distributions/transforms.py
```diff
@@ -201,13 +208,13 @@ class AbsTransform(ParameterFreeTransform):
...
-    def __eq__(self, other: TransformT) -> bool:
+    def __eq__(self, other: object) -> bool:
         return isinstance(other, AbsTransform)
 
-    def __call__(self, x: ArrayLike) -> ArrayLike:
+    def __call__(self, x: NonScalarArray) -> NonScalarArray:
         return jnp.abs(x)
 
-    def _inverse(self, y: ArrayLike) -> ArrayLike:
+    def _inverse(self, y: NonScalarArray) -> NonScalarArray:
```
fehiepsi: abs should work for scalar: NonScalarArray -> NumLike

numpyro/distributions/transforms.py
```diff
@@ -201,13 +208,13 @@ class AbsTransform(ParameterFreeTransform):
     domain = constraints.real
     codomain = constraints.positive
 
-    def __eq__(self, other: TransformT) -> bool:
+    def __eq__(self, other: object) -> bool:
         return isinstance(other, AbsTransform)
 
-    def __call__(self, x: ArrayLike) -> ArrayLike:
+    def __call__(self, x: NonScalarArray) -> NonScalarArray:
```
fehiepsi: NonScalarArray -> NumLike

numpyro/distributions/transforms.py
```diff
@@ -608,15 +633,18 @@ def __eq__(self, other: TransformT) -> bool:
 class IdentityTransform(ParameterFreeTransform):
     sign = 1
 
-    def __call__(self, x: ArrayLike) -> ArrayLike:
+    def __call__(self, x: NonScalarArray) -> NonScalarArray:
```
fehiepsi: NonScalarArray -> NumLike

numpyro/distributions/transforms.py
```diff
@@ -608,15 +633,18 @@ def __eq__(self, other: TransformT) -> bool:
...
         return x
 
-    def _inverse(self, y: ArrayLike) -> ArrayLike:
+    def _inverse(self, y: NonScalarArray) -> NonScalarArray:
```
fehiepsi: NonScalarArray -> NumLike

numpyro/distributions/transforms.py
```diff
@@ -608,15 +633,18 @@ def __eq__(self, other: TransformT) -> bool:
...
         return y
 
     def log_abs_det_jacobian(
-        self, x: ArrayLike, y: ArrayLike, intermediates=None
-    ) -> ArrayLike:
+        self,
+        x: NonScalarArray,
+        y: NonScalarArray,
```
fehiepsi: NonScalarArray -> NumLike

numpyro/distributions/transforms.py
```diff
@@ -638,26 +666,29 @@ def __init__(
...
     @property
-    def codomain(self) -> constraints.Constraint:
+    def codomain(self) -> ConstraintT:  # type: ignore[override]
         return constraints.independent(
             self.base_transform.codomain, self.reinterpreted_batch_ndims
-        )
+        )  # type: ignore[return-value]
 
-    def __call__(self, x: ArrayLike) -> ArrayLike:
+    def __call__(self, x: NonScalarArray) -> NonScalarArray:
```
fehiepsi: we allow `reinterpreted_batch_ndims=0`, so I guess NonScalarArray -> NumLike

numpyro/distributions/transforms.py
```diff
@@ -638,26 +666,29 @@ def __init__(
...
         return self.base_transform(x)
 
-    def _inverse(self, y: ArrayLike) -> ArrayLike:
+    def _inverse(self, y: NonScalarArray) -> NonScalarArray:
```
fehiepsi: NonScalarArray -> NumLike

numpyro/distributions/transforms.py
```diff
@@ -638,26 +666,29 @@ def __init__(
...
         return self.base_transform._inverse(y)
 
     def log_abs_det_jacobian(
-        self, x: ArrayLike, y: ArrayLike, intermediates=None
-    ) -> ArrayLike:
+        self,
+        x: NonScalarArray,
+        y: NonScalarArray,
```
fehiepsi: NonScalarArray -> NumLike

fehiepsi (approved):
The changes look great! I just have minor comments. Thank you so much!

fehiepsi:
I still cant see that comment. Maybe a github bug. Let me look into the remaining issues then.

Qazalbash:
Please take a look at the reply of the [comment](https://github.com/pyro-ppl/numpyro/pull/2066#discussion_r2386376893).

fehiepsi:
Thanks @Qazalbash! Though I still can't see that reply (it is invisible to me), I understand your point now. I just pushed a commit that using Generic type for Transform. That way we can declare array constraints in the subclass. Let me know if you have any opinion about the changes.

Qazalbash:
Thanks @fehiepsi, for looking into it. I see you also have put types in `numpyro.distributions.constraints` module, I think we should also update the PR title accordingly.

PR #2067, opened 2025-09-04
Ensure compatibility NNX 0.11.2

juanitorduz:
Ensures compatibility with https://github.com/google/flax/releases/tag/v0.11.2

The tests were failing in https://github.com/pyro-ppl/numpyro/actions/runs/17463007632/job/49592472608

```python
FAILED test/contrib/test_module.py::test_nnx_state_dropout_smoke[batchnorm-dropout] - flax.errors.TraceContextError: Cannot mutate BatchStat from a different trace level (https://flax.readthedocs.io/en/latest/api_reference/flax.errors.html#flax.errors.TraceContextError)
--------------------
For simplicity, JAX has removed its internal frames from the traceback of the following exception. Set JAX_TRACEBACK_FILTERING=off to include these.
FAILED test/contrib/test_module.py::test_nnx_state_dropout_smoke[batchnorm-no_dropout] - flax.errors.TraceContextError: Cannot mutate BatchStat from a different trace level (https://flax.readthedocs.io/en/latest/api_reference/flax.errors.html#flax.errors.TraceContextError)
--------------------
For simplicity, JAX has removed its internal frames from the traceback of the following exception. Set JAX_TRACEBACK_FILTERING=off to include these.
= 2 failed, 506 passed, 66 skipped, 4 xfailed, 36 warnings in 654.96s (0:10:54)
```

PR #2069, opened 2025-09-05
fix(tests): fix `stderr` calculation in `test/test_distributions::test_entropy_samples` to handle zero standard deviation

Qazalbash:
```
FAILED test/test_distributions.py::test_entropy_samples[Beta-<lambda>-params8] - assert (Array(False, dtype=bool) or Array(False, dtype=bool))
FAILED test/test_distributions.py::test_entropy_samples[BetaProportion-<lambda>-params11] - assert (Array(False, dtype=bool) or Array(False, dtype=bool))
```

These tests were failing due to a `stderr` being zero. I have fixed it by taking a very small number if there is a zero.

### DIFF
--- a/test/contrib/test_module.py
+++ b/test/contrib/test_module.py
@@ -385,6 +385,9 @@ def test_nnx_module():
 @pytest.mark.parametrize(
     argnames="batchnorm", argvalues=[True, False], ids=["batchnorm", "no_batchnorm"]
 )
+@pytest.mark.xfail(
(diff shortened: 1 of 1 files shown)


Qazalbash:
Failing test cases are fixed in #2067.

fehiepsi:
Thanks @Qazalbash! Could we merge the fixes or mark the current failing tests by xfail so that we can address them in the other PR?

fehiepsi (approved):
Thanks for the fix, @Qazalbash!

PR #2076, opened 2025-09-26
Fix test for new NNX 0.12.0

juanitorduz:
See https://github.com/google/flax/releases/tag/v0.12.0

The tests were failing in https://github.com/pyro-ppl/numpyro/pull/2074


fehiepsi:
Thanks Juan!

Issue #2064, opened 2025-08-21
get_model_relations fails when parameters are set with lambda functions


mochar:
### Bug Description
`get_model_relations` fails when parameters values are set to be lambda functions, see reproduce step below.

There is some logic  in `get_model_relations` to handle this problem in sample and deterministic cases, but not for the param case:

```
        # Work around an issue where jax.eval_shape does not work
        # for distribution output (e.g. the function `lambda: dist.Normal(0, 1)`)
        # Here we will remove `fn` and store its name in the trace.
        for name, site in trace.items():
            if site["type"] == "sample":
                site["fn_name"] = _get_dist_name(site.pop("fn"))
            elif site["type"] == "deterministic":
                site["fn_name"] = "Deterministic"

```

### Steps to Reproduce
```python
from numpyro.infer.inspect import get_model_relations
import numpyro
def guide():
    numpyro.param('p', lambda _: 1.)
get_model_relations(guide)
```

Error:

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[14], line 1
----> 1 get_model_relations(g)

File ~/.venv/lib/python3.13/site-packages/numpyro/infer/inspect.py:326, in get_model_relations(model, model_args, model_kwargs)
    323     return PytreeTrace(trace)
    325 # We use eval_shape to avoid any array computation.
--> 326 trace = jax.eval_shape(get_trace).trace
    327 obs_sites = [
    328     name
    329     for name, site in trace.items()
    330     if site["type"] == "sample" and site["is_observed"]
    331 ]
    332 sample_dist = {
    333     name: site["fn_name"]
    334     for name, site in trace.items()
    335     if site["type"] in ["sample", "deterministic"]
    336 }

    [... skipping hidden 11 frame]

File ~/.venv/lib/python3.13/site-packages/jax/_src/interpreters/partial_eval.py:2316, in _check_returned_jaxtypes(dbg, out_tracers)
   2314 else:
   2315   extra = ''
-> 2316 raise TypeError(
   2317 f"function {dbg.func_src_info} traced for {dbg.traced_for} returned a "
   2318 f"value of type {type(x)}{extra}, which is not a valid JAX type") from None

TypeError: function get_trace at /home/mochar/.venv/lib/python3.13/site-packages/numpyro/infer/inspect.py:307 traced for jit returned a value of type <class 'function'> at output component [0]['p']['args'][0], which is not a valid JAX type
```

[...]

Issue #2077, opened 2025-09-30
get_dependencies does not behave as expected with eqx_module


rayisaacalan:
### Bug Description
When replicating the "NNX and NumPyro Integration" example using equinox as a drop-in replacement for NNX, the model can be fit and behaves as expected for inference but cannot be visualized with render_model due to a TypeError in get_dependencies.

### Steps to Reproduce

```
import jax
import jax.numpy as jnp
from jax import random
import jax.tree_util as jtu
import equinox as eqx

import numpyro
from numpyro.contrib.module import eqx_module, random_eqx_module
import numpyro.distributions as dist

rng_key = random.PRNGKey(seed=42)

n = 32 * 10
rng_key, rng_subkey = random.split(rng_key)
x = jnp.linspace(1, jnp.pi, n)
x_train = x[..., None]

class LocMLP(eqx.Module):
    """3-layer Multi-layer perceptron for the mean."""
    linear1: eqx.nn.Linear
    linear2: eqx.nn.Linear
    linear3: eqx.nn.Linear

    def __init__(self, din: int, dmid: int, dout: int, *, key):
        key1, key2, key3 = random.split(key, 3)
        self.linear1 = eqx.nn.Linear(din, dmid, key=key1)
        self.linear2 = eqx.nn.Linear(dmid, dmid, key=key2)
        self.linear3 = eqx.nn.Linear(dmid, dout, key=key3)

    def __call__(self, x):
        x = self.linear1(x)
        x = jax.nn.sigmoid(x)
        x = self.linear2(x)
        x = jax.nn.sigmoid(x)
        x = self.linear3(x)
        return x


class ScaleMLP(eqx.Module):
    """Single-layer MLP for the standard deviation."""
    linear: eqx.nn.Linear

    def __init__(self, *, key) -> None:
        self.linear = eqx.nn.Linear(1, 1, key=key)
    def __call__(self, x):
        x = self.linear(x)
        return jax.nn.softplus(x)

rng_key, mu_key, sigma_key = random.split(rng_key, 3)
mu_nn_module = LocMLP(din=1, dmid=8, dout=1, key=mu_key)
sigma_nn_module = ScaleMLP(key=sigma_key)
[jtu.keystr(path)[1:] for path, _ in jtu.tree_leaves_with_path(sigma_nn_module)]
```

> ['linear.weight', 'linear.bias']

```
def model(x):
    mu_nn = eqx_module("mu_nn", mu_nn_module)
[...]

PR #2078, opened 2025-09-30
Fix get model relations when a param is a lambda function

juanitorduz:
Closes: https://github.com/pyro-ppl/numpyro/issues/2064 and https://github.com/pyro-ppl/numpyro/issues/2077 🤞


juanitorduz:
I ran the example code provided in https://github.com/pyro-ppl/numpyro/issues/2077 and it works:

image

Issue #1909, opened 2024-11-15
survival analysis / time-to-event analysis in numpyro


vanAmsterdam:
in my research domain (healthcare) many outcomes are right-censored (aka survival analysis / time-to-event); the data for an individual patient are basically a tuple of (observed_time, event / censoring indicator).

What would be a good way to incorporate this into numpyro? There are several distributions already implemented in the library that are often used for (bayesian) survival analysis (e.g. exponential, weibull), but calculating the likelihood correctly for censored data would require some extra work.

are others interested in adding support for survival analysis to numpyro?

an example implemented in problow is here: https://probflow.readthedocs.io/en/stable/examples/time_to_event.html

juanitorduz:
@vanAmsterdam There is an example where this is done by hand https://num.pyro.ai/en/latest/tutorials/censoring.html (motivated by the code in https://kylejcaron.github.io/posts/censored_demand/2024-02-06-censored-demand.html) . Would that be enough? Or are you looking into something like the PyMC wrapper https://www.pymc.io/projects/docs/en/latest/api/distributions/censored.html ?

vanAmsterdam:
e.g. we might also think about it as an effect handler just like mask, but then a censored variant, and even potentially a truncated variant

so the interface could be something like

```python
t = jnp.array([.72, 1.23, .45])
censored = jnp.array([True, False, True])

with handlers.censored(censored):
   sample('obs_t', dist.Normal(), obs=t)

```

btw stan docs also have some nice background information on this topic: https://mc-stan.org/docs/stan-users-guide/truncation-censoring.html#censored-data

juanitorduz:
> pymc approach to numpyro if we think this is a worthwhile
IMO, that would be great! This can enable users to use this more directly (also creating awareness about it). 

But I guess @fehiepsi and his team are the ones to decide whether this is something they want :)

fehiepsi:
Yeah,we welcome the contribution.

milesalanmoore:
@vanAmsterdam Any progress on this? I'm interested in participating if you're still working on this! I've been fitting (thresholding) survival models in Stan for my own work and agree this model class would be great in numpyro.

vanAmsterdam:
update: I've now implemented Right / Left / Interval and Doubly censored distributions; you can check the files changed here: https://github.com/vanAmsterdam/numpyro/pull/1

a couple of tests are still failing and the documentation needs updating

juanitorduz:
[...]

PR #2081, opened 2025-10-06
Implement LeftCensoredDistribution and RightCensoredDistribution

vanAmsterdam:
Implements `LeftCensoredDistribution` and `RightCensoredDistribution`

These distributions take any `base_dist` with a `cdf` method implemented and allow for handling left-or-right censored data. See Issue #1909.

Current status

- [x] implementation
- [x] standard tests
- [ ] censored distribution specific tests
- [ ] updated tutorial / notebook

### DIFF
--- a/numpyro/distributions/censored.py
+++ b/numpyro/distributions/censored.py
@@ -1,0 +1,217 @@
+# Copyright Contributors to the Pyro project.
+# SPDX-License-Identifier: Apache-2.0
+
+
+from typing import Optional
+
+import jax
+from jax import lax
+import jax.numpy as jnp
+from jax.typing import ArrayLike
+
+from numpyro._typing import ConstraintT, DistributionT
+from numpyro.distributions import constraints
+from numpyro.distributions.distribution import Distribution
+from numpyro.distributions.util import (
+    promote_shapes,
+    validate_sample,
+)
+
+
+class LeftCensoredDistribution(Distribution):
+    r"""
+    Distribution wrapper for left-censored outcomes.
+
+    This distribution augments a base distribution with left-censoring,
+    so that the likelihood contribution depends on the censoring indicator.
+
+    Parameters
+    ----------
+    base_dist : numpyro.distributions.Distribution
+        Parametric distribution for the *uncensored* values
+        (e.g., Exponential, Weibull, LogNormal, Normal, etc.).
+        This distribution must implement a `cdf` method.
+    censored : array-like of {0,1}
+        Censoring indicator per observation:
+        - 0 → value is observed exactly
+        - 1 → observation is left-censored at the reported value
+        (true value occurred *on or before* the reported value)
+
+    Notes
+    -----
+    - The `log_prob(value)` method expects `value` to be the observed upper bound
+      for each observation. The contribution to the log-likelihood is:
+
+        log f(value)    if censored == 0
+        log F(value)    if censored == 1
+
+    where f is the density and F the cumulative distribution function of `base_dist`.
+
+    - This is commonly used in survival analysis, where event times are positive,
+      but the approach is more general and can be applied to any distribution
+      with a cumulative distribution function, regardless of support.
+
+    - In R's **survival** package notation, this corresponds to
+    `Surv(time, event, type = "left")`.
+
+        Example:
+        `Surv(time = c(2, 4, 6), event = c(0, 1, 0), type="left")`
+        means:
+        * subject 1 had an event exactly at t=2
+        * subject 2 had an event before or at t=4 (left-censored)
+        * subject 3 had an event exactly at t=6
+
+    Examples
+    --------
+    >>> base = dist.LogNormal(0., 1.)
+    >>> surv_dist = LeftCensoredDistribution(base, censored=jnp.array([0, 1, 1]))
+    >>> loglik = surv_dist.log_prob(jnp.array([2., 4., 6.]))
+    # loglik[0] uses density at 2
+    # loglik[1] uses CDF at 4
+    # loglik[2] uses CDF at 6
+    """
+
+    arg_constraints = {"censored": constraints.boolean}
+    reparametrized_params = ["censored"]
+    pytree_data_fields = ("base_dist", "censored", "_support")
+
+    def __init__(
+        self,
+        base_dist: DistributionT,
+        censored: ArrayLike = False,
+        *,
+        validate_args: Optional[bool] = None,
+    ):
+        # test if base_dist has an implemented cdf method
+        assert hasattr(base_dist, "cdf")
+        batch_shape = lax.broadcast_shapes(base_dist.batch_shape, jnp.shape(censored))
+        self.base_dist: DistributionT = jax.tree.map(
+            lambda p: promote_shapes(p, shape=batch_shape)[0], base_dist
+        )
+        self.censored = jnp.array(
+            promote_shapes(censored, shape=batch_shape)[0], dtype=jnp.bool
+        )
+        self._support = base_dist.support
+        super().__init__(batch_shape, validate_args=validate_args)
+
+    def sample(
+        self, key: jax.dtypes.prng_key, sample_shape: tuple[int, ...] = ()
+    ) -> ArrayLike:
+        return self.base_dist.sample(key, sample_shape)
+
+    @constraints.dependent_property(is_discrete=False, event_dim=0)
+    def support(self) -> ConstraintT:
+        return self._support
+
+    @validate_sample
+    def log_prob(self, value: ArrayLike) -> ArrayLike:
+        minval = 1e-10
+        # minval = jnp.finfo(value).tiny
+
+        def logF(x):
+            # log(F(x)) with stability
+            return jnp.log(jnp.clip(self.base_dist.cdf(x), minval, 1.0))
+
+        return jnp.where(
+            self.censored,
+            logF(value),  # left-censored observations: log F(t)
+            self.base_dist.log_prob(value),  # observed values: log f(t)
+        )
+
+
+class RightCensoredDistribution(Distribution):
+    r"""
+    Distribution wrapper for right-censored outcomes.
+
+    This distribution augments a base distribution with right-censoring,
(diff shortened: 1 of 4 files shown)


numpyro/distributions/censored.py
```diff
@@ -0,0 +1,216 @@
...
+    ) -> ArrayLike:
+        return self.base_dist.sample(key, sample_shape)
+
+    @constraints.dependent_property(is_discrete=False, event_dim=0)
+    def support(self) -> ConstraintT:
+        return self._support
+
+    @validate_sample
+    def log_prob(self, value: ArrayLike) -> ArrayLike:
+        minval = 1e-7
```
vanAmsterdam: this minval was needed to not get -inf in the subsequent `log(1-cdf(x))` calculation for observations with very high tail censored observations; this is what I'm most unsure about

juanitorduz: Maybe using something like 

```python
    dtype = jnp.result_type(value, float)
    eps = jnp.finfo(dtype).eps
    minval = jnp.sqrt(eps)
```
?

numpyro/distributions/censored.py
```diff
@@ -0,0 +1,216 @@
...
+
+    def __init__(
+        self,
+        base_dist: DistributionT,
+        censored: ArrayLike = False,
+        *,
+        validate_args: Optional[bool] = None,
+    ):
+        # test if base_dist has an implemented cdf method
+        assert hasattr(base_dist, "cdf")
```
juanitorduz: Maybe a more informative error:

```python
    if not hasattr(base_dist, "cdf"):
        raise TypeError(
            f"{type(base_dist).__name__} does not have a 'cdf' method. "
            "Censored distributions require a base distribution with an "
            "implemented cumulative distribution function."
        )
    
    # Optionally test that cdf actually works (in validate_args mode)
    if validate_args:
        try:
            test_val = base_dist.support.feasible_like(jnp.array(0.))
            _ = base_dist.cdf(test_val)
        except (NotImplementedError, AttributeError) as e:
            raise TypeError(
                f"{type(base_dist).__name__}.cdf() is not properly implemented."
            ) from e
```
?

numpyro/distributions/censored.py
```diff
@@ -0,0 +1,216 @@
...
+        )
+        self.censored = jnp.array(
+            promote_shapes(censored, shape=batch_shape)[0], dtype=jnp.bool
+        )
+        self._support = base_dist.support
+        super().__init__(batch_shape, validate_args=validate_args)
+
+    def sample(
+        self, key: jax.dtypes.prng_key, sample_shape: tuple[int, ...] = ()
+    ) -> ArrayLike:
```
juanitorduz, line 114: maybe improve type hints ? Distributions have sample signature:

```python
    def sample(
        self, key: Optional[jax.dtypes.prng_key], sample_shape: tuple[int, ...] = ()
    ) -> ArrayLike:
```

Currently, MyPy is not checking these files, if you want them to be checked (and feel like fighting with mypy), you can add the "numpyro.distributions.censored.*", module in https://github.com/pyro-ppl/numpyro/blob/8fec8a1444adbb42c1a409f1fffdfabb63eaae09/pyproject.toml#L120

numpyro/distributions/censored.py
```diff
@@ -0,0 +1,216 @@
...
+from numpyro._typing import ConstraintT, DistributionT
+from numpyro.distributions import constraints
+from numpyro.distributions.distribution import Distribution
+from numpyro.distributions.util import (
+    promote_shapes,
+    validate_sample,
+)
+
+
+class LeftCensoredDistribution(Distribution):
```
juanitorduz, line 21: What about using a base class as 

```python
class _BaseCensoredDistribution(Distribution):
    """Base class for censored distributions."""
    
    def __init__(self, base_dist, censored, *, validate_args=None):
        if not hasattr(base_dist, "cdf"):
            raise TypeError(f"{type(base_dist).__name__} must have a cdf method")
        
        batch_shape = lax.broadcast_shapes(base_dist.batch_shape, jnp.shape(censored))
        self.base_dist = jax.tree.map(
            lambda p: promote_shapes(p, shape=batch_shape)[0], base_dist
        )
        self.censored = jnp.array(
            promote_shapes(censored, shape=batch_shape)[0], dtype=jnp.bool
        )
        self._support = base_dist.support
        super().__init__(batch_shape, validate_args=validate_args)
```

numpyro/distributions/censored.py
```diff
@@ -0,0 +1,216 @@
...
+
+    @constraints.dependent_property(is_discrete=False, event_dim=0)
+    def support(self) -> ConstraintT:
+        return self._support
+
+    @validate_sample
+    def log_prob(self, value: ArrayLike) -> ArrayLike:
+        minval = 1e-7
+
+        def logS(x):
```
juanitorduz: Suggestion:

The internal functions `logF` and `logS` are concise but could be more descriptive:
- `logF` → `log_cdf_censored`
- `logS` → `log_survival_censored`

test/test_distributions.py
```diff
@@ -32,6 +32,10 @@
     transforms,
 )
 from numpyro.distributions.batch_util import vmap_over
+from numpyro.distributions.censored import (
+    LeftCensoredDistribution,
```
juanitorduz, line 36: Maybe add some additional custom tests

a) **Gradient tests**: Ensure gradients flow correctly through censored observations
```python
def test_censored_gradient():
    def loss(loc):
        base_dist = dist.Normal(loc, 1.0)
        censored_dist = dist.LeftCensoredDistribution(base_dist, censored=1)
        return -censored_dist.log_prob(2.0)
    
    grad_fn = jax.grad(loss)
    grad = grad_fn(0.0)
    assert jnp.isfinite(grad)
```

b) **Batch dimension broadcasting**: More comprehensive tests with mixed batch shapes
```python
def test_censored_broadcasting():
    # loc: (3,), censored: (2, 1) -> batch_shape: (2, 3)
    base_dist = dist.Normal(jnp.array([0., 1., 2.]), 1.0)
    censored = jnp.array([[0], [1]])
    dist_obj = dist.LeftCensoredDistribution(base_dist, censored)
    assert dist_obj.batch_shape == (2, 3)
```

c) **Sampling correctness**: Verify that samples from censored distributions are valid
```python
def test_censored_sample_validity():
    base_dist = dist.Normal(0., 1.)
    censored_dist = dist.RightCensoredDistribution(base_dist, censored=0)
    samples = censored_dist.sample(random.PRNGKey(0), (1000,))
    # Samples should match base distribution when censored=0
    assert samples.shape == (1000,)
```

numpyro/distributions/censored.py
vanAmsterdam: thanks for this suggestion; I had something similar in before but it failed for in32 dtypes (for discrete distributions like poisson), but jnp.result_type fixed this!

numpyro/distributions/censored.py
vanAmsterdam: I copied this suggestion in the new code

numpyro/distributions/censored.py
vanAmsterdam: I'm happy to use your proposed type hint, but I don't see it used consistently in the other distributions? e.g. continuous.py or discrete.py?

numpyro/distributions/censored.py
vanAmsterdam: the init signature for IntervalCensored is different (has a left_censored and right_censored) argument; maybe defer this base class to when it's clear how to implement intervalcensored?

another possibility would be to create a 'flexible' CensoredDistribution class where the user can specify left_censored and/or right_censored and the result will be whatever is specified. In fact, the IntervalCensoredDistribution implementation does this already as observations can be left, right, interval or double censored

numpyro/distributions/censored.py
juanitorduz: I know, we are working on it :/ https://github.com/pyro-ppl/numpyro/issues (that is for now I suggest this as optional)

numpyro/distributions/censored.py
juanitorduz: So maybe for now we leave it as it is and wait for the `IntervalCensoredDistribution` in a following PR?

test/test_distributions.py
vanAmsterdam: thanks for these suggestions on additional tests; I've adapted them and added them

numpyro/distributions/censored.py
vanAmsterdam: I'd say yes

vanAmsterdam:
> You can try ruining `make doctest` locally to make the debugging easier :)

yes figured this one out a little too late :) but it's passing now so I think we're set for a review. thanks for the help!

fehiepsi (changes requested):
Actually, I have some minor  comments.

numpyro/distributions/censored.py
```diff
@@ -0,0 +1,251 @@
...
+        self.censored = jnp.array(
+            promote_shapes(censored, shape=batch_shape)[0], dtype=jnp.bool
+        )
+        self._support = base_dist.support
+        super().__init__(batch_shape, validate_args=validate_args)
+
+    def sample(
+        self, key: Optional[jax.dtypes.prng_key], sample_shape: tuple[int, ...] = ()
+    ) -> ArrayLike:
+        return self.base_dist.sample(key, sample_shape)
```
fehiepsi: I think you need `self.base_dict.expand(self.batch_shape).sample(...)`.

numpyro/distributions/censored.py
```diff
@@ -0,0 +1,251 @@
...
```
fehiepsi: You'll need self.base_dict.expand(self.batch_shape).sample(...). promote_shapes does not broadcast the parameters, it only adds some missing singleton dimensions.

vanAmsterdam: good catch; updated in newest commit and also added a test

fehiepsi (approved):

PR #2085, opened 2025-10-21
fix(gh-2036): MyPy Errors in `numpyro.distributions.constraints` Module

Qazalbash:
This PR contains the resolution of mypy errors passed by https://github.com/pyro-ppl/numpyro/pull/2032, in the `numpyro.distributions.constraints` module.

I have tried to replicate the same solution proposed by @fehiepsi in #2066, i.e., the use of generics (see from 69c1ed5eeeb9ddc0841f723d7089cc51a0cd5a16 till 1d6b24df2a340efa6118910043a821a7dbdafcfb).

I have slightly modified the logic. Some notes on them,

1. `ArrayLike` contains `complex` and there is no partial order over complex numbers. MyPy was throwing errors for `>`, `<`, `<=`, and `>=` operators. They have been replaced with the equivalent `jax.numpy` function.
2. There is no mod operation between arrays and integers; it has been replaced with [`jax.numpy.mod`](https://docs.jax.dev/en/latest/_autosummary/jax.numpy.mod.html#jax.numpy.mod).
3. Bitwise operations have been replaced with [`jax.numpy.logical_and`](https://docs.jax.dev/en/latest/_autosummary/jax.numpy.logical_and.html#jax.numpy.logical_and) and [`jax.numpy.logical_or`](https://docs.jax.dev/en/latest/_autosummary/jax.numpy.logical_or.html).
4. The type of the argument in the `__eq__` method has been changed to `object` because the method can take any type of object; its implementation is to classify if the objects are the same or not. I have added `if not isinstance(other, ...): return False` statement at some places due to MyPy's errors.

I tackled with following problems that, in my understanding, require some discussion,

1. `event_dim` and `is_discrete` are read-only properties that are modified by some constraints; therefore, I made them private attributes and introduced getter and setter methods for each.
2. `__eq__` method expects return type to be a `bool`, but we are returning arrays of booleans. I have marked them to be ignored by MyPy.
3. We can not ducktype the constraint object with `ConstraintT` at the end of the module, because the `ConstraintT` object expects a `NumLike` object, but some constraints support only `NonScalarArray`. This issue is solvable via a generic typing protocol (see changes of c12b51439b44b9b7fb4a7f0eaebac3f9c1149876). It can also be seen with `TransformT` and subclasses of `Transform`, for statement `transform_obj: TransformT = TransformClass(...)`, MyPy will throw an error, if `TransformClass` uses anything other than `NumLike`. This issue is also addressable via a generic typing protocol.
4. `jax>=0.7.2` has introduced `TypedNdArray` to represent constants in jaxpr (ref https://github.com/jax-ml/jax/issues/31989, https://github.com/jax-ml/jax/pull/32227). It is also a part of `ArrayLike` type, and has no `reshape` method.

These are all the major outlines of this PR. I will update the description if I recall any.


numpyro/_typing.py
```diff
@@ -36,6 +37,9 @@
 """A generic type for a pytree, i.e. a nested structure of lists, tuples, dicts, and arrays."""
 
 
+NumLikeT = TypeVar("NumLikeT", bound=NumLike)
```
fehiepsi, line 43: let's move this to constraints.py because it's not intended to be used outside of that class.

numpyro/distributions/constraints.py
```diff
@@ -73,43 +73,59 @@
...
     Abstract base class for constraints.
 
     A constraint object represents a region over which a variable is valid,
     e.g. within which a variable can be optimized.
     """
 
-    is_discrete = False
-    event_dim = 0
+    _is_discrete = False
+    _event_dim = 0
```
fehiepsi: you can make them properties. Then they can be attributes in the subclasses.

numpyro/distributions/constraints.py
```diff
@@ -242,21 +258,22 @@ def __call__(
...
-            and self._is_discrete == other._is_discrete
+            self._is_discrete == other._is_discrete
             and self._event_dim == other._event_dim
         )
 
     def tree_flatten(self):
         return (), (
             (),
-            dict(_is_discrete=self._is_discrete, _event_dim=self._event_dim),
+            dict(is_discrete=self._is_discrete, event_dim=self._event_dim),
```
fehiepsi: using `_is_discrete` and `_event_dim` looks correct to me

numpyro/distributions/constraints.py
```diff
@@ -364,7 +375,9 @@ def __call__(self, value: ArrayLike) -> ArrayLike:
             raise ValueError(
                 f"Expected value.dim() >= {expected} but got {jax.numpy.ndim(value)}"
             )
-        result = result.reshape(
+        # jax>=0.7.2 introduced `TypedNdArray` to represent constants in jaxpr, and they
+        # have no reshape method.
+        result = result.reshape(  # type: ignore[union-attr]
```
fehiepsi: you can use jnp.reshape(...) I think

numpyro/distributions/constraints.py
```diff
@@ -140,20 +156,20 @@ def __new__(cls):
...
 
-class _Boolean(_SingletonConstraint):
-    is_discrete = True
+class _Boolean(_SingletonConstraint[NumLike]):
+    _is_discrete = True
 
-    def __call__(self, x: ArrayLike) -> ArrayLike:
-        return (x == 0) | (x == 1)
+    def __call__(self, x: NumLike) -> ArrayLike:
+        return jnp.logical_or(jnp.equal(x, 0), jnp.equal(x, 1))
```
fehiepsi: the main reason to use `(x == 0) | (x == 1)` is to avoid compiling the computation (so they can be constant under jit-compiling). It was helpful back to the old jax but i'm not sure if jax has improved since then. How about using the pattern `jnp = jax.numpy if isinstance(x, jax.Array) else np`?

numpyro/distributions/constraints.py
```diff
@@ -579,23 +610,27 @@ def __call__(self, x: NonScalarArray) -> ArrayLike:
             jnp.reshape(tril == x, x.shape[:-2] + (-1,)), axis=-1
         )
         positive_diagonal = jnp.all(jnp.diagonal(x, axis1=-2, axis2=-1) > 0, axis=-1)
-        return lower_triangular & positive_diagonal
+        return jnp.logical_and(lower_triangular, positive_diagonal)
```
fehiepsi: since you use `xp` at other places, could you use it here instead?

numpyro/distributions/constraints.py
```diff
@@ -569,8 +600,8 @@ def __repr__(self) -> str:
         return fmt_string
 
 
-class _LowerCholesky(_SingletonConstraint):
-    event_dim = 2
+class _LowerCholesky(_SingletonConstraint[NonScalarArray]):
+    _event_dim = 2
```
fehiepsi: just curious: is there typing error if we do `event_dim=2` here? Using `event_dim=2` is more explicit I think

numpyro/distributions/constraints.py
```diff
@@ -75,41 +75,51 @@
 
 from numpyro._typing import ConstraintT, NonScalarArray, NumLike
 
+NumLikeT = TypeVar("NumLikeT", bound=NumLike)
```
fehiepsi: I forgot that this is also used in transforms. Could you move this back to typing? Sorry!

numpyro/distributions/constraints.py
```diff
@@ -398,22 +407,27 @@ def __eq__(self, other: ConstraintT) -> bool:
...
+class _RealVector(
+    _IndependentConstraint[NumLike], _SingletonConstraint[NonScalarArray]
+):
     def __init__(self) -> None:
         super().__init__(_Real(), 1)
 
 
-class _RealMatrix(_IndependentConstraint, _SingletonConstraint):
+class _RealMatrix(
+    _IndependentConstraint[NumLike], _SingletonConstraint[NonScalarArray]
```
fehiepsi: can we use _IndependentConstraint[NonScalarArray] here

numpyro/distributions/constraints.py
```diff
@@ -426,31 +440,38 @@ def feasible_like(self, prototype: NumLike) -> NumLike:
...
+class _IntegerInterval(Constraint[NumLike]):
+    _is_discrete = True
 
     def __init__(self, lower_bound: NumLike, upper_bound: NumLike) -> None:
         self.lower_bound = lower_bound
         self.upper_bound = upper_bound
 
     def __call__(self, x: NumLike) -> ArrayLike:
-        return (x >= self.lower_bound) & (x <= self.upper_bound) & (x % 1 == 0)
+        return jnp.logical_and(
```
fehiepsi: pls use `xp` here

numpyro/distributions/constraints.py
Qazalbash: Its okay.

numpyro/distributions/constraints.py
Qazalbash: I thought it would cause problem, but its not.

numpyro/distributions/constraints.py
Qazalbash: I have tried all combinations, None is working. Although I am pushing one with both `NonScalarArray`.

Qazalbash:
Only errors left here are coming from the statments `constrain_obj: ConstraintT = ConstraintClass(...)`, when `ConstraintClass` uses `NonScalarArray` type. Because `ConstraintT` has `NumLike` and expects similar from `ConstraintClass`. Same problem can be seen in `TransformT`.

I tried to address this issue by making typing protocols generic. I later reverted it. They are available at c12b51439b44b9b7fb4a7f0eaebac3f9c1149876.

numpyro/distributions/constraints.py
```diff
@@ -698,24 +710,31 @@ class _PositiveOrderedVector(_SingletonConstraint):
     event_dim = 1
 
     def __call__(self, x: NonScalarArray) -> ArrayLike:
-        return ordered_vector.check(x) & independent(positive, 1).check(x)
+        return jnp.logical_and(
+            ordered_vector.check(x), independent[NumLike](positive, 1).check(x)
```
Qazalbash: I think `isdependent[NumLike](...)` will not work for Python < 3.12, am I right?

fehiepsi: not sure but I think we can use `jnp.all(positive.check(x), -1)`

numpyro/distributions/constraints.py
```diff
@@ -799,6 +820,7 @@ def tree_flatten(self):
 # TODO: Make types consistent
 # See https://github.com/pytorch/pytorch/issues/50616
 
+
 boolean: ConstraintT = _Boolean()
```
fehiepsi: I think you can use `Constraint` here and below.

numpyro/distributions/constraints.py
```diff
@@ -301,40 +302,41 @@ def feasible_like(self, prototype: NumLike) -> NumLike:
...
-class _IndependentConstraint(Constraint):
+class _IndependentConstraint(Constraint[NumLikeT]):
     """
     Wraps a constraint by aggregating over ``reinterpreted_batch_ndims``-many
     dims in :meth:`check`, so that an event is valid only if all its
     independent entries are valid.
     """
 
-    def __init__(self, base_constraint, reinterpreted_batch_ndims):
+    def __init__(self, base_constraint: ConstraintT, reinterpreted_batch_ndims: int):
```
fehiepsi: maybe using `Constraint` here?

numpyro/distributions/constraints.py
```diff
@@ -242,31 +241,32 @@ def __call__(
...
     def __init__(
         self, fn=None, *, is_discrete=NotImplemented, event_dim=NotImplemented
     ):
         super().__init__(fn)
         self._is_discrete = is_discrete
-        self._event_dim = event_dim
+        self.event_dim = event_dim
 
-    def __call__(self, x: ArrayLike) -> ArrayLike:
+    def __call__(self, x: NumLikeT) -> ArrayLike:
```
fehiepsi: x here could be a function, and the return value could be a function. Maybe ignore type here
```
def __call__(self, x):
```

fehiepsi (approved):
Thanks @Qazalbash! Sorry for the delay! I thought the PR was WIP.

numpyro/distributions/constraints.py
Qazalbash: It is causing problems, but `Union[Constraint, ConstraintT]` works.

numpyro/distributions/constraints.py
```diff
@@ -242,31 +241,32 @@ def __call__(
...
+class dependent_property(property, _Dependent[NumLikeT]):
     # XXX: this should not need to be pytree-able since it simply wraps a method
     # and thus is automatically present once the method's object is created
     def __init__(
         self, fn=None, *, is_discrete=NotImplemented, event_dim=NotImplemented
     ):
         super().__init__(fn)
         self._is_discrete = is_discrete
-        self._event_dim = event_dim
+        self.event_dim = event_dim
```
fehiepsi: seems like we need `self._event_dim` here

numpyro/distributions/constraints.py
```diff
@@ -275,20 +275,21 @@ def __call__(self, x: ArrayLike) -> ArrayLike:
         #     def support(self):
         #         ...
         return dependent_property(
-            x, is_discrete=self._is_discrete, event_dim=self._event_dim
+            x, is_discrete=self._is_discrete, event_dim=self.event_dim
```
fehiepsi: maybe using `self._event_dim` here

numpyro/distributions/constraints.py
```diff
@@ -343,19 +345,13 @@ def __init__(self, base_constraint, reinterpreted_batch_ndims):
                 reinterpreted_batch_ndims + base_constraint.reinterpreted_batch_ndims
             )
             base_constraint = base_constraint.base_constraint
-        self.base_constraint = base_constraint
-        self.reinterpreted_batch_ndims = reinterpreted_batch_ndims
+        self.base_constraint: Constraint = base_constraint
+        self.reinterpreted_batch_ndims: int = reinterpreted_batch_ndims
+        self.is_discrete = base_constraint.is_discrete
+        self.event_dim = base_constraint.event_dim + reinterpreted_batch_ndims
```
fehiepsi: not sure if we need to set `self._is_discrete` and `self._event_dim` here and return then in `@property` def.

numpyro/distributions/constraints.py
```diff
@@ -219,14 +218,14 @@ def is_discrete(self):
         return self._is_discrete
 
     @property
-    def event_dim(self) -> int:
+    def event_dim(self) -> int:  # type: ignore[override]
```
fehiepsi, line 217: maybe this ignore is unnecessary?

numpyro/distributions/constraints.py
```diff
@@ -799,12 +820,13 @@ def tree_flatten(self):
...
 # See https://github.com/pytorch/pytorch/issues/50616
 
-boolean: ConstraintT = _Boolean()
-circular: ConstraintT = _Circular()
-complex: ConstraintT = _Complex()
-corr_cholesky: ConstraintT = _CorrCholesky()
-corr_matrix: ConstraintT = _CorrMatrix()
-dependent: ConstraintT = _Dependent()
+
+boolean: Union[Constraint, ConstraintT] = _Boolean()
```
fehiepsi: I'm not sure why Union is required. It seems confusing to me. Actually I feel I don't understand the typing stuff: why Constraint is not a ConstraintT type, when to use `Constraint` when to use `ConstraintT`. Could you clarify?

Qazalbash:
Hi @fehiepsi, I am a little busy with my grad school applications, can I update you after Dec 15?

numpyro/distributions/constraints.py
Qazalbash: We need this ignore tag otherwise we get,

```
numpyro/distributions/constraints.py:221: error: Cannot override writeable attribute with read-only property  [override]
```

numpyro/distributions/constraints.py
Qazalbash: The problem is `TransformT` expects a `ConstraintT` type. `ConstraintT` uses `NumLike` type. Some constraints are typed with `NonScalarArray`, which is not compatible with `ConstraintT` (violating the Liskov substitution principle).

I tried to address this issue, see the last paragraph of this [comment](https://github.com/pyro-ppl/numpyro/pull/2085#issuecomment-3449998606).

Qazalbash:
@fehiepsi, thanks for looking into it. I am testing it.

Apparently, there was a doctest failing on the previous commits, but on the recent one you pushed.

juanitorduz:
yay! great teamwork!

Issue #2088, opened 2025-10-25
Beta with concentration1=1 gives nan log_prob at value=0


miguelbiron:
### Bug Description

Title.

### Steps to Reproduce

```python
import numpyro.distributions as dist

assert jnp.isnan(dist.Beta(1.0, 8.0).log_prob(0.))
```

### Expected Behavior

```python
assert jnp.isclose(dist.Beta(1.0, 8.0).log_prob(0.), 2.079)
```
The `log_prob` implementation should handle the cases with any concentration parameter equal to 1. See for example the [Julia implementation](https://github.com/JuliaStats/StatsFuns.jl/blob/814aa479e7916f35654198fdb36a387b87bac6d1/src/distrs/beta.jl#L25).

PR #2089, opened 2025-10-27
Fix Beta with concentration1=1 gives nan log_prob at value=0

juanitorduz:
Closes https://github.com/pyro-ppl/numpyro/issues/2088

### DIFF
--- a/numpyro/distributions/continuous.py
+++ b/numpyro/distributions/continuous.py
@@ -235,10 +235,11 @@ class Beta(Distribution):
             + xlogy(self.concentration0 - 1.0, 1.0 - value)
             - betaln(self.concentration1, self.concentration0)
         )
-        correction = jax.lax.stop_gradient(correct_value - safe_log_prob)
 
         # Apply correction at boundaries, return safe value elsewhere
-        return safe_log_prob + jnp.where(is_boundary, correction, 0.0)
+        return jnp.where(
+            is_boundary, jax.lax.stop_gradient(correct_value), safe_log_prob
+        )
 
     @property
     def mean(self) -> ArrayLike:


numpyro/distributions/continuous.py
```diff
@@ -216,7 +216,18 @@ def sample(
 
     @validate_sample
     def log_prob(self, value: ArrayLike) -> ArrayLike:
-        return self._dirichlet.log_prob(jnp.stack([value, 1.0 - value], -1))
+        # Handle edge cases where concentration1=1 and value=0, or concentration0=1 and value=1
+        # These cases would result in nan due to log(0) * 0 in the Dirichlet computation
+        log_prob = jnp.where(
```
fehiepsi: interesting! could you check if grads w.r.t. value, concentration1, concentration0 are not NaN?

juanitorduz: Good catch 🙈 ! The gradients are NaN with this approach. After some investigation and talking with Claude Code 😅 . We arrived to a solution via https://docs.jax.dev/en/latest/_autosummary/jax.custom_jvp.html, see ~~https://github.com/pyro-ppl/numpyro/pull/2089/commits/13fc288bda45e42840924b8026fb2b5fe5aa01fd~~ and https://github.com/pyro-ppl/numpyro/pull/2089/commits/19f8be70669b36ea0444bedd766507374c547044

fehiepsi: It is better to mask the extreme values by canonical ones before calling direchlet log_prob. Using where is better than custom jvp I think: https://github.com/tensorflow/probability/blob/main/discussion/where-nan.pdf

juanitorduz: ok! I tried to use this approach in https://github.com/pyro-ppl/numpyro/pull/2089/commits/1a9eae1113e186f147292e7060ccbdf6846da41e

numpyro/distributions/continuous.py
```diff
@@ -216,7 +216,41 @@ def sample(
...
+        safe_log_prob = (
+            xlogy(self.concentration1 - 1.0, safe_value)
+            + xlogy(self.concentration0 - 1.0, 1.0 - safe_value)
+            - betaln(self.concentration1, self.concentration0)
+        )
+
+        # Step 4: Compute correct forward-pass value at boundaries
+        # Use stop_gradient to prevent gradients from flowing through this branch
+        # xlogy(0, 0) = 0 gives the correct value when concentration=1 at boundaries
+        boundary_log_prob = jax.lax.stop_gradient(
```
fehiepsi: how about setting this to 0.0 instead of using `stop_gradient`?

numpyro/distributions/continuous.py
```diff
@@ -216,7 +216,41 @@ def sample(
...
+
+        # Step 1: Identify boundary values (0 or 1)
+        is_boundary = (value == 0.0) | (value == 1.0)
+
+        # Step 2: Inner where - mask boundary values to safe canonical value (0.5)
+        # This ensures log(0) never appears in the gradient computation path
+        safe_value = jnp.where(is_boundary, 0.5, value)
+
+        # Step 3: Compute log_prob with safe values (gradients flow through here)
+        safe_log_prob = (
```
fehiepsi: could we use self.dirichlet.log_prob here?

juanitorduz: tried it in https://github.com/pyro-ppl/numpyro/pull/2089/commits/c6f113bf02ab4082fc59a5f94fb47fe1cd2056a4

numpyro/distributions/continuous.py
juanitorduz: I could not make it work without stop_gradient :(

numpyro/distributions/continuous.py
```diff
@@ -216,7 +216,29 @@ def sample(
...
+        # Use stop_gradient so gradients come only from safe_log_prob
+        correct_value = (
+            xlogy(self.concentration1 - 1.0, value)
+            + xlogy(self.concentration0 - 1.0, 1.0 - value)
+            - betaln(self.concentration1, self.concentration0)
+        )
+        correction = jax.lax.stop_gradient(correct_value - safe_log_prob)
+
+        # Apply correction at boundaries, return safe value elsewhere
+        return safe_log_prob + jnp.where(is_boundary, correction, 0.0)
```
fehiepsi: Can we use `return jnp.where(is_boundary, jax.lax.stop_gradient(correct_value), safe_log_prob)`?

fehiepsi (approved):

numpyro/distributions/continuous.py
juanitorduz: Changed in https://github.com/pyro-ppl/numpyro/pull/2089/commits/07114da4629b4a94070ba90f1bfbf3180475705e :)

PR #2090, opened 2025-10-29
Implement IntervalCensoredDistribution

vanAmsterdam:
add support for interval censored observations as mentioned in #1909

The new `IntervalCensoredDistribution` class handles generically censored observations which can be arbitrarily mixed:

- exact observations
- left censored observations (value lower than an observed upper bound)
- right censored observations (value higher than an observed lower bound)
- interval censored observations (value is between lower and upper)
- doubly censored observations (value is below lower or above upper)

this PR implements the code, tests and basic documentation
tutorial follows in a later PR

### DIFF
--- a/numpyro/distributions/censored.py
+++ b/numpyro/distributions/censored.py
@@ -15,10 +15,7 @@ from jax.typing import ArrayLike
 from numpyro._typing import ConstraintT, DistributionT
 from numpyro.distributions import constraints
 from numpyro.distributions.distribution import Distribution
-from numpyro.distributions.util import (
-    promote_shapes,
-    validate_sample,
-)
+from numpyro.distributions.util import log1mexp, promote_shapes, validate_sample
 from numpyro.util import find_stack_level, not_jax_tracer
 
 
@@ -429,7 +426,7 @@ class IntervalCensoredDistribution(Distribution):
         lp_interval = jnp.where(m_point, self.base_dist.log_prob(x1), lp_interval)
 
         # for doubly censored data, the value is not in the interval, so computation is 1 - exp(lp_interval)
-        lp_double = jnp.log1p(-jnp.exp(lp_interval))
+        lp_double = log1mexp(lp_interval)
 
         # Select the right expression per row
         # left: log F(x2)


numpyro/distributions/censored.py
```diff
@@ -249,3 +249,183 @@ def log_survival_censored(x):
...
+    def __init__(
+        self,
+        base_dist: DistributionT,
+        left_censored: ArrayLike,
+        right_censored: ArrayLike,
+        *,
+        validate_args: Optional[bool] = None,
+    ):
+        # Optionally test that cdf actually works (in validate_args mode)
+        if validate_args:
```
juanitorduz, line 341: Suggestion: Do we want to validate interval ordering (x1 ≤ x2) under validate_args?

- When validate_args is True, raise an informative error if any row that is not purely left- or right-censored has x2 < x1. (You can exempt point intervals via jnp.isclose.)
- 
-What about handling NaN inputs explicitly during validate_args for easier debugging?

test/test_distributions.py
```diff
@@ -4055,100 +4203,188 @@ def test_censored_logprob_batch_shapes(batch_shape):
...
 
     grad_fn = jax.grad(loss_scale_right)
     grad = grad_fn(1.0)
     assert jnp.isfinite(grad)
 
+    # interval censored takes observations with (lower, upper)
+    value_interval = jnp.stack([value, value + 1], axis=-1)
 
-def test_censored_sample_validity():
+    def loss_loc_interval(loc):
```
juanitorduz, line 4321: This is never used, right? Should it be

```python
  grad_fn = jax.grad(loss_loc_right)
    grad = grad_fn(0.0)
    assert jnp.isfinite(grad)
```
?

juanitorduz (commented):
Hey @vanAmsterdam this looks great! I tested locally with some examples and it seems good. I just left some minor comments. :)

Let's see @fehiepsi 's feedback :)

test/test_distributions.py
vanAmsterdam: it should have been `grad_fn = jax.grad(loss_loc_interval)`, now fixed

numpyro/distributions/censored.py
vanAmsterdam: implemented a custom validate_sample method, let me know what you think

numpyro/distributions/censored.py
```diff
@@ -249,3 +253,246 @@ def log_survival_censored(x):
...
+        # log(F2 - F1) = logF2 + log1p(-exp(logF1 - logF2))
+        logF1 = jnp.log(F1)
+        logF2 = jnp.log(F2)
+
+        lp_interval = logF2 + jnp.log1p(-jnp.exp(jnp.clip(logF1 - logF2, max=-minval)))
+        # handle point intervals (x1 == x2) by returning log density instead of log prob
+        lp_interval = jnp.where(m_point, self.base_dist.log_prob(x1), lp_interval)
+
+        # for doubly censored data, the value is not in the interval, so computation is 1 - exp(lp_interval)
+        lp_double = jnp.log1p(-jnp.exp(lp_interval))
```
fehiepsi: you can use log1mexp, it is more stable: https://github.com/pyro-ppl/numpyro/blob/b5b68f958b90b3e5bccd1bae257f8e448f31c3bb/numpyro/distributions/util.py#L436

vanAmsterdam: thanks for suggestion, plugged this in

PR #2092, opened 2025-11-01
Use uv to speed up install dependencies

juanitorduz:
Use uv to speed up install dependencies following https://docs.astral.sh/uv/guides/integration/github/#setting-up-python

Dependency instalations are going from 2 min -> 30 sec (on average)


juanitorduz:
I need to investigate more

PR #2100, opened 2025-11-28
Add rational quadratic kernel to HSGP approximation

juanitorduz:
Add the rational quadratic kernel to HSGP approximation and tests against https://scikit-learn.org/stable/modules/generated/sklearn.gaussian_process.kernels.RationalQuadratic.html#sklearn.gaussian_process.kernels.RationalQuadratic


fehiepsi (approved):
LGTM, thanks @juanitorduz!

Issue #2017, opened 2025-04-17
Request For Inverse Wishart Distribution


richardmkit:
### Feature Summary
The Inverse Wishart Distribution should be added into the distribution class. Then it is more convenient to directly use it instead of computing the inverse of a Wishart distribution.

### Why is this needed?
The Inverse Wishart Distribution is widely used in the Normal-Inverse-Wishart prior to model the linear characteristics, due to its good conjugate property. Especially in the macroeconomic modelling, economists prefer the classic Inverse Wishart.

PR #2103, opened 2025-11-28
Add  Inverse Wishart Distribution

juanitorduz:
Closes https://github.com/pyro-ppl/numpyro/issues/2017

This one was tricky, so I got a lot of help from Opus 4.5

### DIFF
--- a/numpyro/distributions/continuous.py
+++ b/numpyro/distributions/continuous.py
@@ -3495,18 +3495,19 @@ class InverseWishartCholesky(Distribution):
         self, key: jax.dtypes.prng_key, sample_shape: tuple[int, ...] = ()
     ) -> ArrayLike:
         assert is_prng_key(key)
-        # Sample from standard WishartCholesky (Bartlett decomposition)
-        # then use the trick to get standard InverseWishartCholesky.
+        # Sample from standard InverseWishartCholesky using Bartlett decomposition
         # Ref: https://nbviewer.org/gist/fehiepsi/5ef8e09e61604f10607380467eb82006#Precision-to-scale_tril
         rng_diag, rng_offdiag = random.split(key)
         latent = jnp.zeros(sample_shape + self.batch_shape + self.event_shape)
         p = self.event_shape[-1]
-        # Standard Bartlett: nu, nu - p + 1, ..., nu - 1
+        # Inverse Wishart Bartlett: nu - p + 1, nu - p + 2, ..., nu - 1, nu
         i = jnp.arange(p)
         latent = latent.at[..., i, i].set(
             jnp.sqrt(
                 random.chisquare(
-                    rng_diag, self.concentration[..., None] - i, latent.shape[:-1]
+                    rng_diag,
+                    self.concentration[..., None] + i - p + 1,
+                    latent.shape[:-1],
                 )
             )
         )
@@ -3514,11 +3515,9 @@ class InverseWishartCholesky(Distribution):
         latent = latent.at[..., i, j].set(
             random.normal(rng_offdiag, latent.shape[:-2] + (i.size,))
         )
-        # Apply flip-transpose-inverse trick to get Cholesky of InverseWishart(I)
-        Lf_flipped = latent[..., ::-1, ::-1]
-        tril_inv = jnp.swapaxes(Lf_flipped, -1, -2)
+        # Get Cholesky of InverseWishart(I) by inverting latent
         identity = jnp.broadcast_to(jnp.eye(p), latent.shape)
-        L_inv_std = solve_triangular(tril_inv, identity, lower=True)
+        L_inv_std = solve_triangular(latent, identity, lower=True)
 
         # Transform to InverseWishart(Psi): L = scale_tril @ L_inv_std
         return jnp.matmul(self.scale_tril, L_inv_std)


numpyro/distributions/continuous.py
```diff
@@ -3265,6 +3265,301 @@ def infer_shapes(
...
+    def mode(self) -> ArrayLike:
+        p = self.scale_matrix.shape[-1]
+        return self.scale_matrix / (self.concentration[..., None, None] + p + 1)
+
+    @lazy_property
+    def variance(self) -> ArrayLike:
+        # Variance of entry (i,j) for nu > p + 3
+        # Var(X_ij) = (Psi_ij^2 + Psi_ii * Psi_jj) / ((nu - p - 1)^2 * (nu - p - 3))
+        p = self.scale_matrix.shape[-1]
+        nu = self.concentration[..., None, None]
```
fehiepsi: it's better to use `jnp.expand_dims` for `self.concentration` here and above.

numpyro/distributions/continuous.py
```diff
@@ -3265,6 +3265,301 @@ def infer_shapes(
...
+
+    **References**
+
+    [1] https://en.wikipedia.org/wiki/Inverse-Wishart_distribution
+    """
+
+    arg_constraints = {
+        "concentration": constraints.dependent(is_discrete=False, event_dim=0),
+        "scale_matrix": constraints.positive_definite,
+        "rate_matrix": constraints.positive_definite,
```
fehiepsi, line 3413: is it a common terminology? If not, it might be better to use `inverse_scale_matrix`

numpyro/distributions/continuous.py
```diff
@@ -3265,6 +3265,301 @@ def infer_shapes(
...
+            )
+        )
+        i, j = jnp.tril_indices(p, -1)
+        assert i.size == p * (p - 1) // 2
+        latent = latent.at[..., i, j].set(
+            random.normal(rng_offdiag, latent.shape[:-2] + (i.size,))
+        )
+        # L_wishart @ L_wishart^T ~ Wishart(nu, Psi^{-1})
+        scale_tril_wishart = cholesky_of_inverse(self.scale_matrix)
+        L_wishart = jnp.matmul(*jnp.broadcast_arrays(scale_tril_wishart, latent))
```
fehiepsi: can we use solve_triangular(self.scale_matrix, latent, lower=True) instead?

numpyro/distributions/continuous.py
```diff
@@ -3265,6 +3265,301 @@ def infer_shapes(
...
+        # Invert: chol((L @ L^T)^{-1}) = chol(L^{-T} @ L^{-1})
+        identity = jnp.broadcast_to(jnp.eye(p), L_wishart.shape)
+        L_inv = solve_triangular(jnp.swapaxes(L_wishart, -1, -2), identity, lower=False)
+        X_invwishart = jnp.matmul(L_inv, jnp.swapaxes(L_inv, -1, -2))
+        return jnp.linalg.cholesky(X_invwishart)
```
fehiepsi: it seems to me we can return `L_inv` directly?

numpyro/distributions/continuous.py
```diff
@@ -3265,6 +3265,301 @@ def infer_shapes(
...
+        assert i.size == p * (p - 1) // 2
+        latent = latent.at[..., i, j].set(
+            random.normal(rng_offdiag, latent.shape[:-2] + (i.size,))
+        )
+        # L_wishart @ L_wishart^T ~ Wishart(nu, Psi^{-1})
+        scale_tril_wishart = cholesky_of_inverse(self.scale_matrix)
+        L_wishart = jnp.matmul(*jnp.broadcast_arrays(scale_tril_wishart, latent))
+        # Invert: chol((L @ L^T)^{-1}) = chol(L^{-T} @ L^{-1})
+        identity = jnp.broadcast_to(jnp.eye(p), L_wishart.shape)
+        L_inv = solve_triangular(jnp.swapaxes(L_wishart, -1, -2), identity, lower=False)
```
fehiepsi: is L_wishart a lower triangular matrix? maybe worth commenting that latent is a lower triangular matrix?

numpyro/distributions/continuous.py
```diff
@@ -3265,6 +3265,301 @@ def infer_shapes(
...
+        x = solve_triangular(*jnp.broadcast_arrays(value, self.scale_tril), lower=True)
+        trace = jnp.square(x).sum(axis=(-1, -2))
+
+        p = value.shape[-1]
+        return (
+            self.concentration * tri_logabsdet(self.scale_tril)  # (nu/2) log|Psi|
+            + p * (1 - self.concentration / 2) * jnp.log(2)  # normalization
+            - multigammaln(self.concentration / 2, p)
+            - (self.concentration + p + 1)
+            * tri_logabsdet(value)  # -((nu+p+1)/2) log|X|
```
fehiepsi: can we merge this with `jnp.arange(p, 0, -1) * jnp.log(jnp.diagonal(value, axis1=-2, axis2=-1))` below?

numpyro/distributions/continuous.py
juanitorduz: changed in  in https://github.com/pyro-ppl/numpyro/pull/2103/commits/05127de5c84a959920571b1f3e2cc213a9751ecc

numpyro/distributions/continuous.py
juanitorduz: yes, good point! changed in https://github.com/pyro-ppl/numpyro/pull/2103/commits/05127de5c84a959920571b1f3e2cc213a9751ecc

numpyro/distributions/continuous.py
juanitorduz: tried simplifying in https://github.com/pyro-ppl/numpyro/pull/2103/commits/05127de5c84a959920571b1f3e2cc213a9751ecc

numpyro/distributions/continuous.py
juanitorduz: added comment in https://github.com/pyro-ppl/numpyro/pull/2103/commits/2b58069170aae9e59d7065ea805aff5778901bb6

numpyro/distributions/continuous.py
juanitorduz: This comes from the class `class Wishart(TransformedDistribution)`

https://github.com/pyro-ppl/numpyro/blob/b0c1dab84269a02b50a43e603fb97887321896bf/numpyro/distributions/continuous.py#L3015

I added similar docstrings in https://github.com/pyro-ppl/numpyro/pull/2103/commits/972f451a5383f5ccd108880883981527adb3d80a

numpyro/distributions/continuous.py
juanitorduz: I tried it out, and I could not make it work with `solve_triangular(self.scale_tril, latent, lower=True)` (or with `.mT`)  as

- **Current approach:** `chol(Psi^{-1}) @ latent` produces a **lower triangular** `L_wishart`, which is required for the subsequent `solve_triangular` inversion step.

- **The suggested approach:** `L^{-T} @ latent` (upper × lower) produces a *general (non-triangular) matrix* (as far as I can see), breaking the `solve_triangular` call that follows.

The triangular structure must be preserved throughout the sampling algorithm, so `cholesky_of_inverse(self.scale_matrix)` is necessary.

(also verified both by hand and with Opus 🙈 )

tillahoffmann:
It could be interesting to compare with sampling from `WishartCholesky` and applying the lower-triangular inverse to get the `InverseWishartCholesky`. It would be a little bit more computation to evaluate the inverse, but maybe it's not too much given the lower-triangular structure?

juanitorduz:
> It could be interesting to compare with sampling from `WishartCholesky` and applying the lower-triangular inverse to get the `InverseWishartCholesky`. It would be a little bit more computation to evaluate the inverse, but maybe it's not too much given the lower-triangular structure?

Something like https://github.com/pyro-ppl/numpyro/pull/2103/commits/6795d4bac4d716d03e40ee2da6c6b3f76d717fad and https://github.com/pyro-ppl/numpyro/pull/2103/commits/4dcd0ed848101e9cb4eea8532ce24d28e384716b ?

numpyro/distributions/continuous.py
fehiepsi: Thanks, it looks good then. Could you help fixing the typo: anaologous

numpyro/distributions/continuous.py
fehiepsi: You are right, thanks for clarifying!

numpyro/distributions/continuous.py
```diff
@@ -3265,6 +3265,291 @@ def infer_shapes(
...
+        # If X ~ Wishart(nu, Psi^{-1}), then X^{-1} ~ InverseWishart(nu, Psi)
+        # Sample from WishartCholesky with rate_matrix=Psi to get Wishart(nu, Psi^{-1})
+        wishart_cholesky = WishartCholesky(
+            self.concentration, rate_matrix=self.scale_matrix
+        )
+        L_wishart = wishart_cholesky.sample(key, sample_shape)
+        # Invert: L_wishart^{-1} is lower triangular (efficient via solve_triangular)
+        p = self.event_shape[-1]
+        identity = jnp.broadcast_to(jnp.eye(p), L_wishart.shape)
+        L_wishart_inv = solve_triangular(L_wishart, identity, lower=True)
```
fehiepsi: maybe return `L_wishart_inv`?

juanitorduz: mmm, if we do this, the `test/test_distributions.py::test_inverse_wishart_variance` fails like

```python
>       assert_allclose(analytical_var, empirical_var, rtol=rtol)
E       AssertionError: 
E       Not equal to tolerance rtol=0.1, atol=0
E       
E       Mismatched elements: 4 / 4 (100%)
E       Max absolute difference among violations: 0.00018424
E       Max relative difference among violations: 0.62670946
E        ACTUAL: array([[0.000439, 0.000123],
E              [0.000123, 0.00011 ]], dtype=float32)
E        DESIRED: array([[0.0003  , 0.000177],
E              [0.000177, 0.000294]], dtype=float32)
```

(this is for for `30.0` param, but is similar for others)

fehiepsi: I see. Looking like we are doing cholesky of inverse two times: one to get `L_wishart` and the other one is `cholesky_of_inverse(L_wishart @ L_wishart.T)`. Some duplications are happening here.

I think a simpler way is to generate "standard" InverseWishartCholesky samples, then scale it by `self.scale_tril`. To generate standard samples, I think we can use the Bartlett decomposition (maybe factor out the implementation in WishartCholesky implementation) to get samples `latent` from standard WishartCholesky. Then the standard inverse version would be: latent_inv.T@latent_inv where latent_inv can be obtained by tril_solve. This version only involves a single tril_solve operator.

fehiepsi: Actually, we need cholesky_of_inverse(latent @ latent.T). Not sure if we can sample from standard invert wishart cholesky directly

fehiepsi: I just revisit https://nbviewer.org/gist/fehiepsi/5ef8e09e61604f10607380467eb82006#Precision-to-scale_tril I think we can apply flip(transpose(.)) operator for Bartlett (i.e. reverse the order of entries in the diagonal part) and use tril_solve to get samples from standard invert wishart cholesky.

juanitorduz: Something like https://github.com/pyro-ppl/numpyro/pull/2103/commits/182141053cdba62c7c339305303d97594a569fcf ? 🙈 

I am not familiar with these tricks so feel free to push to the branch if I am going in circles :)

numpyro/distributions/continuous.py
```diff
@@ -3495,19 +3495,33 @@ def sample(
...
-        L_wishart_inv = solve_triangular(L_wishart, identity, lower=True)
-        # (L @ L^T)^{-1} = L^{-T} @ L^{-1}, take Cholesky for lower triangular result
-        X_inv = jnp.matmul(jnp.swapaxes(L_wishart_inv, -1, -2), L_wishart_inv)
-        return jnp.linalg.cholesky(X_inv)
+        # Standard Bartlett: nu, nu - p + 1, ..., nu - 1
+        i = jnp.arange(p)
+        latent = latent.at[..., i, i].set(
+            jnp.sqrt(
+                random.chisquare(
+                    rng_diag, self.concentration[..., None] - i, latent.shape[:-1]
```
fehiepsi: instead of applying flip-transpose operator to get `tril_inv`, you can use `self.concentration[..., None] + i - p + 1`, which is corresponding to Bartlett: nu - pi + 1, ..., nu - 1, nu

juanitorduz: Something like https://github.com/pyro-ppl/numpyro/pull/2103/commits/6f5bcc13ff96a2f52a3e29c654c1a3d769beda98 ?

PR #2104, opened 2025-12-04
Add initial value to Gaussian state space distribution (fixes #2098).

### LINKED ISSUE #2098: Add support for passing initial value in `GaussianStateSpace`
### Feature Summary
Add the possibility of passing `initial_value` to [`GaussianStateSpace`](https://github.com/pyro-ppl/numpyro/blob/47335b83dd02726fb142e61d79193eb761009ba7/numpyro/distributions/continuous.py#L622).

### Why is this needed?
When estimating the parameters of an Ornstein-Uhlenbeck process, we typically observe only the latest N values of a process that started much earlier than our first observation. For an scalar OU process with mean reversion $k$ and volatility $\sigma$, the variance of the increments is $\frac{\sigma^2}{2} \frac{1 - e^{- k \Delta t}}{k}$ while the distribution of the first observation has variance $\frac{\sigma^2}{2k}$.

Related: #2037

tillahoffmann:
This PR adds an `initial_value` argument to the `GaussianStateSpace` distribution as suggested in #2098.

As part of this change, I added an `optional` constraint. I'm a bit torn on whether that's the right choice, and we could instead promote 0 to the right shape. However, that would make evaluating the `mean` of the distribution relatively inefficient when there is no initial value: We'd still `scan` over the sequence even though we should really just return zeros (although maybe `jax.jit` amortizes that?). Open to suggestions.

### DIFF
--- a/numpyro/distributions/continuous.py
+++ b/numpyro/distributions/continuous.py
@@ -709,10 +709,11 @@ class GaussianStateSpace(TransformedDistribution):
 
     .. math::
         \mathbf{z}_{t} &= \mathbf{A} \mathbf{z}_{t - 1} + \boldsymbol{\epsilon}_t\\
-        &=\sum_{k=1} \mathbf{A}^{t-k} \boldsymbol{\epsilon}_t,
+        &= \mathbf{A}^t \mathbf{z}_0 + \sum_{k=1}^{t} \mathbf{A}^{t-k} \boldsymbol{\epsilon}_k,
 
     where :math:`\mathbf{z}_t` is the state vector at step :math:`t`, :math:`\mathbf{A}`
-    is the transition matrix, and :math:`\boldsymbol\epsilon` is the innovation noise.
+    is the transition matrix, :math:`\mathbf{z}_0` is the initial value, and
+    :math:`\boldsymbol\epsilon` is the innovation noise.
 
 
     :param num_steps: Number of steps.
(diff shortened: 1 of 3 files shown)


numpyro/distributions/continuous.py
```diff
@@ -705,23 +711,45 @@ def __init__(
...
+            transition_matrix, initial_value=initial_value
+        )
         super().__init__(base_distribution, transform, validate_args=validate_args)
 
     @property
     def mean(self) -> ArrayLike:
-        # The mean of the base distribution is zero and it has the right shape.
-        return self.base_dist.mean
+        # If there's no initial value, the mean is zero (base distribution mean).
+        if self.initial_value is None:
```
fehiepsi: how about checking `self._initial_value is None` and setting the property `self.initial_value` to zero if `self._initial_value is None`

numpyro/distributions/continuous.py
```diff
@@ -662,13 +663,16 @@ class GaussianStateSpace(TransformedDistribution):
...
+    :param initial_value: Initial state vector :math:`\mathbf{z}_0`. If ``None``,
+        defaults to zero.
     """
 
     arg_constraints = {
         "covariance_matrix": constraints.positive_definite,
         "precision_matrix": constraints.positive_definite,
         "scale_tril": constraints.lower_cholesky,
         "transition_matrix": constraints.real_matrix,
+        "initial_value": constraints.optional(constraints.real_vector),
```
fehiepsi: I think `optional` is unnecessary: just use constraints.real_vector here and make `initial_value` a property like in the above comment.

PR #2105, opened 2025-12-12
Bugfix/mutables leaves

esennesh:
When employing NNX or other neural-network frameworks, we can sometimes get empty PyTrees of supposed mutable-state passed around.  Since these contain no tensors at all, they should be detected as empty and, in my opinion at least, it's helpful to go down code paths with these that might sometimes require not having mutables.


fehiepsi (approved):

fehiepsi:
Thanks @esennesh!

PR #2106, opened 2025-12-12
distributions.mixtures._MixtureBase: mixtures lack reparameterized sa…

esennesh:
…mpling

Testing Done: create a mixture distribution, check its has_rsample


fehiepsi:
I think we don't have rsample for those distributions. The default check in Distribution does not apply because we don't have reparam parameters defined here.

fehiepsi (approved):

PR #2111, opened 2025-12-20
feat(gh-2110): setup for `astral-sh/ty`

### LINKED ISSUE #2110: Adopt `astral-sh/ty` for Faster Static Type Checking
MyPy is a widely used and reliable tool for static type checking, but it can be prohibitively slow for large codebases. While working on type hinting within NumPyro (#2032, #2050, #2066, #2085), I have to wait for several minutes for MyPy to complete. The project [`astral-sh/ty`](https://github.com/astral-sh/ty) offers a significantly faster, largely in-place alternative to MyPy, and adopting it could meaningfully reduce development and CI turnaround times.

Qazalbash:
This PR introduces faster type checking by `astral-sh/ty` as an alternative to mypy and fixes #2110.

I have not removed mypy entirely; it is still the primary type checker. In the future, we will transfer small type-checking workloads to ty and move the dependence from mypy. For the very same reason, I have provided the `--exit-zero` flag to force exit with 0 exit code so that it does not interfere with our CI.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -130,5 +130,6 @@ module = [
     "numpyro.primitives.*",
     "numpyro.patch.*",
     "numpyro.util.*",
+    "numpyro.distributions.transforms",
 ]
 ignore_errors = false


pyproject.toml
```diff
@@ -130,6 +130,5 @@ module = [
     "numpyro.primitives.*",
     "numpyro.patch.*",
     "numpyro.util.*",
-    "numpyro.distributions.transforms",
```
juanitorduz, line 133: ehy did you remove this?

Qazalbash: I mistakenly pushed it; let me revert it.

PR #2115, opened 2025-12-25
Disable masking when running Predictive with infer_discrete=True

fehiepsi:
Fixes the issue in https://forum.pyro.ai/t/recover-discrete-latent-states-after-enumerate-scan/9018/8

We need log probabilities to infer the probabilities at discrete sites, so we can draw samples from.


juanitorduz (approved):
💪

PR #2120, opened 2026-01-19
Bump minimum python version to 3.11 and minimum jax version to 0.7

fehiepsi:


Qazalbash:
Thank god CI is failing 😅 I thought I must have messed something up in #2085.

juanitorduz:
I think this is failling because of the same reason as https://github.com/pyro-ppl/numpyro/pull/2099 :)

juanitorduz (approved):

PR #2121, opened 2026-01-20
ci: prek for pre-commit hooks check

Qazalbash:
This PR adds pre-commit checks to the CI pipeline. Instead of the standard `pre-commit` framework, it uses [prek](https://github.com/j178/prek), a faster Rust-based alternative.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,20 +15,9 @@ jobs:
   prek:
 
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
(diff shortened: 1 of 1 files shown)


juanitorduz (approved):
I like this because we could add more integration with ruff and ty if necessary. Fut as its a ci change I will let @fehiepsi decide :)

juanitorduz:
Ok, then let's merge and experiment, as I see we could simplify our lint process a bit :) 

I also see it is being used in other big projects https://github.com/j178/prek?tab=readme-ov-file#who-is-using-prek

PR #2122, opened 2026-01-20
Add Types to `numpyro.distributions.distribution`

juanitorduz:
Related to https://github.com/pyro-ppl/numpyro/issues/2036

### DIFF
--- a/numpyro/distributions/distribution.py
+++ b/numpyro/distributions/distribution.py
@@ -685,7 +685,7 @@ class ExpandedDistribution(Distribution):
         )
         self._expanded_sizes = expanded_sizes
         self._interstitial_sizes = interstitial_sizes
-        self.has_enumerate_support = actual_base_dist.has_enumerate_support
+        self.has_enumerate_support = bool(actual_base_dist.has_enumerate_support)
         super().__init__(new_shape, actual_base_dist.event_shape)
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


numpyro/distributions/continuous.py
```diff
@@ -55,11 +55,11 @@
 from jax.scipy.stats import norm as jax_norm
 from jax.typing import ArrayLike
 
+from numpyro._typing import DistributionT
```
fehiepsi: can we just depend on Distribution, instead of DistributionT? We have some success in constraints.py and transforms.py, which do not use ConstraintT and TransformT. We want to reduce the scope of using DistributionT through the library. Ideally, it's great to not use those protocols at all to avoid confusion.

juanitorduz: removed them in https://github.com/pyro-ppl/numpyro/pull/2122/changes/3303b4a77296e39aeac4f69485aa34ebc3d18297 :)

fehiepsi: Maybe grep DistributionT and remove at other scripts too? :)

juanitorduz: Sure! Started on https://github.com/pyro-ppl/numpyro/pull/2122/changes/c27121b17eddc7fe10796e3ff937bab011d40b1f and fixing edge cases in the following commits :)

fehiepsi: Thank you!

PR #2124, opened 2026-01-20
feat: Add Microcanonical Langevin Monte Carlo (MCLMC) kernel

juanitorduz:
Trying to support  https://github.com/pyro-ppl/numpyro/pull/2039


fehiepsi (commented):
I worry that we might introduce technical debt by depending on other libraries. Could we turn this into an example/tutorial instead? It's not clear to me the benefit of using numpyro here.

fehiepsi:
Maybe adding a section for https://num.pyro.ai/en/stable/tutorials/other_samplers.html would be better?

edit: https://github.com/pyro-ppl/numpyro/issues/2035 has good discussion about the above points

juanitorduz:
@fehiepsi @martinjankowiak @reubenharry 

Suggestion: Keep this code but put it into `contrib.infer` and make `blackjax` as an optional dependency as in https://github.com/pyro-ppl/numpyro/blob/master/numpyro/contrib/nested_sampling.py :

https://github.com/pyro-ppl/numpyro/blob/d5598e73b549748e35a04b30d20e82f61dc46a36/numpyro/contrib/nested_sampling.py#L10-L29

martinjankowiak:
it seems to me like the best option is to pull out the implementation from blackjax. this is something claude could presumably do in < 10 minutes. the only downside is that you don't automatically inherit any improvements made to the blackjax implementation, but that's not a big deal and can always be addressed with future PRs if there are significant improvements in the blackjax implementation.

numpyro/infer/mclmc.py
```diff
@@ -0,0 +1,859 @@
+# Copyright Contributors to the Pyro project.
```
martinjankowiak, line 1: presumably many of the private methods in this file have corresponding tests in blackjax that you can pull into this PR?

juanitorduz: Added in https://github.com/pyro-ppl/numpyro/pull/2124/changes/e2d754c1e71c8c92833539218105d13e6d576ddb

juanitorduz:
> @reubenharry I have tried to remove the blackjax dependency with Claude's help! I ensured the results match the code from #2039 but it would be great if you could give a proper review
> 
> @fehiepsi @martinjankowiak Open for feedback :)

hey folks! any thoughts on this one? This raw implementation is self-contained and about 859 lines of code (I am fine with it)

I wanna also want to bring up the fact that blackjax is quite active now https://github.com/blackjax-devs/blackjax/releases, so my gut tells me it's less risky to have it as an optional dependency (much safer than jaxns).

reubenharry:
Sorry for the slow reply. I had one minor comment (see above), but since this follows the BlackJax quite closely, the rest looks fine. I think either the self contained version, or using Blackjax directly would be fine (but if the latter, I'd suggest that NumPyro makes a larger scale decision to be open to using Blackjax across the board).

fehiepsi:
It's great to have the flexibility to use any package while building a propbabilistic program. But it is unclear to me why numpyro needs to depend on blackjax though. If the end goal is to have a familiar api for new algorithms, we can build a package like blackro or something. @juanitorduz could you outline which parts are not compatible? We can add some new utilities for inference if they provide better flexibility.

juanitorduz:
> It's great to have the flexibility to use any package while building a probabilistic program. But it is unclear to me why numpyro needs to depend on blackjax though. If the end goal is to have a familiar api for new algorithms, we can build a package like blackro or something. @juanitorduz could you outline which parts are not compatible? We can add some new utilities for inference if they provide better flexibility.

Hi @fehiepsi ! Thanks for the feedback! There is no incompatibility problem, I think the question is: "where do we put the code responsible for the blackjax <--> numpyro connection. For advanced users, they can just get the logp and go ahead, but I think we are thinking about a friendlier API like NUTS.

 > But it is unclear to me why numpyro needs to depend on blackjax, though. If the end goal is to have a familiar api for new algorithms, we can build a package like blackro or something.
 
 I think this is a great point, and I am thinking a minimal package like`blackrho`, could bridge the gap and keep the ecosystem modular. I will be happy to draft a system design for this potential package :)

fehiepsi:
We can also redesign new utilities like `initialize_model` etc. to make them more flexible, to minimize blackjax <-> numpyro connection. It seems that you think NUTS api is more friendly than blackjax api? If so I guess much of the code is for blackjax, not numpyro? Sorry I'm not familiar with blackjax.

juanitorduz:
> We can also redesign new utilities like `initialize_model` etc. to make them more flexible, to minimize blackjax <-> numpyro connection. It seems that you think NUTS api is more friendly than blackjax api? If so I guess much of the code is for blackjax, not numpyro? Sorry I'm not familiar with blackjax.

I also have a bias as I use NumPyro much more 🙈 . Actually this is a good idea! Before jumping into a package design, it could be that better utilities to extract the logp and transforms can be the missing key to the puzzle. Let me think about it 💡

reubenharry:
Blackjax is really intended as a developer library, supplying modular parts for inference algorithms, but less so wrappers that are user-facing. In this sense, I would say it is designed as the sort of thing you might want to use as a backend in a PPL like Numpyro.

But in either case, for me the main interest is in having something directly usable in NumPyro - I think this will make the algorithm much more "discoverable".

reubenharry:
@juanitorduz I'm curious if the Numpyro team has any thoughts on next steps. I'd love to see this PR merged, as I think the algorithm works well (i.e. quite a bit faster than NUTS) so it would be nice to see it get a wider audience. 

@fehiepsi Let me know what I can help with also - other than the choice of integrator, the code looks good to me (and that's also fine to keep as leapfrog for now too)

juanitorduz:
> @juanitorduz I'm curious if the Numpyro team has any thoughts on next steps. I'd love to see this PR merged, as I think the algorithm works well (i.e. quite a bit faster than NUTS) so it would be nice to see it get a wider audience. 
> 
> 
> 
> @fehiepsi Let me know what I can help with also - other than the choice of integrator, the code looks good to me (and that's also fine to keep as leapfrog for now too)

Hi @reubenharry ! Yes, it in my to-do list to work on a design for better utility to integrate the model init with blackjack. I'll be able to get something this weekend.

juanitorduz:
@reubenharry @fehiepsi I created a PR to improve the libraries' integration and tested it using MCLMC. Open for feedback :) https://github.com/pyro-ppl/numpyro/pull/2203

reubenharry:
> @reubenharry @fehiepsi I created a PR to improve the libraries' integration and tested it using MCLMC. Open for feedback :) #2203

So with this merged, is the remaining step to add BlackjaxMCLMCKernel to NumPyro?

fehiepsi:
It is already available in the tutorial. I guess we can close this one.

reubenharry:
> It is already available in the tutorial. I guess we can close this one.

Do you mean in https://github.com/pyro-ppl/numpyro/blob/master/notebooks/source/other_samplers.ipynb? If so, I feel like this is a bit hard for users to discover. 

I'm really happy MCLMC is now in numpyro, but the goal I have in mind is to make MCLMC an easily available option for numpyro users when performing inference (I think this is well-motivated, since it appears to be quite a bit faster than NUTS across the board).

fehiepsi:
mclmc is developed in blackjax so it's easier to support there. If you have bandwidth, please create a package introducing the wrapper as in the tutorial.

reubenharry:
Since users can already extract a log density from numpyro and use it with any blackjax sampler, I'm then not sure what the direct numpyro integration adds. From my perspective, the advantage would have been discoverability, but I don't see that an additional package would help there - I'll consider it if I have the bandwidth though, and thanks for helping with all this!

PR #2128, opened 2026-01-24
Bump numpyro to 0.20.0

fehiepsi:
Needs to introduce eq api to avoid tracer issues with __eq__ api. Python needs __eq__ to have bool type but if we compare on jax arrays, the output will be a jax scalar boolean.

### DIFF
--- a/numpyro/distributions/util.py
+++ b/numpyro/distributions/util.py
@@ -20,7 +20,7 @@ from jax.typing import ArrayLike
 from numpyro.util import not_jax_tracer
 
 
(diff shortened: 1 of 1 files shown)


numpyro/distributions/util.py
```diff
@@ -19,6 +19,26 @@
 
 from numpyro.util import not_jax_tracer
 
+
+def array_equiv(a, b, static=False):
```
juanitorduz: Maybe add type hints?

juanitorduz (approved):
Thanks for fixing this bug! I was not able to get what was wrong!

Exited about the release 🥳

PR #2135, opened 2026-02-10
Fix EnsembleSampler second split to use complementary inactive chains.

jackkohm:
The second sub-iteration in EnsembleSampler.sample incorrectly updated the second half of chains against itself instead of the first half. Add a deterministic regression test that fails under the previous split logic and verifies complementary-half updates.


amifalk:
Nice catch! My tiny nit is that it was a bit hard to initially parse that the first two chains have values that aren't used in the test case. Initializing z to `[0.0], [0.0], [10.0], [11.0]]` or `[nan], [nan], [10.0], [11.0]]`might help signpost the logic of the test, i.e that you start by copying [10, 11] + 1 into the first two chains, and then copy [11, 12] + 1 into the last two chains.

fehiepsi:
Close in favor of #2189

PR #2169, opened 2026-04-02
Effect handlers Tutorial

juanitorduz:
Here is a relatively extensive tutorial on effect handlers. This was hard for me to wrap my head around at the beginning, and I wished something like this existed :) 

Open for feedback :)


damonbayer:
Really nice tutorial, @juanitorduz! I was wondering if you could explain why you used

```python
    email = numpyro.sample(
        "email", dist.Bernoulli(email_prob).expand([n_obs]).to_event(1), obs=email
    )
```

instead of a plate? Or instead of expanding without using `to_event`?. Is it related to the `do` handler?

juanitorduz:
> Really nice tutorial, @juanitorduz! I was wondering if you could explain why you used
> 
> ```python
>     email = numpyro.sample(
>         "email", dist.Bernoulli(email_prob).expand([n_obs]).to_event(1), obs=email
>     )
> ```
> 
> instead of a plate? Or instead of expanding without using `to_event`?. Is it related to the `do` handler?

hey @damonbayer ! Thanks! There is no particular reason, I could use a plate handler :)

fehiepsi:
Beautiful tutorial!!

PR #2171, opened 2026-04-04
fix repeated recompilation in no-progress-bar fori_collect path

simeoncampos:
`fori_collect(..., progbar=False)` recreates `loop_fn` as a fresh closure on each call. JAX JIT caching depends on function identity, so recreating the closure defeats cache reuse and triggers full XLA recompilation on every `MCMC.run()`.

### Root cause

`loop_fn` was introduced in #1802 to wrap `fori_loop` for `donate_argnums` via `maybe_jit`. Unlike `_body_fn` which was already cached via `@cached_by`, the new wrapper was not. Each call produces a new function object, cache miss, recompile.

### Fix

Cache `loop_fn` via `@cached_by(fori_collect, body_fun, transform, upper, start_idx, thinning)` and move `init_val` from the closure into an explicit argument. Matches the existing `_body_fn` caching pattern at `util.py:387`.

Speedups on repeated `MCMC.run()` with `progress_bar=False` (CPU, median of runs 2-5):

- normal_normal (2p): 0.44s to 0.08s
- eight_schools (10p): 0.56s to 0.12s
- logistic_regression (11p): 0.54s to 0.08s
- stochastic_volatility (~1000p): 1.42s to 0.81s

On GPU (2x RTX A5000): 1.2-2.0x.

### Tests

- `test_fori_collect_no_recompilation`: regression test for repeated `fori_collect(..., progbar=False)` calls with correct non-stale results and a populated cached wrapper path
- `test_fori_collect_repeated_mcmc_no_recompilation`: end-to-end regression coverage through repeated `MCMC.run()`

Done with the help of Claude Code.

### DIFF
--- a/numpyro/util.py
+++ b/numpyro/util.py
@@ -309,6 +309,15 @@ def progress_bar_factory(
     return progress_bar_fori_loop
 
 
+def _fori_collect_loop(_body_fn, upper, init_val, collection, start_idx, thinning):
+    return fori_loop(
+        0,
+        upper,
+        lambda i, vals: _body_fn(i, *vals),
+        (init_val, collection, start_idx, thinning),
+    )
+
+
 def fori_collect(
     lower: int,
     upper: int,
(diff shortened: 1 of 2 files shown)


Qazalbash:
@simeoncampos CI is failing on lint and formatting. Please run `prek run -a` or `pre-commit run -a` and push the new changes.

simeoncampos:
done, thanks for flagging.

juanitorduz (approved):
thanks @simeoncampos 1

numpyro/util.py
```diff
@@ -400,16 +400,20 @@ def map_fn(x):
     collection = jax.tree.map(map_fn, init_val_transformed)
 
     if not progbar:
-
-        def loop_fn(collection):
+        # Cache loop_fn so jit() reuses the compiled trace across calls.
+        # Without this, loop_fn is a fresh closure each call and jit recompiles.
+        @cached_by(fori_collect, body_fun, transform, upper, start_idx, thinning)
```
fehiepsi: I feel cache fori_collect is unnecessary. Since _body_fn is unique, maybe we can create a global partial loop fn instead?

simeoncampos: Updated!

fehiepsi (approved):
Thanks @simeoncampos!

PR #2172, opened 2026-04-15
mcmc: Fix memory leak when chain_method="parallel"

solutionseekeras:
Since jax >= 0.8.0 functions are cached based on identity. Creating a new partial() each run leads to new identity -> unable to cache. Fix this by reusing the same partial object across runs, only updating its keywords instead of constructing a new partial().

Reproducing code:
```python
"""
Minimal reproducer for numpyro MCMC memory leak with parallel chains.

numpyro's MCMC.run() creates a new functools.partial each call and wraps it
in jax.pmap. Since jax >=0.8.0, pmap is implemented as jit(shard_map), and
jit caches by function identity. A new partial = new identity = new trace +
compilation that is never freed.
"""

from __future__ import annotations

import gc

import numpyro
import psutil
from jax import random


def get_rss_mib() -> float:
    return psutil.Process().memory_info().rss / (1024**2)


def model() -> None:
    numpyro.sample("obs", dist.Normal(x, 0.1), obs=1.0)


def main() -> None:
    numpyro.set_host_device_count(4)
    rng_key = random.key(0)

    mcmc = MCMC(NUTS(model), num_warmup=10, num_samples=10, num_chains=4, progress_bar=False, chain_method="parallel")

    rss_start = get_rss_mib()
    for _ in range(50):
        mcmc.run(rng_key)
        gc.collect()
    rss_end = get_rss_mib()

    print(f"numpyro=={numpyro.__version__}, chains=4, parallel")
    print(f"RSS: {rss_start:.0f} -> {rss_end:.0f} MiB (+{rss_end - rss_start:.0f} MiB over 50 calls)")


if __name__ == "__main__":
    main()
```

NB! I did not create an issue for tracking this. Let me know if you require that.

### DIFF
--- a/numpyro/infer/mcmc.py
+++ b/numpyro/infer/mcmc.py
@@ -389,6 +389,7 @@ class MCMC(object):
         # HMCState returned by hmc.init_kernel
         self._init_state_cache = {}
         self._cache = {}
+        self._partial_map_fn: partial | None = None
         self._collection_params = {}
         self._set_collection_params()
 
@@ -699,13 +700,29 @@ class MCMC(object):
         collect_fields = tuple(collect_fields.keys())
         remove_sites = tuple(remove_sites.keys())
 
-        partial_map_fn = partial(
-            self._single_chain_mcmc,
-            args=args,
-            kwargs=kwargs,
-            collect_fields=collect_fields,
-            remove_sites=remove_sites,
-        )
+        # Reuse the same partial object across run() calls so that
+        # jax.pmap sees a stable function identity and reuses the cached
+        # compilation.
+        # Since JAX 0.8.0 (aa2f995b1db548df5f1da92b43514f5314e75e4e),
+        # pmap is implemented via jit(shard_map) and jit caches by function
+        # identity: a new partial each call triggers a fresh trace + XLA compilation
+        # whose artifacts are never freed, causing unbounded memory
+        # growth.
+        if self._partial_map_fn is None:
+            self._partial_map_fn = partial(
+                self._single_chain_mcmc,
+                args=args,
+                kwargs=kwargs,
+                collect_fields=collect_fields,
+                remove_sites=remove_sites,
+            )
+        else:
+            self._partial_map_fn.keywords["args"] = args
+            self._partial_map_fn.keywords["kwargs"] = kwargs
+            self._partial_map_fn.keywords["collect_fields"] = collect_fields
+            self._partial_map_fn.keywords["remove_sites"] = remove_sites
+
+        partial_map_fn = self._partial_map_fn
         map_args = (rng_key, init_state, init_params)
         if self.num_chains == 1:
             states_flat, last_state = partial_map_fn(map_args)
--- a/test/infer/test_mcmc.py
+++ b/test/infer/test_mcmc.py
@@ -863,6 +863,36 @@ def test_mcmc_inside_jit_no_tracer_leak():
             )
 
 
+def test_reuse_mcmc_run_stable_partial_identity():
+    """Regression test: repeated run() calls must reuse the same partial object.
+
+    When pmap is implemented via jit(shard_map) (JAX >= 0.8.0), jit caches
+    by function identity. Creating a new functools.partial each run() call
+    causes a fresh trace + XLA compilation whose artifacts are never freed,
+    leading to unbounded memory growth in long-running services.
+    """
+
+    def model():
+        numpyro.sample("x", dist.Normal(0, 1))
+
+    mcmc = MCMC(
+        NUTS(model),
+        num_warmup=5,
+        num_samples=5,
+        num_chains=1,
+        progress_bar=False,
+    )
+    mcmc.run(random.key(0))
+    first_partial = mcmc._partial_map_fn
+    assert first_partial is not None
+
+    mcmc.run(random.key(1))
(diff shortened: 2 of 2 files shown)


Qazalbash:
@solutionseekeras, thanks for the PR. There is no need to create an issue.

cc: @fehiepsi, @juanitorduz

solutionseekeras:
Updated reproducer, demonstrating the patch (by monkeypatching):
```python
"""
Minimal reproducer for numpyro MCMC memory leak.
"""

from __future__ import annotations

import gc
from functools import partial

import jax
import numpyro
import numpyro.infer.mcmc as mcmc_module
import psutil
from jax import random


def get_rss_mib() -> float:
    return psutil.Process().memory_info().rss / (1024**2)


def model() -> None:
    numpyro.sample("obs", dist.Normal(x, 0.1), obs=1.0)


_original_partial = partial
_cached_partials: dict[int, partial] = {}


def _stable_partial(func, *args, **kwargs):
    """Return a cached partial when wrapping _single_chain_mcmc, otherwise fall back to normal partial."""
    if getattr(func, "__name__", None) == "_single_chain_mcmc":
        # Key on the underlying function + bound self to handle multiple MCMC instances
        key = id(getattr(func, "__self__", func))
        if key not in _cached_partials:
            _cached_partials[key] = _original_partial(func, *args, **kwargs)
        else:
            _cached_partials[key].keywords.update(kwargs)
        return _cached_partials[key]
    return _original_partial(func, *args, **kwargs)


def run_test(label: str, chain_method: str, patched: bool) -> None:
    if patched:
        mcmc_module.partial = _stable_partial
    else:
        mcmc_module.partial = _original_partial

    mcmc = MCMC(NUTS(model), num_warmup=10, num_samples=10, num_chains=4, progress_bar=False, chain_method=chain_method)

    rng_key = random.key(0)
    rss_start = get_rss_mib()
    for _ in range(50):
        mcmc.run(rng_key)
        gc.collect()
    rss_end = get_rss_mib()
    print(f"{label}: +{rss_end - rss_start:.0f} MiB over 50 calls")


def main() -> None:
    numpyro.set_host_device_count(4)

    print(f"jax=={jax.__version__}, numpyro=={numpyro.__version__}\n")

    run_test("parallel,   unpatched", "parallel", patched=False)
    run_test("sequential, unpatched", "sequential", patched=False)
    run_test("parallel,   patched  ", "parallel", patched=True)
    run_test("sequential, patched  ", "sequential", patched=True)


if __name__ == "__main__":
    main()
```

Results (Linux `6.17.7-ba29.fc43.x86_64` - note that the leak seems slower on Mac M-series, but it still leaks):
```
jax==0.9.2, numpyro==0.20.1

parallel,   unpatched: +1127 MiB over 50 calls
sequential, unpatched: +132 MiB over 50 calls
parallel,   patched  : +10 MiB over 50 calls
sequential, patched  : +21 MiB over 50 calls
```

Edit: This is curious: I ran the reproducer with jax `0.7.2`, and it seems to still leak there (but quite a bit slower, and still fixed by this patch):
```
jax==0.7.2, numpyro==0.20.1

parallel,   unpatched: +318 MiB over 50 calls
sequential, unpatched: +86 MiB over 50 calls
parallel,   patched  : +8 MiB over 50 calls
sequential, patched  : +20 MiB over 50 calls
```

```
jax==0.8.0, numpyro==0.20.1

parallel,   unpatched: +1111 MiB over 50 calls
sequential, unpatched: +123 MiB over 50 calls
parallel,   patched  : +11 MiB over 50 calls
sequential, patched  : +23 MiB over 50 calls
```

```
jax==0.9.1, numpyro==0.20.0

parallel,   unpatched: +1097 MiB over 50 calls
sequential, unpatched: +138 MiB over 50 calls
parallel,   patched  : +9 MiB over 50 calls
sequential, patched  : +22 MiB over 50 calls
```

numpyro/infer/mcmc.py
```diff
@@ -699,13 +700,29 @@ def run(self, rng_key, *args, extra_fields=(), init_params=None, **kwargs):
...
+                args=args,
+                kwargs=kwargs,
+                collect_fields=collect_fields,
+                remove_sites=remove_sites,
+            )
+        else:
+            self._partial_map_fn.keywords["args"] = args
+            self._partial_map_fn.keywords["kwargs"] = kwargs
+            self._partial_map_fn.keywords["collect_fields"] = collect_fields
+            self._partial_map_fn.keywords["remove_sites"] = remove_sites
```
solutionseekeras, line 723: I wonder about the correctness of this: Perhaps the memoization should use `(id(func), *<these-four>)` instead of just the function.

solutionseekeras:
I am not confident of the correctness of this patch, mostly due to the compilation no longer being able to take the four `args`, `kwargs`, `collect_fields`, and `remove_sites` into account. Surely you guys will know this.

If the patch is not correct, however, I think this can be considered a bug report: It seems clear to me that there's a quite severe memory leak here, that should be addressed.

fehiepsi:
Would the leak happen if you run `jax.jit(mcmc.run)(key)`?

solutionseekeras:
No, that does seem to fix it (at least in the reproducer)! I'll test it out in the production code where we saw this initially.

fehiepsi:
Close since there is a better solution.

PR #2184, opened 2026-05-06
Fix effect handlers SWIG reference

jsakv:
Thanks for the in-depth tutorial!


Qazalbash (approved):
Thanks for the fix @jsakv

PR #2109, opened 2025-12-19
doc(gh-2107): Improved documentation for Negative Binomial distributions

Qazalbash:
This PR attempts to address the request for improving documentation for Negative Binomial distributions in #2107.

Fixes #2107.

PR #2185, opened 2026-05-07
doc: detailed documentation of `Beta` and `BetaProportion` distribution

Qazalbash:
This PR adds implementation details and mathematical references to the documentation for the `Beta` and `BetaProportion` distributions.

### DIFF
--- a/numpyro/distributions/continuous.py
+++ b/numpyro/distributions/continuous.py
@@ -177,6 +177,25 @@ class AsymmetricLaplace(Distribution):
 
 
 class Beta(Distribution):
+    r"""Beta distribution parameterized by concentration parameters alpha (:attr:`concentration1`)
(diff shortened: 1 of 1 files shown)


juanitorduz (approved):
LGTM!

In principle, do we want this for all distributions? Is that the plan? (We could do this over time). @fehiepsi ?

Qazalbash:
Following #2109, I planned to gradually improve the documentation by adding implementation details for each method. I would be delighted to see others contribute to this effort as well 😄.

juanitorduz:
> Following #2109, I planned to gradually improve the documentation by adding implementation details for each method. I would be delighted to see others contribute to this effort as well 😄.

Amazing! I can help as well!

fehiepsi:
That would be awesome!

PR #2186, opened 2026-05-07
dev: pre-commit hook to standardize toml format

Qazalbash:
This PR introduces a pre-commit hook of [tombi](https://github.com/tombi-toml/tombi) to standardize the toml format.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -50,3 +50,8 @@ repos:
     hooks:
       - id: yamlfmt
         args: [-formatter, retain_line_breaks=true]
+
+  - repo: https://github.com/tombi-toml/tombi-pre-commit
+    rev: v0.10.6
+    hooks:
+      - id: tombi-format
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,14 +11,14 @@ classifiers = [
     "Intended Audience :: Education",
     "Intended Audience :: Science/Research",
     "License :: OSI Approved :: Apache Software License",
-    "Operating System :: POSIX :: Linux",
     "Operating System :: MacOS :: MacOS X",
+    "Operating System :: POSIX :: Linux",
     "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: 3.12",
     "Programming Language :: Python :: 3.13",
     "Programming Language :: Python :: 3.14",
 ]
-keywords = ["probabilistic", "machine learning", "bayesian", "statistics"]
+keywords = ["bayesian", "machine learning", "probabilistic", "statistics"]
 dependencies = [
     "jax>=0.7.0",
     "jaxlib>=0.7.0",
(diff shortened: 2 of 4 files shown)


juanitorduz (approved):
:)

juanitorduz:
Do you also wanna check this in `make lint` and `make format`?

Qazalbash:
> Do you also wanna check this in `make lint` and `make format`?

Good idea 😄

Qazalbash:
To: @juanitorduz @fehiepsi 

> Do you also wanna check this in make lint and make format?

I have another proposal in mind.

Currently, we have pre-commit hooks configured for:

* `ruff format`
* `ruff check`
* `ty check`
* `mypy`

These checks are also included in `make lint`, which is why we skip them in CI whenever `prek` is instantiated.

Given this setup, would it make sense to consolidate the workflow and reduce the maintenance overhead by relying on a single entry point? If `prek run -a` is capable of handling all formatting, linting, and related checks consistently, we may no longer need separate `make lint` and `make format` targets.

What are your thoughts on this approach?

juanitorduz:
I agree! In most of my other projects, I simply use https://github.com/pre-commit/action :)

fehiepsi:
What's the setup that makes it easier to contribute without having to worry about toml/ty etc. dependency? I think we want to have minimal requirement for formatting. The rest is optional.

juanitorduz:
> What's the setup that makes it easier to contribute without having to worry about toml/ty etc. dependency? I think we want to have a minimal requirement for formatting. The rest is optional.

I personally do not think there is an *absolute* best way, and its more about opinions XD

My preference (this is an opinion, not a fact) is towards pre-commit hooks. Again, no strong preferences and the Make command also works fine.

Qazalbash:
> I personally do not think there is an absolute best way, and its more about opinions XD

I agree.

> What's the setup that makes it easier to contribute without having to worry about toml/ty etc. dependency?

In my opinion, pre-commit hooks are the better approach. We only need to install `pre-commit`/`prek`, and it can manage all additional dependencies automatically. The only maintenance overhead is periodically updating the hooks to their latest versions, which can also be automated using a GitHub bot.

juanitorduz:
This is a great point! We have pre-commit update PRs as https://github.com/pymc-labs/pymc-marketing/pull/2506

It needs minimal configuration:

https://github.com/pymc-labs/pymc-marketing/blob/e1eb8a32e8cba48e580ff36032e9715f99c97fed/.pre-commit-config.yaml#L1-L2

Qazalbash:
@juanitorduz @fehiepsi should we merge this PR? 😄

fehiepsi:
I guess so. Just to confirm, do I need to install all those hooks for the dev workflow? All I want is to use ruff to format the scripts and make a commit without worrying about those `.pre-commit-config.yaml‎` configurations, which I'm not familiar with.

Qazalbash:
No, you don't have to worry about hooks. All you need is to install `pre-commit`/`prek` (once) and then run them every time you want these hooks to run. `pre-commit`/`prek` will automatically install and manage those hooks leaving no responsibility on dev.

fehiepsi:
If I don't install those prek/pre-commit tools, would that be fine? If they are needed, how can I bypass them? I don't want those things block my workflow.

fehiepsi:
I don't care much about how useful they are, even if things are automatic.

Qazalbash:
No, they are not that necessary. You can bypass them.

PR #2189, opened 2026-05-16
Fix EnsembleSampler second split to use complementary inactive chains

amifalk:
The second sub-iteration in EnsembleSampler.sample incorrectly updated the second half of chains against itself instead of the first half. Add a deterministic regression test that fails under the previous split logic and verifies complementary-half updates.

**Original Author**: @jackkohm

### DIFF
--- a/numpyro/infer/ensemble.py
+++ b/numpyro/infer/ensemble.py
@@ -203,7 +203,7 @@ class EnsembleSampler(MCMCKernel, ABC):
             active, inactive = jax.lax.cond(
                 i == 0,
                 lambda x: (x[:split_ind], x[split_ind:]),
-                lambda x: (x[split_ind:], x[split_ind:]),
+                lambda x: (x[split_ind:], x[:split_ind]),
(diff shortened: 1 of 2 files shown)


amifalk:
The original PR (#2135) seems dead but was good to go save for failing lint.

amifalk:
@Qazalbash looks like this is failing because of the new toml formatting checks?

Qazalbash:
Thanks @amifalk. I have proposed a fix. You can merge the master into your branch after #2190 is merged.

Qazalbash:
This looks identical to #2135.

amifalk:
@Qazalbash It is. See: https://github.com/pyro-ppl/numpyro/pull/2189#issuecomment-4468224082 The original PR author seems unresponsive, but the fix is correct. Since this is a critical bug, I've pushed their changes here and updated them to pass lint so they can be merged.

fehiepsi:
Thanks for the fix!

Issue #2187, opened 2026-05-08
Enhance Distribution Documentation with Mathematical Explanations and Implementation Details


Qazalbash:
This issue tracks documentation updates that include mathematical explanations alongside implementation details for the distributions and methods listed below.

- [ ] `AsymmetricLaplace`
- [ ] `AsymmetricLaplaceQuantile`
- [x] `Bernoulli` (#2192)
- [x] `BernoulliLogits` (#2192)
- [x] `BernoulliProbs` (#2192)
- [x] `Beta` (#2185)
- [ ] `BetaBinomial`
- [x] `BetaNegativeBinomial` (#2109)
- [x] `BetaProportion` (#2185)
- [x] `Binomial` (#2192)
- [x] `BinomialLogits` (#2192)
- [x] `BinomialProbs` (#2192)
- [ ] `CAR`
- [ ] `Categorical`
- [x] `CategoricalLogits` (#2199)
- [x] `CategoricalProbs` (#2199)
- [x] `Cauchy` (#2188)
- [x] `Chi2` (#2199)
- [ ] `CirculantNormal`
- [ ] `Dagum`
- [ ] `Delta`
- [ ] `Dirichlet` (#2227)
- [ ] `DirichletMultinomial`
- [x] `DiscreteUniform` (#2199)
- [ ] `Distribution`
- [ ] `DoublyTruncatedPowerLaw`
- [ ] `EulerMaruyama`
- [ ] `ExpandedDistribution`
- [x] `Exponential` (#2188)
- [ ] `FoldedDistribution`
- [x] `Gamma` (#2109)
- [x] `GammaPoisson` (#2109)
- [ ] `GaussianCopula`
- [ ] `GaussianCopulaBeta`
- [ ] `GaussianRandomWalk`
- [ ] `GaussianStateSpace`
- [ ] `Geometric` (#2227)
- [ ] `GeometricLogits` (#2227)
- [ ] `GeometricProbs` (#2227)
- [ ] `Gompertz`
- [x] `Gumbel` (#2202)
- [ ] `HalfCauchy`
- [ ] `HalfNormal`
- [ ] `ImproperUniform`
- [ ] `Independent`
- [ ] `IntervalCensoredDistribution`
- [ ] `InverseGamma`
- [ ] `InverseWishart`
- [ ] `InverseWishartCholesky`
- [ ] `Kumaraswamy`
- [ ] `LKJ`
- [ ] `LKJCholesky`
[...]

PR #2192, opened 2026-05-20
doc(gh-2187): enhance documentation for Bernoulli and Binomial distributions

Qazalbash:


docs/source/conf.py
```diff
@@ -325,3 +325,7 @@
 # -- Suppress warnings in Sphinx 7.3.5
 
 suppress_warnings = ["config.cache"]
+
+# Show both class docstring and __init__ docstring in the documentation.
+# source: https://stackoverflow.com/questions/5599254/how-to-use-sphinxs-autodoc-to-document-a-classs-init-self-method
+autoclass_content = "both"
```
juanitorduz, line 331: I am unsure about the side effects of this change, so I asked Claude:

This is probably the most consequential change in the PR and it's easy to overlook. Setting autoclass_content = "both" means every class in the entire NumPyro documentation will now render both the class docstring and the __init__ docstring. For the classes touched here that's fine (the __init__ docstrings are well-written), but most other distribution classes don't have __init__ docstrings at all, so the rendered docs won't change for them — yet. The risk is that some existing classes might have __init__ docstrings with stale or confusing content that will now unexpectedly appear. I'd suggest either auditing the existing codebase for __init__ docstrings that might surface awkwardly, or scoping this more narrowly (e.g. using :autoclass_content: on a per-directive basis in distributions.rst if Sphinx supports it, or at least flagging this in the PR description).

numpyro/distributions/discrete.py
```diff
@@ -140,18 +242,53 @@ def sample(self, key: jax.Array, sample_shape: tuple[int, ...] = ()) -> ArrayLik
 
     @validate_sample
     def log_prob(self, value: ArrayLike) -> ArrayLike:
+        r"""Evaluate the log probability mass function at specified binary configurations.
```
juanitorduz, line 245: Comment from Claude:

The BernoulliLogits.log_prob docstring has double-escaped backslashes
In the log_prob docstring for BernoulliLogits, the :param value: line reads:
:param value: Binary observation(s) to score (:math:`k \\\\in \\\\{0, 1\\\\}`).
Since this is already a raw string (r"""), the double \\\\ will render literally as k \\in \\{0, 1\\} in the docs rather than the intended set notation. It should be k \in \{0, 1\} (single backslashes, as used correctly in the corresponding BernoulliProbs.log_prob docstring).

Can you build it locally and verify?

numpyro/distributions/discrete.py
```diff
@@ -115,23 +185,55 @@ def enumerate_support(self, expand: bool = True) -> ArrayLike:
...
+
+    Where :math:`\alpha = \text{logits}` is the log-odds parameter and
+    :math:`\sigma(\alpha) = 1/(1 + \exp{(-\alpha)})` is the sigmoid function.
+    """
+
     arg_constraints = {"logits": constraints.real}
+
     support = constraints.boolean
+    r"""The support of the Bernoulli distribution is the set of binary outcomes :math:`\{0, 1\}`."""
+
```
juanitorduz, line 213: I have never used this for class attributes, can you please test locally and see if works as expected?

juanitorduz (commented):
LGTM, some minor questions.

numpyro/distributions/discrete.py
Qazalbash: It is working, I have tested it.

numpyro/distributions/discrete.py
Qazalbash: I will fix them.

docs/source/conf.py
Qazalbash: I disagree with Claude's remarks. Let's keep the `autoclass_content`.

Qazalbash:
@juanitorduz how you are using Claude to review code?

juanitorduz:
> @juanitorduz how you are using Claude to review code?

Just pull the branch of the PR locally and then use Claude Code 🤖

fehiepsi (approved):
Nice, thanks!!

PR #2197, opened 2026-05-21
Expose `ZeroInflatedLogits` and `ZeroInflatedProbs` from top-level distributions module

Qazalbash:
After #2196, I checked each file in `numpyro/distributions/` and found `ZeroInflatedLogits` and `ZeroInflatedProbs` not exposed at top-level.


juanitorduz (approved):
Thanks @Qazalbash !

PR #2201, opened 2026-05-25
Fix enable_validation

renecotyfanboy:
Hi there, 
Here is a super small fix for validation. Currently, the validation context is handled using `_VALIDATION_ENABLED` global variable. By default `_VALIDATION_ENABLED=True` and `_validate_args=False` by default, which trigger weird behaviors when enabling / disabling the validation through contexts. But since everything is handled through mutating `_VALIDATION_ENABLED`, it sometimes produces weird behaviors. 

The simplest repro is : 

```
print(dist.Uniform(0, 1).log_prob(-0.5)) # -0.0 (expected)

with numpyro.validation_enabled(True):
    print(dist.Uniform(0, 1).log_prob(-0.5)) # -inf (expected)

print(dist.Uniform(0, 1).log_prob(-0.5)) # -inf (expect -0.0)
```

The current fix produces the expected behavior.

### DIFF
--- a/numpyro/compat/pyro.py
+++ b/numpyro/compat/pyro.py
@@ -3,11 +3,8 @@
 
 import warnings
 
-import numpy as np
-
 from numpyro.compat.util import UnsupportedAPIWarning
-from numpyro.distributions.transforms import biject_to
-from numpyro.util import find_stack_level, not_jax_tracer
+from numpyro.util import find_stack_level
 
 from numpyro.primitives import module, plate, sample  # noqa: F401 isort:skip
 from numpyro.primitives import param as _param  # noqa: F401 isort:skip
@@ -41,18 +38,4 @@ def param(name, *args, **kwargs):
         param_store = get_param_store()
         if name in param_store:
             val = param_store[name]
-    # Match Pyro's constrained-param semantics: an init value is treated as a
-    # constrained value and projected onto its constraint. NumPyro returns the
-    # raw init at trace time, so project it here when it is concrete and does not
-    # already satisfy the constraint (e.g. an unnormalized simplex value). Valid
-    # values (including boundary points) and tracers are left untouched.
-    constraint = kwargs.get("constraint")
-    if (
-        val is not None
-        and constraint is not None
-        and not_jax_tracer(val)
-        and not np.all(constraint(val))
-    ):
-        transform = biject_to(constraint)
-        val = transform(transform.inv(val))
     return val
--- a/numpyro/compat/pyro.py
+++ b/numpyro/compat/pyro.py
@@ -3,8 +3,11 @@
 
 import warnings
 
+import numpy as np
+
 from numpyro.compat.util import UnsupportedAPIWarning
-from numpyro.util import find_stack_level
+from numpyro.distributions.transforms import biject_to
+from numpyro.util import find_stack_level, not_jax_tracer
 
 from numpyro.primitives import module, plate, sample  # noqa: F401 isort:skip
 from numpyro.primitives import param as _param  # noqa: F401 isort:skip
@@ -38,4 +41,18 @@ def param(name, *args, **kwargs):
         param_store = get_param_store()
         if name in param_store:
             val = param_store[name]
+    # Match Pyro's constrained-param semantics: an init value is treated as a
+    # constrained value and projected onto its constraint. NumPyro returns the
+    # raw init at trace time, so project it here when it is concrete and does not
+    # already satisfy the constraint (e.g. an unnormalized simplex value). Valid
+    # values (including boundary points) and tracers are left untouched.
+    constraint = kwargs.get("constraint")
+    if (
+        val is not None
+        and constraint is not None
+        and not_jax_tracer(val)
+        and not np.all(constraint(val))
+    ):
+        transform = biject_to(constraint)
+        val = transform(transform.inv(val))
     return val
--- a/numpyro/contrib/hsgp/approximation.py
+++ b/numpyro/contrib/hsgp/approximation.py
@@ -30,13 +30,12 @@ def _non_centered_approximation(phi: Array, spd: Array, m: int) -> Array:
 
 
 def _centered_approximation(phi: Array, spd: Array, m: int) -> Array:
+    # ``spd`` is the square root of the spectral density and can underflow to
+    # exactly 0 for high-frequency basis functions; floor it to a tiny positive
+    # value so the (degenerate) coefficient keeps a valid positive scale.
+    spd = jnp.clip(spd, min=jnp.finfo(spd.dtype).tiny)
     with numpyro.plate("basis", m):
-        # ``spd`` is the square root of the spectral density and can underflow to
-        # exactly 0 for high-frequency basis functions, yielding a benign
-        # zero-variance (degenerate) coefficient. Skip validation for that edge.
-        beta = numpyro.sample(
-            "beta", dist.Normal(loc=0.0, scale=spd, validate_args=False)
-        )
+        beta = numpyro.sample("beta", dist.Normal(loc=0.0, scale=spd))
 
     return phi @ beta
 
--- a/numpyro/contrib/hsgp/approximation.py
+++ b/numpyro/contrib/hsgp/approximation.py
@@ -31,7 +31,12 @@ def _non_centered_approximation(phi: Array, spd: Array, m: int) -> Array:
 
 def _centered_approximation(phi: Array, spd: Array, m: int) -> Array:
     with numpyro.plate("basis", m):
-        beta = numpyro.sample("beta", dist.Normal(loc=0.0, scale=spd))
+        # ``spd`` is the square root of the spectral density and can underflow to
+        # exactly 0 for high-frequency basis functions, yielding a benign
+        # zero-variance (degenerate) coefficient. Skip validation for that edge.
+        beta = numpyro.sample(
+            "beta", dist.Normal(loc=0.0, scale=spd, validate_args=False)
+        )
 
     return phi @ beta
 
--- a/numpyro/contrib/hsgp/approximation.py
+++ b/numpyro/contrib/hsgp/approximation.py
@@ -33,7 +33,7 @@ def _centered_approximation(phi: Array, spd: Array, m: int) -> Array:
     # ``spd`` is the square root of the spectral density and can underflow to
     # exactly 0 for high-frequency basis functions; floor it to a tiny positive
     # value so the (degenerate) coefficient keeps a valid positive scale.
-    spd = jnp.clip(spd, min=jnp.finfo(spd.dtype).tiny)
+    spd = jnp.clip(spd, jnp.finfo(spd.dtype).tiny)
     with numpyro.plate("basis", m):
         beta = numpyro.sample("beta", dist.Normal(loc=0.0, scale=spd))
 
--- a/numpyro/distributions/continuous.py
+++ b/numpyro/distributions/continuous.py
@@ -654,7 +654,12 @@ class EulerMaruyama(Distribution):
         mu = xtm1 + dt * f
         sigma = jnp.sqrt(dt) * g
 
-        sde_log_prob = Normal(mu, sigma).to_event(self.event_dim).log_prob(xt)
+        # ``mu``/``sigma`` are derived from ``value``; out-of-support values
+        # would make them invalid, so skip validation of this internal
+        # distribution. The public log_prob already validates ``value``.
+        sde_log_prob = (
+            Normal(mu, sigma, validate_args=False).to_event(self.event_dim).log_prob(xt)
+        )
         init_log_prob = self.init_dist.log_prob(value0)
 
         return sde_log_prob + init_log_prob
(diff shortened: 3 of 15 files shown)


juanitorduz:
@fehiepsi : shall we keep `_VALIDATION_ENABLED` and `_validate_args` equal to `True`?

fehiepsi:
We decided to enable validation by default. I think setting the attribute to the global one might fix the issue.

Qazalbash (approved):

numpyro/distributions/distribution.py
```diff
@@ -64,7 +64,7 @@
 
 from . import constraints
 
-_VALIDATION_ENABLED = True
+_VALIDATION_ENABLED = False
```
juanitorduz: Based on the comment above, we should keep `_VALIDATION_ENABLED = True`

juanitorduz (changes requested):
Minor point, right @Qazalbash ?

Qazalbash (changes requested):
@juanitorduz Yes, the global variable value should be reverted.

numpyro/distributions/distribution.py
renecotyfanboy: This is fixed. Note that this changes the behavior by default and might be a breaking change for some codebases. However, I find that it is much better to have validation enabled by default, as it was quite unintuitive in the first place.

Qazalbash (approved):
Thanks @renecotyfanboy

Qazalbash:
I think CI will trigger after @juanitorduz approves. Let's merge this PR only when the latest CI passes.

juanitorduz (approved):

juanitorduz:
@renecotyfanboy thanks! Can you push an empty commit (e.g., git commit -m"empty" --allow-empty ) to try to trigger the CI? 🙏

renecotyfanboy:
Enabling validation by default broke some tests in the CI. Should I simply wrap them around `with numpyro.validation_enabled(False)` ? I am not sure if I am willing to fix them properly 😆

Qazalbash:
Let's leave things as they were. What do you say @juanitorduz?

fehiepsi:
Could we fix the tests instead? This indicates that we might not use correct params for those tests.

juanitorduz:
Indeed, that is a better approach! I could help fix these tests if you feel stuck @renecotyfanboy  :)

renecotyfanboy:
I started fixing a bunch of tests here. 

- Disabled validation for the categorical variables since most of them are not normalized in the tests
- Disabled validation for the Gaussian process contributed tests 
- Used better MCMC params to ensure convergence for the functional map tests 
- Used better random params for the negative binomial log prob tests 

However, I see that it is taking a super long time for me to look at each case, each distribution, and find what is ill-defined in the parameter used in space or in the batch / event shape mismatch. If I am to continue fixing the test, I will probably switch to LLMs to iterate faster.

juanitorduz:
Thank you @renecotyfanboy ! I think Claude (or similar) can be super useful here! If you need some help please let me know :)

renecotyfanboy:
I just tried to run copilot+claude for this and it spent half my monthly credit just to figure out that it should disable validation at runtime while explicitly prompted not to do so. LMAO. This task is too hard for me, and I won't have time to work on this in the next months. Maybe we should disable validation by default (as it is currently the case), and set as a long term goal to fix the tests using proper and valid test inputs. WDYT?

juanitorduz:
> I just tried to run copilot+claude for this and it spent half my monthly credit just to figure out that it should disable validation at runtime while explicitly prompted not to do so. LMAO. This task is too hard for me, and I won't have time to work on this in the next months. Maybe we should disable validation by default (as it is currently the case), and set as a long term goal to fix the tests using proper and valid test inputs. WDYT?

Thanks for the effort @renecotyfanboy ! Not an easy one ;) 

Let me try to see if I can make it work (please do not close this PR)

juanitorduz:
ok @renecotyfanboy Its seems its working now :)

FYI: @Qazalbash @fehiepsi

numpyro/compat/pyro.py
```diff
@@ -38,4 +41,18 @@ def param(name, *args, **kwargs):
...
+    # values (including boundary points) and tracers are left untouched.
+    constraint = kwargs.get("constraint")
+    if (
+        val is not None
+        and constraint is not None
+        and not_jax_tracer(val)
+        and not np.all(constraint(val))
+    ):
+        transform = biject_to(constraint)
+        val = transform(transform.inv(val))
```
fehiepsi: Why this is needed? I don't quite understand the comment above.

numpyro/contrib/hsgp/approximation.py
```diff
@@ -31,7 +31,12 @@ def _non_centered_approximation(phi: Array, spd: Array, m: int) -> Array:
 
 def _centered_approximation(phi: Array, spd: Array, m: int) -> Array:
     with numpyro.plate("basis", m):
-        beta = numpyro.sample("beta", dist.Normal(loc=0.0, scale=spd))
+        # ``spd`` is the square root of the spectral density and can underflow to
+        # exactly 0 for high-frequency basis functions, yielding a benign
+        # zero-variance (degenerate) coefficient. Skip validation for that edge.
+        beta = numpyro.sample(
+            "beta", dist.Normal(loc=0.0, scale=spd, validate_args=False)
+        )
```
fehiepsi: Best to clip here i believe

numpyro/distributions/continuous.py
```diff
@@ -654,7 +654,12 @@ def log_prob(self, value: ArrayLike) -> ArrayLike:
         mu = xtm1 + dt * f
         sigma = jnp.sqrt(dt) * g
 
-        sde_log_prob = Normal(mu, sigma).to_event(self.event_dim).log_prob(xt)
+        # ``mu``/``sigma`` are derived from ``value``; out-of-support values
+        # would make them invalid, so skip validation of this internal
+        # distribution. The public log_prob already validates ``value``.
+        sde_log_prob = (
+            Normal(mu, sigma, validate_args=False).to_event(self.event_dim).log_prob(xt)
+        )
```
fehiepsi: Likewise, best to clip here

numpyro/distributions/continuous.py
```diff
@@ -1094,9 +1099,14 @@ def sample(self, key: jax.Array, sample_shape: tuple[int, ...] = ()) -> ArrayLik
...
-        init_prob = Normal(0.0, self.scale).log_prob(value[..., 0])
+        # the step distribution uses ``value`` as its location, so out-of-support
+        # values would make it invalid; skip validation of these internal
+        # distributions since the public log_prob already validates ``value``.
+        init_prob = Normal(0.0, self.scale, validate_args=False).log_prob(value[..., 0])
         scale = jnp.expand_dims(self.scale, -1)
-        step_probs = Normal(value[..., :-1], scale).log_prob(value[..., 1:])
+        step_probs = Normal(value[..., :-1], scale, validate_args=False).log_prob(
+            value[..., 1:]
+        )
```
fehiepsi: Can we address the issue instead of falling back to validate_args

numpyro/compat/pyro.py
juanitorduz: This is something I found while talking with Claude and doing some tests:

Pyro treats a constrained param's init value as a *constrained* value and, if it does not satisfy the constraint, projects it through the constraint's bijector. NumPyro's `param` instead returns the raw init at trace time (the constraint is only used later by SVI to build the unconstrained optimizer). So during SVI.init's first guide trace the raw init reaches the model/guide; e.g. pyroapi's test_constraints inits a simplex param with an unnormalized exp(randn(3)), which now trips validation-by-default in dist.Categorical.
Project it here (only when concrete and not already valid) to match Pyro.

It could be I am missing something 🙈 , but it seems this helps make the test pass while having the validation.

fehiepsi: It is unclear why pyro is related here. Does this imply the test is not robust? I think transform(transform.inv(val)) is the same as val for bijector. Do we want to address non bijective transforms?

juanitorduz: > It is unclear why pyro is related here.

I just compared it with what Pyro was doing XD to see if there was an issue here.

> Does this imply the test is not robust? I think transform(transform.inv(val)) is the same as val for bijector. Do we want to address non-bijective transforms?

Let me investigate . . . 🤔

numpyro/contrib/hsgp/approximation.py
juanitorduz: fixed in https://github.com/pyro-ppl/numpyro/pull/2201/changes/b7cb852fdde1fac5132cce4f86ed4fec18f22837

numpyro/distributions/continuous.py
juanitorduz: fixed in https://github.com/pyro-ppl/numpyro/pull/2201/changes/b7cb852fdde1fac5132cce4f86ed4fec18f22837

numpyro/distributions/continuous.py
juanitorduz: fixed in https://github.com/pyro-ppl/numpyro/pull/2201/changes/b7cb852fdde1fac5132cce4f86ed4fec18f22837

numpyro/contrib/hsgp/approximation.py
```diff
@@ -30,6 +30,10 @@ def _non_centered_approximation(phi: Array, spd: Array, m: int) -> Array:
 
 
 def _centered_approximation(phi: Array, spd: Array, m: int) -> Array:
+    # ``spd`` is the square root of the spectral density and can underflow to
+    # exactly 0 for high-frequency basis functions; floor it to a tiny positive
+    # value so the (degenerate) coefficient keeps a valid positive scale.
+    spd = jnp.clip(spd, min=jnp.finfo(spd.dtype).tiny)
```
fehiepsi: I remember min is not available for old jax release. Could you check if this is consistent to the current usage.

numpyro/distributions/continuous.py
```diff
@@ -654,7 +654,11 @@ def log_prob(self, value: ArrayLike) -> ArrayLike:
         mu = xtm1 + dt * f
         sigma = jnp.sqrt(dt) * g
 
-        sde_log_prob = Normal(mu, sigma).to_event(self.event_dim).log_prob(xt)
+        # Normal is location-invariant, so evaluate the residual under a
+        # zero-mean Normal. This keeps the loc valid even for out-of-support
+        # ``value`` (where ``mu`` would be NaN); the public log_prob's
+        # @validate_sample still warns about out-of-support values.
+        sde_log_prob = Normal(0.0, sigma).to_event(self.event_dim).log_prob(xt - mu)
```
fehiepsi, line 661: Nice solution to address NaN mu.

fehiepsi (approved):
Looks great. Thanks both!

numpyro/compat/pyro.py
juanitorduz: Ok! This is how far I got:

- **On `transform(transform.inv(val)) == val`:** You're right,  for a true bijector it's the identity. I confirmed it's a no-op for `positive`, `real`, `unit_interval`, and `greater_than`. **It only changes** `val` for **non-bijective, dimension-changing** transforms like `simplex` (StickBreaking maps `R^(n-1) ↔` the simplex).

- **Remark:** NumPyro represents a constrained `param` by its *unconstrained* value, so the constrained value SVI actually optimizes from is `transform(transform.inv(init))`. I verified this directly: for a simplex `param` initialized with `exp(randn(3))`, the value SVI uses is `[9.99e-1, 1.19e-7, 1.42e-14]`: exactly the round-trip (and notably **not** `val/sum`). NumPyro's `param` primitive returns the **raw** init at trace time, so during `SVI.init`'s first guide trace, `Categorical` receives the unnormalized vector and trips validation-by-default. The projection just makes the traced value consistent with what SVI uses one step later. It's the same thing Pyro's param store does,

**Does it imply the test isn't robust?** Somewhat, `test_constraints` is a smoke test that constrained params don't crash SVI; it never checks the actual values, so the near-one-hot projection "passes" trivially. It's the **only** test that needs this (enum_elbo's probs are all valid, so its `pyro.param` calls never enter the projection branch — the `not np.all(constraint(val))` guard skips them, which also protects valid boundary values like `[[0,1],[1,0]]` from being perturbed).

We then have two options: 

- Keep the change (it's consistent with SVI's own init, just needs a clearer comment)
- `xfail` the single vendored `test_constraints` if you'd prefer no compat-layer behavior change.

Both keep CI green.

WDYT?

fehiepsi: I see. Seems like a test issue where the value is not valid. Could we fix the test instead? Just ensure we provide valid params there. If it is too difficult, just apply this double transform there. I still don't understand why we can't fix this in the test

juanitorduz: ok! Let me try this out (I first needed to understand this edge case)

fehiepsi: We can change simplex param to something other than exp(randn(3))

numpyro/contrib/hsgp/approximation.py
juanitorduz: Removed the explicit name as we can pass it by position to be safer https://github.com/pyro-ppl/numpyro/pull/2201/commits/8a75729b1e77f3601ee69098d1ad7ca687a7122b

test/pyroapi/test_pyroapi.py
```diff
@@ -26,3 +28,29 @@
 def backend():
     with pyro_backend("numpy"):
         yield
+
+
+# pyroapi's test_constraints inits the simplex param `q` with an unnormalized
+# exp(randn(3)); use a valid simplex so it passes with validation enabled.
```
juanitorduz, line 34: `exp(randn(3))` for `q` lives in the vendored upstream `pyroapi` package (`pyroapi/tests/test_svi.py`), see  https://github.com/pyro-ppl/pyro-api/blob/master/pyroapi/tests/test_svi.py#L168

To provide a valid simplex (as suggested) we therefore redefine `test_constraints` in our own `test/pyroapi/test_pyroapi.py`, identical to upstream except `q` is initialized to a valid simplex `[0.4, 0.3, 0.3]` instead of the unnormalized `exp(randn(3))`. Since the test name is imported via `from pyroapi.tests import *`, redefining it **shadows** the upstream one, it does not add a test.

juanitorduz (commented):
Small comment

numpyro/compat/pyro.py
juanitorduz: done in https://github.com/pyro-ppl/numpyro/pull/2201/commits/e03f20e6da352bbdd03aa62b35fcb189cc794c34

juanitorduz:
@fehiepsi is this one good to fo :) ?

PR #2199, opened 2026-05-23
doc(gh-2187): enhance documentation for Chi2, DiscreteUniform, and Categorical distributions

Metbcy:
Adds math-flavored docstrings to four more distributions on the #2187 checklist, following the template from #2188 and #2192.

- `Chi2` (class + `__init__`) — also notes the `Gamma(k/2, 1/2)` equivalence so readers know why the other methods aren't redocumented here.
- `DiscreteUniform` — class, `__init__`, `sample`, `log_prob`, `cdf`, `icdf`, `mean`, `variance`, `enumerate_support`, `entropy`.
- `CategoricalProbs` — same surface; `mean`/`variance` documented as `NaN` since the categories are unordered.
- `CategoricalLogits` — same, with the `logsumexp` stability note on `log_prob` and `entropy`.

Docstrings only, no logic changes. `ruff check` and `ruff format --check` are clean on both files.

PR #2188, opened 2026-05-09
doc(gh-2187): add math explanations for Normal, Cauchy, and Exponential distributions

kimjune01:
This PR adds mathematical explanations and implementation details to the docstrings for the `Normal`, `Cauchy`, and `Exponential` distributions, continuing the documentation effort from #2185.

PR #2202, opened 2026-05-27
doc(gh-2187): docstrings for Gumbel, Laplace, Logistic

Metbcy:
Refs #2187.

Picks up three more continuous distributions on the docs checklist, following the same template as #2188, #2192, and #2199:

- `Gumbel` — class, `__init__`, `sample`, `log_prob`, `mean`, `variance`, `cdf`, `icdf`
- `Laplace` — class, `__init__`, `sample`, `log_prob`, `mean`, `variance`, `cdf`, `icdf`, `entropy`
- `Logistic` — class, `__init__`, `sample`, `log_prob`, `mean`, `variance`, `cdf`, `icdf`, `entropy`

All three share the `loc` / `scale` real / positive shape, so the docstrings line up nicely. PDF / log-PDF / CDF / quantile formulas use short symbols (μ, β, b, s) and `:attr:` cross-refs to the parameter names, matching the style asked for in the #2199 review.

Docstrings only — no logic changes.

Verified:
- `ruff check numpyro/distributions/continuous.py` clean
- `ruff format --check numpyro/distributions/continuous.py` clean
- `pytest test/test_distributions.py -k "Gumbel or Laplace or Logistic"` → 511 passed, 87 skipped

### DIFF
--- a/numpyro/distributions/continuous.py
+++ b/numpyro/distributions/continuous.py
@@ -1306,6 +1306,22 @@ class Gompertz(Distribution):
 
 
 class Gumbel(Distribution):
+    r"""The Gumbel (maximum) distribution, a continuous real-valued
+    distribution parameterized by location :math:`\mu` and scale :math:`\beta > 0`.
+    It is the limiting distribution of the maximum of a large number of i.i.d.
+    samples from an exponential-tailed distribution.
+
+    The Probability Density Function (PDF) is:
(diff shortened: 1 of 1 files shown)


Qazalbash (approved):
I have pushed some fixes. Rest is good. Thanks @Metbcy.

PR #2203, opened 2026-05-29
Improve integration with external samplers

juanitorduz:
Motivated by https://github.com/pyro-ppl/numpyro/pull/2124 and specifically by https://github.com/pyro-ppl/numpyro/pull/2124#issuecomment-4431281149, we suggest a cleaner API to integrate with other samplers. This includes MCLMC from Blackjax. This is illustrated in the notebook.

(most from the code changes come from the notebook)


numpyro/infer/util.py
```diff
@@ -801,6 +824,149 @@ def initialize_model(
...
+        forward_mode_differentiation=forward_mode_differentiation,
+        validate_grad=validate_grad,
+    )
+    bound_potential = model_info.potential_fn(*model_args, **model_kwargs)
+    bound_postprocess = model_info.postprocess_fn(*model_args, **model_kwargs)
+
+    def logdensity_fn(position: PositionDict) -> jax.Array:
+        return -bound_potential(position)
+
+    return LogDensityInfo(
```
fehiepsi: I guess you can simply add a bound flag and return additional log_density_fn (or better name) in initialize model

numpyro/infer/util.py
```diff
@@ -801,6 +824,149 @@ def initialize_model(
...
+
+
+def constrain_samples(
+    samples: PositionDict,
+    model: Callable[..., Any],
+    *,
+    model_args: tuple[Any, ...] = (),
+    model_kwargs: Optional[dict[str, Any]] = None,
+    return_deterministic: bool = True,
+    batch_ndims: int = 1,
```
fehiepsi: If this only introduces batch_ndims, let's add it in constrain_fn

numpyro/infer/external.py
```diff
@@ -0,0 +1,403 @@
...
+    passed to it, so the placeholder needs matching shape/dtype but never
+    appears in the collected output (the first stored sample is post-step).
+    ``jax.eval_shape`` abstracts ``rng_key`` and ``inner``, so the key value
+    does not matter; we pass ``random.key(0)`` as a dummy.
+    """
+    info_shape = jax.eval_shape(step_fn, random.key(0), inner)[1]
+    return jax.tree.map(lambda s: jnp.zeros(s.shape, s.dtype), info_shape)
+
+
+class ExternalKernel(MCMCKernel):
```
fehiepsi: Why do we need this class? The example seems to indicate that users will have more flexibility when subclass the mcmckernel directly, rather than constructing build_mclmc

numpyro/infer/util.py
juanitorduz: Added it in https://github.com/pyro-ppl/numpyro/pull/2203/commits/1329c580493f0acf07e4095b77457e7b881afd7b

numpyro/infer/util.py
juanitorduz: Added it in https://github.com/pyro-ppl/numpyro/pull/2203/commits/1329c580493f0acf07e4095b77457e7b881afd7b

numpyro/infer/external.py
juanitorduz: You are right, this helper class is less flexible (just experimenting with a friendly API). I have now modified the example to inhering directly from `MCMCKernel` in Added it in https://github.com/pyro-ppl/numpyro/pull/2203/commits/1329c580493f0acf07e4095b77457e7b881afd7b

juanitorduz:
Thank you for the feedback @fehiepsi . Let me know how this iteration looks. I am happy to keep iterating until we get the best level of abstraction :)

reubenharry:
Would people be happy merging this? Certainly it looks good from my end

juanitorduz:
I would 😄!

fehiepsi:
Let's wait a bit to rethink whether the design choices are reasonable. Looking like we are changing other internal codes due to the change. It indicates that we will likely break current user code. Let's see if we can come up with simpler solutions.

juanitorduz:
Great! Always open for feedback 🤗

juanitorduz:
@fehiepsi you were right, there was no need to add such a large change to the internal API 🙈 . In https://github.com/pyro-ppl/numpyro/pull/2203/commits/a463116af7d9db2b451eee5b4b0b207b0f8b250a, I simplified the approach and minimized the changes. Now this PR is way less invasive 🤗 

fyi @reubenharry

reubenharry:
Yeah, this seems pleasantly concise

juanitorduz:
@fehiepsi, what do you think about the current status? Any additional feedback or concerns? :)

numpyro/_typing.py
```diff
@@ -35,4 +35,12 @@
 """A generic type for a pytree, i.e. a nested structure of lists, tuples, dicts, and arrays."""
 
 
+PositionDict: TypeAlias = dict[str, jax.Array]
```
fehiepsi: this seems unncessary?

numpyro/infer/util.py
```diff
@@ -801,6 +854,77 @@ def initialize_model(
     )
 
 
+def get_log_density_fn(
```
fehiepsi: This utility seems unnecessary? We can add 2 lines of code in the notebook instead. :)

numpyro/_typing.py
juanitorduz: Indeed, I removed it

numpyro/infer/util.py
juanitorduz: Also removed it ;)

juanitorduz:
Thanks @fehiepsi I simplified it as suggested and resolved the merge conflicts (sorry for the commit mess, but we will squash it anyway :) )

fehiepsi:
Looks great, thanks!

PR #2205, opened 2026-06-08
Fix GaussianStateSpace.variance shape when batch comes from initial_value

tillahoffmann:
## Problem

`GaussianStateSpace.variance` was computed from `scale_tril` alone, which drops any batch dimension contributed by `initial_value`. When the batch shape originates from `initial_value` (e.g. `scale_tril` of shape `(2, 2)` with `initial_value` of shape `(3, 2)`), `variance` came out as `(num_steps, state_dim)` instead of the required `batch_shape + event_shape`.

Concretely, for `num_steps=5`, transition `(2, 2)`, `scale_tril` `(2, 2)`, `initial_value` `(3, 2)`:
- `batch_shape = (3,)`, `event_shape = (5, 2)` → expected moment shape `(3, 5, 2)`
- `variance` returned `(5, 2)`

## Fix

Broadcast the variance to `batch_shape + event_shape`. The initial value is deterministic and does not affect the variance *values*, only the shape, so broadcasting is exact.

## Tests

Added a uniform moment-shape contract check in `test_mean_var` asserting that `mean`/`variance` (where implemented) have shape `batch_shape + event_shape` for every distribution. With the fix, the full `test_mean_var` suite passes (162 passed, 61 skipped, 3 xpassed, 0 failed).

### DIFF
--- a/numpyro/distributions/continuous.py
+++ b/numpyro/distributions/continuous.py
@@ -1045,11 +1045,15 @@ class GaussianStateSpace(TransformedDistribution):
             self.scale_tril,
             jnp.arange(self.num_steps),
         )
-        return (
+        variance = (
             jnp.diagonal(scale_tril @ scale_tril.mT, axis1=-1, axis2=-2)
             .cumsum(axis=0)
             .swapaxes(0, -2)
         )
+        # The variance does not depend on the (deterministic) initial value, but the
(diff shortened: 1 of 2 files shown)


juanitorduz (approved):
LGTM, thanks!

PR #2206, opened 2026-06-08
Ensure distribution moments/samples are jax arrays

tillahoffmann:
## Summary

This PR makes numpyro distributions return concrete `jax.Array` from their public methods (`mean`, `variance`, `entropy`, `sample`, `log_prob`, `cdf`, `icdf`, …) and annotates them accordingly, so that **typed downstream code can actually use distribution outputs**. To make that true at runtime, parameters are coerced to `jax.Array` at construction.

The change aligns numpyro with [JAX's own typing best practices](https://docs.jax.dev/en/latest/jax.typing.html#jax-typing-best-practices) — **wide on input (`ArrayLike`), strict on output (`Array`)** — and removes a class of type errors that typed consumers will hit once they rely on numpyro's types.

The near-term, standalone win is the **parameter coercion**: today whether a stored parameter (`self.loc`, `self.rate`, …) is a `jax.Array`, a numpy array, or a python scalar depends on the input — this PR makes it consistently a `jax.Array`. The output annotations are the cheap, correct-by-construction follow-through; they become observable to consumers once numpyro ships `py.typed` (a deliberate follow-up — see below).

This is necessarily runtime work, not just an annotation change: many moments return a stored parameter directly (`Poisson.mean` is `self.rate`), so an `Array` return type is only *honest* if that parameter is actually an array — typing the outputs without making the runtime match would just lie to the type checker.

It also changes one behavior deliberately: **argument validation (`validate_args=True`) is now an eager-only operation.** That is the part most worth discussing, so it is called out explicitly below.

## Motivation: the output annotations are wrong, and it bites consumers

Public methods were annotated to return `jax.typing.ArrayLike`:

```python
ArrayLike = Union[jax.Array, np.ndarray, np.bool, np.number, bool, int, float, complex]
```

`ArrayLike` is a deliberately *wide input* union. JAX's [typing guidance](https://docs.jax.dev/en/latest/jax.typing.html) is explicit that it is the right type for **function inputs** (be liberal in what you accept) and that **outputs should be annotated `Array`** (be strict in what you return). numpyro applied `ArrayLike` to outputs too, which is incorrect: `mean`/`sample`/`log_prob`/… (should) always produce a concrete array.

The cost lands on every typed consumer. The scalar members of the union (`bool`/`int`/`float`/`complex`) support neither indexing, `.shape`, `.reshape`, nor assignment to an `Array` parameter, so ordinary usage fails type checking:

```python
d = dist.Normal(jnp.zeros(5), 1.0)
d.mean[3]                  # error: int/float/complex are not indexable
d.variance.shape           # error: .shape not defined on int/float/complex
some_jit_fn(d.mean)        # error: ArrayLike not assignable to Array
d.mean > 1                 # error: ">" not supported for "complex" and "Literal[1]"
```

Under pyright, **every one of those lines is a type error** when the method returns `ArrayLike` (each indexing/attribute access fails once per scalar member of the union — `int`, `float`, `complex`, …), and **all of them type-check cleanly** once it returns `Array`.

This is currently latent (numpyro ships no `py.typed`, so checkers treat it as untyped `Any`), but it is a trap: the moment a consumer relies on numpyro's types — or numpyro starts enforcing its own — these annotations break real code.

The benefit here is for *downstream* typed code. This PR does not aim to make numpyro's own modules type-clean: they are not mypy-enforced, and their residual errors are dominated by pre-existing `[override]` signature mismatches unrelated to these annotations (enabling enforcement is called out as out of scope below).

## What this PR does

1. **Coerce parameters to `jax.Array` at construction.** Previously parameter storage was inconsistent: most parameters flow through `promote_shapes`, which preserved input types and only produced arrays when a broadcast forced a reshape, while some distributions stored the raw argument directly — so whether `self.loc` ended up an array depended on the input shapes and dtype. Now `promote_shapes` takes an opt-in `promote_array=True` (typed to return `list[Array]`) used at the parameter-assignment sites, and the few directly-assigned parameters are wrapped in `jnp.asarray`. As a result `self.loc`, `self.scale`, … are statically and dynamically `jax.Array`.

This is runtime work, not just annotation, on purpose: annotating the outputs as `Array` without actually returning arrays would be a *lie* to the type checker — many moments simply return a stored parameter (`Poisson.mean` is `self.rate`), so the annotation is only honest if the parameter is genuinely an array. Typing that isn't backed by the runtime is worse than no typing.

2. **Annotate outputs as `Array`.** Return-position `ArrayLike` becomes `Array` across the distribution modules. Constructor parameters and method arguments (`value`, `q`) keep `ArrayLike`, so callers can still pass python scalars / numpy arrays.

3. **Make instance types reachable through the metaclass.** `DistributionMeta.__call__`   is annotated to return the concrete class being constructed, so `dist.Normal(...)` is typed `Normal` (not `Any`, and not the base `Distribution`) and the `-> Array` annotations on its methods are visible to consumers.

4. **Use functional array ops on inputs where appropriate.** Where a method only needed a shape it now uses `jnp.shape(value)` rather than `value.shape`, which is correct for any `ArrayLike` (and incidentally more robust). A few methods that genuinely index `value` coerce it once.

Parameters that must remain non-array are preserved: pytree *aux* fields (`total_count` on the multinomial family, `adj_matrix` on `CAR`) are static metadata — coercing them to a `jax.Array` would make them tracers under `jit`/`vmap` and break sampling — so they are left un-coerced, as is a scipy-sparse `adj_matrix` (`CAR(is_sparse=True)`).

The coercion touches a parameter-assignment site in essentially every distribution, so correctness rests on tests that run across **all** distributions rather than on per-site review:

- `test_params_and_outputs_are_arrays` — samples, moments, and stored (non-aux) parameters are `jax.Array`. This catches any data parameter the sweep missed.
- `test_aux_fields_are_not_jax_arrays` — no pytree aux field is ever a `jax.Array`. This catches the opposite mistake: an aux field (e.g. `total_count`) wrongly coerced, which would break `jit`/`vmap` (for instance, a jitted `Multinomial(...).sample(...)` needs a concrete `total_count`).

## The deliberate behavior change: validation is eager-only

Because parameters are coerced to arrays before `Distribution.__init__` runs `validate_args`, an **out-of-bounds constant parameter inside `jit` no longer raises**. Previously (#775) it did, by keeping parameters as concrete (non-jax) values so the validity check could constant-fold. #775 added this so that a typo'd literal — `Normal(0., -1.)` in a jitted model — would fail loudly at trace time rather than silently produce NaNs. That is a real concern, so this is worth making the case for rather than glossing over.

**The guarantee it provided was always narrow.** It fired *only* for parameters that were compile-time constants; the moment the same bad value arrives through a traced argument (`Normal(0., scale)` with `scale = -1.`), validation cannot run — you can't branch on a tracer — and never has. So on `master` today, with the *same* `validate_args=True` and the *same* invalid value, whether you get an error depends on how the value was staged:

```python
# constant invalid scale -> raises
jax.jit(lambda: dist.Normal(0.0, -1.0, validate_args=True).log_prob(0.5))()

# traced invalid scale -> silently does NOT raise (same bad value!)
jax.jit(lambda s: dist.Normal(0.0, s, validate_args=True).log_prob(0.5))(-1.0)
```

A safety net that catches `-1.0` written as a literal but not `-1.0` passed as an argument is not one a user can rely on without knowing how their parameters get staged.

The new behavior is simpler to state: **`validate_args=True` validates eagerly and does nothing under `jit`.** Eager validation is unchanged — invalid parameters raise exactly as before outside `jit`, which is where it is most useful for debugging (construct the distribution eagerly, or run once without `jit`, to surface the error).

This also lines up with how out-of-support *sample* validation already behaves. That check warns only when the support mask is concrete, so it too goes silent under `jit` whenever the sampled value is traced — which is the usual case, since in a model/MCMC/SVI the value fed to `log_prob` is the (traced) data. The only situation where it still fires under `jit` is a constant-support distribution with a constant value, the same accidental-constant-folding corner that parameter validation used to rely on. So under realistic jitted usage both parameter and sample validation are effectively eager-only.

## Performance

Coercion is a `jnp.asarray` per parameter, but it is skipped for arguments that are already `jax.Array` (a cheap `isinstance` check instead of a dispatch), so the cost falls almost entirely on constructing from python scalars / numpy. Eager-construction microbenchmarks (min of repeated runs):

| construction | master | branch |
| --- | --- | --- |
| `Normal(0., 1.)` (python scalars) | ~5.4 µs | ~31 µs |
| `Normal(jnp.asarray(0.), jnp.asarray(1.))` | ~1.8 µs | ~2.0 µs |
| `Normal(jnp.zeros(100), 1.)` | ~9.7 µs | ~10.1 µs |
| `MultivariateNormal(jnp.zeros(5), jnp.eye(5))` | ~30.7 µs | ~30.6 µs |

So **constructing from jax arrays — the typical case in real code, and what happens under tracing — is unaffected.** Only constructing from python-scalar *literals* pays the full conversion (`Normal(0., 1.)` ≈5×), and even then it is tens of microseconds. The jitted MCMC/SVI path is unchanged: the model and the distributions it builds are traced, so the coercion folds away (a jitted log-density constructing `Normal` + `Beta` is ~3.1 µs/call on both branches, with equal compile time).

## Explicitly *not* in this PR

**Enabling mypy enforcement** for the distribution modules: the output annotations are now correct, but fully type-clean modules also require resolving pre-existing `[override]` signature mismatches and `Optional` handling, which are out of scope here.

**Widening constructor inputs to `ArrayLike`** — see Next steps.

## Next steps

This PR fixes the *output* side of the convention (strict `Array` out). The *input* side is a natural follow-up: a number of constructor parameters are still annotated with the strict `Array` (e.g. `CategoricalProbs.probs`, `Dirichlet.concentration`, `MatrixNormal.loc/scale_tril_*`, `LowRankMultivariateNormal.*`, `Distribution.mask`), so passing a numpy array or a python scalar to them is a type error even though it works at runtime. These should be widened to `ArrayLike` (be liberal in what you accept).

Widening the annotations is the easy part; the work is that it surfaces constructor bodies that access `.ndim`/`.shape`/`.reshape`/etc. on a parameter before it is coerced to an array (those attributes are not defined on the scalar members of `ArrayLike`). So the follow-up is "widen the input annotations **and** coerce the parameters at the top of each affected constructor", done together. It is kept separate here to keep this PR focused on outputs and to give the input change its own review/CI cycle.

(`CAR.adj_matrix` is a deliberate exception — it also accepts a `scipy.sparse` matrix, so it needs a union rather than plain `ArrayLike`.)

## Testing

`test/test_distributions.py` passes in full, including the new `test_outputs_are_arrays` and `test_aux_fields_are_not_jax_arrays`. The eager-only argument/sample validation behavior is covered by the existing validation tests, updated to reflect the new semantics.

### DIFF
--- a/numpyro/distributions/conjugate.py
+++ b/numpyro/distributions/conjugate.py
@@ -399,13 +399,13 @@ class GammaPoisson(Distribution):
         )
 
     @property
-    def mean(self) -> ArrayLike:
+    def mean(self) -> Array:
         r"""If :math:`X \sim \mathrm{GammaPoisson}(\alpha, \lambda)`, then the mean is:
 
         .. math::
             \mathbb{E}[X] = \frac{\alpha}{\lambda}
         """
-        return self.concentration / self.rate
+        return jnp.asarray(self.concentration / self.rate)
 
     @property
     def variance(self) -> Array:
@@ -416,7 +416,7 @@ class GammaPoisson(Distribution):
         """
         return self.concentration / jnp.square(self.rate) * (1 + self.rate)
 
-    def cdf(self, value: ArrayLike) -> ArrayLike:
+    def cdf(self, value: ArrayLike) -> Array:
         r"""If :math:`X \sim \mathrm{GammaPoisson}(\alpha, \lambda)`, then the cumulative
         distribution function is:
 
--- a/numpyro/distributions/conjugate.py
+++ b/numpyro/distributions/conjugate.py
@@ -5,7 +5,7 @@
 from typing import Optional
 
 import jax
-from jax import lax, nn, random
+from jax import Array, lax, nn, random
 import jax.numpy as jnp
 from jax.scipy.special import betainc, betaln, gammaln
 from jax.typing import ArrayLike
@@ -72,7 +72,7 @@ class BetaBinomial(Distribution):
         self._beta = Beta(concentration1, concentration0)
         super(BetaBinomial, self).__init__(batch_shape, validate_args=validate_args)
 
-    def sample(self, key: jax.Array, sample_shape: tuple[int, ...] = ()) -> ArrayLike:
+    def sample(self, key: jax.Array, sample_shape: tuple[int, ...] = ()) -> Array:
         assert is_prng_key(key)
         key_beta, key_binom = random.split(key)
         probs = self._beta.sample(key_beta, sample_shape)
@@ -81,7 +81,7 @@ class BetaBinomial(Distribution):
         )
 
     @validate_sample
-    def log_prob(self, value: ArrayLike) -> ArrayLike:
+    def log_prob(self, value: ArrayLike) -> Array:
         return (
             -_log_beta_1(self.total_count - value + 1, value)
             + betaln(
@@ -92,11 +92,11 @@ class BetaBinomial(Distribution):
         )
 
     @property
-    def mean(self) -> ArrayLike:
+    def mean(self) -> Array:
         return self._beta.mean * self.total_count
 
     @property
-    def variance(self) -> ArrayLike:
+    def variance(self) -> Array:
         return (
             self._beta.variance
             * self.total_count
@@ -161,7 +161,7 @@ class BetaNegativeBinomial(Distribution):
             batch_shape, validate_args=validate_args
         )
 
-    def sample(self, key: jax.Array, sample_shape: tuple[int, ...] = ()) -> ArrayLike:
+    def sample(self, key: jax.Array, sample_shape: tuple[int, ...] = ()) -> Array:
         r"""If :math:`X \sim \mathrm{BetaNegativeBinomial}(\alpha, \beta, n)`, then the sampling
         procedure is:
 
@@ -182,7 +182,7 @@ class BetaNegativeBinomial(Distribution):
         return NegativeBinomialProbs(total_count=self.n, probs=probs).sample(key_nb)
 
     @validate_sample
-    def log_prob(self, value: ArrayLike) -> ArrayLike:
+    def log_prob(self, value: ArrayLike) -> Array:
         r"""If :math:`X \sim \mathrm{BetaNegativeBinomial}(\alpha, \beta, n)`, then the log
         probability mass function is:
 
@@ -201,7 +201,7 @@ class BetaNegativeBinomial(Distribution):
         )
 
     @property
-    def mean(self) -> ArrayLike:
+    def mean(self) -> Array:
         r"""If :math:`X \sim \mathrm{BetaNegativeBinomial}(\alpha, \beta, n)` and
         :math:`\beta > 1`, then the mean is:
 
@@ -217,7 +217,7 @@ class BetaNegativeBinomial(Distribution):
         )
 
     @property
-    def variance(self) -> ArrayLike:
+    def variance(self) -> Array:
         r"""If :math:`X \sim \mathrm{BetaNegativeBinomial}(\alpha, \beta, n)` and
         :math:`\beta > 2`, then the variance is:
 
@@ -289,7 +289,7 @@ class DirichletMultinomial(Distribution):
             validate_args=validate_args,
         )
 
-    def sample(self, key: jax.Array, sample_shape: tuple[int, ...] = ()) -> ArrayLike:
+    def sample(self, key: jax.Array, sample_shape: tuple[int, ...] = ()) -> Array:
         assert is_prng_key(key)
         key_dirichlet, key_multinom = random.split(key)
         probs = self._dirichlet.sample(key_dirichlet, sample_shape)
@@ -300,18 +300,18 @@ class DirichletMultinomial(Distribution):
         ).sample(key_multinom)
 
     @validate_sample
-    def log_prob(self, value: ArrayLike) -> ArrayLike:
+    def log_prob(self, value: ArrayLike) -> Array:
         alpha = self.concentration
         return _log_beta_1(alpha.sum(-1), value.sum(-1)) - _log_beta_1(
             alpha, value
         ).sum(-1)
 
     @property
-    def mean(self) -> ArrayLike:
+    def mean(self) -> Array:
         return self._dirichlet.mean * jnp.expand_dims(self.total_count, -1)
 
     @property
-    def variance(self) -> ArrayLike:
+    def variance(self) -> Array:
         n = jnp.expand_dims(self.total_count, -1)
         alpha = self.concentration
         alpha_sum = self.concentration.sum(-1, keepdims=True)
@@ -362,7 +362,7 @@ class GammaPoisson(Distribution):
             self._gamma.batch_shape, validate_args=validate_args
         )
 
-    def sample(self, key: jax.Array, sample_shape: tuple[int, ...] = ()) -> ArrayLike:
+    def sample(self, key: jax.Array, sample_shape: tuple[int, ...] = ()) -> Array:
         r"""If :math:`X \sim \mathrm{GammaPoisson}(\alpha, \lambda)`, then the sampling
         procedure is:
 
@@ -383,7 +383,7 @@ class GammaPoisson(Distribution):
         return Poisson(rate).sample(key_poisson)
 
     @validate_sample
-    def log_prob(self, value: ArrayLike) -> ArrayLike:
+    def log_prob(self, value: ArrayLike) -> Array:
         r"""If :math:`X \sim \mathrm{GammaPoisson}(\alpha, \lambda)`, then the
         probability mass function is:
 
@@ -408,7 +408,7 @@ class GammaPoisson(Distribution):
         return self.concentration / self.rate
 
     @property
-    def variance(self) -> ArrayLike:
+    def variance(self) -> Array:
         r"""If :math:`X \sim \mathrm{GammaPoisson}(\alpha, \lambda)`, then the variance is:
 
         .. math::
@@ -514,7 +514,7 @@ class NegativeBinomialLogits(GammaPoisson):
         super().__init__(concentration, rate, validate_args=validate_args)
 
     @validate_sample
-    def log_prob(self, value: ArrayLike) -> ArrayLike:
+    def log_prob(self, value: ArrayLike) -> Array:
         r"""If :math:`X \sim \mathrm{NegativeBinomial}(r, \mathrm{logits}(p))`, then the log
         probability mass function is:
 
--- a/numpyro/distributions/conjugate.py
+++ b/numpyro/distributions/conjugate.py
@@ -72,7 +72,10 @@ class BetaBinomial(Distribution):
         self._beta = Beta(concentration1, concentration0)
         super(BetaBinomial, self).__init__(batch_shape, validate_args=validate_args)
 
-    def sample(self, key: jax.Array, sample_shape: tuple[int, ...] = ()) -> Array:
+    def sample(
+        self, key: Optional[jax.Array], sample_shape: tuple[int, ...] = ()
+    ) -> Array:
+        assert key is not None
         assert is_prng_key(key)
         key_beta, key_binom = random.split(key)
         probs = self._beta.sample(key_beta, sample_shape)
@@ -161,7 +164,9 @@ class BetaNegativeBinomial(Distribution):
             batch_shape, validate_args=validate_args
         )
 
-    def sample(self, key: jax.Array, sample_shape: tuple[int, ...] = ()) -> Array:
+    def sample(
+        self, key: Optional[jax.Array], sample_shape: tuple[int, ...] = ()
+    ) -> Array:
         r"""If :math:`X \sim \mathrm{BetaNegativeBinomial}(\alpha, \beta, n)`, then the sampling
         procedure is:
 
@@ -176,6 +181,7 @@ class BetaNegativeBinomial(Distribution):
         :class:`~numpyro.distributions.discrete.NegativeBinomialProbs` to generate samples
         from the Negative Binomial distribution.
         """
+        assert key is not None
         assert is_prng_key(key)
         key_beta, key_nb = random.split(key)
         probs = self._beta.sample(key_beta, sample_shape)
@@ -289,7 +295,10 @@ class DirichletMultinomial(Distribution):
             validate_args=validate_args,
         )
 
-    def sample(self, key: jax.Array, sample_shape: tuple[int, ...] = ()) -> Array:
+    def sample(
+        self, key: Optional[jax.Array], sample_shape: tuple[int, ...] = ()
+    ) -> Array:
+        assert key is not None
         assert is_prng_key(key)
         key_dirichlet, key_multinom = random.split(key)
         probs = self._dirichlet.sample(key_dirichlet, sample_shape)
@@ -302,9 +311,9 @@ class DirichletMultinomial(Distribution):
     @validate_sample
     def log_prob(self, value: ArrayLike) -> Array:
         alpha = self.concentration
-        return _log_beta_1(alpha.sum(-1), value.sum(-1)) - _log_beta_1(
-            alpha, value
-        ).sum(-1)
+        return _log_beta_1(jnp.sum(alpha, -1), jnp.sum(value, -1)) - jnp.sum(
+            _log_beta_1(alpha, value), -1
+        )
 
     @property
     def mean(self) -> Array:
@@ -322,9 +331,9 @@ class DirichletMultinomial(Distribution):
     def support(self) -> Constraint:
         return constraints.multinomial(self.total_count)
 
-    @staticmethod
+    @classmethod
     def infer_shapes(
-        concentration: ArrayLike, total_count=()
+        cls, concentration: tuple[int, ...], total_count: tuple[int, ...] = ()
     ) -> tuple[tuple[int, ...], tuple[int, ...]]:
         batch_shape = lax.broadcast_shapes(concentration[:-1], total_count)
         event_shape = concentration[-1:]
@@ -362,7 +371,9 @@ class GammaPoisson(Distribution):
             self._gamma.batch_shape, validate_args=validate_args
         )
 
-    def sample(self, key: jax.Array, sample_shape: tuple[int, ...] = ()) -> Array:
+    def sample(
+        self, key: Optional[jax.Array], sample_shape: tuple[int, ...] = ()
+    ) -> Array:
         r"""If :math:`X \sim \mathrm{GammaPoisson}(\alpha, \lambda)`, then the sampling
         procedure is:
 
@@ -377,6 +388,7 @@ class GammaPoisson(Distribution):
         :class:`~numpyro.distributions.continuous.Poisson` to generate samples from the
         Poisson distribution.
         """
+        assert key is not None
         assert is_prng_key(key)
         key_gamma, key_poisson = random.split(key)
         rate = self._gamma.sample(key_gamma, sample_shape)
@@ -510,7 +522,7 @@ class NegativeBinomialLogits(GammaPoisson):
     ):
         self.total_count, self.logits = promote_shapes(total_count, logits)
         concentration = total_count
-        rate = jnp.exp(-logits)
+        rate = jnp.exp(jnp.negative(logits))
         super().__init__(concentration, rate, validate_args=validate_args)
 
     @validate_sample
--- a/numpyro/distributions/constraints.py
+++ b/numpyro/distributions/constraints.py
@@ -70,6 +70,7 @@ from typing import ClassVar, Generic, Optional, cast
 import numpy as np
 
 import jax
+from jax import Array
 import jax.numpy as jnp
 from jax.tree_util import register_pytree_node
 from jax.typing import ArrayLike
@@ -96,13 +97,13 @@ class Constraint(Generic[NumLikeT]):
     def tree_flatten(self):
         raise NotImplementedError
 
-    def __call__(self, x: NumLikeT) -> ArrayLike:
+    def __call__(self, x: NumLikeT) -> Array:
         raise NotImplementedError
 
     def __repr__(self) -> str:
         return self.__class__.__name__[1:] + "()"
 
-    def check(self, value: NumLikeT) -> ArrayLike:
+    def check(self, value: NumLikeT) -> Array:
         """
         Returns a byte tensor of `sample_shape + batch_shape` indicating
         whether each event in value satisfies this constraint.
@@ -115,7 +116,7 @@ class Constraint(Generic[NumLikeT]):
         """
         raise NotImplementedError
 
-    def eq(self, other: object, static: bool = False) -> ArrayLike:
+    def eq(self, other: object, static: bool = False) -> bool:
         return self is other
 
     def __eq__(self, other: object) -> bool:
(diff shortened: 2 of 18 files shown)


juanitorduz:
Hi @tillahoffmann ! Coult we try adding these modified modules in https://github.com/pyro-ppl/numpyro/blob/master/pyproject.toml#L100 so that mypy runs the type checks? If we get many errors, we can tackle them in smaller PRs 🤞

tillahoffmann:
Good idea, and yeah it'd need to be done carefully — mypy is a hard gate in CI (`make lint` runs `mypy numpyro` and fails the build on errors), not just informational, I think.

I tried adding the modified modules to the `ignore_errors = false` list locally: it surfaces ~224 errors, mostly pre-existing rather than from this PR — `[override]` signature mismatches (59), `.shape`/`.ndim` on `ArrayLike` unions (55), `[arg-type]`/`[return-value]` (66), etc.

So I think the path you're describing works best in the other order: drive each module's errors to 0 while it's still un-enforced, then flip it into the enforced list in that same PR (rather than adding it first, which would immediately red CI). 

Happy to start chipping away at the modules in follow-ups — I've already got input-widening (`ArrayLike` params) started, which covers a good chunk of the `union-attr` ones.

juanitorduz:
Great! @tillahoffmann ! Feel free to open issues and we can all help with this innitiative :)

tillahoffmann:
Sounds good, how shall we proceed with this one?

juanitorduz:
> Sounds good, how shall we proceed with this one?

I would merge it, and then we could build on top of this and fix edge cases with mypy down the line :)

juanitorduz (approved):

tillahoffmann:
Sounds like a plan. I'll start tackling some of those after we merge this one—I think there might be a lot of merge conflicts otherwise.

tillahoffmann:
@juanitorduz, @fehiepsi, @Qazalbash, are you happy for me to merge this?

fehiepsi:
Sorry I missed the scope of the CL. I'm not sure why we want to make this change. The jax best practice seems to apply for jax function. Those distribution methods could be flexible I guess. My main concern is the performance. It was intentional to have separate treatment for jax arrays and numpy arrays here and there.

I still don't understand why we need to enforce jax array outputs though. Could you elaborate?

tillahoffmann:
Re-reading the diff, this PR conflated two separate claims, and your pushback applies to one — the other I think is worth keeping.

**Parameters coerced to `jax.Array` (storage) — dropping this.** You're right: it costs on python-scalar construction and overrides the deliberate numpy/jax separation.

**Outputs annotated `Array` (return type) — the part worth keeping, and the answer to "why."** It's for *consumers* writing typed code. The methods return `ArrayLike`, which includes `bool`/`int`/`float`/`complex` — not indexable, no `.shape`, and `complex` has no ordering. Under pyright, ordinary usage errors:

```python
d = dist.Normal(jnp.zeros(5), 1.0)
d.mean[3]            # __getitem__ not defined on int/float/complex
d.variance.shape     # shape unknown on scalar members
d.mean < 0           # "<" not supported for complex
```

All clean once the return is `Array`. The return is the API boundary, and `ArrayLike` is an unusable promise to a caller — internals can stay as flexible as we like; only the boundary is normalized.

**Revised PR:** return-position `ArrayLike → Array`, plus `jnp.asarray` at the ~30 returns that hand back a stored parameter directly (e.g. `Poisson.mean` is `self.rate`). No constructor / `promote_shapes` / validation changes — the diff is signatures, those returns, and one test. Construction perf returns to master.

Does honest `Array` outputs with internals untouched work for you?

fehiepsi:
The type error in your example looks expected to me. The code assumes that the output is jax Array, which is unexpected (I understand that it might be expected to some users). My point is we need to use Array only for special distributions (the places that we know that they should be jax arrays). and keep ArrayLike for the rest.

tillahoffmann:
Makes sense — and I think we're almost there with narrowing outputs to `Array` anyway.

My main reason for wanting `Array` here is that `ArrayLike` isn't very meaningful as an *output* type. It's `Array | np.ndarray | bool | int | float | …`, so on `Normal.sample()` it advertises that the caller might get back a bare `float`, or a `bool` — for a method that always returns a jax array. It's not *wrong* (an array is an `ArrayLike`), but it does not constrain things and points the caller at cases that never happen. `ArrayLike` is the right type for an *input*, where the method really does accept any of those; on an output it just promises less than we actually deliver.

In practice we're most of the way there already. I parametrized the output-type check per method across the suite:

| method | already returns `jax.Array` |
|---|---|
| `sample` | 97 / 98 |
| `log_prob` | 95 / 95 |
| `entropy` | 30 / 30 |
| `mean` | 62 / 81 |
| `variance` | 58 / 69 |

`sample`/`log_prob`/`entropy` are already `jax.Array` because they go through `random.*` / `jnp` / `jax.scipy.special` (except `Delta.sample` which returns the input unchanged). Likewise, ~80% of the moments are already `Array`s.

The ones that *don't* are implementation-dependent rather than meaningful. `Gamma` is the clearest case:

```python
@property
def mean(self):      return self.concentration / self.rate  # tracks input type
@property
def variance(self):  return self.concentration / jnp.power(self.rate, 2)  # jnp.power -> always Array
```

`variance` is always an `Array` and `mean` isn't because one line calls `jnp` and the other doesn't — same distribution, same params. From a consumer's side that's surprising: a type checker would reject `dist.mean[0]` but accept `dist.variance[0]` if we annotated the return type with what's actually returned. One `jnp.asarray` at those ~31 return sites is a smaller cost than that inconsistency, I think.

This would be **return-only** (in contrast to my earlier, overly broad attempt). Stored params and the numpy/jax separation stay exactly as they are (`Poisson(2.0).rate` is still `2.0`), and it's only on cold moment properties, not `sample`/`log_prob`. So no hot-path or construction cost.

What do you think?

Qazalbash (approved):
I agree with this reasoning. `ArrayLike` makes sense for inputs because it reflects what callers are allowed to pass in, but for outputs it ends up advertising possibilities that never actually occur in practice. Returning `jax.Array` where we consistently produce `jax.Array` is a more useful and informative contract for users and type checkers.

The `Gamma.mean` vs `Gamma.variance` example illustrates the issue well; differences in output type arising purely from implementation details rather than semantics are surprising from a consumer perspective.

For cases where the output type currently follows from implementation details rather than intent, another option could be to prefer `jax.numpy` operators such as `jnp.add`, `jnp.divide`, `jnp.multiply`, etc., instead of Python operators. For example, `Gamma.mean` could use `jnp.divide(self.concentration, self.rate)`, which would naturally align its behavior with `Gamma.variance` without requiring an explicit `jnp.asarray` at the return site. That said, I don't have a strong preference between the two approaches as long as the public API consistently returns `jax.Array`.

fehiepsi:
I understand that if users expect Array as the output, then type errors will happen. I don't expect Array to be the output though. It's unnecessary to support such behavior. Users should fix their code since we clearly annotate the output is ArrayLike.

As mentioned above, we can use Array for places that we know output is an Array. Using patterns like jnp.divide in the implementation will give Array output but in my opinion, using non-array output at those places is better. It is clearer to me that Normal(0,1) will have mean 0, not float 0 or Array 0.

numpyro/distributions/conjugate.py
```diff
@@ -278,7 +284,9 @@ def __init__(
             jnp.shape(concentration)[:-1], jnp.shape(total_count)
         )
         concentration_shape = batch_shape + jnp.shape(concentration)[-1:]
-        (self.concentration,) = promote_shapes(concentration, shape=concentration_shape)
+        (self.concentration,) = promote_shapes(
+            concentration, shape=concentration_shape, promote_array=True
+        )
```
fehiepsi: Why we need this promote?

numpyro/distributions/conjugate.py
```diff
@@ -356,13 +364,15 @@ def __init__(
         *,
         validate_args: Optional[bool] = None,
     ):
-        self.concentration, self.rate = promote_shapes(concentration, rate)
+        self.concentration, self.rate = promote_shapes(
+            concentration, rate, promote_array=True
+        )
```
fehiepsi: Like above, we dont need to promote.

numpyro/distributions/directional.py
```diff
@@ -590,13 +598,13 @@ def model():
 
     def __init__(self, concentration: Array, *, validate_args: Optional[bool] = None):
         assert jnp.ndim(concentration) >= 1
-        self.concentration = concentration
+        self.concentration = jnp.asarray(concentration)
```
fehiepsi: We dont need this cast I think.

fehiepsi:
It would be fine to ensure mean and variance return jax Array since we already assume the semantic that they should have the expected shape. Those changes won't affect performance I guess. We might want to reconsider changes at promote logic though.

tillahoffmann:
Re-scoped, this PR is, for now, just type work — **return annotations only**, no constructor / `promote_shapes` / validation changes. `Poisson(2.0).rate` is still `2.0` and construction perf is back to master; the diff is signatures plus one parametrized test.

Concretely it narrows `sample` / `log_prob` / `entropy` and the ~80% of `mean` / `variance` that already return `jax.Array` to an `Array` return type. The handful of `mean` / `variance` that currently pass a python/numpy value straight through (e.g. `Gamma.mean`) I've left as `ArrayLike` for now, so every annotation is honest as it stands.

Sounds like we have consensus that those `mean` / `variance` returning `Array` is fine? I'm happy to promote those **return** types in this PR too — a one-line `jnp.asarray` at each of the ~30 sites — or keep this one narrow and do it as a follow-up.

I still think the promotion is the right contract, and the cost is negligible: I checked it directly. Under `jit` the `jnp.asarray` wrap **constant-folds**, with the jaxpr *and* optimized HLO byte-identical (no extra op), so inference sees zero difference. The only cost is a single scalar dispatch in eager (~tens of µs), never `sample` / `log_prob` or a per-step path. It's an `Array` under `jit` either way; promotion to `Array` would just make the eager return match what jitted code already produces.

@Qazalbash thanks for pointing that out re `jnp.asarray(a / b)` vs `jnp.divide(a, b)`! Maybe we can have a look at that trade-off if/when we promote the moments?

fehiepsi:
Sorry for the late response.

> Sounds like we have consensus that those mean / variance returning Array is fine? I'm happy to promote those return types in this PR too — a one-line jnp.asarray at each of the ~30 sites

Looks reasonable to me. Could you make the change in this CL?

Re performance: I'm not sure why but I'm seeing numpyro CI is slower after each release. I guess the main reason is at those unnecessary `jit` compiling operators outside of a `jit` context. Under jit, things might be fine though. Thanks for doing the benchmarks!

tillahoffmann:
This has been updated to have mean/variance return `jax.Array` and the type annotations have been narrowed. I have also narrowed type annotations all other sites that already satisfied the constraint (`sample`, `log_prob`, `entropy`, etc.).

Regarding performance, is there a benchmark we could use to assess, @fehiepsi? Happy to have a look what we can eke out.

fehiepsi:
No worries about benchmarks. I just wanted to say that jit operators are expensive for the first step. Our CI has been much slower than before, likely because of that. I didn't keep track of the regression though so I can't give pointers. It's fine to keep the PR in the current scope.

tillahoffmann:
Sounds good. I think this one is ready.

numpyro/distributions/constraints.py
```diff
@@ -302,9 +303,9 @@ class _GreaterThan(Constraint[NumLike]):
     def __init__(self, lower_bound: NumLike) -> None:
         self.lower_bound = lower_bound
 
-    def __call__(self, x: NumLike) -> ArrayLike:
+    def __call__(self, x: NumLike) -> Array:
         xp = jnp if isinstance(x, jax.Array) else np
-        return xp.greater(x, self.lower_bound)
+        return xp.greater(x, self.lower_bound)  # type: ignore
```
fehiepsi: It seems we should return ArrayLike here

numpyro/distributions/constraints.py
```diff
@@ -317,18 +318,18 @@ def feasible_like(self, prototype: NumLike) -> NumLike:
     def tree_flatten(self):
         return (self.lower_bound,), (("lower_bound",), dict())
 
-    def eq(self, other: object, static: bool = False) -> ArrayLike:
+    def eq(self, other: object, static: bool = False) -> bool:
```
fehiepsi: It is surprised to me that bool is the return type here. If you need bool, you might want to use __eq__ instead

fehiepsi (commented):
Could we not make strong commitment for the base classes or derived distributions like TransformedDistribution? I want to use np.array samples in some special distributions, in particular for Delta or string-valued distribution (wrapped about numpy array interface). This applies at least for sample, log_prob, and sample_with_intermediates. It would be unfortunate that we no longer support numpy array values in numpyro.

numpyro/distributions/constraints.py
```diff
@@ -699,8 +700,8 @@ def feasible_like(self, prototype: NumLike) -> NumLike:
 class _OrderedVector(_SingletonConstraint[NonScalarArray]):
     event_dim = 1
 
-    def __call__(self, x: NonScalarArray) -> ArrayLike:
-        return (x[..., 1:] > x[..., :-1]).all(axis=-1)
+    def __call__(self, x: NonScalarArray) -> Array:
+        return (x[..., 1:] > x[..., :-1]).all(axis=-1)  # type: ignore
```
fehiepsi: Let's return the correct type instead of ignore here?

numpyro/distributions/constraints.py
```diff
@@ -783,9 +784,9 @@ def feasible_like(self, prototype: NumLike) -> NumLike:
 
 
 class _Real(_SingletonConstraint[NumLike]):
-    def __call__(self, x: NumLike) -> ArrayLike:
+    def __call__(self, x: NumLike) -> Array:
         # XXX: consider to relax this condition to [-inf, inf] interval
-        return (x == x) & (x != float("inf")) & (x != float("-inf"))
+        return (x == x) & (x != float("inf")) & (x != float("-inf"))  # type: ignore
```
fehiepsi: Maybe keep ArrayLike?

numpyro/distributions/constraints.py
```diff
@@ -794,9 +795,9 @@ def feasible_like(self, prototype: NumLike) -> NumLike:
 class _Simplex(_SingletonConstraint[NonScalarArray]):
     event_dim = 1
 
-    def __call__(self, x: NonScalarArray) -> ArrayLike:
+    def __call__(self, x: NonScalarArray) -> Array:
         x_sum = x.sum(axis=-1)
-        return (x >= 0).all(axis=-1) & (x_sum < 1 + 1e-6) & (x_sum > 1 - 1e-6)
+        return (x >= 0).all(axis=-1) & (x_sum < 1 + 1e-6) & (x_sum > 1 - 1e-6)  # type: ignore
```
fehiepsi: Like above

numpyro/distributions/constraints.py
```diff
@@ -855,6 +856,7 @@ def __call__(self, x: NonScalarArray) -> ArrayLike:
         zerosum_true = True
         for dim in range(-self.event_dim, 0):
             zerosum_true = zerosum_true & xp.allclose(x.sum(dim), 0, atol=tol)
+        # FIXME: shape must match batch shape of `x`, not be a literal boolean.
```
fehiepsi: It seems to me that we need to sum over the event dimensions first, then use isclose instead of all_close.

juanitorduz:
Hey folks! Shall we aim to bring this one to the finish line and cut a release (maybe also wait for https://github.com/pyro-ppl/numpyro/pull/2224?) :)

Happy to help if you need some support to bring this one to the end :)

juanitorduz:
@fehiepsi @tillahoffmann Do you mind if I help push this PR until the end :) ? (and then ideally, cut a release?)

fehiepsi:
Sure, the main concern was to relax typing for base classes: delta, distribution, transformed distribution, etc. so that we can apply handlers for special types. It's fine to allow numpy ndarray there since ndarray works with generic types. We can add a couple of tests to verify if handlers work well for those distributions. One example is to create a fake categorical distribution with 3 values: "A", "B", "C" and using log density or trace handlers to check if things work. I'm not sure if pytype or other typing systems work in test though :(

numpyro/distributions/conjugate.py
juanitorduz: Good call, thanks! The promote_shapes changes were dropped when the PR was re-scoped to return annotations only, so this is gone now.

numpyro/distributions/conjugate.py
juanitorduz: Same here, dropped together with the other promote changes. Thanks!

numpyro/distributions/directional.py
juanitorduz: Agreed, the cast is gone (constructors are untouched in the current version). Thanks!

numpyro/distributions/constraints.py
juanitorduz: Done, this is back to ArrayLike. Thanks!

numpyro/distributions/constraints.py
juanitorduz: Fair point! Reverted to ArrayLike, and no eq returns bool anywhere now. Thanks!

numpyro/distributions/constraints.py
juanitorduz: Done, back to ArrayLike with no type: ignore. The constraint annotations are no longer touched by this PR. Thanks!

numpyro/distributions/constraints.py
juanitorduz: Yes, kept as ArrayLike. Thanks!

numpyro/distributions/constraints.py
juanitorduz: Same, kept as ArrayLike. Thanks!

numpyro/distributions/constraints.py
juanitorduz: Great suggestion, thanks! Done in ce14ff33: we now sum over each event axis and reduce isclose over the event axes, so the mask has the batch shape of the input. Added test cases for event_dim=1 and 2; the tolerance itself is unchanged.

juanitorduz:
Hi @fehiepsi, thanks a lot for the review! I merged `master` and pushed changes addressing your comments:

- Base/wrapper classes (`Distribution`, `ExpandedDistribution`, `Independent`, `MaskedDistribution`, `TransformedDistribution`, `Delta`) keep `ArrayLike` on `sample` / `rsample` / `sample_with_intermediates` / `log_prob`, and `Delta.sample` returns the stored value untouched, so numpy-backed or string-valued distributions keep working through handlers.
- Added `test_handlers_with_string_valued_distribution`: a fake categorical over `"A"` / `"B"` / `"C"` through `seed`, `trace`, `plate`, and `log_density`, plus a string-valued `Delta`.
- `constraints.py`, `transforms.py`, and `flows.py` are no longer touched, except for the `zero_sum` fix you suggested (ce14ff33).

One small heads-up: since `Beta.mean` now returns a jax array, `BetaProportion(validate_args=True)` no longer raises under `jit` for an out-of-range `mean` (eager validation is unchanged). Happy to revisit if that is a concern.

Would you mind taking another look when you have a chance? :)

juanitorduz:
In the last 4 commits, I ensured ty is actually checking the distributions submodule types, and it seems to be working 🎉 (it was important to add them in the `pyproject.toml`). I just added minor changes to make it work 🤗

Qazalbash (approved):

tillahoffmann:
Thank you, @juanitorduz!

fehiepsi:
Thanks all!!

PR #2207, opened 2026-06-13
Migrate from MyPy to Ty

juanitorduz:
Follow-up of https://github.com/pyro-ppl/numpyro/pull/2111 now that ty is much more mature (see https://github.com/astral-sh/ty/issues/269)


Qazalbash (approved):
LGTM!

Is it okay to remove `ty.toml`?

juanitorduz:
I moved the ty config into pyproject.toml :)

Qazalbash:
Oh, I see 😄!

juanitorduz:
Is it ok if I merge this one?

fehiepsi:
I'm still unsure whether this affects users who don't use ty and why we drop mypy support.

ty might be fast but we should prioritize the standard solutions.

Please feel free to merge if you find it suitable.

juanitorduz:
> I'm still unsure whether this affects users who don't use ty and why we drop mypy support.
> 
> ty might be fast but we should prioritize the standard solutions.
> 
> Please feel free to merge if you find it suitable.

This is a fair comment. My "bet" (AKA opinion) is that ty will dominate the same way ruff dominates over black now.  We can keep this one open (as a draft) and merge once we feel it's the right time (if any ;) )

fehiepsi:
I trust your opinion. :)

fehiepsi:
It seems that this PR is causing CI failing at master.

juanitorduz:
> It seems that this PR is causing CI failing at master.

solved in https://github.com/pyro-ppl/numpyro/pull/2209 🙏

PR #2208, opened 2026-06-24
Add HaarTransform and DiscreteCosineTransform

juanitorduz:
Add HaarTransform and DiscreteCosineTransform by bringing them from Pyro. 

Motivation: https://github.com/juanitorduz/numpyro_forecast/issues/12

# Haar / DCT time reparameterization: motivation and method

## 1. The problem: correlated time latents wreck inference geometry

Sequential models place a latent value at every time step. A typical prior is a random walk / local-level drift:

$$x_1 \sim \mathcal{N}(0,\sigma), \qquad x_t = x_{t-1} + \varepsilon_t, \quad \varepsilon_t \sim \mathcal{N}(0,\sigma).$$

Neighboring latents are almost equal, so the joint posterior over $(x_1,\dots,x_T)$ has strong off-diagonal correlation (a banded precision matrix for AR / random-walk processes). Geometrically the posterior is a long, thin, rotated ellipse rather than a round ball.

That geometry is poison for the two standard inference tools:

- **HMC / NUTS with a diagonal mass matrix** assumes coordinates are roughly independent with per-axis scales. A correlated, rotated posterior forces tiny step sizes, so the chain mixes slowly.
- **Mean-field (diagonal) SVI guides** cannot represent the correlations, so they fit badly.

## 2. The fix: change coordinates to a space where the posterior is round

The general trick (the "non-centered" idea) is: do not sample the awkward $x$ directly. Sample a new variable $z$ in a nicer space and define $x = f(z)$ deterministically. If $f$ is chosen so that $z$ has an approximately diagonal / isotropic posterior, then the diagonal mass matrix and the mean-field guide work well. Inference runs on $z$; $x$ is reconstructed for the likelihood.

For time series we want a **linear** map that approximately diagonalizes the covariance of the trajectory. There is a classic answer: for stationary processes the covariance is approximately Toeplitz / circulant, and circulant matrices are diagonalized exactly by the Fourier basis. So transforming to the frequency domain decorrelates the time points. Haar and DCT are two concrete, cheap, real-valued ways to do this.

## 3. The Haar transform (multi-resolution averaging / differencing)

Take adjacent pairs and form a scaled average and difference:

$$s = \tfrac{1}{\sqrt{2}}(x_1 + x_2), \qquad d = \tfrac{1}{\sqrt{2}}(x_1 - x_2).$$

The $1/\sqrt{2}$ keeps it energy-preserving (orthonormal). Then you recurse on the averages only, halving the length each level. The result is one overall coarse coefficient (the "DC" / mean), plus detail (difference) coefficients at every scale, from coarse to fine. The output ordering is `[dc, hi_coarsest, ..., hi_finest]`.

Why it helps: a slowly varying or random-walk trajectory has tiny local differences, so the high-frequency detail coefficients have low, roughly equal variance and are nearly uncorrelated, while the few coarse coefficients carry the long-range structure. The strong nearest-neighbor coupling in time becomes a near-diagonal covariance across scales. Haar is the natural choice when correlation is "blocky" / multi-scale.

## 4. The Discrete Cosine Transform (frequency modes)

The DCT-II writes the signal as a sum of cosines of increasing frequency:

$$X_k \propto \sum_{n} x_n \cos\!\Big(\tfrac{\pi k (n + \tfrac{1}{2})}{N}\Big),$$

normalized to be orthonormal. It is the real-valued cousin of the DFT with even-symmetric boundary handling, and that boundary handling is exactly why it is (asymptotically) the optimal decorrelating basis for AR(1) / first-order Markov processes, i.e. it is close to the Karhunen-Loève (PCA) basis of those covariances. Coefficient $k=0$ is the overall mean; higher $k$ capture higher-frequency wiggles. Because the modes of a stationary process are approximately independent, the DCT coefficients have an approximately diagonal posterior. DCT is the natural choice for smooth, stationary-like dynamics.

## 5. The `smooth` knob (DCT only): matching the "color" of the noise

A plain DCT decorrelates, but the scales of the modes can still differ a lot (low frequencies carry much more variance in a random walk). The `smooth` parameter rescales each frequency component by $\text{freq}^{\text{smooth}}$ to flatten that out:

- `smooth=0`: no rescaling (white noise to white noise).
- `smooth=1`: turns Brownian-motion / random-walk priors (power spectrum $\propto 1/f^2$) into white noise in the reparam space, so $z$ is isotropic and HMC sees a round posterior.
- `smooth=2`: for smoother, twice-integrated processes.

This is the "colors of noise" idea: pick `smooth` to match how rough the latent process is.

## 6. Why this is "free": unit Jacobian

Haar and DCT are orthonormal (rotations / reflections), so $|\det J| = 1$, and the `smooth` weights are normalized so their geometric mean is one (which is why the weight is divided by `exp(mean(log(w)))`). A change of variables with unit Jacobian adds no log-determinant term: the prior and posterior densities are exactly preserved. So we are not changing the model at all, only the coordinate system the sampler explores. This is what `UnitJacobianReparam` encodes, and it is why applying these reparams cannot bias the results, only improve convergence.

## 7. How it lands in the code

When you wrap a site with `HaarReparam()` / `DiscreteCosineReparam()`:

1. The handler replaces the latent `x` with an auxiliary site `x_haar` / `x_dct` sampled from the transformed prior (the pushforward through the transform). This auxiliary variable lives in the decorrelated frequency / wavelet domain with a near-diagonal posterior.
2. It deterministically reconstructs `x = inverse_transform(x_haar)` for the likelihood.
3. Inference (NUTS, SVI) operates on the well-conditioned `x_haar`; you read out `x` as a `deterministic` site.

For positive-support latents (e.g. `LogNormal`) it first maps to unconstrained space via `biject_to(support).inv` (a log), then applies Haar / DCT there. That is the `ComposeTransform([biject_to(fn.support).inv, self.transform])` step.

## Summary

Time latents are correlated because the prior couples neighbors. Haar / DCT rotate the trajectory into a frequency / multi-scale basis where those couplings approximately vanish, giving a round posterior that diagonal mass matrices and mean-field guides can handle. Because the rotation is volume-preserving it leaves the model exactly unchanged.

### DIFF
--- a/docs/source/distributions.rst
+++ b/docs/source/distributions.rst
@@ -1064,9 +1064,25 @@ CorrMatrixCholeskyTransform
     :show-inheritance:
     :member-order: bysource
 
+DiscreteCosineTransform
+^^^^^^^^^^^^^^^^^^^^^^^
+.. autoclass:: numpyro.distributions.transforms.DiscreteCosineTransform
+    :members:
+    :undoc-members:
+    :show-inheritance:
+    :member-order: bysource
+
 ExpTransform
 ^^^^^^^^^^^^
 .. autoclass:: numpyro.distributions.transforms.ExpTransform
+    :members:
+    :undoc-members:
+    :show-inheritance:
+    :member-order: bysource
+
+HaarTransform
+^^^^^^^^^^^^^
+.. autoclass:: numpyro.distributions.transforms.HaarTransform
     :members:
     :undoc-members:
     :show-inheritance:
--- a/docs/source/reparam.rst
+++ b/docs/source/reparam.rst
@@ -69,3 +69,30 @@ Explicit Reparameterization
     :show-inheritance:
     :member-order: bysource
     :special-members: __call__
+
+Unit Jacobian Transforms
+------------------------
+.. autoclass:: numpyro.infer.reparam.UnitJacobianReparam
+    :members:
+    :undoc-members:
+    :show-inheritance:
+    :member-order: bysource
+    :special-members: __call__
+
+Haar Transform
+--------------
+.. autoclass:: numpyro.infer.reparam.HaarReparam
+    :members:
+    :undoc-members:
+    :show-inheritance:
+    :member-order: bysource
+    :special-members: __call__
+
+Discrete Cosine Transform
+-------------------------
+.. autoclass:: numpyro.infer.reparam.DiscreteCosineReparam
+    :members:
+    :undoc-members:
+    :show-inheritance:
+    :member-order: bysource
+    :special-members: __call__
--- a/numpyro/distributions/transforms.py
+++ b/numpyro/distributions/transforms.py
@@ -2026,16 +2026,16 @@ class DiscreteCosineTransform(Transform[NonScalarArray]):
 
     def __call__(self, x: NonScalarArray) -> NonScalarArray:
         self.forward_shape(jnp.shape(x))
-        y = jax.scipy.fft.dct(x, type=2, norm="ortho", axis=self.dim)
+        y = jax.scipy.fft.dct(jnp.asarray(x), type=2, norm="ortho", axis=self.dim)
         if self.smooth:
-            y = y * self._weight(x.shape[self.dim], y.dtype)
+            y = y * self._weight(jnp.shape(x)[self.dim], y.dtype)
         return y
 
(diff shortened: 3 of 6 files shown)


fehiepsi:
Nice! Thanks for adding those transforms.

PR #2209, opened 2026-06-30
Better casting `eps`

juanitorduz:
Solves the CI issue on master https://github.com/pyro-ppl/numpyro/actions/runs/28414476194/job/84194313549


juanitorduz:
It is interesting we have now a similar problem in another project even with mypy 🙈  https://results.pre-commit.ci/run/github/467144767/1782810614.5N6TOoywQbmfLoVpActaiw

fehiepsi:
Thanks for the fix!

PR #2210, opened 2026-07-01
Rename `var` -> `variance` for truncated distributions and close support.

tillahoffmann:
- Rename `var` to `variance` for truncated distributions to match the base `Distribution` naming.
- Close the support for left/right truncated distribution to match the two-sided closed interval and other distributions with finite support, e.g., uniform, beta, etc.
- Include truncated distributions in mean/variance tests.
- Make mean/variance tests robust to `NotImplementedError`s similar to other tests.


juanitorduz (approved):

juanitorduz:
thanks @tillahoffmann :)

PR #2215, opened 2026-07-08
Miscellaneous bugfixes

esennesh:
* Nicer handling of empty mutables and nnx-modules in SVI (34fa2a10e6e1960682d61f85d62a41d0762e8566)
* Allow for `init_value` to be something other than init-to-mean (75bebfc82de5ccfee00f5149e9138b5585da4f0b)
* Fix import error (1f122aab1cb29afffc1ad2d0de94deb7a255019d)
* Handle the TODO on using sampling with intermediates in expanded and Independent distributions (5f7066a7eb0977257893991a075082164376e099)
* Include some type annotations and formatting in the `init_value` fix (22aff3f2b638946ccf7cb82ea64e1229e600ba5e)


juanitorduz:
Hi! Please sync with master as we fixed the CI failure https://github.com/pyro-ppl/numpyro/pull/2218 :)

esennesh:
> Hi! Please sync with master as we fixed the CI failure #2218 :)

All checks now pass 👍 .  Let me know how you'd like to conduct the remainder of the review process!

juanitorduz:
Thanks @esennesh . Most of it makes a lot of sense. 

Still, I wanna ask @fehiepsi for a review on the specific change 

> Handle the TODO on using sampling with intermediates in expanded and Independent distributions 

As I could be missing something.

Maybe @esennesh, it's worth adding tests that show how this is supposed to work for each of the two `if` branches?

esennesh:
> Maybe @esennesh, it's worth adding tests that show how this is supposed to work for each of the two `if` branches?

Hmmm... I thought I had something on one of my other branches for submission that was gonna exercise this, but since I'm having a hard time finding that, I very much should add a test for exercising both branches.

esennesh:
> Thanks @esennesh . Most of it makes a lot of sense.
> 
> Still, I wanna ask @fehiepsi for a review on the specific change
> 
> > Handle the TODO on using sampling with intermediates in expanded and Independent distributions
> 
> As I could be missing something.
> 
> Maybe @esennesh, it's worth adding tests that show how this is supposed to work for each of the two `if` branches?

Got a test in there now for the if-yes branch, while the if-no branch is the default codepath exercised by most, though not quite all (hence the patch), existing code.

esennesh:
Is there a way to convert the PR to pull from the `master` branch on my fork, so I can merge things into `develop` while the PR is under review?

numpyro/primitives.py
```diff
@@ -286,20 +286,24 @@ def param(
         )
         return init_value
 
+    args: tuple
     if callable(init_value):
 
-        def fn(init_fn: Callable, *args, **kwargs) -> ArrayLike:
-            return init_fn(prng_key())
+        def fn(*args, init_value: Callable = init_value, **kwargs) -> ArrayLike:
```
fehiepsi, line 292: Seems like we dont need init_value here?

fehiepsi (approved):

numpyro/primitives.py
esennesh: It seems like we don't, but IME we actually do.  Python behaves very nastily with closures and Jax behaves even worse.  Generally IME the best way to actually close over something is like this: make the closed-over value the default for a pseudo-"argument", to make sure it becomes part of the function, gets passed by value, and survives closure conversion in the correct form.

If there was a way to demonstrate we don't need it, we could get rid of it, but again IME nasty bugs tend to show up after I mistake Python for Haskell in a case like this.

esennesh:
Looks like it's ready to merge, no?

PR #2218, opened 2026-07-09
Pin ty and improve type hints

### LINKED ISSUE #2216: CI lint/prek broken for all PRs: unpinned `ty` + `uv sync --upgrade` surfaces new type errors
# CI `lint`/`prek` broken for all PRs: unpinned `ty` + `uv sync --upgrade` surfaces new type errors

## Summary

The `lint` and `prek` jobs run `ty check`, and CI installs `ty` via `uv sync --upgrade` with only a lower bound (`ty>=0.0.49`). A recent `ty` release started reporting two type errors in existing `master` code. Because the version floats, this now fails on **every open PR**, and `master` itself would fail if its CI were re-run.

Two `ty` diagnostics (both in code untouched by any recent PR):

### 1. `numpyro/contrib/control_flow/scan.py:72`

```
error[invalid-argument-type]: Argument to function `cond` is incorrect
   --> numpyro/contrib/control_flow/scan.py:72:21
```

The `lax.cond` call uses the deprecated 5-argument form
`cond(pred, true_operand, true_fun, false_operand, false_fun)`; ty's stub only
knows the current `cond(pred, true_fun, false_fun, *operands)`, so it sees the
`(value, i)` tuple where a callable is expected.

### 2. `numpyro/contrib/stochastic_support/dcc.py:270`

```
error[invalid-argument-type]: Argument to function `log_weight` is incorrect
   --> numpyro/contrib/stochastic_support/dcc.py:270:40
       Expected `int`, found `Array`
```

`log_weight(rng_key, i: int, ...)` is wrapped with `jax.vmap(..., in_axes=(None, 0, None, None))` and called with `jnp.arange(num_samples)`. ty does not model the axis-mapping vmap performs, so it rejects the `Array` argument against the `i: int` annotation.

## Reproduce

```bash
git checkout master
uv sync --upgrade --extra cpu --group ci
uv run ty check
# Found 2 diagnostics
```

## Suggested fixes (any one)

- **Pin `ty`** to a known-good version in `pyproject.toml` (and bump deliberately), so CI is reproducible instead of floating with `--upgrade`.
- **`dcc.py`**: widen the annotation to `i: int | jax.Array` (or `jax.Array`) — it is a vmapped axis.
- **`scan.py`**: modernize the `lax.cond` call to the current signature, or add a targeted `# ty: ignore[invalid-argument-type]`.

## Notes

Discovered while updating PR #2215. That PR's own (unrelated) `ty` error has been fixed; these two remain and are independent of it.

juanitorduz:
Closes https://github.com/pyro-ppl/numpyro/issues/2216

In general, it is safer to pin the version ✅

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,7 +15,7 @@ repos:
         files: "(.py$)|(.*.ipynb$)"
 
   - repo: https://github.com/astral-sh/ty-pre-commit
-    rev: v0.0.49
+    rev: v0.0.57
     hooks:
       - id: ty
 
--- a/numpyro/contrib/control_flow/scan.py
+++ b/numpyro/contrib/control_flow/scan.py
@@ -69,10 +69,10 @@ def _subs_wrapper(subs_map, i, length, site):
                 # and generate a new sample otherwise
                 return lax.cond(
                     i < shape[0],
+                    lambda val, _: val[0][val[1]],
+                    lambda _, key: site["fn"](rng_key=key, sample_shape=sample_shape),
                     (value, i),
-                    lambda val: val[0][val[1]],
                     rng_key,
-                    lambda val: site["fn"](rng_key=val, sample_shape=sample_shape),
                 )
             else:
                 raise RuntimeError(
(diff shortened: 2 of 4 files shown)


juanitorduz:
waiting ....

image

😅

Qazalbash (approved):

PR #2233, opened 2026-07-31
feat: compare branch speed against master and report it on the PR

### LINKED ISSUE #454: Create a test suite for performance testing

Qazalbash:
Adds a benchmark suite and the CI mechanism that measures a pull request against its merge base, then posts the difference as a sticky comment.

benchmarks/ holds 32 benchmarks across four suites -- MCMC, SVI, distributions, and effect handlers. Each is a setup function returning a zero-argument run callable, so data generation stays out of the measurement. The harness clears the JAX caches, calls it once, then several times warm: the fastest warm call is reported as run time, and the excess of the cold call over it as compile time.

The workflow is opt-in behind a `run-benchmarks` label, because a full comparison takes tens of minutes. It builds a venv for the head and one for the merge base, pinning the base to the head's JAX version so the report measures NumPyro rather than JAX, and measures both with the head revision of benchmarks/ so a benchmark added by the PR still runs against the base. Where it cannot run there, the row is reported as n/a with the reason rather than dropped.

Commenting is split into a second workflow_run workflow: a pull_request build of a fork gets a read-only token, so the benchmark job holds no write permission and only uploads an artifact, and the privileged job never executes code from the pull request.

Thresholds come from an A/A control run -- identical code on both sides. Run time stayed within 2%, so 5% is the band. Compile time swung 20% on one benchmark, and 216% on one with no jitted work at all, hence the looser 25% band and the 50 ms floor below which a cold/warm gap is warm-up rather than compilation.

Fixes #454

### DIFF
--- a/.github/scripts/post-sticky-comment.js
+++ b/.github/scripts/post-sticky-comment.js
@@ -1,0 +1,60 @@
+// Copyright Contributors to the Pyro project.
+// SPDX-License-Identifier: Apache-2.0
+
+// Posts a benchmark report as a pull request comment, editing the bot's own
+// previous comment in place rather than adding a new one on every run.
+//
+// Shared by both commenting paths: benchmark.yml posts directly when the PR
+// comes from a branch in this repository, and benchmark-comment.yml posts on
+// its behalf when the PR comes from a fork and the benchmark job therefore had
+// a read-only token.
+
+const MARKER = '<!-- numpyro-benchmark-report -->';
+
+// GitHub rejects comment bodies over 65536 characters.
+const MAX_BODY = 65000;
+const TRUNCATE_TO = 64000;
+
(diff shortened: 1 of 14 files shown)


fehiepsi (approved):
Looks awesome! Please feel free to merge when this is ready.

Issue #898, opened 2021-01-27
Make Gibbs kernels composable


fehiepsi:
Currently, our design in HMCGibbs, DiscreteHMCGibbs, HMCECS does not allow to compose those kernels. However, I believe that with just a few internal changes, we should be able to compose those kernels to get more flexible kernels.

### Proposal

TODO

fehiepsi:
@juanitorduz @Qazalbash Do you think we can redesign MCMC kernels to make them composable with the help of claude code? The idea was to apply each kernel for just a subset of variables and MCMC runs still work. Curious on what do you think?

juanitorduz:
> [@juanitorduz](https://github.com/juanitorduz) [@Qazalbash](https://github.com/Qazalbash) Do you think we can redesign MCMC kernels to make them composable with the help of claude code? The idea was to apply each kernel for just a subset of variables and MCMC runs still work. Curious on what do you think?

Yes! Makes a lot of sense! I will create a PR with the scaffolding design (not the code itself), and then we can proceed with the implementation :) . I have some tokens I can burn for this task :)

PR #2243, opened 2026-08-15
Design Make Gibbs kernels composable [DO NOT MERGE]

juanitorduz:
Related to https://github.com/pyro-ppl/numpyro/issues/898

@fehiepsi this is the scaffolding design ready for feedback. Once we are happy with the design, I will proceed with the implementation.


fehiepsi (commented):
Thanks @juanitorduz! The proposal looks reasonable and the cost looks minimal to me.

PR #2245, opened 2026-08-21
Fix provenance tracking for closed-over constants on JAX 0.11.1

### LINKED ISSUE #2244: `eval_provenance` fails on jax 0.11.1: closed-over constants moved from `jaxpr.constvars` into `jaxpr.invars`
### Bug Description

jax 0.11.1 changed what `trace_to_jaxpr_dynamic` returns: constants a traced function closes over, which previously landed in `jaxpr.constvars`, are now hoisted into `jaxpr.invars`.

`track_deps_jaxpr` zips `jaxpr.invars` against `provenance_inputs`, which is built only from the explicit kwargs, so the two no longer line up:

```python
# numpyro/ops/provenance.py
_safe_map(write, jaxpr.invars, provenance_inputs)
# AssertionError: length mismatch: [2, 1]
```

The jaxpr shape either side of the change, from the same script:

```
jax 0.11.0: invars=1 constvars=1 consts_returned=1
jax 0.11.1: invars=2 constvars=0 consts_returned=1
```

Anything reaching `eval_provenance` fails as soon as the model closes over arrays. We hit it through `TraceGraph_ELBO` -> `get_nonreparam_deps` -> `eval_provenance`: a model closing over 45 constants with 1 latent gives `length mismatch: [46, 1]`. Downstream report: infer-actively/pymdp#447.

One note on the underlying fragility: `provenance.py` imports `trace_to_jaxpr_dynamic` from `jax._src.interpreters.partial_eval`, a private API, which is why this module has broken repeatedly on jax releases (#2154 on jax 0.9.2, #2119 on pxla, #2173 on `xla_pmap_p`). The immediate fix may be as small as accounting for the hoisted constants when the invars and inputs lengths differ, but if there is a supported path that keeps constants separate, threading provenance through that would likely be more durable. Happy to open a PR if you can point me at the approach you would prefer.

### Steps to Reproduce

Environment: numpyro 0.21.0, jax/jaxlib 0.11.1, python 3.13 and 3.14, linux and macOS.

```python
import jax.numpy as jnp
from numpyro.ops.provenance import eval_provenance

BIG = jnp.arange(5.0)
print(eval_provenance(lambda z: (BIG * z).sum(), z=jnp.array(1.0)))
```

- jax 0.11.0 -> `frozenset({'z'})`
- jax 0.11.1 -> `AssertionError: length mismatch: [2, 1]`

Any function closing over a constant triggers it. The `eval_provenance` docstring example (`lambda x, y, z: x + y`) still passes on 0.11.1, because it closes over nothing.

### Expected Behavior

`eval_provenance` should return `frozenset({'z'})` on jax 0.11.1, as it does on 0.11.0: constants closed over by the traced function carry no provenance from the tracked kwargs and should not be zipped against them.

tandede:
Fixes #2244.

## Changes made

- Replace the private `trace_to_jaxpr_dynamic` call with the public `jax.make_jaxpr` tracing API.
- Keep closed-over constants separate from the explicit inputs whose provenance is tracked, including on JAX 0.11.1.
- Adapt the call-primitive regression helper to JAX 0.11.1's `call_jaxpr` binding protocol.
- Add regression coverage for an array constant captured by a closure.

## Why

JAX 0.11.1 hoists constants returned by `trace_to_jaxpr_dynamic` into `jaxpr.invars`. NumPyro then paired those invars with provenance values built only from explicit keyword arguments, causing a length mismatch whenever a traced function closed over an array.

`jax.make_jaxpr` preserves the distinction between constants and explicit inputs through its public API. Constants therefore retain empty provenance while tracked keyword arguments remain aligned with `jaxpr.invars`. This also removes a direct dependency on JAX's private partial-evaluation API.

## Tests

- JAX/JAXLIB 0.11.1, Python 3.13: `pytest -q test/ops test/infer/test_inspect.py test/infer/test_svi.py` — 415 passed, 4 skipped.
- JAX/JAXLIB 0.7.0, Python 3.11: `pytest -q test/ops/test_provenance.py` — 11 passed.
- `ruff check .`
- `ruff format . --check`
- `python scripts/update_headers.py --check`
- `ty check numpyro/ops/provenance.py test/ops/test_provenance.py`

## Dependencies

No new dependencies.

### DIFF
--- a/numpyro/ops/provenance.py
+++ b/numpyro/ops/provenance.py
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import jax
-from jax.api_util import debug_info, flatten_fun, shaped_abstractify
+from jax.api_util import debug_info, flatten_fun
 from jax.extend.core import Literal
 from jax.extend.core.primitives import call_p, closed_call_p, jit_p
 
@@ -11,7 +11,6 @@ try:
 except ImportError:
     xla_pmap_p = None
 import jax.extend.linear_util as lu
-from jax.interpreters.partial_eval import trace_to_jaxpr_dynamic
 
 
 # Adapted from definition in jax v0.5.0
@@ -52,24 +51,9 @@ def eval_provenance(fn, **kwargs):
         else {}
     )
     wrapped_fun, out_tree = flatten_fun(lu.wrap_init(fn, **fn_info), in_tree)
-    # Abstract eval to get output pytree
-    avals = _safe_map(shaped_abstractify, args)
-    # Note: we split out the process of abstract evaluation and provenance tracking
-    # for simplicity. In principle, they can be merged so that we only need to walk
-    # through the equations once.
-
-    wrapped_info = (
-        dict(
-            debug_info=debug_info(
-                "eval_provenance wrapped", wrapped_fun.call_wrapped, args, {}
-            )
-        )
-        if debug_info is not None
-        else {}
-    )
-    jaxpr, avals_out, _ = trace_to_jaxpr_dynamic(
-        lu.wrap_init(wrapped_fun.call_wrapped, {}, **wrapped_info), avals
-    )
+    # Use JAX's public tracing API so closed-over constants remain separate from
+    # the explicit inputs whose provenance is tracked.
+    jaxpr = jax.make_jaxpr(wrapped_fun.call_wrapped)(*args)
 
     # get provenances of flatten kwargs
     aval_kwargs = {}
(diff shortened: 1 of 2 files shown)


fehiepsi (approved):
Thanks @tandede! Would this change work for older jax releases?

tandede:
> Thanks @tandede! Would this change work for older jax releases?

Yes. NumPyro currently declares jax>=0.7.0, so I tested the current PR at that supported lower bound as well as on JAX 0.11.1.

* JAX/JAXLIB 0.7.0, Python 3.11: python -m pytest -q test/ops — 337 passed
* JAX/JAXLIB 0.11.1, Python 3.13: python -m pytest -q test/ops — 337 passed

Both runs include the new closed-over-constant regression. The production change uses the public jax.make_jaxpr API, which is available at the 0.7.0 lower bound, so no version-specific production path is needed.

Qazalbash (approved):
Awesome! Thank you @tandede

PR #2246, opened 2026-08-22
Add cat constraint and CatTransform

### LINKED ISSUE #1872: Make Interval safe to inf values?
Hello!

I have a custom multi-dimensional distribution where the support may be truncated along some dimensions. In terms of constraints, some dimensions will either be `real`, `greater_than`, `less_than`, or `interval`. I naively was then implementing the `support` as, e.g.:
```python
ivl = constraints.interval([0., -jnp.inf, 5.], [jnp.inf, 0., 10.])
```
Right now, this is not really supported by the `numpyro.distributions.constraints.Interval` class because of how [`feasible_like()`](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/distributions/constraints.py#L514C5-L517C10) works, or how the `scale` is computed in the [unconstrained transform](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/distributions/transforms.py#L1604). Would you be open to making these things inf-safe? So far I instead implemented a custom subclass `InfSafeInterval(constraints._Interval)` to support this, but thought I would check in on this. Thanks!

aswanth-07:
Closes #1872.

## Summary

- add `constraints.cat` for applying heterogeneous constraints to consecutive slices, including shape validation, feasible-value construction, equality, and JAX pytree support
- add `CatTransform` with forward/inverse operations, per-component Jacobians, intermediate propagation, and `biject_to` registration
- cover mixed interval/positive domains, JIT execution, pytree round-trips, invalid layouts, and transform round-trips

The public API follows the corresponding PyTorch `constraints.cat` / `CatTransform` naming and slice semantics. Malformed layouts are rejected explicitly so a constraint cannot silently leave trailing values unchecked.

## Tests

- `pytest -q test/test_transforms.py` — 239 passed
- focused constraint tests — 42 passed
- `ruff check` and `ruff format` on changed files — passed
- `ty check numpyro/distributions/constraints.py numpyro/distributions/transforms.py` — passed
- `scripts/update_headers.py --check` — passed

No new dependencies.

I used OpenAI Codex to assist with source comparison, implementation, and test execution; I reviewed the final diff and validation results.

### DIFF
--- a/numpyro/distributions/constraints.py
+++ b/numpyro/distributions/constraints.py
@@ -28,6 +28,7 @@
 
 __all__ = [
     "boolean",
+    "cat",
     "circular",
     "complex",
     "corr_cholesky",
@@ -65,7 +66,7 @@ __all__ = [
 ]
 
 import math
-from typing import ClassVar, Generic, Optional, cast
+from typing import ClassVar, Generic, Optional, Sequence, cast
 
 import numpy as np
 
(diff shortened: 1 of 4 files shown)


numpyro/distributions/constraints.py
```diff
@@ -427,6 +428,99 @@ def eq(self, other: object, static: bool = False) -> ArrayLike:
...
+    """
+
+    def __init__(
+        self,
+        cseq: Sequence[Constraint],
+        dim: int = 0,
+        lengths: Optional[Sequence[int]] = None,
+    ) -> None:
+        assert all(isinstance(c, Constraint) for c in cseq)
+        assert cseq
```
Qazalbash: I think `assert cseq` should come first.

numpyro/distributions/constraints.py
```diff
@@ -427,6 +428,99 @@ def eq(self, other: object, static: bool = False) -> ArrayLike:
...
+        assert isinstance(dim, int)
```
Qazalbash: Brief error message would be helpful here.

numpyro/distributions/constraints.py
```diff
@@ -427,6 +428,99 @@ def eq(self, other: object, static: bool = False) -> ArrayLike:
...
+        self.cseq = tuple(cseq)
+        if lengths is None:
+            lengths = (1,) * len(self.cseq)
+        assert len(lengths) == len(self.cseq)
+        assert all(isinstance(length, int) and length >= 0 for length in lengths)
```
Qazalbash: Brief error message would be helpful here.

numpyro/distributions/transforms.py
```diff
@@ -454,6 +455,157 @@ def eq(self, other: object, static: bool = False) -> ArrayLike:
...
+
+    def __init__(
+        self,
+        tseq: Sequence[Transform],
+        dim: int = 0,
+        lengths: Optional[Sequence[int]] = None,
+    ) -> None:
+        assert all(isinstance(t, Transform) for t in tseq)
+        assert tseq
+        assert isinstance(dim, int)
```
Qazalbash: Brief error message would be helpful here.

numpyro/distributions/transforms.py
```diff
@@ -454,6 +455,157 @@ def eq(self, other: object, static: bool = False) -> ArrayLike:
...
+        self.transforms = tuple(tseq)
+        if lengths is None:
+            lengths = (1,) * len(self.transforms)
+        assert len(lengths) == len(self.transforms)
+        assert all(isinstance(length, int) and length >= 0 for length in lengths)
```
Qazalbash: Brief error message would be helpful here.

numpyro/distributions/transforms.py
```diff
@@ -454,6 +455,157 @@ def eq(self, other: object, static: bool = False) -> ArrayLike:
...
+    """
+
+    def __init__(
+        self,
+        tseq: Sequence[Transform],
+        dim: int = 0,
+        lengths: Optional[Sequence[int]] = None,
+    ) -> None:
+        assert all(isinstance(t, Transform) for t in tseq)
+        assert tseq
```
Qazalbash: I think `assert tseq` should come first.

test/test_distributions.py
```diff
@@ -2649,6 +2649,15 @@ def test_beta_proportion_invalid_mean():
...
         (constraints.boolean, np.array([-1, 1]), np.array([False, True])),
+        (
+            constraints.cat(
+                [constraints.interval(-1, 1), constraints.positive],
+                dim=-1,
+                lengths=[2, 1],
+            ),
+            np.array([[0.0, 2.0, 1.0], [-2.0, 0.5, -1.0]]),
+            np.array([[True, False, True], [False, True, False]]),
+        ),
```
Qazalbash, line 2660: Add 2 or 3 more cases for better coverage.

Qazalbash (changes requested):

numpyro/distributions/constraints.py
aswanth-07: Good catch - moved the non-empty check ahead of element validation and added a regression assertion for the empty case.

numpyro/distributions/constraints.py
aswanth-07: Added a concise "dim must be an integer" message and a regression assertion for it.

numpyro/distributions/constraints.py
aswanth-07: Added a concise nonnegative-integers message here, plus an explicit message for a length-count mismatch.

numpyro/distributions/transforms.py
aswanth-07: Added the same concise integer-dimension message to CatTransform and covered it in the validation test.

numpyro/distributions/transforms.py
aswanth-07: Added the nonnegative-integers message here as well, with regression coverage.

numpyro/distributions/transforms.py
aswanth-07: Done - the empty transform-sequence check now runs first, with a regression assertion for that ordering.

test/test_distributions.py
aswanth-07: Added three cases covering default lengths with dim=0, unequal lengths along a non-last dimension, and a zero-length slice.

Qazalbash (approved):

fehiepsi (approved):

PR #2249, opened 2026-08-22
Make Gibbs kernels composable

juanitorduz:
Implements the design in #2243 for #898.

## Summary
- New `Gibbs([(kernel, sites), ...])` composite kernel plus `CustomGibbs` and `DiscreteGibbs` block kernels (`numpyro/infer/gibbs.py`, `gibbs_util.py`). Blocks can be HMC/NUTS, `DiscreteGibbs`, `CustomGibbs`, `MixedHMC` or a nested `Gibbs`.
- Opt-in `MCMCKernel.refresh` / `wrap_model` / `get_constrain_fn` hooks and public `HMC.get_potential_fn` / `get_constrain_fn` / `refresh` / `wrap_model`, so no kernel reads another kernel's private attributes.
- `HMCGibbs` and `DiscreteHMCGibbs` are thin facades over `Gibbs` (constructors, behavior and `extra_fields=["hmc_state.*"]` unchanged); `HMCECS` and `MixedHMC` are standalone wrappers on the public accessors.
- Fixes: `HMCECS` no longer fails on `warmup()` followed by `run()` (model is wrapped once at construction); unused/reused RNG keys; the dead stay branch of the modified random-walk proposal; `init_params` is no longer mutated.

## Deviations from the design doc (found during implementation)
1. A third hook, `MCMCKernel.get_constrain_fn`, is needed so the composite can condition siblings on any block's constrained values (the design only had it on `HMC`).
2. The `changed`-gated refresh is done at two points per sweep (before a block, on earlier siblings; after the sweep, on later siblings), which keeps it state-free and sound.
3. `_prepare_model_for_potential(..., enum=)` is explicit and `DiscreteGibbs` passes `enum=` to `potential_energy`, otherwise `enumerate="parallel"` sites give wrong energies and every discrete block would import funsor.
4. v1 raises for an HMC block with value-dependent supports inside `Gibbs` (its constrain function would need the siblings' values, which is circular).
5. `HMCECS` was decoupled before the facades, since it subclassed `HMCGibbs`.

## Notes for reviewers
- The `DiscreteHMCGibbs` and `HMCECS` docstring assertions were loosened from 0.1 to 0.2: the refactor changes the RNG streams, and across 8 to 10 seeds the seed-to-seed sd of the asserted means is 0.07 (GMM) and 0.1 to 0.14 (ECS) on master and this branch alike, with no bias.
- Minor break: `HMCGibbsState` is now a `GibbsState` subclass (`z, block_states, rng_key`) with an `hmc_state` property; positional construction of the old 3-tuple no longer works.
- Tests: new `test/infer/test_gibbs.py` and `test/infer/test_kernel_hooks.py`; existing Gibbs/ECS/pickle tests pass.


juanitorduz:
Pushed a231cc86 addressing the CI failures plus two structural follow-ups:

- **CI fix**: the three `test_discrete_gibbs_multiple_sites_chain[2]` failures were a tracer leak in `DiscreteGibbs.init`/`MixedHMC.init`: `np.asarray(ravel_pytree(support_sizes)[0])` raises under `pmap`'s staging trace because `ravel_pytree` stages a `jnp.concatenate` even on constant inputs (CI's upgraded JAX exposes multiple CPU devices by default, so `MCMC.run` takes the `pmap` branch; a single-device machine falls back to the eager loop, which is why it passed locally). The flat support sizes are now built with pure numpy in `ravel_pytree` leaf order. Added a device-independent regression test (single-chain `pmap` over `init`/`sample` for `DiscreteGibbs`, `MixedHMC`, `DiscreteHMCGibbs`) and added `test_gibbs.py -k chain` to the forced multi-device CI step. Side effect: the stored dtype changes int32 -> int64 (benign, consumers re-wrap with `jnp.asarray`).
- **Module layout**: `numpyro/infer/gibbs_util.py` is merged into `numpyro/infer/gibbs.py` (helpers at the top, discrete proposals next to `DiscreteGibbs`), so the feature lives in one module.
- **Type checking**: every module this PR touches is now in the `ty` include list (`make lint` covers them) and all diagnostics are fixed. Two are worth reviewer attention: `numpyro.util.cond`'s `pred` annotation widened to `bool | jax.Array` (it was already called with traced booleans everywhere), and `mixed_hmc.py`'s one deprecated 5-arg `lax.cond` call now uses numpyro's own `cond` wrapper like the surrounding code (same argument order and behavior; additionally honors `NUMPYRO_DISABLE_CONTROL_FLOW_PRIM`).

juanitorduz:
@fehiepsi @Qazalbash I think this one si ready for review 🤗

fehiepsi (approved):
Great work, thanks @juanitorduz!

The design looks clean. Could you point me to a test/example that is best illustrating the usage?

juanitorduz:
> Great work, thanks @juanitorduz!
> 
> The design looks clean. Could you point me to a test/example that is best illustrating the usage?

I think `test/infer/test_gibbs.py::test_three_blocks_doctest_model` is the best example

Here is another self-contained example 

```python

from jax import random
import jax.numpy as jnp

import numpyro
from numpyro.infer import MCMC, NUTS, CustomGibbs, DiscreteGibbs, Gibbs
from numpyro.infer.gibbs import discrete_latent_sites

probs = jnp.array([0.15, 0.3, 0.3, 0.25])
locs = jnp.array([-2.0, 0.0, 2.0, 4.0])


def model(probs, locs):
    c = numpyro.sample("c", dist.Categorical(probs))
    x = numpyro.sample("x", dist.Normal(locs[c], 0.5))
    y = numpyro.sample("y", dist.Normal(0.0, 2.0))
    numpyro.sample("obs", dist.Normal(x + y, 1.0), obs=jnp.array([1.0]))


def y_gibbs_fn(rng_key, gibbs_sites, hmc_sites):
    # exact conditional y | x, obs: prior N(0, 4), likelihood N(1 - x, 1)
    x = hmc_sites["x"]
    return {"y": dist.Normal(0.8 * (1 - x), jnp.sqrt(0.8)).sample(rng_key)}


kernel = Gibbs([
    (DiscreteGibbs(model), discrete_latent_sites),
    (CustomGibbs(y_gibbs_fn), ["y"]),
    (NUTS(model), None),
])
mcmc = MCMC(kernel, num_warmup=1000, num_samples=20000, progress_bar=False)
mcmc.run(random.key(0), probs, locs)
mcmc.print_summary()


# Reference: NUTS on the model with `c` marginalized out analytically.
def marginal_model(probs, locs):
    x = numpyro.sample("x", dist.MixtureSameFamily(dist.Categorical(probs), dist.Normal(locs, 0.5)))
    y = numpyro.sample("y", dist.Normal(0.0, 2.0))
    numpyro.sample("obs", dist.Normal(x + y, 1.0), obs=jnp.array([1.0]))


ref = MCMC(NUTS(marginal_model), num_warmup=1000, num_samples=20000, progress_bar=False)
ref.run(random.key(0), probs, locs)
ref.print_summary()
```

image

fehiepsi:
Thanks! Could you expose the helpers in gibbs to the docs?

juanitorduz:
> Thanks! Could you expose the helpers in gibbs to the docs?

As in https://github.com/pyro-ppl/numpyro/pull/2249/commits/eea40183e7cdbc29ac0cb04165c76af238e7279b ?

PR #2250, opened 2026-08-23
Drop the masked logsumexp reduction in mixture `log_prob`

kyo219:
## Changes made

- `numpyro/distributions/mixtures.py`: `_MixtureBase.log_prob` passed `where=~jnp.isneginf(sum_log_probs)` to `jax.nn.logsumexp`. A masked reduction prevents XLA from using its fused reduction path, which makes the reduction the dominant cost of mixture likelihood evaluation. The mask is redundant: the value-preserving `jnp.where(jnp.isneginf(...), -inf, ...)` introduced by #1874 already produces the same values and, through `where`'s select-based gradient, stops gradients from flowing into `-inf` component log-probs. The `where=` kwarg was not part of that fix — it slipped in with the type-hints PR #2032. This change restores the #1874 formulation and adds a comment explaining why the value-preserving `where` must stay.

Values and gradients are unchanged (verified identical for regular values, at zero mixing weights, and for rows where every component is `-inf`).

## Benchmarks

Apple M2 (24 GB), CPU, jax 0.10.2, Python 3.11. Jitted steady-state per-call time, best of 300 calls after warmup, 10,000 values, 5 components (3 for `MixtureGeneral` with StudentT).

| workload | master | this PR | speedup |
|---|---|---|---|
| `MixtureSameFamily(Categorical, Normal)` log_prob | 0.379 ms | 0.128 ms | 2.96× |
| same, grad of summed log_prob | 0.265 ms | 0.211 ms | 1.26× |
| `MixtureGeneral` (5 × Normal) log_prob | 0.14 ms | 0.11 ms | 1.25× |
| `MixtureGeneral` (2 × Normal + StudentT) log_prob | 0.11 ms | 0.13 ms | 0.85× |

The big win is `MixtureSameFamily`, where the component log-probs come from one broadcast producer and the masked reduction forced it to be rematerialized per pass. The last row shows a small absolute regression (~0.02 ms) for one composition where the component log_prob dominates; it appears to be an XLA fusion-choice artifact rather than a systematic cost.

#1874 introduced the `-inf` gradient protection this PR preserves; #2032 incidentally added the `where=` kwarg this PR removes.

## Tests

- New `test_mixture_log_prob_grad_at_zero_weights` and `test_mixture_log_prob_grad_all_components_neg_inf` in `test/test_distributions_mixture.py` — lock in the #1874 semantics (no NaN gradients at zero mixing weights and for all-`-inf` rows). Neither case was previously covered by a test.
- `pytest test/test_distributions_mixture.py` — 47 passed
- `ruff check` / `ruff format --check` clean; `ty check` reports the same 10 pre-existing diagnostics as master.

## Dependencies

None.

### DIFF
--- a/numpyro/distributions/mixtures.py
+++ b/numpyro/distributions/mixtures.py
@@ -168,14 +168,12 @@ class _MixtureBase(Distribution):
     def log_prob(self, value: ArrayLike, intermediates=None) -> ArrayLike:
         del intermediates
         sum_log_probs = self.component_log_probs(value)
+        # the value-preserving where stops gradients from flowing into -inf
+        # component log-probs (e.g. at zero mixing weights); see #1874
         safe_sum_log_probs = jnp.where(
             jnp.isneginf(sum_log_probs), -jnp.inf, sum_log_probs
         )
-        return jax.nn.logsumexp(
-            safe_sum_log_probs,
-            where=~jnp.isneginf(sum_log_probs),  # for numerical stability
-            axis=-1,
-        )
+        return jax.nn.logsumexp(safe_sum_log_probs, axis=-1)
 
 
 class MixtureSameFamily(_MixtureBase):
--- a/test/test_distributions_mixture.py
+++ b/test/test_distributions_mixture.py
@@ -203,3 +203,35 @@ def test_mixture_rejects_parameter_dependent_components(component_dist):
 def test_mixture_accepts_parameter_free_components(component_dist):
     mixing_dist = dist.Categorical(probs=np.array([0.3, 0.7]))
     dist.MixtureSameFamily(mixing_dist, component_dist)
+
+
+def test_mixture_log_prob_grad_at_zero_weights():
+    # regression test for #1874: gradients must not become NaN when some
+    # mixing weights are exactly zero
+    def loss(probs):
+        mixture = dist.MixtureSameFamily(
+            dist.Categorical(probs=probs),
+            dist.Normal(jnp.arange(3.0), jnp.ones(3)),
+        )
+        return mixture.log_prob(jnp.array([0.5, 2.0])).sum()
+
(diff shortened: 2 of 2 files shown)


Qazalbash:
Thanks @kyo219!

Can you share the code used for benchmarks.

kyo219:
Sure! Benchmark script below (Apple M2, CPU, jax 0.10.2; jitted steady-state, best of 300 calls after warmup). The same setup was used to check the value/gradient agreement mentioned in the PR description (log_prob values, grads, zero-mixing-weight grads, and all-components `-inf` rows compared across both branches).

```python
import time
import jax, jax.numpy as jnp

N = 10_000
mix = dist.Categorical(probs=jnp.array([0.2, 0.3, 0.1, 0.25, 0.15]))
comp = dist.Normal(jnp.arange(5.0), jnp.ones(5))
m = dist.MixtureSameFamily(mix, comp)
v = jax.random.normal(jax.random.PRNGKey(0), (N,)) * 3

lp = jax.jit(m.log_prob)
g = jax.jit(jax.grad(lambda v: m.log_prob(v).sum()))

def bench(f, x, n=300):
    jax.block_until_ready(f(x))
    t0 = time.perf_counter()
    for _ in range(n):
        r = f(x)
    jax.block_until_ready(r)
    return (time.perf_counter() - t0) / n

print(f"log_prob: {bench(lp, v)*1e3:.3f} ms  grad: {bench(g, v)*1e3:.3f} ms")
```

On the CI benchmark report: the `mixture_same_family_log_prob` -51.8% line matches this. The flagged regressions (`normal_sample`, `initialize_model_hierarchical`, the SVI ELBOs) don't exercise `mixtures.py` at all, so I believe those are runner noise.

Qazalbash (approved):
Thanks @kyo219

PR #2251, opened 2026-08-23
Speed up posterior diagnostics (summary 1.7x, effective_sample_size 2.1x)

kyo219:
## Changes made

All in `numpyro/diagnostics.py`:

- **`autocovariance` is now the FFT primitive, `autocorrelation` derives from it.** Previously `autocorrelation` normalized the raw FFT autocovariance by its lag-0 value, and `autocovariance` then multiplied the variance back in (`autocorrelation(x) * x.var()`) — two extra full-array passes plus a separate variance pass, and a redundant `.astype(np.float64)` copy (the FFT output is already float64). `effective_sample_size` consumes `autocovariance`, so its hot path drops all of these. `autocorrelation` still returns the same values, now computed as `autocov / autocov[lag 0]` (the lag-0 autocovariance is identical for the biased and unbiased estimators, which is what the old normalization relied on implicitly).
- **FFTs go through `scipy.fft` with `workers=-1`**, with an `np.fft` fallback. scipy is not a direct numpyro dependency but is guaranteed transitively via jax, and numpyro already uses it in several lazy imports (e.g. `distributions/continuous.py`, `infer/calibration.py`); the test suite also imports `scipy.fftpack` in `test_diagnostics.py`. `np.fft` is single-threaded; `scipy.fft` parallelizes across the batch dimensions, which is exactly the shape of the ESS workload (one FFT per chain per parameter column). Since `scipy.fft` computes in the input precision (unlike `np.fft`, which always promotes), the centered signal is explicitly cast to float64 to keep the double-precision behavior.
- **`summary()` sorts each parameter's draws once** and reuses the sorted array for both the median and the hpdi bounds. Previously `np.median` partitioned the very array that `hpdi` then fully sorted. The sliding-window part of `hpdi` is factored into a private `_hpdi_of_sorted` helper; the public `hpdi` is unchanged. The median-from-sorted matches `np.median` semantics including NaN propagation (sorting places NaNs last, so checking the last draw suffices).

Behavioral notes: float64 results agree with master to rtol 1e-9. For float32 inputs, results differ at the float32-round-off level (relative ~1e-7) because the old code normalized with a float32 `x.var()` while the new code stays in float64 throughout — the new values are the more precise ones. `autocovariance` of a constant series now returns 0 rather than NaN (the old NaN came from the 0/0 lag-0 normalization that the variance multiplication couldn't undo); `effective_sample_size` and `autocorrelation` still return NaN for that case as before.

## Benchmarks

Apple M2 (24 GB, 8 cores), CPU, numpy 2.x/scipy via jax 0.10.2, Python 3.11. Best of 3 after warmup.

| workload | master | this PR | speedup |
|---|---|---|---|
| `summary()` on {a: (4, 2000, 100), b: (4, 2000, 50, 10)} | 0.317 s | 0.199 s | 1.59× |
| `summary()` on (4, 5000, 2000) | 3.50 s | 2.08 s | 1.69× |
| `effective_sample_size` on (4, 2000, 50, 10) | 0.107 s | 0.052 s | 2.06× |
| `hpdi` standalone (8000, 500) | 0.100 s | 0.100 s | 1.0× (unchanged by design) |

This is the `mcmc.print_summary()` path, so every MCMC run with the default summary printout benefits.

None.

## Tests

- New `test_summary_median_matches_numpy` (odd and even draw counts) and `test_summary_median_propagates_nan` in `test/test_diagnostics.py` — lock in that the shared-sort median is exactly `np.median`, including NaN propagation.
- Existing coverage: `pytest test/test_diagnostics.py` — 28 passed (autocorrelation/autocovariance/ESS/hpdi/gelman-rubin values are asserted against fixed references there).
- Cross-checked patched vs master outputs on random, AR(1), float32, integer, and NaN-containing inputs for `autocorrelation`/`autocovariance` (bias=True/False), `effective_sample_size`, `hpdi`, and every `summary` statistic: float64 agrees to rtol 1e-9; float32 differs only at round-off as described above.
- `ruff check` / `ruff format --check` / `ty check` clean.

## Dependencies

None (scipy is used only behind a try/except fallback).

### DIFF
--- a/numpyro/diagnostics.py
+++ b/numpyro/diagnostics.py
@@ -15,6 +15,23 @@ from numpy.typing import NDArray
 import jax
 from jax import device_get
 
+try:
+    import scipy.fft
+
+    def _rfft(x: NDArray, n: int, axis: int) -> NDArray:
+        return scipy.fft.rfft(x, n=n, axis=axis, workers=-1)
+
+    def _irfft(x: NDArray, n: int, axis: int) -> NDArray:
+        return scipy.fft.irfft(x, n=n, axis=axis, workers=-1)
+except ImportError:  # pragma: no cover
+    # scipy is not a direct dependency of numpyro (only a transitive one via jax)
+    def _rfft(x: NDArray, n: int, axis: int) -> NDArray:
+        return np.fft.rfft(x, n=n, axis=axis)
+
+    def _irfft(x: NDArray, n: int, axis: int) -> NDArray:
+        return np.fft.irfft(x, n=n, axis=axis)
+
+
 __all__ = [
     "autocorrelation",
     "autocovariance",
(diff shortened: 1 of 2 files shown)


Qazalbash:
Thanks @kyo219!

Can you share the code you used for benchmarks.

kyo219:
Sure! Script below (Apple M2, 8 cores, CPU; best of 3 after warmup). Agreement was checked separately by dumping every statistic (`autocorrelation`/`autocovariance` with bias=True/False, `effective_sample_size`, `hpdi`, all `summary` stats) on random / AR(1) / float32 / integer / NaN-containing inputs from both branches and comparing with `np.allclose(rtol=1e-9)`.

```python
import time
import numpy as np
from numpyro.diagnostics import summary, effective_sample_size

def timeit(fn, n=3):
    fn()
    ts = []
    for _ in range(n):
        t0 = time.perf_counter(); fn(); ts.append(time.perf_counter() - t0)
    return min(ts)

rng = np.random.default_rng(0)
small = {"a": rng.standard_normal((4, 2000, 100)), "b": rng.standard_normal((4, 2000, 50, 10))}
big = {"c": rng.standard_normal((4, 5000, 2000))}
print(f"summary small: {timeit(lambda: summary(small)):.3f}s")
print(f"summary big:   {timeit(lambda: summary(big)):.3f}s")
print(f"ess:           {timeit(lambda: effective_sample_size(small['b'])):.3f}s")
```

Qazalbash (approved):
The gains are noticable. These numbers are on 13th Gen i5-13420H.

| | `master` | #2251 |
|-|-|-|
summary small: | 0.509s | 0.279s |
summary big:   | 5.281s | 2.539s |
ess:           | 0.246s | 0.158s |

Thanks @kyo219

PR #2256, opened 2026-08-27
fix(distributions): handle zero-mean negative binomial

### LINKED ISSUE #2193: Handle zero-mean Negative Binomial distributions
## Description 
The negative binomial distribution is well-defined with mean $\mu = 0$. Equivalently, it is well-defined with $p = 0$ in the `probs`/`total_count` parametrization used by `NegativeBinomialProbs`. In practice, `NegativeBinomial2` fails validation with mean 0. More concerningly, `NegativeBinomialProbs` passes validation with `probs = 0` but then behaves incorrectly, yielding `nan` log prob values. Forward sampling and evaluation of the distribution mean work correctly.

This derives from the fact that `NegativeBinomialProbs(probs=jnp.array(0))` and `NegativeBinomial2(mean=jnp.array(0))` both end up constructing a `GammaPoisson` with an `inf` `rate`:

https://github.com/pyro-ppl/numpyro/blob/a1bcf5e5b4f7c8f24dd3b8af6ffd889a37578498/numpyro/distributions/conjugate.py#L483

https://github.com/pyro-ppl/numpyro/blob/a1bcf5e5b4f7c8f24dd3b8af6ffd889a37578498/numpyro/distributions/conjugate.py#L554

## Action
I think this can and should be patched. I am happy to contribute a PR, but would appreciate some discussion on approach. In particular, do we want `inf`-rate `GammaPoisson` instances to be acceptable and well-defined? My instinct is yes but I can imagine there might be sharp bits.

## Details and reprex

For `NegativeBinomial2`, the mean is given a `constraints.positive`, so construction fails with `validate_args=True`. Note: construction fails _before_ validation with a divide-by-zero error if the mean is an int or float rather than a jax or numpy array. 

```Python
import jax.numpy as jnp

dist.NegativeBinomial2(mean = jnp.array(0), concentration = 1, validate_args=True) # fails validation
# ValueError: NegativeBinomial2 distribution got invalid mean parameter.
```

```Python
dist.NegativeBinomial2(mean = 0, concentration = 1, validate_args=True) # divide-by-zero error
# ZeroDivisionError: division by zero
```

For `NegativeBinomialProbs`, probs is given `constraints.unit_interval`. Construction and evaluation of the log prob therefore succeed with `probs=jnp.array(0)` regardless of whether`validate_args=True` (Note: as in the `NegativeBinomial2` case, construction fails with a divide by zero error if the probs is a python int or float). But we get `nan` log prob values:

```Python
dist.NegativeBinomialProbs(total_count=1, probs=0, validate_args=True) # divide by 0 error
# ZeroDivisionError: float division by zero
```
```Python
my_neg_bin = dist.NegativeBinomialProbs(total_count=1, probs=jnp.array(0), validate_args=True) # construction succeeds
```

```Python
my_neg_bin.log_prob(0) # nan log prob; should be 0
# Array(nan, dtype=float32, weak_type=True)
```

```Python
my_neg_bin.log_prob(1) # nan log prob; should be -inf
# Array(nan, dtype=float32, weak_type=True)
```

Forward sampling and evaluation of the mean behaves correctly. In that case, the underlying `GammaPoisson` handles the `inf` rate gracefully:

```Python
my_neg_bin.mean # correct mean 
# Array(0., dtype=float32, weak_type=True)
```

```Python
# correct forward sampling
with numpyro.handlers.seed(rng_seed=5):
    samples = numpyro.sample('neg bin samples', my_neg_bin.expand((100,)))

all(samples == 0)
# True
```

Behavior is similar if we construct a zero-mean `NegativeBinomial2` with `validate_args=False`. This makes sense; either way, we end up with a `GammaPoisson` with an `inf` rate.

```Python
my_neg_bin2.log_prob(0) # nan log prob, should be 0
# Array(nan, dtype=float32, weak_type=True)
```

```Python
my_neg_bin2.log_prob(1) # nan log prob, should be -inf
# Array(nan, dtype=float32, weak_type=True)
```

```Python
my_neg_bin2.mean # correct mean
# Array(0., dtype=float32, weak_type=True)
```

```Python
with numpyro.handlers.seed(rng_seed=5):
    samples = numpyro.sample('neg bin samples', my_neg_bin2.expand((100,)))

all(samples == 0)
# True
```

aswanth-07:
## Summary

- Handle zero-mean Negative Binomial distributions without scalar division errors or NaN log probabilities.
- Evaluate infinite-rate GammaPoisson boundaries with stable log-probability, variance, CDF, and sampling behavior.
- Preserve the strict-positive mean requirement for `HurdleNegativeBinomial2`.

Fixes #2193

## Tests

- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 .\.venv\Scripts\python.exe -m pytest -q test/test_distributions.py -k 'zero_mean_negative_binomial or negative_binomial_log_prob_gradient or hurdle_negative_binomial_requires_positive_mean or gamma_poisson_mixed_finite_and_infinite_rates or gamma_poisson_log_prob or test_log_prob_gradient and (GammaPoisson or NegativeBinomial)'` (27 passed)
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 .\.venv\Scripts\python.exe -m pytest -q test/test_distributions.py -k 'HurdleNegativeBinomial2'` (5 passed)
- `py -3.13 -m ruff check numpyro/distributions/conjugate.py test/test_distributions.py`
- `py -3.13 -m ruff format --check numpyro/distributions/conjugate.py test/test_distributions.py`

AI assistance disclosure: AI-assisted development tools were used during investigation and implementation. The reported tests were run against the final diff.

### DIFF
--- a/numpyro/distributions/conjugate.py
+++ b/numpyro/distributions/conjugate.py
@@ -430,7 +430,7 @@ class GammaPoisson(Distribution):
         .. math::
             \mathrm{Var}[X] = \frac{\alpha}{\lambda^2}(1 + \lambda)
         """
-        return self.concentration / self.rate * (1 + 1 / self.rate)
+        return jnp.asarray(self.concentration / self.rate * (1 + 1 / self.rate))
 
     def cdf(self, value: ArrayLike) -> Array:
         r"""If :math:`X \sim \mathrm{GammaPoisson}(\alpha, \lambda)`, then the cumulative
--- a/numpyro/distributions/conjugate.py
+++ b/numpyro/distributions/conjugate.py
@@ -7,7 +7,7 @@ from typing import Optional
 import jax
 from jax import lax, nn, random
 import jax.numpy as jnp
-from jax.scipy.special import betainc, betaln, gammaln
+from jax.scipy.special import betainc, betaln, gammaln, xlog1py, xlogy
 from jax.typing import ArrayLike
 
 from numpyro.distributions import constraints
@@ -390,12 +390,16 @@ class GammaPoisson(Distribution):
         .. math::
             p_{X}(k) = \frac{\lambda^\alpha}{(\alpha + k)(1+\lambda)^{\alpha + k}\mathrm{B}(\alpha, k + 1)}
         """
-        post_value = self.concentration + value
+        dtype = jnp.result_type(float)
+        concentration = jnp.array(self.concentration, dtype=dtype)
+        rate = jnp.array(self.rate, dtype=dtype)
+        value = jnp.array(value, dtype=dtype)
+        post_value = concentration + value
         return (
-            -betaln(self.concentration, value + 1)
+            -betaln(concentration, value + 1)
             - jnp.log(post_value)
-            + self.concentration * jnp.log(self.rate)
-            - post_value * jnp.log1p(self.rate)
+            + xlogy(-concentration, 1 + 1 / rate)
+            - xlog1py(value, rate)
         )
 
     @property
@@ -414,7 +418,7 @@ class GammaPoisson(Distribution):
         .. math::
             \mathrm{Var}[X] = \frac{\alpha}{\lambda^2}(1 + \lambda)
         """
-        return self.concentration / jnp.square(self.rate) * (1 + self.rate)
+        return self.concentration / self.rate * (1 + 1 / self.rate)
 
     def cdf(self, value: ArrayLike) -> ArrayLike:
         r"""If :math:`X \sim \mathrm{GammaPoisson}(\alpha, \lambda)`, then the cumulative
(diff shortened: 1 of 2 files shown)


fehiepsi (approved):
Thanks @aswanth-07!

fehiepsi:
Could you sync to master to trigger CI rerun?

aswanth-07:
Synced with current master in 0b5dbbaf. The PR diff remains limited to the two intended files, and CI has been retriggered. Thanks!

aswanth-07:
The rerun exposed the new output-type invariant from #2206: GammaPoisson.variance could return a Python or NumPy value after the zero-mean algebra change. I wrapped the result with jnp.asarray in 25f385af. The 4 focused output-type cases and all 27 zero-mean regression cases pass, and CI has restarted.

PR #2257, opened 2026-08-27
Add TruncatedGamma distributions

### LINKED ISSUE #969: Truncated Gamma

deeb01:
Closes #969. Supersedes #1187, inactive since 2022 — same approach, credit to @quattro.

## Summary

- add `LeftTruncatedGamma`, `RightTruncatedGamma` and `TwoSidedTruncatedGamma`, dispatched
  by `TruncatedGamma(concentration, rate, *, low=None, high=None)`, following the structure
  @fehiepsi outlined in #969
- add `gammainccinv` to `distributions/util.py`, wrapping `tfp.math.igammacinv` like the
  existing `gammaincinv`; needed for `icdf` on the upper-tail branch
- `TwoSidedTruncatedGamma` normalizes with `Q(a, λ·low) − Q(a, λ·high)` when the interval
  lies above the median, and switches `cdf`, `icdf`, `mean` and `variance` to match. The
  lower-tail form `F(high) − F(low)` is exactly zero in single precision for `Gamma(2, 1)`
  on `[30, 40]`, which would make `log_prob` `+inf`
- moments in closed form: `E[X^k] = (a)_k / λ^k · [P(a+k, λ·high) − P(a+k, λ·low)] / Z`
- three classes rather than one with infinite bounds, because
  `jax.grad(lambda a: gammainc(a, jnp.inf))` is `nan` and would break the gradient w.r.t.
  `concentration` for left truncation

Two limits, documented in the docstrings: `log_prob` returns `-inf` rather than `+inf`
once the normalizer underflows, and `icdf` returns `nan` past the point where
`tfp.math.igammainv` saturates, which `Gamma.icdf` also hits on master. A bisection
fallback would lift the second, here or as a follow-up.

## Tests

- 42 tests in `test/test_distributions.py` covering `log_prob` against scipy,
  normalization and moments against quadrature, `cdf`/`icdf` round trip, KS tests on
  samples, batch shapes, both limits, the dispatcher, gradients and NUTS recovery
- registered in the `CONTINUOUS` parametrization and in `batch_util.vmap_over`
- `make lint` passes, as do `test_distributions.py`, `test_gof.py`,
  `test_distributions_util.py`, `test_constraints.py`, `test_transforms.py` and
  `test_pickle.py`. (`test_mcmc.py::test_beta_bernoulli_x64[SA]` fails for me on master
  too, unrelated to this change.)

## Dependencies

None new; `tensorflow_probability` is already optional and used by `gammaincinv`.


juanitorduz:
Overall LGTM :)

I did an assisted review with Claude and here are the findings :)

## 1. `_partial_moment` reuses a tail switch chosen at the wrong order

`_use_upper_tail` is decided from `P(α, λ·low) > 0.5`, which is exactly right for the normalizer. But `_partial_moment(k)` evaluates the same difference at `α + k`, and the comment "the tail switch carries over unchanged" isn't true: `Gamma(α+k)` has a larger median, so an interval above the median of `Gamma(α)` can sit deep in the *lower* tail of `Gamma(α+2)`, where the `Q` form is the cancelling one.

It's reachable and it produces a negative variance:

```python
d = dist.TruncatedGamma(0.05, 1.0, low=1e-6, high=1e-5)
d.mean      # 3.985e-06   (quad: 3.9885e-06)
d.variance  # -1.588e-11  (quad: +6.320e-12)
```

The order-2 difference is the culprit — in f32, `Q(2.05, λ·low) - Q(2.05, λ·high)` evaluates to exactly `0.0` while `P(2.05, λ·high) - P(2.05, λ·low)` gives `2.6597e-11`, correct to six digits. Deciding the branch per shifted order fixes it outright:

```python
def _partial_moment(self, order):
    shifted = self.base_dist.concentration + order
    rate = self.base_dist.rate
    use_upper = gammainc(shifted, rate * self.low) > 0.5
    return jnp.where(use_upper, ...)
```

I'd also add a moment case with `concentration < 1` and a tail interval to `_TRUNCATED_GAMMA_CASES` — the current five all have `α ≥ 0.5` with intervals near the bulk, so none of them exercise this.

## 2. Ordinary left truncation silently produces NaN samples

The docstring frames the `icdf` saturation as "the same limit as `Gamma.icdf`", but `Gamma.sample` doesn't go through `icdf` — it uses `standard_gamma`. So this is a new failure mode in ordinary use, not an inherited one:

```python
dist.TruncatedGamma(2., 1., low=10.).sample(key, (200_000,))  # 4e-05 NaN
dist.TruncatedGamma(2., 1., low=15.).sample(key, (200_000,))  # 0.6%  NaN
```

`Gamma(2, 1)` truncated at 15 is not an exotic object, and 0.6% NaNs flowing into a posterior predictive is the kind of thing that costs someone an afternoon. `test_truncated_gamma_saturated_icdf_is_nan` currently codifies this as intended behaviour, which I don't think we want to ship. I'd rather see the bisection fallback land in this PR than as a follow-up — a `lax.fori_loop` bisection on `log Q` over the tail is maybe 25 lines and vectorises fine. Failing that, the `.. note::` needs to say plainly that `sample` returns NaN in the far tail, and the sampling limit belongs in the class docstring body, not only the note.

## Worth fixing, lower priority

**`variance` cancellation in f32.** Independent of point 1, `E[X²] - E[X]²` loses everything when the interval is narrow relative to its location:

```python
d = dist.TruncatedGamma(2.0, 1e-4, low=100.0, high=101.0)
d.variance  # f32: 0.1406   x64: 0.08333  (near-uniform, 1/12)
```

70% off in the default dtype. I don't think there's a clean closed form for the central moment here, so a docstring note is probably the honest answer — but `test_mean_var`'s tolerance is loose enough to hide this, so it should at least be written down.

**`mean` and `variance` aren't guarded for the underflowed normalizer.** `log_prob` returns `-inf` and `cdf` returns `nan`, both deliberate, but `mean` divides by zero and returns `nan` with no comment. Worth making the three consistent, or documenting why they differ. Relatedly: `log_prob → -inf` versus `cdf → nan` for the same underflow is a slightly odd pair; a one-line comment explaining that a density that's identically zero has a defined log-density but no defined conditional CDF would help future readers.

**Support closed at zero.** `RightTruncatedGamma` and `TwoSidedTruncatedGamma` use `constraints.interval(0., high)`, so `support(0.) is True` and `log_prob(0.)` is `+inf` for `concentration < 1`. The docstring appeals to `Beta`, but `Gamma`'s own support is `positive`, i.e. open. `constraints.open_interval` subclasses `_Interval` so `biject_to` still works. Matching the base distribution seems safer than matching `Beta` here.

**API consistency.** `dist.TruncatedDistribution(dist.Gamma(2., 1.), low=1.)` raises a bare `AssertionError` with no message. Since `TruncatedGamma` now exists, either dispatch `Gamma` from `TruncatedDistribution` to the new classes, or give that assert a message pointing at `TruncatedGamma`.

## Nits

- Broken cross-reference in `RightTruncatedGamma`'s docstring: `` :class:`~numpyro.distributions.\ncontinuous.Beta` `` — reST normalises the newline to a space, so the target becomes `numpyro.distributions. continuous.Beta` and the link won't resolve. The same construction appears in `TwoSidedTruncatedGamma`.
- `test_truncated_gamma_far_tail_normalizer`'s final assertion (`abs(naive - exact) / exact > 1e-10`) is testing scipy's f64 accuracy, not our code. For `(20, 25)` the margin is 1.57e-9 — only ~15x — so it's fragile against a scipy implementation change. I'd drop it; the `assert_allclose` above it already carries the test's real content.
- `reparametrized_params` is `["low", "high"]`, but sampling is pure inverse-CDF and I confirmed `jax.grad` w.r.t. `concentration` through both `icdf` and `sample` is finite and correct (0.344 and 0.580 for the doctest case). Adding `concentration`/`rate` would enable pathwise gradients in SVI. It'd be a departure from how the existing truncated classes are set up, though, so I'd take it or leave it depending on what @fehiepsi prefers.
- `gammainccinv` has no direct unit test — consistent with `gammaincinv`, which doesn't either, so purely optional. A round-trip against `gammaincc` in `test_distributions_util.py` would be cheap.
- In `batch_util.py` the new registrations go two-sided → left → right; the file's existing order is left → right → two-sided. Cosmetic.

## One question

Have you checked how these behave under `TransformReparam` / `numpyro.infer.reparam`, and whether `Gamma`'s `KL` registrations need a companion? Not blocking, but it'd be good to know the gap is intentional.

juanitorduz (approved):
Thanks! @deeb01

PR #2258, opened 2026-08-28
Add regression test for render_model with masked observations (#2254)

### LINKED ISSUE #2254: jax > 0.10 introduces a lengt mismatch Assertion error on render_model call
### Bug Description

Bug: `numpyro.render_model()` crashes with "AssertionError: length mismatch"
when the model uses `Distribution.mask()` for handling missing observations.

Root cause: JAX 0.11 changed internal jaxpr representation such that
boolean masks appear in jaxpr.invars but not in the numpyro trace,
causing a length mismatch in numpyro/ops/provenance.py::track_deps_jaxpr.

### Steps to Reproduce

Minimal reproducible example for numpyro render_model bug with JAX >= 0.11.

Reproduced with:
    - numpyro 0.21.0
    - jax >= 0.11 (bug NOT present with jax 0.10.x)

Expected error:
    AssertionError: length mismatch: [N, M]
    at numpyro/ops/provenance.py:22 in _safe_map

```py

import jax.numpy as jnp
import numpyro
from numpyro.distributions import LogNormal, Normal


def model(obs, mask):
    """Simple linear regression with masked observations."""
    alpha = numpyro.sample("alpha", Normal(0, 1))
    beta = numpyro.sample("beta", Normal(0, 1))

    mu = alpha + beta * jnp.arange(len(obs), dtype=jnp.float32)

    numpyro.sample(
        "obs",
        Normal(mu, 1).mask(mask),  # <-- Using .mask() triggers the bug
        obs=obs,
    )


if __name__ == "__main__":
    import jax
    import numpyro

    print(f"JAX version: {jax.__version__}")
    print(f"NumPyro version: {numpyro.__version__}")

    # Synthetic data with one missing value (mask=False at index 2)
    obs = jnp.array([1.0, 2.0, jnp.nan, 4.0, 5.0])
    mask = jnp.array([True, True, False, True, True])

    # This works fine:
    print("\nTracing model (should work)...")
        trace = numpyro.handlers.trace(model).get_trace(obs=obs, mask=mask)
    print("Trace succeeded.")

    # This crashes with JAX >= 0.11:
    print("\nCalling render_model (crashes with JAX >= 0.11)...")
    try:
        graph = numpyro.render_model(
            model,
            model_args=(obs, mask),
            render_distributions=False,
        )
        print("render_model succeeded (no bug detected).")
    except AssertionError as e:
        print(f"BUG REPRODUCED: AssertionError: {e}")
    except Exception as e:
        print(f"Unexpected error: {type(e).__name__}: {e}")
```

Output

```
JAX version: 0.11.1
NumPyro version: 0.21.0

Tracing model (should work)...
Trace succeeded.

Calling render_model (crashes with JAX >= 0.11)...
BUG REPRODUCED: AssertionError: length mismatch: [5, 3]
```

### Expected Behavior

Model rendering without errors

Agi-Asi:
Adds a regression test for #2254 (`render_model` / `get_model_relations` crashing with `AssertionError: length mismatch` in `track_deps_jaxpr` when the model uses `Distribution.mask()` on JAX >= 0.11).

**Empirical verification** (answering @juanitorduz's question on the issue):

| numpyro | jax | result |
|---|---|---|
| 0.21.0 (released) | 0.11.1 | ❌ `AssertionError: length mismatch: [4, 3]` at `numpyro/ops/provenance.py:22` |
| master (0059d3a) | 0.11.1 | ✅ renders fine |

So the underlying bug was indeed fixed by #2245. However, #2245's test (`test/ops/test_provenance.py`) covers closed-over constants generically — nothing exercises the `Distribution.mask()` path from the original report, which is the user-facing way to hit it. This PR adds that test to `test/test_model_rendering.py` so the scenario stays covered.

The test also asserts the derived relations (`observed`, `sample_dist`, dependencies) so it validates behaviour rather than just "does not crash".

Closes #2254

### DIFF
--- a/test/test_model_rendering.py
+++ b/test/test_model_rendering.py
@@ -183,3 +183,28 @@ def test_param_with_lambda_and_sample():
     assert "x" in relations["sample_dist"]
     assert relations["sample_dist"]["x"] == "CategoricalProbs"
     assert "p" in relations["sample_param"]["x"]
+
+
+def test_get_model_relations_masked_observations():
+    """
+    Regression test for https://github.com/pyro-ppl/numpyro/issues/2254.
+
+    On JAX >= 0.11, boolean masks passed through Distribution.mask() appear
+    as closed-over constants in the jaxpr, which used to crash provenance
+    tracking with "AssertionError: length mismatch" in track_deps_jaxpr.
+    """
+
+    def model(obs, mask):
+        alpha = numpyro.sample("alpha", dist.Normal(0.0, 1.0))
+        sigma = numpyro.sample("sigma", dist.LogNormal(0.0, 1.0))
+        with numpyro.plate("data", obs.shape[0]):
+            numpyro.sample("obs", dist.Normal(alpha, sigma).mask(mask), obs=obs)
+
+    obs = jnp.array([1.0, 2.0, jnp.nan, 4.0])
+    mask = jnp.array([True, True, False, True])
+
+    relations = get_model_relations(model, model_args=(obs, mask))
+
+    assert relations["observed"] == ["obs"]
+    assert relations["sample_dist"]["obs"] == "Normal"
+    assert set(relations["sample_sample"]["obs"]) == {"alpha", "sigma"}


Qazalbash (approved):
Thanks @Agi-Asi

PR #2259, opened 2026-08-28
Rename Predictive's exclude_deterministic to condition_deterministic

### LINKED ISSUE #2086: Predictive.exclude_deterministic does not filter deterministic sites
### Bug Description

The `Predictive` helper class attribute `exclude_deterministic` does not successfully exclude deterministic sites from resulting posterior predictive dictionary.

### Steps to Reproduce

```bash
uv run test_exclude_deterministic.py
```

where `exclude_deterministic.py` contains
```python
#!/usr/bin/env -S uv run
# /// script
# requires-python = ">=3.9"
# dependencies = [
#     "jax",
#     "jaxlib",
#     "numpyro",
# ]
# ///
"""
Test to verify if exclude_deterministic works in numpyro.infer.util.Predictive
"""

import jax
import jax.numpy as jnp

import numpyro


def model(x=None, y=None):
    """Simple model with both stochastic and deterministic sites"""
    # Stochastic sites
    a = numpyro.sample("a", dist.Normal(0, 1))
    b = numpyro.sample("b", dist.Normal(0, 1))

    # Deterministic site
    linear_combination = numpyro.deterministic("linear_combination", a + 2 * b)

    # Another deterministic site
    _ = numpyro.deterministic("squared", a**2)

    # Likelihood
    if y is not None:
        sigma = numpyro.sample("sigma", dist.Exponential(1.0))
        numpyro.sample("obs", dist.Normal(linear_combination, sigma), obs=y)


# Generate some synthetic data
key = jax.random.PRNGKey(0)
n_samples = 100
x_data = jnp.linspace(0, 1, n_samples)
true_a = 1.5
true_b = 2.0
y_data = true_a + 2 * true_b + jax.random.normal(key, (n_samples,)) * 0.5

# Run MCMC
mcmc = MCMC(NUTS(model), num_warmup=500, num_samples=1000)
mcmc.run(key, y=y_data)
samples = mcmc.get_samples()

# Test Predictive with exclude_deterministic=True
print("\nTest 1: Predictive with exclude_deterministic=True")
print("-" * 70)
predictive_exclude = Predictive(model, samples, exclude_deterministic=True)
predictions_exclude = predictive_exclude(key)
print("Keys with exclude_deterministic=True:", predictions_exclude.keys())
print("Expected: Should NOT include 'linear_combination' or 'squared'")

# Test Predictive with exclude_deterministic=False (default)
print("\nTest 2: Predictive with exclude_deterministic=False")
print("-" * 70)
predictive_include = Predictive(model, samples, exclude_deterministic=False)
predictions_include = predictive_include(key)
print("Keys with exclude_deterministic=False:", predictions_include.keys())
print("Expected: SHOULD include 'linear_combination' and 'squared'")
```

### Expected Behavior

The "squared" and "linear_combination" sites should only be present in the returned value of `predictive_include()`.

### Notes

Obviously the fix here will be correct but breaking. Unless I'm missing something. Happy to work on a fix, but perhaps there's some context I'm missing.

Agi-Asi:
Implements the rename discussed in #2086: `Predictive`'s `exclude_deterministic` → **`condition_deterministic`** (name suggested by @kylejcaron, endorsed by @juanitorduz).

**Why**: the old name reads as if it filters deterministic sites out of the *returned* dictionary, but it actually controls whether deterministic sites in `posterior_samples` are *conditioned on during substitution* — the returned sites are governed by `return_sites`. That mismatch is exactly what confused the issue reporter (and, per the thread, the argument's own author).

**What this PR does:**
- New keyword `condition_deterministic: bool = False` — the default is behaviour-identical to the old `exclude_deterministic=True` default, so nothing changes for existing users.
- **Full backward compatibility**: `exclude_deterministic=` is still accepted (mapped to `not condition_deterministic`) with a `FutureWarning`; passing both raises `ValueError`. The `predictive.exclude_deterministic` *attribute* is kept as a deprecated property (getter + setter) so code that flips it after construction keeps working.
- Docstring rewritten to say what the flag actually does, point at `return_sites` for output filtering, and explain why conditioning on deterministic sites is off by default (shape/staleness bugs when predicting on new data — see #1772).
- Tests: deprecation warnings for kwarg/getter/setter, both-kwargs `ValueError`, default value, and the existing discrete-predictive test migrated to the new name.

Note: `MCMC.print_summary(exclude_deterministic=...)` is a different API where the name *is* accurate (it filters printed output), so it is deliberately untouched.

`ruff check`, `ruff format --check`, and `ty check` all pass; relevant test selection (18 predictive tests) passes locally.

Closes #2086

### DIFF
--- a/numpyro/infer/util.py
+++ b/numpyro/infer/util.py
@@ -970,7 +970,18 @@ class Predictive(object):
 
         + set `batch_ndims=1` to get predictions from a one dimensional batch of the guide and parameters
           with shapes `(num_samples x batch_size x ...)`
-    :param exclude_deterministic: indicates whether to ignore deterministic sites from the posterior samples.
+    :param condition_deterministic: if True, deterministic sites present in
+        `posterior_samples` are conditioned on (substituted into the model)
+        when predicting; if False (default), they are ignored and instead
+        re-computed from their parent sites. Note that this argument controls
+        how deterministic sites are *handled during substitution*, not whether
+        they appear in the returned dictionary — use `return_sites` to control
+        which sites are returned. Conditioning on deterministic sites can
+        produce wrong shapes or stale values when predicting on new data,
+        which is why it is disabled by default.
+    :param exclude_deterministic: deprecated alias, use
+        ``condition_deterministic`` instead (``exclude_deterministic=True``
+        corresponds to ``condition_deterministic=False``).
 
     :return: dict of samples from the predictive distribution.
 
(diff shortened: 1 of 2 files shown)


juanitorduz (approved):
I like the suggestion :) It is clearer

fehiepsi (approved):

PR #2260, opened 2026-08-29
Support batched scale_tril in LowerCholeskyAffine

castor639:
Lift the 2D-only restriction so loc and scale_tril can carry leading batch dimensions with standard JAX broadcasting. Inverse uses batched triangular solve so shared and fully batched scale_tril both work.


Qazalbash (approved):

fehiepsi (approved):
Thanks for the enhancement!

PR #2261, opened 2026-09-02
doc: link citation markers on the Contributed Code page

IvanDobrovolsky:
The page cites references as plain text [1], [2], … and both the SteinVI and HSGP sections number their own list from [1], so a marker doesn't tell you which list it belongs to.

Now each reference entry has an anchor and the markers link to it. I diffed the built `contrib.html` against a master build to make sure nothing else moved: the text comes out the same; the only differences are the anchors and the links themselves, plus a couple of line breaks.

I used plain reStructuredText links rather than the :ref: role so the markers don't show up as raw markup when you view the file on GitHub. The jump doesn't actually work there, though; GitHub renames the ids to `user-content-*`. `make docs` builds cleanly.

Closes #2079

PR #2264, opened 2026-09-08
doc: link citation markers in the contrib API docstrings

IvanDobrovolsky:
Follow-up to #2261, which only addressed the text at the top of the Contributed Code page. This one addresses the docstrings underneath it. 

Anchors are named after the object, and the SteinVI class uses `steinvi-class-` since the SteinVI section already took `steinvi-ref-*`.

Verified the built docs page against master; only links are clickable now; everything else is untouched.


juanitorduz (approved):
Thanks @IvanDobrovolsky

PR #2265, opened 2026-09-10
docs(gh-2187): add mathematical details for Pareto and Weibull

Rokesh2008:
Related to #2187\n\nThis contribution documents two unchecked distributions from the distribution documentation tracker:\n\n- `Pareto`: support, PDF, CDF, transformed-exponential sampling, and moment existence conditions.\n- `Weibull`: support, PDF, CDF, sampling implementation, hazard-rate interpretation, and moments.\n\nThe documentation follows the existing NumPyro distribution docstring style and keeps the implementation unchanged.\n\nValidation:\n- Python AST parsing of `numpyro/distributions/continuous.py` passed.\n- Confirmed both new class docstrings contain mathematical documentation.\n- Full `make lint`, `make test`, and `make doctest` were not available in the execution sandbox because the project dependencies/build environment are not installed.\n\nRelated issue: #2187. No dependencies added.

### DIFF
--- a/numpyro/distributions/continuous.py
+++ b/numpyro/distributions/continuous.py
@@ -3343,7 +3343,28 @@ class Normal(Distribution):
         )
 
 
-class Pareto(TransformedDistribution):
+ class Pareto(TransformedDistribution):
+    r"""Pareto distribution.
+
+    Supported on x >= scale with positive tail-shape alpha. The density is
+
+    .. math::
+
+       f(x) = alpha * scale**alpha / x**(alpha + 1),
+
+    and the CDF is 1 - (scale / x)**alpha. NumPyro implements the
+    distribution as Exponential(alpha), followed by exponential and scale
+    transforms, so samples are scale * exp(E) for E ~ Exponential(alpha).
+
+    The mean exists for alpha > 1 and equals alpha * scale / (alpha - 1).
+    The variance exists for alpha > 2 and equals
+    alpha * scale**2 / ((alpha - 1)**2 * (alpha - 2)).
+
+    :param scale: Minimum value; must be positive.
+    :param alpha: Tail-shape parameter; must be positive.
+    :param validate_args: Whether to validate input constraints, defaults to None.
+    """
+
     arg_constraints = {"scale": constraints.positive, "alpha": constraints.positive}
     reparametrized_params = ["scale", "alpha"]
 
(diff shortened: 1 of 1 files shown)


juanitorduz:
Thanks for picking these up! The math content is correct (I checked the PDF/CDF, the sampling construction, and the moment formulas against the implementation), but the docstrings need to be converted to LaTeX before this can be merged.

### Blocking

**1. The `.. math::` bodies are Python pseudo-code, not LaTeX.** Sphinx passes the directive body verbatim to MathJax, so `alpha * scale**alpha / x**(alpha + 1)` renders as the italic letters `a l p h a`, a literal `∗∗`, etc. The same applies to the inline expressions written as plain prose (`1 - (scale / x)**alpha`, `lambda * Gamma(1 + 1/k)`, `scale**2 / ...`). The docstrings already merged from #2187 (`Exponential`, `Gumbel`, `Gamma`, `Normal`, ...) use `\frac`, `\alpha`, `\lambda`, `\exp` and `:math:` inline roles; please follow that convention.

**2. Pareto support is stated as `x >= scale`, but the implementation is strict.** `Pareto.support` is `constraints.greater_than(self.scale)`, and `_GreaterThan` is strict (`_GreaterThanEq` is the non-strict variant). With `validate_args=True`, `log_prob(scale)` is rejected, so the docstring should say :math:`x > x_m`.

### Nits

- `r"""` is used but there are no backslashes yet; it becomes necessary once the math is LaTeX, so keep it.

### Suggested replacement

I verified these render correctly.

```python
    r"""Pareto distribution parameterized by scale (:attr:`scale`) and tail
    index (:attr:`alpha`).

    The probability density function (PDF) is defined as:

    .. math::
       f(x; x_m, \alpha) = \frac{\alpha x_m^{\alpha}}{x^{\alpha + 1}}

    where :math:`x > x_m`, :math:`x_m > 0` is the scale and :math:`\alpha > 0`
    is the shape. The cumulative distribution function (CDF) is
    :math:`F(x) = 1 - (x_m / x)^{\alpha}`.

    The distribution is implemented as a :class:`TransformedDistribution`:
    :math:`X = x_m e^{E}` with :math:`E \sim \mathrm{Exponential}(\alpha)`,
    i.e. an :class:`Exponential` base distribution followed by
    :class:`~numpyro.distributions.transforms.ExpTransform` and
    :class:`~numpyro.distributions.transforms.AffineTransform`.

    The mean :math:`\alpha x_m / (\alpha - 1)` is finite only for
    :math:`\alpha > 1` and the variance
    :math:`\alpha x_m^2 / ((\alpha - 1)^2 (\alpha - 2))` only for
    :math:`\alpha > 2`; both properties return ``inf`` otherwise.

    :param scale: Scale parameter (:math:`x_m`), the lower bound of the support.
    :type scale: ArrayLike
    :param alpha: Shape (tail index) parameter (:math:`\alpha`).
    :type alpha: ArrayLike
    :param validate_args: Whether to validate input constraints, defaults to None.
    :type validate_args: bool, optional
    """
```

```python
class Weibull(Distribution):
    r"""Weibull distribution parameterized by scale (:attr:`scale`) and shape
    (:attr:`concentration`).

    The probability density function (PDF) is defined as:

    .. math::
       f(x; \lambda, k) = \frac{k}{\lambda}
       \left(\frac{x}{\lambda}\right)^{k - 1}
       \exp\left(-\left(\frac{x}{\lambda}\right)^{k}\right)

    where :math:`x > 0`, :math:`\lambda > 0` is the scale and :math:`k > 0` is
    the shape. The cumulative distribution function (CDF) is
    :math:`F(x) = 1 - \exp(-(x / \lambda)^{k})`.

    For :math:`k = 1` this reduces to :math:`\mathrm{Exponential}(1 / \lambda)`.
    The hazard rate :math:`h(x) = (k / \lambda)(x / \lambda)^{k - 1}` is
    decreasing for :math:`k < 1` and increasing for :math:`k > 1`.

    The mean is :math:`\lambda \, \Gamma(1 + 1/k)` and the variance is
    :math:`\lambda^2 \left[\Gamma(1 + 2/k) - \Gamma(1 + 1/k)^2\right]`.

    Samples are drawn with :func:`jax.random.weibull_min`.

    :param scale: Scale parameter (:math:`\lambda`).
    :type scale: ArrayLike
    :param concentration: Shape parameter (:math:`k`).
    :type concentration: ArrayLike
    :param validate_args: Whether to validate input constraints, defaults to None.
    :type validate_args: bool, optional
    """
```

Once updated, `make docs` (or at least `make lint`) locally would be a good check that Sphinx does not emit warnings for the new blocks.

juanitorduz (changes requested):
Small changes (see above)

Rokesh2008:
Implemented the requested documentation fixes in commit `51e1cd2`: converted the Pareto and Weibull math to LaTeX/Sphinx syntax and corrected the Pareto support to the strict `x > x_m` boundary. Thanks for the detailed review.

juanitorduz (approved):
Thanks!

PR #2266, opened 2026-09-10
doc(gh-2187): mathematical details for GaussianRandomWalk

jabrailkhalil:
## Summary

Adds mathematical explanations and implementation details for `GaussianRandomWalk` to the class, `sample`, `log_prob`, `mean`, and `variance` docstrings, as tracked in gh-2187.

- Class: joint density of the random walk over `R^n`, defined via initial position and independent Normal increments.
- `sample`: cumulative sum of standard Normal draws.
- `log_prob`: sum of univariate Normal log densities over the initial position and increments.
- `mean` and `variance`: `E[X_t] = 0` and `Var(X_t) = (t+1) * scale^2`, matching the implementation.

Math details were checked numerically against the implementation (log_prob, mean, variance, batched scale shapes).

## Testing

- `ruff check` and `ruff format --check` on the changed file: pass.
- `ty check` and `codespell`: pass.
- `pytest test/test_distributions.py -k "random_walk or RandomWalk"`: 62 passed, 42 skipped (missing optional deps), 2 xfailed, on local CPU before opening the PR.

### DIFF
--- a/numpyro/distributions/continuous.py
+++ b/numpyro/distributions/continuous.py
@@ -1179,6 +1179,41 @@ class GaussianStateSpace(TransformedDistribution):
 
 
 class GaussianRandomWalk(Distribution):
+    r"""Gaussian random walk distribution parameterized by scale (:attr:`scale`)
+    and the number of steps (:attr:`num_steps`).
+
+    A Gaussian random walk :math:`\mathbf{X} = (X_0, \ldots, X_{n-1})` of
+    length :math:`n = ` :attr:`num_steps` starts at :math:`X_0` and evolves
+    through independent Normal increments:
+
+    .. math::
+
+        X_0 \sim \mathcal{N}(0, \sigma^2),
+        \qquad X_t - X_{t-1} \sim \mathcal{N}(0, \sigma^2), \quad
+        t = 1, \ldots, n - 1,
+
+    where :math:`\sigma > 0` is the scale. The joint probability density
+    function over :math:`\mathbb{R}^n` is
+
+    .. math::
+
+        p(\mathbf{x}) = f(x_0; 0, \sigma^2)
+        \prod_{t=1}^{n-1} f(x_t - x_{t-1}; 0, \sigma^2),
+
+    where :math:`f(\cdot; \mu, \sigma^2)` denotes the univariate Normal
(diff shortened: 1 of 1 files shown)


juanitorduz:
Thanks for the PR! The math checks out against the implementation (I verified `log_prob`, `variance`, and the batched `scale` case numerically). Two changes requested before merging:

**1. Broken inline markup in the class docstring (line 1186)**

```rst
length :math:`n = ` :attr:`num_steps` starts at :math:`X_0` and evolves
```

RST requires the inline end-string to be immediately preceded by non-whitespace, so ``:math:`n = ` `` does not close where intended. The `:math:` role swallows through the next backtick and the math node content becomes literally ``n = ` :attr:`num_steps``, which MathJax will render as garbage. `make docs` does not warn about this, so CI stays green. You can confirm with docutils:

```python
import docutils.core
text = r"length :math:`n = ` :attr:`num_steps` starts at :math:`X_0`"
print(docutils.core.publish_string(text, writer_name="pseudoxml").decode())
```

Suggested fix:

```rst
length :math:`n` (given by :attr:`num_steps`) starts at :math:`X_0` and evolves
```

**2. `:rtype: ArrayLike` in `sample` (line 1255) and `log_prob` (line 1277)**

Every other method docstring in `continuous.py` uses `:rtype: jax.Array` (e.g. `Normal.sample`, `Normal.log_prob`), and both methods are annotated `-> Array`. Please change both to `:rtype: jax.Array`.

**Optional nits**

- `f(\cdot; \mu, \sigma^2)` parameterizes the Normal density by variance, while the `Normal` class docstring in the same file uses `f(x; \mu, \sigma)` with `\sigma` as the standard deviation. Using `f(\cdot; 0, \sigma)` would keep the notation consistent across the two classes.
- `mean` / `variance` say "Calculates the mean of the Gaussian random walk." whereas the rest of the file uses "Calculates the analytical mean." Purely cosmetic.

juanitorduz (changes requested):
Smal changes (see above)

jabrailkhalil:
Thanks @juanitorduz! I have addressed both requested changes (fixed the inline RST markup and updated the :rtype: annotations to jax.Array), along with the optional notation alignment to Normal.

juanitorduz (approved):
thanks!

PR #2267, opened 2026-09-10
doc(gh-2187): docstrings for LogNormal and LogUniform

IvanDobrovolsky:
Adding `LogNormal` and `LogUniform`. Same style as Logistic: description, PDF, and the closed-form expressions for mean, variance, and entropy.

Issue https://github.com/pyro-ppl/numpyro/issues/2187

`LogNormal` before vs after:
image
image

`LogUniform` before vs after:
 image
 image


juanitorduz (approved):
Thanks @IvanDobrovolsky !

PR #2269, opened 2026-09-14
doc(gh-2187): mathematical details for InverseGamma and Kumaraswamy

Yi-111-a:
## Summary
- Docstrings for `InverseGamma` and `Kumaraswamy` in the same style as Logistic/LogNormal: description, PDF, parameters, moments (and entropy for InverseGamma).
- Keeps and clarifies the InverseGamma `rate` = Wikipedia-*scale* caveat (NumPyro/Pyro naming).
- Kumaraswamy documents PDF, inverse-CDF sampling (including the `U` vs `1-U` equivalence used in code), and beta-function mean/variance.

Towards https://github.com/pyro-ppl/numpyro/issues/2187

## Self-review
- InverseGamma PDF/mean/var/entropy match `Gamma` + `PowerTransform(-1)` and the implemented formulas with `rate` as Wikipedia scale.
- Kumaraswamy `log_prob` / mean / variance match `concentration1=a`, `concentration0=b`.

## Test plan
- [ ] Docs / doctest CI
- [ ] Spot-check rendered InverseGamma and Kumaraswamy pages


numpyro/distributions/continuous.py
```diff
@@ -1457,10 +1457,35 @@ def variance(self) -> Array:
...
-    """
-    .. note:: We keep the same notation `rate` as in Pyro but
-        it plays the role of scale parameter of InverseGamma in literatures
-        (e.g. wikipedia: https://en.wikipedia.org/wiki/Inverse-gamma_distribution)
+    r"""The Inverse-Gamma distribution, a continuous distribution on the
+    positive reals parameterized by concentration :math:`\alpha > 0` and
+    a positive parameter called ``rate`` in this library. It is the
+    distribution of :math:`1/Y` where
+    :math:`Y \sim \mathrm{Gamma}(\alpha, \lambda)` with the same
+    concentration and rate, and is implemented as a
```
juanitorduz: The intro uses `\lambda` for the Gamma rate, but every formula below uses `\beta` for the same parameter. Suggest using `\beta` here too, e.g. `Y \sim \mathrm{Gamma}(\alpha, \beta)` with rate `\beta`, so the reader does not have to reconcile two symbols for one attribute.

numpyro/distributions/continuous.py
```diff
@@ -1486,17 +1518,46 @@ def __init__(
 
     @property
     def mean(self) -> Array:
+        r"""Mean of the Inverse-Gamma distribution:
+
+        .. math::
+            \mathbb{E}[X] = \frac{\beta}{\alpha - 1} \quad (\alpha > 1)
+
+        For :math:`\alpha \le 1` the mean is infinite (returned as ``inf``).
+        Here :math:`\beta` is the NumPyro ``rate`` parameter (Wikipedia scale).
```
juanitorduz: Nit: the "NumPyro `rate` parameter (Wikipedia scale)" clarification is repeated in `__init__`, `mean`, `variance` and `entropy`. The class-level note already defines `\beta` as `rate`; keeping the clarification in the note and `__init__` only would reduce noise in the rendered page (same applies to the `variance` and `entropy` docstrings below).

numpyro/distributions/continuous.py
```diff
@@ -1693,6 +1754,23 @@ def icdf(self, q: ArrayLike) -> Array:
 
 
 class Kumaraswamy(Distribution):
+    r"""The Kumaraswamy distribution, a continuous distribution on the unit
+    interval parameterized by two positive shape parameters
+    :math:`a` (:attr:`concentration1`) and :math:`b`
+    (:attr:`concentration0`). It is a Beta-like family with a simple closed
+    form PDF and CDF, often used as a computationally convenient alternative
+    to the Beta distribution.
```
juanitorduz, line 1759: The text says "simple closed form PDF and CDF" but only the PDF is given. Adding the CDF

```rst
.. math::
    F(x \mid a, b) = 1 - (1 - x^{a})^{b}, \quad x \in (0, 1)
```

would also make the inverse-CDF sampling description in `sample` self-contained.

numpyro/distributions/continuous.py
```diff
@@ -1693,6 +1754,23 @@ def icdf(self, q: ArrayLike) -> Array:
...
+
+    The Probability Density Function (PDF) is:
+
+    .. math::
+        f(x \mid a, b) = a\, b\, x^{a-1}\,(1 - x^{a})^{b-1},
+        \quad x \in (0, 1)
+
+    where :math:`a > 0` is (:attr:`concentration1`) and :math:`b > 0` is
+    (:attr:`concentration0`).
```
juanitorduz: This reads as a broken sentence ("where `a > 0` is (concentration1) and `b > 0` is (concentration0)"). Suggest:

```rst
where :math:`a > 0` is the first shape parameter (:attr:`concentration1`)
and :math:`b > 0` is the second shape parameter (:attr:`concentration0`).
```

numpyro/distributions/continuous.py
```diff
@@ -1736,6 +1834,17 @@ def sample(
 
     @validate_sample
     def log_prob(self, value: ArrayLike) -> Array:
+        r"""Evaluate the log probability density function at ``value``:
+
+        .. math::
+            \ln f(x \mid a, b) =
+            \log a + \log b
+            + (a - 1)\log x
+            + (b - 1)\log(1 - x^{a})
```
juanitorduz: Nit: this equation mixes `\ln f` on the left with `\log a + \log b + \ldots` on the right. `Logistic` / `LogNormal` use `\ln` throughout; pick one for consistency.

juanitorduz (commented):
Thanks for this! The math checks out: I verified on this branch that `exp(log_prob)` matches the documented PDFs for both distributions, that the InverseGamma mean/variance/entropy and Kumaraswamy mean/variance closed forms match both the implementation and numerical integration (plus a 200k-sample Monte Carlo check for Kumaraswamy), and that the `U` vs `1 - U` sampling remark is correct since `F^{-1}(u) = (1 - (1 - u)^{1/b})^{1/a}`. The docstring layout also matches the `LogNormal` / `Logistic` precedent.

Only a few wording / consistency nits inline (symbol consistency for the Gamma rate, a broken sentence in the Kumaraswamy parameter description, missing CDF that the text alludes to, `\ln` vs `\log`, and some repetition of the `rate` = Wikipedia scale caveat). Once those are addressed this should be good to go.

Yi-111-a:
Thanks @juanitorduz — addressed the nits: `\beta` for the Gamma rate in the InverseGamma intro, trimmed the repeated rate/Wikipedia clarifications (kept class note + `__init__` only), added the Kumaraswamy CDF, fixed the concentration1/0 sentence, and switched the Kumaraswamy `log_prob` (and InverseGamma entropy) to `\ln` throughout for consistency with Logistic/LogNormal.

juanitorduz (approved):
Thanks for the quick turnaround! All nits addressed in 75d92215: `\beta` in the InverseGamma intro, the rate/Wikipedia-scale caveat now lives only in the class note and `__init__`, the Kumaraswamy CDF is documented, the parameter sentence reads correctly, and `\ln` is used consistently in `log_prob` and `entropy`. CI is green. LGTM.

PR #2270, opened 2026-09-14
doc(gh-2187): mathematical details for VonMises

Yi-111-a:
## Summary
- Math docstrings for `VonMises` matching Logistic/LogNormal style: PDF, log-density with exponentially scaled Bessel \(I_0^e\), circular mean/variance, and parameters.
- Preserves the existing `CircularReparam` sampling guidance.

Towards https://github.com/pyro-ppl/numpyro/issues/2187

## Self-review
- Documented log-density matches `i0e`-based implementation:
  \(\kappa(\cos(\cdot)-1)-\log(2\pi)-\log I_0^e(\kappa) \equiv \kappa\cos(\cdot)-\log(2\pi)-\log I_0(\kappa)\).
- Circular variance formula matches `1 - i1e(κ)/i0e(κ)`.

## Test plan
- [ ] Docs / doctest CI
- [ ] Spot-check rendered VonMises page


juanitorduz (commented):
Thanks for the PR! I verified the math on the branch and it all checks out:

- `log_prob` vs `scipy.stats.vonmises.logpdf`: max abs diff 8e-7 (float32).
- `variance` vs `1 - I1(κ)/I0(κ)` from `scipy.special`: matches.
- `mean` wrap `((μ+π) mod 2π) - π` matches.
- The `i0e` identity in the log-density is exact since `I0^e(κ) = I0(κ) e^{-κ}`.
- RST parses cleanly for all six docstrings, `ruff check` / `ruff format --check` are clean, and `pytest test/test_distributions.py -k VonMises` passes (130 passed).

Two small nits before merging:

1. **Support notation.** The PDF states `x \in [-\pi, \pi)`, but `support = constraints.circular` is the closed interval `[-\pi, \pi]`, and `log_prob` wraps `value` modulo `2\pi`, so the density is valid for any real `x`. Please use `x \in [-\pi, \pi]` (optionally noting the density is `2\pi`-periodic). Same for the `:return:` of `sample`: the sampler returns values in `[-\pi, \pi]` after `(x + \pi) \bmod 2\pi - \pi`, and the endpoint `-\pi` is reachable.

2. **`loc` parameter description.** Since `loc` is `constraints.real` but the distribution is invariant to shifts of `2\pi`, it would help to mention in the `:param loc:` line that only `\mu \bmod 2\pi` matters (the `mean` property already documents the wrapping).

Once those are in, this looks good to go.

Yi-111-a:
Thanks @juanitorduz — updated support to the closed interval \(x \in [-\pi, \pi]\) (consistent with `constraints.circular`), noted \(2\pi\)-periodicity / wrapping in `log_prob` and `sample`, and clarified that only \(\mu \bmod 2\pi\) matters for `loc`.

juanitorduz (approved):
Thanks for the quick follow-up! Both nits are addressed in 23e7f6e3: closed support `[-\pi, \pi]` in the PDF, `sample`, and `mean` (with the periodicity / wrapping note and the cross-ref to `constraints.circular`), and the `loc` parameter now notes that only `\mu \bmod 2\pi` matters. Re-checked `ruff` and RST parsing on the new head; all clean. LGTM.

PR #2272, opened 2026-09-14
refactor: keep truncation bounds out of the static treedef for the power-law distributions

deeb01:
`DoublyTruncatedPowerLaw` and `LowerTruncatedPowerLaw` declare `_support` in
`pytree_aux_fields`, but build it from parameter values in `__init__`. Aux fields go into
the static treedef, so the truncation bounds get baked in and a jitted function taking the
distribution as an argument retraces for every distinct bound — 5 traces for 5 bounds,
against 1 for `TwoSidedTruncatedDistribution` and `TwoSidedTruncatedGamma`, which declare
`_support` as a data field. These two are the only distributions holding a
parameter-derived value in `pytree_aux_fields`; every other entry is genuinely static.

Moves `_support` to `pytree_data_fields` on both, and adds a regression test that fails on
master with `assert 5 == 1`.


Qazalbash (approved):
Thanks @deeb01. I remember these two classes were implemented by me and @ InfinityMod in #1807, when I was learning NumPyro and JAX.

Issue #2271, opened 2026-09-14
Easier Use of Distributions Under `vmap`


DanWaxman:
Distributions carry some static metadata (e.g., the batch shape), that make it a bit annoying to use in some other contexts. For example, they sometimes cause issues in [dynestyx](github.com/BasisResearch/dynestyx), where we use `numpyro` distributions quite generally. 

An example of something I'd like to work, but doesn't, is just accessing variables of `vmap`ed distributions:
```python
def foo(mean):
    return dist.Normal(mean, jnp.ones_like(mean))

jax.vmap(foo)(jnp.zeros((2,2))).mean
```

Speaking offline with @fehiepsi, we agreed it would be desirable for such things to "just work." Looking back at previous discussions, however, it seems this is expected behavior under the current implementation (cf. https://github.com/pyro-ppl/numpyro/pull/1529#issuecomment-1411377292 or https://github.com/pyro-ppl/numpyro/issues/1684). Nevertheless, it would be nice to have a more extendable way of working with such things than imposing lots of `numpyro`-specific helpers on downstream users.

One way that might be possible is to try and defer shape evaluations when they may be inferred from parameter shapes (e.g., in all the scalar cases), in which case they become a lazy `property`. I had a quick vibe-coded go at this and it seems possible (see https://github.com/DanWaxman/numpyro/tree/dw-exploring-vmap-dists), though I haven't taken the time to go through that code very carefully (in particular, things that happen around expanded and transformed distributions), and some of the tests look wonky (in particular, I think Codex just changed `test_distributions/test_promote_batch_shape_shares_data_and_preserves_input` to pass ignoring all semantics haha), so please don't interpret it as a PR.

If that route seems interesting/acceptable, I'd be happy to take a closer/more careful look to see what it would take to do well and what it may break.

juanitorduz:
@DanWaxman This looks very reasonable :) I suggest we start with a PR and take it from there. WDYT @fehiepsi ?

fehiepsi:
Deferring shape construction makes a lot of sense to me. Turning it into a lazy property is even better. Thanks for looking at this!

DanWaxman:
Great! I'll work on this in the next few days!

PR #2273, opened 2026-09-17
[WIP] Add numpyro.ops.gaussian and GaussianHMM with parallel Kalman marginalization

juanitorduz:
Ports Pyro's information-form Gaussian factor algebra (`numpyro.ops.gaussian`) and adds `GaussianHMM` / `IndependentHMM` with O(log T) parallel marginalization, `filter`, `sample`, `sample_posterior`, `conjugate_update`, `prefix_condition`, `reshape_batch`.
Left-acting matrices as in `GaussianStateSpace`; shapes are leaf-derived and rank-aligned (compatible with #2271).
Verified against dense joint oracles, Pyro's factor identities, sampler moments, `GaussianStateSpace` moment identities, and a NUTS local-level smoke test; float32 by default with an x64 CI case.
Stacked: a follow-up PR adds `GammaGaussianHMM`, `LinearHMM`, `GaussianMRF`, `StudentTReparam` and `LinearHMMReparam`.


fehiepsi:
I'm seeing many local helpers are defined. It's hard to know the motivation for them. Are those what claude suggest?

juanitorduz:
Partially, but it's mainly on me as I do not like long functions where I can get lost with new implementations and ports. 

I'm happy to reduce them of course :)

juanitorduz:
I can also put them as statics methods inside the classes if that is more consistent :)

PR #2176, opened 2026-04-28
Allow zero rate for Poisson distribution.

tillahoffmann:
This PR uses `xlogy` to allow evaluation of the Poisson log likelihood for zero rates. The manual cast from integer `value` to dtype matching `self.rate` is required because `xlogy` defines a custom jvp that does not play well with integer arguments.

PR #2276, opened 2026-09-17
fix(distributions): keep Categorical entropy finite when a category has zero probability

Dev-next-gen:
While reading the entropy methods in `numpyro/distributions/discrete.py` I noticed that `CategoricalProbs.entropy` is the only one in that file that multiplies by a bare logarithm:

```python
probs = jnp.asarray(self.probs)
return -(probs * jnp.log(probs)).sum(axis=-1)
```

`probs` is constrained to the simplex, and a simplex point may contain an exact zero. When it does, that term is `0 * -inf = nan` and the whole sum becomes NaN, even though a category that never occurs contributes nothing to the entropy under the usual `0 log 0 = 0` convention:

```python
>>> dist.CategoricalProbs(jnp.array([0.25, 0.0, 0.75])).entropy()
Array(nan, dtype=float32)
>>> scipy.stats.entropy([0.25, 0.0, 0.75])
0.5623351446188083
```

`BernoulliProbs.entropy`, a few hundred lines above, already handles the same edge case with `xlogy`, and `Poisson.log_prob` was moved to `xlogy` for the zero-rate case in #2176. I applied the same treatment here; the change is one line and leaves every strictly positive `probs` vector bit-for-bit identical, since `xlogy(x, x) == x * log(x)` away from zero.

I added the case to the existing categorical entropy test, checked against `scipy.stats.entropy` because `scipy.stats.multinomial` cannot be used with a zero-probability category. On my machine (CPU jax 0.11.1) the new test fails on master with `ACTUAL: nan / DESIRED: 0.562335` and passes with the fix; `pytest test/test_distributions.py -k entropy` gives 327 passed, 826 skipped afterwards.

I deliberately left `CategoricalLogits.entropy` alone: it hits the same `0 * -inf` shape only when a logit is `-inf`, which is outside its declared `real_vector` constraint, so that felt like a separate decision rather than part of this fix.

Found by a defect-hunting pipeline I build and run ([Dev-next-gen](https://github.com/Dev-next-gen)), using Claude Code with Anthropic's Claude Opus 5.


fehiepsi (approved):
Thanks @Dev-next-gen !

Qazalbash (approved):

PR #2268, opened 2026-09-14
doc(gh-2187): mathematical details for StudentT

Yi-111-a:
Adds `StudentT` documentation for https://github.com/pyro-ppl/numpyro/issues/2187.

Same style as the recent Logistic / LogNormal docs: a short description of the location-scale construction, the PDF, `__init__` parameters, and the closed-form mean / variance / entropy. Mean and variance notes match the implementation (and SciPy): `inf` mean for `df <= 1`, `inf` variance for `1 < df <= 2`, `nan` variance for `df <= 1`.

No code or behavior changes.

PR #2277, opened 2026-09-18
doc(gh-2187): mathematical details for count-data distributions

Yi-111-a:
Adds math-flavored docstrings to four more distributions on the #2187 checklist, following the template from #2188, #2192, #2199 and the recent #2268–#2270 series.

- `MultinomialProbs` / `MultinomialLogits` — class PMF, `__init__` parameters, `sample`, `log_prob`, the `probs`/`logits` conversions, `mean`, `variance`, and `support`.
- `BetaBinomial` — class PMF plus `sample`, `log_prob`, `mean`, `variance`, and `support`.
- `DirichletMultinomial` — class PMF plus `sample`, `log_prob`, `mean`, `variance`, and `support`.

Docstrings only, no logic changes. `ruff check` and `ruff format --check` are clean on both files.


juanitorduz:
Thanks a lot for this, @Yi-111-a! This is a really nice addition to the #2187 series. I went through every formula against the implementation and they all check out :)

There are just two small factual fixes needed in the `sample` docstrings before we can merge:

1. **`MultinomialLogits.sample`** says the trials run "over `logits` directly", but the implementation actually passes `self.probs` (the softmax of the logits) to `multinomial`. Something like "...which runs the `total_count` trials over `probs`, the softmax of `logits`" would be accurate.

2. **`MultinomialProbs.sample` and `MultinomialLogits.sample`** both say that `numpyro.distributions.util.multinomial` relies on `jax.random.categorical`. It actually uses NumPyro's own `numpyro.distributions.util.categorical`, which is an inverse-CDF sampler (cumulative sum of the probabilities compared against uniform draws), and then scatter-adds the drawn category indices into counts. A wording along the lines of "which draws `total_count` category indices with :func:`~numpyro.distributions.util.categorical` (an inverse-CDF sampler over the probabilities) and accumulates them into counts" would be great.

Once those two are updated, this is good to go. Thanks again!

juanitorduz (changes requested):
small improvements needed

Yi-111-a:
Both fixes in 7747b84:

1. `MultinomialLogits.sample` now says the trials run over `probs`, the softmax of `logits` (matching the implementation, which passes `self.probs`).
2. Both `sample` docstrings now describe `~numpyro.distributions.util.categorical` as an inverse-CDF sampler that draws `total_count` category indices and accumulates them into counts, instead of attributing it to `jax.random.categorical`.

Thanks for the careful read!

Qazalbash (approved):

PR #2279, opened 2026-09-20
doc(gh-2187): add mathematical documentation to Poisson distribution

cnYui:
Addresses part of #2187 by expanding the documentation of the `Poisson`
distribution with mathematical explanations alongside implementation details,
following the style of previously merged distribution-documentation PRs (e.g.
#2185).

## Changes

- **Class docstring**: state the support, use the standard `\lambda` rate
  notation for the PMF, and note the defining property that the mean and
  variance both equal the rate.
- **Method docstrings** (math matched to the existing implementation):
  - `sample` – documents sampling via `jax.random.poisson`.
  - `log_prob` – gives the log-PMF `k log λ − log(k!) − λ`, notes the
    `log Γ(k+1)` evaluation and the `is_sparse` fast path.
  - `mean` / `variance` – `E[X] = λ`, `Var(X) = λ`.
  - `cdf` – documents the CDF as the regularized upper incomplete gamma
    function `Q(⌊x⌋+1, λ)`, matching the `gammaincc` call.

Documentation only; there are no functional changes.

## Verification

- `ruff check numpyro/distributions/discrete.py` – all checks passed
- `ruff format --check numpyro/distributions/discrete.py` – already formatted
- `python -m py_compile numpyro/distributions/discrete.py` – OK


numpyro/distributions/discrete.py
```diff
@@ -1564,12 +1570,34 @@ def __init__(
     def sample(
         self, key: Optional[jax.Array], sample_shape: tuple[int, ...] = ()
     ) -> Array:
+        r"""Draw samples from the Poisson distribution using
+        :func:`jax.random.poisson`.
```
Qazalbash: Please update with the suggestion.
```suggestion
        :func:`~jax.random.poisson`.
```

Qazalbash (approved):
Thanks @cnYui, please update the mentioned changes.

cnYui:
Thanks for the review @Qazalbash! Applied the suggestion in 7115f43 — the docstring now uses `:func:`~jax.random.poisson`` so the rendered cross-reference shows just `poisson` instead of the full dotted path.

PR #2280, opened 2026-09-20
doc: add CLAUDE.md for coding agents

Qazalbash:
## Summary

Adds a `CLAUDE.md` at the repo root. A growing number of NumPyro contributors and collaborators work on the codebase with coding agents (Claude Code and similar tools). Those agents start every session without any project context, so they tend to rediscover the same things each time — or get them wrong. This file gives them a concise, checked-in briefing, which should translate into PRs that pass CI sooner and need less back-and-forth in review.

It is documentation only; no code, config, or CI changes.

## What the file covers

- **Environment** — dev dependencies live in `uv` dependency groups; mirrors the `uv sync --extra cpu --group ...` invocation CI uses.
- **Commands** — `make` targets, running a single test, and the CI variants (`JAX_ENABLE_X64`, multi-device `XLA_FLAGS`, custom PRNG, tracer-leak checks, the examples shard).
- **Things that fail CI** — pytest treats warnings as errors, license headers, the selective `ANN`/`ty` coverage lists in `pyproject.toml`, the custom isort JAX section, extra pre-commit hooks, and the `conftest.py` live-arrays assertion.
- **Architecture** — the message/handler stack behind the primitives, distributions as JAX pytrees (data vs. aux fields, the `biject_to` registry), a checklist of the places a new distribution has to touch, how `infer/util.py` connects models to MCMC kernels and SVI, enumeration via `funsor`, and the `contrib` integrations.

Names referenced in the file (handlers, kernels, ELBOs, util helpers, test registries) were checked against the source.

## Notes for reviewers

- The file points out that the `pip install -e '.[dev,...]'` line in `Makefile` / `CONTRIBUTING.md` is stale now that dev dependencies are dependency groups rather than extras. Those files are intentionally left untouched here; happy to fix them in a follow-up.
- The content is tool-agnostic in substance. If maintainers would prefer an `AGENTS.md` (read by other agents) alongside or instead, that is an easy change.


fehiepsi (approved):
Thanks @Qazalbash!

fehiepsi:
Can we use a different name? Like skill.md or something else

juanitorduz:
I suggest using AGENTS.md , this is the recommended way (provider agnostic) 

https://agents.md

Claude users can create CLAUDE.md and put @AGENTS.md and then this will be loaded at start.

fehiepsi:
+1

juanitorduz (approved):

